PR #1115, opened 2023-04-25
8296590: StraightLineTest fails always on Linux and sometimes on other platforms

jaybhaskar:
Issue: The test fails, in case snapshot is taken before web view finish web content load
Solution: Introduce a sleep , before taking snapshot of web view , to test straight line

### DIFF
--- a/tests/system/src/test/java/test/javafx/scene/web/StraightLineTest.java
+++ b/tests/system/src/test/java/test/javafx/scene/web/StraightLineTest.java
@@ -154,6 +154,8 @@ public class StraightLineTest {
         });
 
         assertTrue("Timeout when waiting for focus change ", Util.await(webViewStateLatch));
+        //introduce sleep , so that web contents would be loaded , then take snapshot for testing
+        Util.sleep(1000);
 
         Util.runAndWait(() -> {
             WritableImage snapshot = straightLineTestApp.primaryStage.getScene().snapshot(null);


johanvos:
We have seen similar issues with snapshots containing WebView.
I agree that adding a sleep fixes the issue here, but it would be great if we can have a clear path that indicates when a snapshot can successfully be taken.

jaybhaskar:
Tested on Linux, There was an issue with the HTML source, the div text was wrapping, due to which the underline property becomes wrapped. Like 
<div>
UNDERLINE TEST
_______________
</div>

becomes 
<div>
UNDERLINE 
__________
TEST
_____
</div>

So , it needs to be nowarp.

tests/system/src/test/java/test/javafx/scene/web/StraightLineTest.java
```diff
@@ -108,7 +108,6 @@ public void setupTestObjects() {
 
     @Test public void testLine() {
         // JDK-8296590
```
kevinrushforth, line 110: You need to remove this comment as well.

kevinrushforth (commented):
The fix looks good, and works for me on Linux. By way of cleanup, please remove the comment with the Bug ID (see inline).

Minor: I think you can also remove the now-unused import of `org.junit.Assume.assumeFalse`.

kevinrushforth (approved):

kevinrushforth:
> We have seen similar issues with snapshots containing WebView. I agree that adding a sleep fixes the issue here, but it would be great if we can have a clear path that indicates when a snapshot can successfully be taken.

Yes, we should consider an RFE to provide some sort of indication of when a page has been rendered.

PR #1111, opened 2023-04-22
Feature/space distributor

hjohn:
Adds a `SpaceDistributor` class that, given a set of spaces bounded by their minimum and maximum values, can calculate how to best distribute a given space over them.

---------
### Progress
- [ ] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [ ] Commit message must refer to an issue

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1111/head:pull/1111` \
`$ git checkout pull/1111`

Update a local copy of the PR: \
`$ git checkout pull/1111` \
`$ git pull https://git.openjdk.org/jfx.git pull/1111/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 1111`

View PR using the GUI difftool: \
`$ git pr show -t 1111`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jfx/pull/1111.diff">https://git.openjdk.org/jfx/pull/1111.diff</a>

</details>

PR #1118, opened 2023-04-27
JDK-8306990: The guarantees given by Region's floor and ceiling functions should work for larger values

hjohn:
Region has floor and ceiling functions that ensure that calling them twice in a row will yield the same result:

     ceil(x) = ceil(ceil(x))

However, due to use of a constant `EPSILON` which is added/subtracted before doing the rounding, this only works for small numbers (in the range of 0-50 approximately).  For larger values and scales, rounding errors can easily occur.  This is visible as artifacts on screen where controls are a pixel wider than they should be.

The use of the `EPSILON` constant is incorrect, as its value depends on the magnitude of the value in question (as magnitude increases, the fractional precision decreases).

The Math class offers the function `ulp` that should be used here.  It represents the smallest possible change in value for a given double.

Extending the existing test case `snappingASnappedValueGivesTheSameValueTest` to use larger magnitude numbers exposes the problems.

---------
### Progress

### Issue
 * [JDK-8306990](https://bugs.openjdk.org/browse/JDK-8306990): The guarantees given by Region's floor and ceiling functions should work for larger values

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1118/head:pull/1118` \
`$ git checkout pull/1118`

`$ git checkout pull/1118` \
`$ git pull https://git.openjdk.org/jfx.git pull/1118/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1118`

`$ git pr show -t 1118`

</details>

<a href="https://git.openjdk.org/jfx/pull/1118.diff">https://git.openjdk.org/jfx/pull/1118.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1118#issuecomment-1525405225)


kevinrushforth:
At first glance, this fix looks correct to me. We'll want to do a reasonable amount of testing.

/reviewers 2

kevinrushforth:
@andy-goryachev-oracle can you also review this?

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/RegionTest.java
```diff
@@ -1286,20 +1287,20 @@ public void clearDirty() {
         for (double scale : scales) {
             stage.setRenderScaleX(scale);
             for (int j = 0; j < 1000; j++) {
-                double value = new Random().nextDouble() * 100 - 50;
+                double value = new Random().nextDouble() * Integer.MAX_VALUE;
```
andy-goryachev-oracle, line 1290: I'd suggest to set and **print** random seed here and on line 1300.  If the test ever fails with some random value, we should be able to reproduce the issue using that particular seed.

(+ use the same Random instance in the second part of the test)

andy-goryachev-oracle:
Note: this PR came out of review https://github.com/openjdk/jfx/pull/1111

I tried to test *every* double value in the range from 100.0 to 200.0.  In several hours the test did not even reach 101.0, so testing it exhaustively is probably not feasible.  Using 32 bit floats, it ran through ~8M values in the same range almost instantly.  I don't think we need to add these exhaustive tests, at the same time using random values is also of little utility give the large space of all possible doubles (still good to have though).

andy-goryachev-oracle:
For completeness sake:

```
    @Test
    public void testFloat() {
        int ct = 0;
        for(float x = 0.0f; x<Integer.MAX_VALUE; ) {
            t(x, 1.75);
            x = Math.nextUp(x);
            ct++;
        }
        System.out.println("count=" + ct);
    }
    
    private static void t(double x, double scale)
    {
        double x1 = scaledCeil2(x, scale);
        double x2 = scaledCeil2(x1, scale);
        if(x1 != x2) {
            throw new Error("x=" + x + " x1=" + x1 + " x2=" + x2 + " ==" + (x1 == x2));
        }
    }

    private static double scaledCeil2(double value, double scale) {
        double d = value * scale;
        return Math.ceil(d - Math.ulp(d)) / scale;
    }
```

andy-goryachev-oracle (approved):
looks good with minor change requests.  if you agree to make them, i'll re-approve.

andy-goryachev-oracle:
@hjohn could you please merge in the latest master branch?  I'd like to re-run the test suite on windows with fractional scale.  thanks!

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/RegionTest.java
hjohn: I've changed this. It now prints the seed only if there is a failure.

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/RegionTest.java
```diff
@@ -1259,48 +1259,60 @@ public void snappingASnappedValueGivesTheSameValueTest() {
...
-                double value = new Random().nextDouble() * Integer.MAX_VALUE;
+                double value = random.nextDouble() * Integer.MAX_VALUE;
                 double snappedValue = region.snapSizeX(value);
                 double snapOfSnappedValue = region.snapSizeX(snappedValue);
-                assertEquals(snappedValue, snapOfSnappedValue, 0.0);
+                assertEquals(failMessage, snappedValue, snapOfSnappedValue, 0.0);
             }
         }
 
+        random.setSeed(seed);
```
andy-goryachev-oracle, line 1281: it might be better to set the seed only once at the beginning.  otherwise the next 2 tests use the same sequence of doubles due to the seed being the same.  
or is this intentional?

hjohn: What are we trying to achieve here I'm wondering? Because if it is to slowly exhaustively test more and more of the problem space during each build, then I really disagree with the methodology.

We're confident that this is the correct solution at this time.  The test is only there to prevent regressions to the old broken situation. 

If you argued that it's a poor test, I'd agree; it is a poor test, and has been since the beginning; I can trivially pass the test with a code change that it would not detect as faulty (returning a constant, using round, etc. would all pass this test).

andy-goryachev-oracle: Not sure I quite understand what you are asking, so -

1. the test is good, it's a reasonable approach for a situation with a large space.  An exhaustive tests using floats that I ran passes, no need (I think) to run it on every build.  A similar test using doubles is not practical.  So we are good here.

2. printing seed for tests involving Random.  I think this should be a general rule to print the seed because if/when the test actually fails, we can reproduce with that particular seed.

3. resetting seed at the beginning of each test block.  I am not sure that this is a good idea because we are cutting the amount of "randomness" to a third, **unless** this is intentional.  Is it intentional?

hjohn: I'll clarify what I mean.  The purpose of unit tests is to verify code works correctly, by putting it through various scenarios and verifying the results are what is expected. A test should be reliable, isolated, fast and accurate.

- Reliable: when run multiple times, it should either always fail or always pass, the initial state of the machine should not affect the test outcome
- Focused: the goal is to test an isolated unit of code only; if it can't be fully isolated, than assume that any dependencies work as specified
- Fast: test a reasonable subset of values and edge cases, exhaustive testing is impossible and not a goal
- Accurate: the results should be verified to be correct against an independent source (constants, test computed values, etc.)

The existing test is poor in that regard. It runs random tests, which depend on the initial state and timing of the machine, making it not reliable.  It's not accurate because it is verifying results with a call to the function under test; if that function is broken then we'd be using a broken function to verify results.

So, to improve this test case we could:
- Make it reliable: use a fixed seed (use of `Random` with a fixed seed is okay, it's specified exactly)
- Verify the results independently; currently the function could be written as `return 0` and this test would pass

andy-goryachev-oracle: I think the test is good, I just recommend removing the setSeed() in all but the first case.

We cannot test the whole range, than would take too long.  But we can put some effort to test a small subset of random values, which I think also serves a purpose.  Don't forget that these tests are run by multiple agents many times over, to over time the number of tested values increases.

To make the test repeatable, we print the seed.  If it ever fails, we can reproduce the condition.

I am not sure how you arrived at the last two points, but I think we are very close to finishing this PR.  Perhaps @kevinrushforth could take a look?

andy-goryachev-oracle:
> I'm not able to reproduce it.

I am trying to run the code after a clean build, and can't - the MonkeyTester freezes going in an infinite loop in HBox.growOrShrinkAreaWidths() on line 517 when launching to an HBox page.

And when launching to a different page, the layout all gets messed up:

Screenshot 2023-05-01 132041

also, looks like it's entering infinite loop when moving the window from primary monitor on windows 11 (scale=100%) to the secondary (scale=175%)

hjohn:
> > I'm not able to reproduce it.
> 
> I am trying to run the code after a clean build, and can't - the MonkeyTester freezes going in an infinite loop in HBox.growOrShrinkAreaWidths() on line 517 when launching to an HBox page.
> 
> And when launching to a different page, the layout all gets messed up:
> 
> Screenshot 2023-05-01 132041
> 
> also, looks like it's entering infinite loop when moving the window from primary monitor on windows 11 (scale=100%) to the secondary (scale=175%)

Okay, I will investigate, I have an idea what this can be, just very surprised that the rounding fixes would cause any issues in this code.

andy-goryachev-oracle:
> Okay, I will investigate, I have an idea what this can be, just very surprised that the rounding fixes would cause any issues in this code.

Thank you. 
I wonder if other places might also exhibit a similar failure mode.  `scaledCeil` is called by `snapSizeX/snapSizeY` as well as `layoutInArea()` which are ... everywhere.

andy-goryachev-oracle:
The latest commit still completely breaks Monkey Tester layout when a scale other than 100% is used:
The Monkey Tester uses SplitPanes, BorderPanes and VBox:

Screenshot 2023-05-03 123236


(a mirror as a standalone project has been moved to https://github.com/andy-goryachev-oracle/MonkeyTest )

andy-goryachev-oracle:
even 200% breaks the layout.

andy-goryachev-oracle:
I wonder if this change requires HBox/VBox changes as well and therefore cannot be made separately from https://github.com/openjdk/jfx/pull/1111 as we expected earlier...

hjohn:
> I wonder if this change requires HBox/VBox changes as well and therefore cannot be made separately from #1111 as we expected earlier...

@andy-goryachev-oracle thanks for testing this (again).  

I think it was important to track down the root cause here, so I put some time in tracking this down.  Turns out it is a real bug introduced with this tiny code change. The problem was that `Math.ulp` will return infinity when given an infinity, and when you then use that in a subtraction, the result is `Double.NaN` -- this caused controls to go a bit crazy :)

I've moved the math functions to a new class to make them easier to test, but also added an extra test case that uses `Region` directly, just in case.

I've confirmed the Monkey Tester works now, so if you could take a look one more time :-)

andy-goryachev-oracle:
> I think it was important to track down the root cause here

You are so right!  I'll try to take a look today.
Without looking at the code, could we add a test (unless already done) which exercise all the possible corner cases with Double - NaN, Infinities, 0, ulp, 1, ...?

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/layout/ScaledMath.java
```diff
@@ -0,0 +1,79 @@
+package com.sun.javafx.scene.layout;
```
andy-goryachev-oracle, line 1: please add a standard header

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/layout/ScaledMath.java
```diff
@@ -0,0 +1,79 @@
...
+     * The value is floored for a given scale using Math.floor.
+     * When the absolute value of the given value multiplied by the
+     * current scale is less than 10^15, then this method guarantees that:
+     *
+     * <pre>scaledFloor(scaledFloor(value, scale), scale) == scaledFloor(value, scale)</pre>
+     *
+     * The limit is about 10^15 because double values will no longer be able to represent
+     * larger integers with exact precision beyond this limit.
+     *
+     * @param value The value that needs to be floored
```
andy-goryachev-oracle, line 18: is this method only for positive values?

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/layout/ScaledMath.java
```diff
@@ -0,0 +1,79 @@
...
+     * <pre>scaledCeil(scaledCeil(value, scale), scale) == scaledCeil(value, scale)</pre>
+     *
+     * The limit is about 10^15 because double values will no longer be able to represent
+     * larger integers with exact precision beyond this limit.
+     *
+     * @param value The value that needs to be ceiled
+     * @param scale The scale that will be used
+     * @return value ceiled with scale
+     */
+    public static double ceil(double value, double scale) {
```
andy-goryachev-oracle, line 46: same comment about negative values

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/RegionTest.java
```diff
@@ -1250,56 +1250,130 @@ public void clearDirty() {
...
                 double snappedValue = region.snapSizeY(value);
                 double snapOfSnappedValue = region.snapSizeY(snappedValue);
-                assertEquals(snappedValue, snapOfSnappedValue, 1.0e-14);
+                assertEquals(failMessage, snappedValue, snapOfSnappedValue, 0.0);
             }
         }
 
         // test snapPortionX/snapPortionY methods
 
+        random.setSeed(seed);
```
andy-goryachev-oracle, line 1356: I still think we need to remove setSeed() here and below.

andy-goryachev-oracle (commented):
This version does not seem to crash the Monkey Tester, both VBox and HBox continue to work.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/layout/ScaledMath.java
hjohn: No, it can be used for any value.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/layout/ScaledMath.java
hjohn: No, it can be used for any value.

hjohn:
@kevinrushforth I think this one is ready to go, and should help to already solve some of the off-by-one pixel errors for controls that are larger than 100 pixels (in any direction) that make use of the ceiling or floor functions

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/RegionTest.java
```diff
@@ -1250,34 +1250,102 @@ public void clearDirty() {
...
+    @Test
+    public void snappingASnappedValueGivesTheSameValueTest() {
         Stage stage = new Stage();
         Region region = new Region();
         Scene scene = new Scene(region);
         stage.setScene(scene);
 
-        double[] scales = new double[] {1.0, 1.25, 1.5, 1.75, 2.0, 1.374562997};
+        double[] scales = new double[] {1.0, 1.25, 1.5, 1.75, 2.0, 1.374562997, 20.0};
+        Random random = new Random();
```
mstr2, line 1323: In my opinion, using a fixed seed would have been preferable here.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/layout/ScaledMath.java
```diff
@@ -0,0 +1,104 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package com.sun.javafx.scene.layout;
+
+/**
+ * Math functions which deal with rounding scaled values.
+ */
+public class ScaledMath {
```
mstr2, line 31: If you want, you could make this class final and give it a private constructor. But since this is not public API, it probably doesn't matter all that much.

mstr2 (approved):
LGTM

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/RegionTest.java
hjohn: I agree with you, but I'll like to leave it now as-is as it was already discussed, and it won't make much of a difference at this point as there are better tests in place directly on `ScaledMath`.

PR #1120, opened 2023-04-29
8307076: gradle test should always run tests

kevinrushforth:
"gradle test" won't rerun a test that has already passed. This is surprising behavior, especially for UI toolkit tests where there might be many reasons to want to run the tests again, even if none of that tests "inputs" have changed. This is especially bothersome when debugging an intermittent test failure. Unlike rebuilding the JavaFX runtime or the docs, where there is no need to rerun a task that was successful if none of the inputs to that task have changed, I expect gradle to run my test when I tell it to.

This PR will make test tasks always run when specified.

---------
### Progress

### Issue
 * [JDK-8307076](https://bugs.openjdk.org/browse/JDK-8307076): gradle test should always run tests

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1120/head:pull/1120` \
`$ git checkout pull/1120`

`$ git checkout pull/1120` \
`$ git pull https://git.openjdk.org/jfx.git pull/1120/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1120`

`$ git pr show -t 1120`

</details>

<a href="https://git.openjdk.org/jfx/pull/1120.diff">https://git.openjdk.org/jfx/pull/1120.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1120#issuecomment-1528795620)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2006,6 +2006,9 @@ allprojects {
     test {
         useJUnitPlatform();
 
+        // Always run tests
+        outputs.upToDateWhen { false }
+
         executable = JAVA;
         enableAssertions = true;
         testLogging.exceptionFormat = "full";


johanvos (approved):

johanvos:
This is a major enhancement. The non-rerunning tests have been extremely annoying, glad this could be fixed with such a simple PR!

lukostyra:
FYI Gradle 7.6 introduced `--rerun` command which kind of solves that problem. There used to be a `--rerun-tasks` option which invalidates all tasks (including dependencies), but now the process is improved - if you run for example `gradle --info :web:test --rerun` it's going to qualify only the `:web:test` target as out of date, but all dependent tasks will remain up-to-date and will be skipped.

I do, however, agree that this behavior might be more intuitive when it comes to running tests.

kevinrushforth:
> This is a major enhancement. The non-rerunning tests have been extremely annoying

I wanted to do this 10 years ago, but was talked out of it. It recently came up as a source of much wasted time for a developer when tracking down another test bug, so I figured it's past time to reconsider it.

> FYI Gradle 7.6 introduced --rerun command which kind of solves that problem

For rerunning tests, that's not appreciably better than the currently recommended approach of adding the `cleanTest` target: in both cases you have to remember to do it, and can be surprised if you don't.

This RFE has a closed-side part to it, so I'll wait to integrate this PR until the other half is ready (which should be soon).

build.gradle
```diff
@@ -2006,6 +2006,9 @@ allprojects {
     test {
         useJUnitPlatform();
 
+        // Always run tests
+        outputs.upToDateWhen { false }
+
```
andy-goryachev-oracle, line 2011: Should the 'testing' section of
https://wiki.openjdk.org/display/OpenJFX/Building+OpenJFX

be updated after this PR is integrated?

Are there any other wiki pages that might be impacted?

kevinrushforth: Updating the testing section is a good idea. Thanks. I'll do that once I integrate.

I am not aware of any other page that needs to be updated.

PR #1121, opened 2023-05-01
8296919: Make system tests that detect memory leaks more robust by using JMemoryBuddy utility

lukostyra:
Verified these changes on all platforms and saw no abnormalities in test execution.

This change is based on a preliminary patch done by @aghaisas , with some minor changes and the addition of one Leak-related web test. EventListenerLeak test was not touched, as it is a separate manual test app instead of a JUnit test.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3465,6 +3465,7 @@ project(":web") {
         implementation project(":graphics")
         implementation project(":controls")
         implementation project(":media")
+        testImplementation project(":base").sourceSets.test.output
     }
 
     compileJava.dependsOn updateCacheIfNeeded
--- a/modules/javafx.web/src/test/java/test/javafx/scene/web/LeakTest.java
+++ b/modules/javafx.web/src/test/java/test/javafx/scene/web/LeakTest.java
@@ -47,6 +47,7 @@ import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.w3c.dom.NodeList;
 import static org.junit.Assert.*;
+import test.util.memory.JMemoryBuddy;
 
 public class LeakTest extends TestBase {
 
@@ -79,7 +80,7 @@ public class LeakTest extends TestBase {
 
     @Test public void testGarbageCollectability() throws InterruptedException {
         final int count = 3;
-        Reference<?>[] willGC = new Reference[count];
+        WeakReference<?>[] willGC = new WeakReference[count];
 
         submit(() -> {
             WebView webView = new WebView();
@@ -90,19 +91,9 @@ public class LeakTest extends TestBase {
 
         Thread.sleep(SLEEP_TIME);
 
-        for (int i = 0; i < 5; i++) {
-            System.gc();
-
-            if (isAllElementsNull(willGC)) {
-                break;
-            }
-
-            Thread.sleep(SLEEP_TIME);
+        for (WeakReference ref : willGC) {
+            JMemoryBuddy.assertCollectable(ref);
         }
-
-        assertNull("WebView has not been GCed", willGC[0].get());
-        assertNull("WebEngine has not been GCed", willGC[1].get());
-        assertNull("WebPage has not been GCed", willGC[2].get());
     }
 
     private static boolean isAllElementsNull(Reference<?>[] array) {
@@ -116,7 +107,7 @@ public class LeakTest extends TestBase {
 
     @Test public void testJSObjectGarbageCollectability() throws InterruptedException {
         final int count = 10000;
-        Reference<?>[] willGC = new Reference[count];
+        WeakReference<?>[] willGC = new WeakReference[count];
 
         submit(() -> {
             for (int i = 0; i < count; i++) {
@@ -127,17 +118,9 @@ public class LeakTest extends TestBase {
 
         Thread.sleep(SLEEP_TIME);
 
-        for (int i = 0; i < 5; i++) {
-            System.gc();
-
-            if (isAllElementsNull(willGC)) {
-                break;
-            }
-
-            Thread.sleep(SLEEP_TIME);
+        for (WeakReference ref : willGC) {
+            JMemoryBuddy.assertCollectable(ref);
         }
-
-        assertTrue("All JSObjects are GC'ed", isAllElementsNull(willGC));
     }
 
     // JDK-8170938
@@ -177,7 +160,7 @@ public class LeakTest extends TestBase {
     // JDK-8176729
     @Test public void testDOMNodeDisposeCount() throws InterruptedException {
         int count = 7;
-        Reference<?>[] willGC = new Reference[count];
+        WeakReference<?>[] willGC = new WeakReference[count];
         final String html =
                 "<html>\n" +
                 "<head></head>\n" +
@@ -237,19 +220,8 @@ public class LeakTest extends TestBase {
         });
 
         Thread.sleep(SLEEP_TIME);
-
-        for (int i = 0; i < 5; i++) {
-            System.gc();
-
-            if (isAllElementsNull(willGC)) {
-                break;
-            }
-
-            Thread.sleep(SLEEP_TIME);
+        for (WeakReference ref : willGC) {
+            JMemoryBuddy.assertCollectable(ref);
         }
-
-        // Give disposer a chance to run
-        Thread.sleep(SLEEP_TIME);
-        assertEquals("NodeImpl HashCount after dispose", initialHashCount, NodeImplShim.test_getHashCount());
     }
 }
--- a/tests/system/src/test/java/test/javafx/embed/swing/SwingNodeDnDMemoryLeakTest.java
+++ b/tests/system/src/test/java/test/javafx/embed/swing/SwingNodeDnDMemoryLeakTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2019, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -33,6 +33,7 @@ import java.util.ArrayList;
 import java.util.concurrent.CountDownLatch;
 
 import javax.swing.JLabel;
+import java.lang.reflect.InvocationTargetException;
 
 import javafx.application.Application;
 import javafx.embed.swing.SwingNode;
@@ -45,12 +46,12 @@ import org.junit.BeforeClass;
 import org.junit.Test;
 
 import test.util.Util;
+import test.util.memory.JMemoryBuddy;
 
 public class SwingNodeDnDMemoryLeakTest {
 
     final static int TOTAL_SWINGNODE = 10;
     static CountDownLatch launchLatch = new CountDownLatch(1);
-    final static int GC_ATTEMPTS = 10;
     ArrayList<WeakReference<SwingNode>> weakRefArrSN =
                                       new ArrayList(TOTAL_SWINGNODE);
 
@@ -65,12 +66,14 @@ public class SwingNodeDnDMemoryLeakTest {
     }
 
     @Test
-    public void testSwingNodeMemoryLeak() {
+    public void testSwingNodeMemoryLeak() throws InvocationTargetException, InterruptedException {
         Util.runAndWait(() -> {
             testSwingNodeObjectsInStage();
         });
-        attemptGCSwingNode();
-        assertEquals(TOTAL_SWINGNODE, getCleanedUpSwingNodeCount());
+
+        for (WeakReference<SwingNode> ref : weakRefArrSN) {
+            JMemoryBuddy.assertCollectable(ref);
+        }
     }
 
     private void testSwingNodeObjectsInStage() {
@@ -108,21 +111,6 @@ public class SwingNodeDnDMemoryLeakTest {
             if (tempStage[i] != null) {
                 tempStage[i].close();
                 tempStage[i] = null;
-            }
-        }
-    }
-
-    private void attemptGCSwingNode() {
-        // Attempt gc GC_ATTEMPTS times
-        for (int i = 0; i < GC_ATTEMPTS; i++) {
-            System.gc();
-            if (getCleanedUpSwingNodeCount() == TOTAL_SWINGNODE) {
-                break;
-            }
-            try {
-                Thread.sleep(500);
-            } catch (InterruptedException e) {
-                System.err.println("InterruptedException occurred during Thread.sleep()");
             }
         }
     }
--- a/tests/system/src/test/java/test/javafx/embed/swing/SwingNodeMemoryLeakTest.java
+++ b/tests/system/src/test/java/test/javafx/embed/swing/SwingNodeMemoryLeakTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -29,10 +29,12 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assume.assumeTrue;
 
 import java.lang.ref.WeakReference;
+import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.concurrent.CountDownLatch;
 
 import javax.swing.JLabel;
+import javax.swing.SwingUtilities;
 
 import javafx.application.Application;
 import javafx.embed.swing.SwingNode;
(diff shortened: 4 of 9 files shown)


kevinrushforth:
> EventListenerLeak test was not touched, as it is a separate manual test app instead of a JUnit test.

In addition to the manual test, the `javafx.web` module also has an automated [EventListenerLeakTest](https://github.com/openjdk/jfx/blob/fbee1d2201270e7b3beb054ecb9fdf65ae11c43d/modules/javafx.web/src/test/java/test/javafx/scene/web/EventListenerLeakTest.java) that could benefit from using JMemoryBuddy, but it would not be a trivial change, so it might be better to file a follow-on bug for that.

build.gradle
```diff
@@ -3465,6 +3465,7 @@ project(":web") {
         implementation project(":graphics")
         implementation project(":controls")
         implementation project(":media")
+        testImplementation project(":base").sourceSets.test.output
```
andy-goryachev-oracle, line 3468: does the header year need to be changed?

modules/javafx.web/src/test/java/test/javafx/scene/web/LeakTest.java
```diff
@@ -47,6 +47,7 @@
 import org.w3c.dom.Element;
 import org.w3c.dom.NodeList;
 import static org.junit.Assert.*;
+import test.util.memory.JMemoryBuddy;
```
andy-goryachev-oracle, line 50: same comment - year in the header?

tests/system/src/test/java/test/javafx/embed/swing/SwingNodeDnDMemoryLeakTest.java
```diff
@@ -65,12 +66,14 @@ public static void teardownOnce() {
...
     @Test
-    public void testSwingNodeMemoryLeak() {
+    public void testSwingNodeMemoryLeak() throws InvocationTargetException, InterruptedException {
         Util.runAndWait(() -> {
             testSwingNodeObjectsInStage();
         });
-        attemptGCSwingNode();
-        assertEquals(TOTAL_SWINGNODE, getCleanedUpSwingNodeCount());
+
+        for (WeakReference<SwingNode> ref : weakRefArrSN) {
```
andy-goryachev-oracle, line 74: I wonder if we should add a utility method to JMemoryBuddy to operate on an array/collection, so we don't have to repeat this code again and again?

tests/system/src/test/java/test/javafx/embed/swing/SwingNodeMemoryLeakTest.java
```diff
@@ -69,18 +72,20 @@ public static void teardownOnce() {
...
             testSwingNodeObjectsInStage();
         });
-        attemptGCSwingNode();
-        assertEquals(TOTAL_SWINGNODE, getCleanedUpSwingNodeCount());
 
-        attemptGCJLabel();
-        assertEquals(TOTAL_SWINGNODE, getCleanedUpJLabelCount());
+        SwingUtilities.invokeAndWait(() -> Util.sleep(500));
+
+        for (WeakReference<SwingNode> ref : weakRefArrSN) {
```
andy-goryachev-oracle, line 82: same comment as before about a new utility method

tests/system/src/test/java/test/javafx/scene/control/TabPaneHeaderLeakTest.java
```diff
@@ -92,16 +93,8 @@ public void testTabPaneHeaderLeak() throws Exception {
...
-                break;
-            }
-            Util.sleep(500);
-        }
-        // Ensure that Tab and TextField are GCed.
-        Assert.assertNull("Couldn't collect Tab", tabWeakRef.get());
-        Assert.assertNull("Couldn't collect TextField", textFieldWeakRef.get());
+
+        JMemoryBuddy.assertCollectable(tabWeakRef);
+        JMemoryBuddy.assertCollectable(textFieldWeakRef);
```
andy-goryachev-oracle, line 98: perhaps the utility method operating on an array should accept varargs, like so:

`assertCollectable(WeakRef a, WeakRef ... b)`

(to differentiate it from a single arg method, and to make sure there are 2+ args)

andy-goryachev-oracle (commented):
LG, with minor suggestions.  What do you think?

tests/system/src/test/java/test/javafx/scene/control/TabPaneHeaderLeakTest.java
lukostyra: Done

tests/system/src/test/java/test/javafx/embed/swing/SwingNodeMemoryLeakTest.java
lukostyra: Done

tests/system/src/test/java/test/javafx/embed/swing/SwingNodeDnDMemoryLeakTest.java
lukostyra: Done

modules/javafx.web/src/test/java/test/javafx/scene/web/LeakTest.java
lukostyra: Done

build.gradle
lukostyra: Done

lukostyra:
> > EventListenerLeak test was not touched, as it is a separate manual test app instead of a JUnit test.
> 
> In addition to the manual test, the `javafx.web` module also has an automated [EventListenerLeakTest](https://github.com/openjdk/jfx/blob/fbee1d2201270e7b3beb054ecb9fdf65ae11c43d/modules/javafx.web/src/test/java/test/javafx/scene/web/EventListenerLeakTest.java) that could benefit from using JMemoryBuddy, but it would not be a trivial change, so it might be better to file a follow-on bug for that.

I actually missed that. I looked briefly into it and I agree it is a larger change, I will file a follow-up bug for it after this PR is merged.

modules/javafx.base/src/test/java/test/util/memory/JMemoryBuddy.java
```diff
@@ -91,6 +92,25 @@ public static void assertCollectable(WeakReference weakReference) {
         }
     }
 
+    public static void assertCollectable(WeakReference[] weakReferences) {
+        for (WeakReference wr : weakReferences) {
+            assertCollectable(wr);
+        }
+    }
+
+    public static <T> void assertCollectable(ArrayList<WeakReference<T>> weakReferences) {
```
andy-goryachev-oracle, line 101: Could we change the argument to Collection<...>?

modules/javafx.web/src/test/java/test/javafx/scene/web/LeakTest.java
```diff
@@ -90,19 +91,9 @@ public class LeakTest extends TestBase {
...
-        for (int i = 0; i < 5; i++) {
-            System.gc();
-
-            if (isAllElementsNull(willGC)) {
-                break;
-            }
-
-            Thread.sleep(SLEEP_TIME);
+        for (WeakReference ref : willGC) {
+            JMemoryBuddy.assertCollectable(ref);
```
andy-goryachev-oracle, line 95: should we use newly added methods here?

modules/javafx.web/src/test/java/test/javafx/scene/web/LeakTest.java
```diff
@@ -127,17 +118,9 @@ private static boolean isAllElementsNull(Reference<?>[] array) {
...
```
andy-goryachev-oracle, line 122: ditto, new methods?

modules/javafx.web/src/test/java/test/javafx/scene/web/LeakTest.java
```diff
@@ -237,19 +220,8 @@ private State getLoadState() {
...
```
andy-goryachev-oracle, line 224: and here

modules/javafx.web/src/test/java/test/javafx/scene/web/LeakTest.java
lukostyra: Done + others.

modules/javafx.base/src/test/java/test/util/memory/JMemoryBuddy.java
lukostyra: Done

andy-goryachev-oracle:
The changes look good!
Unfortunately, the project won't build in Eclipse due to

`The type test.util.memory.JMemoryBuddy is not accessible	LeakTest.java	/web/src/test/java/test/javafx/scene/web`

To fix that, could you please modify line 45 of modules/web/.classpath file to this (er, better include the whole file):

``
<classpath>
	<classpathentry kind="src" path="src/main/java"/>
	<classpathentry kind="src" path="src/main/native/Source/WebCore/bindings/java/dom3/java"/>
	<classpathentry kind="src" output="testbin" path="src/shims/java">
		<attributes>
			<attribute name="test" value="true"/>
		</attributes>
	</classpathentry>
	<classpathentry kind="src" output="testbin" path="src/test/java">
		<attributes>
			<attribute name="test" value="true"/>
			<attribute name="optional" value="true"/>
		</attributes>
	</classpathentry>
	<classpathentry kind="src" path="src/main/resources">
		<attributes>
			<attribute name="optional" value="true"/>
		</attributes>
	</classpathentry>
	<classpathentry kind="src" output="testbin" path="src/test/resources">
		<attributes>
			<attribute name="test" value="true"/>
			<attribute name="optional" value="true"/>
		</attributes>
	</classpathentry>
	<classpathentry combineaccessrules="false" kind="src" path="/media">
		<attributes>
			<attribute name="module" value="true"/>
		</attributes>
	</classpathentry>
	<classpathentry combineaccessrules="false" kind="src" path="/controls">
		<attributes>
			<attribute name="module" value="true"/>
		</attributes>
	</classpathentry>
	<classpathentry combineaccessrules="false" kind="src" path="/graphics">
		<attributes>
			<attribute name="module" value="true"/>
		</attributes>
	</classpathentry>
	<classpathentry combineaccessrules="false" kind="src" path="/base">
		<attributes>
			<attribute name="module" value="true"/>
			<attribute name="add-exports" value="javafx.base/com.sun.javafx=javafx.web:javafx.base/test.util.memory=javafx.web"/>
		</attributes>
	</classpathentry>
	<classpathentry kind="con" path="org.eclipse.jdt.junit.JUNIT_CONTAINER/5">
		<attributes>
			<attribute name="test" value="true"/>
		</attributes>
	</classpathentry>
	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER">
		<attributes>
			<attribute name="module" value="true"/>
			<attribute name="add-reads" value="javafx.web=java.management"/>
		</attributes>
	</classpathentry>
	<classpathentry kind="output" path="bin"/>
</classpath>
```

**Review by andy-goryachev-oracle — CHANGES_REQUESTED** [review]
looks good, one .classpath file change is requested.

File: modules/javafx.base/src/test/java/test/util/memory/JMemoryBuddy.java [hunk 10]
```diff
@@ -91,6 +92,25 @@ public static void assertCollectable(WeakReference weakReference) {
         }
     }
 
+    public static void assertCollectable(WeakReference[] weakReferences) {
```
kevinrushforth, line 95: Can you add javadoc comments?

tests/system/src/test/java/test/javafx/embed/swing/SwingNodeMemoryLeakTest.java
```diff
@@ -69,18 +72,15 @@ public static void teardownOnce() {
...
+    public void testSwingNodeMemoryLeak() throws InterruptedException, InvocationTargetException {
         Util.runAndWait(() -> {
             testSwingNodeObjectsInStage();
         });
-        attemptGCSwingNode();
-        assertEquals(TOTAL_SWINGNODE, getCleanedUpSwingNodeCount());
 
-        attemptGCJLabel();
-        assertEquals(TOTAL_SWINGNODE, getCleanedUpJLabelCount());
+        SwingUtilities.invokeAndWait(() -> Util.sleep(500));
```
kevinrushforth, line 80: Sleeping on either the JavaFX Application thread or the Swing EDT (as in this case) is not generally desirable. What is the sleep trying to achieve? Is there a better way, possibly sleeping on the test thread and then doing a `SwingUtilities.invokeAndWait` (maybe with a no-op runnable? not sure)?.

kevinrushforth (commented):
The changes generally look good, although I left a comment about sleeping on the EDT.

tests/system/src/test/java/test/javafx/embed/swing/SwingNodeMemoryLeakTest.java
lukostyra: This is something that stayed in the original patch - I'm no Swing expert so I originally figured I should leave it here. My only guess was that it (among other `Thread.sleep()`-s in this patch) is trying to let the other threads run for a bit and chew through the resources to make sure they're collectable.

Upon closer inspection it looks like JMemoryBuddy is already doing those sleeps, so I'll remove them - tests should work fine without those.

lukostyra:
I removed the sleeps, ran the test multiple times on all platforms and didn't notice any abnormalities.

tests/system/src/test/java/test/javafx/embed/swing/SwingNodeMemoryLeakTest.java
```diff
@@ -77,8 +77,6 @@ public void testSwingNodeMemoryLeak() throws InterruptedException, InvocationTar
             testSwingNodeObjectsInStage();
         });
 
-        SwingUtilities.invokeAndWait(() -> Util.sleep(500));
-
         JMemoryBuddy.assertCollectable(weakRefArrSN);
```
kevinrushforth, line 80: On my Mac (MacBook Pro x64 running macOS 13.2.1) I saw a one-time test failure here. So there might be a need to wait for the EDT to at least do something, possibly followed by a sleep (on the test thread, not on the EDT).

kevinrushforth (commented):
I saw a one-time failure that suggests waiting for the EDT to run something, and possibly a sleep, might still be needed.

tests/system/src/test/java/test/javafx/embed/swing/SwingNodeMemoryLeakTest.java
lukostyra: Interesting, I will add a `runAndWait()` noop and a test thread sleep then.

kevinrushforth: I think you mean `SwingUtilities::invokeAndWait`, right?

Btw, I also got a similar one-time failure in `SwingNodeDnDMemoryLeakTest`, so perhaps the same could be added there?

lukostyra: Yes, I meant `invokeAndWait`. A force of habit from using our `Util` class...

I also added it to `SwingNodeDnDMemoryLeakTest` just to be sure. After a few runs it works fine on my end (but also, it did earlier so I can't really fully test it).

lukostyra:
@andy-goryachev-oracle I'll wait for your last look at it before I integrate it

modules/javafx.base/src/test/java/test/util/memory/JMemoryBuddy.java
```diff
@@ -91,6 +92,38 @@ public static void assertCollectable(WeakReference weakReference) {
         }
     }
 
+    /**
+     * Checks whether an array of WeakReference objects can be collected.
```
andy-goryachev-oracle, line 96: I think it should rather say something like "whether content of all references in the array", rather than array itself.

modules/javafx.base/src/test/java/test/util/memory/JMemoryBuddy.java
```diff
@@ -91,6 +92,38 @@ public static void assertCollectable(WeakReference weakReference) {
...
+     * @param weakReferences WeakReference objects to check.
+     */
+    public static void assertCollectable(WeakReference[] weakReferences) {
+        for (WeakReference wr : weakReferences) {
+            assertCollectable(wr);
+        }
+    }
+
+    /**
+     * Checks whether a Collection of WeakReference objects can be collected.
```
andy-goryachev-oracle, line 106: same comment: "content of all references in a collection"

modules/javafx.base/src/test/java/test/util/memory/JMemoryBuddy.java
```diff
@@ -91,6 +92,38 @@ public static void assertCollectable(WeakReference weakReference) {
...
+     * @param weakReferences WeakReference objects to check.
+     */
+    public static <T> void assertCollectable(Collection<WeakReference<T>> weakReferences) {
+        for (WeakReference<T> wr : weakReferences) {
+            assertCollectable(wr);
+        }
+    }
+
+    /**
+     * Checks whether provided WeakReference objects can be collected.
```
andy-goryachev-oracle, line 116: same comment: "content of references..."

andy-goryachev-oracle (approved):
minor javadoc changes, I'll re-approve if you want to make them.

modules/javafx.base/src/test/java/test/util/memory/JMemoryBuddy.java
lukostyra: Done (+ others)

kevinrushforth (approved):

andy-goryachev-oracle (approved):
LGTM, thanks!

PR #1123, opened 2023-05-03
JDK-8298104: NPE on synchronizeSceneNodes()

buedi:
A null pointer exception occurs on the ScenePulseListener when iterating through the dirty node list.
A null check is needed on the node before calling node.getScene().

The error occurs occasionally and causes the application to crash.
/signed

Issue: [JDK-8298104: NPE on synchronizeSceneNodes()](https://bugs.openjdk.org/browse/JDK-8298104)

---------
### Progress

### Issue
 * [JDK-8298104](https://bugs.openjdk.org/browse/JDK-8298104): NPE on synchronizeSceneNodes() (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1123/head:pull/1123` \
`$ git checkout pull/1123`

`$ git checkout pull/1123` \
`$ git pull https://git.openjdk.org/jfx.git pull/1123/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1123`

`$ git pr show -t 1123`

</details>

<a href="https://git.openjdk.org/jfx/pull/1123.diff">https://git.openjdk.org/jfx/pull/1123.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1123#issuecomment-1533712897)

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
@@ -2480,7 +2480,7 @@ public class Scene implements EventTarget {
                 for (int i = 0 ; i < dirtyNodesSize; ++i) {
                     Node node = dirtyNodes[i];
                     dirtyNodes[i] = null;
-                    if (node.getScene() == Scene.this) {
+                    if (node != null && node.getScene() == Scene.this) {
                             node.syncPeer();
                         }
                     }


buedi:
/signed

[https://oca.opensource.oracle.com/api/v1/oca-requests/12557/documents/12517/download](https://oca.opensource.oracle.com/api/v1/oca-requests/12557/documents/12517/download)

hjohn:
Hi @buedi, I've taken a look at this, but would like to know the root cause. I get the impression that this happens a lot for you, could you share when this happens so I can reproduce this as well?

You could also try this code and report back if you get any of the exceptions I added:

        private void synchronizeSceneNodes() {
            Toolkit.getToolkit().checkFxUserThread();

            Scene.inSynchronizer = true;

            // if dirtyNodes is null then that means this Scene has not yet been
            // synchronized, and so we will simply synchronize every node in the
            // scene and then create the dirty nodes array list
            if (Scene.this.dirtyNodes == null) {
                // must do this recursively
                syncAll(getRoot());
                dirtyNodes = new Node[MIN_DIRTY_CAPACITY];

            } else {
                if (peer == null) {
                    throw new IllegalStateException("peer shouldn't be null here: " + dirtyNodesSize);
                }

                // This is not the first time this scene has been synchronized,
                // so we will only synchronize those nodes that need it
                for (int i = 0 ; i < dirtyNodesSize; ++i) {
                    try {
                        Node node = dirtyNodes[i];
                        dirtyNodes[i] = null;
                        if (node.getScene() == Scene.this) {
                            node.syncPeer();
                        }
                    }
                    catch (Throwable e) {
                        throw new IllegalStateException("exception will break dirtyNodes!", e);
                    }
                }
                dirtyNodesSize = 0;
            }

            if (!Scene.inSynchronizer) {
                throw new IllegalStateException("Synchronizer flag was reset, reentrant call?");
            }

            Scene.inSynchronizer = false;
        }

buedi:
Hello John, this issue occurs very rarely, approximately 1-2 times per month, on systems that run 24/7. However, when it does occur, the JavaFx application gets stuck in an endless loop (--> [logError-JavaFxScene.txt](https://github.com/openjdk/jfx/files/11397119/logError-JavaFxScene.txt)) and cannot recover from it. Currently, I am unable to replicate this phenomenon. I fully agree with you that it's essential to address the problem at its root cause. However, adding an additional check for a null value would at least solve a rare but blocking issue.

sceneError_

hjohn:
> Hello John, this issue occurs very rarely, approximately 1-2 times per month, on systems that run 24/7. However, when it does occur, the JavaFx application gets stuck in an endless loop (--> [logError-JavaFxScene.txt](https://github.com/openjdk/jfx/files/11397119/logError-JavaFxScene.txt)) and cannot recover from it. Currently, I am unable to replicate this phenomenon. I fully agree with you that it's essential to address the problem at its root cause. However, adding an additional check for a null value would at least solve a rare but blocking issue.
> 
> sceneError_

The problem is, this kind of fix is unlikely to be accepted as it hides the root cause of perhaps a much more serious issue. Skipping `null`s for example will mean that some nodes probably didn't get synced in the last run due to an exception or some other kind of bug in the code (I was looking at Node reuse as a possible cause for this problem).

The NPE you show is probably not the first exception that is logged, but the one that is most visible because it keeps being repeated once the dirt nodes list gets into a bad state.

You said you already fixed this yourself, would you be willing to add diagnostics to your build that you could share so the root cause can be tracked down? A few log lines when a bad state is detected that could give us insights into what went wrong could go a long way. I will take another look tonight to see if I can find how this list is getting corrupted; most likely it is an exception thrown by `syncPeer`; if that could be caught and logged it will probably tell us the rest of the story.

hjohn:
I've looked a bit through the JavaFX bugs, and I noticed that many problems involving `synchronizeSceneNodes` are caused by modifying the SceneGraph outside the FX thread which would sometimes throw `ConcurrentModificationException`. As you see this problem only occasionally, it's a pretty good fit.  It might be worthwhile to check your logs for such an exception.

buedi:
Hi John, thank you for this information. I couldn't find any instances of `ConcurrentModificationException`, but about an hour ago before the NPE, we did have an `IllegalStateException`. Could this be the cause of the issue?
image 
Two weeks ago, we fixed the issue by implementing a check to see if the dialog header was being called from a JavaFX thread.
image

hjohn:
> Hi John, thank you for this information. I couldn't find any instances of `ConcurrentModificationException`, but about an hour ago before the NPE, we did have an `IllegalStateException`. 

No, I don't think that is it.  The check there prevented things from going wrong, and only alerted you that you were about to manipulate the SceneGraph on the wrong thread.  These checks are not everywhere though, so there are many places where you can still change the SceneGraph without being alerted.

buedi:
Hi John,

I've been working on moving our project from Java 8 to Java 17 and JavaFX 21 for the past month. But I can't seem to fix a bug that keeps showing up, even after running the applications for over 10 days on JavaFX version 21.0.1.
Now, I'm having trouble compiling javafx-graphics for Windows. Any ideas on how to figure out what's causing the problem?

best regards,
Daniel

image


2023-12-15 11:00:37,026 ERROR [JavaFX Application Thread] c.m.a.e.AdviseUncaughtExceptionHandler [?:?] AdviseUncaughtExceptionHandler uncaughtException detects NullPointerException
2023-12-15 11:00:37,026 ERROR [JavaFX Application Thread] c.m.a.e.AdviseUncaughtExceptionHandler [?:?] NullPointerException => 
java.lang.NullPointerException: Cannot invoke "javafx.scene.Node.getScene()" because "<local2>" is null
                at javafx.scene.Scene$ScenePulseListener.synchronizeSceneNodes(Scene.java:2483)
                at javafx.scene.Scene$ScenePulseListener.pulse(Scene.java:2630)
                at com.sun.javafx.tk.Toolkit.lambda$runPulse$2(Toolkit.java:401)
                at java.base/java.security.AccessController.doPrivileged(AccessController.java:399)
                at com.sun.javafx.tk.Toolkit.runPulse(Toolkit.java:400)
                at com.sun.javafx.tk.Toolkit.firePulse(Toolkit.java:430)
                at com.sun.javafx.tk.quantum.QuantumToolkit.pulse(QuantumToolkit.java:592)
                at com.sun.javafx.tk.quantum.QuantumToolkit.pulse(QuantumToolkit.java:572)
                at com.sun.javafx.tk.quantum.QuantumToolkit.pulseFromQueue(QuantumToolkit.java:565)
                at com.sun.javafx.tk.quantum.QuantumToolkit.lambda$runToolkit$11(QuantumToolkit.java:352)
                at com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:95)
                at com.sun.glass.ui.win.WinApplication._runLoop(Native Method)
                at com.sun.glass.ui.win.WinApplication.lambda$runLoop$3(WinApplication.java:185)
                at java.base/java.lang.Thread.run(Thread.java:840)
2023-12-15 11:00:37,027 ERROR [JavaFX Application Thread] c.m.a.e.AdviseUncaughtExceptionHandler [?:?] Exception 
java.lang.NullPointerException: Cannot invoke "javafx.scene.Node.getScene()" because "<local2>" is null
                at javafx.scene.Scene$ScenePulseListener.synchronizeSceneNodes(Scene.java:2483)
                at javafx.scene.Scene$ScenePulseListener.pulse(Scene.java:2630)
                at com.sun.javafx.tk.Toolkit.lambda$runPulse$2(Toolkit.java:401)
                at java.base/java.security.AccessController.doPrivileged(AccessController.java:399)
                at com.sun.javafx.tk.Toolkit.runPulse(Toolkit.java:400)
                at com.sun.javafx.tk.Toolkit.firePulse(Toolkit.java:430)
                at com.sun.javafx.tk.quantum.QuantumToolkit.pulse(QuantumToolkit.java:592)
                at com.sun.javafx.tk.quantum.QuantumToolkit.pulse(QuantumToolkit.java:572)
                at com.sun.javafx.tk.quantum.QuantumToolkit.pulseFromQueue(QuantumToolkit.java:565)
                at com.sun.javafx.tk.quantum.QuantumToolkit.lambda$runToolkit$11(QuantumToolkit.java:352)
                at com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:95)
                at com.sun.glass.ui.win.WinApplication._runLoop(Native Method)
                at com.sun.glass.ui.win.WinApplication.lambda$runLoop$3(WinApplication.java:185)
                at java.base/java.lang.Thread.run(Thread.java:840)
2023-12-15 11:00:37,029 ERROR [JavaFX Application Thread] c.m.a.e.AdviseUncaughtExceptionHandler [?:?]  BÄÄHHMM, get more informations => java.lang.NullPointerException: Cannot invoke "javafx.scene.Node.getScene()" because "<local2>" is null
                at javafx.scene.Scene$ScenePulseListener.synchronizeSceneNodes(Scene.java:2483)

hjohn:
If you are not using modules, you don't need to compile javafx.graphics to attempt to analyze the problem.  You can include a copy of `Scene` with your sources (making sure that it is in the original package `javafx.scene`) to debug this further.

If you can reproduce this problem in a small program that you can share, we can try and find out why there is a NPE there -- if only a single thread is using the `dirtyNodes` array, I don't think there can be a `null` in there, but perhaps there is a mistake in the code.

As noted before, a possible cause of this problem is accessing FX components that are currently part of the Scene graph on the wrong thread.  This is not allowed, and such accesses must be wrapped in `Platform::runLater`.  I realize that in a large application it might be hard to find what is doing this, but if you have multiple threads, then inspect if any thread is accessing FX components (like `Node`s, `Timeline`s, etc) directly.  Not all such calls will inform you of problems, so assuming that a call is allowed because you didn't get an exception is incorrect.

elfoxus:
I'm interested if you @buedi handled this issue. In my current project I am facing the same problem. It happens hardly ever and only on one machine in my client's environment, so debugging is kinda hard.

PR #1125, opened 2023-05-03
8307208: Add GridPane constructor that accepts hGap and vGap values

Maran23:
This PR adds a new constructor to `GridPane` that accepts a `hgap` and a `vgap`.
As also written in the ticket, this is a small enhancement to improve the creation of a layout involving a `GridPane` without FXML.
In my experience, the `hgap` and the `vgap` are almost always set when creating a layout with the `GridPane`, as a space between the content just looks much better.

This also fits well with other classes like `HBox` and `VBox`, which also provide a constructor that accepts the spacing.

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/layout/GridPane.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/layout/GridPane.java
@@ -735,6 +735,20 @@ public class GridPane extends Pane {
     }
 
     /**
+     * Creates a GridPane layout with the given {@link #hgapProperty() hgap} and {@link #vgapProperty() vgap}.
+     *
+     * @param hgap the width of the horizontal gaps between columns
+     * @param vgap the height of the vertical gaps between rows
+     *
+     * @since 21
+     */
+    public GridPane(double hgap, double vgap) {
+        this();
+        setHgap(hgap);
+        setVgap(vgap);
+    }
+
+    /**
      * The width of the horizontal gaps between columns.
      * @return the width of the horizontal gaps between columns
      */
--- a/modules/javafx.graphics/src/test/java/test/javafx/scene/layout/GridPaneTest.java
+++ b/modules/javafx.graphics/src/test/java/test/javafx/scene/layout/GridPaneTest.java
@@ -3158,4 +3158,15 @@ public class GridPaneTest {
 
         assertEquals(160, gridpane.prefHeight(-1), 1e-100);
     }
+
+    @Test
+    public void testGridPaneHgapVgapConstructor() {
+        assertEquals(0, gridpane.getHgap(), 0);
+        assertEquals(0, gridpane.getVgap(), 0);
+
+        gridpane = new GridPane(3, 8);
+
+        assertEquals(3, gridpane.getHgap(), 0);
+        assertEquals(8, gridpane.getVgap(), 0);
+    }
 }


modules/javafx.graphics/src/main/java/javafx/scene/layout/GridPane.java
```diff
@@ -734,6 +734,20 @@ public GridPane() {
         getChildren().addListener((Observable o) -> requestLayout());
     }
 
+    /**
```
andy-goryachev-oracle, line 737: we probably need to update the year `2011, 2023,` in the header

andy-goryachev-oracle: also in the test.

andy-goryachev-oracle (changes requested):
minor change requested, otherwise looks good!

modules/javafx.graphics/src/main/java/javafx/scene/layout/GridPane.java
```diff
@@ -734,6 +734,20 @@ public GridPane() {
...
+     * Creates a GridPane layout with the given {@link #hgapProperty() hgap} and {@link #vgapProperty() vgap}.
```
nlisker, line 738: "GridPane" should be in `{@code}`.

modules/javafx.graphics/src/main/java/javafx/scene/layout/GridPane.java
```diff
@@ -734,6 +734,20 @@ public GridPane() {
...
+     *
+     * @param hgap the width of the horizontal gaps between columns
+     * @param vgap the height of the vertical gaps between rows
```
nlisker, line 741: I think that either
"the width of the gaps between columns" or
"the size of the horizontal gaps between columns"
otherwise it gives the illusion that there is a height for the horizontal gap.

Still reads fine as it is.

nlisker (commented):
Look simple enough. `FlowPane` and `TilePane` have this constructor as well.

modules/javafx.graphics/src/main/java/javafx/scene/layout/GridPane.java
nlisker: There is a script that does this automatically.

andy-goryachev-oracle:
Not related to these changes specifically, but why does GridPane adds a listener in the constructor
```
getChildren().addListener((Observable o) -> requestLayout());
```
and VBox/HBox do not?

modules/javafx.graphics/src/main/java/javafx/scene/layout/GridPane.java
Maran23: I copied that mostly from the javadoc of `hgapProperty` and `vgapProperty`.
Tbh I like `the size of the horizontal gaps between columns` more as well. But then we are not consistent with the javadoc of the property. I'm fine either way, just that was my intention. :)

modules/javafx.graphics/src/main/java/javafx/scene/layout/GridPane.java
andy-goryachev-oracle: I thought it's only in jdk, not jfx.  @kevinrushforth could you confirm that?

nlisker: We use the script every release.

modules/javafx.graphics/src/main/java/javafx/scene/layout/GridPane.java
nlisker: I prefer the better phrasing over the consistent one :)

Maran23:
> Not related to these changes specifically, but why does GridPane adds a listener in the constructor
> 
> ```
> getChildren().addListener((Observable o) -> requestLayout());
> ```
> 
> and VBox/HBox do not?

I don't think it is necessary. If you actually look at the `children` list it already takes care about layouting stuff and more. So requesting it here for every change seems wrong. But this should be tested carefully.

andy-goryachev-oracle:
> I don't think it is necessary. If you actually look at the `children` list it already takes care about layouting stuff and more. So requesting it here for every change seems wrong. But this should be tested carefully.

that's what I thought too.
just to clarify, that was a question, not a change request.

andy-goryachev-oracle (approved):
looking good 🕶

Maran23:
> that's what I thought too. just to clarify, that was a question, not a change request.

no problem, I asked myself the same. 😄

modules/javafx.graphics/src/main/java/javafx/scene/layout/GridPane.java
kevinrushforth: Yes, we run a copyright update script a couple times a release, so I don't bother updating the copyright year for existing files. It will get picked up soon enough. If you do want to update them, no one will object, but it's optional.

nlisker (approved):
You can create a CSR, Kevin will need to review it.

hjohn (approved):

kevinrushforth (approved):
Looks good. I reviewed the CSR, so you can Finalize it now.

PR #718, opened 2022-01-17
8278924: [Linux] Robot key test can fail if multiple keyboard layouts are installed

beldenfox:
The Robot implementation on Linux did not consult the current layout when mapping from a KeyCode to a hardware code. Internally it retrieved results for all the layouts but just picked the first one it saw leading to random effects. Though not part of the original bug report, the code also ignored the shift level when choosing which result to pick. On a French layout the dollar sign is on two keys (AltGr 4 is the second one) and the code could choose either one. Same is true for pound.

This PR consults the current layout and only on shift level 0 which is the same level used in get_glass_key to figure out which KeyCode to assign when generating a KeyEvent.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [ ] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issue
 * [JDK-8278924](https://bugs.openjdk.org/browse/JDK-8278924): [Linux] Robot key test can fail if multiple keyboard layouts are installed (**Bug** - P4)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/718/head:pull/718` \
`$ git checkout pull/718`

Update a local copy of the PR: \
`$ git checkout pull/718` \
`$ git pull https://git.openjdk.org/jfx.git pull/718/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 718`

View PR using the GUI difftool: \
`$ git pr show -t 718`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jfx/pull/718.diff">https://git.openjdk.org/jfx/pull/718.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/718#issuecomment-1014867700)

PR #694, opened 2021-12-13
8273743: KeyCharacterCombination for "+" does not work on US QWERTY keyboard layout

beldenfox:
The algorithm in `KeyCharacterCombination.match` relies on the call `Toolkit.getKeyCodeForChar` which is difficult to implement correctly. It defies the way most keyboard API’s work and no platform has got it right yet. In particular the Mac and Linux implementations have to resort to a brute-force approach which monitors keystrokes to learn the relationship between keys and characters.

This PR introduces an alternative mechanism which directly asks the platform whether a given key can generate a specific character. It also allows the platform to attach identifying key information to each KeyEvent to make it easier to answer the question (much, much easier).

This is mostly dumb plumbing. On the front-end there’s a new call `View.notifyKeyEx` that takes an additional platform-specific `hardwareCode` parameter. It also returns a boolean indicating whether the event was consumed or not so I can fix JDK-8087863. If you want to follow the path visit the files in this order:

	View.java
	GlassViewEventHandler.java
	TKSceneListener.java
	Scene.java

The `KeyEvent` class has been expanded with an additional `hardwareCode` member that can only be accessed internally. See KeyEvent.java and KeyEventHelper.java.

On the back-end `KeyCharacterCombination.match` calls a new routine `Toolkit.getKeyCanGenerateCharacter` which unpacks the `KeyEvent` information and sends it on to the Application. The default implementation falls back to the old `getKeyCodeForChar` call but platform specific Applications can send it on to the native glass code.

	KeyCharacterCombination.java
	Toolkit.java
	QuantumToolkit.java
	Application.java
	GtkApplication.java

The glass code can use the `hardwareCode` to answer the question directly. It also has enough information to fall back on the old `getKeyCodeForChar` logic while also enabling the keypad (a common complaint is that Ctrl+’+’ only works on the main keyboard and not the keypad, see JDK-8090275).

This PR improves the situation for key events generated by keystrokes. Manually constructed key events won’t work any better or worse than they already do. Based on the bug database I don't think this is an issue.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [ ] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issue
[...]

PR #425, opened 2021-03-12
8150709: Mac OSX and German Keyboard Layout (Y/Z)

beldenfox:
This PR adds code to ensure that KeyCodeCombinations match KeyEvents as expected by more accurately mapping from a Mac key code to a Java key code based on the user’s active keyboard layout (the existing code assumes a US QWERTY layout). The new code first identifies a set of Mac keys which can produce different characters based on the user’s keyboard layout. A Mac key code outside that area is processed exactly as before. For a key inside the layout-sensitive area the code calls UCKeyTranslate to translate the key to an unshifted ASCII character based on the active keyboard and uses that to determine the Java key code.

When performing the reverse mapping for the Robot the code first uses the old QWERTY mapping to find a candidate key. If it lies in the layout-sensitive area the code then scans the entire area calling UCKeyTranslate until it finds a match. If the key lies outside the layout-sensitive area it’s processed exactly as before.

There are multiple duplicates of these bugs logged against Mac applications built with JavaFX.

https://bugs.openjdk.java.net/browse/JDK-8090257 Mac: Inconsistent KeyEvents with alternative keyboard layouts
https://bugs.openjdk.java.net/browse/JDK-8088120 [Accelerator, Mac] CMD + Z accelerator is not working with French keyboard
https://bugs.openjdk.java.net/browse/JDK-8087915 Mac: accelerator doesn't take into account azerty keyboard layout
https://bugs.openjdk.java.net/browse/JDK-8150709 Mac OSX and German Keyboard Layout (Y/Z)

PR #1126, opened 2023-05-04
8089373: Translation from character to key code is not sufficient

beldenfox:
Note: the Java-side changes in this PR are also in #694 which fixes the same issue (and more) on Linux. Unfortunately the Linux Robot code is not working making it difficult to test on that platform (see #718).

KeyCharacterCombinations allow the specification of accelerators based on characters whose KeyCodes vary across keyboard layouts. For example, the + character is on KeyCode.EQUALS on a U.S. English layout, KeyCode.PLUS on a German layout, and KeyCode.DIGIT1 on a Mac Swiss German layout. KeyCharacterCombinations finds the correct KeyCode by calling `Toolkit.getKeyCodeForChar`.

`getKeyCodeForChar` can only return one KeyCode for a given character so it can't easily handle characters which appear in more than one location, like + which is on the main keyboard and the numeric keypad. It's also reliant on KeyCodes which prevents KeyCharacterCombinations from working on keys with no codes (e.g. the base character contains a diacritic). It also relies on the platform to map from a character to a key which is the reverse of how key mapping normally works making it slow and/or imprecise to implement on Mac and Linux (Windows is the only platform with a system call to do this).

This PR introduces a new way for a platform to pass key information to the Java core. `View.notifyKeyEx` takes an additional platform-specific `hardwareCode` which identifies the key and is tracked in a private field in the KeyEvent. This is opt-in; a platform can continue to call the old `View.notifyKey` method and allow the `hardwareCode` to default to -1.

On the back-end `KeyCharacterCombination.match` calls the new routine `Toolkit.getKeyCanGenerateCharacter` which unpacks the KeyEvent information and sends it on to the Application. This is also opt-in; the default implementation falls back to the Application's `getKeyCodeForChar` call. Platforms which call `View.notifyKeyEx` will be handed the `hardwareCode` for the key in addition to the Java KeyCode.

The new `View.notifyKeyEx` returns a boolean indicating whether the event was consumed or not. This plays no role here but will be used later to fix [JDK-8087863](https://bugs.openjdk.org/browse/JDK-8087863).

For testing I've included the manual KeyboardTest app that also appears in PR #425. Tests with keypad combinations should now work.

Note: this PR only fixes Windows. Fixes for Mac and Linux but can't be submitted until #425 and #718 are integrated.

---------
### Progress

### Issues
 * [JDK-8089373](https://bugs.openjdk.org/browse/JDK-8089373): Translation from character to key code is not sufficient (**Bug** - P4)
 * [JDK-8274967](https://bugs.openjdk.org/browse/JDK-8274967): KeyCharacterCombinations for punctuation and symbols fail on non-US keyboards (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1126/head:pull/1126` \
`$ git checkout pull/1126`

`$ git checkout pull/1126` \
`$ git pull https://git.openjdk.org/jfx.git pull/1126/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1126`

`$ git pr show -t 1126`

</details>

<a href="https://git.openjdk.org/jfx/pull/1126.diff">https://git.openjdk.org/jfx/pull/1126.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1126#issuecomment-1535142007)

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinApplication.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinApplication.java
@@ -377,13 +377,13 @@ final class WinApplication extends Application implements InvokeLaterDispatcher.
 
     @Override
     protected int _getKeyCodeForChar(char c) {
-        // This platform has migrated to getKeyCanGenerateCharacter
+        // This platform has migrated to canKeyGenerateCharacter
         // so getKeyCodeForChar will no longer be called.
-        return 0;
+        throw new UnsupportedOperationException("Windows uses canKeyGenerateCharacter");
     }
 
     @Override
-    protected native boolean _getKeyCanGenerateCharacter(int hardwareCode, int vkCode, char c);
+    protected native boolean _canKeyGenerateCharacter(int hardwareCode, int vkCode, char c);
 
     @Override
     protected native int _isKeyLocked(int keyCode);
--- a/modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinApplication.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinApplication.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -376,7 +376,14 @@ final class WinApplication extends Application implements InvokeLaterDispatcher.
     }
 
     @Override
-    protected native int _getKeyCodeForChar(char c);
+    protected int _getKeyCodeForChar(char c) {
+        // This platform has migrated to getKeyCanGenerateCharacter
+        // so getKeyCodeForChar will no longer be called.
+        return 0;
+    }
+
+    @Override
+    protected native boolean _getKeyCanGenerateCharacter(int hardwareCode, int vkCode, char c);
 
     @Override
     protected native int _isKeyLocked(int keyCode);
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/TKSceneListener.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/TKSceneListener.java
@@ -65,7 +65,7 @@ public interface TKSceneListener {
      * Pass a key event to the scene to handle
      *
      * @param keyEvent The key event
-     * @return true iff the event was consumed
+     * @return {@code true} if the event was consumed
      */
     public boolean keyEvent(KeyEvent keyEvent);
 
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/TKSceneListener.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/TKSceneListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2009, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -63,8 +63,11 @@ public interface TKSceneListener {
 
     /**
      * Pass a key event to the scene to handle
+     *
+     * @param keyEvent The key event
+     * @return true iff the event was consumed
      */
-    public void keyEvent(KeyEvent keyEvent);
+    public boolean keyEvent(KeyEvent keyEvent);
 
     /**
      * Pass an input method event to the scene to handle
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/Toolkit.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/Toolkit.java
@@ -703,6 +703,15 @@ public abstract class Toolkit {
                            float dashOffset);
 
     public abstract int getKeyCodeForChar(String character);
+    /**
+     * The default implementation bridges into the existing getKeyCodeForChar call.
+     */
+    public boolean getKeyCanGenerateCharacter(KeyEvent event, String character) {
+        if (event.getCode() != KeyCode.UNDEFINED) {
+            return getKeyCodeForChar(character) == event.getCode().getCode();
+        }
+        return false;
+    }
     public abstract Dimension2D getBestCursorSize(int preferredWidth, int preferredHeight);
     public abstract int getMaximumCursorColors();
     public abstract PathElement[] convertShapeToFXPath(Object shape);
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/Toolkit.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/Toolkit.java
@@ -706,7 +706,7 @@ public abstract class Toolkit {
     /**
      * The default implementation bridges into the existing getKeyCodeForChar call.
      */
-    public boolean getKeyCanGenerateCharacter(KeyEvent event, String character) {
+    public boolean canKeyGenerateCharacter(KeyEvent event, String character) {
         if (event.getCode() != KeyCode.UNDEFINED) {
             return getKeyCodeForChar(character) == event.getCode().getCode();
         }
--- a/modules/javafx.graphics/src/main/java/javafx/scene/input/KeyEvent.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/input/KeyEvent.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -26,6 +26,8 @@
 package javafx.scene.input;
 
 import com.sun.javafx.tk.Toolkit;
+import com.sun.javafx.scene.input.KeyEventHelper;
+
 import javafx.beans.NamedArg;
 import javafx.event.EventTarget;
 import javafx.event.EventType;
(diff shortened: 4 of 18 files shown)


mstr2:
I'm not quite sold on having `notifyKey` and `notifyKeyEx` be two separate methods. Why not just have one? Not changing the existing call sites doesn't seem to be a sufficient reason to expand the toolkit API surface.

kevinrushforth:
This will need careful review and lots of testing.

/reviewers 2

beldenfox:
> I'm not quite sold on having `notifyKey` and `notifyKeyEx` be two separate methods. Why not just have one? Not changing the existing call sites doesn't seem to be a sufficient reason to expand the toolkit API surface.

The long term goal is to have one method. Eventually the Mac and Linux platforms will transition over, PR #694 has already been entered for Linux and bugs like [JDK-8087863](https://bugs.openjdk.org/browse/JDK-8087863) require the new method on the Mac. Once the transition is done I'll enter a bug to remove the old method and rename the new one.

I thought it would be easiest to review and test one platform at a time so I kept `notifyView` around as a compatibility method. It just calls into `notifyViewEx` with a `hardwareCode` of -1.

modules/javafx.graphics/src/main/java/javafx/scene/input/KeyEvent.java
```diff
@@ -378,6 +382,32 @@ public EventType<KeyEvent> getEventType() {
         return (EventType<KeyEvent>) super.getEventType();
     }
 
+    /**
+     * The hardware key code which is private to the implementation.
+     */
+    private int hardwareCode;
```
hjohn, line 388: Does it need to be private? Events are public, and I think it should be possible to make your own events that act exactly like an event created by the system, which would preclude hidden variables.

modules/javafx.graphics/src/main/native-glass/win/KeyTable.cpp
```diff
@@ -343,21 +343,46 @@ BOOL IsExtendedKey(UINT vkey) {
...
+    if (vkCode == com_sun_glass_events_KeyEvent_VK_DELETE) {
+        return character == 0x7F;
+    }
+
+    // This will match keys on the numeric keypad as well as characters on the
+    // main keyboard that don't require modifiers
+    if (hardwareCode >= 0) {
+        UINT keyChar = ::MapVirtualKeyEx(UINT(hardwareCode), 2, layout);
+        // Filter out dead keys
+        BOOL isDead = (keyChar & 0x80000000) != 0;
```
hjohn, line 366: Wouldn't `hardwareCode >= 0` already filter out dead keys since when the highest bit is set the value is negative?

tests/manual/events/KeyboardTest.java
```diff
@@ -0,0 +1,855 @@
...
+        }
+    }
+
+    /**
+     * List of keys to test for one layout
+     */
+    @SuppressWarnings("serial")
+    static private class KeyList extends ArrayList<KeyData> {}
+
+    static private class KeyListBuilder {
```
hjohn, line 134: minor: standard keyword order is `private static class`

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/TKSceneListener.java
```diff
@@ -63,8 +63,11 @@ public void mouseEvent(EventType<MouseEvent> type, double x, double y, double sc
 
     /**
      * Pass a key event to the scene to handle
+     *
+     * @param keyEvent The key event
+     * @return true iff the event was consumed
```
hjohn, line 68:
```suggestion
     * @return {@code true} if the event was consumed
```

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinApplication.java
```diff
@@ -376,7 +376,14 @@ public String getDataDirectory() {
     }
 
     @Override
-    protected native int _getKeyCodeForChar(char c);
+    protected int _getKeyCodeForChar(char c) {
+        // This platform has migrated to getKeyCanGenerateCharacter
+        // so getKeyCodeForChar will no longer be called.
+        return 0;
```
hjohn, line 382: If you are sure it is never called, then how about going all in and throwing an exception here? Currently it just returns an incorrect result.

hjohn (commented):
Not entirely sure it is good to have a hidden variable in `KeyEvent`, but looks good other than that.

hjohn:
> I'm not quite sold on having `notifyKey` and `notifyKeyEx` be two separate methods. Why not just have one? Not changing the existing call sites doesn't seem to be a sufficient reason to expand the toolkit API surface.

I also think it isn't really needed; Mac and Linux can just ignore the extra parameter for now?

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/Toolkit.java
```diff
@@ -703,6 +703,15 @@ protected static final double clampStopOffset(double offset) {
                            float dashOffset);
 
     public abstract int getKeyCodeForChar(String character);
+    /**
+     * The default implementation bridges into the existing getKeyCodeForChar call.
+     */
+    public boolean getKeyCanGenerateCharacter(KeyEvent event, String character) {
```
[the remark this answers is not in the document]
hjohn: I forgot the `KeyEvent` is needed to extract the hardware code. Depending on whether this hardware code remains a hidden variable this may be for the best (or can pass both `KeyCode` and the hardware code).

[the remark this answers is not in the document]
beldenfox: I'll change the call over to `canKeyGenerateCharacter`, the longer form was starting to annoy me too.

modules/javafx.graphics/src/main/java/javafx/scene/input/KeyEvent.java
beldenfox: At this point I don't have a strong opinion on whether this field is private or not. Originally I wanted it to be private because I thought of it as "whatever magic value makes KeyCharacterCombinations work" and was concerned we might want to tweak that over time. That's less of a concern now that I've prototyped the implementation on three different platforms.

This is largely a policy issue that's above my pay grade. This is an intrinsically platform-specific bit of information we would be exposing to developers.

modules/javafx.graphics/src/main/native-glass/win/KeyTable.cpp
beldenfox: I just realized that I never actually said what `hardwareCode` means on this platform.

A negative number means the `hardwareCode` isn't known (in other words, the KeyEvent wasn't constructed by glass). If it's not negative it's a Windows VK code that identifies the key the user pressed. That code doesn't reflect whether the key is dead or not. Here I'm asking what character that key would generate if pressed without any modifiers and ignoring it if the key is dead.

hjohn: I misread partially, somehow I read `keyChar` as the same as `hardwareCode`. LGTM then.

beldenfox:
> > I'm not quite sold on having `notifyKey` and `notifyKeyEx` be two separate methods. Why not just have one? Not changing the existing call sites doesn't seem to be a sufficient reason to expand the toolkit API surface.
> 
> I also think it isn't really needed; Mac and Linux can just ignore the extra parameter for now?

I now realize that there's no reason for `notifyKey` and `notifyKeyEx` to have unique names since `GetMethodID` has no problem distinguishing between Java overloads. I'll be tweaking this PR to rename `notifyKeyEx` to `notifyKey` so I won't have to get rid of the clunky name in a separate pass.

Mac and Linux will migrate naturally to the new version of `notifyKey` as I fix bugs related to accelerator handling. The remaining clients would be iOS and Monocle. If we're still determined to stamp out the older version it would make sense to tweak iOS at the same time as the Mac since that's where the cross-compilation would happen.

Both versions of `notifyKey` are one-liners that pass their parameters on to the same internal routine (the older version just defaults the `hardwareCode`). I'm still not convinced that it's worth the effort to mess with iOS and Monocle to make the older version go away.

beldenfox:
I added JDK-8274967 as an issue this PR fixes. Support for `Ctrl+'+'` and `Ctrl+'-'` is erratic on Windows without this PR. Currently the platform code finds the correct key but fails to generate the correct Java KeyCode (it uses a hard-coded table that may not correspond to the current layout). With this PR it just compares the platform key code it found to the `hardwareCode` passed in.

(BTW, bugs like this are under-represented in the Java bug tracking system. I was looking through the issues list for a JavaFX desktop and saw at least four bugs related to KeyCharacterCombinations spread across all three platforms. App developers have a hard time sorting this out since the behavior varies so much between platforms and layouts and the specific character being targeted.)

andy-goryachev-oracle:
> App developers have a hard time sorting this out

There was a discussion on the mailing list about adding a doc/ folder with markdown  informational notes.  (Build instructions and snapping rules were the prime candidates for such notes).

If you could write such an explanatory note, we could put it there (there is no JBS issue for the doc/ folder, I'll create one).

Another suggestion is that we could create an umbrella task linking all related issues, as we did, for example with skin changing [JDK-8241364](https://bugs.openjdk.org/browse/JDK-8241364).

What do you think?

beldenfox:
In the mailing list discussion the doc/ folder was intended for internal JavaFX developers and not for users of JavaFX. I was planning on updating the public JavaDocs but that means spreading the information out across the KeyCode, KeyCombination, KeyCharacterCombination, and KeyCodeCombination classes. A short overview document would be nice to have but would that be suitable for the proposed doc/ folder?

I think what most developers need is a) for us to fix our bugs and b) to know that KeyCharacterCombinations should be used for accelerators involving symbols and punctuation and KeyCodeCombinations should be used for everything else. I should probably add that to the KeyCombination documentation sooner rather than later.

modules/javafx.graphics/src/main/java/javafx/scene/input/KeyEvent.java
beldenfox: A KeyEvent that doesn’t originate from Glass will not have a hardwareCode but will have a KeyCode. When matching it against a char combo the platform can just invoke the old getKeyCodeForChar logic and compare the result against the KeyEvent’s KeyCode. This is the same test that used to happen in the Java core so there should be no change in behavior.

Unfortunately the current behavior is all over the map. On the Mac KeyCharacterCombinations don’t work when JavaFX is embedded in a JFXPanel (long story). On Windows and Linux they do work. I could make char combos work on the Mac but that would just lead to other bugs because AWT on the Mac does a poor job of assigning KeyCodes for punctuation and symbol keys. I will take a look at the VKFX skin but I also doubt it assigns KeyCodes “correctly” for char combos, it just doesn’t have enough information to do the job right.

(In general we should only match KeyCharacterCombinations against KeyEvents that Glass generates. Otherwise we’re depending on the creator of the KeyEvent to assign the same code that Glass would assign to that key. That’s a tall order.)

So there’s nothing in this PR that would make matters worse but given the current behavior I don’t know how to write tests to ensure that.

PR #1129, opened 2023-05-05
8307538: Memory leak in TreeTableView when calling refresh

andy-goryachev-oracle:
Fixed a memory leak in TreeTableView by using WeakInvalidationListener (which is the right approach in this particular situation) - the leak is specific to TreeTableRowSkin.

Added a unit test.

Added Refresh buttons and Skin menu to the Monkey Tester (will expand these to other controls as a part next MT update).

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -28,7 +28,7 @@ package javafx.scene.control.skin;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-
+import javafx.beans.WeakInvalidationListener;
 import javafx.beans.property.DoubleProperty;
 import javafx.beans.property.ObjectProperty;
 import javafx.collections.FXCollections;
@@ -48,8 +48,6 @@ import javafx.scene.control.TreeTableColumn;
 import javafx.scene.control.TreeTablePosition;
 import javafx.scene.control.TreeTableRow;
 import javafx.scene.control.TreeTableView;
-
-import com.sun.javafx.scene.control.ListenerHelper;
 import com.sun.javafx.scene.control.behavior.BehaviorBase;
 import com.sun.javafx.scene.control.behavior.TreeTableRowBehavior;
 
@@ -98,46 +96,43 @@ public class TreeTableRowSkin<T> extends TableRowSkinBase<TreeItem<T>, TreeTable
 
         updateTreeItem();
 
-        ListenerHelper lh = ListenerHelper.get(this);
-
-        lh.addChangeListener(control.indexProperty(), (ev) -> {
+        control.indexProperty().addListener(new WeakInvalidationListener((x) -> {
             updateCells = true;
-        });
+        }));
 
-        lh.addChangeListener(control.treeItemProperty(), (ev) -> {
+        control.treeItemProperty().addListener(new WeakInvalidationListener((x) -> {
             updateTreeItem();
             // There used to be an isDirty = true statement here, but this was
             // determined to be unnecessary and led to performance issues such as
             // those detailed in JDK-8143266
-        });
+        }));
 
         setupTreeTableViewListeners();
     }
 
     // FIXME: replace listener to fixedCellSize with direct lookup - JDK-8277000
     private void setupTreeTableViewListeners() {
-        ListenerHelper lh = ListenerHelper.get(this);
         TreeTableView<T> treeTableView = getSkinnable().getTreeTableView();
         if (treeTableView == null) {
-            lh.addInvalidationListener(getSkinnable().treeTableViewProperty(), (ev) -> {
+            getSkinnable().treeTableViewProperty().addListener(new WeakInvalidationListener((x) -> {
                 unregisterInvalidationListeners(getSkinnable().treeTableViewProperty());
                 setupTreeTableViewListeners();
-            });
+            }));
         } else {
-            lh.addChangeListener(treeTableView.treeColumnProperty(), (ev) -> {
+            treeTableView.treeColumnProperty().addListener(new WeakInvalidationListener((x) -> {
                 // Fix for RT-27782: Need to set isDirty to true, rather than the
                 // cheaper updateCells, as otherwise the text indentation will not
                 // be recalculated in TreeTableCellSkin.calculateIndentation()
                 isDirty = true;
                 getSkinnable().requestLayout();
-            });
+            }));
 
             DoubleProperty fixedCellSizeProperty = getTreeTableView().fixedCellSizeProperty();
             if (fixedCellSizeProperty != null) {
-                lh.addChangeListener(fixedCellSizeProperty, (ev) -> {
+                fixedCellSizeProperty.addListener(new WeakInvalidationListener((x) -> {
                     fixedCellSize = fixedCellSizeProperty.get();
                     fixedCellSizeEnabled = fixedCellSize > 0;
-                });
+                }));
                 fixedCellSize = fixedCellSizeProperty.get();
                 fixedCellSizeEnabled = fixedCellSize > 0;
 
@@ -147,7 +142,9 @@ public class TreeTableRowSkin<T> extends TableRowSkinBase<TreeItem<T>, TreeTable
                 // be required (because we remove all cells that are not visible).
                 VirtualFlow<TreeTableRow<T>> virtualFlow = getVirtualFlow();
                 if (virtualFlow != null) {
-                    lh.addChangeListener(getVirtualFlow().widthProperty(), (ev) -> treeTableView.requestLayout());
+                    getVirtualFlow().widthProperty().addListener(new WeakInvalidationListener((x) -> {
+                        treeTableView.requestLayout();
+                    }));
                 }
             }
         }
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableRowSkinTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableRowSkinTest.java
@@ -25,7 +25,11 @@
 
 package test.javafx.scene.control.skin;
 
-import com.sun.javafx.tk.Toolkit;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import java.lang.ref.WeakReference;
 import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
 import javafx.geometry.Insets;
@@ -44,14 +48,11 @@ import javafx.scene.control.skin.TreeTableRowSkin;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import com.sun.javafx.tk.Toolkit;
 import test.com.sun.javafx.scene.control.infrastructure.StageLoader;
 import test.com.sun.javafx.scene.control.infrastructure.VirtualFlowTestUtils;
 import test.com.sun.javafx.scene.control.test.Person;
-
-import static org.junit.jupiter.api.Assertions.assertNotEquals;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertEquals;
+import test.util.memory.JMemoryBuddy;
 
 public class TreeTableRowSkinTest {
 
@@ -318,6 +319,20 @@ public class TreeTableRowSkinTest {
     @Test
     public void invisibleColumnsShouldRemoveCorrespondingCellsInRow() {
         invisibleColumnsShouldRemoveCorrespondingCellsInRowImpl();
+    }
+
+    /** TreeTableView.refresh() must release all discarded cells JDK-8307538 */
+    @Test
+    public void cellsMustBeCollectableAfterRefresh() {
+        IndexedCell<?> row = VirtualFlowTestUtils.getCell(treeTableView, 0);
+        assertNotNull(row);
+        WeakReference<Object> ref = new WeakReference<>(row);
+        row = null;
+
+        treeTableView.refresh();
+        Toolkit.getToolkit().firePulse();
+
+        JMemoryBuddy.assertCollectable(ref);
     }
 
     @AfterEach
--- a/tests/manual/monkey/src/com/oracle/tools/fx/monkey/MonkeyTesterApp.java
+++ b/tests/manual/monkey/src/com/oracle/tools/fx/monkey/MonkeyTesterApp.java
@@ -26,17 +26,13 @@ package com.oracle.tools.fx.monkey;
 
 import java.util.Arrays;
 import java.util.Comparator;
-import com.oracle.tools.fx.monkey.pages.DemoPage;
-import com.oracle.tools.fx.monkey.settings.FxSettings;
-import com.oracle.tools.fx.monkey.util.FX;
-import com.oracle.tools.fx.monkey.util.Native2AsciiPane;
-import com.oracle.tools.fx.monkey.util.TestPaneBase;
 import javafx.application.Application;
 import javafx.application.Platform;
 import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
 import javafx.geometry.Insets;
 import javafx.geometry.NodeOrientation;
+import javafx.scene.Node;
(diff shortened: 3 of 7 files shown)


andy-goryachev-oracle:
will separate Monkey Tester changes into a separate PR.

hjohn:
I've seen the problem with Views and Cell creation often enough now that I think we may want to talk about the fundamental problem that is causing these issues.  The lifecycle of a Cell or Row is not the same as the View that contains them. Cells are created on demand, and should either be destroyed or unlinked.

In JavaFX, we have chosen to not have an explicit `destroy` method for cells, but we do have a way to unlink cells.  For a `TreeTableRow` this is a simple as calling `TreeTableRow#updateTreeTableView(null)`. The call already does exactly what you would expect: it unregisters all troublesome listeners (ones that were registered on the view).  The skin also responds to this setting to `null` as it adds listeners to the `treeTableViewProperty` of its skinnable.

In other words, a `TreeTableView` which creates cells on demand with a different lifecycle than itself should also be responsible to clean them up when it no longer needs them -- this is where the real bug is.  It is managing the lifecycle of these cells and should not rely on these cells doing this indirectly by using a lot of weak listeners.  If `TreeTableView` would correctly unlink cells and rows it doesn't need, it would remove the need for **all** weak listeners (and also references) in both `TreeTableRow` and `TreeTableRowSkin`, making their code much simpler, more predictable and easier to test.

Specifically for this PR, only the listeners registered as part of `setupTreeTableViewListeners` need to be weak, the others are not the issue.

andy-goryachev-oracle:
This is a very good suggestion, thanks!

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableRowSkinTest.java
```diff
@@ -208,14 +209,15 @@ public void treeTableRowWithCellSizeShouldHonorPadding() {
...
     @Test
     public void treeTableRowWithFixedCellSizeShouldIgnoreVerticalPadding() {
-        int top = 10;
-        int right = 20;
-        int bottom = 30;
-        int left = 40;
+        int top = 11;
+        int right = 23;
+        int bottom = 37;
+        int left = 49;
```
kevinrushforth, line 215: After the fix and test change, this test now fails for me.

kevinrushforth (commented):
If I run just the `TreeTableRowSkinTest` test class, I now get a test failure:

```
$ gradle --info -PTEST_ONLY=true :controls:test --tests TreeTableRowSkinTest
TreeTableRowSkinTest > treeTableRowWithFixedCellSizeShouldIgnoreVerticalPadding() FAILED
    org.opentest4j.AssertionFailedError: expected: <18.0> but was: <23.0>
        at app//org.junit.jupiter.api.AssertionUtils.failNotEqual(AssertionUtils.java:62)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:81)
        at app//org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1010)
        at app//test.javafx.scene.control.skin.TreeTableRowSkinTest.treeTableRowWithFixedCellSizeShouldIgnoreVerticalPadding(TreeTableRowSkinTest.java:251)
```

Given that the GHA run passes, as does a full controls test run on my system, that suggests something odd is going on with that test, possibly it is affected by tests in another class. If so, then it could fail randomly anyway, since the tests are run in an unspecified (and unpredictable) order.

kevinrushforth:
@aghaisas can you review this?

hjohn:
> For the purposes of fixing the regression bug, I would like to proceed with this fix as is.

I think that's fine, but I do think you should remove the changes that don't contribute to the fix (I'm pretty sure the weak listeners in the constructor are not a problem).

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
```diff
@@ -98,46 +96,45 @@ public TreeTableRowSkin(TreeTableRow<T> control) {
 
         updateTreeItem();
 
-        ListenerHelper lh = ListenerHelper.get(this);
-
-        lh.addChangeListener(control.indexProperty(), (ev) -> {
+        control.indexProperty().addListener(new WeakInvalidationListener((x) -> {
             updateCells = true;
-        });
+        }));
```
hjohn, line 101: I don't think this needed changing, the index property is directly available on `TreeTableRow` and so shouldn't cause GC issues.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
```diff
@@ -98,46 +96,45 @@ public TreeTableRowSkin(TreeTableRow<T> control) {
...
 
-        lh.addChangeListener(control.treeItemProperty(), (ev) -> {
+        control.treeItemProperty().addListener(new WeakInvalidationListener((x) -> {
             updateTreeItem();
             // There used to be an isDirty = true statement here, but this was
             // determined to be unnecessary and led to performance issues such as
             // those detailed in JDK-8143266
-        });
+        }));
```
hjohn, line 108: As above, I don't think this needs to be weak.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
```diff
@@ -332,7 +331,7 @@ public final DoubleProperty indentProperty() {
 
     private void updateTreeItem() {
         unregisterInvalidationListeners(graphicProperty());
-        treeItem = getSkinnable().getTreeItem();
+        treeItem = (getSkinnable() == null) ? null : getSkinnable().getTreeItem();
```
hjohn, line 334: This is not needed if the weak listener change is reverted in the constructor. The cause for it being `null` is that the listener is now no longer removed by `dispose` where before (with `ListenerHelper` or `regsisterXXXListener`) it would have been.

Also, I think when `dispose` is called, this skin should still do its utmost best to remove all listeners immediately.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableRowSkinTest.java
hjohn: I'm not quite seeing why these changes are needed.  How are they related to the memory leak?

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableRowSkinTest.java
```diff
@@ -320,6 +323,20 @@ public void invisibleColumnsShouldRemoveCorrespondingCellsInRow() {
...
+    /** TreeTableView.refresh() must release all discarded cells JDK-8307538 */
+    @Test
+    public void cellsMustBeCollectableAfterRefresh() {
+        IndexedCell<?> row = VirtualFlowTestUtils.getCell(treeTableView, 0);
+        assertNotNull(row);
+        WeakReference<Object> ref = new WeakReference<>(row);
+        row = null;
+
+        treeTableView.refresh();
+        Toolkit.getToolkit().firePulse();
```
hjohn, line 335: I think cells should also be collectable in other situations, like replacing the row factory.  Is it worth adding a test for that?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
andy-goryachev-oracle: The problem is, `dispose()` is never called for cells discarded by the refresh() method.

In the absence of explicit release, these cells are still listening - so using a weak listener in this case is, I think, the right approach.  Adding explicit release per your (absolutely valid) suggestion is a good idea, but much, much more complicated task.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableRowSkinTest.java
andy-goryachev-oracle: this test reflects the exact failure scenario described in the ticket.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
kevinrushforth: Even if you are using weak listeners, `dispose` should remove them.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableRowSkinTest.java
kevinrushforth: They shouldn't be needed. This points to a problem either with the fix, or a preexisting problem with this test.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableRowSkinTest.java
kevinrushforth: If it's easy to add another test for additional missing cases that would provide better coverage for your fix, it seems worth doing.

kevinrushforth (commented):
I left a couple inline comments.

I also wonder whether the `index` and `treeItem` properties need to be weak (I don't think they do).

The failing test is something that will need to be tracked down. If I revert your changes to `treeTableRowWithFixedCellSizeShouldIgnoreVerticalPadding`, the test passes when that class is run by itself, but fails when the entire controls test suite is run. The modified test fails with that class is run by itself, but passes when the entire controls test suite is run.

This suggests a possible problem where the (now weak) listeners are getting GCed sometimes but not others, and that whether or not it is GCed causes differences that are visible to the test. So either the test is relying on an implementation detail, or there is a functional problem caused by the listeners being GCed too early or not early enough (the latter could point to the need to remove them in dispose).

andy-goryachev-oracle:
it looks like the test is unstable - it works fine in Eclipse, but sometimes fails with gradle.  investigating.

kevinrushforth:
> it looks like the test is unstable - it works fine in Eclipse, but sometimes fails with gradle. investigating.

It's definitely affected by GC, so I'd start by looking there. I can change whether the the test passes or fails by changing the heap size or adding a call to gc before each test.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
andy-goryachev-oracle: revered back to using register*Listener, which creates a weak listener and also disconnects them in dispose()

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableRowSkinTest.java
andy-goryachev-oracle: added one more test case.

hjohn (changes requested):
I confirmed the tests fail before the changes in `setupTreeTableViewListeners` and pass with those changes.  Removing the constructor and dispose changes did not fail any tests.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
```diff
@@ -116,43 +109,59 @@ public TreeTableRowSkin(TreeTableRow<T> control) {
...
 
+    private void updateCachedFixedSize() {
+        if (getSkinnable() != null) {
+            TreeTableView<T> t = getSkinnable().getTreeTableView();
+            if (t != null) {
+                DoubleProperty p = t.fixedCellSizeProperty();
+                if (p != null) {
+                    fixedCellSize = p.get();
+                    fixedCellSizeEnabled = fixedCellSize > 0;
+                }
```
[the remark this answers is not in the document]
andy-goryachev-oracle: good point, thanks!

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
```diff
@@ -98,13 +93,11 @@ public TreeTableRowSkin(TreeTableRow<T> control) {
...
-        lh.addChangeListener(control.indexProperty(), (ev) -> {
+        registerChangeListener(control.indexProperty(), (x) -> {
             updateCells = true;
         });
 
-        lh.addChangeListener(control.treeItemProperty(), (ev) -> {
+        registerChangeListener(control.treeItemProperty(), (obs) -> {
             updateTreeItem();
             // There used to be an isDirty = true statement here, but this was
             // determined to be unnecessary and led to performance issues such as
```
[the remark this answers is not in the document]
andy-goryachev-oracle: Even though I don't particularly like register*Listener() because of its asymmetric nature (when it comes to removing listeners), here we do need to create weak listeners that get unregistered upon `dispose()`.  We need weak listeners because TreeTableView does not explicitly "disconnect" unused cells (e.g. refresh()), and we need `dispose()` due to skin life cycle.

So, in this particular case, I think this change should be ok.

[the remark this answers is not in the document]
hjohn: Sorry, I think we may be misunderstanding each other.  I'm specifically talking about the two listeners added in `TreeTableRowSkin`'s constructor on `indexProperty` and `treeItemProperty`.  These are part of the `TreeTableRow` which is also discarded when `refresh` is called.

1) The test passes if these changes are reverted. This is a clear indication that either the test is insufficient, or that your assumption, that these must be weak, is incorrect. If you can construct a test that requires these listeners to be weak, then I think the changes are warranted.

2) At the risk of stating the obvious, the listeners in the constructor added using `ListenerHelper` are also correctly removed in `dispose`. `ListenerHelper` does this in a similar way to the `register` methods.

3) The `indexProperty` and `treeItemProperty` listeners are attached to the `TreeTableRow`, not the `TreeTableView`.  The refresh function replaces the entire row, including the skin. There is no need to use weak listeners for this purpose.  Compare this to the listeners that are attached to the `TreeTableView` or the `VirtualFlow`; these have a much longer lifecycle and can survive multiple refreshes and row factory replacements, and hence should be weak (for now).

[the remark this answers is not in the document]
andy-goryachev-oracle: you are right, of course.  and I should know - ListenerHelper is my baby.  I guess the fear of another regression overtook me.

reverting.

thank you for your patience and persistence!

hjohn (approved):

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
```diff
@@ -116,43 +112,57 @@ public TreeTableRowSkin(TreeTableRow<T> control) {
...
             }
         }
     }
 
+    private void updateCachedFixedSize() {
+        if (getSkinnable() != null) {
+            TreeTableView<T> t = getSkinnable().getTreeTableView();
+            if (t != null) {
+                double sz = t.getFixedCellSize();
+                fixedCellSize = sz;
```
Maran23, line 161: Why you don't just assign the variable directly?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
```diff
@@ -116,43 +112,57 @@ public TreeTableRowSkin(TreeTableRow<T> control) {
...
+                        }
+                    });
                 }
             }
         }
     }
 
+    private void updateCachedFixedSize() {
+        if (getSkinnable() != null) {
+            TreeTableView<T> t = getSkinnable().getTreeTableView();
```
Maran23, line 158: I know this is a short method, but I would rather see a more descriptive variable name here

andy-goryachev-oracle: I prefer not to drag long names if a very descriptive type is right there: `TreeTableView<T> t`

aghaisas (approved):
The fix looks good to me!

kevinrushforth (approved):
Looks good.

@hjohn Did you want to re-review (there was only one cleanup change since you approved)?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
```diff
@@ -116,43 +112,56 @@ public TreeTableRowSkin(TreeTableRow<T> control) {
 
     // FIXME: replace listener to fixedCellSize with direct lookup - JDK-8277000
     private void setupTreeTableViewListeners() {
-        ListenerHelper lh = ListenerHelper.get(this);
         TreeTableView<T> treeTableView = getSkinnable().getTreeTableView();
         if (treeTableView == null) {
-            lh.addInvalidationListener(getSkinnable().treeTableViewProperty(), (ev) -> {
+            registerInvalidationListener(getSkinnable().treeTableViewProperty(), (x) -> {
```
Maran23, line 117: One question here: Why does this prevent the leak but the ListenerHelper does not?

andy-goryachev-oracle: The difference is that registerInvalidationListener() adds a weak listener, while ListenerHelper adds a strong listener.

It is possible to use ListenerHelper here, at the expense of more complicated code since we'd need to explicitly disconnect the listener when tableViewProperty value gets set.  

Another solution would involve adding a method to add a weak listener to the ListenerHelper to avoid explicit cleanup, or

Go back to the original code which used register/unregister*Listener **in this particular case**.

Maran23: Okay, makes sense.

Maran23 (approved):
Looks good to me.

PR #1130, opened 2023-05-09
8278422: Replace use of deprecated single string variant of Runtime.exec method

lukostyra:
Calls of `Runtime.getRuntime().exec()` were changed to `String[]` variant.

Only Windows and macOS parts were affected, tests work good on both platforms.

I looked through the code and didn't find any other cases of `Runtime.getRuntime().exec(String)` being used, so that should cover all possible deprecation warnings for this issue.

---------
### Progress

### Issue
 * ⚠️ Temporary failure when trying to retrieve information on issue `8278422`.

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1130/head:pull/1130` \
`$ git checkout pull/1130`

`$ git checkout pull/1130` \
`$ git pull https://git.openjdk.org/jfx.git pull/1130/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1130`

`$ git pr show -t 1130`

</details>

<a href="https://git.openjdk.org/jfx/pull/1130.diff">https://git.openjdk.org/jfx/pull/1130.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1130#issuecomment-1540218385)

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/application/HostServicesDelegate.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/application/HostServicesDelegate.java
@@ -135,11 +135,16 @@ public abstract class HostServicesDelegate {
             String osName = System.getProperty("os.name");
             try {
                 if (osName.startsWith("Mac OS")) {
-                    Runtime.getRuntime().exec(
-                            "open " + uri);
+                    Runtime.getRuntime().exec(new String[] {
+                        "open",
+                        uri
+                    });
                 } else if (osName.startsWith("Windows")) {
-                    Runtime.getRuntime().exec(
-                            "rundll32 url.dll,FileProtocolHandler " + uri);
+                    Runtime.getRuntime().exec(new String[] {
+                        "rundll32",
+                        "url.dll,FileProtocolHandler",
+                        uri
+                    });
                 } else { //assume Unix or Linux
                     String browser = null;
                     for (String b : browsers) {


modules/javafx.graphics/src/main/java/com/sun/javafx/application/HostServicesDelegate.java
```diff
@@ -135,11 +135,16 @@ public void showDocument(final String uri) {
             String osName = System.getProperty("os.name");
             try {
                 if (osName.startsWith("Mac OS")) {
-                    Runtime.getRuntime().exec(
-                            "open " + uri);
+                    Runtime.getRuntime().exec(new String[] {
```
andy-goryachev-oracle, line 138: not a review, but a general question:
do we have (do we want) an enum identifying the platform?

nlisker: Don't think we have an enum (maybe we should), but `PlatformUtil` already takes care of identifying the OS, also using `System.getProperty("os.name")`. You can use `PlatformUtil.isMac()`. If we make an enum, that's where is should be.

andy-goryachev-oracle: There was a recent PR
https://github.com/openjdk/jdk/pull/13357
adding an enum deep in the jdk bowels

we probably can't use that

kevinrushforth: No, we can't use that. More to the point, it is unrelated to this PR (and thus out of scope).

kevinrushforth: But I do think it would be a good idea to use `PlatformUtil` here, so maybe file a cleanup issue to look for places where we explicitly parse `os.name`, and consider calling the appropriate `PlatformUtil` method instead?

andy-goryachev-oracle: +1 for PlatformUtil.

kevinrushforth (approved):
Looks good.

PR #1009, opened 2023-01-23
8223373: Remove IntelliJ IDEA specific files from the source code repository

tsayao:
This PR does:

- Remove specific Idea files and let it be imported from gradle;
- Adds checkstyle (to use with checkstyle plugin - it will let you know style mistakes);
- Configures auto-format to sun style (with the changes mentioned in [Code Style Rules](https://wiki.openjdk.org/display/OpenJFX/Code+Style+Rules));
- Automatically sets Copyright notice (updates year too);

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issue
 * [JDK-8223373](https://bugs.openjdk.org/browse/JDK-8223373): Remove IntelliJ IDEA specific files from the source code repository

### Reviewers
 * [Marius Hanl](https://openjdk.org/census#mhanl) (@Maran23 - Committer)
 * [Ambarish Rapte](https://openjdk.org/census#arapte) (@arapte - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1009/head:pull/1009` \
`$ git checkout pull/1009`

Update a local copy of the PR: \
`$ git checkout pull/1009` \
`$ git pull https://git.openjdk.org/jfx.git pull/1009/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 1009`

View PR using the GUI difftool: \
`$ git pr show -t 1009`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jfx/pull/1009.diff">https://git.openjdk.org/jfx/pull/1009.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1009#issuecomment-1407752304)

PR #1132, opened 2023-05-10
JDK-8307807 Replace use of System.getProperty("os.name") with PlatformUtil calls

EstelonAgarwaen:
Replaced querying of `System.getProperty("os.name")` to check for OS with the use of the methods provided by `PlatformUtils`

---------
### Progress

### Issue
 * [JDK-8307807](https://bugs.openjdk.org/browse/JDK-8307807): Replace use of System.getProperty("os.name") with PlatformUtil calls

### Reviewers
 * [Lukasz Kostyra](https://openjdk.org/census#lkostyra) (@lukostyra - Author)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1132/head:pull/1132` \
`$ git checkout pull/1132`

`$ git checkout pull/1132` \
`$ git pull https://git.openjdk.org/jfx.git pull/1132/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1132`

`$ git pr show -t 1132`

</details>

<a href="https://git.openjdk.org/jfx/pull/1132.diff">https://git.openjdk.org/jfx/pull/1132.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1132#issuecomment-1542670667)


modules/javafx.media/src/main/java/com/sun/media/jfxmedia/locator/Locator.java
```diff
@@ -358,7 +359,7 @@ public void init() throws URISyntaxException, IOException, FileNotFoundException
             }
 
             // On non-Windows systems, replace "/~/" with home directory path + "/".
-            if (System.getProperty("os.name").toLowerCase().indexOf("win") == -1
+            if (!PlatformUtil.isWindows()
                     && protocol.equals("file")) {
```
nlisker, line 363: Can be on the same line now.

nlisker (commented):
Found an occurrences:
* `com.sun.javafx.application.LauncherImpl` line 386.
* `com.sun.media.jfxmediaimpl.HostUtils` seems to be recreating the `PlatformUtil` API, can probably be replaced. `"os.arch"` could be moved to `PlatformUtil` probably.
* `com.sun.webkit.network.URLLoader` line 244.
* `ClipboardExtImageTest` line 59.
* `BigGlyphIDTest` line 39.
* `INVISIBLE_GLYPH_IDTest` line 39.
* There are occurrences in the apps folder in ensemble. 

Creating an enum for the OS is not a bad idea by the way.

EstelonAgarwaen:
I cannot use the PlatformUtils in the `com.sun.webkit.network.URLLoader` class, as the PlatformUtils are not Exported to the corresponding module.

nlisker:
> I cannot use the PlatformUtils in the `com.sun.webkit.network.URLLoader` class, as the PlatformUtils are not Exported to the corresponding module.

I wonder if this was done on purpose.

EstelonAgarwaen:
> > I cannot use the PlatformUtils in the `com.sun.webkit.network.URLLoader` class, as the PlatformUtils are not Exported to the corresponding module.
> 
> I wonder if this was done on purpose.

i assumed it was.

kevinrushforth:
> > > I cannot use the PlatformUtils in the `com.sun.webkit.network.URLLoader` class, as the PlatformUtils are not Exported to the corresponding module.
> > 
> > 
> > I wonder if this was done on purpose.
> 
> i assumed it was.

In this case, it just wasn't needed at the time `module-info.java` was created. There would be no problem adding `javafx.web` to the list of modules that `com.sun.javafx` is exported to.

EstelonAgarwaen:
> > > > I cannot use the PlatformUtils in the `com.sun.webkit.network.URLLoader` class, as the PlatformUtils are not Exported to the corresponding module.
> > > 
> > > 
> > > I wonder if this was done on purpose.
> > 
> > 
> > i assumed it was.
> 
> In this case, it just wasn't needed at the time `module-info.java` was created. There would be no problem adding `javafx.web` to the list of modules that `com.sun.javafx` is exported to.

then i will do so.

modules/javafx.base/src/main/java/module-info.java
```diff
@@ -50,7 +50,7 @@
         javafx.graphics,
         javafx.fxml,
         javafx.media,
-        javafx.swing;
+        javafx.swing, javafx.web;
```
kevinrushforth, line 53: Minor: can you move `javafx.web` onto its own line to match the formatting in the rest of the file?

.idea/jarRepositories.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kevinrushforth, line 1: This is not relevant to the bug fix. You need to revert the addition of this file.

kevinrushforth (commented):
Looks OK with two comments.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/monocle/LinuxPlatformFactory.java
Glavo, line 1: https://github.com/openjdk/jfx/blob/91ac578372e0676967151315556fa31556927725/modules/javafx.graphics/src/main/java/com/sun/glass/ui/monocle/LinuxPlatformFactory.java#L30-L31

redundant imports

.idea/jarRepositories.xml
Maran23: Note that when this PR https://github.com/openjdk/jfx/pull/1009 is merged, this file is inside the `.gitignore` and mistakes like this can not happen anymore.

modules/javafx.graphics/src/main/java/com/sun/javafx/application/LauncherImpl.java
```diff
@@ -383,7 +384,7 @@ private static Class<?> loadClass(final ModuleAccess mainModule, final String cl
             } catch (ClassNotFoundException | NoClassDefFoundError cnfe) {}
         }
 
-        if (clz == null && System.getProperty("os.name", "").contains("OS X")
+        if (clz == null && PlatformUtil.isMac()
                     && Normalizer.isNormalized(className, Normalizer.Form.NFD)) {
```
nlisker, line 388: Can also be in 1 line.

nlisker (commented):
Looks good. Do you want to replace `HostUtils`'s OS detection with `PlatformUtil` (or were they split on purpose)?

lukostyra (approved):
Looks good, tests are passing as expected on all platforms

modules/javafx.graphics/src/main/java/com/sun/glass/ui/Platform.java
```diff
@@ -62,15 +64,13 @@ else if (userPlatform.equals("ios"))
                 return type;
             }
 
-            String osName = System.getProperty("os.name");
-            String osNameLowerCase = osName.toLowerCase(Locale.ROOT);
-            if (osNameLowerCase.startsWith("mac") || osNameLowerCase.startsWith("darwin")) {
+            if (PlatformUtil.isMac()) {
```
andy-goryachev-oracle, line 67: this change might be ok, but it is not strictly equivalent.
before, MAC path would be selected if os.name lowercase starts from either "mac" or "darwin", after the change, only starts with "Mac".
Do we want to support "darwin"?

andy-goryachev-oracle: I'd still prefer to use PlatformUtil, since this is a cleaner solution that the variety of methods used before:
```
osNameLowerCase.startsWith("wind"
"ios".equals(System.getProperty("os.name")
osName.startsWith("Windows"
osName.startsWith("Mac OS"
("os.name").toLowerCase().indexOf("win")
```

kevinrushforth: I initially had the same thought. What this actually highlights, though, is the danger of having this sort of check scattered in multiple places, but done differently. If it was important that "darwin" be recognized (it isn't), then it would be wrong to have this check recognize it, but all other "isMac" checks not recognize it.

kevinrushforth:
> Looks good. Do you want to replace `HostUtils`'s OS detection with `PlatformUtil` (or were they split on purpose)?

I think this might be a good change, but I'll leave it up to @EstelonAgarwaen . To minimize the intrusion, you could just change the implementation of the four `HostUtils` methods that check `os.name` to call the equivalent `PlatformUtil` methods rather than changing all of the call sites.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/Platform.java
andy-goryachev-oracle: Do we officially support Darwin OS?

andy-goryachev-oracle: Wikipedia says it's currently at the level of macOS Ventura

22.4.0 | March 27, 2023 | macOS 13.3, iOS 16.4, iPadOS 16.4, watchOS 9.4 and tvOS 16.4

https://en.wikipedia.org/wiki/Darwin_(operating_system)#Darwin_20_onwards

lukostyra: Darwin is the kernel name of macOS/OS X, so technically when we speak of Darwin we speak of macOS. My guess for what the goal was here is to support some older naming where `os.name` prop had what `uname` command returns (which returns what kernel reports, which is its name "Darwin"). So, technically, we do support it because we support macOS?

EstelonAgarwaen: Darwin as its own OS stopped existing after 2002 (at least german wilipedia sais so)

kevinrushforth: Apple still calls their macOS Kernel "Darwin", but that's not the important bit. If it was _ever_ reported as the value of `os.name` (I'm not sure it was), it would have been in the pre-Java 6 time frame. It's a long-dead holdover.

andy-goryachev-oracle (approved):
thank you for clarification!

Maran23 (approved):
Looks good to me too.

EstelonAgarwaen:
> > Looks good. Do you want to replace `HostUtils`'s OS detection with `PlatformUtil` (or were they split on purpose)?
> 
> I think this might be a good change, but I'll leave it up to @EstelonAgarwaen . To minimize the intrusion, you could just change the implementation of the four `HostUtils` methods that check `os.name` to call the equivalent `PlatformUtil` methods rather than changing all of the call sites.

Can you open a ticket in the Bugtracker? Ill create a seperate PR for it.

kevinrushforth:
> > > Looks good. Do you want to replace `HostUtils`'s OS detection with `PlatformUtil` (or were they split on purpose)?
> > 
> > 
> > I think this might be a good change, but I'll leave it up to @EstelonAgarwaen . To minimize the intrusion, you could just change the implementation of the four `HostUtils` methods that check `os.name` to call the equivalent `PlatformUtil` methods rather than changing all of the call sites.
> 
> Can you open a ticket in the Bugtracker? Ill create a separate PR for it.

https://bugs.openjdk.org/browse/JDK-8308028

PR #1133, opened 2023-05-11
8307960: Create Table Column PopupMenu lazily

Maran23:
This PR changes the `columnPopupMenu`, so that it is created lazily.

The problem here is, that the `columnPopupMenu` is always initialized and updated via bindings, even if the table menu button is never shown (`setTableMenuButtonVisible(false)`) or the user never clicked on it.
This problem can be solved by creating the `columnPopupMenu` and related bindings when it should be shown the first time.

I also added many tests to ensure that everything still works (there are no tests for that area as of now).

Side note: There are a bunch of tickets with the wish to customize the Popup shown by the table menu button or show it programmatically. This ticket will prepare this, as now all Popup related code is on one place and in the future we can think of implementing a way to override this behaviour in a way that the Popup and all related bindings are never created and therefore do not decrease performance.

---------
### Progress

### Issue
 * [JDK-8307960](https://bugs.openjdk.org/browse/JDK-8307960): Create Table Column PopupMenu lazily

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1133/head:pull/1133` \
`$ git checkout pull/1133`

`$ git checkout pull/1133` \
`$ git pull https://git.openjdk.org/jfx.git pull/1133/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1133`

`$ git pr show -t 1133`

</details>

<a href="https://git.openjdk.org/jfx/pull/1133.diff">https://git.openjdk.org/jfx/pull/1133.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1133#issuecomment-1544581486)

### DIFF
--- a/modules/javafx.controls/src/shims/java/javafx/scene/control/skin/TableHeaderRowShim.java
+++ b/modules/javafx.controls/src/shims/java/javafx/scene/control/skin/TableHeaderRowShim.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2023 Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -24,6 +24,8 @@
  */
 package javafx.scene.control.skin;
 
+import javafx.scene.control.ContextMenu;
+import javafx.scene.layout.Pane;
 import javafx.scene.control.TableColumnBase;
 
 public class TableHeaderRowShim {
@@ -32,4 +34,11 @@ public class TableHeaderRowShim {
         return tr.getColumnHeaderFor(col);
     }
 
+    public static ContextMenu getColumnPopupMenu(TableHeaderRow tableHeaderRow) {
+        return tableHeaderRow.getColumnPopupMenu();
+    }
+
+    public static Pane getCornerRegion(TableHeaderRow tableHeaderRow) {
+        return tableHeaderRow.getCornerRegion();
+    }
 }
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2023, Oracle and/or its affiliates. All rights reserved.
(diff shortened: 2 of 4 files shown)


modules/javafx.controls/src/shims/java/javafx/scene/control/skin/TableHeaderRowShim.java
```diff
@@ -32,4 +34,11 @@ public static TableColumnHeader getColumnHeaderFor(TableHeaderRow tr, final Tabl
         return tr.getColumnHeaderFor(col);
     }
 
+    public static ContextMenu getColumnPopupMenu(TableHeaderRow tableHeaderRow) {
+        return tableHeaderRow.getColumnPopupMenu();
```
lukostyra, line 38: Minor: I would keep `tr` instead of `tableHeaderRow` for consistency with already existing static method.

modules/javafx.controls/src/shims/java/javafx/scene/control/skin/TableHeaderRowShim.java
```diff
@@ -32,4 +34,11 @@ public static TableColumnHeader getColumnHeaderFor(TableHeaderRow tr, final Tabl
...
+    }
+
+    public static Pane getCornerRegion(TableHeaderRow tableHeaderRow) {
+        return tableHeaderRow.getCornerRegion();
```
lukostyra, line 42: As above

lukostyra (commented):
Left a couple very minor style-related comments

modules/javafx.controls/src/shims/java/javafx/scene/control/skin/TableHeaderRowShim.java
andy-goryachev-oracle: I would too - avoids wrapping the line, and there is no need to spell the name for a one line method.

andy-goryachev-oracle:
There is also a similar property in `TreeTableView` which also goes through the same code path in `TableHeaderRow`.  Should we have a separate test or set of tests for TTV?

andy-goryachev-oracle:
Verified that the code works as expected with Tree- and TableView, with no ill effects (on Mac), using 

andy-goryachev-oracle (changes requested):
Would it be possible to add test(s) for TreeTableView?  Otherwise, looks good.
edit: tests are there, I must be blind!

andy-goryachev-oracle (approved):

modules/javafx.controls/src/shims/java/javafx/scene/control/skin/TableHeaderRowShim.java
Maran23: I am more a fan of descriptive variable names. I know this is a one-line method - but on the other hand `tr` is just not a good understandable name, even for a very short method.
Since you read code more often than you write it, I'm for keeping the descriptive name, also in this simple case :)

PR #1134, opened 2023-05-13
JDK-8308028 Replace more uses of System.getProperty("os.name") with PlatformUtil calls

EstelonAgarwaen:
replaced implementation of OS checks in HostUtils with calls of PlatformUtil

---------
### Progress

### Issue
 * [JDK-8308028](https://bugs.openjdk.org/browse/JDK-8308028): Replace more uses of System.getProperty("os.name") with PlatformUtil calls

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1134/head:pull/1134` \
`$ git checkout pull/1134`

`$ git checkout pull/1134` \
`$ git pull https://git.openjdk.org/jfx.git pull/1134/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1134`

`$ git pr show -t 1134`

</details>

<a href="https://git.openjdk.org/jfx/pull/1134.diff">https://git.openjdk.org/jfx/pull/1134.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1134#issuecomment-1546699432)


modules/javafx.media/src/main/java/com/sun/media/jfxmediaimpl/HostUtils.java
```diff
@@ -57,19 +59,19 @@ public static boolean is64Bit() {
...
-        return osName.startsWith("linux");
+        return PlatformUtil.isLinux();
     }
 
     public static boolean isIOS() {
-        return osName.startsWith("ios");
+        return PlatformUtil.isIOS();
     }
 
     /**
```
Maran23, line 77: `isEmbedded` can also be replaced by `PlatformUtil`.

modules/javafx.media/src/main/java/com/sun/media/jfxmediaimpl/HostUtils.java
```diff
@@ -57,19 +59,19 @@ public static boolean is64Bit() {
     }
 
     public static boolean isWindows() {
-        return osName.startsWith("windows");
```
Maran23, line 60: `osName` is unused now

nlisker: `osArch` is also unused. It's just a local variable.

EstelonAgarwaen: @nlisker osArch seems to be used for is64bit still

Maran23: Yes, but what @nlisker means is that `osArch` does not need to be a member variable but only a local one inside the static initializer above.

EstelonAgarwaen:
I have looked at `HostUtils` further. The only method it implements itself is the `is64bit`, which checks whether the architechture is 64 bit. i could not find any uses os this method anywhere. 

i think it may be a valid option to replace calls of `HostUtils` by calls of `PlatformUtil` and move `is64bit` in the `PlatformUtil` as well. It may also be an option to remove it completely.

modules/javafx.media/src/main/java/com/sun/media/jfxmediaimpl/HostUtils.java
```diff
@@ -25,57 +25,53 @@
...
-    private static boolean is64bit = false;
+    private static boolean is64bit;
 
     static {
         @SuppressWarnings("removal")
-        boolean tmp = AccessController.doPrivileged((PrivilegedAction<Boolean>) () -> {
-            osName = System.getProperty("os.name").toLowerCase();
-            osArch = System.getProperty("os.arch").toLowerCase();
+        boolean temp = AccessController.doPrivileged((PrivilegedAction<Boolean>) () -> {
+            String osArch = System.getProperty("os.arch").toLowerCase();
```
Glavo, line 42:
```suggestion
            String osArch = System.getProperty("os.arch").toLowerCase(Locale.ROOT);
```

Glavo:
> I have looked at `HostUtils` further. The only method it implements itself is the `is64bit`, which checks whether the architechture is 64 bit. i could not find any uses os this method anywhere.
> 
> i think it may be a valid option to replace calls of `HostUtils` by calls of `PlatformUtil` and move `is64bit` in the `PlatformUtil` as well. It may also be an option to remove it completely.

Sounds great, I was just about to say the same.

Maran23:
> i think it may be a valid option to replace calls of `HostUtils` by calls of `PlatformUtil` and move `is64bit` in the `PlatformUtil` as well. It may also be an option to remove it completely.

Since `is64bit` is unused, `HostUtils` is basically a delegate for `PlatformUtil`. So we might as well delete it completely and replace all occurrences with `PlatformUtil`. And as written by Kevin in the ticket, this is a valid choice so we might as well do it.

nlisker:
`isEmbedded()` is also unused as far as I can tell.

The question is why this and `is64bit` were created. Before removing unused code I would try to find out if it is *supposed* to be used.

> i think it may be a valid option to replace calls of HostUtils by calls of PlatformUtil and move is64bit in the PlatformUtil as well. It may also be an option to remove it completely.

This is what I would have done. No need to have duplicate functionalities.

EstelonAgarwaen:
Since `HostUtils` is internal API and not used outside of JFX itself, and `is64bit` is not used anywhere i would say its safe to remove. 

From its implementation I saw that it checked whether the operating systems architechture was either intel/amd 64 bit or intel itanium(ia64). In any case of an ARM-Architecture, even on ARM64, the method would have returned false.

nlisker (approved):

Maran23 (approved):
Looks good.

PR #1135, opened 2023-05-13
8091153: Customize the Table Button Menu

Maran23:
This PR implements a way to override the table column menu. 
When the `cornerRegion` is pressed, it will now call the `showColumnMenu` method. This new method is protected and therefore can be overidden by developers. If not overridden, the known default column 'ContextMenu' is created and shown.

This PR also resolves [JDK-8091419 ](https://bugs.openjdk.org/browse/JDK-8091419) (The method `showColumnMenu` can be overridden and made public now)

This PR also helps with [JDK-8092148](https://bugs.openjdk.org/browse/JDK-8092148), but does not address all the points mentioned in the ticket. 
Now that everyone can provide their own implementation, we can think of treating the current implementation of the table column menu as an [MVP](https://en.wikipedia.org/wiki/Minimum_viable_product) and any other requested feature/idea should be fulfilled by providing your own implementaion of the menu.

/issue add JDK-8091419

---------
### Progress
- [x] Change requires CSR request [JDK-8308121](https://bugs.openjdk.org/browse/JDK-8308121) to be approved

### Issues
 * [JDK-8091153](https://bugs.openjdk.org/browse/JDK-8091153): Customize the Table Button Menu
 * [JDK-8091419](https://bugs.openjdk.org/browse/JDK-8091419): TableView: invoke table menu button programmatically
 * [JDK-8308121](https://bugs.openjdk.org/browse/JDK-8308121): Customize the Table Button Menu (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1135/head:pull/1135` \
`$ git checkout pull/1135`

`$ git checkout pull/1135` \
`$ git pull https://git.openjdk.org/jfx.git pull/1135/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1135`

`$ git pr show -t 1135`

</details>

<a href="https://git.openjdk.org/jfx/pull/1135.diff">https://git.openjdk.org/jfx/pull/1135.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1135#issuecomment-1546734028)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
@@ -465,11 +465,11 @@ public class TableHeaderRow extends StackPane {
     }
 
     /**
-     * Shows a menu containing all leaf columns as item.
+     * Shows a menu containing all leaf columns as items.
      * An item can be selected/deselected to make the corresponding column visible/invisible.
      *
-     * @implNote This method can be overridden to create and show a custom menu.
-     * @param mouseEvent the {@link MouseEvent} which was generated when the table menu button was pressed.
+     * @implNote This method can be overridden to create and show a custom menu
+     * @param mouseEvent the {@link MouseEvent} which was generated when the table menu button was pressed
      * @since 21
      */
     protected void showColumnMenu(MouseEvent mouseEvent) {
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
@@ -469,7 +469,7 @@ public class TableHeaderRow extends StackPane {
      * An item can be selected/deselected to make the corresponding column visible/invisible.
      *
      * @implNote This method can be overridden to create and show a custom menu.
-     * @param mouseEvent the {@link MouseEvent} which was generated when the table menu button was pressed
+     * @param mouseEvent the {@code MouseEvent} which was generated when the table menu button was pressed
      * @since 21
      */
     protected void showColumnMenu(MouseEvent mouseEvent) {
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
@@ -465,11 +465,12 @@ public class TableHeaderRow extends StackPane {
     }
 
     /**
-     * Shows a menu with all leaf columns.
-     * These can selected/deselected to make the corresponding column visible/invisible.
-     * This method can be overridden if there is a desire to create and show a custom menu.
+     * Shows a menu containing all leaf columns as item.
+     * An item can be selected/deselected to make the corresponding column visible/invisible.
      *
-     * @param mouseEvent the {@link MouseEvent} which was generated when the corner region was pressed.
+     * @implNote This method can be overridden to create and show a custom menu.
+     * @param mouseEvent the {@link MouseEvent} which was generated when the table menu button was pressed.
+     * @since 21
      */
     protected void showColumnMenu(MouseEvent mouseEvent) {
         if (columnPopupMenu == null) {
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
@@ -50,6 +50,7 @@ import javafx.scene.control.ContextMenu;
 import javafx.scene.control.Control;
 import javafx.scene.control.Label;
 import javafx.scene.control.TableColumnBase;
+import javafx.scene.input.MouseEvent;
 import javafx.scene.layout.Pane;
 import javafx.scene.layout.Region;
 import javafx.scene.layout.StackPane;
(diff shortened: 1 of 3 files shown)


kevinrushforth:
This adds new public API and, as such, will need additional discussion. Can you start a thread on the openjfx-dev mailing list?

/reviewers 2
/csr

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
```diff
@@ -475,7 +464,25 @@ protected NestedTableColumnHeader createRootHeader() {
         return new NestedTableColumnHeader(null);
     }
 
-
+    /**
+     * Shows a menu with all leaf columns.
+     * These can selected/deselected to make the corresponding column visible/invisible.
+     * This method can be overridden if there is a desire to create and show a custom menu.
+     *
+     * @param mouseEvent the {@link MouseEvent} which was generated when the corner region was pressed.
```
andy-goryachev-oracle, line 472: probably needs
`@since 21`

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
```diff
@@ -475,7 +464,26 @@ protected NestedTableColumnHeader createRootHeader() {
         return new NestedTableColumnHeader(null);
     }
 
-
+    /**
+     * Shows a menu containing all leaf columns as item.
+     * An item can be selected/deselected to make the corresponding column visible/invisible.
+     *
+     * @implNote This method can be overridden to create and show a custom menu.
+     * @param mouseEvent the {@link MouseEvent} which was generated when the table menu button was pressed.
```
kevinrushforth, line 472: Minor: by convention, the text following `@param` and `@return` tags should not end with a period.

kevinrushforth (commented):
The new API looks OK to me. I left a couple comments on the API docs.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
Maran23: Changed. And thanks for the hint, I actually didn't know that this is a convention. 
Personally, I also use the variant without a period, so it's good to know that this is indeed the convention. 😄

andy-goryachev-oracle (approved):
tested on macOS with Tree/TableView, working as expected.

I wish the default implementation did not hide the popup on mouse click, to allow for toggling multiple columns, but it is out of scope for this PR.

Maran23:
> I wish the default implementation did not hide the popup on mouse click, to allow for toggling multiple columns, but it is out of scope for this PR.

I agree that this behavior is rather uncomfortable.
The problem is that we are changing a behavior that may or may not be expected. 
Regardless, we should evaluate it. I think the value is higher than the risk.

In any case, anyone can now customize this to their needs without sacrificing performance or anything else.
This also gives us the possibility to close improvements that we think don't fit to a standard implementation with the hint that this can be implemented by yourself in this case.

kevinrushforth (commented):
The API and docs looks fine, as does the implementation and the tests.

Do you have an example application that overrides this new method? That would help ensure that this new API is sufficient to allow an app to create a custom menu. The main reason I ask is that the default implementation uses internal classes and methods that an application does not have access to. Knowing that an app can do what it needs using only public API will be useful. I expect that it is sufficient, but an example will verify this.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
```diff
@@ -475,7 +464,26 @@ protected NestedTableColumnHeader createRootHeader() {
         return new NestedTableColumnHeader(null);
     }
 
-
+    /**
+     * Shows a menu containing all leaf columns as items.
+     * An item can be selected/deselected to make the corresponding column visible/invisible.
+     *
+     * @implNote This method can be overridden to create and show a custom menu.
+     * @param mouseEvent the {@link MouseEvent} which was generated when the table menu button was pressed
```
nlisker, line 472: There's no need to link to method parameters (`MouseEvent`) since they already linked in the method description.

Maran23: okay. Replaced with `@code`.

andy-goryachev-oracle (approved):

Maran23:
> Do you have an example application that overrides this new method? That would help ensure that this new API is sufficient to allow an app to create a custom menu. The main reason I ask is that the default implementation uses internal classes and methods that an application does not have access to. Knowing that an app can do what it needs using only public API will be useful. I expect that it is sufficient, but an example will verify this.

I created a gist with a sample implementation and some notes here:
https://gist.github.com/Maran23/ad2f974912981788dc50880f8031d7eb
The implementation looks like this:
image

While the `TableSkinUtils` is not public API, in this case it only helps to retrieve the columns as we do not know if the underlying table is a `TreeTableView` or a `TableView`.

nlisker:
For comparison, this is the Win 10 menu:

image

nlisker:
Do you think that the "Size all columns to fit" option should be added here?

kevinrushforth:
> Do you think that the "Size all columns to fit" option should be added here?

Added where? In his example custom menu? Maybe, but it isn't directly relevant to the PR. Or did you mean something else?

Maran23:
> This looks good. If you haven't already done so, can you change `@link` to `@code` in the CSR to match? I'll review that next.

done. Thanks for the reminder!

PR #1136, opened 2023-05-15
8301763: Adding children to wrong index leaves inconsistent state in Parent#childrenSet 

lukostyra:
This issue happened because `childSet` member of Parent was modified during `onProposedChange()` call - that call did not recognize negative indexes as invalid, which caused an exception when actually adding the Node to a List.

This seemed like the simplest solution which doesn't rework a lot of code underneath. Exceptions coming from a backing list/collection technically are handled by `VetoableListDecorator`'s try-catch clauses, however `VetoableListDecorator` does not provide an interface to react when such an exception happens - without it we cannot revert `childSet` back to its original state.

---------
### Progress

### Issue
 * [JDK-8301763](https://bugs.openjdk.org/browse/JDK-8301763): Adding children to wrong index leaves inconsistent state in Parent#childrenSet

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1136/head:pull/1136` \
`$ git checkout pull/1136`

`$ git checkout pull/1136` \
`$ git pull https://git.openjdk.org/jfx.git pull/1136/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1136`

`$ git pr show -t 1136`

</details>

<a href="https://git.openjdk.org/jfx/pull/1136.diff">https://git.openjdk.org/jfx/pull/1136.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1136#issuecomment-1547809277)

### DIFF
--- a/modules/javafx.base/src/main/java/com/sun/javafx/collections/VetoableListDecorator.java
+++ b/modules/javafx.base/src/main/java/com/sun/javafx/collections/VetoableListDecorator.java
@@ -33,6 +33,8 @@ import java.util.ConcurrentModificationException;
 import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
+import java.util.Objects;
+
 import javafx.beans.InvalidationListener;
 import javafx.collections.ListChangeListener;
 import javafx.collections.ObservableList;
@@ -110,6 +112,7 @@ public abstract class VetoableListDecorator<E> implements ObservableList<E> {
 
     @Override
     public boolean setAll(Collection<? extends E> col) {
+        Objects.requireNonNull(col);
         onProposedChange(Collections.unmodifiableList(new ArrayList<>(col)), 0, size());
         try {
             modCount++;
@@ -161,6 +164,7 @@ public abstract class VetoableListDecorator<E> implements ObservableList<E> {
 
     @Override
     public void remove(int from, int to) {
+        Objects.checkFromToIndex(from, to, size());
         onProposedChange(Collections.<E>emptyList(), from, to);
         try {
             modCount++;
@@ -230,6 +234,7 @@ public abstract class VetoableListDecorator<E> implements ObservableList<E> {
 
     @Override
     public boolean addAll(Collection<? extends E> c) {
+        Objects.requireNonNull(c);
         onProposedChange(Collections.unmodifiableList(new ArrayList<>(c)), size(), size());
         try {
             modCount++;
@@ -245,6 +250,8 @@ public abstract class VetoableListDecorator<E> implements ObservableList<E> {
 
     @Override
     public boolean addAll(int index, Collection<? extends E> c) {
+        Objects.requireNonNull(c);
+        Objects.checkIndex(index, size() + 1);
         onProposedChange(Collections.unmodifiableList(new ArrayList<>(c)), index, index);
         try {
             modCount++;
@@ -260,6 +267,7 @@ public abstract class VetoableListDecorator<E> implements ObservableList<E> {
 
     @Override
     public boolean removeAll(Collection<?> c) {
+        Objects.requireNonNull(c);
         removeFromList(this, 0, c, false);
         try {
             modCount++;
@@ -275,6 +283,7 @@ public abstract class VetoableListDecorator<E> implements ObservableList<E> {
 
     @Override
     public boolean retainAll(Collection<?> c) {
+        Objects.requireNonNull(c);
         removeFromList(this, 0, c, true);
         try {
             modCount++;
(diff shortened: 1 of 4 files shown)


hjohn:
I think it may be better to do this check in the callers of `onProposedChange` (using the assertions in `Objects` to test for valid indices.).  This will also allow to check when the index is too large, which is a similar problem. Note that you can't do the "too large" check in `onProposedChange` as for `remove`, `size()` would be too large, while for `add(int, E)` `size()` would still be valid.

Also, for the cases where `onProposedChange` is called with multiple sets of indices, those don't need a check (they come from calls like `removeAll`, which will all be valid).

Furthermore, it should be an `IndexOutOfBoundsException` as this is specified by the `List` interface.

I noticed there are similar problems in `ObservableListWrapper`.  This method for example:

    @Override
    protected void doAdd(int index, E element) {
        if (elementObserver != null)
            elementObserver.attachListener(element);
        backingList.add(index, element);
    }

This will attach a listener to the element (assuming it is an `Observable`) even if `backingList.add` fails...

lukostyra:
Thanks for your notes, I was wondering if there was a better approach. I'll rework the patch.

lukostyra:
As a small note, the issue is solved with current patches, however I think the best solution would be what I mentioned in the original PR message - an interface in `VetoableListDecorator` which would let us "roll-back" any changes from `onProposedChange` (ex. revert `childSet` to original state) when backing lists throw an exception. This would require a more thorough analysis of what bits use `VetoableListDecorator` and if any roll-backs are necessary there in addition to Parent code.

modules/javafx.base/src/main/java/com/sun/javafx/collections/VetoableListDecorator.java
```diff
@@ -110,6 +112,7 @@ public boolean setAll(E... elements) {
 
     @Override
     public boolean setAll(Collection<? extends E> col) {
+        Objects.requireNonNull(col);
         onProposedChange(Collections.unmodifiableList(new ArrayList<>(col)), 0, size());
```
hjohn, line 116: It's good to make it explicit.  It already threw NPE because of `new ArrayList<>(col)`, so this is not a change.

modules/javafx.base/src/main/java/com/sun/javafx/collections/VetoableListDecorator.java
```diff
@@ -260,6 +267,7 @@ public boolean addAll(int index, Collection<? extends E> c) {
 
     @Override
     public boolean removeAll(Collection<?> c) {
+        Objects.requireNonNull(c);
```
hjohn, line 270: Looks like a good change, if `c` was `null` and the backing list was empty, this would not throw NPE, but now it does.

modules/javafx.graphics/src/test/java/test/javafx/scene/ParentTest.java
```diff
@@ -838,6 +838,60 @@ public void testPickingChildNodeWithDirtyViewOrder() {
...
+        assertThrows(IndexOutOfBoundsException.class, () -> g.getChildren().remove(-1));
+        assertEquals(preTestSize, g.getChildren().size());
+
+        assertThrows(IndexOutOfBoundsException.class, () -> g.getChildren().remove(g.getChildren().size()));
+        assertEquals(preTestSize, g.getChildren().size());
+
+        // below call should throw no exception - if it does, internal state is corrupted
+        g.getChildren().remove(0);
+    }
+
```
hjohn, line 894: I think you should also add tests for the `null` cases, if they aren't part of this test already.

hjohn (changes requested):
I think `ObservableListWrapper#remove(int, int)` also needs a check index check, or it may end up remove some items and then throw an exception (exceptions thrown from collections should not modify its state).

hjohn:
> As a small note, the issue is solved with current patches, however I think the best solution would be what I mentioned in the original PR message - an interface in `VetoableListDecorator` which would let us "roll-back" any changes from `onProposedChange` (ex. revert `childSet` to original state) when backing lists throw an exception. This would require a more thorough analysis of what bits use `VetoableListDecorator` and if any roll-backs are necessary there in addition to Parent code.

I'm not convinced this is needed or desirable. The `VetoableListDecorator` is currently set up that its owner can already make "preparations" for the change when it knows it won't veto it (there can be only one vetoer, making this reasonable).  Furthermore, it has the wrapped list under its own control, so it knows what to expect (no exceptions can occur (after this fix) if it decides not to veto).

A roll back mechanism would just complicate the code for very little benefit (a correct FX application won't run into these situations, you'll only be likely to encounter them during development). The whole mechanism surrounding the children list and bad additions should be seen as a best effort early warning system (like ConcurrentModificationExeption).

With this fix I don't see any other ways you could get it in a bad state.

hjohn (approved):

kevinrushforth (approved):
Looks good.

PR #1146, opened 2023-05-30
9075324: Add Linux/riscv64 support

hack3ric:
RISC-V is a new free and open-source ISA that is being widely adopted in the recent years. Currently WebKit is ready for riscv64, and only minor changes is needed to support the platform.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -301,6 +301,7 @@ ext.ARCH_NAME = "x64"
 ext.IS_64 = OS_ARCH.toLowerCase().contains("64")
 ext.IS_AARCH64 = OS_ARCH.toLowerCase().contains("aarch64")
 ext.IS_LOONGARCH64 = OS_ARCH.toLowerCase().contains("loongarch64")
+ext.IS_RISCV64 = OS_ARCH.toLowerCase().contains("riscv64")
 ext.IS_MAC = OS_NAME.contains("mac") || OS_NAME.contains("darwin")
 ext.IS_WINDOWS = OS_NAME.contains("windows")
 ext.IS_LINUX = OS_NAME.contains("linux")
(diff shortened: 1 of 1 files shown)


hack3ric:
> Can you enable GitHub Actions for your personal fork to help verify that this doesn't impact builds on our core platforms?

Sure, enabled.

kevinrushforth:
Can you trigger a GHA run? Since it wasn't enabled when you first pushed your branch, I don't know if you can trigger it manually or not. You might need to push a new commit (either an empty commit or by merging with the latest upstream master) to trigger a GHA run.

hack3ric:
Sorry I messed up a bit and force pushed, but the workflow should be triggered.

kevinrushforth:
The changes look OK to me.

@jaybhaskar Can you also review this?

jaybhaskar (commented):
Webkit has web assembly implementation, in the case of enabling risc64, we need to check any building issue with WebKit, by enabling wasm. The build error usually happens in case of debug build.

hack3ric:
I just noticed Oracle added 'to Webkit' to my bug report title. This patch only allows OpenJFX to build on riscv64 and WebKit already did that.

kevinrushforth:
> I just noticed Oracle added 'to Webkit' to my bug report title. This patch only allows OpenJFX to build on riscv64 and WebKit already did that.

@hack3ric  I just now changed the bug title to:

```
Allow building JavaFX on Linux/riscv64
```

Please change the PR title to match the updated bug title.

FWIW, as long as it doesn't break any supported platforms, I think it is OK to take this fix. Since it isn't WebKit-specific, perhaps @arapte can be the second reviewer?

kevinrushforth (approved):

johanvos (approved):
looks good.
In general, I think the concept of the split between the build.gradle and the separate platform-specific files in buildSrc (e.g. linux.gradle) needs some rethinking as we are increasing the already huge build.gradle. But that should not be part of this PR, hence +1.

PR #1147, opened 2023-05-30
8308191: [macOS] VoiceOver decorations are shifted on second monitor

azuev-java:
The coordinate system of the Mac OS X is reversed to the JavaFX coordinate system meaning that zero point [0,0] is the bottom left corner of the main screen and positive Y coordinate means elevation of the point above the zero line. Taking that into account we should use the main screen's height in order to calculate the correct vertical offset of the window - even if coordinates are physically on the secondary screen, otherwise on screens with different height that are not aligned along the bottom border we will have a wrong vertical position.

---------
### Progress

### Issue
 * [JDK-8308191](https://bugs.openjdk.org/browse/JDK-8308191): [macOS] VoiceOver decorations are shifted on second monitor

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1147/head:pull/1147` \
`$ git checkout pull/1147`

`$ git checkout pull/1147` \
`$ git pull https://git.openjdk.org/jfx.git pull/1147/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1147`

`$ git pr show -t 1147`

</details>

<a href="https://git.openjdk.org/jfx/pull/1147.diff">https://git.openjdk.org/jfx/pull/1147.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1147#issuecomment-1569120160)

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacAccessible.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacAccessible.java
@@ -964,9 +964,7 @@ final class MacAccessible extends Accessible {
     }
 
     private Bounds flipBounds(Bounds bounds) {
-        View view = getRootView((Scene)getAttribute(SCENE));
-        if (view == null || view.getWindow() == null) return null;
-        Screen screen = view.getWindow().getScreen();
+        Screen screen = Screen.getMainScreen();
         float height = screen.getHeight();
         return new BoundingBox(bounds.getMinX(),
                                height - bounds.getMinY() - bounds.getHeight(),


kevinrushforth:
@andy-goryachev-oracle Can you also review?

andy-goryachev-oracle:
> Can you also review?

as an originator of this bug report, most certainly will.

andy-goryachev-oracle:
The problem described in the ticket seems to be fixed, as tested with macOS Ventura 13.3.1(a), using retina (scale=2) and external monitor (scale=1) in all possible arrangements relative to each other (8 in total).

Noticed two intermittent issues when moving the same javafx window between the monitors:

1. at some point, the window turned completely black.  moving it on another monitor fixed the issue:

Screenshot 2023-05-30 at 15 37 38

2. sometimes, especially when monitors were arranged diagonally, accessibility focus rectangle, while placed correctly, was outlining a much taller area.  I can't get this to reproduce reliably, but it feels like it happens after the window was positioned to straddle two monitors:

Screenshot 2023-05-30 at 15 33 14

azuev-java:
> Noticed two intermittent issues when moving the same javafx window between the monitors:
> 
> 1. at some point, the window turned completely black.  moving it on another monitor fixed the issue:

That might be because when moving between displays with the different magnification ratios there is some sort of animation happens to the entirety of the window. Probably sometimes it is stuck on the animation and does not update the window content. 

> 2. sometimes, especially when monitors were arranged diagonally, accessibility focus rectangle, while placed correctly, was outlining a much taller area.  I can't get this to reproduce reliably, but it feels like it happens after the window was positioned to straddle two monitors:

Yes, the way coordinates translation happens on Mac involves virtual screen coordinates and a real ones and when we report the shape of the accessible component we are reporting in the virtual pixels and then OS recalculates the size and thickness of the accessibility cursor based on the magnification factor of the display and the assumed magnification ratio of the control. So it might be possible that we are partially on a different screen and we report the control dimensions and system scales them incorrectly because it thinks that the control belongs to the frame placed on another display.
I am not sure there is a simple fix for such corner case.

andy-goryachev-oracle (approved):
Right.  Both scenarios are intermittent and relatively hard to reproduce, cosmetic, and recoverable simply by moving to another screen.
The scenario described in the ticket is fixed.

kevinrushforth (approved):

PR #1156, opened 2023-06-15
8299753: Tree/TableView: Column Resizing With Fractional Scale

andy-goryachev-oracle:
Modified the resize algorithm to work well with fractional scale, thanks for deeper understanding of the problem thanks to  @hjohn and @mstr2 .

Removed earlier manual tester in favor of the monkey tester.

It is important to note that even though the constraints are given by the user in unsnapped coordinates, they are converted to snapped values, since the snapped values correspond to the actual pixels on the display.  This means the tests that validate honoring constraints should, in all the cases where (scale != 1.0), assume possibly error not exceeding (1.0 / scale).

---------
### Progress

### Issues
 * [JDK-8299753](https://bugs.openjdk.org/browse/JDK-8299753): Tree/TableView: Column Resizing With Fractional Scale (**Bug** - P4)
 * [JDK-8299755](https://bugs.openjdk.org/browse/JDK-8299755): Tree/TableView: Cursor Decouples From Divider When Resizing With Fractional Scale (**Bug** - P4)

### Reviewers
 * [Karthik P K](https://openjdk.org/census#kpk) (@karthikpandelu - Committer) Review applies to [8eab7970](https://git.openjdk.org/jfx/pull/1156/files/8eab7970c48eb3d97609ee9458fb51ecb9547c0c)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1156/head:pull/1156` \
`$ git checkout pull/1156`

`$ git checkout pull/1156` \
`$ git pull https://git.openjdk.org/jfx.git pull/1156/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1156`

`$ git pr show -t 1156`

</details>

<a href="https://git.openjdk.org/jfx/pull/1156.diff">https://git.openjdk.org/jfx/pull/1156.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1156#issuecomment-1593851109)
</details>


karthikpandelu (changes requested):
Tested the fix in Windows 11 and MacOS 13.3.1. The changes fixes the issue in Windows and in mac, the behavior is as expected with and without the change.
Provided couple of minor comments inline.

andy-goryachev-oracle:
fixed, thanks!

andy-goryachev-oracle:
@hjohn would you mind taking a look at this?  using the ideas from your SpaceDistributor

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/ResizeHelper.java
```diff
@@ -67,81 +69,191 @@ public ResizeHelper(ResizeFeaturesBase rf,
...
+
+        double delta = target - sumWidths;
+        if (delta < 0.0) {
+            // shrink
+            if (target < sumPrefs) {
+                distribute(delta, min);
+            } else {
+                distribute(delta, pref);
+            }
+        } else if (delta > 0.0) {
```
karthikpandelu, line 112: Do we need delta = 0.0 condition here?

karthikpandelu (approved):
Left a inline comment.
Otherwise looks good to me

hjohn:
> @hjohn would you mind taking a look at this? using the ideas from your SpaceDistributor

Will take a look this weekend

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/ResizeHelper.java
andy-goryachev-oracle: excellent question.  we don't - zero delta means no change, so no action is needed.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/ResizeHelper.java
```diff
@@ -56,7 +58,7 @@ public ResizeHelper(ResizeFeaturesBase rf,
         this.snap = (rf.getTableControl().isSnapToPixel() ? rf.getTableControl() : null);
```
hjohn, line 58: Obtaining the `Region` here is a bit hacky, this may be out of scope, but I would say `snap` should be a boolean, and the `snapXXX` helper methods in this class should call `ScaledMath`.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/ResizeHelper.java
```diff
@@ -67,81 +69,191 @@ public ResizeHelper(ResizeFeaturesBase rf,
...
+            } else {
+                distribute(delta, pref);
             }
+        }
+    }
+
+    /** distibuting delta (positive when growing and negative when shrinking) */
+    private void distribute(double delta, double[] desired) {
+        double threshold = snapRound(SMALL_DELTA);
+        if (Math.abs(delta) > threshold) {
```
hjohn, line 125: The threshold is pretty arbitrary, I don't think it benefits from snap rounding here.  I'd be more inclined to eliminate the difference between these two functions; I would expect that if a user drags a column resizer for 100 pixels, that it should make no difference whether that was a slow drag or a fast one, and the end visuals and column sizes should be exactly the same regardless.

hjohn (commented):
I only reviewed this partially.

My observation is that this algorithm seems unable to provide a proper user resizing experience as it seems to discard important information it would need to do so.  The `ResizeHelper` is short-lived, and created/recreated many times during a resize operation. I'm not even sure why it is being instantiated at all (it literally is created and discarded in a single method).  The `SMALL_DELTA` constant that changes behavior on how large a "drag" was registered is a red flag; this shouldn't matter.  The sizes should always be based on what they were initially (at the start of the drag), and where the cursor is currently, regardless of what path the cursor took to get there (ie. there should be no memory effects, the algorithm should only need the initial sizes + current position).

I'd expect:

1. User starts a resize
2. All sizes and positions are stored -- this may be a good time to create something like the `ResizeHelper`.  This `ResizeHelper` should never modify the initial sizes, as they're needed for each new calculation until the drag ends.
3. User drags the cursor all over the place
4. Depending on where the cursor is + the initial stored values, a new set of column sizes is calculated and displayed; the algorithm should not have "memory" effects of non-final column sizes of positions where the cursor was before
5. When the drag ends, the final position is exactly what would have happened if the drag was a single instant move, in other words, moving the cursor from `0 -> 100` should be the same as if it was moved from `0 -> -10 -> 90 -> 200 -> 100`

I haven't checked completely how the column resizing works, but I don't see how this could possible be a good algorithm currently. So I'm left wondering what value this change then brings as I think these changes would more than likely all be discarded once the UX is implemented correctly.

A direct JUnit test on this complicated code that verifies all the edge cases would be something I'd like to see added as a minimum, but looking at the code, I don't see this being a good algorithm at all...

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/ResizeHelper.java
```diff
@@ -67,81 +69,191 @@ public ResizeHelper(ResizeFeaturesBase rf,
...
             TableColumnBase<?,?> c = columns.get(i);
-            size[i] = c.getWidth();
+            size[i] = snapCeil(c.getWidth());
 
             if (c.isResizable()) {
-                double cmin = c.getMinWidth();
-                double cmax = c.getMaxWidth();
+                double cmin = snapCeil(c.getMinWidth()); // always honor min width!
+                double cmax = snapFloor(c.getMaxWidth()); // TableColumnBase.doSetWidth() clamps to (min,max) range
                 min[i] = cmin;
```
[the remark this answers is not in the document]
andy-goryachev-oracle: For some reason, I've decided to leave that as is, but since you pointed it out, I think it ought to be fixed.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/ResizeHelper.java
andy-goryachev-oracle: Here i would disagree.  I specifically do not want to replicate the snap code in the Region, since the Region publishes its snapping API.  If there is a change in the snapping APIs in the Region, that change would need to be replicated again here, which I think is a bad idea.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/ResizeHelper.java
andy-goryachev-oracle: theoretically, may be.  I could not achieve it with the algorithm that distributes space for large deltas, as it breaks down with small deltas, see my previous comment.

hjohn:
> > My observation is that this algorithm seems unable to provide a proper user resizing experience as it seems to discard important information it would need to do so.
> 
> please elaborate, or point to a specific problem. It is entirely possible that a better algorithm might exist, but it might be out of scope _at the moment_, as this is a follow-up for a specific issue.

It's hard to point to a specific problem when most of the algorithm used would be unnecessary if it used the initial conditions + current resize position as its basis for calculating the column sizes.  My problem with this implementation is that it takes what is fundamentally a very simple algorithm (columns have sizes X,Y,Z and Y is resized 10 pixels larger, what should the layout be?) and turns it into a frame rate dependent, mouse movement dependent delta calculation.  The initial conditions are discarded, and so a single drag resize of 10 pixels is NOT the same as a drag resize that captured several individual steps (1 +  2 +  3  + 4 pixels), while it really should be...

On top of that, if indeed the algorithm is flawed, as I think it is, then there is no way to really fix it apart from some cosmetic changes.  This then would be a lot of wasted effort.  As I noted, there is no JUnit test for this code as of yet, and for such a complicated algorithm to be verified to be correct, I think it would need one to pass review.  If we're willing to forego that, then I suppose a casual fix is in the cards, but I can't really see whether or not it would be correct without extensive manual testing.

> > The `SMALL_DELTA` constant that changes behavior on how large a "drag" was registered is a red flag; this shouldn't matter. The sizes should always be based on what they were initially (at the start of the drag), and where the cursor is currently, regardless of what path the cursor took to get there (ie. there should be no memory effects, the algorithm should only need the initial sizes + current position).
> 
> This is not my experience. Specifically, the difference in behavior between small changes (when the users resizes the columns slowly and we get small deltas of 1 pixel) and large changes (e.g. when initially resizing the table, or the user moves the mouse quickly) are significant.

Yes, it would be needed with this algorithm as it is dependent on mouse cursor speed and frame rate as it has no idea of what the initial positions were and how it arrived at the current state.

> For example, I tried to use your SpaceDistributor from #1111 and it suffered from the same problem as bypassing the small delta path (by setting SMALL_DELTA=0) - when the user resizes the columns slowly, the same column gets the pixel and grows wider than the rest.

It would not be sufficient to just replace `ResizeHelper` with something that uses the space distributor as the information it needs would still be lost. When writing algorithms that resize a UI, using the current size of the elements in your UI + a resize amount will never result in a consistent looking resize.  It always must be based on their size constraints (which can be min/pref/max based for controls that can't be individually resized) or the actual sizes when the action started (for splitters or columns). Once the action is finished, only then do the new sizes become the initial sizes for the next action -- not at every mouse event or frame that elapses.

`HBox` for example doesn't use its **current** sizes to calculate its "new" size -- it always goes back to the basic min/pref/max sizes, then looks at the space available and computes new sizes based on only that information.  If it used the current sizes as its base, you would notice that in certain scenario's the final sizes (if you slowly vs quickly resized a window) would not necessarily always be the same for the same available width -- that's exactly what I think is happening with the column sizes; it's not deterministic and can't be because it lacks the information to do so and hence has to resort to `SMALL_DELTA` tricks to differentiate between "fast" and "slow" resizes.  Relying on such things (which can vary with mouse hardware, current framerate, CPU/GPU load, amount of visible `Node`s) makes for an inconsistent resize experience, even when the column was resized by the exact same amount of pixels when the resize ends.

This can I think be fixed relatively easily; all it really would take is to track when the drag starts, store the sizes, use these sizes to calculate new sizes each time, and when the drag ends, discard the stored sizes.  The `ResizeHelper` would then only need a single much simpler method that takes an array of sizes, the resizing mode, the space available and which column was resized and by how much -- or if you want, you can associate the `ResizeHelper` with the drag resize start, and use the same helper while the resize is in progress.

andy-goryachev-oracle:
Thank you for a detailed write up, @hjohn .

One thing to note: the new code is a modification of the earlier work on constrained column resize policies, that logic is unaffected.  The only difference is to recognize the fact that in an over-constrained situation sone constraints must be relaxed, and that eliminates multiple passes.

Another consideration is that we are not re-writing Tree/TableView skin column resizing code.  The fact that the current public APIs do not provide us with the initial condition and instead invoke the policy resize methods giving small deltas is a limitation this PR is not going to address.

mstr2:
I agree with John that a layout algorithm that uses incremental calculations will always be flawed in principle. The correct approach is to store the initial configuration, and then for each configuration change, go back to the initial configuration and recompute the layout solution.

Now, we might still accept a bugfix for a flawed algorithm. But JDK-8299753 is an enhancement, not a bugfix. I'm not sure what to make of this: it's obviously a flawed approach, and basing an enhancement on a flawed approach means that someone would have to come back to this issue in the future and solve it correctly.

I don't think that the issue at hand is so severe that it's a forced move to integrate this interim solution.

andy-goryachev-oracle:
I respectfully disagree, @mstr2 .

This fix is not an interim solution - unlike any theoretical considerations of "flawed approach", in practice this code does work as expected with integer and fractional scales, and, given our situation, it's highly unlikely that any alternative solutions will ever be considered, especially those that would affect existing public APIs.

mstr2:
> I respectfully disagree, @mstr2 .
> 
> This fix is not an interim solution - unlike any theoretical considerations of "flawed approach", in practice this code does work as expected with integer and fractional scales, and, given our situation, it's highly unlikely that any alternative solutions will ever be considered, especially those that would affect existing public APIs.

It isn't just a theoretical issue. The proposed patch fails to keep the divider precisely at the cursor location, depending on frame rate and mouse movement speed. This is how the behavior manifests on my machine:

https://github.com/openjdk/jfx/assets/43553916/79cf04be-bd18-4cfe-9c34-912978ee96ee

andy-goryachev-oracle:
Thank you for the video, @mstr2 

This is an intrinsic problem with the Tree/TableView - you can observe it even with the UNCONSTRAINED_RESIZE_POLICY.  You both are right in that it's impossible to fix this without remembering the initial MOUSE_PRESSED position.  I wonder if it's possible to fix it without making API changes.

However, fixing this is out of scope for this PR.  This PR deals with correctly handling of fractional scale (within the constraints of the current resize policy API) and cursor decoupling even when moving very slowly.

andy-goryachev-oracle:
@kevinrushforth please take a look.  thanks!

andy-goryachev-oracle:
please review

andy-goryachev-oracle:
@karthikpandelu @hjohn @kevinrushforth could you please review?

karthikpandelu:
Tested the changes in Windows 11 with different scale values. The fix works as expected.
I will review the code soon and complete the review.

andy-goryachev-oracle:
keep alive

andy-goryachev-oracle:
keep alive

andy-goryachev-oracle:
Happy birthday to me #1156

andy-goryachev-oracle:
please review

andy-goryachev-oracle:
@kevinrushforth please review

kevinrushforth:
It's on  my list to review, but at a low priority. I have concerns over the proposed solution, specifically around the creation and use of the (not public) snapInnerSpace method.

andy-goryachev-oracle:
> snapInnerSpace method.

The main idea is to handle the case when computing snapped coordinates within the unsnapped container.  Granted, this is somewhat unusual case, so I am ok with using simple rounding here.

FYI, I've added Page -> Snapped Split Panes menu option to the monkey tester to test exactly this case.  I don't think I see much difference on Windows with fractional scale.

andy-goryachev-oracle:
this toddler PR needs attention

andy-goryachev-oracle:
keep alive

andy-goryachev-oracle:
please review

andy-goryachev-oracle:
keep-alive

andy-goryachev-oracle:
keep-alive

modules/javafx.controls/src/main/java/javafx/scene/control/ResizeFeaturesBase.java
```diff
@@ -96,7 +97,24 @@ public ResizeFeaturesBase(@NamedArg("column") TableColumnBase<S,?> column, @Name
    *
    * @since 20
    */
-  public void setColumnWidth(TableColumnBase<S,?> col, double width) {
-      col.doSetWidth(width);
+  public void setColumnWidth(TableColumnBase<S, ?> col, double width) {
+      Control c = getTableControl();
+      if (c.isSnapToPixel()) {
+          double min = c.snapSizeX(col.getMinWidth());
+          double max = RegionHelper.snapInnerSpaceX(c, col.getMaxWidth());
```
kevinrushforth, line 104: I'm still not convinced that min and max should be snapped differently (floor vs ceiling).

modules/javafx.controls/src/main/java/javafx/scene/control/ResizeFeaturesBase.java
```diff
@@ -96,7 +97,24 @@ public ResizeFeaturesBase(@NamedArg("column") TableColumnBase<S,?> column, @Name
...
+              if (width < min) {
+                  // safety check in case floor(max) < ceil(min)
+                  width = min;
+              }
+          } else {
+              width = c.snapPositionX(width);
+          }
+      }
+      // can set width directly because all constraints have been checked
+      col.setWidth(width);
```
kevinrushforth, line 118: The constraints have been checked only in the default snap-to-pixel case. This will bypass clamping if snap-to-pixel is false. I recommend moving the `setWidth` call into the if statement, and calling `doSetWidth` in an `else` block.

kevinrushforth (commented):
I reread the discussion on this, and have a few thoughts.

1. [JDK-8299753](https://bugs.openjdk.org/browse/JDK-8299753) is a pretty noticeable bug when dealing with non-integer render scales (especially 125%) in the default snap-to-pixel case.

2. I agree with @hjohn and @mstr2 that the existing `TableView` resize-on-mouse-drag algorithm is flawed. The way that successive delta values are accumulated, which leads to different results for slow vs fast mouse moves, highlights this problem (SMALL_DELTA is a partial workaround, but it's only needed in the first place because the algorithm is flawed). I haven't looked closely enough to see whether it would be possible to fix this without public API changes, but I can imagine that it might not be. Either way it would be a large effort to fix the algorithm with a lot of testing needed. Long term, it might be the right thing to do, but it's out of scope for this bug fix.

Given that rewriting the algorithm is out of scope here, the question becomes whether this bug is worth the effort to fix (I think it is) and whether this particular change is sound, given the existing algorithm. The main thing I would want to ensure is that we don't regress in the case of an integer screen scale.

My testing of it so far looks good, but I want to review the code changes to satisfy myself that the risk of regression is small. I'm still a bit skeptical of the need for the (not-public) snapInnerSpace method, but I'll take a closer look.

We'll need to testing with snap-to-pixel disabled as well (I spotted what I think is a bug in that case, but that was just something I happened to notice -- I still  have not done a thorough review of the code).

If we are going to get this in, doing so early in the jfx25 release would be good so it gets more testing time.

modules/javafx.controls/src/main/java/javafx/scene/control/ResizeFeaturesBase.java
andy-goryachev-oracle: we will violate the max constraint if we snap to a larger value.

so for the max constraint, we should `snapInnerSpace()` which is `Region.snapPortionX()`.

kevinrushforth: > we will violate the max constraint if we snap to a larger value.

Well, it's inherently an over-constrained problem. Conceptually, though, this does not violate the constraint. When snap-to-pixel is enabled, all values, including min and max values are snapped to a pixel that can be almost 1 pixel larger or smaller than it's given value. Within that tolerance, we are honoring the min  and max values.

It isn't required that we always ceil the min value and floor the max value (not to mention that doing that is what causes the odd min > max).

So I don't buy this argument. Is there some other reason to do it?

andy-goryachev-oracle: no other reason.  are you suggesting to simply `round()`?

andy-goryachev-oracle:
this PR currently does not build

andy-goryachev-oracle:
changed to `snapSizeX()` and `doSetWidth` in `else`.

andy-goryachev-oracle:
@aghaisas or @arapte would you please take a look?

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/ResizeHelper.java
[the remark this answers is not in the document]
kevinrushforth: As with my earlier comment in `ResizeFeaturesBase` (which you  fixed), I don't think we should be using different snapping functions for min and max. I recommend `snapCeil` for  both.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/ResizeHelper.java
```diff
@@ -67,81 +69,190 @@ public ResizeHelper(ResizeFeaturesBase rf,
...
 
             if (c.isResizable()) {
-                double cmin = c.getMinWidth();
-                double cmax = c.getMaxWidth();
+                double cmin = snapCeil(c.getMinWidth()); // always honor min width!
+                double cmax = snapFloor(c.getMaxWidth());
                 min[i] = cmin;
                 max[i] = cmax;
-                pref[i] = clip(c.getPrefWidth(), cmin, cmax);
+                pref[i] = clip(snapRound(c.getPrefWidth()), cmin, cmax);
```
kevinrushforth, line 79: Since `prefWidth` is a size, should it also be `snapCeil`?

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/ResizeHelper.java
```diff
@@ -444,169 +551,64 @@ protected double resize(int ix, double delta) {
...
+        if (snap != null) {
+            return snap.snapPositionX(x);
         }
-        return sum;
+        return x;
     }
 
-    protected static boolean isZero(double x) {
-        return Math.abs(x) < EPSILON;
+    private double snapFloor(double x) {
```
kevinrushforth, line 588: This method can be removed if you change max to use `snapCeil`.

kevinrushforth (commented):
All my testing looks good. I left a couple more comments about using ceil for max as well as min.

kevinrushforth (approved):
LGTM now.

arapte (approved):
LGTM. Tested a few samples of TableView.

andy-goryachev-oracle:
woo hoo!

andy-goryachev-oracle:
Thank you everyone who helped with this PR over the years!   :-)

PR #1157, opened 2023-06-21
8306083: Text.hitTest is incorrect when more than one Text node in TextFlow

karthikpandelu:
The text run selected in `PrismTextLayout::getHitInfo()` method for character index calculation was not correct when hitTest was invoked for Text node in a TextFlow with more than one Text child. Hence wrong character index value was calculated.

Since only x, y coordinates were available in the above mentioned method, sending the text as a parameter to this method is necessary so as to know if the text run selected for character index calculation is correct. Along with this change modified the `PrismTextLayout::getHitInfo()` method to calculate the correct character index.

Added tests to validate the changes.

---------
### Progress

### Issue
 * [JDK-8306083](https://bugs.openjdk.org/browse/JDK-8306083): Text.hitTest is incorrect when more than one Text node in TextFlow (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1157/head:pull/1157` \
`$ git checkout pull/1157`

`$ git checkout pull/1157` \
`$ git pull https://git.openjdk.org/jfx.git pull/1157/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1157`

`$ git pr show -t 1157`

</details>

<a href="https://git.openjdk.org/jfx/pull/1157.diff">https://git.openjdk.org/jfx/pull/1157.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1157#issuecomment-1600267261)


andy-goryachev-oracle:
Looks like Text.hitTest() does not return the right information (charIndex=0, leading) for text on either the second line, or a wrapped line.  (mouse over over 2nd/3rd lines):

Screenshot 2023-06-21 at 15 56 30

karthikpandelu:
It works when content of one Text node exceeds one line or is wrapped but this is not working when more than one node is present in TextFlow and nodes exceeds one line. This is an issue. I will look into this.

karthikpandelu:
The cause for above mentioned issue is that, the `line.getRuns()` call in line no.439 in PrismTextLayout.java returns text runs only for the first Text node. Because of this issue correct character index is not calculated.
It looks like the issue is in line creation. I will fix this issue along with the character index issue we are fixing already in this PR.

karthikpandelu:
Fixed issue with wrapped text.

andy-goryachev-oracle:
does not seem to work right still - if you arrange the split panes to make the emoji wrap (see the next screenshot), I think not only it shows incorrect insertion index when you move the mouse across that emoji (in Text.hitTest goes from 0 to 1, while expected change is 2), but I think it's also wrong because that emoji is the last symbol in that Text instance, so insertion index should be different (larger).
Are you reporting insertion index relative to the Text's text or TextRun?

Screenshot 2023-06-27 at 09 07 30

Screenshot 2023-06-27 at 09 07 36

karthikpandelu:
> I think not only it shows incorrect insertion index when you move the mouse across that emoji (in Text.hitTest goes from 0 to 1, while expected change is 2), but I think it's also wrong because that emoji is the last symbol in that Text instance, so insertion index should be different (larger). 

Looks like the wrong insertion index is getting reported after 2nd line. I'll fix this issue.

>Are you reporting insertion index relative to the Text's text or TextRun?

The insertion index is reported relative to the Text's text incase of Text node.

andy-goryachev-oracle:
> The insertion index is reported relative to the Text's text incase of Text node.

right.  in the case of the (highlighted) emoji character - the insertion index starts with 0 which I think cannot be correct, because all of the text should be added as a single Text instance, see 

https://github.com/andy-goryachev-oracle/MonkeyTest/blob/e361b7800b5497da5e23a7eda26084e800410f38/src/com/oracle/tools/fx/monkey/util/Templates.java#L60

and

https://github.com/andy-goryachev-oracle/MonkeyTest/blob/e361b7800b5497da5e23a7eda26084e800410f38/src/com/oracle/tools/fx/monkey/pages/TextFlowPage.java#L208

karthikpandelu:
> right. in the case of the (highlighted) emoji character - the insertion index starts with 0 which I think cannot be correct, because all of the text should be added as a single Text instance, see
> 
If a Text node's content is wrapped, then character index will start from 0 again in the new line or do we expect it to be relative to text node and continue with character index from previous line?
If the latter behavior is expected then I couldn't find a way to differentiate between a text node which is wrapped and which is not. 
Even if warp information is available, the `getOffsetAtX` method returns the offset (character index) based on the x coordinate value which is relative to the new line. If we use run's start value along with the offset, then we get absolute value of character index relative to textflow. So it looks to me that if a Text node's content is wrapped, then character index in the new line will start from 0. In this case insertion index starting from 0 looks to be correct.

karthikpandelu:
Fixed the character index and insertion index issue mentioned above. Please check.
One more issue I observed is that when Text node is displayed in single line, correct x coordinate is picked up by Mouse event, but when text is displayed in more than one line, x coordinate picked up by Mouse event at the same position is different.
image
x value in highlighted position is 7.01

image
x value in highlighted position is 83.0.

Looks like there is issue in MouseEvent or PickResult.

andy-goryachev-oracle:
> Looks like there is issue in MouseEvent or PickResult.

Do you mean MouseVent.getX()?  I cannot reproduce that (in the Monkey Tester).  
I've added a mouse handler to the control (Text or TextFlow, depending on the page) and ev.getX() always returns correct result.

andy-goryachev-oracle:
Part of the issue seems fixed, but i noticed an irregularity with the tone skin emoji.  Now we do not support grapheme clusters (yet), but still, the insertion index reported by Text suddenly jumps from a correct value to what looks like a too big of a value.

In the following screenshots, if the mouse pointer is hovering to the left of the red caret line, the insertion index is 4.  But if you move it to the right side, the index jumps to 33 which does not look right.

Mouse to the left of the line:
Screenshot 2023-06-30 at 10 14 53

Mouse to the right of the line:

Screenshot 2023-06-30 at 10 15 00

I can tell that the red line on the screenshots corresponds to a boundary between two different Text instances (see TextFlowPage:204).

The TextFlow portion does seem to work correctly.

karthikpandelu:
> Do you mean MouseVent.getX()? 

The x coordinate reported by `MouseEvent.getPickResult().getIntersectedPoint()` method. I can reproduce this issue in MonkeyTester as well. When I checked the `Point2D` object sent to `Text::hitTest()` method, it is different when RichText is displayed in multiple lines like how it is shown in the above screenshot and when it is displayed in single line as shown below. 
image

In both the cases cursor is hovered just over the right side of caret line. The x coordinate value reported is given below.
Single line - x: 102
Multi line - x: 1
Because of this issue, the character index calculated for the same character changes.

However I agree that bug is present in the character index calculation like you mentioned above. I'll work on fixing that.

karthikpandelu:
> However I agree that bug is present in the character index calculation like you mentioned above. I'll work on fixing that.

Correction: The character index calculation looks to be correct when x coordinate passed to it is correct.
One more case I observed, if we further reduce the width of TextFlow pane by moving the splitter as shown in screenshot below, the x coordinate value reported is correct again. 
image
If cursor is hovered just after the caret line, x value reported is 1 which is correct. Character index calculated in this case is also correct.

karthikpandelu:
The x coordinate value returned from mouse event is correct and it is not a bug as the x coordinate value is relative to the Text node bounds and  not relative to the start of the first character. Hence fixed the character index calculation issue as part of this PR.
Made changes in text as well to wrap Text node content.

modules/javafx.graphics/src/main/java/com/sun/javafx/text/PrismTextLayout.java
```diff
@@ -702,14 +748,31 @@ public boolean setTabSize(int spaces) {
...
+                        }
+                        isTextPresent = true;
+                        break;
+                    }
+                }
+            }
             bottom += lines[index].getBounds().getHeight() + spacing;
             if (index + 1 == lineCount) bottom -= lines[index].getLeading();
-            if (bottom > y) break;
+            if (bottom > y && isTextPresent) break;
```
andy-goryachev-oracle, line 775: I would recommend adding { }'s to ifs here and on line 774.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/text/TextLayout.java
```diff
@@ -199,7 +199,7 @@ public Hit(int charIndex, int insertionIndex, boolean leading) {
      */
     public boolean setTabSize(int spaces);
 
-    public Hit getHitInfo(float x, float y);
+    public Hit getHitInfo(float x, float y, String text);
```
andy-goryachev-oracle, line 202: Perhaps we could add javadoc with an explanation for the 'text' parameter, that it's expected to be null in the case of TextFlow and non-null in the case of Text.

modules/javafx.graphics/src/main/java/com/sun/javafx/text/PrismTextLayout.java
```diff
@@ -440,24 +443,67 @@ public Hit getHitInfo(float x, float y) {
...
 
                 insertionIndex = charIndex;
                 if (getText() != null && insertionIndex < getText().length) {
                     if (!leading) {
                         BreakIterator charIterator = BreakIterator.getCharacterInstance();
-                        charIterator.setText(new String(getText()));
+                        if (text != null) {
+                            charIterator.setText(text);
+                        } else {
+                            charIterator.setText(new String(getText()));
```
andy-goryachev-oracle, line 505: getText() is possibly an expensive operation when the layout contains a lot of text.
Since we are looking only for the following "character", is it possible to optimize this and only use a subset of spans or limit the amount of text passed to the break iterator in some other way?
What do you think?

(it's likely to be out of scope for this PR)

modules/javafx.graphics/src/main/java/com/sun/javafx/text/PrismTextLayout.java
```diff
@@ -702,14 +748,31 @@ public boolean setTabSize(int spaces) {
      *                                                                         *
      **************************************************************************/
 
-    private int getLineIndex(float y) {
+    private int getLineIndex(float y, String text) {
         int index = 0;
         float bottom = 0;
+        boolean isTextPresent = text == null ? true : false;
```
andy-goryachev-oracle, line 754: this looks very confusing to me... what text is present?

modules/javafx.graphics/src/main/java/com/sun/javafx/text/TextRun.java
```diff
@@ -393,7 +393,7 @@ public int getGlyphAtX(float x, int[] trailing) {
         float runX = 0;
         for (int i = 0; i < glyphCount; i++) {
             float advance = getAdvance(i);
-            if (runX + advance > x) {
+            if (runX + advance >= x) {
```
andy-goryachev-oracle, line 396: this is very interesting.  how did it work before?
could there be any scenarios that might have failed before and are fixed with this change?

tests/system/src/test/java/test/robot/javafx/scene/TextCharacterIndexTest.java
```diff
@@ -0,0 +1,456 @@
...
+
+    boolean isLeading;
+    boolean isSurrogatePair;
+    int charIndex;
+    int insertionIndex;
+    int textFlowCharIndex;
+
+    private void mouseClick(double x, double y) {
+        Util.runAndWait(() -> {
+            robot.mouseMove((int) (scene.getWindow().getX() + scene.getX() + x),
```
andy-goryachev-oracle, line 136: minor: local Window variable eliminates multiple invocations of getWindow()
(I am too lazy to see if javac is smart enough to optimizes it away)

tests/system/src/test/java/test/robot/javafx/scene/TextCharacterIndexTest.java
```diff
@@ -0,0 +1,456 @@
...
+                    textFlow.getLayoutY() + y);
+    }
+
+    private void addTwoTextNodes() {
+        Util.runAndWait(() -> {
+            textOne = new Text("This is Text");
+            textOne.setFont(new Font(48));
+            textTwo = new Text("This is Text");
+            textTwo.setFont(new Font(48));
+            textFlow.getChildren().clear();
```
andy-goryachev-oracle, line 153: is this clear() a necessary part of the test?
setAll() "Clears the ObservableList and adds all the elements..." according to its javadoc

this comment applies here and in all other places (line 164, etc.)

tests/system/src/test/java/test/robot/javafx/scene/TextCharacterIndexTest.java
```diff
@@ -0,0 +1,456 @@
...
+
+            isLeading = hitInfo.isLeading();
+            charIndex = hitInfo.getCharIndex();
+            insertionIndex = hitInfo.getInsertionIndex();
+            isSurrogatePair = Character.isSurrogate(t.getText().charAt(charIndex));
+        }
+
+        // String testString = textOne.getText();
+        // testString += textTwo.getText();
+        // isSurrogatePair = Character.isSurrogate(testString.charAt(charIndex));
```
andy-goryachev-oracle, line 403: clean up?

andy-goryachev-oracle (commented):
The updated code works correctly, as far as I can tell.  Left some minor comments.
This PR definitely needs another pair of eyes.

modules/javafx.graphics/src/main/java/com/sun/javafx/text/PrismTextLayout.java
karthikpandelu: Added {} as suggested

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/text/TextLayout.java
karthikpandelu: Added javadoc comment

modules/javafx.graphics/src/main/java/com/sun/javafx/text/PrismTextLayout.java
karthikpandelu: I agree that the naming and the conditionals were confusing. So optimized the code for better readability.

modules/javafx.graphics/src/main/java/com/sun/javafx/text/TextRun.java
karthikpandelu: Without this condition when `runX + advance` becomes equal to `x`, character index calculated will be wrong. 
For example if we have 2 emojis (😀😃), when we move from first emoji to second, character index returned will be 1 instead of 2 when we just hit the second emoji. This is a failing scenario.

tests/system/src/test/java/test/robot/javafx/scene/TextCharacterIndexTest.java
karthikpandelu: Made changes as suggested. As far as I checked, javac does not optimize this.

tests/system/src/test/java/test/robot/javafx/scene/TextCharacterIndexTest.java
karthikpandelu: Yes clear is not required. Removed from all instances.

modules/javafx.graphics/src/main/java/com/sun/javafx/text/PrismTextLayout.java
```diff
@@ -751,33 +751,35 @@ public boolean setTabSize(int spaces) {
...
+            }
+            if (bottom > y && textFound) {
+                break;
+            }
             index++;
         }
         return index;
     }
 
+
```
andy-goryachev-oracle, line 782: is there an unnecessary newline here?

tests/system/src/test/java/test/robot/javafx/scene/TextCharacterIndexTest.java
```diff
@@ -133,8 +134,9 @@ public class TextCharacterIndexTest {
 
     private void mouseClick(double x, double y) {
         Util.runAndWait(() -> {
-            robot.mouseMove((int) (scene.getWindow().getX() + scene.getX() + x),
-                    (int) (scene.getWindow().getY() + scene.getY() + y));
+            Window st = scene.getWindow();
+            robot.mouseMove((int) (st.getX() + scene.getX() + x),
+                    (int) (st.getY() + scene.getY() + y));
```
andy-goryachev-oracle, line 139: two things:
1. May be `Window w` would be better than `st`
2. Should it be (int) or (int)Math.round(...) (or (int)Math.rint(...) ?)

andy-goryachev-oracle (commented):
looks good, have some very minor comments.

modules/javafx.graphics/src/main/java/com/sun/javafx/text/PrismTextLayout.java
karthikpandelu: Yes. Removed unnecessary newline

tests/system/src/test/java/test/robot/javafx/scene/TextCharacterIndexTest.java
karthikpandelu: >     1. May be `Window w` would be better than `st`
> 

Yes. It is much better. Updated the code.

 >     2. Should it be (int) or (int)Math.round(...) (or (int)Math.rint(...) ?)

I think (int) is sufficient in this case.

andy-goryachev-oracle (approved):
thank you for addressing the comments!  lgtm now.

kevinrushforth:
@prrace would you be able to review this?

andy-goryachev-oracle:
@prrace could you please take a look at this?

prrace (changes requested):

modules/javafx.graphics/src/main/java/com/sun/javafx/text/PrismTextLayout.java
```diff
@@ -702,14 +748,32 @@ public boolean setTabSize(int spaces) {
...
+    private int getLineIndex(float y, String text) {
         int index = 0;
         float bottom = 0;
+        /* Initializing textFound as true when text is null
+         * because when this function is called for TextFlow text parameter will be null */
+        boolean textFound = (text == null);
+
         int lineCount = getLineCount();
         while (index < lineCount) {
+            if (text != null) {
```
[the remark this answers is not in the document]
karthikpandelu: Updated code to use `textFound`

modules/javafx.graphics/src/main/java/com/sun/javafx/text/PrismTextLayout.java
```diff
@@ -702,14 +748,32 @@ public boolean setTabSize(int spaces) {
...
+                for (TextRun r : lines[index].runs) {
+                    if (r.getTextSpan() == null || r.getTextSpan().getText().equals(text)) {
```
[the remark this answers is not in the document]
karthikpandelu: > (1) If you are looking for a particular match of text, what if there are 2 lines or spans that have the exact same text ? 

If 2 runs have the same text as it can happen in the case of emojis, line number will be calculated correctly. But if 2 text nodes have the exact same text and both are present in the same line, the character index calculation fails. I will check this case.

>(2) What is that the logic behind this line ? At least currently you only pass in text for a Text node. Can you break down what exactly it is doing ?

The line index calculation was failing for the Text node. In the case of Text node, even before the actual text line/run is reached, `bottom > y` was becoming true since it is relative to the text node. Hence added this condition of `r.getTextSpan().getText().equals(text)` to make sure that line index is calculated for the correct Text node.

karthikpandelu:
Fixed the issue when more than one text nodes with same text is present in text node. Added new test cases and improved current tests.
@andy-goryachev-oracle and @prrace please re-review.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/text/TextLayout.java
```diff
@@ -210,7 +210,7 @@ public Hit(int charIndex, int insertionIndex, boolean leading) {
      *             and non-null in the case of {@link javafx.scene.text.Text}
      * @return returns a {@link Hit} object containing character index, insertion index and position of cursor on the character.
      */
-    public Hit getHitInfo(float x, float y, String text);
+    public Hit getHitInfo(float x, float y, String text, int textRunStart, int curRunStart);
```
andy-goryachev-oracle, line 213: please add new `@param` blocks to javadoc

modules/javafx.graphics/src/main/java/com/sun/javafx/text/PrismTextLayout.java
```diff
@@ -455,16 +455,17 @@ public Hit getHitInfo(float x, float y, String text) {
             } else {
                 for (int i = 0; i < lineIndex; i++) {
                     for (TextRun r: lines[i].runs) {
-                        if (r.getTextSpan() != null && r.getTextSpan().getText().equals(text)) {
+                        if (r.getTextSpan() != null && r.getTextSpan().getText().equals(text) && r.getStart() >= textRunStart) {
```
andy-goryachev-oracle, line 458: speedup: you may want to compute `>=textRunStart` condition before `.equals(text)` to save CPU cycles (basically, swap the two conditions here)

modules/javafx.graphics/src/main/java/com/sun/javafx/text/PrismTextLayout.java
```diff
@@ -455,16 +455,17 @@ public Hit getHitInfo(float x, float y, String text) {
...
                             textWidthPrevLine += r.getLength();
                         }
                     }
                 }
                 int prevNodeLength = 0;
                 boolean isPrevNodeExcluded = false;
                 for (TextRun r: runs) {
-                    if (!r.getTextSpan().getText().equals(text)) {
+                    if (!r.getTextSpan().getText().equals(text) || (r.getTextSpan().getText().equals(text) && r.getStart() < textRunStart)) {
```
andy-goryachev-oracle, line 466: same suggestion about swapping conditions to save some CPU cycles

modules/javafx.graphics/src/main/java/com/sun/javafx/text/PrismTextLayout.java
```diff
@@ -757,9 +758,9 @@ private int getLineIndex(float y, String text) {
 
         int lineCount = getLineCount();
         while (index < lineCount) {
-            if (text != null) {
+            if (!textFound) {
                 for (TextRun r : lines[index].runs) {
-                    if (r.getTextSpan() == null || r.getTextSpan().getText().equals(text)) {
+                    if (r.getTextSpan() == null || (r.getTextSpan().getText().equals(text) && r.getStart() == runStart)) {
```
andy-goryachev-oracle, line 763: and definitely here `==` should go before `.equals()`

modules/javafx.graphics/src/main/java/javafx/scene/text/Text.java
```diff
@@ -1022,7 +1023,21 @@ public final HitInfo hitTest(Point2D point) {
...
         double x = point.getX() - getX();
         double y = point.getY() - getY() + getYRendering();
-        TextLayout.Hit h = layout.getHitInfo((float)x, (float)y, getText());
+        GlyphList[] runs = getRuns();
+        int runIndex = 0;
+        if (runs.length != 0) {
+            Point2D ptInParent = localToParent(x, y);
+            while (runIndex < runs.length - 1) {
+                if (ptInParent.getY() > runs[runIndex].getLocation().y
+                        && ptInParent.getY() < runs[runIndex + 1].getLocation().y) {
```
andy-goryachev-oracle, line 1032: [minor]: the formatting looks off.  If we rename 'ptInParent` to `p` it will fit on one line.
Also, ptInParent.getY() gets called twice, may be use a local variable?  Also will make it shorter.

modules/javafx.graphics/src/main/java/javafx/scene/text/Text.java
```diff
@@ -1022,7 +1023,21 @@ public final HitInfo hitTest(Point2D point) {
...
+                    break;
+                }
+                runIndex++;
+            }
+        }
+        int textRunStart = runs.length == 0 ? 0 : ((TextRun) runs[0]).getStart();
+        int curRunStart = runs.length == 0 ? 0 : ((TextRun) runs[runIndex]).getStart();
```
andy-goryachev-oracle, line 1039: very minor: we are computing run.length == 0 twice.  I know it might look neater than
```
int textRunStart;
int curRunStart;
if(runs.length == 0) {
  textRunStart = 0;
  curRunStart = 0;
} else {
 ...
}
```

andy-goryachev-oracle (commented):
I think this looks good (with minor suggestions).  Tested with the MonkeyTester and with a rich text that contains multiple Text instances with exactly the same text.

andy-goryachev-oracle:
Another observation, and it may be a different issue.  If i try to insert a Text instance with -fx-scale-x:200% into the flow, the scale affects the glyph rendering but neither caretShape nor hitInfo:

```
                t("BOLD ", f, "-fx-font-weight:bold;"),
                t("BOLD ", f, "-fx-font-weight:100; -fx-scale-x:200%;"),
                t("BOLD ", f, "-fx-font-weight:900;"),
```

Screenshot 2023-10-12 at 12 33 36

karthikpandelu:
Made changes in code as suggested above.

> Another observation, and it may be a different issue. If i try to insert a Text instance with -fx-scale-x:200% into the flow, the scale affects the glyph rendering but neither caretShape, preferredWidth, nor hitInfo:

This looks like a different issue. I will raise a new bug if it is not reported already.

karthikpandelu:
Created [JDK-8318079](https://bugs.openjdk.org/browse/JDK-8318079)

andy-goryachev-oracle (approved):
🕶 looks good, thanks!

karthikpandelu:
@prrace could you please re-review?

prrace (approved):

PR #1159, opened 2023-06-21
8180066: Convert package.html files to package-info.java

kevinrushforth:
This is a doc cleanup task to convert all `package.html` files to `package-info.java`, which is the recommended way to provide package documentation.

The first commit is the result of running a shell script that @jonathan-gibbons provided. The second commit removes an errant `<h2>` tag from the `package-info.java` for one package, `javafx.util.converter` that is unneeded and unwanted. That same file also had a redundant first sentence, which was coincidentally removed by the tool, so once I removed the `<h2>` tag the docs for that package now looks good and matches the other packages. There are no other changes in the javadoc-generated API docs.

One thing I didn't do is add a ` *` at the beginning of the lines that are now part of a javadoc-style comment block. This minimizes the diffs, but it means we won't be using the usual pattern for those comment blocks. I'm inclined to file a follow-up issue rather than changing that now.

---------
### Progress

### Issue
 * [JDK-8180066](https://bugs.openjdk.org/browse/JDK-8180066): Convert package.html files to package-info.java (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1159/head:pull/1159` \
`$ git checkout pull/1159`

`$ git checkout pull/1159` \
`$ git pull https://git.openjdk.org/jfx.git pull/1159/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1159`

`$ git pr show -t 1159`

</details>

<a href="https://git.openjdk.org/jfx/pull/1159.diff">https://git.openjdk.org/jfx/pull/1159.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1159#issuecomment-1601046672)


andy-goryachev-oracle:
String converters link got lost in allpackages-index and index-all.  is this intentional?

Screenshot 2023-06-21 at 11 34 39

andy-goryachev-oracle:
The rest of diff are not important as they all contain whitespace changes that are normally ignored by the browser.

kevinrushforth:
> String converters link got lost in allpackages-index and index-all. is this intentional?

It's the other way around. That link is _not_ present in the existing JavaFX API docs because of the redundant first sentence. It is now there because I got rid of the redundant first sentence (the one without the link) and the unwanted `<h2>` tag, making the following the first sentence:

```
 This package is for standard {@link javafx.util.StringConverter string converters}
 for JavaFX.
```

So the link wasn't lost, it was found.

kevinrushforth:
Unfortunately, with all of the whitespace diffs, it's now hard to see the actual changes. It might be worth looking at the diffs from the first two commits, and then separately confirming that the third commit is just whitespace changes.

andy-goryachev-oracle:
right, got my diff panels swapped.

so all the non-whitespace diffs have newfound link, we are good!

andy-goryachev-oracle (approved):
Looking good.
I guess we don't change the copyright year because we don't change the actual content?
Or should it be changed?

kevinrushforth:
> I guess we don't change the copyright year because we don't change the actual content? Or should it be changed?

Right. Even if we did want to change it, I would just let our copyright scripts update it, rather than doing it manually (which is somewhat error-prone), but in this case, we'll exclude this commit when running the tool, so we will skip the copyright year update altogether.

kevinrushforth:
> One thing I didn't do is add a  * at the beginning of the lines that are now part of a javadoc-style comment block. This minimizes the diffs, but it means we won't be using the usual pattern for those comment blocks. I'm inclined to file a follow-up issue rather than changing that now.

I filed [JDK-8310667](https://bugs.openjdk.org/browse/JDK-8310667) to track normalizing the javadoc comment blocks.

PR #1165, opened 2023-07-01
8311216: DataURI can lose information in some charset environments

mstr2:
DataURI uses the following implementation to decode the percent-encoded payload of a "data" URI:

```
...
String data = uri.substring(dataSeparator + 1);
Charset charset = Charset.defaultCharset();
...
URLDecoder.decode(data.replace("+", "%2B"), charset).getBytes(charset)
```

This approach only works if the charset that is passed into `URLDecoder.decode` and `String.getBytes` doesn't lose information when converting between `String` and `byte[]` representations, as might happen in a US-ASCII environment.

This PR solves the problem by not using `URLDecoder`, but instead simply decoding percent-encoded escape sequences as specified by RFC 3986, page 11.

**Note to reviewers**: the failing test can only be observed when the JVM uses a default charset that can't represent the payload, which can be enforced by specifying the `-Dfile.encoding=US-ASCII` VM option.

---------
### Progress

### Issue
 * [JDK-8311216](https://bugs.openjdk.org/browse/JDK-8311216): DataURI can lose information in some charset environments (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1165/head:pull/1165` \
`$ git checkout pull/1165`

`$ git checkout pull/1165` \
`$ git pull https://git.openjdk.org/jfx.git pull/1165/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1165`

`$ git pr show -t 1165`

</details>

<a href="https://git.openjdk.org/jfx/pull/1165.diff">https://git.openjdk.org/jfx/pull/1165.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1165#issuecomment-1616164021)

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/util/DataURI.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/util/DataURI.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2021, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -25,8 +25,9 @@
 
 package com.sun.javafx.util;
 
-import java.net.URLDecoder;
-import java.nio.charset.Charset;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
 import java.util.Arrays;
 import java.util.Base64;
 import java.util.Collections;
@@ -103,7 +104,6 @@ public class DataURI {
         }
 
         String data = uri.substring(dataSeparator + 1);
-        Charset charset = Charset.defaultCharset();
 
         return new DataURI(
             uri,
@@ -112,9 +112,7 @@ public class DataURI {
             mimeSubtype,
             nameValuePairs,
             base64,
-            base64 ?
-                Base64.getDecoder().decode(data) :
-                URLDecoder.decode(data.replace("+", "%2B"), charset).getBytes(charset));
+            base64 ? Base64.getDecoder().decode(data) : decodePercentEncoding(data));
     }
 
     private final String originalUri;
@@ -206,6 +204,69 @@ public class DataURI {
         int result = Objects.hash(mimeType, mimeSubtype, base64);
         result = 31 * result + Arrays.hashCode(data);
         return result;
+    }
+
+    private static byte[] decodePercentEncoding(String input) {
+        try (var output = new ByteArrayOutputStream(input.length())) {
+            decodePercentEncodingToStream(input, output);
+            return output.toByteArray();
+        } catch (IOException ignored) {
+            // can never happen for ByteArrayOutputStream
+            return null;
+        }
+    }
+
+    /**
+     * Decodes percent-encoded text as specified by RFC 3986, page 11.
+     * This method does not make any assumptions about the allowed character set.
+     */
+    private static void decodePercentEncodingToStream(String input, OutputStream output) throws IOException {
+        enum ParseState {
+            DEFAULT,
+            FIRST_HEX_DIGIT,
+            SECOND_HEX_DIGIT
+        }
+
+        ParseState parseState = ParseState.DEFAULT;
+        int firstDigit = 0;
+
+        for (int i = 0; i < input.length(); ++i) {
+            char c = input.charAt(i);
+
+            parseState = switch (parseState) {
+                case DEFAULT -> {
+                    if (c == '%') {
+                        yield ParseState.FIRST_HEX_DIGIT;
+                    } else {
+                        output.write(c);
+                        yield ParseState.DEFAULT;
+                    }
+                }
+
+                case FIRST_HEX_DIGIT -> {
+                    firstDigit = hexDigit(c);
+                    yield ParseState.SECOND_HEX_DIGIT;
+                }
+
+                case SECOND_HEX_DIGIT -> {
+                    output.write(firstDigit << 4 | hexDigit(c));
+                    yield ParseState.DEFAULT;
+                }
+            };
+        }
+
+        if (parseState != ParseState.DEFAULT) {
+            throw new IllegalArgumentException("Incomplete character escape sequence");
+        }
+    }
+
+    private static int hexDigit(char c) {
+        int digit = Character.digit(c, 16);
+        if (digit < 0) {
+            throw new IllegalArgumentException("Invalid symbol in character escape sequence");
+        }
+
+        return digit;
     }
 
 }
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/util/DataURI.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/util/DataURI.java
@@ -25,9 +25,6 @@
 
 package com.sun.javafx.util;
 
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.io.OutputStream;
 import java.util.Arrays;
 import java.util.Base64;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


modules/javafx.graphics/src/main/java/com/sun/javafx/util/DataURI.java
```diff
@@ -112,9 +112,7 @@ public static DataURI tryParse(String uri) {
             mimeSubtype,
             nameValuePairs,
             base64,
-            base64 ?
-                Base64.getDecoder().decode(data) :
-                URLDecoder.decode(data.replace("+", "%2B"), charset).getBytes(charset));
+            base64 ? Base64.getDecoder().decode(data) : decodePercentEncoding(data));
```
andy-goryachev-oracle, line 115: I wonder if this is all necessary.  The data is supposed to be url-encoded, so it's essentially ASCII, no?

passing default charset to getBytes() is not right, it probably should be

URLDecoder.decode(data.replace("+", "%2B"), charset).getBytes(StandardCharsets.US_ASCII));

or am I missing something?

andy-goryachev-oracle: From https://datatracker.ietf.org/doc/html/rfc3986#page-11

```
Therefore, the





Berners-Lee, et al.         Standards Track                    [Page 11]

[RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986)                   URI Generic Syntax               January 2005


   integer values used by the ABNF must be mapped back to their
   corresponding characters via US-ASCII in order to complete the syntax
   rules.
```

mstr2: > I wonder if this is all necessary. The data is supposed to be url-encoded, so it's essentially ASCII, no?
> 
> passing default charset to getBytes() is not right, it probably should be
> 
> URLDecoder.decode(data.replace("+", "%2B"), charset).getBytes(StandardCharsets.US_ASCII));
> 
> or am I missing something?

The payload of a data URI is just a sequence of bytes, not characters. Only when the numeric value of a byte, assuming ASCII encoding, is a *safe URL character*, it is left as-is; otherwise percent-encoding is used to encode the byte value. The [specification](https://datatracker.ietf.org/doc/html/rfc2397) points out:
```
Without ";base64", the data (as a sequence of octets) is represented using
ASCII encoding for octets inside the range of safe URL characters and using
the standard %xx hex encoding of URLs for octets outside that range.
```

Decoding the payload back to a byte array is done by simply converting each assumed ASCII character to its numeric value, and decoding percent-encoded bytes to their hex value. Note that the assumed ASCII encoding only refers to the URL, but not to the payload. The payload is not a string, and it doesn't contain characters; it's a sequence of bytes.

`URLDecoder` is not a general-purpose class to decode a percent-encoded sequence of bytes. It's specifically meant to take a HTML forms string and decode it into a string with some defined charset, using additional rules that don't generally apply to percent-encoded byte sequences. For example, a space character is encoded as `+` (that's where the kind-of-hacky `data.replace("+", "%2B")` comes from).

Using `URLDecoder` kind of works (if we use a sufficiently rich charset for both `URLDecoder.decode` and `String.getBytes`), but only by accident. While it accepts almost any percent-encoded data, the Javadoc for `URLDecoder` says:
```
There are two possible ways in which this decoder could deal with illegal strings.
It could either leave illegal characters alone or it could throw an IllegalArgumentException.
Which approach the decoder takes is left to the implementation.
```

modules/javafx.graphics/src/test/java/test/com/sun/javafx/util/DataURITest.java
```diff
@@ -174,4 +176,31 @@ public void testLongUriToString() {
         assertEquals("data:text/plain;charset=utf-8;base64,SGVsbG9Xb3JsZE...RIZWxsb1dvcmxk", uri.toString());
     }
 
+    @Test
+    public void testPercentEncodedTextIsDecodedAccordingToRFC3986() {
+        // We use URLEncoder here to escape the emoji character using percent-encoding.
+        // When DataURI parses its payload, it automatically converts percent-encoded characters back to octets.
+        String input = URLEncoder.encode("🙂", StandardCharsets.UTF_8);
```
andy-goryachev-oracle, line 183: would it make sense to try several different strings that include +, \n, \t, data:, charset:, %, empty string, &, _, %zz?

modules/javafx.graphics/src/main/java/com/sun/javafx/util/DataURI.java
andy-goryachev-oracle: thank you for clarifications!  your approach does make sense.

modules/javafx.graphics/src/test/java/test/com/sun/javafx/util/DataURITest.java
```diff
@@ -174,4 +176,31 @@ public void testLongUriToString() {
...
+    @Test
+    public void testPercentEncodedTextContainsInvalidEscapeSequence() {
+        var ex = assertThrows(IllegalArgumentException.class, () -> DataURI.tryParse("data:,%XY"));
+        assertTrue(ex.getMessage().startsWith("Invalid"));
+
+        ex = assertThrows(IllegalArgumentException.class, () -> DataURI.tryParse("data:,%5G"));
+        assertTrue(ex.getMessage().startsWith("Invalid"));
+
+        ex = assertThrows(IllegalArgumentException.class, () -> DataURI.tryParse("data:,%0"));
+        assertTrue(ex.getMessage().startsWith("Incomplete"));
```
andy-goryachev-oracle, line 203: "%", "", null ?

mstr2: I've added tests for these inputs.

modules/javafx.graphics/src/test/java/test/com/sun/javafx/util/DataURITest.java
mstr2: Most of these cases should already be covered by existing tests (`testMissingDataSeparatorIsInvalid`, `testParametersListWithoutKeyValuePairsIsInvalid`, `testLeadingOrTrailingWhitespaceIsAcceptable`).

andy-goryachev-oracle (approved):
+1 for added tests, thank you.
LGTM

kevinrushforth:
This one could use a second pair of eyes on it.

mstr2:
Let's keep this open.

modules/javafx.graphics/src/main/java/com/sun/javafx/util/DataURI.java
```diff
@@ -208,4 +206,67 @@ public int hashCode() {
         return result;
     }
 
+    private static byte[] decodePercentEncoding(String input) {
+        try (var output = new ByteArrayOutputStream(input.length())) {
```
hjohn, line 210: If deemed important to optimize this a bit:

This could allocate an array that's a factor 3 too large, which depending on how these URI's are used can become quite big (although for large amounts of data, base64 is recommended).

I noticed that in the URLDecoder they limit the size of the buffer `new StringBuilder(numChars > 500 ? numChars / 2 : numChars);` -- perhaps you could do something similar (`new ByteArrayOutputStream(input.length() > 500 ? input.length() / 2 : input.length())`)

Another alternative would be to calculate the exact size first (by counting `%`s) avoiding the need to reallocate the `byte[]` -- performance is (IMHO) likely to be better:

- ByteArrayOutputStream method
  - allocate output array: zeroes n*3 bytes
  - decode: reads whole string once (n to n*3 reads)
  - decode: writes n bytes
  - reallocate: reads whole output once (n)
  - reallocate: writes n bytes
  - garbage created: n*3 bytes
  - max memory use: n*3 + n bytes

Total read/writes: 4n + 5n

- Count method
  - count: reads whole string once (n to n*3 reads)
  - allocate output array: zeroes n bytes
  - decode: reads whole string again (n to n*3 reads)
  - decode: writes n bytes
  - garbage created: none
  - max memory use: n bytes

Total read/writes: 6n + 2n

modules/javafx.graphics/src/test/java/test/com/sun/javafx/util/DataURITest.java
hjohn: minor(?): Is it common practice for the FX code base to include non-ASCII characters?  If not, perhaps encode this instead of using the smiley directly (or use an ASCII character that requires % escaping, like `%`).

hjohn (approved):
This change is probably a good move; not only is URLDecoder not doing the right thing, it's not that efficient when decoding a large amount of data (as it was intended for short URL's).

modules/javafx.graphics/src/main/java/com/sun/javafx/util/DataURI.java
mstr2: Solved by calculating the exact size first.

hjohn: Thanks, I see you went this route, you could eliminate the `ByteArrayOutputStream` now, perhaps have the signature of `decodePercentEncodingToStream` changed to:

    private static byte[] decodePercentEncodingToStream(String input, int size)

hjohn (approved):

modules/javafx.graphics/src/main/java/com/sun/javafx/util/DataURI.java
mstr2: I didn't to exactly that, but something quite like it.

modules/javafx.graphics/src/main/java/com/sun/javafx/util/DataURI.java
```diff
@@ -207,55 +207,75 @@ public int hashCode() {
     }
 
     private static byte[] decodePercentEncoding(String input) {
-        try (var output = new ByteArrayOutputStream(input.length())) {
+        try (var output = new ByteArrayOutputStream(size(input))) {
             decodePercentEncodingToStream(input, output);
```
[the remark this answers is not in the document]
mstr2: I removed `ByteArrayOutputStream`.

hjohn (approved):
LGTM

modules/javafx.graphics/src/main/java/com/sun/javafx/util/DataURI.java
hjohn: Was just a suggestion, I like your solution!

modules/javafx.graphics/src/main/java/com/sun/javafx/util/DataURI.java
```diff
@@ -208,4 +203,83 @@ public int hashCode() {
...
+     */
+    private static byte[] decodePercentEncoding(String input) {
+        enum ExpectedCharacter {
+            DEFAULT,
+            FIRST_HEX_DIGIT,
+            SECOND_HEX_DIGIT
+        }
+
+        ExpectedCharacter expectedCharacter = ExpectedCharacter.DEFAULT;
+        byte[] output = new byte[computePayloadSize(input)];
```
andy-goryachev-oracle, line 221: just a thought: what if the url contains no %?  in this case one can simply return the input string.
I think we could return -1 from computePayloadSize() to indicate that condition.

mstr2: The return value is not a string, it's a byte array. So we need to do the conversion anyway, which is what this method is doing quite efficiently by now. It's probably not useful to use another conversion method instead, at least not in terms of performance.

modules/javafx.graphics/src/test/java/test/com/sun/javafx/util/DataURITest.java
```diff
@@ -174,4 +186,34 @@ public void testLongUriToString() {
...
+
+        ex = assertThrows(IllegalArgumentException.class, () -> DataURI.tryParse("data:,%"));
+        assertTrue(ex.getMessage().startsWith("Incomplete"));
+    }
+
```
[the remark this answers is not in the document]
mstr2: I haven't been able to find out whether there's a recommended style in OpenJFX with regards to empty lines at the end of a class declaration. The codebase is very inconsistent about that...

PR #1172, opened 2023-07-07
8285700: [TreeTableView] graphic property of TreeItem is still visible after collapsing tree

karthikpandelu:
In `TreeTableRowSkin`, graphic was not updated along with tree item update.

Made changes to update graphics of TreeTableView row in `updateTreeItem()` method.

Added options in monkey tester to add graphics and subnodes to `TreeTableView` rows.

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
@@ -343,6 +343,7 @@ public class TreeTableRowSkin<T> extends TableRowSkinBase<TreeItem<T>, TreeTable
         unregisterInvalidationListeners(graphicProperty());
         treeItem = getSkinnable().getTreeItem();
         registerInvalidationListener(graphicProperty(), e -> updateTreeItemGraphic());
+        updateDisclosureNodeAndGraphic();
     }
 
     private TreeTableView<T> getTreeTableView() {
--- a/tests/manual/monkey/src/com/oracle/tools/fx/monkey/pages/TreeTableViewPage.java
+++ b/tests/manual/monkey/src/com/oracle/tools/fx/monkey/pages/TreeTableViewPage.java
@@ -39,6 +39,8 @@ import javafx.scene.control.TreeTableView;
 import javafx.scene.control.TreeTableView.ResizeFeatures;
 import javafx.scene.layout.BorderPane;
 import javafx.scene.layout.Pane;
+import javafx.scene.paint.Color;
+import javafx.scene.shape.Rectangle;
 import javafx.scene.text.Text;
 import javafx.util.Callback;
 import com.oracle.tools.fx.monkey.util.OptionPane;
@@ -114,6 +116,8 @@ public class TreeTableViewPage extends TestPaneBase {
     protected final ComboBox<ResizePolicy> policySelector;
     protected final ComboBox<Selection> selectionSelector;
     protected final CheckBox nullFocusModel;
+    protected final CheckBox addGraphics;
+    protected final CheckBox addSubNodes;
     protected TreeTableView<String> tree;
 
     public TreeTableViewPage() {
@@ -150,6 +154,18 @@ public class TreeTableViewPage extends TestPaneBase {
             updatePane();
         });
 
+        addGraphics = new CheckBox("add graphics");
+        addGraphics.setId("addGraphics");
+        addGraphics.selectedProperty().addListener((s, p, c) -> {
+            updatePane();
+        });
+
+        addSubNodes = new CheckBox("add sub-nodes");
+        addSubNodes.setId("addSubNodes");
+        addSubNodes.selectedProperty().addListener((s, p, c) -> {
+            updatePane();
+        });
+
         Button clearButton = new Button("Clear Items");
         clearButton.setOnAction((ev) -> {
             tree.setRoot(new TreeItem(null));
@@ -167,6 +183,8 @@ public class TreeTableViewPage extends TestPaneBase {
         p.label("Selection Model:");
         p.option(selectionSelector);
         p.option(nullFocusModel);
+        p.option(addGraphics);
+        p.option(addSubNodes);
         setOptions(p);
 
         demoSelector.getSelectionModel().selectFirst();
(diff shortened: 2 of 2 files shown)


andy-goryachev-oracle (approved):
tested using the modified MonkeyTester, works with the fix and fails in the master branch.
good job!

Maran23:
Is it possible to write a Unit test as well?

johanvos:
+1 on @Maran23 to add a unit test.
For these kinds of changes, I think it is important to have a performance check as well (e.g. count the number of invocations to updateItem etc).

karthikpandelu:
> Is it possible to write a Unit test as well?

I tried to write unit test but since it should be checked if the graphic is cleared while collapsing the tree and added back while expanding, I couldn't find out a way to do this. I tried to use `VirtualFlowTestUtils.assertGraphicIsVisible()` method similar to how it is used in `TreeTableViewTest` to do the same but it did not help in this case. Please let me know if you have any suggestions.

> For these kinds of changes, I think it is important to have a performance check as well (e.g. count the number of invocations to updateItem etc).

I have checked my changes using the program given in [JDK-8143266](https://bugs.openjdk.org/browse/JDK-8143266). I did not find performance degradation as a result of my changes. Do you suggest writing separate test for this? Please let me know your suggestion.

kleopatra:
> > Is it possible to write a Unit test as well?
> 
> I tried to write unit test but since it should be checked if the graphic is cleared while collapsing the tree and added back while expanding, I couldn't find out a way to do this. 

I would try to test the state of the graphic of the concrete item - keep track of it on setup and before showing the treetable - it should/not be part of the visible hierarchy (as appropriate) after expanding/collapsing. When it should be visible also verify that it is at the expected location.

karthikpandelu:
Added unit test as suggested above.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableRowSkinTest.java
```diff
@@ -356,6 +358,51 @@ public void cellsMustBeCollectableAfterRowFactoryChange() {
...
+        assertNotEquals(treeItemSubNodeAfterExpand, treeItemSubNode.getGraphic().localToScene(treeItemSubNode.getGraphic().getBoundsInLocal()).getMinY());
+
+        treeItem.setExpanded(true);
+        Toolkit.getToolkit().firePulse();
+        assertTrue(treeItem.isExpanded());
+        assertEquals(treeItemSubNodeAfterExpand, treeItemSubNode.getGraphic().localToScene(treeItemSubNode.getGraphic().getBoundsInLocal()).getMinY());
+    }
+
     @AfterEach
     public void after() {
```
andy-goryachev-oracle, line 407: unrelated to the current PR, but could this method be more defensive, to avoid any problems with the invocation order?

```
    @AfterEach
    public void after() {
        if (stageLoader != null) {
            stageLoader.dispose();
            stageLoader = null;
        }
    }
```

andy-goryachev-oracle (approved):
the test passes with the fix, fails in the master branch.

one minor suggestion, unrelated, but probably worth addressing in to this PR.

andy-goryachev-oracle:
github says it fails pre-submit check - probably some kind of spurious failure, local build and headful tests (minus web tests) are ok on macOS.

karthikpandelu:
> github says it fails pre-submit check - probably some kind of spurious failure, local build and headful tests (minus web tests) are ok on macOS.

Ran the pre-submit checks again. All checks are passing now

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableRowSkinTest.java
karthikpandelu: This looks like a good check to have.
Made above suggested code changes.

karthikpandelu:
@johanvos could you please review the latest updates?

johanvos (approved):
Looks ok now. Thanks @kleopatra for the guidance on the test.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableRowSkinTest.java
```diff
@@ -356,9 +358,57 @@ public void cellsMustBeCollectableAfterRowFactoryChange() {
...
+        for (int i = 0; i < 4; i++) {
+            TreeItem parent = new TreeItem("item - " + i, new Rectangle(10, 10));
+            TreeItem subNode = new TreeItem("sub-node - " + i, new Rectangle(10, 10));
+            parent.getChildren().addAll(subNode);
+            tree.getRoot().getChildren().add(parent);
+        }
+        tree.getRoot().setExpanded(true);
+
+        int itemNodeSize = tree.getRoot().getChildren().size();
+        TreeItem treeItem = (TreeItem)tree.getRoot().getChildren().get(itemNodeSize - 2);
```
Maran23, line 380: Cast is not needed here and below

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableRowSkinTest.java
```diff
@@ -356,9 +358,57 @@ public void cellsMustBeCollectableAfterRowFactoryChange() {
...
+    public void testGraphicVisibilityOfTreeItem() {
+        TreeTableView<String> tree = new TreeTableView<>();
+        stageLoader = new StageLoader(tree);
+
+        TreeTableColumn<String, String> column = new TreeTableColumn<>("Column");
+        tree.getColumns().add(column);
+        tree.setRoot(new TreeItem("Root"));
+
+        for (int i = 0; i < 4; i++) {
+            TreeItem parent = new TreeItem("item - " + i, new Rectangle(10, 10));
```
Maran23, line 372: Generics should be added here

karthikpandelu:
Made changes according to @Maran23's review comments.
Please re-approve the changes.

Maran23 (approved):
Looks good. Can confirm that the test fails before.
Thanks!

andy-goryachev-oracle (approved):
re-approving latest changes.

karthikpandelu:
Thanks @andy-goryachev-oracle , @johanvos and @Maran23 for the review and thanks @kleopatra for the guidance on the the test.

PR #1173, opened 2023-07-09
8251240: Menus inaccessible on Linux with i3 wm

tsayao:
The bug happens because `gdk_window_get_frame_extents` is not returning the correct position when o i3-wm, probably by some bug on the wm itself.

Te fix replaces the usage of the function for already known extents value calculation.

---------
### Progress

### Issue
 * [JDK-8251240](https://bugs.openjdk.org/browse/JDK-8251240): Menus inaccessible on Linux with i3 wm (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1173/head:pull/1173` \
`$ git checkout pull/1173`

`$ git checkout pull/1173` \
`$ git pull https://git.openjdk.org/jfx.git pull/1173/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1173`

`$ git pr show -t 1173`

</details>

<a href="https://git.openjdk.org/jfx/pull/1173.diff">https://git.openjdk.org/jfx/pull/1173.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1173#issuecomment-1627782537)


jperedadnr (approved):
Looks good to me!
Tested on Ubuntu and Ubuntu with i3 window manager. 
Passes on the latter after this patch, and fails before it.

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
```diff
@@ -1010,14 +1010,10 @@ void WindowContextTop::process_configure(GdkEventConfigure* event) {
     }
 
     int x, y;
-
+    gdk_window_get_origin(gdk_window, &x, &y);
     if (frame_type == TITLED) {
-        GdkRectangle rect;
-        gdk_window_get_frame_extents(gdk_window, &rect);
```
johanvos, line 1016: Looking at potential regression: in case of a TITLED frame, before this patch you would use `get_frame_extents` and now you will use `get_origin` in all cases. Are there scenario's where this might lead to regression, or is there a way to test this?

tsayao: `geometry.x` and `y` should point to the top left corner of the window including decorations (on JavaFX logic), then the view position will have the decoration size as x and y. `gdk_window_get_frame_extents` returns the top left position with frame extents (decorations) - all other Gdk/Gtk functions will return position without decorations (because on X11 they are drawn on client side). Somehow `gdk_window_get_frame_extents` is broken on i3-wm, but we already have the frame extents values. I just replaced it by getting normal origins (does not include decorations) and if the frame is TITLED, subtract the frame extents to have the window position accounting decorations.

To test this I:

- [ ] Run the bug example with UNDECORATED window on i3 and mutter (the default gnome wm);
- [ ] Run the bug example with DECORATED window on i3 and mutter (the default gnome wm);

tsayao:
Can someone sponsor this?

johanvos (approved):
I am very careful when generic glass code is changed, but with the explanation added, I don't see a potential case for regression, so +1 from me.
We should encourage developers to use 22-ea builds to make sure there is indeed no regression.

kevinrushforth:
@tsayao This is ready to integrate when you are ready.

PR #1174, opened 2023-07-10
8311806: Class ButtonAccessibility is implemented twice

azuev-java:
To avoid confusion on the os x dynamic linker side i renamed native classes from ButtonAccessibility to JFXButtonAccessibility. I will use JFX prefix down the line to avoid any confusion when running with the latest JDK that also migrated to the new a11y API.

---------
### Progress

### Issue
 * [JDK-8311806](https://bugs.openjdk.org/browse/JDK-8311806): Class ButtonAccessibility is implemented twice (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1174/head:pull/1174` \
`$ git checkout pull/1174`

`$ git checkout pull/1174` \
`$ git pull https://git.openjdk.org/jfx.git pull/1174/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1174`

`$ git pr show -t 1174`

</details>

<a href="https://git.openjdk.org/jfx/pull/1174.diff">https://git.openjdk.org/jfx/pull/1174.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1174#issuecomment-1629706252)

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/mac/a11y/AccessibleBase.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/a11y/AccessibleBase.m
@@ -40,10 +40,10 @@ static NSMutableDictionary * rolesMap;
      */
     rolesMap = [[NSMutableDictionary alloc] initWithCapacity:4];
 
-    [rolesMap setObject:@"ButtonAccessibility" forKey:@"BUTTON"];
-    [rolesMap setObject:@"ButtonAccessibility" forKey:@"DECREMENT_BUTTON"];
-    [rolesMap setObject:@"ButtonAccessibility" forKey:@"INCREMENT_BUTTON"];
-    [rolesMap setObject:@"ButtonAccessibility" forKey:@"SPLIT_MENU_BUTTON"];
+    [rolesMap setObject:@"JFXButtonAccessibility" forKey:@"BUTTON"];
+    [rolesMap setObject:@"JFXButtonAccessibility" forKey:@"DECREMENT_BUTTON"];
+    [rolesMap setObject:@"JFXButtonAccessibility" forKey:@"INCREMENT_BUTTON"];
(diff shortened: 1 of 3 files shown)


kevinrushforth:
The fix looks fine and should avoid the problem of name collision. It seems unfortunate that the namespace is global, especially in this case, since the symbols are not needed outside of libglass.dylib. Is there a way to hide the symbols so that they are not exported? If not, then adopting the pattern you chose is OK.

azuev-java:
> Is there a way to hide the symbols so that they are not exported?

I do not think we can hide these symbols from the library - they have to be public for us to instantiate them dynamically based on the accessibility role of the FX node.

arapte:
May I suggest an alternate naming pattern similar to AccessibleBase.h.
Can we name our files as AccessibleButton, (AccessibleList, ...  and so forth) ?

azuev-java:
> May I suggest an alternate naming pattern similar to AccessibleBase.h. Can we name our files as AccessibleButton, (AccessibleList, ... and so forth) ?

The slight issue with this approach is that in my opinion the class names will be too similar for the first look which will make code less readable. Also that will make code in JDK and JFX less similar which might be a problem for anyone who will support accessibility on both sides. For that reason i'm trying to keep naming and structure of the files as close as possible.

kevinrushforth:
> The slight issue with this approach is that in my opinion the class names will be too similar for the first look which will make code less readable. Also that will make code in JDK and JFX less similar which might be a problem for anyone who will support accessibility on both sides. For that reason i'm trying to keep naming and structure of the files as close as possible.

OK, that  explanation makes sense to me.

kevinrushforth (approved):

arapte (approved):

PR #1069, opened 2023-03-25
JDK-8304439: Subscription based listeners

hjohn:
Makes `Subscription` public (removing some of its methods that are unnecessary), and adds methods that can provide `Subscription`s in `ObservableValue`.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))
- [x] Change requires CSR request [JDK-8311123](https://bugs.openjdk.org/browse/JDK-8311123) to be approved

### Issues
 * [JDK-8304439](https://bugs.openjdk.org/browse/JDK-8304439): Subscription based listeners (**Enhancement** - P4)
 * [JDK-8311123](https://bugs.openjdk.org/browse/JDK-8311123): Subscription based listeners (**CSR**)

### Reviewers
 * [Nir Lisker](https://openjdk.org/census#nlisker) (@nlisker - **Reviewer**)
 * [Kevin Rushforth](https://openjdk.org/census#kcr) (@kevinrushforth - **Reviewer**)
 * [Michael Strauß](https://openjdk.org/census#mstrauss) (@mstr2 - Committer)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1069/head:pull/1069` \
`$ git checkout pull/1069`

Update a local copy of the PR: \
`$ git checkout pull/1069` \
`$ git pull https://git.openjdk.org/jfx.git pull/1069/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 1069`

View PR using the GUI difftool: \
`$ git pr show -t 1069`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jfx/pull/1069.diff">https://git.openjdk.org/jfx/pull/1069.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1069#issuecomment-1483863302)

PR #1177, opened 2023-07-13
JDK-8312058: Documentation improvements for subscription based listeners

hjohn:
Incorporates documentation review comments from #1069.

This PR should be kept minimal so it can be backported to JFX21; it should only contain documentation changes.

---------
### Progress

### Issue
 * [JDK-8312058](https://bugs.openjdk.org/browse/JDK-8312058): Documentation improvements for subscription based listeners (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1177/head:pull/1177` \
`$ git checkout pull/1177`

`$ git checkout pull/1177` \
`$ git pull https://git.openjdk.org/jfx.git pull/1177/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1177`

`$ git pr show -t 1177`

</details>

<a href="https://git.openjdk.org/jfx/pull/1177.diff">https://git.openjdk.org/jfx/pull/1177.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1177#issuecomment-1634932804)


kevinrushforth:
As long as this is limited to clarification of the existing docs, no CSR is needed.

modules/javafx.base/src/main/java/javafx/beans/Subscription.java
```diff
@@ -59,7 +59,8 @@ static Subscription combine(Subscription... subscriptions) {
     /**
      * Cancels this subscription, or does nothing if already cancelled.<p>
      *
-     * Implementors must ensure the implementation is idempotent.
+     * Implementors must ensure the implementation is idempotent (a no-op
+     * if called more than once).
```
nlisker, line 63: I now think this should be in an `@implSpec` tag since it specifies the implementation conditions.

kevinrushforth (commented):
These changes look good to me, pending the addition of the `@implSpec` as noted by @nlisker 

If we decide to move the Subscription class to `javafx.util`, additional changes would be done as part of that fix (which would need a CSR).

kevinrushforth:
Now that `Subscription` has moved to `javafx.util`, this PR can resume. You first need to merge in the latest master and fix any merge conflicts.

andy-goryachev-oracle (approved):
LvGtm

mstr2 (approved):
LGTM

nlisker:
I think that the documentation on `Subscription` still needs to be improved beyond a single sentence. It should include how to use the interface and give use cases. I'll need to think about a good first paragraph. For the use cases, maybe the ones you outlines in https://github.com/openjdk/jfx/pull/1069/files#r1263857263. I also think that a short review of the methods it offers is useful:
```
Subscriptions can also be combined using {@link #combine} and {@link #and}, which allows for
multiple subscriptions to be unsubscribed together. This is useful when they share the same
lifecycle, for example, when performing cleanup for the same object.
```

Then there is the question of what to do with regards to its use in JavaFX. The previous description I gave in https://github.com/openjdk/jfx/pull/1069#discussion_r1261873147 is centered around `Observable/ObservableValue`. I think that it's still better to include this description somewhere rather than repeat a part of it in every method. Maybe it fits better in `Observable/ObservableValue`.

hjohn:
> I think that the documentation on `Subscription` still needs to be improved beyond a single sentence. It should include how to use the interface and give use cases. I'll need to think about a good first paragraph. For the use cases, maybe the ones you outlines in https://github.com/openjdk/jfx/pull/1069/files#r1263857263. I also think that a short review of the methods it offers is useful:
> 
> ```
> Subscriptions can also be combined using {@link #combine} and {@link #and}, which allows for
> multiple subscriptions to be unsubscribed together. This is useful when they share the same
> lifecycle, for example, when performing cleanup for the same object.
> ```
> 
> Then there is the question of what to do with regards to its use in JavaFX. The previous description I gave in [#1069 (comment)](https://github.com/openjdk/jfx/pull/1069#discussion_r1261873147) is centered around `Observable/ObservableValue`. I think that it's still better to include this description somewhere rather than repeat a part of it in every method. Maybe it fits better in `Observable/ObservableValue`.

I took a stab at creating a good first paragraph, and used your suggestion as the second paragraph. What do you think?

modules/javafx.base/src/main/java/javafx/util/Subscription.java
```diff
@@ -29,8 +29,17 @@
 import java.util.Objects;
 
 /**
- * A subscription encapsulates how to cancel it without having
- * to keep track of how it was created.
+ * A {@code Subscription} represents the encapsulation of the result of an action
+ * that can be cancelled. Subscriptions can be obtained, for example, as a result
```
nlisker, line 33: The first sentence is a bit cryptic to me. I'm looking at `Runnable` that is basically a less fancy version of `Subscription` (in a way, `Subscription` could extend `Runnable`). Maybe something like
"Represents a cancel/cleanup operation for an action that can be cancelled/allocates resources".
You can also choose either.

I would also add a general example after the first paragraph, maybe like this:

```java
class Person {

    public Subscription subscribeToNewspaper(Newspaper paper) {
        register(paper);
        return () -> unregister(paper);
    }
```

hjohn: Thanks, that's a good suggestion.  I also thought the wording was a bit too hard to follow, but wanted to see what you thought.  I updated it to use your wording and added an example.

modules/javafx.base/src/main/java/javafx/util/Subscription.java
```diff
@@ -29,13 +29,24 @@
 import java.util.Objects;
 
 /**
- * A {@code Subscription} represents the encapsulation of the result of an action
- * that can be cancelled. Subscriptions can be obtained, for example, as a result
+ * Represents a cancel or cleanup operation for an action be cancelled or
+ * allocated resources. Subscriptions can be obtained, for example, as a result
```
nlisker, line 33: Somethign is missing in the sentence, should it be "...for an action that can be cancelled or that allocates resources"

hjohn: Sorry, missed that somehow, fixed.

nlisker:
Looks good now. Do you still prefer to repeat the behavior of the subscription in each of the new methods over the centralized one in https://github.com/openjdk/jfx/pull/1069#discussion_r1261873147?

modules/javafx.base/src/main/java/javafx/beans/value/ObservableValue.java
```diff
@@ -338,9 +340,11 @@ default Subscription subscribe(BiConsumer<? super T, ? super T> changeSubscriber
...
 
     /**
-     * Creates a {@link Subscription} on this value which immediately provides
-     * the current value to the given {@code valueSubscriber}, followed by any
-     * subsequent values whenever its value changes.
+     * Creates a {@code Subscription} on this {@code ObservableValue} which immediately
+     * provides the current value to the given {@code valueSubscriber}, followed by any
+     * subsequent values whenever its value changes. The {@code valueSubscriber} is called
+     * for convenience since usually the user will want to initialize a value
+     * and then update on changes.
```
kevinrushforth, line 347: Only the initial call is for convenience. Subsequent calls are necessary. Do you think it's worth calling this out? Something like:

```
     * ...The {@code valueSubscriber} is called immediately
     * for convenience, since usually the user will want to initialize a value
     * and then update on changes.
```

(I  might also add a comma before since, but that is more a matter of style)

kevinrushforth (approved):
Looks good. I made one suggestion for you to consider.

hjohn:
> Looks good now. Do you still prefer to repeat the behavior of the subscription in each of the new methods over the centralized one in [#1069 (comment)](https://github.com/openjdk/jfx/pull/1069#discussion_r1261873147)?

I'm really not quite sure where to put that.  I don't think it would go well with `Subscription` since it assumes observables, and things like `ObservableValue#when`.  We'd need to word it a lot more general then, and it would be hard to refer to `ObservableValue#when`, or general rules about garbage collection (I think that would depend on the method providing the subscriptions, which may not have such restrictions).

Reading `ObservableValue` documentation though, I think it could use some updating.  It mentions change and invalidation events, and their listeners, but nothing about the subscriptions as of yet.

kevinrushforth (approved):

nlisker:
> > Looks good now. Do you still prefer to repeat the behavior of the subscription in each of the new methods over the centralized one in [#1069 (comment)](https://github.com/openjdk/jfx/pull/1069#discussion_r1261873147)?
> 
> I'm really not quite sure where to put that. I don't think it would go well with `Subscription` since it assumes observables, and things like `ObservableValue#when`. We'd need to word it a lot more general then, and it would be hard to refer to `ObservableValue#when`, or general rules about garbage collection (I think that would depend on the method providing the subscriptions, which may not have such restrictions).
> 
> Reading `ObservableValue` documentation though, I think it could use some updating. It mentions change and invalidation events, and their listeners, but nothing about the subscriptions as of yet.

Yes, on `Observable` (invalidation) or `ObservableValue` (change) is the place I was thinking of.

andy-goryachev-oracle (approved):
changes since last review look good.

mstr2 (approved):

kevinrushforth:
I think we should get this in sooner than later, so we can backport it to jdk21 while there is still time. We can file a follow-up issue for any further changes, as long as we are all happy with the state of what's here now.

@nlisker Any other feedback?

nlisker:
> @nlisker Any other feedback?

Not anything beyond the comment on updating `Observable/Value`'s documentation for the new methods.

hjohn:
I think that might be better; changing docs on `Observable` and `ObservableValue` is a good idea, but may take some time to get agreement on. I would like to also make clear in such a change when you can expect listener / subscribers to fire, basically documenting the undocumented rules we already have (invalidations fire first then changes; listeners fire in the order they were added, allowing vetoing (which is used in and outside of JavaFX).

The general rules would be something like:
- Observers on super types fire before observers on sub types (ie. `Observable` before `ObservableValue`)
- Observers added on a single type fire in the order they were added (ie. value subscribers, change subscribers and change listeners are not separate lists as they're all offered by `ObservableValue`)

I've done an attempt to update the `Observable` class documentation with all recent developments:

```java
/**
 * An {@code Observable} is an entity that wraps content and allows to
 * observe the content for invalidations.
 *
 * <h2>Lazy evaluation and events</h2>
 *
 * An implementation of {@code Observable} may support lazy evaluation,
 * which means that the content is not immediately recomputed after changes, but
 * lazily the next time it is requested. All bindings and properties in
 * this library support lazy evaluation.
 * <p>
 * Implementations of this class should strive to generate as few events as
 * possible to avoid wasting too much time in event handlers. Implementations in
 * this library mark themselves as invalid when the first invalidation event
 * occurs. They do not generate anymore invalidation events until their value is
 * recomputed and valid again.
 *
 * <h2>Observing an {@code Observable}</h2>
 *
 * Observation is possible using listeners and subscribers. These are
 * interchangeable and have the same semantics:
 * <li>Observers are notified in the order they were added</li>
 * <li>When an observer is registered more than once, it will be notified more than once</li>
 * <li>Observers may be safely registered for different {@code Observable}s</li>
 * <p>
 * When observing an {@code Observable}, it will hold a strong reference to the observer which will
 * prevent it from being garbage collected as long as the {@code Observable} is reachable.
 * If the lifecycle of the {@code Observable} is longer than a registered observer, care
 * must be taken to unregister the observer after use to avoid unwanted retention of the
 * observer, and any other objects it may refer to, in memory (also known as a memory leak).
 *
 * <h2>Unregistering observers</h2>
 *
 * Unregistering observers when their use ends can be achieved in several ways:
 * <li>For listeners, call the corresponding {@code removeListener} method, ensuring to pass the exact same instance as was used to register it</li>
 * <li>For subscribers, call {@link Subscription#unsubscribe()} on each {@link Subscription}, or on a combined subscription</li>
 * <li>If the {@code Observable} is an {@code ObservableValue}, {@link ObservableValue#when(ObservableValue)} can be used to automatically
 *     decouple the lifecycle of the observer from the original {@code ObservableValue} when a condition (like visibility) holds:
 *     <pre>
 *     {@code
 *     observableValue.when(condition).subscribe(...)
 *     }</pre>
 * </li>
 * <li>For listeners, it's possible to use a weak wrapper, like {@link WeakInvalidationListener}, to automatically unregister the listener when it becomes weakly reachable</li>
 * <p>
 * Note that the use of weak listener wrappers is no longer recommended when it can be
 * avoided as it requires intimate knowledge of how weak references work in Java and how
 * they interact with the garbage collector. The timing of when a weak listener is
 * unregistered may be very different from the intended or expected timing. It can
 * depend on many external factors, including the JVM used, its version, its (GC) settings,
 * but also on the system type and its memory availability.
 * <li>Garbage collection may be deferred for long periods (or even indefinitely), resulting in weak listeners being removed much later than expected;
 * until such time, they will keep receiving events, and hold on to any referenced objects</li>
 * <li>Weak listeners may be cleaned up almost immediately, possibly resulting in a listener being removed if care wasn't taken
 * to reference it directly</li>
 * <li>Registering the same or a similar weak listener again may see strange behavior as a previous weak listener may still be present</li>
 * <p>
 * The above can lead to hard to reproduce issues that may vary from system to system.
 * <p>
 * It is therefore highly recommended to always remove observers which observe an {@code Observable}
 * that have a longer lifecycle. This can be done either manually (in a clean up or dispose method),
 * or automatically using a condition with {@code ObservableValue#when}.
 *
 * @see javafx.beans.value.ObservableValue
 * @see javafx.collections.ObservableList
 * @see javafx.collections.ObservableMap
 *
 *
 * @since JavaFX 2.0
 */
```

nlisker (approved):
The changes to `Observable/Value` seem significant enough to not be part of this change. Approving the current changes.

PR #1178, opened 2023-07-18
8255248: NullPointerException in JFXPanel due to race condition in HostContainer

prsadhuk:
Due to transient datatype of scenePeer, it can become null which can result in NPE in scenarios where scene is continuously been reset and set, which warrants a null check, as is done in other places for the same variable.

---------
### Progress

### Issue
 * [JDK-8255248](https://bugs.openjdk.org/browse/JDK-8255248): NullPointerException in JFXPanel due to race condition in HostContainer (**Bug** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1178/head:pull/1178` \
`$ git checkout pull/1178`

`$ git checkout pull/1178` \
`$ git pull https://git.openjdk.org/jfx.git pull/1178/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1178`

`$ git pr show -t 1178`

</details>

<a href="https://git.openjdk.org/jfx/pull/1178.diff">https://git.openjdk.org/jfx/pull/1178.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1178#issuecomment-1639553112)


aghaisas:
This avoids the NPE.

Overall, I think that we should choose between below options- 
- Option 1 : A null check should be added to every usage of `scenePeer` and `stagePeer` as close to the usage as possible.
- Option 2: The usage of `scenePeer` and `stagePeer` should be `synchronized`.

I am not well versed with this area of the code and hence cannot comment on ramifications of Option 2.

Also, can we add a test? (either automated or manual) - I see that there is a sample in JBS that demonstrates this issue.

mstr2:
The proposed patch does not appear to fix the issue. Here's how the code currently looks like (after the patch is applied):

```java
786    @Override
787    protected void paintComponent(Graphics g) {
788        if (scenePeer == null) {
789            return;
790        }
791        if (pixelsIm == null) {
792            createResizePixelBuffer(scaleFactorX, scaleFactorY);
793            if (pixelsIm == null) {
794                return;
795            }
796        }
797        DataBufferInt dataBuf = (DataBufferInt)pixelsIm.getRaster().getDataBuffer();
798        int[] pixelsData = dataBuf.getData();
799        IntBuffer buf = IntBuffer.wrap(pixelsData);
800        if (scenePeer != null && !scenePeer.getPixels(buf, pWidth, pHeight)) {
801            // In this case we just render what we have so far in the buffer.
802        }
```

Note that in L788, we return early if `scenePeer` is null. The fact that `scenePeer` can sometimes be null in L800 means that the field is changed on a different thread. Since that's the case, no amount of null checking will ever be good enough: `scenePeer` can just as easily be null after the check succeeds in L800.

Similiar issues can be observed in other places, for example later in the same file:
```java
    private class HostContainer implements HostInterface {
        @Override
        public void setEmbeddedScene(EmbeddedSceneInterface embeddedScene) {
            ...
            scenePeer = embeddedScene;
            ....
            invokeOnClientEDT(() -> {
                ....
                if (scenePeer != null) {
                    scenePeer.setDragStartListener(dnd.getDragStartListener());
                }
            });
        }
    }
```

Note that `scenePeer` is assigned on the thread that calls `setEmbeddedScene`, and then its value is read from the EDT.

Since it appears that `JFXPanel` is inherently multi-threaded, every usage of an unprotected shared field is potentially faulty. At the very least, every field that can be written to or read from multiple threads must be protected by a lock. Note that is is not enough to just protect all writes; all reads have to be protected as well.

In addition to that, we need to account for concurrent code execution. What happens if thread A runs code on `scenePeer`, while thread B has already created a new scene peer? Is this safe? If not, then we need to prevent such concurrent execution.

modules/javafx.swing/src/main/java/javafx/embed/swing/JFXPanel.java
```diff
@@ -405,8 +407,25 @@ private Point2D convertSwingToFxPixel(GraphicsConfiguration g, double wx, double
...
+        synchronized(LOCK) {
+            lStagePeer = stagePeer;
+        }
+        return lStagePeer;
+    }
+
     private void sendMouseEventToFX(MouseEvent e) {
-        if (scenePeer == null || !isFxEnabled()) {
+        EmbeddedSceneInterface lScenePeer = getScenePeer();
+        if (lScenePeer == null || !isFxEnabled()) {
```
Maran23, line 428: Probably minor, but maybe we should first check `!isFxEnabled()` and return fast before we actually call `getScenePeer()` and thus locking it

prsadhuk: ok..sure..

mstr2:
After looking at the class for a bit longer and seeing how its state is read and mutated on different threads, it might be a better approach to enforce serialized code execution in all cases. This might be done as follows:

1. Remove all `volatile` modifiers from the fields of the class. (Also, make `disableCount` an `int` field.)
2. Mark all non-private methods which are entry points into the class with the `synchronized` modifier. Private methods should not be marked with `synchronized`.
3. Wrap code in all other entry points (lambdas, anonymous implementations) in a `synchronized` block that synchronizes on the `JFXPanel` instance.

If done correctly, no method of the class will ever run concurrently with another method of the class, and the memory effects of `synchronized` will ensure that we can always see the latest field values. What do you think about this approach?

modules/javafx.swing/src/main/java/javafx/embed/swing/JFXPanel.java
```diff
@@ -1029,15 +1085,22 @@ public void setEmbeddedScene(EmbeddedSceneInterface embeddedScene) {
                 return;
             }
             if (pWidth > 0 && pHeight > 0) {
-                scenePeer.setSize(pWidth, pHeight);
+                lScenePeer.setSize(pWidth, pHeight);
+            }
+            lScenePeer.setPixelScaleFactors((float) scaleFactorX, (float) scaleFactorY);
+            synchronized(LOCK) {
+                scenePeer = lScenePeer;
```
andy-goryachev-oracle, line 1092: I wonder if the whole method should be synchronized?
the logic around scenePeer is asking to be atomic, isn't it?

tests/system/src/test/java/test/javafx/embed/swing/JFXPanelNPETest.java
```diff
@@ -0,0 +1,126 @@
...
+        failure = new AtomicBoolean(false);
+        Thread.setDefaultUncaughtExceptionHandler(new Thread.UncaughtExceptionHandler() {
+            @Override
+            public void uncaughtException(Thread t, Throwable e) {
+                e.printStackTrace();
+                failure.set(true);
+            }
+        });
+        SwingUtilities.invokeAndWait(JFXPanelNPETest::createUI);
+        for (int i = 0; i < 300; i++) {
```
andy-goryachev-oracle, line 98: is 300 sufficient?

tests/system/src/test/java/test/javafx/embed/swing/JFXPanelNPETest.java
```diff
@@ -0,0 +1,126 @@
...
+            SwingUtilities.invokeLater(contentPane::repaint);
+            Platform.runLater(() -> contentPane.setScene(null));
+            Thread.sleep(100);
```
andy-goryachev-oracle, line 101: I would recommend replacing a fixed number with a random value.

(Typically, when a random is added to the test, it would make sense to print its seed to stdout, for the sake of reproducing a possible failure later; but in this case it might be unnecessary because we are dealing with multiple threads and that adds its own degree of randomness, completely eliminating reproducibility.  Or, perhaps, we still need to print the seed to maximize the probability of reproducing the failed scenario.)

andy-goryachev-oracle:
> If done correctly,

I think it makes sense, especially since there is at least one place where the whole method might need to be synchronized (JXFPanel:1090)

The key word is **if done correctly**.   One thing to look for is the possibility of introducing a deadlock and regression, so we may need to analyze the callers and perform more extensive testing.

kevinrushforth:
> The key word is **if done correctly**. One thing to look for is the possibility of introducing a deadlock and regression, so we may need to analyze the callers and perform more extensive testing.

That is exactly my concern. It would need a lot more very careful analysis before making this synchronous.

mstr2:
> That is exactly my concern. It would need a lot more very careful analysis before making this synchronous.

It seems to me that this whole class is asking for more careful analysis, considering that there seem to be several concurrency bugs in there. I don't quite like the idea of a system test "proving" that there's no race condition. That should be done through rigorous analysis instead. The first (and incorrect) proposed fix shows that the system test would have failed in demonstrating that the defects are no longer there.

tests/system/src/test/java/test/javafx/embed/swing/JFXPanelNPETest.java
hjohn, line 98: I think this is way too long for a unit test.  300 iterations, 200 ms each = 60 seconds

andy-goryachev-oracle: good point!

at the same time, we probably want to cover both extremes (short and long delays).
what would be a reasonable upper limit for a test?  10 seconds?

hjohn (commented):
It's not entirely clear to me which thread is calling which methods in this class, and it would help to clearly mark which part of the class can be called from the FX thread and which parts from the AWT thread.

I also question how some of the other fields are used.  Many are marked `volatile` which is insufficient if you don't want to observe partial changes of x/y coordinates for example.

I agree with @mstr2 that a more careful analysis couldn't hurt. There seem to be only 2 threads involved, the AWT and FX Thread.  The fields they are sharing should be grouped and marked with a comment that they're shared. Any access to these fields should then be only while holding a lock.

prsadhuk:
> > The key word is **if done correctly**. One thing to look for is the possibility of introducing a deadlock and regression, so we may need to analyze the callers and perform more extensive testing.
> 
> That is exactly my concern. It would need a lot more very careful analysis before making this synchronous.

I guess if we have to make all public methods in JFXPanel synchronous, then the same might be called for SwingNode too, ..

modules/javafx.swing/src/main/java/javafx/embed/swing/JFXPanel.java
```diff
@@ -998,28 +1039,43 @@ private class HostContainer implements HostInterface {
...
+                }
+                if (tmpStagePeer != null && JFXPanel.this.isFocusOwner()) {
+                    tmpStagePeer.setFocused(true, AbstractEvents.FOCUSEVENT_ACTIVATED);
                 }
             });
             sendMoveEventToFX();
         }
 
         @Override
         public void setEmbeddedScene(EmbeddedSceneInterface embeddedScene) {
```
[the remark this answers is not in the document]
prsadhuk: ok

tests/system/src/test/java/test/javafx/embed/swing/JFXPanelNPETest.java
prsadhuk: It used to be 3000 for JBS reproducer which I reduced to 300 for consistent issue reproduction, below which the issue might not be seen so consistently even without the fix and I guess the test is supposed to fail without the fix and pass with it..

modules/javafx.swing/src/main/java/javafx/embed/swing/JFXPanel.java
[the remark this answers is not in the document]
andy-goryachev-oracle: this code looks much better - eliminates TOC/TOU concern
https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use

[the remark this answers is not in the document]
mstr2: This code is still accessing shared state (`dnd`, `pWidth`, `pHeight`, `scaleFactorX`, `scaleFactorY`). Without further analysis of when and where these values are written, it is unclear whether this is the right move. In general, allowing interleaved code execution on multiple threads is very hard to get right. Better to not do it, if at all possible.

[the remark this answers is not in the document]
andy-goryachev-oracle: No doubt.  `dnd`, even though it's commented as "// Accessed on EDT only" (line 152) seems to be accessed from FX thread via setFxEnabled line 866.
I tend to agree with @mstr2 that this component deserves a re-think and perhaps a re-design.  I do recall seeing a number of issues with swing/fx interop in jdk8.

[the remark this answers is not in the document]
prsadhuk: I have synchronized other shared resources.

modules/javafx.swing/src/main/java/javafx/embed/swing/JFXPanel.java
```diff
@@ -405,8 +407,60 @@ private Point2D convertSwingToFxPixel(GraphicsConfiguration g, double wx, double
...
+        synchronized(LOCK) {
+            return pWidth;
+        }
+    }
+
+    private int getCompHeight() {
+        synchronized(LOCK) {
+            return pHeight;
+        }
+    }
```
andy-goryachev-oracle, line 444: perhaps we could add setCompSize(int, int) to set both pHeight and pWidth at the same time, in order to avoid multiple LOCK'ed blocks down below (L638, L1068)?

modules/javafx.swing/src/main/java/javafx/embed/swing/JFXPanel.java
```diff
@@ -405,8 +407,60 @@ private Point2D convertSwingToFxPixel(GraphicsConfiguration g, double wx, double
...
+
+    private double getScaleFactorX() {
+        synchronized(LOCK) {
+            return scaleFactorX;
+        }
+    }
```
andy-goryachev-oracle, line 450: same idea, possibly add setScaleFactor(double, double)?

modules/javafx.swing/src/main/java/javafx/embed/swing/JFXPanel.java
```diff
@@ -792,19 +877,24 @@ private void sendInputMethodEventToFX(InputMethodEvent e) {
      */
     @Override
     protected void paintComponent(Graphics g) {
-        if (scenePeer == null) {
+        EmbeddedSceneInterface lScenePeer = getScenePeer();
+        int lWidth = getCompWidth();
+        int lHeight = getCompHeight();
+        double lScaleFactorX = getScaleFactorX();
+        double lScaleFactorY = getScaleFactorY();
+        if (lScenePeer == null) {
```
andy-goryachev-oracle, line 885: minor: move check on L885 right after L880, no need to lock 4 times if lScenePeer is null

modules/javafx.swing/src/main/java/javafx/embed/swing/JFXPanel.java
```diff
@@ -997,47 +1095,59 @@ private class HostContainer implements HostInterface {
...
+            embeddedScene.setPixelScaleFactors((float) lScaleFactorX, (float) lScaleFactorY);
 
             invokeOnClientEDT(() -> {
-                dnd = new SwingDnD(JFXPanel.this, scenePeer);
-                dnd.addNotify();
-                if (scenePeer != null) {
-                    scenePeer.setDragStartListener(dnd.getDragStartListener());
-                }
+                SwingDnD lDnD = getDnD();
+                lDnD = new SwingDnD(JFXPanel.this, embeddedScene);
```
andy-goryachev-oracle, line 1147: is there a need for getting getDnd() on L1146 and immediately overwriting it on L1147?
should it be
`SwingDnD lDnD = new SwingDnD(JFXPanel.this, embeddedScene);`
?

tests/system/src/test/java/test/javafx/embed/swing/JFXPanelNPETest.java
```diff
@@ -0,0 +1,126 @@
...
+import javax.swing.SwingUtilities;
+import javax.swing.WindowConstants;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.CountDownLatch;
+
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
```
andy-goryachev-oracle, line 44: should we be using junit5?

tests/system/src/test/java/test/javafx/embed/swing/JFXPanelNPETest.java
andy-goryachev-oracle: Despite downvote, I still think randomness in the unit test is a good thing.

For more background, please take a look at the second answer in 
https://stackoverflow.com/questions/32458/random-data-in-unit-tests

andy-goryachev-oracle (commented):
overall looks good - the new LOCK'ing logic appears I have some minor comments

modules/javafx.swing/src/main/java/javafx/embed/swing/JFXPanel.java
```diff
@@ -148,6 +148,8 @@ public class JFXPanel extends JComponent {
     private transient volatile EmbeddedWindow stage;
     private transient volatile Scene scene;
 
+    private final Object LOCK = new Object();
+
     // Accessed on EDT only
```
andy-goryachev-oracle, line 153: is this comment valid anymore?

perhaps we should mention the fact that access to these fields should be LOCK'ed instead?

andy-goryachev-oracle (commented):
overall looks good - the new LOCK'ing logic seems well designed.
added some minor suggestions - I hope my comments are still there after I accidentally clicked on 'Cancel Review'...

hjohn (commented):
The changes as they are now are IMHO entirely inadequate.  Simply surrounding all accesses to shared fields with a narrowly scoped `synchronized` block is not a magic solution. You will want the fields to be coherent for a longer period than that of a single field access.  For example:

        int lWidth = getCompWidth();
        int lHeight = getCompHeight();

The new methods `getCompWidth` and `getCompHeight` are synchronized, but the overall code here is not.  This means that if the control resizes from `1000 x 1` to `1 x 1000` you may end up with a `lWidth` and `lWeight` of `1000 x 1000` or `1 x 1`.

IMHO the steps to resolve this to proper code is:
1. Find all external entry points (all non-private methods, and any private methods used by a listener callback)
2. For the entry points found above, mark which are called by FX and which by AWT
3. For the AWT methods note down all fields it accesses
4. For the FX methods note down all fields it accesses
5. Find the fields that are accessed in both the AWT and FX lists
6. Mark any methods that access those fields as `synchronized` -- this may be too course grained if these methods are large, do I/O, call back into other systems -- in that case you may need to move some code around to do all the code that needs synchronization first or last and use a `synchronized` block.

andy-goryachev-oracle:
You do have a valid point, @hjohn .
Perhaps an easier approach would be to separate fields into awt and fx ones (perhaps renaming the fields to have awt- and fx- prefix?), and use invokeLater/runLater for inter-thread communications?

This way no locking/synchronization is required, we have no shared fields, and everything happens in the right thread (at the expense of some memory allocation).

What do you guys think?

mstr2:
I agree with @hjohn and will add one more comment: even if all local effects are taken into account, the code in this class might be racing against entirely unrelated code in the JavaFX framework. These races can't be reasonably addressed with local synchronization.

Coordinating multiple threads with potentially unknown side effects requires a bigger analysis effort, and the changes here do not inspire confidence that this has happened. I see no reason to replace the existing defective implementation with another potentially defective implementation.

PR #1186, opened 2023-07-27
8313227: Correct attenuation indicator for removed lights

nlisker:
A simple fix of changing the attenuation toggle from 1 to 0 for removed lights.

---------
### Progress

### Issue
 * [JDK-8313227](https://bugs.openjdk.org/browse/JDK-8313227): Correct attenuation indicator for removed lights (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1186/head:pull/1186` \
`$ git checkout pull/1186`

`$ git checkout pull/1186` \
`$ git pull https://git.openjdk.org/jfx.git pull/1186/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1186`

`$ git pr show -t 1186`

</details>

<a href="https://git.openjdk.org/jfx/pull/1186.diff">https://git.openjdk.org/jfx/pull/1186.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1186#issuecomment-1652865067)

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/sg/prism/NGShape3D.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/sg/prism/NGShape3D.java
@@ -255,7 +255,7 @@ public abstract class NGShape3D extends NGNode {
         meshView.setLight(lightIndex,
                 0, 0, 0,    // x y z
                 0, 0, 0, 0, // r g b lightOn
-                1, 0, 0, 1, 0, // ca la qa isAttenuated maxRange
+                1, 0, 0, 0, 0, // ca la qa isAttenuated maxRange
                 0, 0, 1,    // dirX Y Z
                 0, 0, 0);   // inner outer falloff
     }


nlisker:
@kevinrushforth I think that this is simple enough to be able to get into RDP1. There is some performance improvement regained here that was lost in one of my previous enhancements because of my mistake of setting this parameter to 1.

I think that 1 reviewer is enough.

By the way, the lighting in the LightingSample test application looks a bit off to me (regardless of this patch). It's as if the point and spot lights are too dim even at close range. Do you also notice that? Might need to look at previous revisions to see when it started,

kevinrushforth:
> @kevinrushforth I think that this is simple enough to be able to get into RDP1. There is some performance improvement regained here that was lost in one of my previous enhancements because of my mistake of setting this parameter to 1.
> 
> I think that 1 reviewer is enough.

Yes, I agree that a single reviewer should be fine. This seems a reasonable candidate to backport to `jfx21` during RDP1 (no approval beyond the code review of the backport is needed).

> By the way, the lighting in the LightingSample test application looks a bit off to me (regardless of this patch). It's as if the point and spot lights are too dim even at close range. Do you also notice that? Might need to look at previous revisions to see when it started,

I'll take a quick look on both Mac and Windows and let you know what I find.

kevinrushforth:
> > By the way, the lighting in the LightingSample test application looks a bit off to me (regardless of this patch). It's as if the point and spot lights are too dim even at close range. Do you also notice that? Might need to look at previous revisions to see when it started,
> 
> I'll take a quick look on both Mac and Windows and let you know what I find.

Without attenuation, it looks fine to me on both OpenGL and D3D. When I turn on attenuation, it seems like it is attenuated more quickly (and thus dimmer) than I might expect, but I can't be sure. It might be worth checking it against earlier versions (so might need to comment out SpotLight and DirectionalLight so you can go back far enough).

It's not related to this PR, though, so I'll approve it now.

kevinrushforth (approved):

PR #1188, opened 2023-07-28
8307536: FileAlreadyExistsException from NativeLibLoader when running concurrent applications with empty cache

jperedadnr:
This PR adds a file lock to the cache directory, allowing access from multiple processes (that is, from more than one JVM) to that directory. 

This helps solving the issue where the cache is empty or doesn't exist yet, and two JavaFX applications start at the same time, and both try to copy the same native libraries to the same cache.

No tests have been added to the PR though, since this requires a complex scenario: building the SDK and publishing to local Maven repository, creating two simple JavaFX applications that use those artifacts, using a temporary folder for cache (via `javafx.cachedir`, and launching in parallel both applications.

I've tested successfully such scenario on Linux, Mac and Windows.

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
@@ -30,7 +30,10 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.InputStream;
 import java.io.IOException;
+import java.io.RandomAccessFile;
 import java.net.URI;
+import java.nio.channels.FileChannel;
+import java.nio.channels.FileLock;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.security.AccessController;
@@ -41,7 +44,6 @@ import java.security.PrivilegedAction;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Locale;
 
 public class NativeLibLoader {
 
@@ -301,7 +303,22 @@ public class NativeLibLoader {
         }
         if (write) {
             Path path = f.toPath();
-            Files.copy(is, path);
+            File lockFile = new File(cacheDir, ".lock");
+            try (RandomAccessFile randomAccessFile = new RandomAccessFile(lockFile, "rw");
+                 FileChannel fileChannel = randomAccessFile.getChannel();
+                 FileLock fileLock = fileChannel.lock()) {
+                try {
+                    if (!Files.exists(path)) {
+                        Files.copy(is, path);
+                    }
+                } finally {
+                    if (fileLock != null) {
+                        fileLock.release();
+                    }
+                }
+            } catch (IOException ex) {
+                throw new IOException("Error copying library " + path + " to cache: " + ex.getMessage(), ex);
+            }
         }
 
         String fp = f.getAbsolutePath();


andy-goryachev-oracle:
Would it still fail if an earlier JVM tries to write to the cache dir?

Just a wild idea: what if we double-check (the file hash?) after writing files or something like that?  I know, that still won't help if an older JVM writes to the same dir...

kevinrushforth:
Since there is no robust solution that work in the presence of an older JavaFX writing to the cache concurrently, I'm not sure it would be worth the complexity to further narrow the window, but @jperedadnr can comment on that.

The proposed solution seems reasonable to me. I presume you've tested it?

/reviewers 2

jperedadnr:
This PR tries to solve the issue of two (or more) concurrent attempts to do the same thing when different JavaFX applications are launched at the same time, in case the cache was empty (or gets cleaned for some reason), and I've tested successfully such scenario on Linux, Mac and Windows.

The cache directory doesn't get permanently locked, but only for a short period of time, while a file is being copied to it from resources. 

Therefore, I don't think there should be any issue with old running JVMs. Typically all the access to the cache is done during the launch of an application, where native libraries are loaded. And if you remove the cache folder _after_ the application was launched, it will still continue working. If for any reason it needs to access the cache after that, as I said, it won't be locked.

kevinrushforth:
Thanks for the additional explanation. One more thing I realized is that there won't ever be a conflict between an old (without the fix) and a new (with the fix) JavaFX release, because the version number is included in the directory name. So this fix looks good to me.

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
```diff
@@ -301,7 +303,22 @@ private static String cacheLibrary(InputStream is, String name, Class caller) th
...
         if (write) {
             Path path = f.toPath();
-            Files.copy(is, path);
+            File lockFile = new File(cacheDir, ".lock");
+            try (RandomAccessFile randomAccessFile = new RandomAccessFile(lockFile, "rw");
+                 FileChannel fileChannel = randomAccessFile.getChannel();
+                 FileLock fileLock = fileChannel.lock()) {
+                try {
+                    if (!Files.exists(path)) {
+                        Files.copy(is, path);
```
andy-goryachev-oracle, line 312: I noticed `is` is not closed (also in the original code).
Would that be a problem?

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
```diff
@@ -301,7 +303,22 @@ private static String cacheLibrary(InputStream is, String name, Class caller) th
         }
         if (write) {
             Path path = f.toPath();
-            Files.copy(is, path);
+            File lockFile = new File(cacheDir, ".lock");
+            try (RandomAccessFile randomAccessFile = new RandomAccessFile(lockFile, "rw");
+                 FileChannel fileChannel = randomAccessFile.getChannel();
+                 FileLock fileLock = fileChannel.lock()) {
+                try {
+                    if (!Files.exists(path)) {
```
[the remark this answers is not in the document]
johanvos: This is a different type of a problem, that imho should be addressed as well, but in a separate issue which also contains a test strategy. This is something that requires a hash-based solution, but it is something that needs to be done with great care, so I personally wouldn't rush this into this PR (as it might cause criticial regression if not done correctly).

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
Maran23: Probably something to evaluate, but in a separate ticket

jperedadnr: I agree, this should probably be considered as well, as part of a separate ticket. 

Note that, when the file already exists in the cache, the initial InputStream object is used by the DigestInputStream, and it is manually closed and reread again, so special care should be taken here.

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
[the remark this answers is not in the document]
jperedadnr: Agree to do it separately, but in any case, if there was already a file at the cache directory, its checksum is verified, and in case it is not the same, the file is removed from the cache, before it is copied again.
Does this cover the possible scenarios you are referring to?

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
kevinrushforth: Unless we suspect a problem, I don't see a need to do anything differently.

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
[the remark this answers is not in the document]
kevinrushforth: I agree with Johan and Jose: This should be fixed in a separate issue. @jperedadnr can you file a follow-up bug to investigate?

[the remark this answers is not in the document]
jperedadnr: Sure, I've filed: https://bugs.openjdk.org/browse/JDK-8313688

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
andy-goryachev-oracle: A possible problem is we might reach the maximum of open file handles in the underlying OS (I've experienced this before).  I think we should definitely fix it (in a separate ticket).

Curiously, the code example for Files.copy(InputStream in, Path target, CopyOption... options) does not include closing of input stream, though it has some verbiage about promptly closing the input stream in case of an error.    I think javadoc there might need to be improved.

andy-goryachev-oracle (approved):
This change fixes the problem at hand, though I'd suggest one more follow up ticket (to close the input stream after Files.copy()).

johanvos:
> This change fixes the problem at hand, though I'd suggest one more follow up ticket (to close the input stream after Files.copy()).

I agree. It's always tempting to add some (un)related changes into a PR, but it also often dangerous to do so, hence I prefer these other changes to be done in other PR's with clear and different commit messages.

johanvos (approved):
Good to go.

modules/javafx.graphics/src/main/java/com/sun/glass/utils/NativeLibLoader.java
jperedadnr: I've filed https://bugs.openjdk.org/browse/JDK-8314500 for closing the inputStream.

PR #1190, opened 2023-07-29
8311527: Region.snapInnerSpace*()

andy-goryachev-oracle:
Introduces Region.snapInnerSpaceX/Y() methods for dealing with inner space (using Math.floor), see for instance [JDK-8299753](https://bugs.openjdk.org/browse/JDK-8299753), using existing methods Region.snapPortionX/Y().

---------
### Progress
- [ ] Change requires CSR request [JDK-8314534](https://bugs.openjdk.org/browse/JDK-8314534) to be approved

### Issues
 * [JDK-8311527](https://bugs.openjdk.org/browse/JDK-8311527): Region.snapInnerSpace*() (**Enhancement** - P4)
 * [JDK-8314534](https://bugs.openjdk.org/browse/JDK-8314534): Region.snapInnerSpace*() (**CSR**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1190/head:pull/1190` \
`$ git checkout pull/1190`

`$ git checkout pull/1190` \
`$ git pull https://git.openjdk.org/jfx.git pull/1190/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1190`

`$ git pr show -t 1190`

</details>

<a href="https://git.openjdk.org/jfx/pull/1190.diff">https://git.openjdk.org/jfx/pull/1190.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1190#issuecomment-1682829970)


kevinrushforth:
I see this in the two new snap methods:

> For negative values returns 0.0.

Clamping to zero is not something any of the other snap methods do, and would be unexpected for a general purpose method such as this. If your use case really does require clamping to 0 (does it?), that should either be done by the caller, or it will need a name that implies clamping (e.g., has "clamp" or "positive" or similar in the name).

So the first question to answer is whether this really needs to clamp to zero?

Possibly worth noting is that this new method is similar to the existing private method `snapPortion`, but with a clamp to 0. I don't know if that might help inform the name (since it is non-public, it might not). 

/reviewers 2
/csr

andy-goryachev-oracle:
re: clamping.
On one hand, this method deals with the amount of space, so negative values make no sense.
On the other hand, there are indeed snapPortion*() which floors the positive values and ceils the negative.

One may argue that this is done to allow the developer to additionally encode an extra bit (sign) for a let's say direction or some other property while preserving the mathematical semantics of this method -- then we can certainly accept that and remove the clamping.  My initial thought was to fail early and clearly show that one can't have negative lengths by clamping to 0.

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/RegionTest.java
```diff
@@ -1282,6 +1282,17 @@ public void snapFunctionsShouldHandleExtremelyLargeValuesWithoutReturningNaN() {
         assertEquals(Long.MIN_VALUE, region.snapSpaceX(Double.NEGATIVE_INFINITY), 0.0);
         assertEquals(Long.MIN_VALUE, region.snapSpaceY(Double.NEGATIVE_INFINITY), 0.0);
 
+        // Inner space
+        assertEquals(Double.POSITIVE_INFINITY, region.snapInnerSpaceX(Double.MAX_VALUE), 0.0);
+        assertEquals(Double.POSITIVE_INFINITY, region.snapInnerSpaceY(Double.MAX_VALUE), 0.0);
+        assertEquals(Double.NEGATIVE_INFINITY, region.snapInnerSpaceX(-Double.MAX_VALUE), 0.0);
+        assertEquals(Double.NEGATIVE_INFINITY, region.snapInnerSpaceY(-Double.MAX_VALUE), 0.0);
```
kevinrushforth, line 1289: Why is the expected value infinity here?

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
```diff
@@ -1802,6 +1802,41 @@ public double snapPositionY(double value) {
         return snapPortionY(value, isSnapToPixel());
     }
 
+    /**
+     * If this region's snapToPixel property is true, then the value is either floored (positive values) or
+     * ceiled (negative values) with a scale. When the absolute value of the given value
+     * multiplied by the current scale is less than 10^15, then this method guarantees that:
```
kevinrushforth, line 1808: 1. This doesn't match the descriptions of the other snap methods. All of them say "to the nearest pixel" without mentioning "the given scale" (which isn't accurate anyway, since the scale isn't "given" here). I recommend doing the same here. If we want to add something about the scale, then it should be done for all of the snap methods, and in a follow-on issue. If we do this, I would add it after the first sentence, and it would need to say that it is the "render" scale, explaining that the scale is used so that it will be rounded/floored/ceiled to the nearest screen pixel after taking scale into account.

2. None of the other snap methods specify the guarantee you are making in this method, so we shouldn't add it here. This is also something we could do in a follow-on if needed.

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
```diff
@@ -1802,6 +1802,41 @@ public double snapPositionY(double value) {
...
+     *
+     * <pre>snapInnerSpaceX(snapInnerSpaceX(value)) == snapInnerSpaceX(value)</pre>
+     *
+     * The limit is about 10^15 because double values will no longer be able to represent
+     * larger integers with exact precision beyond this limit.
+     *
+     * @since 22
+     * @param value The value that needs to be snapped
+     * @return value either as passed, or floored or ceiled with scale, based on snapToPixel property
```
kevinrushforth, line 1817: same comment here: "with scale" --> "to the nearest pixel".

kevinrushforth (commented):
I have a couple comments on the javadoc and one inline comment on the test. 

Also, I recommend adding a basic functional test of the new methods (I know that the same could be said of the others), meaning a test with in-range values to make sure that it does a floor (or ceil) as expected.

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
```diff
@@ -1802,6 +1802,40 @@ public double snapPositionY(double value) {
         return snapPortionY(value, isSnapToPixel());
     }
 
+    /**
+     * If this region's snapToPixel property is true, returns a value floored
+     * to the nearest pixel in the horizontal direction, else returns the
+     * same value, for any non-negative value.  For negative values returns 0.0.
```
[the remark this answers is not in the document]
kevinrushforth: Weird. When I submitted my review, GitHub posted this old pending comment on an outdated revision. Please ignore.

hjohn:
Are you sure you'll be needing these methods for solving the table column resizing issues? For dealing with space, there are already the `snapSpace` methods, which do rounding.

The linked ticket does not really explain why these would be needed, just that they might be needed for a potential new algorithm that doesn't exist yet. The ticket also mentions that the table column resizing might need an algorithm that does multiple passes; I'm not sure if that's really needed -- it's tempting to write a column resizing algorithm by incrementally adjusting sizes during the resize, but that will lead to an accumulation of small errors eventually resulting in the mouse cursor no longer tracking the column edge being resized.  If the algorithm were rewritten to track the initial state of all columns, then apply a calculation that takes initial state + current drag location for as long as the drag is in progress, then it becomes much more predictable and testable, ie:

inputs:
   initial sizes: [10, 10, 10, 10]
   column being resized: 2
   start mouse location: 20
   current mouse location: 15

output (depending on the resize algorithm):
   sizes: [10, 5, 10, 15]

When the mouse is moved again, the calculation is again done from the initial inputs, discarding anything that was calculated in the first calculation:

inputs:
   initial sizes: [10, 10, 10, 10]
   column being resized: 2
   start mouse location: 20
   current mouse location: 25

output (depending on the resize algorithm):
   sizes: [10, 15, 10, 5]

This will avoid errors accumulating, and moving the cursor around will give the same result column sizes as long as the drag operation is ongoing.  Once the drag finishes, only then are the sizes finalized.

Having a small helper class that can be easily unit tested for the various resize options may be good.

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/RegionTest.java
andy-goryachev-oracle: Good question!

ScaledMath.floor() goes through 
```
Math.floor(d + Math.ulp(d)) / scale;
```
which results in a value greater than Double.MAX_VALUE for positive input.

Similarly, the negative input goes through
```
Math.ceil(d - Math.ulp(d)) / scale;
```
and the result is less than -Double.MAX_VALUE, so negative infinity.

It's not really important because all these are way beyond the values acceptable as coordinates.

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
andy-goryachev-oracle: the javadoc was lifted almost verbatim from snapPortionX/Y (I did not find the words "given scale" though).

Changing to be the same as snapSizeX/Y, although, technically speaking, that description is not correct either: the value in both sets of methods is ceiled/floored for positive values and floored/ceiled for negative values.

We might reword both cases saying something like "rounded to the nearest pixel with larger/smaller value" which will capture the behavior exactly.

andy-goryachev-oracle:
> Are you sure you'll be needing these methods for solving the table column resizing issues?

Se here is the rules the way I see it:
- For snapping the **min** constraint, we should be using snapSize, since the expected result should be the same or larger so as not to violate the constraint.
- For snapping the **max** constraint, we should be using snapInnerSpace, since the expected result should be the same or smaller so as not to violate the constraint.
- When both **min** and **max** constraints are set, we must pick the one than makes sense, typically the **min** one, and use the method corresponding to that.

In the context of layouts, we should not have a situation when we lay out an inner Region and it goes beyond the boundaries of the enclosing container by one pixel, right?  That's the rationale for having snapInnerSpace().

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
kevinrushforth: > the javadoc was lifted almost verbatim from snapPortionX/Y (I did not find the words "given scale" though).

That's an internal method, not API docs.

> Changing to be the same as snapSizeX/Y, although, technically speaking, that description is not correct either: the value in both sets of methods is ceiled/floored for positive values and floored/ceiled for negative values.

Interesting. But I didn't mean to copy the entire thing, just the "to the nearest pixel" part (and don't say anything about scale, since we don't anywhere else).

As for the fact that snapSize doesn't say anything about what it does for negative values, that should be fixed in a follow-on bug.

> We might reword both cases saying something like "rounded to the nearest pixel with larger/smaller value" which will capture the behavior exactly. What do you think?

I wouldn't use the word "rounded" for a floor or ceiling operation, since it will be too confusing.

kevinrushforth:
> Are you sure you'll be needing these methods for solving the table column resizing issues? For dealing with space, there are already the snapSpace methods, which do rounding.

That's a good point. These methods might or might not be needed in layout or to address table column resizing, so until we know for sure that they are needed, and are generally useful in other cases, we don't need to add them to the public API.

It seems best to move this PR to Draft until we are sure we need it.

andy-goryachev-oracle:
I see at least two use cases:

1. When laying out an unsnapped container which contains snapped children.  In order to get the available (snapped) space, we'd need to floor the unsnapped size.

2. When laying out a complex constraint layout, such as a variant of TableLayout (see for example https://github.com/andy-goryachev/FxDock/blob/master/doc/CPane.md ), where there might exist external constraints on one or more columns and/or rows, there might be a need to snap the constraint value.  How it needs to be snapped depends on the actual constraint, for example when we have a **maximum width** constraint in an otherwise unconstrained situation, we need snapInnerSpace().

hjohn:
> > Are you sure you'll be needing these methods for solving the table column resizing issues?
> 
> Se here is the rules the way I see it:
> 
> * For snapping the **min** constraint, we should be using snapSize, since the expected result should be the same or larger so as not to violate the constraint.
> * For snapping the **max** constraint, we should be using snapInnerSpace, since the expected result should be the same or smaller so as not to violate the constraint.

Are we going to be changing this, as that's not what current containers are doing?  All calculations (min/pref/max) use the same function.  That's probably a good idea as otherwise `min >= max` won't hold anymore if you use two different roundings.

I see very little problem in exceeding max in order to do snapping.  Snapping by definition has the leeway to slightly alter the constraints to align to pixels.  

> In the context of layouts, we should not have a situation when we lay out an inner Region and it goes beyond the boundaries of the enclosing container by one pixel, right? That's the rationale for having snapInnerSpace().

The off-by-one-pixel problem we've been seeing does not have its cause in using the wrong rounding AFAIK. It's because of accumulating errors in the float calculations that are not correctly handled.  The problem also occurs when containers have not reached their maximum size, so it does not have its cause in the max size specified (often max size for resizable containers is `Double.MAX_VALUE` or infinity anyway).

There's also a good chance that if you change the rounding without dealing with the accumulating errors, you'll just have the opposite problem (instead of the child being one pixel to big, it sometimes will be one pixel too small).

> I see at least two use cases:
> 
> 1. When laying out an unsnapped container which contains snapped children.  In order to get the available (snapped) space, we'd need to floor the unsnapped size.

I don't think we will want to start using different snap functions depending on the snapped status of the parent.  Snapping doesn't work when any parent is unsnapped, as the snapping occurs relative to the parent.   This was brought up before, and it's practically impossible to snap to nearest pixels when contained in an unsnapped parent (you also may not want to, because if the parent is animated this attempting to resnap a child will become visible as a weird jitter effect).

There are also problems that when a parent has its start and end point not aligned to pixels, you can't actually render the child correctly without either foregoing some snapping (at the borders, which will impact the inner borders and margins in turn), or misaligning the child's borders with the parent ones (which can show artifacts like a thin line of the parent's background showing through).

> 2. When laying out a complex constraint layout, such as a variant of TableLayout (see for example https://github.com/andy-goryachev/FxDock/blob/master/doc/CPane.md ), where there might exist external constraints on one or more columns and/or rows, there might be a need to snap the constraint value.  How it needs to be snapped depends on the actual constraint, for example when we have a **maximum width** constraint in an otherwise unconstrained situation, we need snapInnerSpace().

See above, I think min/pref/max should use the same roundings, and there is no need to strictly adhere to "min" or "max" when snapping is enabled -- snapping is allowed to ignore these values to align to the nearest pixel -- the docs can be made to mention this: `When snapping is enabled, positions, spacings and sizes may be violated up to a maximum of one pixel to ensure alignment to pixel boundaries` -- or you could be even more specific and mention that it is half a pixel up or down for spacings,  up to one pixel more for sizes, etc...

kevinrushforth:
> I see at least two use cases:
> 
> 1. When laying out an unsnapped container which contains snapped children.  In order to get the available (snapped) space, we'd need to floor the unsnapped size.

Even if this is a valid use case (which I'm not sure it is), I don't see that we should do anything different based on whether ot not the parent is snapped. So this doesn't motivate the need for this method.

> 2. When laying out a complex constraint layout, such as a variant of TableLayout (see for example https://github.com/andy-goryachev/FxDock/blob/master/doc/CPane.md ), where there might exist external constraints on one or more columns and/or rows, there might be a need to snap the constraint value.  How it needs to be snapped depends on the actual constraint, for example when we have a **maximum width** constraint in an otherwise unconstrained situation, we need snapInnerSpace().

Yes, if there is a complex case of layout that needs this, then it would be a good argument for adding this method. What I would want to see first, though, is an actual use case -- not a theoretical one -- that needs this method, and uses it in a way that shows it to be a generally useful method to add to the API.

andy-goryachev-oracle:
I don't understand the comment about "theoretical use case" - the case of CPane is practical enough for me.  The fact that javafx does not provide a good table layout Pane (and a GridPane is as clunky as its ancestor GridBagLayout) and we have not yet encountered the need for it is not, in my opinion, a good argument.

Can this functionality be written by an app developer?  Yes, so I am not that concerned.

I am even ok to withdraw this PR, but I still think there is a value in providing this API as a part of the core.

kevinrushforth:
It's theoretical, because we haven't done a thorough evaluation as to whether such a layout would actually _need_ `snapInnerSpace` to correctly do snapping. You are hypothesizing that it would, but until we have a concrete use case to evaluate and test, it's just that -- a hypothesis.

So my recommendation is not to withdraw this PR, but put it into Draft until we do the work needed to show that it is generally useful.

We should keep this potential need in mind when we review your "Column Resizing With Fractional Scale" fix in PR #1156 and when we pursue the general issues of layout being explored in PR #1111. I think we'll have a better idea of whether we will need this once we do that.

andy-goryachev-oracle:
staying aliiive

PR #1194, opened 2023-08-02
JDK-8311983: ListView sometimes throws an IndexOutOfBoundsException

Maran23:
An IOOBE was thrown when scrolling up via the trough (-> `VirtualScrollBar#adjustValue`).
This happened only when it has bigger cells than the viewport. 
If the the uppermost cell with the index 0 is only visible (although not completely scrolled to the top) and then an attempt is made to scroll up again, the `VirtualFlow` will try to scroll to the next cell, subtracting index 0 by 1, resulting in -1 -> IOOBE.

The code now guards against any under or overflow.

This is technically a regression from https://bugs.openjdk.org/browse/JDK-8173321

Note: While testing with very big cells, I found out that scrolling via the trough may not work after the first time. 
This is because the `VirtualFlow` still creates 2 cells, although only one can be visible at a time (at least when scrolling to the next cell) (and `VirtualScrollBar` does this check, which will never be true then: `firstVisibleCell == lastVisibleCell`). This is unrelated to this fix. I can create a ticket when I have more information.

---------
### Progress

### Issue
 * [JDK-8311983](https://bugs.openjdk.org/browse/JDK-8311983): ListView sometimes throws an IndexOutOfBoundsException (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1194/head:pull/1194` \
`$ git checkout pull/1194`

`$ git checkout pull/1194` \
`$ git pull https://git.openjdk.org/jfx.git pull/1194/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1194`

`$ git pr show -t 1194`

</details>

<a href="https://git.openjdk.org/jfx/pull/1194.diff">https://git.openjdk.org/jfx/pull/1194.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1194#issuecomment-1663053033)

### DIFF
--- a/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/VirtualScrollBar.java
+++ b/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/VirtualScrollBar.java
@@ -130,28 +130,29 @@ public class VirtualScrollBar extends ScrollBar {
             adjusting = true;
             double oldValue = flow.getPosition();
             double newValue = ((getMax() - getMin()) * Utils.clamp(0, pos, 1)) + getMin();
-            /**
+            /*
              * Scroll one cell further in the direction the user has clicked if only one cell is shown.
              * Otherwise, a click on the trough would have no effect when cell height > viewport height.
              */
-            IndexedCell firstVisibleCell = flow.getFirstVisibleCell();
-            IndexedCell lastVisibleCell = flow.getLastVisibleCell();
+            IndexedCell<?> firstVisibleCell = flow.getFirstVisibleCell();
+            IndexedCell<?> lastVisibleCell = flow.getLastVisibleCell();
             if (firstVisibleCell != null && firstVisibleCell == lastVisibleCell) {
                 int index = firstVisibleCell.getIndex();
                 if (newValue < oldValue) {
-                    flow.scrollTo(index - 1);
+                    index = Math.max(0, index - 1);
                 } else {
-                    flow.scrollTo(index + 1);
+                    index = Math.max(flow.getCellCount(), index + 1);
                 }
+                flow.scrollTo(index);
             } else {
                 if (newValue < oldValue) {
-                    IndexedCell cell = firstVisibleCell;
+                    IndexedCell<?> cell = firstVisibleCell;
                     if (cell == null) {
                         return;
                     }
                     flow.scrollToBottom(cell);
                 } else if (newValue > oldValue) {
-                    IndexedCell cell = lastVisibleCell;
+                    IndexedCell<?> cell = lastVisibleCell;
                     if (cell == null) {
                         return;
                     }
--- a/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/VirtualScrollBar.java
+++ b/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/VirtualScrollBar.java
@@ -141,7 +141,7 @@ public class VirtualScrollBar extends ScrollBar {
                 if (newValue < oldValue) {
                     index = Math.max(0, index - 1);
                 } else {
-                    index = Math.max(flow.getCellCount(), index + 1);
+                    index = Math.min(flow.getCellCount(), index + 1);
                 }
                 flow.scrollTo(index);
             } else {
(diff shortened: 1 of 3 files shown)


modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/VirtualFlowTest.java
```diff
@@ -34,6 +34,7 @@
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
```
kevinrushforth, line 37: We do not want to mix JUnit 4 and JUnit 5 calls in the same test class. Since converting this class to use JUnit 5 would be out of scope for a bug fix such as this, please rewrite it using JUnit 4 API calls only (which probably means a try/catch with a "fail" if you don't get the expected exception.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/VirtualScrollBar.java
```diff
@@ -130,28 +130,29 @@ public final BooleanProperty virtualProperty() {
...
+            IndexedCell<?> firstVisibleCell = flow.getFirstVisibleCell();
+            IndexedCell<?> lastVisibleCell = flow.getLastVisibleCell();
             if (firstVisibleCell != null && firstVisibleCell == lastVisibleCell) {
                 int index = firstVisibleCell.getIndex();
                 if (newValue < oldValue) {
-                    flow.scrollTo(index - 1);
+                    index = Math.max(0, index - 1);
                 } else {
-                    flow.scrollTo(index + 1);
+                    index = Math.max(flow.getCellCount(), index + 1);
```
kevinrushforth, line 144: Shouldn't this be Math.min?

kevinrushforth (commented):
I'll let others who are more familiar with VirtualFlow formally review it, but I left a comment on the fix and test.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/VirtualScrollBar.java
Maran23: Yes indeed, leftover from testing.

kevinrushforth:
@andy-goryachev-oracle and @johanvos can you review ?

andy-goryachev-oracle (approved):
The change looks good, tested with the test app in the ticket and the Monkey Tester.

I wonder if we ought to consider some kind of `dirty` flag in the virtual layout to ensure that the layout is always performed when needed.  I understand this might require a larger change and it's definitely out of the scope for this PR.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/VirtualFlowTest.java
```diff
@@ -1767,6 +1767,77 @@ public void testAddCellWithBigCurrentOne() {
...
+        // Scroll up.
+        flow.shim_getVbar().adjustValue(0.05263157894736842);
+
+        // This should not throw any exception. It used to throw an IndexOutOfBoundsException.
+        flow.shim_getVbar().adjustValue(0.05263157894736842);
+    }
+
+    @Test
+    public void testScrollBarValueAdjustmentShouldScrollOneDown() {
+        flow = new VirtualFlowShim<>();
```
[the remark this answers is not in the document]
Maran23: No.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/VirtualFlowTest.java
```diff
@@ -1767,6 +1767,77 @@ public void testAddCellWithBigCurrentOne() {
...
+        flow.setFixedCellSize(512);
+        flow.setCellFactory(fw -> new CellStub(flow));
+        flow.setCellCount(2);
+        flow.resize(250, 300);
+
+        pulse();
+
+        // Scroll down.
+        flow.shim_getVbar().adjustValue(0.9605263157894737);
+        // Scroll up.
```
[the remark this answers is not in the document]
Maran23: Not sure, but it is fishy that it is called twice sometimes. So it is worth checking out if this can be optimized.

mstr2 (approved):
The solution is sound, as it constrains the index to a valid range. The tests work as expected.

PR #1199, opened 2023-08-05
8313648: JavaFX application continues to show a black screen after graphic card driver crash

tsx84:
Hi,

I did open the bug report. Some notes to this PR:

My colleagues and I are able to reproduce this bug regularly, even though it takes sometimes up to 3 or 4 weeks until the D3DERR_DEVICEHUNG error shows up. We are currently evaluating two versions of fixes, but until now we do not have any results. I will post them as soon as I got them.

Version 1 (this version): Based on the observation, that the TestCooperativeLevel/CheckDeviceState method returns D3D_OK again after about 20 - 60 seconds, the reinitialize is called after the first time the state is returning D3D_OK. The 'isHung' flag stores the information until then.

Version 2: calls reinitialize directly after D3DERR_DEVICEHUNG has been returned. Basically
if (hr == D3DERR_DEVICEREMOVED || hr == D3DERR_DEVICEHUNG  ) { .. }

I did not modify the validatePresent method, as for our workaround (see ticket) it was not necessary. At least the native call swapchain->present dows not return that error code (https://learn.microsoft.com/en-us/windows/win32/api/d3d9/nf-d3d9-idirect3dswapchain9-present). I did not look decisively into all the native calls behind D3DRTTexture#readPixels.

As I said I will post the results (prism.verbose output) for the 2 versions later as a base for discussions.

---------
### Progress

### Issue
 * [JDK-8313648](https://bugs.openjdk.org/browse/JDK-8313648): JavaFX application continues to show a black screen after graphic card driver crash (**Bug** - P3)

### Reviewers
 * [Marius Hanl](https://openjdk.org/census#mhanl) (@Maran23 - Committer) 🔄 Re-review required (review applies to [61a5ec15](https://git.openjdk.org/jfx/pull/1199/files/61a5ec15dbffd644c530139cce92ca5f96d9e2f5))

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1199/head:pull/1199` \
`$ git checkout pull/1199`

`$ git checkout pull/1199` \
`$ git pull https://git.openjdk.org/jfx.git pull/1199/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1199`

`$ git pr show -t 1199`

</details>

<a href="https://git.openjdk.org/jfx/pull/1199.diff">https://git.openjdk.org/jfx/pull/1199.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1199#issuecomment-1742113052)


kevinrushforth:
This version seems like a straight-forward fix, so if it resolves the problem, it may well be the best option. It uses the existing well-tested logic to reinitialize once the graphics device reports that it is OK again.

We can discuss your findings regarding your "version 2" when you are ready, but we would need a pretty compelling reason to want to make any intrusive changes.

/reviewers 2

Maran23:
As I also worked on this in https://github.com/openjdk/jfx/pull/1200 and have a much better understanding of the code there (and the communication with Direct3d9) I agree this looks like the right thing to do in this situation.

tsx84:
Thank you @Maran23 for taking a look into this! Sadly it did not work out as expected.

Our observations with current proposal ("Version 1"): When the error occurred, the state did not go into D3D_OK again, but it stayed at D3DERR_DEVICEHUNG. The D3DERR_DEVICEHUNG error message was printed over and over (for > 1 day).

For "Version 2" (directly calling reiniztialize after the error): One day we saw the Prism initialization text 3 times in a row and the app was running fine. As it happened during a holiday week, we are unsure if there were 3 distinct errors or if there was only one error that took 3 attempts/iterations. (We turned application logging off to not pollute the console log.)

Especially that the current proposal did not work out is very unfortunate, becuase either we did not analyze correctly what happend or the error behavior is not that predictable. Either way, we are now re-running our workaround version withe the 5-minute loop (see bug ticket) and include timestamps and "Version 2" also with additional timestamp information.
We are going to rerun both version mutliple times, to have somewhat reliable information. I'll update findings here.

Maran23:
> Our observations with current proposal ("Version 1"): When the error occurred, the state did not go into D3D_OK again, but it stayed at D3DERR_DEVICEHUNG. The D3DERR_DEVICEHUNG error message was printed over and over (for > 1 day).

That is weird. I would also expect that Direct3d will eventually report `D3D_OK` when the device recovered..

tsx84:
Another update. This time for the version with a 15-minute timer (and periodically checking the D3D error status every second). The reinitialization happened after the 15 minutes went by. So it seems that the Version 2 (immediately calling reinitialze) is the most promising solution. But we will run those tests again, at least one time each.
[15_min_timer_with_timestamp.txt](https://github.com/openjdk/jfx/files/12665424/15_min_timer_with_timestamp.txt)

tsx84:
Another time Version 2 hung and recovered immediately. At the moment this version seems promising to us.

tsx84:
Last week the error happend again, and calling reinitialize immediately (-> aka Version 2) was again a reliable way to handle it (screenshot is attached). Everything runs normal and we did not see multiple reinitializations for the same error. The 2 errors are about an hour appart, which indicates two distict failures. The former fix did sadly not work, so I updated this PR. We see it as a solution and would be glad to integrate it in the code base. @kevinrushforth if you have any further questions, please let me know.
screenshot

mstr2 (approved):
Microsoft's [documentation](https://learn.microsoft.com/en-us/windows/win32/direct3d9/dx9lh) states the following for Direct3D 9Ex devices:

> Devices are now only lost under two circumstances; when the hardware is reset because it is hanging, and when the device driver is stopped. When hardware hangs, the device can be reset by calling [ResetEx](https://learn.microsoft.com/en-us/windows/desktop/api/d3d9/nf-d3d9-idirect3ddevice9ex-resetex). If hardware hangs, texture memory is lost.

The current D3D pipeline doesn't call `ResetEx` when the device hangs, it releases and recreates the device instead. However, I think it is okay to do this. [Here's](https://learn.microsoft.com/en-us/windows/win32/direct3d9/d3derr) what the documentation says about `D3DERR_DEVICEHUNG`:

> The device that returned this code caused the hardware adapter to be reset by the OS. Most applications should destroy the device and quit. Applications that must continue should destroy all video memory objects (surfaces, textures, state blocks etc) and call Reset() to put the device in a default state. If the application then continues rendering in the same way, the device will return to this state.

On a side note, I think we should remove the non-9Ex code paths from the D3D pipeline. We're not supporting Windows XP, are we?

kevinrushforth:
The current version (what you called "version 2" in the Description) looks good to me. Despite my initial concern when you first mentioned it, it's actually _less_ intrusive than "version 1" and seems like the right approach: handle `D3DERR_DEVICEHUNG` in the same manner as `D3DERR_DEVICEREMOVED`. I'll do some minimal testing, and approve it if I don't spot any problems.

kevinrushforth:
> On a side note, I think we should remove the non-9Ex code paths from the D3D pipeline. We're not supporting Windows XP, are we?

This might be a good follow-up issue to file. We no longer support nor run on Windows XP (or Windows Vista).

mintykat:
We have a mission critical application for Navy air traffic control that is losing javaFX display continuity after the graphics driver crashes on an HP Pavlion G6, which happens randomly every few hours. Of course, while the error says the system recovers, the javaFX doesn't, so we have to reboot the workstation. I see that a similar issue was fixed for remote desktop. I hope we can get an Oracle bug fix ASAP.

modules/javafx.graphics/src/main/native-prism-d3d/D3DContext.cc
```diff
@@ -758,6 +758,9 @@ D3DContext::TestCooperativeLevel()
     case D3DERR_DEVICEREMOVED:
         TraceLn1(NWT_TRACE_INFO, "  device %d has been removed", adapterOrdinal);
         break;
+    case D3DERR_DEVICEHUNG:
```
Maran23, line 761: you probably need to adjust the copyright here as well

Maran23 (approved):
Looks good to me as well.

mintykat:
We are currently using jre 11.0.19 with javaFX as a browser. Since it appears that the checks have passed, we will want to obtain the fix asap () although the Oracle bug report doesn't indicate a fix is ready). Please indicate when we can get the fix, and whether it be integrated into another version of the jre or whether it will be a separate javaFX download. Thx.

mintykat:
I have put this in D3DContext.java, a slight variation of the suggested fix. Just wondering if I should just reinitialize directly and not wait loop: in testLostState in D3DContext.java
       if (hr == D3DERR_DEVICEHUNG || hr == D3DERR_DEVICEREMOVED) {
            setLost();

            long retryMillis = TimeUnit.MINUTES.toMillis(5);
            long sleepMillis = TimeUnit.SECONDS.toMillis(1);

            for (int i = 0; i < retryMillis; i += sleepMillis) {
                int cooperativeLevel = D3DResourceFactory.nTestCooperativeLevel(pContext);
                System.err.println("Checking Cooperative Level: " + cooperativeLevel);

                if (cooperativeLevel == D3D_OK) {
                    break;
                } else {
                    try {
                        Thread.sleep(sleepMillis);
                    } catch (InterruptedException e) {
                        e.printStackTrace();
                    }
                }
            }

            // Reinitialize after 5 mins anyway, even if result is not OK.

            // Reinitialize the D3DPipeline. This will dispose and recreate
            // the resource factory and context for each adapt
            D3DPipeline.getInstance().reinitialize();
            LOGGER.warn("Reinit after graphics hang.");
        }

mstr2 (approved):

mintykat:
Thx!These hangs tend to only happen with Pavilion G6 chips, not G4 or G9. We've tried switching to different graphics drivers to no avail. I put some LOGGER.warn messages in D3DContext.java to help us pinpoint. I'm inexperienced with Intellij, so I'm not sure where to put this javafx.graphics code, so I stuck it on the same level as the Browser.java which used javaFX as a "browser" - i.e. for graphics. Its code calls some source from a zip file.Since the reviews are complete, a release fix, say, a new jre version, would be the best option. 

    On Monday, November 27, 2023 at 02:57:42 PM EST, Thorsten Fischer ***@***.***> wrote:  
 
 


Hello @mintykat , the loop is not necessary, in fact it is not recommended as it makes the whole application (window) unresponsive due to the Thread.sleep. It was just an approach to generate some debug output in order to see how the system and D3D is responding after the failure happens. Best is to call reinitialize directly after the check for the D3DERR_DEVICEHUNG error code.

I'm wondering a little bit, if JavaFX may be somehow responsible for the crash or if its just old drivers? This fix is going to keep the application running, but those crashes will still happen. You said, that you can reproduce this error every couple of hours? If you have the capacity, maybe you can track down the root cause? But I guess it is not a trivial task, I just found here that some debug.dlls are needed. We had this error 'only' every few days (sometimes weeks), and I wasn't quite motivated for such a long winding deep-dive bug hunt :)

mintykat:
Thx, Thorsten. No, the crashes only happen about once every 8 hours and only on a few of the G6's. I finally found the javaFX code in a zip file under a production folder, and we made the change there (after unzippimg it). We put in a few logger debug statements and are only looping ojne minute - we'll remove those after this try. 
Looks like you guys are ready to release this fix.

kevinrushforth (approved):
Looks good. I don't have a way to provoke a real device hung state, but I simulated it and it works as expected (the same as device reset).

kevinrushforth:
@arapte Do you have any concerns about this PR? Since there are already two reviewers, you can review it or not as you prefer.

arapte:
As pointed out by @tsx84 [here](https://github.com/openjdk/jfx/pull/1199#issuecomment-1828503142): This fix is to recover from a device hung situation but does not fix the root cause as why did the device hung.
But the fix itself seems safe, I have no concerns about it. thanks @kevinrushforth for waiting on me.

mintykat:
Thorsten,Thanks for helping to resolve this! I just would like to know where I can download the relevant jar file with fixed javafx components. 
    On Wednesday, November 29, 2023 at 03:10:25 AM EST, Thorsten Fischer ***@***.***> wrote:  
 
 


/integrate

mintykat:
Thanks for helping fix this! Would like to know where I can download the updated jfx graphics jar.-Dave
    On Wednesday, November 29, 2023 at 06:59:37 AM EST, Kevin Rushforth ***@***.***> wrote:  
 
 


/sponsor

tsx84:
Hi @mintykat. thanks, I'm glad I also could help others :) Usually the fix will be available with the next main release. I'm not 100% sure, but it should be JavaFX 22. I don't know the exact relase date, but it seems it will be well into next year (I think around March?).. All in all your options depend on which JDK/Java version you are bound to. Since you 'still' use JavaFX 11, it seems you need a backport of this fix. If you are up to the task, you can build your own version with the adapted source code (Steps to build JavaFX are described here: https://wiki.openjdk.org/display/OpenJFX/Building+OpenJFX). Or you might ask the project leads  (@kevinrushforth or @johanvos) to backport this fix. I don't know how requests for backports are handled, but it might be, that you need some sort of support plan from Gluon. But I'm also not sure about this.

kevinrushforth:
That's a pretty good summary of when it will be available in a production version: JavaFX 22, which will be shipped in March along with JDK 22. Early access builds will be available sooner.

There is a short window of time (until next Monday, 4-Dec-2023) to get this into jfx21u for JavaFX 21.0.2 if someone wants to backport it. This is a safe enough fix, that I will approve it if you do.

Getting this into jfx17u-dev and/or jfx11u-dev is possible, but I don't know how likely it is nor what would be involved in getting a build of JavaFX 17.0.x / 11.0.x with the fix. Check with @johanvos of Gluon (probably offline).

mintykat:
Thx, I guess a 21u jar would be ok. I copied Adam Owsley for a possible comment. Comments from Johan Vos would also be appreciated.
    On Wednesday, November 29, 2023 at 12:45:43 PM EST, Kevin Rushforth ***@***.***> wrote:  
 
 


That's a pretty good summary of when it will be available in a production version: JavaFX 22, which will be shipped in March along with JDK 22. Early access builds will be available sooner.

There is a short window of time (until next Monday, 4-Dec-2023) to get this into jfx21u for JavaFX 21.0.2 if someone wants to backport it. This is a safe enough fix, that I will approve it if you do.

Getting this into jfx17u-dev and/or jfx11u-dev is possible, but I don't know how likely it is nor what would be involved in getting a build of JavaFX 17.0.x / 11.0.x with the fix. Check with @johanvos of Gluon (probably offline).

johanvos:
I added the fix-request label to backport to 21.0.2
I agree it would be good to backport this to jfx17u.
The 5-year support for jfx11u ended in september, and we now focus on LTS for 17/21. 
Contact me offline in case you need something specific (e.g. for 11).

mintykat:
Thx. We will have to upgrade Spring Boot for the 21 jar, but maybe that's okay. 
    On Wednesday, November 29, 2023 at 01:44:25 PM EST, Johan Vos ***@***.***> wrote:  
 
 


I added the fix-request label to backport to 21.0.2
I agree it would be good to backport this to jfx17u.
The 5-year support for jfx11u ended in september, and we now focus on LTS for 17/21.
Contact me offline in case you need something specific (e.g. for 11).

PR #1200, opened 2023-08-05
JDK-8283401: ArrayIndexOutOfBoundsException when disconnecting screen(s)

Maran23:
When the `D3DPipeline` is reinitialized, the adapter ordinal of all the `Screen`s are outdated.
As a consequence, when a `D3DResourceFactory` is created for a `Screen` (adapter ordinal), the code may fail with an `ArrayIndexOutOfBoundsException` as the ordinal is higher than what we would expect it to be.

Example:
We have 3 screens. Adapter ordinal 0, 1, 2.
If we now disconnect 2 screens, we only have one screen left. 
The `D3DPipeline` is reinitialized as a [D3DERR_DEVICEREMOVED](https://learn.microsoft.com/en-us/windows/win32/direct3d9/d3derr) is reported from Direct3d9.
Now the remaining screen may have the adapter ordinal 1 (from before). But since we only have 1 screen left, it should have the adapter ordinal 0. 

Direct3d9 will also return 0 as adapter ordinal if you call it. And that is what we do to update the `Screen` and fix this problem.

See also the ticket for information how to reproduce this problem (Windows only).

---------
### Progress

### Issue
 * [JDK-8283401](https://bugs.openjdk.org/browse/JDK-8283401): ArrayIndexOutOfBoundsException when disconnecting screen(s) (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1200/head:pull/1200` \
`$ git checkout pull/1200`

`$ git checkout pull/1200` \
`$ git pull https://git.openjdk.org/jfx.git pull/1200/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1200`

`$ git pr show -t 1200`

</details>

<a href="https://git.openjdk.org/jfx/pull/1200.diff">https://git.openjdk.org/jfx/pull/1200.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1200#issuecomment-1666501229)

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/prism/d3d/D3DPipeline.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/prism/d3d/D3DPipeline.java
@@ -199,6 +199,12 @@ public final class D3DPipeline extends GraphicsPipeline {
 
         d3dInitialized = true;
         factories = new D3DResourceFactory[nGetAdapterCount()];
+
+        // Reassign the adapter ordinal of all the screens because the configuration has been changed
+        // and the old adapter ordinal may be outdated.
+        for (Screen screen : Screen.getScreens()) {
+            screen.setAdapterOrdinal(getAdapterOrdinal(screen));
+        }
     }
 
     @Override


kevinrushforth:
Thanks for tracking this down. I'll review and do some testing of this next week.

/reviewers 2

mstr2:
I tested the patch with a docking station and two external displays like this:
1. Put Windows into sleep mode
2. Disconnect the docking station / external monitor
3. Resume Windows

Without the patch, the exception occurs. With the patch applied, everything works as expected.

I notice that the logic to assign adapter ordinals is now duplicated in two places in the codebase. Have you thought about moving the implementation into the `Screen` class? Here's how that could work:

Remove `Screen.setAdapterOrdinal(int)` and add the following method to the `Screen` class:
```java
    public static void updateAdapterOrdinals() {
        for (Screen screen : getScreens()) {
            screen.adapter = GraphicsPipeline.getPipeline().getAdapterOrdinal(screen);
        }
    }
```

Remove `QuantumToolkit.assignScreenAdapters()` and call `Screen.updateAdapterOrdinals()` from the two call sites instead. You can then just call `Screen.updateAdapterOrdinals()` from `D3DPipeline` as well.

For good measure, you could initialize `Screen.adapter` to `-1` and change `Screen.getAdapterOrdinal()` as follows:
```java
    public int getAdapterOrdinal() {
+       if (adapter < 0) {
+           throw new IllegalStateException("Adapter ordinal is not initialized.");
+       }
+
        return adapter;
    }
```

Maran23:
> I notice that the logic to assign adapter ordinals is now duplicated in two places in the codebase. Have you thought about moving the implementation into the `Screen` class? 

I actually considered this but I really want to keep the diff in this PR as small as possible, as this is a very sensitive area. If something is broken here, it will affect a lot of Windows users. 
IMO this is a good idea for a follow up PR, as soon as this one is merged.

mstr2:
> I actually considered this but I really want to keep the diff in this PR as small as possible, as this is a very sensitive area. If something is broken here, it will affect a lot of Windows users. IMO this is a good idea for a follow up PR, as soon as this one is merged.

Wouldn't this just impose the compatibility risk again onto the next PR? The diff is very small in both cases.

kevinrushforth:
> > I actually considered this but I really want to keep the diff in this PR as small as possible, as this is a very sensitive area. If something is broken here, it will affect a lot of Windows users. IMO this is a good idea for a follow up PR, as soon as this one is merged.
> 
> Wouldn't this just impose the compatibility risk again onto the next PR? The diff is very small in both cases.

I prefer the approach taken in this PR to minimize the diffs. One benefit is that I could imagine, for example, wanting to backport this small fix once it is integrated and has baked for a while. I would have no desire to backport what is otherwise just a refactoring to eliminate duplicate code.

kevinrushforth (approved):
The code change looks good.

I am not able to provoke the bug, but I did test it enough to validate the logic, by simulating device lost and verifying that the adapter ordinals are all reinitialized correctly.

mstr2 (approved):

Maran23:
Thanks for the reviews!
@mstr2 I will file a follow-up in the next days to avoid the code duplications.

PR #1203, opened 2023-08-07
8313709: Wrong layout of a FlowPane in a BorderPane in a ScrollPane, if padding is too big

jperedadnr:
So far, BorderPane does the calculation for the children min/pref width/height taken into account only the margin applied to them, if any, but not the total padding that could be applied as well to the BorderPane itself.

However, this padding needs to be taken into account as well, and this PR modifies BorderPane to subtract its insets from its size while doing the children min/pref width/height calculations.

A parameterized test has been included. 

It is a simplified version of the test case attached to https://bugs.openjdk.org/browse/JDK-8313709, but still shows how without this patch, two of the cases (padding with or without margin) fail, while pass with it.

---------
### Progress

### Issue
 * [JDK-8313709](https://bugs.openjdk.org/browse/JDK-8313709): Wrong layout of a FlowPane in a BorderPane in a ScrollPane, if padding is too big (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1203/head:pull/1203` \
`$ git checkout pull/1203`

`$ git checkout pull/1203` \
`$ git pull https://git.openjdk.org/jfx.git pull/1203/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1203`

`$ git pr show -t 1203`

</details>

<a href="https://git.openjdk.org/jfx/pull/1203.diff">https://git.openjdk.org/jfx/pull/1203.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1203#issuecomment-1667643744)

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/layout/BorderPane.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/layout/BorderPane.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -376,6 +376,8 @@ public class BorderPane extends Pane {
     }
 
     @Override protected double computeMinWidth(double height) {
+        final Insets insets = getInsets();
+
         double topMinWidth = getAreaWidth(getTop(), -1, true);
         double bottomMinWidth = getAreaWidth(getBottom(), -1, true);
 
@@ -389,7 +391,8 @@ public class BorderPane extends Pane {
             double topPrefHeight = getAreaHeight(getTop(), -1, false);
             double bottomPrefHeight = getAreaHeight(getBottom(), -1, false);
 
-            double middleAreaHeight = Math.max(0, height - topPrefHeight - bottomPrefHeight);
+            double middleAreaHeight = Math.max(0,
+                    height - insets.getTop() - insets.getBottom() - topPrefHeight - bottomPrefHeight);
 
             leftPrefWidth = getAreaWidth(getLeft(), middleAreaHeight, false);
             rightPrefWidth = getAreaWidth(getRight(), middleAreaHeight, false);
@@ -400,7 +403,6 @@ public class BorderPane extends Pane {
             centerMinWidth = getAreaWidth(getCenter(), -1, true);
         }
 
-        final Insets insets = getInsets();
         return insets.getLeft() +
                 Math.max(leftPrefWidth + centerMinWidth + rightPrefWidth, Math.max(topMinWidth,bottomMinWidth)) +
                 insets.getRight();
@@ -408,6 +410,9 @@ public class BorderPane extends Pane {
 
     @Override protected double computeMinHeight(double width) {
         final Insets insets = getInsets();
+        if (width != -1) {
+            width -= (insets.getLeft() + insets.getRight());
+        }
 
         // Bottom and top are always at their pref height
         double topPrefHeight = getAreaHeight(getTop(), width, false);
@@ -432,6 +437,8 @@ public class BorderPane extends Pane {
     }
 
     @Override protected double computePrefWidth(double height) {
+        final Insets insets = getInsets();
+
         double topPrefWidth = getAreaWidth(getTop(), -1, false);
         double bottomPrefWidth = getAreaWidth(getBottom(), -1, false);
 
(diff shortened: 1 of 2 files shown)


jperedadnr:
Okay, makes sense. 
Should I port the test back to JUnit 4, then? Or convert the existing test class to JUnit 5?

kevinrushforth:
> Should I port the test back to JUnit 4, then? Or convert the existing test class to JUnit 5?

If the changes aren't too intrusive, it would be OK to convert to JUnit 5. Otherwise, port the test back to JUnit 4.

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/BorderPaneTest.java
```diff
@@ -29,10 +29,15 @@
...
 import javafx.scene.ParentShim;
+import javafx.scene.Scene;
 import javafx.scene.layout.BorderPane;
+import javafx.scene.layout.FlowPane;
+import javafx.stage.Stage;
 import org.junit.Before;
 
 import org.junit.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
```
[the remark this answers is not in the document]
jperedadnr: Fixed, migrated the existing tests to JUnit 5

jperedadnr:
Okay, done

modules/javafx.graphics/src/main/java/javafx/scene/layout/BorderPane.java
```diff
@@ -389,7 +391,8 @@ public final ObjectProperty<Node> rightProperty() {
             double topPrefHeight = getAreaHeight(getTop(), -1, false);
             double bottomPrefHeight = getAreaHeight(getBottom(), -1, false);
 
-            double middleAreaHeight = Math.max(0, height - topPrefHeight - bottomPrefHeight);
+            double middleAreaHeight = Math.max(0,
+                    height - insets.getTop() - insets.getBottom() - topPrefHeight - bottomPrefHeight);
```
andy-goryachev-oracle, line 395: should these be snapped?  snappedBottomInset() etc.?

modules/javafx.graphics/src/main/java/javafx/scene/layout/BorderPane.java
```diff
@@ -400,14 +403,16 @@ public final ObjectProperty<Node> rightProperty() {
...
         return insets.getLeft() +
                 Math.max(leftPrefWidth + centerMinWidth + rightPrefWidth, Math.max(topMinWidth,bottomMinWidth)) +
                 insets.getRight();
     }
 
     @Override protected double computeMinHeight(double width) {
         final Insets insets = getInsets();
+        if (width != -1) {
+            width -= (insets.getLeft() + insets.getRight());
+        }
```
andy-goryachev-oracle, line 415: ditto here

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/BorderPaneTest.java
```diff
@@ -915,6 +923,56 @@ public class BorderPaneTest {
...
+            "209.0, 430.0, true, false",  // flowPane 4x1, margin, no padding
+            "211.0, 210.0, true, false",  // flowPane 2x2, margin, no padding
+            "209.0, 430.0, false, true",  // flowPane 4x1, no margin, padding
+            "211.0, 210.0, false, true",  // flowPane 2x2, no margin, padding
+            "209.0, 430.0, true, true",   // flowPane 4x1, margin, padding
+            "241.0, 210.0, true, true",   // flowPane 2x2, margin, padding
+    })
+    @ParameterizedTest
+    public void testFlowPaneCenterChildWithPaddingAndMargin(double width, double minHeight, boolean useMargin, boolean usePadding) {
+        FlowPane center = new FlowPane(10, 10);
```
andy-goryachev-oracle, line 938: I wonder if we should also test the layout-related code using fractional scale + snapping.
what do you think?

andy-goryachev-oracle:
FYI: the test code shows 2x2 arrangement on windows at 100%, but 1x4 arrangement at 125%.  Is this expected?
In both cases scrolling and resizing seems ok.

modules/javafx.graphics/src/main/java/javafx/scene/layout/BorderPane.java
jperedadnr: You might be right. 

In this case, `topPrefHeight` comes from `getAreaHeight()`, that calls `computeChildPrefAreaHeight()` that ultimately uses `snapSpaceY()`. 

However, this would also mean that the returned value should use snapped insets as well?
```
return insets.getLeft() +
                Math.max(leftPrefWidth + centerMinWidth + rightPrefWidth, Math.max(topMinWidth,bottomMinWidth)) +
                insets.getRight();
```
and also for the layout call?

andy-goryachev-oracle: no, I did not propose to change it everywhere.
In this PR, in the code we change, we probably should [start to] use snapped coordinates.  wouldn't you agree?

kevinrushforth: Hard to say. Unless it's somehow related to the changes being made, I can see a good argument for addressing the snapping in a follow-up issue, so as not to introduce what is essentially an unrelated fix.

modules/javafx.graphics/src/main/java/javafx/scene/layout/BorderPane.java
```diff
@@ -400,14 +403,16 @@ public final ObjectProperty<Node> rightProperty() {
...
-        final Insets insets = getInsets();
         return insets.getLeft() +
                 Math.max(leftPrefWidth + centerMinWidth + rightPrefWidth, Math.max(topMinWidth,bottomMinWidth)) +
                 insets.getRight();
     }
 
     @Override protected double computeMinHeight(double width) {
         final Insets insets = getInsets();
+        if (width != -1) {
+            width -= (insets.getLeft() + insets.getRight());
```
mstr2, line 414: Let's say we call `computeMinHeight(10)`, but the left and right insets are 20. This means that `width` is now -10, which probably means "ignore the value" (the spec isn't entirely clear about that, but -10 is not a valid width in any case).

I think the following code might be better:
```java
if (width >= 0) {
    width = Math.max(0, width - insets.getLeft() - insets.getRight());
}
```

hjohn: The `width` value passed to `computeMinHeight` (if not -1) should be the result of a call to `computeMinWidth(-1)` (depending on the result of `getContentBias`; this is specified somewhere); if that function always includes the insets, then subtracting them here should not result in a negative value.

Of course, one can never be too careful.

I don't like the changing of the meaning of the `width` parameter here though using parameter assignment.

PR #1209, opened 2023-08-14
8087700: [KeyCombination, Mac] KeyCharacterCombinations behave erratically

beldenfox:
A KeyCharacterCombination should match a key if the target character is printed on that key. For example, the user should be able to invoke the `Shortcut+'+' ` combination by holding down the Shortcut key and pressing a key that has '+' printed on it. This should work even if '+' is a shifted symbol but the user doesn't hold down the Shift key. 

The Mac implements KeyCharacterCombinations by monitoring keystrokes to discover the relationship between keys and characters. Currently the system only records the character the user typed and no other characters on the same key. This means a shortcut targeting a shifted character may not work until the user types that character using Shift so the system learns the relationship.

This PR keeps the same mechanism in place but always records the shifted and unshifted character for each keystroke.

For the Mac the KeyboardTest app was modified to remove tests for characters accessed using Option. We don't look for these characters because under the hood just about every key has some symbol assigned to the Option modifier that the user probably isn't even aware of. For these character we fall back to the existing logic; once the user types the character it will start working as a shortcut.

---------
### Progress

### Issue
 * [JDK-8087700](https://bugs.openjdk.org/browse/JDK-8087700): [KeyCombination, Mac] KeyCharacterCombinations behave erratically (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1209/head:pull/1209` \
`$ git checkout pull/1209`

`$ git checkout pull/1209` \
`$ git pull https://git.openjdk.org/jfx.git pull/1209/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1209`

`$ git pr show -t 1209`

</details>

<a href="https://git.openjdk.org/jfx/pull/1209.diff">https://git.openjdk.org/jfx/pull/1209.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1209#issuecomment-1677694793)

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
@@ -38,6 +38,7 @@
 
 #import "ProcessInfo.h"
 #import <Security/SecRequirement.h>
+#import <Carbon/Carbon.h>
 
 //#define VERBOSE
 #ifndef VERBOSE
@@ -741,13 +742,38 @@ jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved)
             javaIDs.Application.leaveNestedEventLoop, (jobject)NULL);
 }
 
+static void inputDidChangeCallback(CFNotificationCenterRef center, void *observer, CFNotificationName name, const void *object, CFDictionaryRef userInfo)
+{
+    if (keyCodeForCharMap != nil) {
+        [keyCodeForCharMap removeAllObjects];
+    }
+}
+
 + (void)registerKeyEvent:(NSEvent*)event
 {
     if (!keyCodeForCharMap) {
         keyCodeForCharMap = [[NSMutableDictionary alloc] initWithCapacity:100];
         // Note: it's never released, just like, say, the jApplication reference...
+        // To avoid stale entries when the user switches layout
+        CFNotificationCenterRef center = CFNotificationCenterGetDistributedCenter();
+        CFNotificationCenterAddObserver(center, NULL, inputDidChangeCallback,
+                                        kTISNotifySelectedKeyboardInputSourceChanged,
+                                        NULL, CFNotificationSuspensionBehaviorCoalesce);
     }
-    [keyCodeForCharMap setObject:[NSNumber numberWithUnsignedShort:[event keyCode]] forKey:[event characters]];
+
+    // Add the character the user typed to the map.
+    NSNumber* mapObject = [NSNumber numberWithUnsignedShort:[event keyCode]];
+    [keyCodeForCharMap setObject:mapObject forKey:[event characters]];
+    // getKeyCodeForChar should not just match against a character the user types
+    // directly but any other character printed on the same key.
+    [keyCodeForCharMap setObject:mapObject forKey:[event charactersByApplyingModifiers: 0]];
+    [keyCodeForCharMap setObject:mapObject forKey:[event charactersByApplyingModifiers: NSEventModifierFlagShift]];
+    // On some European keyboards there are useful symbols which are only
+    // accessible via the Option key. We don't query for the Option key
+    // character because on most layouts just about every key has some
+    // random symbol assigned to that modifier which the user probably
+    // isn't even aware of. The user can get that character into the map
+    // by typing it directly.
 }
 
 + (jint)getKeyCodeForChar:(jchar)c;
--- a/tests/manual/events/KeyboardTest.java
+++ b/tests/manual/events/KeyboardTest.java
@@ -83,7 +83,7 @@ import javafx.stage.Stage;
 public class KeyboardTest extends Application {
 
     public static void main(String[] args) {
-        Application.launch(args);
+        Application.launch(KeyboardTest.class, args);
     }
 
     private static final String os = System.getProperty("os.name");
@@ -167,7 +167,7 @@ public class KeyboardTest extends Application {
             list.add(new KeyData(cd, null, null, null));
         }
 
-        /* Add a key with unshifted, shifted, and AltGr/Option characters */
+        /* Add a key with unshifted, shifted, and AltGr characters */
         private void add(KeyCode cd, String base, String shifted, String altGr) {
             list.add(new KeyData(cd, base, shifted, altGr));
         }
@@ -249,9 +249,9 @@ public class KeyboardTest extends Application {
 
             /*
              * ENTER is assigned to both Return and Enter which generate
-             * different characters.
+             * different characters. Platforms should always target Return.
              */
-            add(KeyCode.ENTER, "wild");
+            add(KeyCode.ENTER, "\r");
 
             if (onMac) {
                 add(KeyCode.COMMAND);
@@ -334,8 +334,8 @@ public class KeyboardTest extends Application {
             builder.addCommon();
             builder.addLetters();
 
-            /* Include one combination that involves AltGr/Option. */
-            final String altGrFive = (onMac ? "{" : "[");
+            /* Include one combination that involves AltGr. */
+            final String altGrFive = "[";
 
             /*
              * On a French layout the unshifted top-row keys (which generate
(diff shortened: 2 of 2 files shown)


beldenfox:
Updated the JDK and PR title and added a new test case. The underlying problem is that KeyCharacterCombinations don't work until the user types the character directly which leads to confusing behavior e.g. Cmd+'+' doesn't work on the main keyboard until the user types '+' but then stops working if they type '+' on the numeric keypad.

beldenfox:
Adding a comment to keep the bot away. Still hoping this PR gets some review cycles (it's only a tiny little thin one).

beldenfox:
Adding a comment to keeper the robot away.

beldenfox:
This is a customer facing issue (see defold/defold#5850). Like most shortcut-related problems there are duplicates in the Defold issue list because KeyCharacterCombinations are malfunctioning in different ways on different platforms and the behavior varies based on the keyboard layout.

kevinrushforth:
@andy-goryachev-oracle Can you be the primary reviewer on this?

beldenfox:
@andy-goryachev-oracle You're right, this PR also deals with the problem of the map retaining stale entries when the user switches layout. I added details about this to the JBS ticket. We could treat it as a separate bug if you want, to me it's just the sort of standard housekeeping one does when caching information.

I also added a note about the KeyboardTest app. This PR scales back that test a bit but this is the first time we've been able to run the KeyCharacterCombinations on the Mac so that portion is basically a new test.

Let me know if you want further information in the JBS ticket.

andy-goryachev-oracle:
@beldenfox thank you.  I still not entirely sure what issue is being tested and how to fix it, between this PR and JBS.
For example, when I run the application listed in JBS description with your fix, I get one "HI" - is it expected?  (I also get 1 "HI" on Windows)

On macOS, new KeyCodeCombination(KeyCode.MINUS, KeyCombination.CONTROL_DOWN) does not equal to new KeyCharacterCombination("-", KeyCombination.CONTROL_DOWN) (same on Windows)

Interestingly, Ctrl+- on macOS with the US keyboard produces

```
KeyEvent{type=KEY_PRESSED, character=<\u00>, text=, code=CONTROL, control}
KeyEvent{type=KEY_PRESSED, character=<\u00>, text=, code=MINUS, control}
KeyEvent{type=KEY_TYPED, character=<\u1F>, text=, code=UNDEFINED, control}
KeyEvent{type=KEY_RELEASED, character=<\u00>, text=, code=MINUS, control}
KeyEvent{type=KEY_RELEASED, character=<\u00>, text=, code=CONTROL}
```

but on Windows, we don't get KEY_TYPED event:

beldenfox:
@andy-goryachev-oracle What keyboard layout do you normally use on Windows? And Mac? When talking about the behavior of punctuation this is a crucial bit of information.

> but on Windows, we don't get KEY_TYPED event:
> is this expected?

When you hold down Ctrl and press a key the platform may or may not generate a low-ASCII control code. JavaFX has never tried to make the platforms consistent here, Glass just passes on whatever the OS generates. In any case the TextInputControls treat these TYPED events as nuisances, if you look in TextInputControlBehavior.java and read `defaultKeyTyped` you'll see how it tries to filter out control codes and anything else that looks like it was associated with a shortcut.

andy-goryachev-oracle:
> What keyboard layout do you normally use on Windows?

US layout on both, at least when the events were captured in the previous comment.

andy-goryachev-oracle:
let me test with #1264 integrated.

andy-goryachev-oracle:
All right, so I see "HI" printed twice on macOS and Windows.  I guess we are good here.

A couple of other observations:

1. Using French keyboard on mac (and windows), Ctrl-- is expected to produce the key code RIGHT_PARENTHESIS, correct?
```
KeyEvent{type=KEY_PRESSED, character=<\u00>, text=, code=CONTROL, control}
KeyEvent{type=KEY_PRESSED, character=<\u00>, text=, code=RIGHT_PARENTHESIS, control}
KeyEvent{type=KEY_TYPED, character=<\u1B>, text=, code=UNDEFINED, control}  <-- missing on Windows, as expected
KeyEvent{type=KEY_RELEASED, character=<\u00>, text=, code=RIGHT_PARENTHESIS, control}
KeyEvent{type=KEY_RELEASED, character=<\u00>, text=, code=CONTROL}
```

2. Using KeyboardTest, I am getting:

|Layout|w/o combinations|w/o keypad combinations|with all combinations|
|---|---|---|---|
|US|0|0|0

beldenfox:
> 1. Using French keyboard on mac (and windows), Ctrl-- is expected to produce the key code RIGHT_PARENTHESIS, correct?

Yes, that key turns into right parenthesis with a French layout.

>  2. Using KeyboardTest, I am getting:

You should not see any failures as long as the test and your selected keyboard match. It looks like you were running the French, German, and Spanish tests while your keyboard was set to U.S. English. Unfortunately you have to change the keyboard layout manually, there's no API in Java that allows me to do it. That's also why I can't run all the tests in one go.

andy-goryachev-oracle:
> as long as the test and your selected keyboard match.

ah!  my fault, did not read the javadoc.  let me try again.

andy-goryachev-oracle:
0 failures with German.

Do you think some other layouts should be added to the keyboard test?
(Suggestion: Arabic, Hebrew, Japanese, French, Russian)

kevinrushforth:
> Do you think some other layouts should be added to the keyboard test? (Suggestion: Arabic, Hebrew, Japanese, French, Russian)
> 
> Another suggestion: maybe add the instructions to the KeyboardTest window?

Those both sound like good suggestions. I recommend taking them in a follow-up issue rather than as part of this PR.

andy-goryachev-oracle (approved):

beldenfox:
I created [JDK-8320216](https://bugs.openjdk.org/browse/JDK-8320216) for adding instructions to the test app.

It would be nice to have a non-Latin language in the test so I will look into adding Russian, Arabic, or Hebrew.

I have no idea where to begin with IME-based languages like Japanese and am concerned that the resulting tests would be very platform-specific. In any case that would be an entirely different test app.

kevinrushforth (approved):
Looks good. I tested it only on a US keyboard layout, but Andy has done more extensive testing.

andy-goryachev-oracle:
> I have no idea where to begin with IME-based languages like Japanese

good point.  IME is probably out of scope then.

yososs:
!! **This PR causes block-level problems in Japanese environment.**

When Japanese IME is turned on,
key input is ignored when Japanese IME is turned on after merging this PR.
We confirmed that the problem is not reproduced by reverting back to the code before the correction of this PR.

@kevinrushforth @andy-goryachev-oracle

beldenfox:
@yososs I can't reproduce this with my test apps or with the MonkeyTester. I tried both Kana and Romaji on macOS 14 but didn't try every possible input mode.

- Which version of macOS are you using?
- Which input method and mode are you using?
- What KeyCharacterCombinations does your app use?

yososs:
The environment where I confirmed the problem is as follows.

* macOS Ventura 13.6 (arm64)
* Xcode 15.0.1

The IME settings are as follows.

* [ ] ABC
* [x] 日本語

This issue causes text input and shortcut operations to be ignored when using the IME. Almost all apps are affected.

beldenfox:
I can reproduce this on 13.6. The `charactersByApplyingModifiers` call is modifying the event in some way which is disturbing the IME. This is surprising given how that call is documented and that it's been around since 10.15. It looks like the fix is to do all this processing after we call `handleEvent:` on the NSTextInputContext.

beldenfox:
The problem occurs on macOS 12 and 13 which is as far back as I can test. Calling charactersByApplyingModifiers modifiers the NSEvent in some way that causes IMEs to malfunction. I've verified that Chinese - Traditional is also affected.

The solution is to use the same Carbon calls that I used in PR #425. I've written the code and will perform some more tests before submitting a PR. I'm not sure if the right approach is to re-open the original bug and rejuvenate this PR or submit a separate bug. The fix for the original bug will need to be re-tested.

yososs:
In language environments that use IMEs, the side effects of this change will be considered block-level. For the app developer, I hope that by his next OpenJFX release there will be a plan to keep this side effect out of the release's binaries.

andy-goryachev-oracle:
Do we want to revert the fix?  In any case, we need a new bug report ticket.

beldenfox:
I entered a new ticket [JDK-8320773](https://bugs.openjdk.org/browse/JDK-8320773). I will submit a fix today.

I can test macOS 12, 13, and 14 on Apple Silicon but can't go as far back as 11.

PR #1210, opened 2023-08-15
8313956: focusWithin on parents of a newly-added focused node is not updated

mstr2:
This PR fixes an issue with the way `focusWithin` bits are adjusted in the scene graph. Previously, the `focusWithin` counts of all parents of a removed node would be decreased if the removed node has a non-zero `focusWithin` count. This PR adds logic for the opposite scenario: the `focusWithin` count is increased on all parents of a newly-added node that is already focused (or contains other focused nodes).

---------
### Progress

### Issue
 * [JDK-8313956](https://bugs.openjdk.org/browse/JDK-8313956): focusWithin on parents of a newly-added focused node is not updated (**Bug** - P4)

### Reviewers
 * [Ambarish Rapte](https://openjdk.org/census#arapte) (@arapte - **Reviewer**) 🔄 Re-review required (review applies to [fd85c902](https://git.openjdk.org/jfx/pull/1210/files/fd85c90268bafaaaa8371d7dae9f54d228892810))

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1210/head:pull/1210` \
`$ git checkout pull/1210`

`$ git checkout pull/1210` \
`$ git pull https://git.openjdk.org/jfx.git pull/1210/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1210`

`$ git pr show -t 1210`

</details>

<a href="https://git.openjdk.org/jfx/pull/1210.diff">https://git.openjdk.org/jfx/pull/1210.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1210#issuecomment-1679654885)

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -8322,12 +8322,7 @@ public abstract class Node implements EventTarget, Styleable {
          */
         void adjust(int change) {
             count += change;
-
-            if (change > 0) {
-                set(true);
-            } else if (count == 0) {
-                set(false);
-            }
+            set(count > 0);
         }
     };
 
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -956,7 +956,6 @@ public abstract class Node implements EventTarget, Styleable {
                 @Override
                 protected void invalidated() {
                     if (oldParent != null) {
-                        clearParentsFocusWithin(oldParent);
                         if (nodeTransformation != null && nodeTransformation.listenerReasons > 0) {
                             ((Node) oldParent).localToSceneTransformProperty().removeListener(
                                     nodeTransformation.getLocalToSceneInvalidationListener());
@@ -965,6 +964,10 @@ public abstract class Node implements EventTarget, Styleable {
                     updateDisabled();
                     computeDerivedDepthTest();
                     final Parent newParent = get();
+
+                    // Update the focus bits before calling reapplyCss(), as the focus bits can affect CSS styling.
+                    updateParentsFocusWithin(oldParent, newParent);
+
                     if (newParent != null) {
                         if (nodeTransformation != null && nodeTransformation.listenerReasons > 0) {
                             ((Node) newParent).localToSceneTransformProperty().addListener(
@@ -8171,21 +8174,41 @@ public abstract class Node implements EventTarget, Styleable {
     }
 
     /**
-     * Called when the current node was removed from the scene graph.
-     * If the current node has the focusWithin bit, we also need to clear the focusWithin bits of this
-     * node's parents. Note that a scene graph can have more than a single focused node, for example when
-     * a PopupWindow is used to present a branch of the scene graph. Since we need to preserve multi-level
+     * Called when the current node was removed from or added to the scene graph.
+     * If the current node has the focusWithin bit, we also need to clear and set the focusWithin bits of this
+     * node's old and new parents. Note that a scene graph can have more than a single focused node, for example
+     * when a PopupWindow is used to present a branch of the scene graph. Since we need to preserve multi-level
      * focus, we need to adjust the focus-within count on all parents of the node.
      */
-    private void clearParentsFocusWithin(Node oldParent) {
-        if (oldParent != null && focusWithin.get()) {
+    private void updateParentsFocusWithin(Node oldParent, Node newParent) {
+        if (!focusWithin.get()) {
+            return;
+        }
+
+        if (oldParent != null) {
             Node node = oldParent;
             while (node != null) {
                 node.focusWithin.adjust(-focusWithin.count);
                 node = node.getParent();
             }
+        }
 
+        if (newParent != null) {
+            Node node = newParent;
+            while (node != null) {
+                node.focusWithin.adjust(focusWithin.count);
+                node = node.getParent();
+            };
+        }
+
+        // Since focus changes are atomic, we only fire change notifications after
+        // all changes are committed on all old and new parents.
+        if (oldParent != null) {
             oldParent.notifyFocusListeners();
+        }
+
+        if (newParent != null) {
+            newParent.notifyFocusListeners();
         }
     }
 
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -8185,21 +8185,17 @@ public abstract class Node implements EventTarget, Styleable {
             return;
         }
 
-        if (oldParent != null) {
-            Node node = oldParent;
-            while (node != null) {
-                node.focusWithin.adjust(-focusWithin.count);
-                node = node.getParent();
-            }
+        Node node = oldParent;
+        while (node != null) {
+            node.focusWithin.adjust(-focusWithin.count);
+            node = node.getParent();
         }
 
-        if (newParent != null) {
-            Node node = newParent;
-            while (node != null) {
-                node.focusWithin.adjust(focusWithin.count);
-                node = node.getParent();
-            };
-        }
+        node = newParent;
+        while (node != null) {
+            node.focusWithin.adjust(focusWithin.count);
+            node = node.getParent();
+        };
 
         // Since focus changes are atomic, we only fire change notifications after
         // all changes are committed on all old and new parents.
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -8323,7 +8323,7 @@ public abstract class Node implements EventTarget, Styleable {
         void adjust(int change) {
             count += change;
 
-            if (count == 1) {
+            if (change > 0) {
                 set(true);
             } else if (count == 0) {
                 set(false);
--- a/modules/javafx.graphics/src/test/java/test/javafx/scene/FocusTest.java
+++ b/modules/javafx.graphics/src/test/java/test/javafx/scene/FocusTest.java
@@ -1105,4 +1105,27 @@ public class FocusTest {
         assertNotFocusWithin(node4);
     }
 
+    @Test public void testFocusWithinBitsAreSetOnParentsWhenAddedNodeIsAlreadyFocused() {
+        class N extends Group {
+            N(Node... children) { super(children); }
+            void _setFocused(boolean value) { setFocused(value); }
+        }
+
+        N node1, node2, node3 = new N();
+
+        scene.setRoot(
+            node1 = new N(
+                node2 = new N()
+            ));
+
+        node3._setFocused(true);
+        assertNotFocusWithin(node1);
+        assertNotFocusWithin(node2);
+
+        node2.getChildren().add(node3);
+        assertIsFocusWithin(node1);
+        assertIsFocusWithin(node2);
+        assertIsFocusWithin(node3);
+    }
+
 }
--- a/modules/javafx.graphics/src/test/java/test/javafx/scene/FocusTest.java
+++ b/modules/javafx.graphics/src/test/java/test/javafx/scene/FocusTest.java
@@ -1116,7 +1116,8 @@ public class FocusTest {
         scene.setRoot(
             node1 = new N(
                 node2 = new N()
-            ));
+            )
+        );
 
         node3._setFocused(true);
         assertNotFocusWithin(node1);
@@ -1139,10 +1140,12 @@ public class FocusTest {
         scene.setRoot(
             node1 = new N(
                 node2 = new N()
-            ));
+            )
+        );
 
         node3 = new N(
-            node4 = new N());
+            node4 = new N()
+        );
 
         node3._setFocused(true);
         node4._setFocused(true);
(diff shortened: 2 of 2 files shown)


mstr2:
@Maran23 maybe you could take a look, too.

Maran23:
> @Maran23 maybe you could take a look, too.

Yes! It is on my list.

arapte:
Looks all neat to me. The FocusTest does cover a variety of scenarios.
The change is specific to FocusWithin and seems unlikely to have any side effects.

arapte (approved):

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -8171,22 +8174,38 @@ final void notifyFocusListeners() {
...
+        }
+
+        node = newParent;
+        while (node != null) {
+            node.focusWithin.adjust(focusWithin.count);
+            node = node.getParent();
+        };
+
+        // Since focus changes are atomic, we only fire change notifications after
+        // all changes are committed on all old and new parents.
```
hjohn, line 8201: Is it possible a user can interfere with this process by adding a listener to `focusWithinProperty`?  It fires events when it goes from `true` to `false` and vice versa.

hjohn: I also noticed this code, which I think will not do the correct thing when `change` is not 1 or -1:

        void adjust(int change) {
            count += change;

            if (count == 1) {
                set(true);
            } else if (count == 0) {
                set(false);
            }
        }

For example, if focus within currently is `false`, and it is "adjusted" by 5, the count goes from 0 to 5 in one go.  It will remain `false` then.  Is this intended?

hjohn (commented):
I had a look, but have some questions.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
hjohn: > Is it possible a user can interfere with this process by adding a listener to `focusWithinProperty`? It fires events when it goes from `true` to `false` and vice versa.

Can disregard this, it seems the property uses a special base class that breaks with the general property contract.

Maran23:
One thing I noticed in the example in the ticket is, that when I click the button below the table, the `focusWithin` is still 1.
image

It seems like the `focusWithin` count is always 2 when the `TableView` is selected. When clicking the button, it will go down to 1, but not 0.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
mstr2: > I also noticed this code, which I think will not do the correct thing when `change` is not 1 or -1:
> 
> ```
>     void adjust(int change) {
>         count += change;
> 
>         if (count == 1) {
>             set(true);
>         } else if (count == 0) {
>             set(false);
>         }
>     }
> ```
> 
> For example, if focus within currently is `false`, and it is "adjusted" by 5, the count goes from 0 to 5 in one go. It will remain `false` then. Is this intended?

Good catch! I've added another test for this edge case.

mstr2:
@arapte I've added another test for an edge case, can you review again?

Maran23:
> That's because the table row remains `focused` even if you change the scene's focus owner to the "focus lost" button

I see. But to be honest, the table focus code is not optimal as it contains multiple focus workarounds which can be better solved with `focusWithin`. So we should first look what can be improved here. Thanks for looking that up.

One a side note, I remember that there is a ticket to clarify what to do when the `focusWithin` count is negative. We don't need to solve this in this PR, bur maybe for the future we should consider logging an info or something similar, in case we find another case where this can get negative.

mstr2:
> One a side note, I remember that there is a ticket to clarify what to do when the `focusWithin` count is negative. We don't need to solve this in this PR, bur maybe for the future we should consider logging an info or something similar, in case we find another case where this can get negative.

The `focusWithin` count cannot be negative, and there is no possible way for applications to misuse the API to cause a negative count. If it ends up being negative, that's because of a bug, and I think that bugs should be fixed and not logged.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -8323,7 +8323,7 @@ public String getName() {
         void adjust(int change) {
             count += change;
 
-            if (count == 1) {
+            if (change > 0) {
                 set(true);
             } else if (count == 0) {
                 set(false);
```
hjohn, line 8329: This looks a bit odd due to the use of two different variables, but it works (`count > 0` would also have worked).  I personally would favor something without any if statements as it leaves less room for misinterpretation in this case:

     count += change;

     set(count > 0);

modules/javafx.graphics/src/test/java/test/javafx/scene/FocusTest.java
```diff
@@ -1128,4 +1128,35 @@ class N extends Group {
...
+        N node1, node2, node3, node4;
+
+        scene.setRoot(
+            node1 = new N(
+                node2 = new N()
+            ));
+
+        node3 = new N(
+            node4 = new N());
+
```
hjohn, line 1146: ignore: I think this looks a bit inconsistent (in the first version not all parenthesis are closed on the last line, while they are in the second version).  Personally I favor this, similar to how curly braces are used):
```suggestion
        scene.setRoot(
            node1 = new N(
                node2 = new N()
            )
        );

        node3 = new N(
            node4 = new N()
        );

```

hjohn (approved):
LGTM

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
andy-goryachev-oracle: set(count > 0);
looks so much better

mstr2: Agreed.

modules/javafx.graphics/src/test/java/test/javafx/scene/FocusTest.java
mstr2: Done.

andy-goryachev-oracle:
question: what happens when the user navigates away when editing the cell?  it looks like the fix correctly handles the situation.  are there any other possible scenarios?

Screenshot 2023-08-18 at 13 53 02

hjohn (approved):

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -8171,22 +8174,38 @@ final void notifyFocusListeners() {
...
-                node = node.getParent();
-            }
+    private void updateParentsFocusWithin(Node oldParent, Node newParent) {
+        if (!focusWithin.get()) {
+            return;
+        }
+
+        Node node = oldParent;
+        while (node != null) {
+            node.focusWithin.adjust(-focusWithin.count);
```
andy-goryachev-oracle, line 8190: suppose we are removing a top level Pane with a bunch of Nodes in it that have non-zero focusWithin.count values.
the parent of said Pane gets a zero count (which is correct), whereas the children of said Pane have their counts unchanged (I assume).

what happens when the same Pane gets added back to the scene graph?  wouldn't we enter an inconsistent state?

mstr2: > suppose we are removing a Pane with a bunch of Nodes in it that have non-zero focusWithin.count values. the parent of said Pane gets a zero count (which is correct), whereas the children of said Pane have their counts unchanged (I assume).

Correct.

> what happens when the same Pane gets added back to the scene graph? wouldn't we enter an inconsistent state?

That's what this patch fixes. If we later add the pane that contains a number of focused nodes back to the scene graph, the `focusWithin` count of its new parents (which was 0 before) is now incremented to match the number of contained focused nodes. The code starts in L8194.

mstr2:
> question: what happens when the user navigates away when editing the cell? it looks like the fix correctly handles the situation. are there any other possible scenarios?

I don't think there are, because now we're correctly handling the two scenarios that affect the `focusWithin` count:
1. Removing a non-zero focusWithin node from the scene graph
2. Adding a non-zero focusWithin node from the scene graph

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
andy-goryachev-oracle: thank you for explanations!

one more question: suppose, prior to removal of the Pane it contained a focused (Scene.focusOwner) Node.
1. once the Pane is removed, what happens to Scene.focusOwner?
2. once the Pane is added back, would we have an inconsistent state in respect to Scene.focusOwner (if some other node requested focus in the mean time, wouldn't we have two focused nodes?)

mstr2: > thank you for explanations!
> 
> one more question: suppose, prior to removal of the Pane it contained a focused (Scene.focusOwner) Node.
> 
> 1. once the Pane is removed, what happens to Scene.focusOwner?

There's quite a bit of logic happening. `Scene.focusCleanup()` is a good starting point. Basically, we need to find a new focus owner if the old focus owner is removed from the scene graph.

> 2. once the Pane is added back, would we have an inconsistent state in respect to Scene.focusOwner (if some other node requested focus in the mean time, wouldn't we have two focused nodes?)

It's not inconsistent, but it wouldn't be the same state as before. The `focused` state would remain, but `Scene.focusOwner` would not be restored to the previous node. That part of the focus story hasn't changed with the introduction of `focus-visible` and `focus-within`.

Maran23:
> The `focusWithin` count cannot be negative, and there is no possible way for applications to misuse the API to cause a negative count. If it ends up being negative, that's because of a bug, and I think that bugs should be fixed and not logged.

Yes, the reason I was asking this is that we actually have this problem right now, that it could be negative, although it should not. If we would have logged that, we probably found that bug much earlier. In any case, I think with the new logic and the `count > 0`, it is fine as is.

Maran23 (approved):

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
andy-goryachev-oracle: So, in the scenario # 2, we would arrive at situation when the previously focused Node is still decorated as having the focus, but won't be getting the events since Scene.focusOwner is different?  I suppose it's the way the whole thing is designed to work.
Thank you for the explanations!

andy-goryachev-oracle (approved):

modules/javafx.graphics/src/test/java/test/javafx/scene/FocusTest.java
tcfurrer: FYI: Using jfx21, I have observed a case where the count becomes negative. If I can create a minimal example, I will file a bug report.  However it's happening in a complex application, so it make take time for me to reduce it down to something I can reasonably provide.

tcfurrer:
> > The `focusWithin` count cannot be negative, and there is no possible way for applications to misuse the API to cause a negative count. If it ends up being negative, that's because of a bug, and I think that bugs should be fixed and not logged.
> 
> Yes, the reason I was asking this is that we actually have this problem right now, that it could be negative, although it should not. If we would have logged that, we probably found that bug much earlier. In any case, I think with the new logic and the `count > 0`, it is fine as is.

Just to clarify...
If I have a case (with jfx21) where the count goes negative, is this still a bug that I should file?  Or are you saying that it's understood as expected now?

tcfurrer:
@mstr2 or @Maran23 Is the term "focus owner" equivalent to "Node that currently has input focus"?

I ask because the javadoc on Node.focusWithin says "Indicates whether this Node or any of its descendants currently has the input focus.", but in discussion comments it has been said that focusWithin includes a TableView's "focused item" even if that is not the focus owner.

It would be great if the javadoc for Node.focusWithin could be improved to avoid this kind of widespread confusion about what it really means...

mstr2:
> > > The `focusWithin` count cannot be negative, and there is no possible way for applications to misuse the API to cause a negative count. If it ends up being negative, that's because of a bug, and I think that bugs should be fixed and not logged.
> > 
> > 
> > Yes, the reason I was asking this is that we actually have this problem right now, that it could be negative, although it should not. If we would have logged that, we probably found that bug much earlier. In any case, I think with the new logic and the `count > 0`, it is fine as is.
> 
> @Maran23 Just to clarify... If I have a case (with jfx21) where the count goes negative, is this still a bug that I should file? Or are you saying that it's understood as expected behavior now?

If you have a situation where the focusWithin count is negative, please file it as a bug. That's not expected behavior.

tcfurrer:
@Maran23 Ok, I will try to get a bug filed eventually.  It will take time because I've only observed this in a complex application, and it may be difficult to produce a minimal example.

mstr2:
> @mstr2 or @Maran23 Does "focus owner" == "input focus" == "scene focus owner" == where key events will be sent?

This is true in general. Note that an unfocused window will not receive input events, but the scene can have a focus owner nonetheless.

> I ask because the javadoc on Node.focusWithin says "Indicates whether this Node or any of its descendants currently has the **_input focus_**.", but in discussions it has been said that focusWithin includes a TableView's "focused item" even if that is not the scene's current focus owner (i.e. where key events will be sent).
> 
> It would be great if the javadoc for Node.focusWithin could be improved to avoid this kind of widespread confusion about what it really means.

Yes, the documentation is misleading. We should at the very least fix that.

By the way, there's an ongoing discussion on the OpenJFX mailing list: https://mail.openjdk.org/pipermail/openjfx-dev/2023-October/042864.html

PR #1213, opened 2023-08-17
8313651: Add 'final' keyword to public property methods in controls

andy-goryachev-oracle:
In the Control hierarchy, all property accessor methods must be declared `final`.

Added a test to check for missing `final` keyword and added the said keyword where required.

---------
### Progress
- [x] Change requires CSR request [JDK-8314581](https://bugs.openjdk.org/browse/JDK-8314581) to be approved

### Issues
 * [JDK-8313651](https://bugs.openjdk.org/browse/JDK-8313651): Add 'final' keyword to public property methods in controls (**Bug** - P3)
 * [JDK-8314581](https://bugs.openjdk.org/browse/JDK-8314581): Add 'final' keyword to public property methods in controls (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1213/head:pull/1213` \
`$ git checkout pull/1213`

`$ git checkout pull/1213` \
`$ git pull https://git.openjdk.org/jfx.git pull/1213/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1213`

`$ git pr show -t 1213`

</details>

<a href="https://git.openjdk.org/jfx/pull/1213.diff">https://git.openjdk.org/jfx/pull/1213.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1213#issuecomment-1684466201)


modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
```diff
@@ -0,0 +1,239 @@
...
+        StringBuilder sb = new StringBuilder(64);
+        sb.append(prefix);
+        for (int i = 0; i < name.length(); i++) {
+            char c = name.charAt(i);
+            if (i == 0) {
+                sb.append(Character.toUpperCase(c));
+            } else {
+                sb.append(c);
+            }
+        }
```
hjohn, line 216:
```suggestion
       sb.append(prefix);
       sb.append(Character.toUpperCase(name.charAt(0)));
       sb.append(name, 1, name.length);
```

hjohn (approved):

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
andy-goryachev-oracle: good idea, thanks!

hjohn (approved):

kevinrushforth:
At first glance this looks good, but I'll need time to get back to it. And thanks for adding a test.

andy-goryachev-oracle:
> And thanks for adding a test.

I feel the test can be improved by collecting the classes automatically, so as to catch possible issues when new controls are added (it'll be a full classpath scan though).  While the scan should be limited to Controls, I wonder if other classes can benefit from enforcing the 'final' constraint?

kevinrushforth:
Um, "classpath"?  JavaFX is a collection of modules that are found on the module path. In any case, this seems better done as follow-up issue, if you want to file one.

Btw, having an automatic scan of "everything" seems like overkill, and isn't really desirable, so if you do want to go down this route, I'd keep it simple. Since any non-exported package (i.e., any pacakge that isn't under `javafx.**`) will already need to be explicitly listed in the `test/addExports` file, and since we so very rarely add a new package, you might consider having a hard coded List of packages that you enumerate looking for all classes in each package for ones with properties.

andy-goryachev-oracle:
> Um, "classpath"?

The problem I was referring to is how to enumerate all descendants of Control in javafx.* hierarchy, and I don't have a good solution.  So, no change in the unit test is being planned, and all newly developed controls will have to be added manually.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
```diff
@@ -0,0 +1,233 @@
...
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    private String err(Method method, String message) {
+        return method + " " + message;
+    }
```
nlisker, line 232: (This comment didn't appear in the review for some reason.)

I think that the code could be a bit more clean. I took a stab at it:

```java
    Set<Class<?>> CLASSES = Set.of(/*...*/);

    private void check(Class<?> cls) {
        Map<String, Method> methods = Stream.of(cls.getMethods()).collect(Collectors.toMap(Method::getName, m -> m)); // the loop is also fine
        
        for (var method : methods.values()) {
            String name = method.getName();
            if (name.endsWith("Property")) {
                checkModifiers(method);

                String propName = name.substring(0, name.length() - "Property".length());
                check(methods, propName, "get", 0);
                check(methods, propName, "set", 1);
                check(methods, propName, "is", 0);
            }
        }
    }

    private void check(Map<String, Method> methods, String name, String prefix, int numArgs) {
        String methodName = new StringJoiner("", prefix, name.substring(0, 1).toUpperCase() + name.substring(1)).toString();

        Method m = methods.get(methodName);
        if (m != null && m.getParameterCount() == numArgs) {
            checkModifiers(m);
        }
    }

    private void checkModifiers(Method m) {
        int mod = m.getModifiers();
        if (Modifier.isPublic(mod) && !Modifier.isFinal(mod)) {
            if (FAIL_FAST) {
                throw new AssertionError(errorMessage(m));
            } else {
                System.err.println(errorMessage(m));
            }
        }
    }

    private String errorMessage(Method method) {
        return method + "is not final";
    }
```

andy-goryachev-oracle: Thank you for suggestions!

Frankly, I don't see much value in rewriting of this method.  A more interesting problem is how to enumerate all descendants of the Control class  using the package name, in such a way that will reliably work in an IDE, from a JAR, or in any test framework (jtreg or otherwise).

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
```diff
@@ -0,0 +1,233 @@
...
+ * Currently uses a list of classes, so any new Controls must be added manually.
+ * Perhaps the test should scan classpath and find all the Controls automagically.
+ */
+public class ControlPropertiesTest {
+    private static final boolean FAIL_FAST = true;
+
+    // or perhaps collect all classes in a package as described here:
+    // https://stackoverflow.com/questions/28678026/how-can-i-get-all-class-files-in-a-specific-package-in-java
+    private Class[] allControlClasses() {
+        return new Class[] {
```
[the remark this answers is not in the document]
andy-goryachev-oracle: Respectfully disagree - these classes (in javafx.scene.control.cell) are public and in the Control hierarchy, so are subject of the 'final' limitation.

There are no duplicate classes.

The bigger problem is (my) inability to extract the list/set/array of classes automatically.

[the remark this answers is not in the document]
nlisker: > these classes (in javafx.scene.control.cell) are public and in the Control hierarchy, so are subject of the 'final' limitation.

I'm confused. Which classes should appear in this list exactly (what are the rules to determine that)?

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
nlisker: > edit: about the test itself - I see little difference between the two version, but will gladly use your version if you feel strongly about it.

I don't, I just think it's more readable. Fine to leave as is. What I might strongly suggest is extracting the `checkModifiers` method that is used in 2 places in your version.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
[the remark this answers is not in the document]
andy-goryachev-oracle: > Which classes should appear in this list exactly

subclasses of javafx.scene.control.Control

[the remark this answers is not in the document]
nlisker: Ah, this is not easy. I thought that we need the classes in the `control` package. At this point it might just be simpler to use the hardcoded class list you already made and "hope" that new classes are added. There won't be many of these I suspect.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
andy-goryachev-oracle: done, and also fixed a stupid mistake ... method names are not unique.
many thanks!

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
```diff
@@ -95,12 +94,14 @@
  * Perhaps the test should scan classpath and find all the Controls automagically.
  */
 public class ControlPropertiesTest {
-    private static final boolean FAIL_FAST = true;
 
+    private static final boolean FAIL_FAST = !true;
```
hjohn, line 98: Did you intend to commit this?  It was `true` before which seems good for tests, otherwise I'd really suggest using `false` here.

hjohn (approved):

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
```diff
@@ -0,0 +1,220 @@
...
+            WebColorField.class
+        );
+    }
+
+    /**
+     * Tests for missing final keyword in Control properties and their accessors.
+     */
+    @Test
+    public void testMissingFinalMethods() {
+        for (Class c: allControlClasses()) {
```
nlisker, line 172: I think we use a space before `:`. Don't know if it's being enforced. If you change it there are 2 more places.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
```diff
@@ -0,0 +1,220 @@
...
+            TreeCell.class,
+            TreeTableCell.class,
+            TreeTableRow.class,
+            TreeTableView.class,
+            WebColorField.class
+        );
+    }
+
+    /**
+     * Tests for missing final keyword in Control properties and their accessors.
```
nlisker, line 168: Technically, we're looking at mutators too, not just accessors.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
```diff
@@ -0,0 +1,220 @@
...
+import javafx.scene.control.cell.TextFieldTreeTableCell;
+import org.junit.jupiter.api.Test;
+import com.sun.javafx.scene.control.DoubleField;
+import com.sun.javafx.scene.control.InputField;
+import com.sun.javafx.scene.control.IntegerField;
+import com.sun.javafx.scene.control.WebColorField;
+import com.sun.javafx.scene.control.skin.FXVK;
+
+/**
+ * Tests contract for properties and their accessors in the Control type hierarchy.
```
nlisker, line 91: Mutators too?

nlisker (approved):
Looks good. Gave a couple of minor optional suggestions.

As to the topic of class finding, I think it's fine to use a manual list, at least for now. The complexity of automatically detecting the classes might not be worth the hassle. Leaving it up to you.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
andy-goryachev-oracle: sorry, committed by accident.  thanks for noticing!

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
andy-goryachev-oracle: Adding superfluous spaces increases our carbon footprint ;-)
Modified the formatter to increase the footprint.

kevinrushforth (commented):
The fix looks good. So does the test with a couple comments. I also noted a couple potential follow-up items.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
```diff
@@ -0,0 +1,221 @@
...
+                    checkModifiers(m);
+                    return;
+                }
+            }
+        }
+    }
+
+    private void checkModifiers(Method m) {
+        int mod = m.getModifiers();
+        if (Modifier.isPublic(mod) && !Modifier.isFinal(mod)) {
```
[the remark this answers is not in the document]
andy-goryachev-oracle: easy enough to add, and makes sense.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
```diff
@@ -0,0 +1,225 @@
...
+     */
+    @Test
+    public void testMissingFinalMethods() {
+        for (Class c : allControlClasses()) {
+            check(c);
+        }
+    }
+
+    private void check(Class cls) {
+        Method[] methods = cls.getMethods();
```
kevinrushforth, line 177: This will only return public methods. Since you modified `checkModifiers` to check protected methods, you will need to use `getDeclaredMethods` instead. This will return protected (and package scope) methods as well as public methods. It won't return any method in a superclass, but since the set of classes you are checking includes intermediate classes (such as `ButtonBase` and `Labeled`), you will still get the coverage (and without redundantly checking the same method in those classes multiple times).

andy-goryachev-oracle: Right, what was I thinking??
And we also have Region.setWidth() which is protected and not final by design.  Reverting the last change.

kevinrushforth (approved):
Looks good.

kevinrushforth:
@nlisker @hjohn Can you re-review?

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
```diff
@@ -0,0 +1,219 @@
...
+ * Currently uses a list of classes, so any new Controls must be added manually.
+ * Perhaps the test should scan modulepath and find all the Controls automagically.
+ */
+public class ControlPropertiesTest {
+
+    private static final boolean FAIL_FAST = true;
+
+    // list all current descendants of Control class.
+    private Set<Class> allControlClasses() {
+        return Set.of(
```
nlisker, line 103: Not that it matters much since the method is called once, but why is there a need for a method that returns the same constant set? The set can just be a constant: `private final Set<Class<?>> = Set.of(...);`

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
```diff
@@ -0,0 +1,219 @@
...
+
+    private void checkModifiers(Method m) {
+        int mod = m.getModifiers();
+        if (Modifier.isPublic(mod) && !Modifier.isFinal(mod)) {
+            String msg = m + " is not final.";
+            if (FAIL_FAST) {
+                throw new AssertionError(msg);
+            } else {
+                System.err.println(msg);
+            }
```
nlisker, line 216: No need for an `else` clause since `throw` will exit the scope of the method if it's reached.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
```diff
@@ -0,0 +1,219 @@
...
+ *
+ * Currently uses a list of classes, so any new Controls must be added manually.
+ * Perhaps the test should scan modulepath and find all the Controls automagically.
+ */
+public class ControlPropertiesTest {
+
+    private static final boolean FAIL_FAST = true;
+
+    // list all current descendants of Control class.
+    private Set<Class> allControlClasses() {
```
nlisker, line 102: Please use `Class<?>` instead of raw types.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
```diff
@@ -0,0 +1,219 @@
...
+/**
+ * Tests contract for properties and their accessors and mutators
+ * in the Control type hierarchy.
+ *
+ * Currently uses a list of classes, so any new Controls must be added manually.
+ * Perhaps the test should scan modulepath and find all the Controls automagically.
+ */
+public class ControlPropertiesTest {
+
+    private static final boolean FAIL_FAST = true;
```
nlisker, line 99: Please add a short comment on what this is for. Like "If true, an exception will be thrown on encountering the first violating class, otherwise a list of all violating cases will be printed".

nlisker (commented):
Looks good. Added some minor comments.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
andy-goryachev-oracle: will keep as a method, if we ever get to extracting classes at run time.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
andy-goryachev-oracle: personal style.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlPropertiesTest.java
andy-goryachev-oracle: good point, thanks!

nlisker (approved):

kevinrushforth (approved):

hjohn (approved):

PR #1215, opened 2023-08-18
8313650: Add hasProperties method to MenuItem and Toggle

andy-goryachev-oracle:
1. Creating a new `javafx.scene.ContainsProperties` interface that declares two methods:
- public ObservableMap<Object, Object> getProperties();
- public boolean hasProperties();

2. Node, MenuItem, and Toggle now extend ContainsProperties interface.

3. Making implemented methods in Node, MenuItem, and Toggle final.

While this change is an improvement from a design point of view, it introduces a larger compatibility risk (by adding 'final' keyword similar to [JDK-8313651](https://bugs.openjdk.org/browse/JDK-8313651))

This change requires CSR.

---------
### Progress
- [ ] Change requires CSR request [JDK-8314593](https://bugs.openjdk.org/browse/JDK-8314593) to be approved

### Issues
 * [JDK-8313650](https://bugs.openjdk.org/browse/JDK-8313650): Add hasProperties method to MenuItem and Toggle (**Enhancement** - P4)
 * [JDK-8314593](https://bugs.openjdk.org/browse/JDK-8314593): Add hasProperties method to MenuItem and Toggle (**CSR**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1215/head:pull/1215` \
`$ git checkout pull/1215`

`$ git checkout pull/1215` \
`$ git pull https://git.openjdk.org/jfx.git pull/1215/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1215`

`$ git pr show -t 1215`

</details>

<a href="https://git.openjdk.org/jfx/pull/1215.diff">https://git.openjdk.org/jfx/pull/1215.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1215#issuecomment-1684466259)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/MenuItem.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/MenuItem.java
@@ -25,40 +25,39 @@
 
 package javafx.scene.control;
 
-import com.sun.javafx.beans.IDProperty;
-import javafx.collections.ObservableSet;
-import javafx.css.PseudoClass;
-import javafx.css.Styleable;
-import javafx.css.CssMetaData;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
 import javafx.beans.property.BooleanProperty;
 import javafx.beans.property.ObjectProperty;
 import javafx.beans.property.ObjectPropertyBase;
+import javafx.beans.property.ReadOnlyObjectProperty;
+import javafx.beans.property.ReadOnlyObjectWrapper;
 import javafx.beans.property.SimpleBooleanProperty;
 import javafx.beans.property.SimpleObjectProperty;
 import javafx.beans.property.SimpleStringProperty;
 import javafx.beans.property.StringProperty;
 import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
+import javafx.collections.ObservableMap;
+import javafx.collections.ObservableSet;
+import javafx.css.CssMetaData;
+import javafx.css.PseudoClass;
+import javafx.css.Styleable;
 import javafx.event.ActionEvent;
 import javafx.event.Event;
 import javafx.event.EventDispatchChain;
 import javafx.event.EventHandler;
 import javafx.event.EventTarget;
 import javafx.event.EventType;
+import javafx.scene.ContainsProperties;
 import javafx.scene.Node;
+import javafx.scene.Parent;
+import javafx.scene.control.skin.ContextMenuSkin;
 import javafx.scene.input.KeyCombination;
-
+import com.sun.javafx.beans.IDProperty;
 import com.sun.javafx.event.EventHandlerManager;
 import com.sun.javafx.scene.control.ContextMenuContent;
-import javafx.scene.control.skin.ContextMenuSkin;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-
-import javafx.beans.property.ReadOnlyObjectProperty;
-import javafx.beans.property.ReadOnlyObjectWrapper;
-import javafx.collections.ObservableMap;
-import javafx.scene.Parent;
 
 /**
  * <p>
@@ -94,7 +93,7 @@ MenuBar menuBar = new MenuBar(menu);</code></pre>
  * @since JavaFX 2.0
  */
 @IDProperty("id")
-public class MenuItem implements EventTarget, Styleable {
+public class MenuItem implements EventTarget, Styleable, ContainsProperties {
 
     /* *************************************************************************
      *                                                                         *
@@ -520,7 +519,8 @@ public class MenuItem implements EventTarget, Styleable {
      * @return an observable map of properties on this menu item for use primarily
      * by application developers
      */
-    public ObservableMap<Object, Object> getProperties() {
+    @Override
+    public final ObservableMap<Object, Object> getProperties() {
         if (properties == null) {
             properties = FXCollections.observableMap(new HashMap<>());
         }
@@ -532,7 +532,8 @@ public class MenuItem implements EventTarget, Styleable {
      * @return {@code true} if this menu item has properties.
      * @since 22
      */
-     public boolean hasProperties() {
+    @Override
+    public final boolean hasProperties() {
         return properties != null && !properties.isEmpty();
     }
 
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/MenuItem.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/MenuItem.java
@@ -527,6 +527,16 @@ public class MenuItem implements EventTarget, Styleable {
         return properties;
     }
 
+    /**
+     * Tests if MenuItem has properties.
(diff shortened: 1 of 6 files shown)


modules/javafx.controls/src/main/java/javafx/scene/control/MenuItem.java
```diff
@@ -527,6 +527,16 @@ public ObservableMap<Object, Object> getProperties() {
         return properties;
     }
 
+    /**
+     * Tests if MenuItem has properties.
+     * @return true if node has properties.
```
hjohn, line 532:
```suggestion
     * Tests if the menu item has properties.
     * @return {@code true} if the menu item has properties
```
Or can use `{@code MenuItem}`

modules/javafx.controls/src/main/java/javafx/scene/control/Toggle.java
```diff
@@ -104,4 +104,10 @@ public interface Toggle {
      * by application developers
      */
     ObservableMap<Object, Object> getProperties();
+
+    /**
+     * Tests if Toggle has properties.
+     * @return true if node has properties.
+     */
```
hjohn, line 111:
```suggestion
    /**
     * Tests if this toggle has properties.
     *
     * @return {@code true} if this toggle has properties
     * @since 22
     */
```

hjohn:
On second thought, I don't think you should add `hasProperties` to the `Toggle` interface.  Adding it to `MenuItem` should be enough to make it publically available for `RadioMenuItem`.  `ToggleButton` (which also implements `Toggle`) already inherits this method from `Node`.

This will also make it easier to retain compatibility, as otherwise you'd have to make `hasProperties` a default method in `Toggle` IMHO.

andy-goryachev-oracle:
> I don't think you should add `hasProperties` to the `Toggle` interface

I think it's ok - the interface declares getProperties(), so it makes sense to add it there.  Alternatively, there should have been "ContainsProperties" interface.

modules/javafx.controls/src/main/java/javafx/scene/control/MenuItem.java
andy-goryachev-oracle: I like this, thanks.  I won't be fixing the javadoc in other places - it's not that critical.

kevinrushforth:
Hmm. I also think might not want to put it in `Toggle` (unless we are also willing to create a new shared interface that Node, MenuItem, and Toggle all inherit). In any case, this one will take a bit of thought.

/reviewers 2

andy-goryachev-oracle:
> (unless we are also willing to create a new shared interface that Node, MenuItem, and Toggle all inherit)

that would be a clean(er) design, at the expense of slight runtime overhead traversing the interface table.
could there be any other issues?

mstr2:
> > (unless we are also willing to create a new shared interface that Node, MenuItem, and Toggle all inherit)
> 
> that would be a clean(er) design, at the expense of slight runtime overhead traversing the interface table. could there be any other issues?

I think a new interface should solve a problem, and not merely serve to organize similarly-named methods. What would you use such an interface for? I've previously proposed a content graph (see #1096), which would unify SG nodes and non-SG nodes and represent these graphs in a way that makes them independent from skins. Such a feature might warrant thinking about moving `getProperties()`/`hasProperties()` to a shared interface.

modules/javafx.graphics/src/main/java/javafx/scene/ContainsProperties.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.HashMap;
+import javafx.collections.FXCollections;
+import javafx.collections.ObservableMap;
+
+/**
+ * Indicates that the implementor contains Properties.
+ * @since 22
+ */
+public interface ContainsProperties {
```
hjohn, line 36: This name fails the "is-a" test "is a contains properties", like for example "is an `EventTarget`" or "is a `Styleable`". Suggestions:

`ApplicationPropertyProvider`
`PropertyProvider`
`ApplicationPropertyAccessor`
`PropertyAccessor`

I'd personally go for the most descriptive one (the first one).  It's long, but it's unlikely to be ever encountered as a type for a variable.

Suggestion for the docs:

    /*
     * Provides access to properties for use primarily by application developers.
     */

modules/javafx.graphics/src/main/java/javafx/scene/ContainsProperties.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Returns an observable map of properties on this object for use primarily by application developers.
+     * @return an observable map of properties on this node for use primarily by application developers
+     */
+    public ObservableMap<Object, Object> getProperties();
+
+    /**
+     * Tests if this object has properties.
+     * @return true if this object has properties.
```
hjohn, line 46:
```suggestion
     * @return {@code true} if this object has properties
```

modules/javafx.graphics/src/main/java/javafx/scene/ContainsProperties.java
```diff
@@ -0,0 +1,49 @@
...
+     */
+    public boolean hasProperties();
```
hjohn, line 48:
```suggestion
    boolean hasProperties();
```

modules/javafx.graphics/src/main/java/javafx/scene/ContainsProperties.java
```diff
@@ -0,0 +1,49 @@
...
+ * Indicates that the implementor contains Properties.
+ * @since 22
+ */
+public interface ContainsProperties {
+
+    /**
+     * Returns an observable map of properties on this object for use primarily by application developers.
+     * @return an observable map of properties on this node for use primarily by application developers
+     */
+    public ObservableMap<Object, Object> getProperties();
```
hjohn, line 42: Interface methods are always public, the keyword is just noise.

```suggestion
    ObservableMap<Object, Object> getProperties();
```

modules/javafx.graphics/src/main/java/javafx/scene/ContainsProperties.java
andy-goryachev-oracle: Shouldn't it be `ApplicationPropertiesProvider` then, since multiple properties are involved?

andy-goryachev-oracle:
converting to Draft until the discussion on Trait.* interfaces comes to a resolution.

andy-goryachev-oracle:
staying alive

Maran23:
> converting to Draft until the discussion on Trait.* interfaces comes to a resolution.

Do we want to start with the MVP here, which is just adding the `hasProperties` method and discuss the interface/trait system separately, and in a new PR?

andy-goryachev-oracle:
> Do we want to start with the MVP here, which is just adding the `hasProperties` method and discuss the interface/trait system separately, and in a new PR?

A good question!  We definitely should start with a  proposal outlining the problem statement and the public API.

PR #1216, opened 2023-08-18
8314589: javadoc build only shows the first 100 warnings and errors

kevinrushforth:
This PR bumps the limit for the number of javadoc warnings and error from 100 to 1000.

We currently use the default setting of javadoc which only shows 100 warnings and errors. This is too small, especially for warnings, and can cause us to miss new warnings that arise. Eventually we should fix all our warnings, and we have a task filed to track that, but even then I think the limit is too small.

Without this fix, our current build only shows the "first" 100 warnings it runs into. With this fix, our current build shows all 187 warnings using JDK 19.0.2 and all 191 warnings using JDK 21.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4258,6 +4258,8 @@ task javadoc(type: Javadoc, dependsOn: createMSPfile) {
 
     options.addStringOption("-since").setValue((9..jfxReleaseMajorVersion.toInteger()).join(","))
     options.addStringOption("-since-label").setValue("New API since JavaFX 9")
+    options.addStringOption("Xmaxwarns").setValue("1000")
+    options.addStringOption("Xmaxerrs").setValue("1000")
 
     options.windowTitle("${javadocTitle}")
     options.header("${javadocHeader}")


kevinrushforth:
@andy-goryachev-oracle can you review?

@nlisker You might also want to take a look.

build.gradle
```diff
@@ -4258,6 +4258,8 @@ task javadoc(type: Javadoc, dependsOn: createMSPfile) {
 
     options.addStringOption("-since").setValue((9..jfxReleaseMajorVersion.toInteger()).join(","))
     options.addStringOption("-since-label").setValue("New API since JavaFX 9")
+    options.addStringOption("Xmaxwarns").setValue("1000")
+    options.addStringOption("Xmaxerrs").setValue("1000")
```
andy-goryachev-oracle, line 4262: would it make sense to treat warnings as errors?  at least once we fix JBS-TBD to fix all the warnings?

andy-goryachev-oracle:
why I am getting only 91 warnings?

build.gradle
kevinrushforth: Eventually, yes. Which is what I noted in the JBS issue.

Btw, I filed [JDK-8314590](https://bugs.openjdk.org/browse/JDK-8314590) to track fixing all the warnings. A follow-up to _that_ issue will be to enable treating warnings as errors.

andy-goryachev-oracle: Thank you for clarification, Kevin!
I wish github would copy bug description from the ticket to the PR... I ddi not bother to read JBS this time.

andy-goryachev-oracle (approved):
excellent.

build.gradle
kevinrushforth: > I wish github would copy bug description from the ticket to the PR... I ddi not bother to read JBS this time.

I'm glad it doesn't given what ends up in the Descriptions for many bugs. I should have added that to the PR description.

kevinrushforth:
> why I am getting only 91 warnings?
> 
> edit: gradle javadoc on mac, java 18 (may be that's why).

Very likely. Btw, our CI builds (for releases) and GHA builds use JDK 19.0.2.

JDK 19 added more checks (so we get 187 warnings). JDK 21 also added more checks (we're up to 191).

nlisker:
I didn't test it, but I took a look and the fix seems fine.

PR #1219, opened 2023-08-22
8262518: SwingNode.setContent does not close previous content, resulting in memory leak

prsadhuk:
Issue is when setting the content of a SwingNode, the old content is not garbage collected owing to the fact
JLightweightFrame is never being released by SwingNodeDisposer

The SwingNodeDisposer holds an hard pointer to the JLightweightFrame that prevents its collection

Modified `SwingNode.setContentImpl`  function to use a WeakReference to properly release the memory.

---------
### Progress

### Issue
 * [JDK-8262518](https://bugs.openjdk.org/browse/JDK-8262518): SwingNode.setContent does not close previous content, resulting in memory leak (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1219/head:pull/1219` \
`$ git checkout pull/1219`

`$ git checkout pull/1219` \
`$ git pull https://git.openjdk.org/jfx.git pull/1219/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1219`

`$ git pr show -t 1219`

</details>

<a href="https://git.openjdk.org/jfx/pull/1219.diff">https://git.openjdk.org/jfx/pull/1219.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1219#issuecomment-1687884571)

### DIFF
--- a/modules/javafx.swing/src/main/java/com/sun/javafx/embed/swing/Disposer.java
+++ b/modules/javafx.swing/src/main/java/com/sun/javafx/embed/swing/Disposer.java
@@ -89,6 +89,10 @@ public class Disposer implements Runnable {
         return ref;
     }
 
+    /**
+     * Remove the DisposerRecord object to prevent memory leak
+     * @param ref Weak reference of the object to be removed
+     */
     public static void removeRecord(WeakReference ref) {
         disposerInstance.records.remove(ref);
     }
--- a/modules/javafx.swing/src/main/java/com/sun/javafx/embed/swing/Disposer.java
+++ b/modules/javafx.swing/src/main/java/com/sun/javafx/embed/swing/Disposer.java
@@ -90,7 +90,9 @@ public class Disposer implements Runnable {
     }
 
     /**
-     * Remove the DisposerRecord object to prevent memory leak
+     * Unregisters a previously registered {@link DisposerRecord}, removing it
+     * from the list of records to dispose off when the
+     * original target goes out of scope
      * @param ref Weak reference of the object to be removed
      */
     public static void removeRecord(WeakReference ref) {
--- a/modules/javafx.swing/src/main/java/com/sun/javafx/embed/swing/Disposer.java
+++ b/modules/javafx.swing/src/main/java/com/sun/javafx/embed/swing/Disposer.java
@@ -89,8 +89,8 @@ public class Disposer implements Runnable {
         return ref;
     }
 
-    public static void removeRecord(DisposerRecord rec) {
-        disposerInstance.records.remove(rec);
+    public static void removeRecord(WeakReference ref) {
+        disposerInstance.records.remove(ref);
     }
 
     @Override
--- a/modules/javafx.swing/src/main/java/com/sun/javafx/embed/swing/Disposer.java
+++ b/modules/javafx.swing/src/main/java/com/sun/javafx/embed/swing/Disposer.java
@@ -89,6 +89,10 @@ public class Disposer implements Runnable {
         return ref;
     }
 
+    public static void removeRecord(DisposerRecord rec) {
+        disposerInstance.records.remove(rec);
+    }
+
     @Override
     public void run() {
         while (true) {
--- a/modules/javafx.swing/src/main/java/javafx/embed/swing/SwingNode.java
+++ b/modules/javafx.swing/src/main/java/javafx/embed/swing/SwingNode.java
@@ -73,6 +73,7 @@ import static javafx.stage.WindowEvent.WINDOW_HIDDEN;
 import com.sun.javafx.embed.swing.SwingNodeHelper;
 import com.sun.javafx.embed.swing.SwingEvents;
 import com.sun.javafx.embed.swing.newimpl.SwingNodeInteropN;
+import jdk.swing.interop.LightweightFrameWrapper;
 
 /**
  * This class is used to embed a Swing content into a JavaFX application.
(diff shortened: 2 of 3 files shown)


hjohn:
I should note, there is a memory leak still when using `Disposer`.  Each time content is switched, but the `SwingNode` is re-used, a `DisposerRecord` is created that only gets cleaned up when `SwingNode` goes out of scope.  In other words, a million content changes will leave behind a million `DisposerRecord`s.  

So, either there needs to be a `removeRecord` added to `Disposer`, or you could switch to the `TreeShowingProperty` solution.

prsadhuk:
As suggested, optimized without WeakReference

modules/javafx.swing/src/main/java/com/sun/javafx/embed/swing/Disposer.java
```diff
@@ -89,6 +89,10 @@ public static WeakReference addRecord(Object target, DisposerRecord rec) {
         return ref;
     }
 
+    public static void removeRecord(DisposerRecord rec) {
+        disposerInstance.records.remove(rec);
+    }
```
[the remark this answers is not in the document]
prsadhuk: I am not sure if this `removeRecord `is necessary for swing-interop Disposer...Other javafx Disposer like
modules/javafx.base/src/main/java/com/sun/javafx/property/adapter/Disposer.java
also does not have `removeRecord` method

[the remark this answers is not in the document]
hjohn: See this comment: https://github.com/openjdk/jfx/pull/1219#issuecomment-1687998468

modules/javafx.swing/src/main/java/javafx/embed/swing/SwingNode.java
```diff
@@ -361,7 +365,9 @@ public JComponent getContent() {
      */
     private void setContentImpl(JComponent content) {
         if (lwFrame != null) {
-            swNodeIOP.disposeFrame(lwFrame);
+            rec.dispose();
```
andy-goryachev-oracle, line 368: is it possible for `rec` to be null at this point?
would it make sense to rewrite this code block as
```
if(rec != null) {
  rec.dispose();
  rec = null;
}
if(disposeRecRef != null) {
  Disposer.removeRecord(...)
  disposerRecRef = null;
}
if(lwFrame != null) {
 lsFrame = null;
}
```
?

modules/javafx.swing/src/main/java/com/sun/javafx/embed/swing/Disposer.java
```diff
@@ -89,6 +89,10 @@ public static WeakReference addRecord(Object target, DisposerRecord rec) {
         return ref;
     }
 
+    public static void removeRecord(WeakReference ref) {
```
andy-goryachev-oracle, line 92: may be add a javadoc comment?

modules/javafx.swing/src/main/java/javafx/embed/swing/SwingNode.java
prsadhuk: If `lwFrame` is not null, little chance of `rec` to be null at this point as unlike PR#1178 there's no multiple threads involved, only EDT, so no race condition causing class variables to null out, 
but anyways I have added null check to be safe..

andy-goryachev-oracle (approved):
Tested with the code example in the ticker.  Seems to work, the "Panels in memory" stays around 2, occasionally goes to 30-40 but then comes down to 2 pretty quickly.  Same behavior even when `Thread.sleep(1);`

Questions:
1. the JBS ticket refers to 3 other tickets - are we sure these are exact duplicates?
2. the JBS talks about 1.8.0.271 - do we need to backport this fix to java8?

modules/javafx.swing/src/main/java/com/sun/javafx/embed/swing/Disposer.java
```diff
@@ -89,6 +89,10 @@ public static WeakReference addRecord(Object target, DisposerRecord rec) {
         return ref;
     }
 
+    /**
+     * Remove the DisposerRecord object to prevent memory leak
+     * @param ref Weak reference of the object to be removed
+     */
```
hjohn, line 95: Looks good, however I think you could say this a bit more general, no need to mention that it is to prevent memory leaks -- it's just normal that you should remove the record if there is no need for that specific dispose task anymore because it was already disposed in another way.

So how about:

Unregisters a previously registered {@link DisposerRecord}, removing it from the list of records to dispose off when the original target goes out of scope. This is useful in cases where the original target is still in use, but the record was already disposed off by other means.

modules/javafx.swing/src/main/java/javafx/embed/swing/SwingNode.java
```diff
@@ -365,12 +364,17 @@ public JComponent getContent() {
...
+                rec.dispose();
+                rec = null;
+	    }
             lwFrame = null;
         }
         if (content != null) {
+            if (disposerRecRef != null) {
+                Disposer.removeRecord(disposerRecRef);
+                disposerRecRef = null;
+            }
```
hjohn, line 377: I think this should be part of the first `if`, or an `if` by itself, and not here.  `lwFrame`, `rec` and `disposerRecRef` will all either be `null` or be set to some value.  The only case this may not happen is that after `lwFrame` is set, an exception occurs.  You could make it very safe, or just assume that when `lwFrame` is not `null` that all clean up actions should be taken -- if an exception did occur during set-up, then it's also fine that one can occur during clean-up, no need to make it more robust for situations that can't happen.

There really is no need to start second guessing yourself when the logic is solid.  Either all those variables are non-`null` and checking one of them is sufficient, or they are all `null` and checking each in turn is pointless (it is just confusing because programmers may assume now that is a situation that can actually occur).

Please revert to the old code.

prsadhuk: @hjohn I guess i have updated as per suggestion..Let me know if this is ok as per you..if yes, please approve..

hjohn (approved):

andy-goryachev-oracle (approved):

kevinrushforth:
Is it possible to write an automated test?

prsadhuk:
> Is it possible to write an automated test?

oops..missed this..i can definitely try..maybe in upcoming test-sprint..

kevinrushforth:
Can you file a follow-on test bug?

prsadhuk:
Yes, JDK-8315317

PR #1220, opened 2023-08-22
8296266: TextArea: Navigation breaks with RTL text

andy-goryachev-oracle:
The fix uses character BreakIterator instead of the logic that relies on caretBounds/hitTest/rangeShape in TextInputControl.nextCharacterVisually().

I believe this is a more reliable method of navigation, as it behaves in sync with the jdk break iterator, thought it might work differently around grapheme clusters, considering a recent change JDK-8291660

This change also introduces TextInputControlHelper class (impl. detail) which gives access to character- and word- break iterators cached by TextInputControl (*some say* these iterators and associated editing logic should be a part of Content implementation, but that's a discussion for another day).

---------
### Progress

### Issue
 * [JDK-8296266](https://bugs.openjdk.org/browse/JDK-8296266): TextArea: Navigation breaks with RTL text (**Bug** - P3)

### Reviewers
 * [Karthik P K](https://openjdk.org/census#kpk) (@karthikpandelu - Committer) 🔄 Re-review required (review applies to [e41da6df](https://git.openjdk.org/jfx/pull/1220/files/e41da6dfb82e8dd2379ecbfd4cc100068bffe92e))
 * [Ajit Ghaisas](https://openjdk.org/census#aghaisas) (@aghaisas - **Reviewer**) 🔄 Re-review required (review applies to [e41da6df](https://git.openjdk.org/jfx/pull/1220/files/e41da6dfb82e8dd2379ecbfd4cc100068bffe92e))

### Contributors
 * Karthik P K `<kpk@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1220/head:pull/1220` \
`$ git checkout pull/1220`

`$ git checkout pull/1220` \
`$ git pull https://git.openjdk.org/jfx.git pull/1220/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1220`

`$ git pr show -t 1220`

</details>

<a href="https://git.openjdk.org/jfx/pull/1220.diff">https://git.openjdk.org/jfx/pull/1220.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1220#issuecomment-1693973615)


karthikpandelu (commented):
When the shortcut: cmd + right arrow is used to traverse the text area, caret is displayed on the both sides of the RTL text  as shown in screenshot below. This is not observed if individual character is traversed using right arrow or shortcut option + right arrow is used to traverse word by word.
image
Rest of the changes look good to me.

andy-goryachev-oracle:
Thank you for testing, @karthikpandelu !

> This is not observed if individual character is traversed using right arrow or shortcut option + right arrow...

This statement is not exactly correct.  Cmd+right navigates to a (visible) line end, in this case it's in the middle of the text string (index=8).  So if you place the cursor after the colon (index=7) and press right arrow, you'll get the split caret in exact the same configuration as cmd+right.

karthikpandelu:
> This statement is not exactly correct. Cmd+right navigates to a (visible) line end, in this case it's in the middle of the text string (index=8). So if you place the cursor after the colon (index=7) and press right arrow, you'll get the split caret in exact the same configuration as cmd+right.

Thanks for the details.

andy-goryachev-oracle:
@aghaisas would you take a look please?

aghaisas (approved):
The fix seems logical.

karthikpandelu (approved):
LGTM

andy-goryachev-oracle:
@prrace can you please take a look?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextAreaSkin.java
```diff
@@ -571,38 +572,11 @@ private void nextCharacterVisually(boolean moveRight) {
             moveRight = !moveRight;
```
prrace, line 572: This method confuses me as to what it actually wants to do and why
As in, "move right, no, your other right" (!)

isRTL() checks the NodeOrientation so it seems in such a case "right" means "left".
Have you checked the fix with both orientations ?
Seems to me it reverses the meaning of the arrow keys without regard to the text.

And he BreakIterator is getting the next logical position, but this method claims to move to the next visual position.
The method is called nextCharacterVISUALLY, but with this code its actually moving LOGICALLY,
so I'm not seeing how this works for mixed directional text for either setting of "isRTL()" which seems to me to be very questionable to even check here. IMO the arrow key should work the same way no matter whether the UI is laid out with nodes from RTL or LTR.

Perhaps the word "Visually" was not the best choice of name here ? And "right" is questionable too. But I'm not 100% sure.
if you really do want LOGICALLY, then this code will probably work, except I find the isRTL() questionable even then.

What exactly was wrong in the existing code ? Other than it being very complex. How did its calculations go wrong ?

modules/javafx.controls/src/main/java/javafx/scene/control/TextInputControl.java
```diff
@@ -745,6 +754,32 @@ public void selectForward() {
     private BreakIterator charIterator;
     private BreakIterator wordIterator;
 
+    /**
+     * Returns a cached instance of character break iterator, creating it if necessary.
+     * The instance is initialized with the given text.
```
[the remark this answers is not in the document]
andy-goryachev-oracle: +1 thanks!

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextAreaSkin.java
prrace: PS if I just use simple ASCII text and set Node Orientation to RTL, then with your fix
then left arrow moves right and right arrow moves left.
In JDK 8 left arrow moved left and right arrow moved right.
With an FX 17 build I have around, then the nothing moves.
So it looks like things might have been better when this code was written and some other change broke it.
Your fix unbreaks it, but I'm not entirely sure its restoring intended behaviour.

andy-goryachev-oracle: what if you have a mixture of RTL and LTR text in JDK8?
I am going to double check the behavior in JTextArea as well.

andy-goryachev-oracle: what is going on??  JTextArea inside a Scene with RTL orientation:

Screenshot 2023-10-09 at 15 23 51

prrace: JDK 8u doesn't have the fixes to deal with all of the macOS fonts in current macOS, so its a little hard for me to test Arabic/bidi there since I get garbage text.

I don't see that mirroring that you see, neither setting it on the TextArea or on the Scene, but the text is way off to the right.
Some layout issue.

andy-goryachev-oracle: ^^ this looks like a separate bug, as it merely mirror images the whole SwingNode or its content.
Setting ComponentOrientation produces desired result, so perhaps we'd need to bind SwingNode's content component orientation to its node orientation.

But anyway, within JTextArea, the RIGHT arrow *always* moves to the right, and the LEFT arrow always moves to the left (of screen), regardless of component orientation which I think might be the "objectively correct" behavior.

Curiously, MS Word does not work this way, instead moving cursor *logically*.

The fixed code works similarly to what MS Word does, reversing the apparent direction within RTL/LTR blocks:
- In the LTR orientation, left moves the cursor left and right - to right within LTR text, which is what we want.
- In the RTL orientation, left moves left and right moves right within RTL text, which is also what we want.

I suppose the decision to work similarly to MS Word (and differently from swing?) was made early on and we are stuck with it.

Sooo, I guess the fix works as designed?
(I agree with your other objections to naming and javadoc, will fix).

andy-goryachev-oracle: > JDK 8u doesn't have the fixes to deal with all of the macOS fonts in current macOS, so its a little hard for me to test Arabic/bidi there since I get garbage text.

if I remember correctly, explicitly setting an Arabic font removes the garbage.

> I don't see that mirroring that you see, neither setting it on the TextArea or on the Scene, but the text is way off to the right.

I've added this to the MonkeyTester (standalone revision)
Tools -> JTextArea 
and
Window -> Orientation: RTL

andy-goryachev-oracle:
The solution needs re-evaluation due to unexpected complexity dealing with bidi text - specifically with navigation (visual vs. logical) and behavior at the bidi boundary.

andy-goryachev-oracle:
@nlisker would you be able to check out the new logical navigation in this PR?

There is also a similar change in navigation in TextField/PasswordField https://github.com/openjdk/jfx/pull/1222 which depends on this PR.

nlisker:
I tested a `TextArea`. Here are my observations:

1. There is no way to change the primary direction / justification, which on Windows is done with CTRL+SHIFT. This is a must have for RTL support.
2. The caret doesn't indicate if I'm in an RTL or LTR mode. This is from Notepad, showing the direction:
image
image
3. The split caret is quite interesting, it's the first time I am experiencing it. There is still the problem that I don't know in which place the text will appear unless I look at which language I have selected, then remember which of the carets is primary. It also doesn't seem to solve the ambiguity:
    1. Start with this text and the caret at position 0 (in English LTR mode): `ds גדsaששגד sds`
image
    2. Move 3 times to the right (with the right arrow key). I get this split caret: 
image
    3. Pressing `a` adds the character at the lower caret position:
image
    4. Instead of adding `a`, I can move the caret another 2 spaces to the right and have the same caret display as in `ii`.
    5. Now pressing `a` adds the character at the upper caret position:
image

    This means I can't know where the character will be added. I think that the split caret is going to confuse a lot of people. Do you know anyone who has ever seen this type of caret?

andy-goryachev-oracle:
> 1. There is no way to change the primary direction / justification, which on Windows is done with CTRL+SHIFT. This is a must have for RTL support.

What is your hotkey setting for switching languages?  On my win11, it's ctrl-shift, so in WordPad ctrl-shift seems to control both input language and alignment, looking almost as if both functions are triggered at the same time.

In the case of JavaFX, there is no function to change primary direction with a key stroke.

andy-goryachev-oracle:
> 2\. The caret doesn't indicate if I'm in an RTL or LTR mode.

This is how it works in jfx8.

But here we might have some freedom as we could change the way dual caret looks.  The directionality indicator seems like a good idea (try crossing the LTR/RTL boundary in swing - it takes two key presses during which the indicator changes.  A single caret, even with the directionality indicator does not solve the insertion point question though.  May be both carets in the case of dual caret should have directionality indicator?

andy-goryachev-oracle:
> 3\. This means I can't know where the character will be added. I think that the split caret is going to confuse a lot of people. Do you know anyone who has ever seen this type of caret?

I've only seen the split caret in FX.

It seems like an attempt to provide an indication of insertion point, but the way it works is confusing: what is definition of the primary and the secondary?

For example, with the TextArea set to RTL orientation, and text set to something like  **כעיגכחשENGLISHשגלכחעיגל**, if the user clicks on the left side of "E" we'll have this:

Screenshot 2023-10-19 at 11 23 57

So the primary caret is to the left side of "E" visually.  Then, type a Hebrew symbol and it gets inserted into position indicated by the secondary caret, unexpectedly (isn't RTL a primary in this case?)

At the same time, if we switch TextArea node orientation back to LTR, it works as expected - latin symbols get inserted where the primary caret points, Hebrew symbols go the secondary position.

nlisker:
> What is your hotkey setting for switching languages? On my win11, it's ctrl-shift, so in WordPad ctrl-shift seems to control both input language and alignment, looking almost as if both functions are triggered at the same time.

ctrl+shift only changes the primary direction / justification. alt+shift changes the input language. Maybe in Win 11 this was changed, but prior to that, at least since the days of XP, those were the shortcuts.

> A single caret, even with the directionality indicator does not solve the insertion point question though.

But It can. The "jumping" caret in the document you linked, which is also what I suggested prior, does that. It provides unambiguous insertion points.

> I've only seen the split caret in FX.

Which is why I think we should not be pushing this indicator. People are used to a single caret that shows them where the next character they type will appear. I think that's what we should give them.

andy-goryachev-oracle:
jfx8 already has split caret, so I am not so sure about removing it.

The jumping caret is a rather strange solution, in my opinion (user: do I need to time my key presses to the moment when the caret jumps to the right place?)  I am not a LTR user, so my opinion here carries little weight, but I thought addition of directionality indicators solves the problem and lowers confusion slightly - the user still needs to maintain some mental state with regards to editing - backspace / delete.

FYI, backspace and delete are not symmetrical - while delete is expected to delete the whole symbol (which might consist of multiple chars), backspace deletes the previous char, breaking up clusters (Note: FX does not currently support complex grapheme clusters like flags or skin tones).

nlisker:
Well, I don't have any experience with split carets whatsoever, nor any of the people I've asked, so I can't advise on it. If the user doesn't need to look outside of the text control to see which language is selected, and can know unambiguously where the character will appear, it's viable. Then again, if the primary caret tells me where my text will appear, why do I need the secondary caret?

andy-goryachev-oracle:
> Then again, if the primary caret tells me where my text will appear, why do I need the secondary caret?

Because where the next character will appear depends on the character.  LTR chars appear in one spot, RTL in the other.
I believe this is an issue even with apps that are purely in one language - because of numbers.

ex: means “Reshet 13”
רשת 13

nlisker:
But from what I understand, the primary and secondary carets change. They are not an LTR and RTL carets.

Perhaps you can update this branch with the implementation you have in mind and I will check it.

andy-goryachev-oracle:
> Perhaps you can update this branch with the implementation you have in mind and I will check it.

Quick and dirty prototype of the pennant-like direction indicator suggested by @prrace  (thank you!)
https://github.com/andy-goryachev-oracle/jfx/pull/5
 
Screenshot 2023-10-19 at 15 22 03

 
You can try this using the MonkeyTester TextArea page because it has Window -> RTL menu option for switching the orientation.

nlisker:
> > Perhaps you can update this branch with the implementation you have in mind and I will check it.
> 
> Quick and dirty prototype of the pennant-like direction indicator suggested by @prrace (thank you!) [andy-goryachev-oracle#5](https://github.com/andy-goryachev-oracle/jfx/pull/5)
> 
> Screenshot 2023-10-19 at 15 22 03
> 
> You can try this using the MonkeyTester TextArea page because it has Window -> RTL menu option for switching the orientation. https://github.com/andy-goryachev-oracle/MonkeyTest

This implementation with the dual caret direction indicator is better, but the problems still remain:

1. Visual: The caret, single or split, does not indicate the current input direction (point 2 in https://github.com/openjdk/jfx/pull/1220#issuecomment-1770459622). This is very necessary. This is also why I mentioned that a split caret can be rendered as a single caret whose location depends on the input direction; no need to show a split caret showing me a location where I'm *not* going to enter characters because the input direction doesn't match.
2. Functional: The input is either being done in the wrong place, or the indicator is off.
    1. Start with the caret at this position: 
image
    2. Enter a character in LTR, the result makes sense: 
image
    3. Enter a character in RTL, "how did it get there?": 
image
    and why did the space separating the RTL and LTR texts was transported to somewhere else?
3. If the text starts with RTL:
image
    the split caret doesn't show correctly, and inserting an LTR character adds it somewhere else, far away from the caret:
image

For JavaFX to match what every modern application does (at least on Windows), the following need to be guaranteed:
1. Keyboard shortcut for changing the text orientation (CTRL+SHIFT on Windows).
2. Direction indicator in the single caret. Requiring the user to look elsewhere to check what the input method is is detrimental (and I would go as far as to say unacceptable).

Even the super-simple Notepad does those.

andy-goryachev-oracle:
@nlisker:

Thank you!

1. when you say "The caret, single or split, does not indicate the current input direction", are you referring to this PR or the branch with the direction indicator?  Because there we do have an indicator when the caret is split.  When it is not split, why do you need an indicator? 

I see for example, MS Word does show an indicator but **only** when the RTL input is active:
 
Screenshot 2023-10-24 at 08 22 44

is that the behavior you want, to indicate when RTL input is on?  If so, we have a problem - there is no current way for FX to know, so we might need to introduce a new API.  Another alternative is to rely on NodeOrientation.

2. You are right.  There is something wrong, the split caret is not displayed when expected.

3. I don't know if it's right to handle CTRL - SHIFT: it's an OS function that switches input languages, at least on my win11.  Perhaps we could use a different key stroke, but let me ask this question: what exactly do you mean "text orientation"?  NodeOrientation?  Inserting one of these characters - https://www.w3.org/International/questions/qa-bidi-unicode-controls#basedirection ?

nlisker:
> I see for example, MS Word does show an indicator but **only** when the RTL input is active:
> 
> Screenshot 2023-10-24 at 08 22 44
> is that the behavior you want, to indicate when RTL input is on? If so, we have a problem - there is no current way for FX to know, so we might need to introduce a new API. Another alternative is to rely on NodeOrientation.

Yes, this is what is expected, as I have shown [in point 2](https://github.com/openjdk/jfx/pull/1220#issuecomment-1770459622). The caret needs to show the input direction. The caret shape needs to be bound to the input direction.

> 3. I don't know if it's right to handle CTRL - SHIFT: it's an OS function that switches input languages, at least on my win11.  Perhaps we could use a different key stroke, but let me ask this question: what exactly do you mean "text orientation"?  NodeOrientation?  Inserting one of these characters - https://www.w3.org/International/questions/qa-bidi-unicode-controls#basedirection ?

I don't know how other OS's handle it. Here is Notepad demonstrating the correct behavior:
LTR orientation: image
RTL orientation: image

Here is WhatsApp Desktop:
LTR: image
RTL: image

This is how MS Word shows the orientation change:
LTR: image
RTL: image
You can see the Text Direction and the Alignment properties change.

Here is me writing this comment in the Vivaldi browser:
LTR: image
RTL: image

The scrollbar position also changes.

andy-goryachev-oracle:
Thank you @nlisker for detailed explanation.  But I am confused: I see you indicate that the orientation of the control (watsapp, notepad) changes, but if you look at MS Word 2208 on Windows (that's the version I have on my Win11 system) I see that ctrl-shift changes the orientation of the current paragraph.  When you say that the side of the scrollbar changes, it means the whole control changes its orientation.  So which one do you want, control or paragraph?

And, in the context of TextArea we have no way to control paragraph attributes, but it's definitely a required function for a rich text control.

As for split caret, I just created JDK-8318840 .  But, generally speaking, what do you think of having split caret with direction indicators as in https://github.com/andy-goryachev-oracle/jfx/pull/5 ?  Do you think it's a good solution?

Also, perhaps we should show a directional indicator any time we have bidi text in the paragraph, not only when the node orientation is RTL, would you agree?

nlisker:
> Thank you @nlisker for detailed explanation. But I am confused: I see you indicate that the orientation of the control (watsapp, notepad) changes, but if you look at MS Word 2208 on Windows (that's the version I have on my Win11 system) I see that ctrl-shift changes the orientation of the current paragraph. When you say that the side of the scrollbar changes, it means the whole control changes its orientation. So which one do you want, control or paragraph?
> 
> And, in the context of TextArea we have no way to control paragraph attributes, but it's definitely a required function for a rich text control.

Whether the it's the control or paragraph that changes seems to be dependent on the control type. I don't think there's any user expectation there. What is expected is that the text will be aligned accordingly and that the primary direction will match.

> As for split caret, I just created JDK-8318840 . But, generally speaking, what do you think of having split caret with direction indicators as in [andy-goryachev-oracle#5](https://github.com/andy-goryachev-oracle/jfx/pull/5) ? Do you think it's a good solution?

This was point 1 in https://github.com/openjdk/jfx/pull/1220#issuecomment-1777050203. The split caret doesn't tell me where the next character will appear without checking elsewhere what the input language is, and that's not good. The split caret always has, by definition, one irrelevant (half) caret. If I'm in RTL insertion mode, I don't care about the LTR caret, it's confusing to show it.

This is why I suggested to render the split caret in a way that only shows the correct half of the caret based on the input method.

> Also, perhaps we should show a directional indicator any time we have bidi text in the paragraph, not only when the node orientation is RTL, would you agree?

A directional indicator should show whenever an RTL input method is selected, regardless of paragraph or node orientation, or what text is in the paragraph. Check Notepad, for example, for the correct behavior here.

nlisker:
I will clarify the guiding principle: what users expect is an unambiguous insertion point by looking only at the text control (the caret and the orientation/justification). When I press a key, I'm supposed to know with certainty what character will appear and where it will appear. This is why we need to display only 1 caret with a direction indicator of the current language (even if the underlying mechanism calculates 2 carets). Anything else creates ambiguity and the all-too-common "I'm typing in the wrong place/language".

andy-goryachev-oracle:
> This is why we need to display only 1 caret with a direction indicator of the current language (even if the underlying mechanism calculates 2 carets)

But the insertion point depends on the key being typed.  
Type: 
aleph aleph 1 2 3

you see the dual caret (in Notes.app in macOS), indicating that, depending on the character that will be typed next, the insertion point may be to the left or to the right of the digits:

Screenshot 2023-10-26 at 08 19 16

type a digit 4 and it will be inserted to the right of 3
type aleph and it will be inserted to the left of 1:

Screenshot 2023-10-26 at 08 23 36

How can you get by with just one caret?  That seems to violate the principle "unambiguous insertion point by looking only at the text control".  I mean, it is always ambiguous  at the bidi boundary, thus the two carets, right?

nlisker:
Numbers are unambiguous because they can't represent any other character, at least not in any case I know of (Hebrew and Arabic). Arabic numbers are also LTR while the language it RTL, and even there there're no surprises as long as the caret tells you what direction you are typing in.

> type a digit 4 and it will be inserted to the right of 3
type aleph and it will be inserted to the left of 1:

But the `aleph` character (`א`) can also be the `t` character, depending on your input language, and the dual caret shown doesn't indicate which of the 2 are going to be typed. The Notepad/Wordpad/Word/WhatsApp... carets on Windows show which character is going to be typed, hence its advantage.

> it is always ambiguous at the bidi boundary, thus the two carets, right?

Which is why a jumping caret is a good solution. Depending on what your input is set to, you know what to expect. With numbers you always know where they are going to be typed because they are always LTR regardless of the language (maybe some languages have RTL numbers, but I'm not aware of those).

andy-goryachev-oracle:
> The Notepad/Wordpad/Word/WhatsApp... carets on Windows show which character is going to be typed, hence its advantage.

I see.  So we basically need some a boolean property in Platform (?) to track input language orientation.

nlisker:
Something like that, yes. I'm not sure if it should be boolean because some languages are typed vertically (e.g., https://en.wikipedia.org/wiki/Horizontal_and_vertical_writing_in_East_Asian_scripts). Far be it from me to advise on what to do with these. Maybe an enum is better here?

Also, I don't think that the input language should be tracked in a public API. It should remain an implementation detail, unless I'm missing something in the realm of creating new controls.

By the way, the more I think about it, the *split caret with an indicator on which of the carets is the one for the current language* looks more reasonable. Both the "single jumping caret" and this "asymmetric split caret" tell the user where the next letter is going to be typed. The "asymmetric split caret" has the advantage of helping a bit with numbers in RTL languages, but at the cost of some confusion. Both this caret and the single caret (used in Notepad as shown above) require to know what the input language direction is anyway, and it's mostly a matter of how to represent that.

andy-goryachev-oracle:
Right, ideally there would be some kind of "InputLanguage" object with text direction and name.  You are right about vertical text, though FX is unlikely to support it in the foreseeable future.

I think this Platform property may start as an impl. detail initially, but it should be public eventually, for the sake of anyone who writes their own text input control.

The asymmetric split caret is a very interesting idea, thank you!

andy-goryachev-oracle:
This depends on [JDK-8319050](https://bugs.openjdk.org/browse/JDK-8319050) to be fixed first.

PR #1225, opened 2023-08-29
8315217 JavaFX: Add Printer.printableArea(Paper) method

FlorianKirmaier:
We sometimes want to access the printable area for a given paper.
The printable area is currently defined in com.sun.javafx.print.PrinterImpl and we would like to make it accessible to javafx.print.Printer.
We would also return the value in points instead of inches for a more consistent API.

RFE:
I have to create an RFE for it, correct?
Any point of reference, on how to create one?

---------
### Progress
- [ ] Change requires CSR request [JDK-8315376](https://bugs.openjdk.org/browse/JDK-8315376) to be approved

### Issues
 * [JDK-8315217](https://bugs.openjdk.org/browse/JDK-8315217): JavaFX: Add Printer.printableArea(Paper) method (**Enhancement** - P4)
 * [JDK-8315376](https://bugs.openjdk.org/browse/JDK-8315376): JavaFX: Add Printer.printableArea(Paper) method (**CSR**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1225/head:pull/1225` \
`$ git checkout pull/1225`

`$ git checkout pull/1225` \
`$ git pull https://git.openjdk.org/jfx.git pull/1225/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1225`

`$ git pr show -t 1225`

</details>

<a href="https://git.openjdk.org/jfx/pull/1225.diff">https://git.openjdk.org/jfx/pull/1225.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1225#issuecomment-1697068562)

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/print/Printer.java
+++ b/modules/javafx.graphics/src/main/java/javafx/print/Printer.java
@@ -231,6 +231,20 @@ public final class Printer {
     }
 
     /**
+     * Returns the printable area in points for a given paper.
+     * @param paper The paper to use.
+     * @return Returns the printable area in points
+     */
+    public Rectangle2D printableArea(Paper paper) {
+        Rectangle2D imgArea = impl.printableArea(paper);
+        return new Rectangle2D(
+                imgArea.getMinX() * 72,
+                imgArea.getMinY() * 72,
+                imgArea.getWidth() * 72,
+                imgArea.getHeight() * 72);
+    }
+
+    /**
      * Obtain a new PageLayout instance for this printer using the specified
      * parameters.
      * The paper should be one of the supported papers and


kevinrushforth:
Two things:

1. As indicated in the [New features / API additions section](https://github.com/openjdk/jfx/blob/master/CONTRIBUTING.md#new-features--api-additions) of the CONTRIBUTING guidelines, new API needs prior socialization on the openjfx-dev mailing list to see if there is general agreement that such a feature would be a benefit to applications. In the mean time, I am moving this PR to Draft.
2. An enhancement request titled "Make XXXX public" is rarely the right way to describe a new public API. The fact that there happens to be a non-public method that does something that applications might want to control is an implementation detail. From the viewpoint of the public API, there is no such method called XXXX. Instead, the title should describe the new method or feature that is missing and that you want to implement. It may seem like a small point, but it's an important distinction. To that end, please change the title of the JBS issue and this PR to something like "Add Printer.printableArea(Paper) method".

To answer the question you raised in the Description of this PR:

> I have to create an RFE for it, correct?

RFE stands for "Request For Enhancement". The JBS issue of type "Enhancement" is the RFE, so you have already created it. This Enhancement request will need a CSR as described in the above-referenced section of the CONTRIBUTING guidelines.

/reviewers 2
/csr

FlorianKirmaier:
@kevinrushforth 
Thank you for your quick response!

**Mailing List:**
Is it sufficient that every comment here is sent to the mailing list, or should I write a separate email outside of GitHub?

Nowadays the RFR is also automatically generated, which is really a great improvement!

If we would also automatically create a ticket for every PR - then the development experience would be identical to any other GitHub project!

**CSR:**
I've created a CSR for the ticket: https://bugs.openjdk.org/browse/JDK-8315376

**Some more explanation:**
We had to write a custom print dialog - for some advanced requirements.
But in some cases, we needed more information which is available with the current API.
The printable area is important because some printers can't print the whole area of the page, in this helps us to know what can be done.

kevinrushforth:
> **Mailing List:** Is it sufficient that every comment here is sent to the mailing list, or should I write a separate email outside of GitHub?

It needs to be a separate email discussion thread. The primary reason for having prior discussion before the review is to see if this is a problem that should be solved, and if there is support for adding it. A PR is not the starting point for that discussion, because it presupposes that the feature in question is generally useful enough that we want to add public API that will need to be tested, maintained, etc. It can be tempting to use a "code first" approach for simple features like this, but I encourage you to resist that urge.

> If we would also automatically create a ticket for every PR - then the development experience would be identical to any other GitHub project!

Even if this were technically feasible, we will not implement this. I know this is how many smaller GitHub projects work. It is not how the OpenJDK community works.

> **CSR:** I've created a CSR for the ticket: https://bugs.openjdk.org/browse/JDK-8315376

OK. The CSR should remain in Draft until there is general agreement about the new API.

PR #1226, opened 2023-08-29
8313556: Create implementation of NSAccessibilitySlider protocol

azuev-java:
Create implementation for Slider and Stepper accessibility protocols.
Fix mapping.
Fix performAction parameter type in declaration.

---------
### Progress

### Issues
 * [JDK-8313556](https://bugs.openjdk.org/browse/JDK-8313556): Create implementation of NSAccessibilitySlider protocol (**Enhancement** - P4)
 * [JDK-8313558](https://bugs.openjdk.org/browse/JDK-8313558): Create implementation of NSAccessibilityStepper protocol (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1226/head:pull/1226` \
`$ git checkout pull/1226`

`$ git checkout pull/1226` \
`$ git pull https://git.openjdk.org/jfx.git pull/1226/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1226`

`$ git pr show -t 1226`

</details>

<a href="https://git.openjdk.org/jfx/pull/1226.diff">https://git.openjdk.org/jfx/pull/1226.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1226#issuecomment-1697159747)


andy-goryachev-oracle:
should there be an automated test for the added functionality?

kevinrushforth:
> should there be an automated test for the added functionality?

Automated testing of the screen reader is not feasible. Even if it were, it would be a separate enhancement to add accessibility tests. Btw, this PR isn't providing new functionality. It is providing a new implementation of existing functionality.

Having said that, it would be helpful for Alex to list the controls that will use the new implementation, so reviewers can test it manually. For this PR, I would imagine that the `Slider` and `Spinner` controls would be used to test the `NSAccessibilitySlider` and `NSAccessibilityStepper` protocols, respectively. Perhaps there are also other ways to hit the new code?

kevinrushforth:
Yes, you read it correctly. Good question about how you can tell that it is going through the new implementation. Since it is supposed to transparently do the same thing, the only way I can think of is to instrument the code to add debug logging statement (which is what I did when reviewing the initial changes).

andy-goryachev-oracle:
perhaps the logging statements should be added (commented out) as a part of this PR.

azuev-java:
> perhaps the logging statements should be added (commented out) as a part of this PR.

I can do that - i usually refrained from doing so because code would look not very clean but basically the only way to tell that it goes trough the new code should be debug output - if behavior changes it should be considered as a potential bug (not always because if i see that the old behavior is incorrect i would not go out of my way to recreate the obviously flawed logic) but in general the idea is that new code should behave exactly the same as the old one.

andy-goryachev-oracle:
what I meant is that commented out logging statements should be left in the code, so the reviewers can uncomment them and test.  like this:

```
// System.err.println("it works");
```

azuev-java:
> Having said that, it would be helpful for Alex to list the controls that will use the new implementation, so reviewers can test it manually.

I will definitely do that in the future. In this case you are tight the controls are Spinner and Slider and while we have a dedicated control group for Spinner component in Ensemble there is no such thing for Slider (or i haven't found it) so the best way to test it is to go to the ColorPicker component test, invoke color chooser and click "Custom Color..." link - that will invoke the color selection dialog that has a lot of sliders to play with.

azuev-java:
> what I meant is that commented out logging statements should be left in the code, so the reviewers can uncomment them and test. like this:
> 
> ```
> // System.err.println("it works");
> ```

Did that. I would just appreciate if you or any other reviewer reminds me to remove these statements after we are done with the review :) The code should print out when slider or stepper values are being changed using the a11y shortcuts to manipulate controls. That will ensure that when being accessed trough the a11y handle the component state can be appropriately updated. And of course it will indicate that the new component is invoked.

arapte:
I observed few warnings and behavioral differences.
- Warnings:
1. modules/javafx.graphics/src/main/native-glass/mac/a11y/JFXSliderAccessibility.m:32:17: warning: method 'accessibilityFrame' in protocol 'NSAccessibilityElement' not implemented [-Wprotocol]
2. modules/javafx.graphics/src/main/native-glass/mac/a11y/JFXSliderAccessibility.m:32:17: warning: method 'accessibilityParent' in protocol 'NSAccessibilityElement' not implemented [-Wprotocol]
3. modules/javafx.graphics/src/main/native-glass/mac/a11y/JFXStepperAccessibility.m:32:17: warning: method 'accessibilityFrame' in protocol 'NSAccessibilityElement' not implemented [-Wprotocol]
4. modules/javafx.graphics/src/main/native-glass/mac/a11y/JFXStepperAccessibility.m:32:17: warning: method 'accessibilityParent' in protocol 'NSAccessibilityElement' not implemented [-Wprotocol]
5. modules/javafx.graphics/src/main/native-glass/mac/a11y/AccessibleBase.m:247:21: warning: 'NSObject' may not respond to 'clearParent'

- Behavior differences: 
1. Spinner: Changed values are not read by VoiceOver when we increase or decrease the value using arrow keys or  using Ctrl + Option + Up/Down keys ( Tested with Spinner app in Ensemble )
2. Slider: Values are read without % as suffix ( Tested with Styled tool bar app in Ensemble )

andy-goryachev-oracle:
Testing with macOS 13.3.1a and Voice Over, using MonkeyTester 

Observed a number of things, some possibly not related to this PR:

1. had color picker custom color dialog opened when turning on VoiceOver, saw stderr output:
*** java.lang.instrument ASSERTION FAILED ***: "!errorOutstanding" with message transform method call failed at open/src/java.instrument/share/native/libinstrument/JPLISAgent.c line: 873

2. color picker, custom color: hue slider is not accessible

3. color picker, slider (opacity):
- shows 46 (%), announces: 46.3
- no NSLog debug output if using left/right arrow
- when entering slider via shift-ctrl-option-down arrow, does not announce how to control the slider (as it did with spinner).
- when controlling the spinner with ctrl-option-up/down, I see NSLog debug output:

2023-08-31 08:09:03.087 java[45918:3234828] Slider decrement action invoked
2023-08-31 08:09:04.565 java[45918:3234828] Slider increment action invoked

4. spinner: see debug output when using ctrl-option-up/down, but change in value is not announced
2023-08-31 08:05:12.267 java[45918:3234828] Stepper decrement action invoked
2023-08-31 08:05:13.192 java[45918:3234828] Stepper increment action invoked

5. spinner: when in spinner (via shift-ctrl-option-down), trying to control the spinner via ctrl-option-up/down, releasing the last button moves the focus to the menu instead of staying in the spinner.

azuev-java:
@arapte @andy-goryachev-oracle Ping. Can i ask for the review continuation?

andy-goryachev-oracle:
will do

arapte:
With this change slider is read differently, slider's name is not read
1. Open Ensemble, run Stacked Area Chart sample
2. Click on lowerBound slider, it will be read as:
 - Without fix: 55%, lowerBound, slider 
 - With fix: 55%, slider
 
Spinner is read same as before.
But with, #1291 , the value of spinner is not read
1. Value is not read when initially spinner gets focus
2. Value is not read when changed

I think, I will have to rework #1291 after this is integrated.

arapte:
>        But with, [8284544: [Win] Name-Property of Spinner cannot be changed #1291](https://github.com/openjdk/jfx/pull/1291) , the value of spinner is not read

This is corrected in #1291, Please ignore above comment.
Above is not a side effect of this PR, but it was an issue with #1291
I have corrected the same and Spinner gets read same as before, with #1291  and with both #1291 & this #1226

modules/javafx.graphics/src/main/native-glass/mac/a11y/AccessibleBase.m
```diff
@@ -129,6 +153,30 @@ - (id)accessibilityParent
...
+- (id)accessibilityTitle
+{
+    jobject jresult = NULL;
+    GET_MAIN_JENV;
+    if (env == NULL) return NULL;
+    jresult = (jobject)(*env)->CallLongMethod(env, self->jAccessible,
+                                              jAccessibilityAttributeValue, (jlong)@"AXTitle");
+    GLASS_CHECK_EXCEPTION(env);
+    return variantToID(env, jresult);
+}
```
arapte, line 165: This method is same as `accessibilityLabel`. Should we change implementation to call one method from the other ?

arapte (approved):
LGTM... Tested the behavior of both controls, observed no problem.

andy-goryachev-oracle:
With Slider, it announces the current value as "percent" (of the working envelope?), even though the min/max can be arbitrary.  Is this intentional?

Also, for Spinner, it says "Stepper", is this expected?

macOS 15.3.1 M1

andy-goryachev-oracle:
Is this PR related to  https://git.openjdk.org/jdk/pull/23841 in any way?

modules/javafx.graphics/src/main/native-glass/mac/a11y/AccessibleBase.m
azuev-java: Makes sense.

azuev-java:
> With Slider, it announces the current value as "percent" (of the working envelope?), even though the min/max can be arbitrary. Is this intentional?
> 
> Also, for Spinner, it says "Stepper", is this expected?
> 
> macOS 15.3.1 M1

I need to check that.

azuev-java:
> Is this PR related to https://git.openjdk.org/jdk/pull/23841 in any way?

Not directly. The event delivery in AWT/Swing and in JavaFX works quite differently so i will check if the new implementation here is affected by the same problem - but i doubt so.

andy-goryachev-oracle:
> Spinner is called Stepper

yep, that's what I suspected.  thanks!

azuev-java:
> so the remaining question is whether it should announce "percent" of the slider position or the actual value?

Unfortunately we do not have any control on that. There is no way to tell VoiceOver if it is percent or the raw value - VO deducts it based on its own logic. The only way to customize the VO heuristics is to override accessibilityValueDescription function and return string for VO but then we would need to decode for ourselves how do we convert the value plus it will be our responsibility to do localization of this string if needed.

andy-goryachev-oracle (approved):
yep, you are right: we don't want to deal with interpreting and l10n.

arapte (approved):

PR #1228, opened 2023-09-01
8315317: Add test for JDK-8262518

prsadhuk:
Added automated test for 8262518:SwingNode.setContent does not close previous content, resulting in memory leak

---------
### Progress

### Issue
 * [JDK-8315317](https://bugs.openjdk.org/browse/JDK-8315317): Add test for JDK-8262518 (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1228/head:pull/1228` \
`$ git checkout pull/1228`

`$ git checkout pull/1228` \
`$ git pull https://git.openjdk.org/jfx.git pull/1228/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1228`

`$ git pr show -t 1228`

</details>

<a href="https://git.openjdk.org/jfx/pull/1228.diff">https://git.openjdk.org/jfx/pull/1228.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1228#issuecomment-1702104245)

### DIFF
--- a/tests/system/src/test/java/test/javafx/embed/swing/SwingNodeContentMemoryLeakTest.java
+++ b/tests/system/src/test/java/test/javafx/embed/swing/SwingNodeContentMemoryLeakTest.java
@@ -62,6 +62,7 @@ public class SwingNodeContentMemoryLeakTest {
     private int count = 0;
     private int fail = 0;
     private SwingNode node;
+    private long panelCount;
 
     @BeforeClass
     public static void setupOnce() {
@@ -102,24 +103,17 @@ public class SwingNodeContentMemoryLeakTest {
                     node.setContent(panel);
                 });
 
-                long panelCount = panels.stream().filter(ref ->
+                panelCount = panels.stream().filter(ref ->
                                                      ref.get() != null).count();
-                // Sometimes panel count can shoot upto more than 3 once or twice
-                // due to gc not being guranteed so this check prevents false failure
-                // Without fix, the panel count will increase continuosly so it will
-                // always be more 3 after 1-2 iterations
-                if (panelCount > 3) {
-                    fail++;
-                }
-                System.out.println("iteration " + count + " Panels in memory: "
-                                               + panelCount + " fail " + fail);
-                assertFalse(fail > 2);
+                System.out.println("iteration " + count + " Panels in memory: " + panelCount);
 
                 //I know this doesn't guarantee anything, but prompting a GC gives me more confidence that this
                 //truly is a bug.
                 System.gc();
                 count++;
             }
+            // Check if panelCount has not increased beyond certain threshold
+            assertFalse(panelCount > count/2);
 
         }).start();
 
--- a/tests/system/src/test/java/test/javafx/embed/swing/SwingNodeContentMemoryLeakTest.java
+++ b/tests/system/src/test/java/test/javafx/embed/swing/SwingNodeContentMemoryLeakTest.java
@@ -62,7 +62,7 @@ public class SwingNodeContentMemoryLeakTest {
     private int count = 0;
     private int fail = 0;
     private SwingNode node;
-    
+
     @BeforeClass
     public static void setupOnce() {
         Util.launch(launchLatch, 50, MyApp.class);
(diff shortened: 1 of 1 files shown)


tests/system/src/test/java/test/javafx/embed/swing/SwingNodeContentMemoryLeakTest.java
```diff
@@ -0,0 +1,137 @@
...
+                SwingUtilities.invokeLater(() -> {
+                    JPanel panel = new JPanel();
+                    panels.add(new WeakReference<>(panel));
+                    node.setContent(panel);
+                });
+
+                long panelCount = panels.stream().filter(ref ->
+                                                     ref.get() != null).count();
+                // Sometimes panel count can shoot upto more than 3 once or twice
+                // due to gc not being guranteed so this check prevents false failure
```
andy-goryachev-oracle, line 108: guaranteed

tests/system/src/test/java/test/javafx/embed/swing/SwingNodeContentMemoryLeakTest.java
```diff
@@ -0,0 +1,137 @@
...
+                // Without fix, the panel count will increase continuosly so it will
+                // always be more 3 after 1-2 iterations
+                if (panelCount > 3) {
+                    fail++;
+                }
+                System.out.println("iteration " + count + " Panels in memory: "
+                                               + panelCount + " fail " + fail);
+                assertFalse(fail > 2);
```
andy-goryachev-oracle, line 116: I actually seen the  number to shoot up to 38 (albeit with Thread.sleep(1)).  I wonder if there is a better way to detect failure?  May be the fact that the 'panelCount` is ever increasing is the sign of failure, whereas if at least one time it's dropping we have succeeded.
Or perhaps look at the count when the test complete and fail if panelCount < (attempts / 2) or something

tests/system/src/test/java/test/javafx/embed/swing/SwingNodeContentMemoryLeakTest.java
```diff
@@ -0,0 +1,137 @@
...
+            Scene scene = new Scene(root, 150, 100);
+            stage.setScene(scene);
+        });
+
+        //Kick off a thread that repeatedly creates new JPanels and resets the swing node's content
+        new Thread(() -> {
+            while(count < 50) {
+                //Lets throw in a little sleep so we can read the output
+                try {
+                    Thread.sleep(100);
```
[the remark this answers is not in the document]
andy-goryachev-oracle: they become predictable when the seed is known.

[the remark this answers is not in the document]
kevinrushforth: Yes, but then you would need to modify the test to rerun it with the same seed as the time it failed. If we had a utility that made this easy, then maybe, but I'd still be skeptical. I am not convinced that whatever benefit we might get is worth the effort or the unpredictability.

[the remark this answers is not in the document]
andy-goryachev-oracle: the main benefit is that every new run effectively extends the test coverage.
however, in this particular case, the test won't be reproducible because it depends on many other factors like the platform load.

tests/system/src/test/java/test/javafx/embed/swing/SwingNodeContentMemoryLeakTest.java
prsadhuk: I am not sure I understand the nuances of JMemoryBuddy. I tried using `JMemoryBuddy.assertCollectable` but it fails with 
```
Exception in thread "Thread-7" java.lang.AssertionError: Content of WeakReference was not collected. content: javax.swing.JPanel[,0,0,0x0,layout=java.awt.FlowLayout,alignmentX=0.0,alignmentY=0.0,border=,flags=9,maximumSize=,minimumSize=,preferredSize=]
	at test.util.memory.JMemoryBuddy.assertCollectable(JMemoryBuddy.java:91)
```

maybe because the test doesn't cause SwingNode content count to 0.

But I have modified the test to have better way of detecting failure as suggested and check for total count at end which normally increases without fix..I hope that should be good enough to ensure it fails without the fix and pass with it..

prsadhuk: Any more comments on this test fix after updation @andy-goryachev-oracle ?

andy-goryachev-oracle (approved):
looking good 🕶
even with Thread.sleep(1);

PR #1229, opened 2023-09-02
JDK-8315569: Tests for the contract of SkinBase.layoutChildren(..)

Maran23:
This PR adds a test that verifies the `SkinBase.layoutChildren(..)` method with different scales.
While not explicitly documented, this method will receive the snapped and correctly calculated x, y, width and height values, so that children of the Control can be layouted correctly without requiring to do many more calculations regarding padding.

---------
### Progress

### Issue
 * [JDK-8315569](https://bugs.openjdk.org/browse/JDK-8315569): Tests for the contract of SkinBase.layoutChildren(..) (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1229/head:pull/1229` \
`$ git checkout pull/1229`

`$ git checkout pull/1229` \
`$ git pull https://git.openjdk.org/jfx.git pull/1229/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1229`

`$ git pr show -t 1229`

</details>

<a href="https://git.openjdk.org/jfx/pull/1229.diff">https://git.openjdk.org/jfx/pull/1229.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1229#issuecomment-1703826887)

### DIFF
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlContractTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlContractTest.java
@@ -72,6 +72,7 @@ class ControlContractTest {
         stage.renderScaleXProperty().bind(renderScaleProperty);
         stage.renderScaleYProperty().bind(renderScaleProperty);
 
+        control.setMinSize(200, 300);
         control.setPadding(new Insets(4, 5, 6, 7));
 
         AtomicBoolean layoutChildrenCalled = new AtomicBoolean(false);
@@ -118,6 +119,9 @@ class ControlContractTest {
         stage.show();
 
         assertTrue(layoutChildrenCalled.get());
+
+        assertEquals(200, control.getWidth());
+        assertEquals(300, control.getHeight());
     }
 
 }
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlContractTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlContractTest.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package test.javafx.scene.control;
+
+import javafx.beans.property.DoubleProperty;
+import javafx.beans.property.SimpleDoubleProperty;
+import javafx.geometry.Insets;
+import javafx.scene.Scene;
+import javafx.scene.control.SkinBase;
+import javafx.stage.Stage;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class ControlContractTest {
+
+    private ControlStub control;
+
+    @BeforeEach
+    void setup() {
+        control = new ControlStub();
+    }
+
+    @AfterEach
+    void cleanup() {
+        control = null;
+    }
+
+    /**
+     * Tests the contract of {@link SkinBase#layoutChildren(double, double, double, double)} method that is called
+     * when a {@link javafx.scene.control.Control} should layout itself and his children.
+     * The x, y, width and height should be snapped according to the render scale.
+     * The width and height should subtract the snapped padding insets.
+     *
+     * @param scale the render scale
+     */
+    @ParameterizedTest
+    @ValueSource(doubles = { 1.0, 1.25, 1.5, 1.75, 2.0 })
+    void testLayoutChildrenContract(double scale) {
+        DoubleProperty renderScaleProperty = new SimpleDoubleProperty(scale);
+
+        Stage stage = new Stage();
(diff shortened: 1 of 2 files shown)


kevinrushforth:

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlContractTest.java
```diff
@@ -0,0 +1,127 @@
...
+                assertEquals(expectedX, x);
+
+                double expectedY = control.snappedTopInset();
+                assertEquals(expectedY, y);
+
+                expectedY = control.snapPositionY(padding.getTop());
+                assertEquals(expectedY, y);
+
+                double expectedWidth = control.snapSizeX(control.getWidth()) - control.snappedLeftInset() - control.snappedRightInset();
+                assertEquals(expectedWidth, w);
```
andy-goryachev-oracle, line 101: since we are performing floating point operations, we might accumulate small error.  
So this and 3 subsequent assertEquals() would need an EPSILON = 0.0000001 (an arbitrary value)

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlContractTest.java
```diff
@@ -0,0 +1,127 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class ControlContractTest {
```
andy-goryachev-oracle, line 44: Do you plan to add more tests to this class?  The name does not specify which contract is being tested.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlContractTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+    private ControlStub control;
```
andy-goryachev-oracle, line 46: Do you think it's worth to run this test against all of the existing Controls?

Maran23: Since no `Control` is overriding the `layoutChildren` method, I don't think it will make any difference as of now

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlContractTest.java
Maran23: Do you a better idea? I guess it would make sense to rename it to something more generic. Maybe ControlLayoutTest?

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlContractTest.java
```diff
@@ -0,0 +1,127 @@
...
+    /**
+     * Tests the contract of {@link SkinBase#layoutChildren(double, double, double, double)} method that is called
+     * when a {@link javafx.scene.control.Control} should layout itself and his children.
+     * The x, y, width and height should be snapped according to the render scale.
+     * The width and height should subtract the snapped padding insets.
+     *
+     * @param scale the render scale
+     */
+    @ParameterizedTest
+    @ValueSource(doubles = { 1.0, 1.25, 1.5, 1.75, 2.0 })
```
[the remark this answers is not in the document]
Maran23: sure.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlContractTest.java
Maran23: Will have a look, but it worked without so I did not bothered adding an epsilon to the tests

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlContractTest.java
andy-goryachev-oracle: ControlLayoutChildrenContractTest perhaps, if that's the only thing being tested here (that is, if you don't plan to add anything to this test).

Also, would it make sense to move javadoc from L58 to L43 since it describes the test (assuming no more changes)

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlContractTest.java
andy-goryachev-oracle: I would have added an EPSILON.  I suspect we just got lucky that with the default padding and a small subset of scales everything comes out ok.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlContractTest.java
Maran23: Well right now there is no plan to add more tests but that may change in the future. Will rename it

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlLayoutTest.java
```diff
@@ -0,0 +1,129 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class ControlLayoutTest {
```
andy-goryachev-oracle, line 44: 1. should this class be public?
2. since no more tests are being planned here, perhaps we could rename it to be more specific (ControlLayoutTest still seems to be too generic to me), ControlLayoutChildrenContractTest or something like that?
3. could you please add a javadoc at the class level to explain what generate area this class tests?

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlLayoutTest.java
```diff
@@ -0,0 +1,129 @@
...
+        stage.setScene(scene);
+
+        stage.show();
+
+        assertTrue(layoutChildrenCalled.get());
+
+        assertEquals(200, control.getWidth());
+        assertEquals(300, control.getHeight());
+    }
+
```
andy-goryachev-oracle, line 128: extra newline?

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlLayoutTest.java
```diff
@@ -64,7 +66,7 @@ void cleanup() {
      * @param scale the render scale
      */
     @ParameterizedTest
-    @ValueSource(doubles = { 1.0, 1.25, 1.5, 1.75, 2.0 })
+    @ValueSource(doubles = { 1.0, 1.25, 1.5, 1.75, 2.0, 2.25 })
```
andy-goryachev-oracle, line 69: also tried 0.333333, 0.00111, still looks good :-)

andy-goryachev-oracle (commented):
looks good, with a few minor comments.  please let me know what you think.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ControlLayoutTest.java
Maran23: 1. No, JUnit5 classes can be package private as well and it is actually recommended
2. Well, we don't know. Maybe we will add some at one point, right now I don't have any idea what to add. Therefore I would keep the more generic name so we can reuse it at one point.
3. Same reason as 2., if we add more tests the Javadoc at class level may not be accurate anymore. 
+ There is Javadoc at method level already, which is more than 'necessary' as there is no policy that specifies that javadoc should be added to tests, I just do it voluntarily.

andy-goryachev-oracle (approved):
🕶 looking good

PR #1230, opened 2023-09-04
8308608: [testbug] Use Util::waitForIdle instead of Toolkit::firePulse in system tests

karthikpandelu:
Made changes to use Util::waitForIdle instead of Toolkit::firePulse in system tests. The test will wait for default value of 10 pulses to complete in the scene before executing the subsequent statements.

---------
### Progress

### Issue
 * [JDK-8308608](https://bugs.openjdk.org/browse/JDK-8308608): [testbug] Use Util::waitForIdle instead of Toolkit::firePulse in system tests (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1230/head:pull/1230` \
`$ git checkout pull/1230`

`$ git checkout pull/1230` \
`$ git pull https://git.openjdk.org/jfx.git pull/1230/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1230`

`$ git pr show -t 1230`

</details>

<a href="https://git.openjdk.org/jfx/pull/1230.diff">https://git.openjdk.org/jfx/pull/1230.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1230#issuecomment-1705051170)

### DIFF
--- a/tests/system/src/test/java/test/robot/javafx/scene/ChoiceBoxScrollUpOnCollectionChangeTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/scene/ChoiceBoxScrollUpOnCollectionChangeTest.java
@@ -69,7 +69,6 @@ import test.util.Util;
 
 public class ChoiceBoxScrollUpOnCollectionChangeTest {
     static CountDownLatch startupLatch = new CountDownLatch(1);
-    static CountDownLatch scrollLatch = new CountDownLatch(1);
     static CountDownLatch choiceBoxDisplayLatch = new CountDownLatch(1);
     static CountDownLatch choiceBoxHiddenLatch = new CountDownLatch(1);
     static Robot robot;
@@ -91,16 +90,10 @@ public class ChoiceBoxScrollUpOnCollectionChangeTest {
     }
 
     private void scrollChoiceBox(int scrollAmt) throws Exception {
-        Util.runAndWait(() -> {
-            for (int i = 0; i < scrollAmt; i++) {
-                robot.keyType(KeyCode.DOWN);
-            }
-            scrollLatch.countDown();
-        });
-
-        Util.waitForIdle(scene);
-        Util.waitForLatch(scrollLatch, 5, "Timeout waiting for choicebox to be hidden.");
-        Thread.sleep(400); // Wait for up arrow to get loaded in UI
+        for (int i = 0; i < scrollAmt; i++) {
+            Util.runAndWait(() -> robot.keyType(KeyCode.DOWN));
+            Util.waitForIdle(scene);
+        }
 
         Util.runAndWait(() -> {
             robot.keyType(KeyCode.ENTER);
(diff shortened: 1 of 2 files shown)


tests/system/src/test/java/test/robot/javafx/scene/ChoiceBoxScrollUpOnCollectionChangeTest.java
```diff
@@ -95,11 +94,11 @@ private void scrollChoiceBox(int scrollAmt) throws Exception {
         Util.runAndWait(() -> {
             for (int i = 0; i < scrollAmt; i++) {
                 robot.keyType(KeyCode.DOWN);
-                Toolkit.getToolkit().firePulse();
             }
             scrollLatch.countDown();
         });
 
+        Util.waitForIdle(scene);
```
andy-goryachev-oracle, line 101: is this the right place for waitForIdle?  should it be on L98? since there are multiple key presses involved?

and also, do we still need to .sleep() on L103?

tests/system/src/test/java/test/robot/javafx/scene/ContextMenuNPETest.java
```diff
@@ -101,9 +100,8 @@ private void selectSubmenuItem() throws Exception {
             robot.keyType(KeyCode.DOWN);
             robot.keyType(KeyCode.RIGHT);
             robot.keyType(KeyCode.ENTER);
-            Toolkit.getToolkit().firePulse();
-
         });
+        Util.waitForIdle(scene);
```
andy-goryachev-oracle, line 104: once we have waitForIdle, do we still need to wait() ?

tests/system/src/test/java/test/robot/javafx/scene/ChoiceBoxScrollUpOnCollectionChangeTest.java
karthikpandelu: Ideally yes it should be on L98.  Since we can not call `waitForIdle` from Application thread I added at L101. But if we have to achieve the purpose of using `waitForIdle` in the test then it should be added after each key press. So updated the code to do so.
I had kept the sleep as it is thinking if it might be required in slower systems. Since we are waiting  for 10 pulses with `waitForIdle` it might not be needed. Hence removed it.
Since we are using `Util.runAndWait` there is no need for the `scrollLatch` hence removed it.

tests/system/src/test/java/test/robot/javafx/scene/ContextMenuNPETest.java
karthikpandelu: Same as above comment, I had kept the sleep as it is thinking if it might be required in slower systems. Since we are waiting  for 10 pulses with `waitForIdle` it might not be needed. Hence removed it.

andy-goryachev-oracle (approved):
looks good, thanks!

PR #1240, opened 2023-09-13
8298500: Create test to initially show stage with various attributes (iconified, maximized, full screen)

lukostyra:
PR adds tests mentioned in the title - a new `AttributesTest` class is added testing iconification, maximization and full-screen-ification of a Stage.

All variants are tested with decorated stage style.

Iconification is tested via overlaying two stages on top of one another, and then iconifying the top one - this is similar to already existing `IconifyTest.java` but it tests just the iconfication process and nothing more.

Maximization and FullScreen are both tested by creating two stages _not_ overlapping each other. After maximization/fullscreen top stage (being always on top as well) should cover the bottom stage. Moreover, FullScreen and Maximize are differentiated by checking if window decoration exists - maximized Stage will have its decoration taking space on top of the screen, whereas FullScreen one will not.

**NOTE:** on macOS I had issues with `getColor()` returning a valid color when called a second time. This only happened on macOS and with FullScreen test (others worked fine). Unfortunately I couldn't figure out why it returned (0, 0, 0, 255) or (255, 255, 255, 255). To mitigate that I moved color checks into separate `runAndWait()`-s with a small sleep between them, which seemed to help `getColor()` return proper values.

Verified to work on Windows 11, macOS and Linux.

---------
### Progress

### Issue
 * [JDK-8298500](https://bugs.openjdk.org/browse/JDK-8298500): Create test to initially show stage with various attributes (iconified, maximized, full screen) (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1240/head:pull/1240` \
`$ git checkout pull/1240`

`$ git checkout pull/1240` \
`$ git pull https://git.openjdk.org/jfx.git pull/1240/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1240`

`$ git pr show -t 1240`

</details>

<a href="https://git.openjdk.org/jfx/pull/1240.diff">https://git.openjdk.org/jfx/pull/1240.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1240#issuecomment-1717468480)

### DIFF
--- a/tests/system/src/test/java/test/robot/javafx/stage/AttributesTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/stage/AttributesTest.java
@@ -1,0 +1,187 @@
+/*
+ * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package test.robot.javafx.stage;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import org.junit.After;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import javafx.application.Platform;
+import javafx.scene.Scene;
+import javafx.stage.Stage;
+import javafx.stage.StageStyle;
+import javafx.scene.layout.Pane;
+import javafx.scene.paint.Color;
+import static junit.framework.Assert.assertFalse;
+import static junit.framework.Assert.assertTrue;
+import test.robot.testharness.VisualTestBase;
+
+import static test.util.Util.TIMEOUT;
+
+public class AttributesTest extends VisualTestBase {
+
+    private static final int WIDTH = 400;
+    private static final int HEIGHT = 400;
+
+    private static final Color BOTTOM_COLOR = Color.LIME;
+    private static final Color TOP_COLOR = Color.RED;
+
+    private static final double TOLERANCE = 0.07;
+
+    private Stage bottomStage;
+    private Stage topStage;
+
+    private void setupStages(boolean overlayed) throws InterruptedException {
+        final CountDownLatch bottomShownLatch = new CountDownLatch(1);
+        final CountDownLatch topShownLatch = new CountDownLatch(1);
+
+        runAndWait(() -> {
+            // Bottom stage, should be visible after top stage is iconified
+            bottomStage = getStage(false);
+            bottomStage.initStyle(StageStyle.DECORATED);
+            Scene bottomScene = new Scene(new Pane(), WIDTH, HEIGHT);
+            bottomScene.setFill(BOTTOM_COLOR);
+            bottomStage.setScene(bottomScene);
+            bottomStage.setX(0);
+            bottomStage.setY(0);
+            bottomStage.setOnShown(e -> Platform.runLater(bottomShownLatch::countDown));
+            bottomStage.show();
+        });
+
+        assertTrue("Timeout waiting for bottom stage to be shown",
+            bottomShownLatch.await(TIMEOUT, TimeUnit.MILLISECONDS));
+
+        runAndWait(() -> {
+            // Top stage, will be inconified
+            topStage = getStage(true);
+            topStage.initStyle(StageStyle.DECORATED);
+            Scene topScene = new Scene(new Pane(), WIDTH, HEIGHT);
+            topScene.setFill(TOP_COLOR);
+            topStage.setScene(topScene);
+            if (overlayed) {
+                topStage.setX(0);
+                topStage.setY(0);
+            } else {
+                topStage.setX(WIDTH);
+                topStage.setY(HEIGHT);
+            }
+            topStage.setOnShown(e -> Platform.runLater(topShownLatch::countDown));
+            topStage.show();
+        });
+
+        assertTrue("Timeout waiting for top stage to be shown",
+            topShownLatch.await(TIMEOUT, TimeUnit.MILLISECONDS));
+    }
+
+    public @Test void testIconifiedStage() throws InterruptedException {
+        setupStages(true);
+
+        runAndWait(() -> {
+            Color color = getColor(200, 200);
+            assertColorEquals(TOP_COLOR, color, TOLERANCE);
+
+            topStage.setIconified(true);
+        });
+
+        // wait a bit to let window system animate the change
+        sleep(1000);
+
+        runAndWait(() -> {
+            assertTrue(topStage.isIconified());
+            Color color = getColor(200, 200);
+            assertColorEquals(BOTTOM_COLOR, color, TOLERANCE);
+        });
+    }
+
+    public @Test void testMaximizedStage() throws InterruptedException {
+        setupStages(false);
+
+        runAndWait(() -> {
+            Color color = getColor(200, 200);
+            assertColorEquals(BOTTOM_COLOR, color, TOLERANCE);
+
+            topStage.setMaximized(true);
+        });
+
+        // wait a bit to let window system animate the change
+        sleep(1000);
+
+        runAndWait(() -> {
+            assertTrue(topStage.isMaximized());
+
+            // maximized stage should take over the bottom stage
+            Color color = getColor(200, 200);
+            assertColorEquals(TOP_COLOR, color, TOLERANCE);
+        });
+
+        // wait a little bit between getColor() calls - on macOS the below one
+        // would fail without this wait
+        sleep(100);
+
+        runAndWait(() -> {
+            // top left corner (plus some tolerance) should show decorations (so not TOP_COLOR)
+            Color color = getColor((int)topStage.getX() + 10, (int)topStage.getY() + 10);
+            assertColorDoesNotEqual(TOP_COLOR, color, TOLERANCE);
+            assertColorDoesNotEqual(BOTTOM_COLOR, color, TOLERANCE);
+        });
+    }
+
+    public @Test void testFullscreenStage() throws InterruptedException {
+        setupStages(false);
+
+        runAndWait(() -> {
+            Color color = getColor(200, 200);
+            assertColorEquals(BOTTOM_COLOR, color, TOLERANCE);
+
+            topStage.setFullScreen(true);
+        });
+
+        // wait a bit to let window system animate the change
+        sleep(1000);
+
+        runAndWait(() -> {
+            assertTrue(topStage.isFullScreen());
+
+            // fullscreen stage should take over the bottom stage
+            Color color = getColor(200, 200);
+            assertColorEquals(TOP_COLOR, color, TOLERANCE);
+        });
+
+        // wait a little bit between getColor() calls - on macOS the below one
+        // would fail without this wait
+        sleep(100);
+
+        runAndWait(() -> {
+            // top left corner (plus some tolerance) should NOT show decorations
+            Color color = getColor((int)topStage.getX() + 5, (int)topStage.getY() + 5);
+            assertColorEquals(TOP_COLOR, color, TOLERANCE);
+        });
+    }
+}
--- a/tests/system/src/test/java/test/robot/javafx/stage/StageAttributesTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/stage/StageAttributesTest.java
@@ -32,6 +32,8 @@ import org.junit.After;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
 
 import javafx.application.Platform;
 import javafx.scene.Scene;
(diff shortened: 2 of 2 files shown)


tests/system/src/test/java/test/robot/javafx/stage/AttributesTest.java
```diff
@@ -0,0 +1,187 @@
...
+import javafx.stage.StageStyle;
+import javafx.scene.layout.Pane;
+import javafx.scene.paint.Color;
+import static junit.framework.Assert.assertFalse;
+import static junit.framework.Assert.assertTrue;
+import test.robot.testharness.VisualTestBase;
+
+import static test.util.Util.TIMEOUT;
+
+public class AttributesTest extends VisualTestBase {
```
andy-goryachev-oracle, line 48: a brief javadoc explaining what this test does would be nice.
also, perhaps the class name could be more descriptive, i.e. StageAttributesTest or something like that.

tests/system/src/test/java/test/robot/javafx/stage/AttributesTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+        runAndWait(() -> {
+            Color color = getColor(200, 200);
+            assertColorEquals(BOTTOM_COLOR, color, TOLERANCE);
+
+            topStage.setMaximized(true);
+        });
+
+        // wait a bit to let window system animate the change
+        sleep(1000);
```
andy-goryachev-oracle, line 134: could we use Util.waitForIdle() here instead of the fixed timeout?

tests/system/src/test/java/test/robot/javafx/stage/AttributesTest.java
```diff
@@ -0,0 +1,187 @@
...
+            assertTrue(topStage.isMaximized());
+
+            // maximized stage should take over the bottom stage
+            Color color = getColor(200, 200);
+            assertColorEquals(TOP_COLOR, color, TOLERANCE);
+        });
+
+        // wait a little bit between getColor() calls - on macOS the below one
+        // would fail without this wait
+        sleep(100);
```
andy-goryachev-oracle, line 146: same - waitForIdle?

tests/system/src/test/java/test/robot/javafx/stage/AttributesTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+public class AttributesTest extends VisualTestBase {
+
+    private static final int WIDTH = 400;
+    private static final int HEIGHT = 400;
+
+    private static final Color BOTTOM_COLOR = Color.LIME;
+    private static final Color TOP_COLOR = Color.RED;
+
+    private static final double TOLERANCE = 0.07;
```
andy-goryachev-oracle, line 56: just curious: how did we arrive at this value?

andy-goryachev-oracle (commented):
tested on macOS 13.5.1 (looks good), with some minor comments and suggestions.

tests/system/src/test/java/test/robot/javafx/stage/AttributesTest.java
lukostyra: I'll add the javadoc.

I skipped the `Stage` prefix since this test is in a directory of `Stage`-related tests. Should I still add it despite that?

andy-goryachev-oracle: The more specific the class name, the better, I think (and also, specific names reduce collisions)

tests/system/src/test/java/test/robot/javafx/stage/AttributesTest.java
kevinrushforth: This constant, with the same value of `0.07`, is repeated in most of our robot screen capture tests. It seems a good candidate for moving it to the `VisualTestBase` parent class, as a future test enhancement.

tests/system/src/test/java/test/robot/javafx/stage/AttributesTest.java
```diff
@@ -0,0 +1,187 @@
...
+import javafx.scene.Scene;
+import javafx.stage.Stage;
+import javafx.stage.StageStyle;
+import javafx.scene.layout.Pane;
+import javafx.scene.paint.Color;
+import static junit.framework.Assert.assertFalse;
+import static junit.framework.Assert.assertTrue;
+import test.robot.testharness.VisualTestBase;
+
+import static test.util.Util.TIMEOUT;
```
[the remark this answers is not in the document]
kevinrushforth: I prefer the static import in this case. More to the point, this is what most other tests do.

tests/system/src/test/java/test/robot/javafx/stage/AttributesTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+    public @Test void testIconifiedStage() throws InterruptedException {
+        setupStages(true);
+
+        runAndWait(() -> {
+            Color color = getColor(200, 200);
+            assertColorEquals(TOP_COLOR, color, TOLERANCE);
+
+            topStage.setIconified(true);
+        });
```
kevinrushforth, line 111: This will show the stage before setting it as iconified. Testing correct behavior for a stage that is initially iconified/maximized/fullScreen prior to showing the stage is the main purpose of this test enhancement, so you will need to do something like split the creation and showing of the stages into separate methods or pass flags for the initial state of those three attributes as arguments to the setupStages method.

tests/system/src/test/java/test/robot/javafx/stage/AttributesTest.java
kevinrushforth: I don't think so. This is a small sleep that isn't intended to wait for idle.

tests/system/src/test/java/test/robot/javafx/stage/AttributesTest.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    public @Test void testFullscreenStage() throws InterruptedException {
+        setupStages(false);
+
+        runAndWait(() -> {
+            Color color = getColor(200, 200);
+            assertColorEquals(BOTTOM_COLOR, color, TOLERANCE);
+
+            topStage.setFullScreen(true);
```
kevinrushforth, line 163: Same comment about needing to set fullScreen before showing topStage.

tests/system/src/test/java/test/robot/javafx/stage/AttributesTest.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    public @Test void testMaximizedStage() throws InterruptedException {
+        setupStages(false);
+
+        runAndWait(() -> {
+            Color color = getColor(200, 200);
+            assertColorEquals(BOTTOM_COLOR, color, TOLERANCE);
+
+            topStage.setMaximized(true);
```
kevinrushforth, line 130: Same comment about needing to set maximized before showing topStage.

kevinrushforth (commented):
I left a few inline comment. I still need to test it, but I'll wait until you address my main feedback about the order of operations before I do too much testing.

tests/system/src/test/java/test/robot/javafx/stage/AttributesTest.java
kevinrushforth: The post-show behavior is already covered by other tests (e.g., IconifyTest). It wouldn't hurt to provide additional tests if easy, but it's not really necessary.

lukostyra:
I addressed review remarks. I noticed following behaviors when testing the change:
- On Windows all tests pass
- On macOS `testMaximizedStageBeforeShow` and `testIconifiedStageBeforeShow` fail because stages are not shown already iconified/maximized. Rest of the tests work fine - I assume a bug in functionality that needs to be fixed.
- On Linux all tests fail, on my VM (Fedora 38 with X11) the stages are created without taking into account WIDTH/HEIGHT parameters provided to the Scene. They are created smaller than expected (I guess 100x100), so `getColor()` calls return background colors (terminal background/letters in my case). This could be cured in-test of course, but it also seems like a bug that should be ironed out (we do expect stages to be larger after all). After forcing proper width/height, testMaximizedStageBeforeShow/testFullScreenStageBeforeShow both fail - I noticed top stage is created maximized/fullscreen and immediately brought back to default dimensions, so I think it's another to-be-fixed bug.

kevinrushforth:
> * On macOS `testMaximizedStageBeforeShow` and `testIconifiedStageBeforeShow` fail because stages are not shown already iconified/maximized. Rest of the tests work fine - I assume a bug in functionality that needs to be fixed.

[JDK-8305675](https://bugs.openjdk.org/browse/JDK-8305675) is already tracking the failure of an initially iconified window. You can use that bug ID when adding the "assumeFalse" to exclude the test on macOS.

I don't see a bug tracking the failure of a maximized window. [JDK-8253997](https://bugs.openjdk.org/browse/JDK-8253997) describes a problem where the window is shown normal size and then animated to maximized. Perhaps the behavior has changed on recent macOS systems. Can you check by running that test program attached to that bug? If it behaves the same way as your new test, then we can repurpose that bug; otherwise, please file a new bug.

> * On Linux all tests fail, on my VM (Fedora 38 with X11) the stages are created without taking into account WIDTH/HEIGHT parameters provided to the Scene. They are created smaller than expected (I guess 100x100), so `getColor()` calls return background colors (terminal background/letters in my case). This could be cured in-test of course, but it also seems like a bug that should be ironed out (we do expect stages to be larger after all). After forcing proper width/height, testMaximizedStageBeforeShow/testFullScreenStageBeforeShow both fail - I noticed top stage is created maximized/fullscreen and immediately brought back to default dimensions, so I think it's another to-be-fixed bug.

Yes, it does seem like there are two or three product bugs on Linux. Can you file new bugs, and then use those bug IDs to exclude the failing tests on Linux? I'll test this on Ubuntu and see if that behaves differently than Fedora.

lukostyra:
> I don't see a bug tracking the failure of a maximized window. [JDK-8253997](https://bugs.openjdk.org/browse/JDK-8253997) describes a problem where the window is shown normal size and then animated to maximized

I found out this is a problem when trying to call `setMaximized(true)` after establishing Stage position with `setX()` and `setY()`. Seems like a separate issue, I will file a bug with the test program I just made (also it is macOS exclusive, Windows handles this just fine).

> Yes, it does seem like there are two or three product bugs on Linux. Can you file new bugs, and then use those bug IDs to exclude the failing tests on Linux? I'll test this on Ubuntu and see if that behaves differently than Fedora.

Will do. I'll also attach test programs for these.

tests/system/src/test/java/test/robot/javafx/stage/StageAttributesTest.java
```diff
@@ -0,0 +1,325 @@
...
+            if (topShown) {
+                topStage.setOnShown(e -> Platform.runLater(topShownLatch::countDown));
+                topStage.show();
+            }
+        });
+
+        if (topShown) {
+            assertTrue("Timeout waiting for top stage to be shown",
+                topShownLatch.await(TIMEOUT, TimeUnit.MILLISECONDS));
+        }
```
kevinrushforth, line 109: Several tests fail for me without an additional sleep. I recommend adding `Util.sleep(1000)` here.

tests/system/src/test/java/test/robot/javafx/stage/StageAttributesTest.java
```diff
@@ -0,0 +1,325 @@
...
+
+        runAndWait(() -> {
+            Color color = getColor(200, 200);
+            assertColorEquals(TOP_COLOR, color, TOLERANCE);
+
+            topStage.setIconified(true);
+        });
+
+        // wait a bit to let window system animate the change
+        Util.waitForIdle(topScene);
```
kevinrushforth, line 128: In looking at this more closely, this change should be reverted back to using `sleep(1000)`. `Util.waitForIdle` is intended to wait for scene graph changes to be propagated, not for platform windowing operations to settle down, so using it could lead to intermittent failures.

kevinrushforth (changes requested):
On Windows 10 system, several tests fail for me:

```
StageAttributesTest > testFullScreenStageBeforeShow FAILED
    junit.framework.AssertionFailedError: expected:rgba(0,255,0,255) but was:rgba(42,244,71,255)
        at test.robot.testharness.VisualTestBase.assertColorEquals(VisualTestBase.java:173)
        at test.robot.javafx.stage.StageAttributesTest.lambda$testFullScreenStageBeforeShow$17(StageAttributesTest.java:298)

StageAttributesTest > testIconifiedStageBeforeShow FAILED
    junit.framework.AssertionFailedError: expected:rgba(0,255,0,255) but was:rgba(28,248,48,255)
        at test.robot.testharness.VisualTestBase.assertColorEquals(VisualTestBase.java:173)
        at test.robot.javafx.stage.StageAttributesTest.lambda$testIconifiedStageBeforeShow$12(StageAttributesTest.java:226)

StageAttributesTest > testMaximizedStage FAILED
    junit.framework.AssertionFailedError: expected:rgba(0,255,0,255) but was:rgba(23,249,38,255)
        at test.robot.testharness.VisualTestBase.assertColorEquals(VisualTestBase.java:173)
        at test.robot.javafx.stage.StageAttributesTest.lambda$testMaximizedStage$6(StageAttributesTest.java:147)

StageAttributesTest > testFullScreenStage FAILED
    junit.framework.AssertionFailedError: expected:rgba(0,255,0,255) but was:rgba(28,248,48,255)
        at test.robot.testharness.VisualTestBase.assertColorEquals(VisualTestBase.java:173)
        at test.robot.javafx.stage.StageAttributesTest.lambda$testFullScreenStage$9(StageAttributesTest.java:185)

StageAttributesTest > testIconifiedStage FAILED
    junit.framework.AssertionFailedError: expected:rgba(255,0,0,255) but was:rgba(177,84,13,255)
        at test.robot.testharness.VisualTestBase.assertColorEquals(VisualTestBase.java:173)
        at test.robot.javafx.stage.StageAttributesTest.lambda$testIconifiedStage$4(StageAttributesTest.java:122)

StageAttributesTest > testMaximizedStageBeforeShow FAILED
    junit.framework.AssertionFailedError: expected:rgba(0,255,0,255) but was:rgba(23,249,39,255)
        at test.robot.testharness.VisualTestBase.assertColorEquals(VisualTestBase.java:173)
        at test.robot.javafx.stage.StageAttributesTest.lambda$testMaximizedStageBeforeShow$14(StageAttributesTest.java:258)
```

I tracked this down to a missing sleep in `setupStages` (see inline comments).

tests/system/src/test/java/test/robot/javafx/stage/AttributesTest.java
kevinrushforth: Actually, no. This isn't the intended use case for `waitForIdle`. That method should only be used when waiting for scene graph changes to propagate, not for the platform to finish showing or animating a window.

lukostyra:
Added both the `sleep(1000)` at the end of `setupStages` and replaced `waitForIdle` back to sleeps.

kevinrushforth (approved):
Looks good.

kevinrushforth:
@andy-goryachev-oracle Did you want a change to review this?

andy-goryachev-oracle (approved):
double checked on macOS, looks good.
thanks!

PR #1242, opened 2023-09-15
8255079: RobotTest::testPixelCaptureAverage fails intermittently on Windows with HiDPI scaling

lukostyra:
The instability came from the way Windows sets window's default position. When it is not predefined on Stage creation (which was the case in these Robot tests) the window is created at some arbitrary position chosen by Windows near top-left corner. It also seems like Windows first picks the position assuming 100% scaling and then multiplies it by UI scale, which can provide X/Y coordinates of Stage with a fractional value.

Due to above behavior, there is a chance that the fractional part of X/Y will be above .5 which fails the test (X/Y values were fetched via casting directly to `int`, which drops the fractional part without rounding). Adding rounding makes the test always pass and the assumption in the comment below consistent.

`unstable.test` property check was removed, since this change stabilizes the test and its results.

Verified also on macOS to ensure the change did not affect the tests.

---------
### Progress

### Issue
 * [JDK-8255079](https://bugs.openjdk.org/browse/JDK-8255079): RobotTest::testPixelCaptureAverage fails intermittently on Windows with HiDPI scaling (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1242/head:pull/1242` \
`$ git checkout pull/1242`

`$ git checkout pull/1242` \
`$ git pull https://git.openjdk.org/jfx.git pull/1242/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1242`

`$ git pr show -t 1242`

</details>

<a href="https://git.openjdk.org/jfx/pull/1242.diff">https://git.openjdk.org/jfx/pull/1242.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1242#issuecomment-1721465318)

### DIFF
--- a/tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2018, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


kevinrushforth (approved):
Looks good. The test is now stable on my Windows 10 system using 125% scaling.

PR #1244, opened 2023-09-19
8316518 javafx.print.Paper getWidth / getHeight rounds values, causing errors.

FlorianKirmaier:
The Method javafx.print.getWidth and getHeight returns a double of points.
But the values are always rounded to full integers. This causes especially problems,
with mm based Papers.

I asked in the mailing list, with the conclusion (me and John Hendrix) that is best to just fix this behavior by removing the rounding.

---------
### Progress
- [x] Change must be properly reviewed (4 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 3 [Authors](https://openjdk.org/bylaws#author))

### Issue
 * [JDK-8316518](https://bugs.openjdk.org/browse/JDK-8316518): javafx.print.Paper getWidth / getHeight rounds values, causing errors. (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1244/head:pull/1244` \
`$ git checkout pull/1244`

`$ git checkout pull/1244` \
`$ git pull https://git.openjdk.org/jfx.git pull/1244/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1244`

`$ git pr show -t 1244`

</details>

<a href="https://git.openjdk.org/jfx/pull/1244.diff">https://git.openjdk.org/jfx/pull/1244.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1244#issuecomment-1725375249)

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/print/Paper.java
+++ b/modules/javafx.graphics/src/main/java/javafx/print/Paper.java
@@ -78,9 +78,9 @@ public final class Paper {
      */
     private double getSizeInPoints(double dim) {
         switch (units) {
-        case POINT : return (int)(dim+0.5);
-        case INCH  : return (int)((dim * 72) + 0.5);
-        case MM    : return (int)(((dim * 72) / 25.4) + 0.5);
+            case POINT : return dim;
+            case INCH  : return dim * 72;
+            case MM    : return (dim * 72) / 25.4;
         }
         return dim;
     }
--- a/modules/javafx.graphics/src/main/java/javafx/print/Paper.java
+++ b/modules/javafx.graphics/src/main/java/javafx/print/Paper.java
@@ -77,12 +77,11 @@ public final class Paper {
      * Translate the internally stored dimension into points.
      */
     private double getSizeInPoints(double dim) {
-        switch (units) {
-            case POINT : return dim;
-            case INCH  : return dim * 72;
-            case MM    : return (dim * 72) / 25.4;
-        }
-        return dim;
+        return switch (units) {
+            case POINT -> dim;
+            case INCH -> dim * 72;
+            case MM -> (dim * 72) / 25.4;
+        };
     }
 
     /**
--- a/modules/javafx.graphics/src/test/java/test/javafx/print/PaperUnitsTest.java
+++ b/modules/javafx.graphics/src/test/java/test/javafx/print/PaperUnitsTest.java
@@ -25,6 +25,7 @@
 
 package test.javafx.print;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 import org.junit.Test;
 
@@ -65,11 +66,11 @@ public class PaperUnitsTest {
      double mmWid = 100.0;
      double mmHgt = 200.0;
      Paper p = PrintHelper.createPaper("TestMM", mmWid, mmHgt, Units.MM);
-     int ptsWid = (int)p.getWidth();
-     int ptsHgt = (int)p.getHeight();
-     int expectedPtsWid = (int)(((mmWid * 72) / 25.4) + 0.5);
-     int expectedPtsHgt = (int)(((mmHgt * 72) / 25.4) + 0.5);
-     assertTrue("MM width is not as expected", ptsWid == expectedPtsWid);
-     assertTrue("MM height is not as expected", ptsHgt == expectedPtsHgt);
+     double ptsWid = p.getWidth();
+     double ptsHgt = p.getHeight();
+     double expectedPtsWid = (mmWid * 72) / 25.4;
+     double expectedPtsHgt = (mmHgt * 72) / 25.4;
+     assertEquals("MM width is not as expected", ptsWid, expectedPtsWid, 0.001);
+     assertEquals("MM height is not as expected", ptsHgt, expectedPtsHgt, 0.001);
    }
 }


FlorianKirmaier:
According to my test - I'm still able to print.
I also don't see anything in the code, which implies that the rounding was necessary.
Is there something specific that should be tested on changes in printing?

kevinrushforth:
@prrace needs to review this change.

/reviewers 2

prrace:
FX uses doubles as a rule and it doesn't preclude that the actual values may have no fractional part.
NA paper is measured in INCHes and if we have some FP error and end up with 792.0001 pts then the rounding would help.
So I am not sure about removing it completely. 
What exact problem does it cause you ? Paper matching to A4 etc should be working SFAIK.
Perhaps we can tweak the MM case only. Not rounding to whole values but a close FP representation.

FlorianKirmaier:
Many of the Papers in the Paper Class are measured in MM, and the Paper Class simply returns wrong values.
A rounding like `.00001` should cause less problems than `.465` (A4).
A4 Paper simply returns wrong values.
If the developer uses these values, then the computations are wrong.
By using the correct values, we get visible more correct results.

In the mailing list, I also suggested 2 other solutions:
**Solution2:**
We could add new methods getExactWidth() / getExactHeight() , which returns the exact value without a rounding problem anywhere noticeable.

**Solution3:**
We could provide an API to access the width/height based on a given Unit.
like: getWidth(Unit.MM).
We already have implemented this, in our application code, so it wouldn’t be much effort.
But it would require moving com.sun.javafx.print.Units back to javafx.print.Paper.Units. (basically reverting JDK-8093699)


I could also implement one of these solutions, if that's preferred.
But i still don't see, how this rounding doesn't cause many unknown problems.

modules/javafx.graphics/src/main/java/javafx/print/Paper.java
```diff
@@ -78,9 +78,9 @@ public final String getName() {
      */
     private double getSizeInPoints(double dim) {
         switch (units) {
-        case POINT : return (int)(dim+0.5);
-        case INCH  : return (int)((dim * 72) + 0.5);
-        case MM    : return (int)(((dim * 72) / 25.4) + 0.5);
+            case POINT : return dim;
+            case INCH  : return dim * 72;
+            case MM    : return (dim * 72) / 25.4;
```
andy-goryachev-oracle, line 83: yeah,  I also don't quite understand why we have the rounding (I mean, floor'ing) here.
I would vote to remove the typecast as the OP proposes.

But there is more:
1. why in the world the constructor is not public?  what if I need to print a Commercial 8-5/8 envelope? 
see https://www.papersizes.org/us-envelope-sizes.htm

2. what is going on in hashCode??
```
    public final int hashCode() {
        return (int)width+((int)height<<16)+units.hashCode();
    }
```

kevinrushforth: > yeah, I also don't quite understand why we have the rounding (I mean, floor'ing) here. I would vote to remove the typecast as the OP proposes.

Phil has some concerns that will need to be looked into.

> But there is more:
> 
> 1. why in the world the constructor is not public?  what if I need to print a Commercial 8-5/8 envelope?

That's the wrong question to ask, and this PR is the wrong place to ask it. If I could reformulate your question, I might rather state it as "would it be a useful enhancement to allow Paper to be customizable?". Maybe (or maybe not), but as with any other new feature, it needs proper motivation, etc.

> 2. what is going on in hashCode??
> 
> ```
>     public final int hashCode() {
>         return (int)width+((int)height<<16)+units.hashCode();
>     }
> ```

This is also unrelated to this PR. It may or may not be a great `hashCode` implementation, but It meets the contract of hasCode w.r.t. equals. And if it didn't, it would need a separate bug.

kevinrushforth: > > yeah, I also don't quite understand why we have the rounding (I mean, floor'ing) here. I would vote to remove the typecast as the OP proposes.
> 
> Phil has some concerns that will need to be looked into.

To follow-on, it seems that there are a few options.

1. Pass back the unmodified value. The concern that Phil raised is that this could lead to floating-point error. Of course the current code that truncates could also lead to errors. If the value ends up as, say, 10.9999 inches, we would want to return 11 rather than 10 even if we did want to limit it to integers.
2. Round the value to whole integers. This doesn't really solve the problem that the PR is trying to solve
3. Round to some number of decimal places, e.g., round to the nearest 0.01 or 0.001 inches.
4. Something else.

Option 3 seems like it might be a reasonable solution, but I'd want Phil to weigh in.

andy-goryachev-oracle: what kind of FP error could option 1 lead to?

I think rounding belongs to the UI, to the panel which displays the sizes with precision given by the requirements.  So this method, I think, should return the unmodified value.

Looking at AWT Paper, I see no rounding there.  Where did it come from to FX?

PS. my other comments are indeed unrelated to this PR.  sorry.

hjohn: Some observations on the paper class:

    /**
     * Specifies the North American legal size, 8.5 inches by 14 inches.
     */
    public static final Paper LEGAL = new Paper("Legal", 8.4, 14, INCH);


The comment and actual size contradict each other. 8.4 inches would not result in an integer number of points (604.8) and looking up the legal paper size confirms it should be 8.5 inches

    /**
     * Specifies the Monarch envelope size, 3.87 inch by 7.5 inch.
     */
    public static final Paper MONARCH_ENVELOPE = new Paper("Monarch Envelope", 3.87, 7.5, INCH);

This seems incorrect, the size is 3-7/8” x 7-1/2”, which would be `3.875` x `7.5` -- the difference in this case is small enough that it will be rounded to the same value in points.

hjohn: About the rounding; all these values are fixed values and could have been entered in points as constants directly.

Also realize that all papers measured in inches can be expressed exactly in points (after the above errors are fixed) so no rounding is needed for those at all; the same is not true for the papers specified in millimeters.

Since the papers in inches never should need any rounding, the rounding can therefore be removed for those.  If there is a worry that the multiplication (* 72) may cause floating point errors, then I suggest to specify all papers in points immediately, like this:

    new Paper("Letter", 8.5 * 72, 11 * 72, POINT);

As for the papers specified in millimeters, these can't be expressed in whole points, as a millimeter is equivalent to `2,83465` points.  Rounding them to points will require some guess work to find out what the actual size in millimeters was, for example, when it says 3 points it must have been 1 mm, while 1, 2, 4 or 5 points has no direct mm equivalent.

hjohn: It is also very odd that `Paper` would expose the name (like `A4`), but not the unit (`MM`) and the exact sizes for those.  Representing these in a UI therefore becomes terribly complicated; if I want to show the user, I can show them the name, but then when I want to display its size, I have to show it in points (a useless measurement, for users, for both papers specific in inches and mm).

beldenfox: Typically paper sizes are displayed in whatever units are appropriate for the locale. On my Mac all paper sizes are displayed in whole millimeters if the locale is United Kingdom and inches (to 1/100 of an inch) if the locale is United States.

hjohn:
I did a fix for the incorrect legal paper size here: https://github.com/openjdk/jfx/pull/1248

FlorianKirmaier:
Relating to the wrong Paper sizes - I have this strange suspicion, that sizes were wrong - to get the right roundings at some point in time.

FlorianKirmaier:
It would be nice if something could be decided here.

Another option would be, to rename the current method to something like getWidthInPoints() which then can be used internally. But the old - now correct methods - can be exposed to applications, returning correct values.

But I would still prefer to just fix it, returning the correct values. And if it turns out, at some part we need it in points, then we just have to convert it to this location. I still don't see how the current state doesn't lead to various incorrect behaviors.

modules/javafx.graphics/src/main/java/javafx/print/Paper.java
```diff
@@ -78,9 +78,9 @@ public final String getName() {
...
         }
         return dim;
```
mstr2, line 85: Minor: you can use an enhanced `switch` expression here, which also ensures exhaustiveness:
```java
return switch (units) {
    case POINT -> dim;
    case INCH -> dim * 72;
    case MM -> (dim * 72) / 25.4;
};
```

FlorianKirmaier: done

mstr2 (approved):

andy-goryachev-oracle:
I think the fix is the right one.

hjohn (approved):
LGTM

andy-goryachev-oracle:
please wait before integrating this until @prrace has a chance to take another look at it.

andy-goryachev-oracle:
(testing jcheck fix)

prrace (approved):
Sorry about the very lengthy delay here. I've gone back over what little I have left from 11 years ago and I can't find a specific reason for the rounding.
My best guess is that it was about "UI", meaning ideally a user would not see values like 611.9985349
If that becomes a problem we can revert this i fwe can't handle it elsewhere.

kevinrushforth:
@FlorianKirmaier This PR is ready for you to `/integrate`

PR #1245, opened 2023-09-19
8185831: Pseudo selectors do not appear to work in Node.lookupAll()

SaiPradeepDandem:
**Issue:**
Using pseudo classes in programmatic query using Node.lookupAll() or Node.lookup() gives unexpected results.

**Cause:**
There is no check for checking the psuedo states matching in the applies() method of SimpleSelector.java. So checking for "applies()" alone is not sufficient in lookup() method.

**Fix:**
Included an extra check for the psuedo states to match.

---------
### Progress

### Issue
 * [JDK-8185831](https://bugs.openjdk.org/browse/JDK-8185831): Pseudo selectors do not appear to work in Node.lookupAll() (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1245/head:pull/1245` \
`$ git checkout pull/1245`

`$ git checkout pull/1245` \
`$ git pull https://git.openjdk.org/jfx.git pull/1245/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1245`

`$ git pr show -t 1245`

</details>

<a href="https://git.openjdk.org/jfx/pull/1245.diff">https://git.openjdk.org/jfx/pull/1245.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1245#issuecomment-1725409162)

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -1988,6 +1988,7 @@ public abstract class Node implements EventTarget, Styleable {
      *
      * @param selector The Selector.
      * @param results The results.
+     * @return List of matching nodes. The returned value can be null.
      */
     List<Node> lookupAll(Selector selector, List<Node> results) {
         if (selectorMatches(selector)) {
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -1973,7 +1973,7 @@ public abstract class Node implements EventTarget, Styleable {
      * Finds all {@code Node}s, including this one and any children, which match
      * the given CSS selector. If no matches are found, an empty unmodifiable set is
      * returned. The set is explicitly unordered.
-     *<p>
+     * <p>
      *     For example, if there are multiple nodes with same style class "myStyle", then the lookupAll method can
      *     be used to find all these nodes as follows: <code>scene.lookupAll(".myStyle");</code>.
      * </p>
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -1986,8 +1986,8 @@ public abstract class Node implements EventTarget, Styleable {
      * Used by Node and Parent for traversing the tree and adding all nodes which
      * match the given selector.
      *
-     * @param selector The Selector. This will never be null.
-     * @param results The results. This will never be null.
+     * @param selector The Selector.
+     * @param results The results.
      */
     List<Node> lookupAll(Selector selector, List<Node> results) {
         if (selectorMatches(selector)) {
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -1961,7 +1961,7 @@ public abstract class Node implements EventTarget, Styleable {
     public Node lookup(String selector) {
         if (selector == null) return null;
         Selector s = Selector.createSelector(selector);
-        return s != null && s.applies(this) ? this : null;
+        return selectorMatches(s) ? this : null;
     }
 
     /**
@@ -1990,7 +1990,7 @@ public abstract class Node implements EventTarget, Styleable {
      * @param results The results. This will never be null.
      */
     List<Node> lookupAll(Selector selector, List<Node> results) {
-        if (selector.applies(this)) {
+        if (selectorMatches(selector)) {
             // Lazily create the set to reduce some trash.
             if (results == null) {
                 results = new LinkedList<>();
@@ -2022,6 +2022,19 @@ public abstract class Node implements EventTarget, Styleable {
         if (getParent() != null) {
             getParent().toFront(this);
         }
+    }
+
+    /**
+     * Checks whether the provided selector matches the node with both styles and pseudo states.
+     * @param s selector to match
+     * @return {@code true} if the selector matches
+     */
+    private boolean selectorMatches(Selector s){
+        boolean matches = s != null && s.applies(this);
+        if(matches && !s.createMatch().getPseudoClasses().isEmpty()){
+            matches = s.stateMatches(this, this.getPseudoClassStates());
+        }
+        return matches;
     }
 
     // TODO: need to verify whether this is OK to do starting from a node in
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -1950,18 +1950,15 @@ public abstract class Node implements EventTarget, Styleable {
      * into the branch until it finds a match. If more than one sub-node matches the
      * specified selector, this function returns the first of them.
      * <p>
-     *     If the lookup selector does not specify a pseudo class, the lookup will ignore pseudo class states;
-     *     it will return the first matching node whether or not it contains pseudo classes.
-     * </p>
+     * If the lookup selector does not specify a pseudo class, the lookup will ignore pseudo class states;
+     * it will return the first matching node whether or not it contains pseudo classes.
      * <p>
-     *     For example, if a Node is given the id of "myId", then the lookup method can
-     *     be used to find this node as follows: {@code scene.lookup("#myId");}.
-     * </p>
+     * For example, if a Node is given the id of "myId", then the lookup method can
+     * be used to find this node as follows: {@code scene.lookup("#myId");}.
      * <p>
-     *     For example, if two nodes, NodeA and NodeB, have the same style class "myStyle" and NodeA has
-     *     a pseudo state "myPseudo", then to find NodeA, the lookup method can be used as follows:
-     *     {@code scene.lookup(".myStyle:myPseudo");} or {@code scene.lookup(":myPseudo");}.
-     * </p>
+     * For example, if two nodes, NodeA and NodeB, have the same style class "myStyle" and NodeA has
+     * a pseudo state "myPseudo", then to find NodeA, the lookup method can be used as follows:
+     * {@code scene.lookup(".myStyle:myPseudo");} or {@code scene.lookup(":myPseudo");}.
      *
      * @param selector The css selector of the node to find
      * @return The first node, starting from this {@code Node}, which matches
@@ -1978,15 +1975,15 @@ public abstract class Node implements EventTarget, Styleable {
      * the given CSS selector. If no matches are found, an empty unmodifiable set is
      * returned. The set is explicitly unordered.
      * <p>
-     *     For example, if there are multiple nodes with same style class "myStyle", then the lookupAll method can
-     *     be used to find all these nodes as follows: {@code scene.lookupAll(".myStyle");}.
-     * </p>
+     * If the lookupAll selector does not specify a pseudo class, the lookupAll will ignore pseudo class states;
+     * it will return all matching nodes whether or not the nodes contain pseudo classes.
      * <p>
-     *     For example, if multiple nodes have same style class "myStyle" and few nodes have
-     *     a pseudo state "myPseudo", then to find all nodes with "myPseudo" state, the lookupAll method can be used as follows:
-     *     {@code scene.lookupAll(".myStyle:myPseudo");} or {@code scene.lookupAll(":myPseudo");}. If no pseudo class is specified
-     *     by the lookupAll selector, irrespective of the nodes pseudo states the result will contain all nodes matching the selector.
-     * </p>
+     * For example, if there are multiple nodes with same style class "myStyle", then the lookupAll method can
+     * be used to find all these nodes as follows: {@code scene.lookupAll(".myStyle");}.
+     * <p>
+     * For example, if multiple nodes have same style class "myStyle" and few nodes have
+     * a pseudo state "myPseudo", then to find all nodes with "myPseudo" state, the lookupAll method can be used as follows:
+     * {@code scene.lookupAll(".myStyle:myPseudo");} or {@code scene.lookupAll(":myPseudo");}.
      *
      * @param selector The css selector of the nodes to find
      * @return All nodes, starting from and including this {@code Node}, which match
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -1982,7 +1982,7 @@ public abstract class Node implements EventTarget, Styleable {
      *     If multiple nodes have same style class "myStyle" and few nodes have
      *     a pseudo state "myPseudo", then to find all nodes with "myPseudo" state, the lookupAll method can be used as follows:
      *     {@code scene.lookupAll(".myStyle:myPseudo");} or {@code scene.lookupAll(":myPseudo");}. If no pseudo class is specified
-     *     by the lookupAll selector, irrespective of their pseudo states the result will contain all nodes matching the selector.
+     *     by the lookupAll selector, irrespective of the nodes pseudo states the result will contain all nodes matching the selector.
      * </p>
      *
      * @param selector The css selector of the nodes to find
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -2030,7 +2030,7 @@ public abstract class Node implements EventTarget, Styleable {
      * @param s selector to match
      * @return {@code true} if the selector matches
      */
-    private boolean selectorMatches(Selector s){
+    private boolean selectorMatches(Selector s) {
         boolean matches = s != null && s.applies(this);
         if(matches && !s.createMatch().getPseudoClasses().isEmpty()){
             matches = s.stateMatches(this, this.getPseudoClassStates());
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -1953,6 +1953,11 @@ public abstract class Node implements EventTarget, Styleable {
      *     For example, if a Node is given the id of "myId", then the lookup method can
      *     be used to find this node as follows: <code>scene.lookup("#myId");</code>.
      * </p>
+     * <p>
+     *     If two nodes, lets say NodeA and NodeB have same style class "myStyle" and NodeA has
+     *     a pseudo state "myPseudo", then to find NodeA, the lookup method can be used as follows:
+     *     <code>scene.lookup(".myStyle:myPseudo");</code> or <code>scene.lookup(":myPseudo");</code>
+     * </p>
      *
      * @param selector The css selector of the node to find
      * @return The first node, starting from this {@code Node}, which matches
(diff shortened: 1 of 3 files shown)


kevinrushforth (changes requested):
This causes many test failures, which suggests that this is not the right fix. And no, the solution is not to change the existing tests.

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_lookup_Test.java
```diff
@@ -89,11 +96,40 @@ public class Node_lookup_Test {
...
+        assertTrue(nodes.contains(ac));
 
         nodes = root.lookupAll(".d");
         assertEquals(2, nodes.size());
         assertTrue(nodes.contains(d));
         assertTrue(nodes.contains(d2));
     }
+
+    @Test
+    public void lookupPsuedoTest(){
```
andy-goryachev-oracle, line 114: spelling: lookupPseudoTest
(also in the bug and PR title)

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -1990,7 +1990,7 @@ public Set<Node> lookupAll(String selector) {
      * @param results The results. This will never be null.
      */
     List<Node> lookupAll(Selector selector, List<Node> results) {
-        if (selector.applies(this)) {
+        if (selectorMatches(selector)) {
```
andy-goryachev-oracle, line 1993: line 1990 might be incorrect: the return value **can** be null.
We probably should just correct the javadoc, and since it's a private API no CSR is needed.  All callers of this method do check for null.

andy-goryachev-oracle:
Another question: this PR changes the behavior of Node.lookup() and Node.lookupAll() in respect to pseudo classes.
Now, wouldn't that pose a regression risk for applications?
Should we create a new method with the new semantics instead (+ boolean considerPseudoClasses or some such)?

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_lookup_Test.java
SaiPradeepDandem: Fixed the spelling issue in the class. However, as I have no permissions to edit the bug title, I cannot change the PR title till it is corrected by some one.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
hjohn: I must say I'm again baffled at how this is implemented.  

Red flag one: it uses a `LinkedList` which are known to be rarely the right choice, and I doubt this case is any different.  

Red flag two: a `List` is converted to a `Set`; being backed by a `LinkedList` means set type operations will be terribly slow if that list has any kind of size.  

Red flag three: The choice to return a `Set` in the public API seems to be only motivated to indicate there won't be any duplicates, which is a terrible reason.

Red flag four: `UnmodifiableListSet` talks about insertion speed and hashing overhead, yet uses a `LinkedList` which are slower than `ArrayList` when it comes to insertion speed (not to mention that they require more object allocations and more memory(!)).

andy-goryachev-oracle: You are right, @hjohn - LinkedList seems like a bad choice!  It should be HashSet from the very beginning, shouldn't it?

But Set<Node> as a rerun value for lookups is correct, I think.

We could (should) fix it in a separate PR.

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_lookup_Test.java
```diff
@@ -96,4 +109,41 @@ public class Node_lookup_Test {
         assertTrue(nodes.contains(d));
         assertTrue(nodes.contains(d2));
     }
+
+    @Test
+    public void lookupPseudoTest(){
```
andy-goryachev-oracle, line 114: please insert space
```
public void lookupPseudoTest() {
```

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_lookup_Test.java
```diff
@@ -96,4 +109,41 @@ public class Node_lookup_Test {
...
+
+        nodes = root.lookupAll(".g:testPseudo1");
+        assertEquals(2, nodes.size());
+        assertTrue(nodes.contains(g));
+        assertTrue(nodes.contains(hg));
+
+        nodes = root.lookupAll(":testPseudo1");
+        assertEquals(2, nodes.size());
+        assertTrue(nodes.contains(g));
+        assertTrue(nodes.contains(hg));
```
andy-goryachev-oracle, line 136: I'd add one more test case to verify lack of regression when no pseudo class is specified.  May be something like this (feel free to expand, add negative cases etc.):

```
    /**
     * Verifies that the lookup ignores pseudo classes when selector contains no explicit pseudo class.
     */
    @Test
    public void lookupPseudoTest2() {
        Group root = new Group();
        
        Group ab = new Group();
        ab.getStyleClass().addAll("a", "b");
        
        Group a = new Group();
        a.getStyleClass().addAll("a");
        
        Group a1 = new Group();
        a1.getStyleClass().addAll("a");
        a1.pseudoClassStateChanged(PseudoClass.getPseudoClass("P1"), true);
        
        ParentShim.getChildren(root).addAll(a, a1, ab);
        
        Set<Node> rv;
        
        rv = root.lookupAll(".a");
        assertTrue(rv.contains(a));
        assertTrue(rv.contains(a1));
        
        rv = root.lookupAll(".a:P1");
        assertTrue(rv.contains(a1));
    }
```

andy-goryachev-oracle (commented):
On second though, I think this change will not cause regression: when no pseudo class is specified in the lookup selector, the lookup ignores the pseudo classes as before.

What I strongly recommend is to clarify the behavior of lookup() and lookupAll() in their javadoc (which might require a CSR) with respect to pseudo classes.

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_lookup_Test.java
SaiPradeepDandem: Added another test case to cover the scenario you mentioned. I used the same nodes (have pseudo classes set)  which are used in previous test case. Tried to cover all possible scenarios to check for regression.

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_lookup_Test.java
SaiPradeepDandem: @andy-goryachev-oracle  Thanks for correcting the bug title. Corrected the PR title now.

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_lookup_Test.java
```diff
@@ -96,4 +109,89 @@ public class Node_lookup_Test {
...
+
+        nodes = root.lookupAll(".f:randomPseudo");
+        assertEquals(0, nodes.size());
+    }
+
+    /**
+     * Verifies that the lookup ignores pseudo classes when selector contains no explicit pseudo class.
+     */
+    @Test
+    public void lookupPseudoTest2() {
```
andy-goryachev-oracle, line 154: the reason I've used different nodes is because the situation you have with existing nodes might give false positive.

I wanted to make sure that selector "a." will collect nodes with ".a" and ".a:pseudo", but in the standard set of nodes with g and hg nodes - but they have the same :testPseudo1, so it's not exactly equivalent.

The other option is to add new nodes, one with ".x" and another with ".x:pseudo"
If you do that, it might be easier to change the javadoc format to do something like this:

```
 └─ Pane
     ├─ VFlow (Pane) .flow
     │   ├─ ClippedPane (Pane) .content
```
(but it's just a very minor suggestion)

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_lookup_Test.java
```diff
@@ -194,4 +194,47 @@ public void lookupPseudoTest2() {
...
+        assertTrue(nodes.contains(x));
+        assertTrue(nodes.contains(x1));
+        assertTrue(nodes.contains(xy));
+
+        nodes = root.lookupAll("#root .x:pseudo");
+        assertTrue(nodes.contains(x1));
+
+        nodes = root.lookupAll(".x:random");
+        assertEquals(0, nodes.size());
+
```
andy-goryachev-oracle, line 238: excellent!  thank you.
could you please remove the extra newline before } here ?

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_lookup_Test.java
SaiPradeepDandem: Thanks for the clarification. Now I got the point what you are saying about.
Added a new test case to cover this scenario. Used x, y.. naming to make a clear separation from previous scenarios.

However do you still prefer to keep the lookupPseudoTest2() scenarios? This is almost similar to quickTest() and lookupAllTest() except that all the nodes used in this method have pseudo classes set.

andy-goryachev-oracle: we could keep it - the more the merrier.

ideally, we could use a more thorough approach: let's say, use two or three layers, and create as many nodes as there are combinations, to test every combination of the selector and search result.

but the main result we got is that there should be no regression w.r.t. the old implementation (I think).
two more people should definitely take a look at this PR still.

SaiPradeepDandem: Ok. Thank you :)

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -2024,6 +2025,19 @@ public void toFront() {
...
     }
 
+    /**
+     * Checks whether the provided selector matches the node with both styles and pseudo states.
+     * @param s selector to match
+     * @return {@code true} if the selector matches
+     */
+    private boolean selectorMatches(Selector s) {
+        boolean matches = s != null && s.applies(this);
+        if(matches && !s.createMatch().getPseudoClasses().isEmpty()){
```
mstr2, line 2035: Please add a space after `if` and before `{`.

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_lookup_Test.java
```diff
@@ -58,9 +60,20 @@ public class Node_lookup_Test {
         bc.getStyleClass().addAll("b", "c");
         d2 = new Group();
         d2.getStyleClass().add("d");
-        ParentShim.getChildren(root).addAll(a, bc);
+        f = new Group();
+        f.getStyleClass().add("f");
+        f.pseudoClassStateChanged(PseudoClass.getPseudoClass("testPseudo"),true);
```
mstr2, line 65: Can you add a space after the comma in this line and other lines after that?

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_lookup_Test.java
```diff
@@ -96,4 +109,131 @@ public class Node_lookup_Test {
...
+
+        nodes = root.lookupAll(".random .h");
+        assertEquals(0, nodes.size());
+    }
+
+    /**
+     * Verifies that the lookup ignores pseudo classes when selector contains no explicit pseudo class.
+     */
+    @Test
+    public void lookupPseudoTest3() {
```
mstr2, line 202: Consider naming the test methods according to what they test.

SaiPradeepDandem: Renamed the methods by referring to method names used in other classes. I noticed that descriptive long method names is accepted, so named similarly. The only thing I want to check is: I renamed existing methods also. Is it accepted? or I am not allowed to rename others method names?

hjohn (commented):
Looks good, although I do think this needs to be documented on the public methods (it should not be a separate PR).

andy-goryachev-oracle:
I agree with @hjohn : it would be nice to clarify the behavior of Node.lookup() and Node.lookupAll() in regards to pseudo classes.

I think it will not require a CSR since it may qualify as a clarification.

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_lookup_Test.java
andy-goryachev-oracle: I see no problem renaming: there are no dependencies.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
andy-goryachev-oracle: created * [JDK-8318842](https://bugs.openjdk.org/browse/JDK-8318842) Node.lookupAll to use Set instead of List (Enhancement)

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -1961,13 +1966,22 @@ private void updateDisabled() {
...
      * returned. The set is explicitly unordered.
+     * <p>
+     *     For example, if there are multiple nodes with same style class "myStyle", then the lookupAll method can
+     *     be used to find all these nodes as follows: <code>scene.lookupAll(".myStyle");</code>.
+     * </p>
+     * <p>
+     *     If multiple nodes have same style class "myStyle" and few nodes have
+     *     a pseudo state "myPseudo", then to find all nodes with "myPseudo" state, the lookupAll method can be used as follows:
+     *     <code>scene.lookupAll(".myStyle:myPseudo");</code> or <code>scene.lookupAll(":myPseudo");</code>
+     * </p>
```
andy-goryachev-oracle, line 1984: Added explanation is very good!  I would add one more thing - that if no pseudo class is specified by the lookup selector, the result will contain nodes with pseudo classes (that is, pseudo classes are ignored).

Minor note: should we be using {@code ... } instead of < code > ?

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -1986,11 +2000,12 @@ public Set<Node> lookupAll(String selector) {
      * Used by Node and Parent for traversing the tree and adding all nodes which
      * match the given selector.
      *
-     * @param selector The Selector. This will never be null.
-     * @param results The results. This will never be null.
+     * @param selector The Selector.
+     * @param results The results.
+     * @return List of matching nodes. The returned value can be null.
```
andy-goryachev-oracle, line 2005: minor: I don't think we should capitalize text in @ param and @ return

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -1951,7 +1951,13 @@ private void updateDisabled() {
      * specified selector, this function returns the first of them.
      * <p>
      *     For example, if a Node is given the id of "myId", then the lookup method can
-     *     be used to find this node as follows: <code>scene.lookup("#myId");</code>.
+     *     be used to find this node as follows: {@code scene.lookup("#myId");}.
+     * </p>
+     * <p>
+     *     If two nodes, lets say NodeA and NodeB have same style class "myStyle" and NodeA has
```
kevinrushforth, line 1957: Since this is a separate paragraph, please add "For example, " here to make it clear that this is an example (and not normative specification). Maybe something like this:

```
     * For example, if two nodes, NodeA and NodeB, have the same ...
```

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
kevinrushforth: The second example is good. I echo my above comments about adding "For example, ".

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
kevinrushforth: Agreed. Please change "The" and "List" to lower case.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -1951,7 +1951,13 @@ private void updateDisabled() {
...
+     *     a pseudo state "myPseudo", then to find NodeA, the lookup method can be used as follows:
+     *     {@code scene.lookup(".myStyle:myPseudo");} or {@code scene.lookup(":myPseudo");}. If no pseudo class is specified
```
kevinrushforth, line 1959: The sentence about what happens when no pseudo class is specified is true in general and not really part of this example. I recommend moving it before the two examples. And I think it could be worded more clearly.

```
    * If the lookup selector does not specify a pseudo class, the lookup will ignore pseudo class
    *  states; it will return the first matching node even if it contains pseudo classes.
```

kevinrushforth (commented):
I left a couple comments on the API docs.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
kevinrushforth: Or possibly even more clear:

```
    * If the lookup selector does not specify a pseudo class, the lookup will ignore pseudo class
    * states; it will return the first matching node whether or not it contains pseudo classes.
```

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -1950,8 +1950,17 @@ private void updateDisabled() {
      * into the branch until it finds a match. If more than one sub-node matches the
      * specified selector, this function returns the first of them.
      * <p>
+     *     If the lookup selector does not specify a pseudo class, the lookup will ignore pseudo class states;
+     *     it will return the first matching node whether or not it contains pseudo classes.
+     * </p>
+     * <p>
```
mstr2, line 1956: You don't need to surround a paragraph with `<p>...</p>`, a single `<p>` (without a corresponding `</p>`) between two paragraphs is sufficient. This reduces nesting, as it keeps the subsequent paragraphs on the same indentation.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -1961,13 +1970,23 @@ private void updateDisabled() {
...
      * returned. The set is explicitly unordered.
+     * <p>
+     *     For example, if there are multiple nodes with same style class "myStyle", then the lookupAll method can
+     *     be used to find all these nodes as follows: {@code scene.lookupAll(".myStyle");}.
+     * </p>
+     * <p>
+     *     For example, if multiple nodes have same style class "myStyle" and few nodes have
+     *     a pseudo state "myPseudo", then to find all nodes with "myPseudo" state, the lookupAll method can be used as follows:
+     *     {@code scene.lookupAll(".myStyle:myPseudo");} or {@code scene.lookupAll(":myPseudo");}. If no pseudo class is specified
+     *     by the lookupAll selector, irrespective of the nodes pseudo states the result will contain all nodes matching the selector.
```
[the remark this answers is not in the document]
SaiPradeepDandem: Sorry overlooked that... done now.

kevinrushforth (approved):
The updated docs look good. I also verified that the provided test fails without the fix and passes with the fix.

NOTE: There is a Skara bug that causes the state of `/reviewers` command to be ignored. Do not integrate this until there are three (3) approving reviews.

hjohn (approved):
Looks good, no comments

kevinrushforth:
@SaiPradeepDandem This is now OK to integrate.

SaiPradeepDandem:
> @SaiPradeepDandem This is now OK to integrate.

@kevinrushforth  thanks for confirming. But I am pretty not sure what steps I need to do next. Is it that I need to type a new comment with /integrate ? as mentioned in 

`To flag this PR as ready for integration with the above commit message, type /integrate in a new comment. (Afterwards, your sponsor types /sponsor in a new comment to perform the integration).`

andy-goryachev-oracle:
yes, just type slash-integrate command

PR #1246, opened 2023-09-21
JDK-8316590: Rendering artifact after JDK-8311983

Maran23:
Fixes the regression by basically reverting one change introduced in https://bugs.openjdk.org/browse/JDK-8311983.
The problem is that it is actually required to get the size from a cell with the index -1, which technically does not exist (the accumCell is used then).

Furthermore, unlike the name suggests, the call to `addLeadingCells()` is always needed, even if there are none.
This is because the method does much more than what you would think first. 

In future, it would probably be good to revisit this code. Also performance wise, since this looks like something that can be optimized. But that is another story, not related to this fix.

---------
### Progress

### Issue
 * [JDK-8316590](https://bugs.openjdk.org/browse/JDK-8316590): Rendering artifact after JDK-8311983 (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1246/head:pull/1246` \
`$ git checkout pull/1246`

`$ git checkout pull/1246` \
`$ git pull https://git.openjdk.org/jfx.git pull/1246/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1246`

`$ git pr show -t 1246`

</details>

<a href="https://git.openjdk.org/jfx/pull/1246.diff">https://git.openjdk.org/jfx/pull/1246.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1246#issuecomment-1729711788)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
@@ -1671,11 +1671,10 @@ public class VirtualFlow<T extends IndexedCell> extends Region {
 
             // Add any necessary leading cells
             if (firstCell != null) {
-                int firstIndex = getCellIndex(firstCell);
-                int index = Math.max(0, firstIndex - 1);
-
-                double prevIndexSize = getCellLength(index);
-                addLeadingCells(index, getCellPosition(firstCell) - prevIndexSize);
+                // NOTE: The index might be -1, but the call to addLeadingCells() is still required.
+                int previousIndex = getCellIndex(firstCell) - 1;
+                double prevIndexSize = getCellLength(previousIndex);
+                addLeadingCells(previousIndex, getCellPosition(firstCell) - prevIndexSize);
             } else {
                 int currentIndex = computeCurrentIndex();
 
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/VirtualFlowTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/VirtualFlowTest.java
@@ -1838,6 +1838,29 @@ assertEquals(0, firstCell.getIndex());
         assertEquals(3, flow.getFirstVisibleCell().getIndex());
     }
 
+    /**
+     * The first cell should always be the same when scrolling down just a little bit.
+     * This is mainly a regression test to check that no leading cells are added where they should not be.
+     *
+     * @see <a href="https://bugs.openjdk.org/browse/JDK-8316590">JDK-8316590</a>
+     */
+    @Test
+    public void testFirstCellShouldBeTheSameOnScroll() {
+        flow = new VirtualFlowShim<>();
+        flow.setCellFactory(fw -> new CellStub(flow));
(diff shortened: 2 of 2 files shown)


kevinrushforth:
Reviewers: @andy-goryachev-oracle @karthikpandelu 

/reviewers 2

kevinrushforth:
@johanvos You also might want to take a look.

johanvos:
The PR indeed fixes a regression caused by the fix for PR-8311983. Calling addLeadingCells with a wrong start index (the index of the cell before the top-one) and a wrong offset causes the issue you observed. I agree that the name "addLeadingCells" is a bit confusing, as is the name "index". 
As for the test: I understand it will detect issues with a new cell being reassigned to content that should still be owned by its original cell. However, that doesn't have to be theoretically wrong. There is no guarantee in the VirtualFlow API about order and shuffles in the cells it is using internally, so a change in internal implementation might cause this test to fail.
It might be better to check for updateIndex() to be called on the Cell implementation, and fail if unexpectedly an item is assigned to a cell that is drawn at a wrong position.

The latter is what you can observe with the current issue: 
inside `addLeadingCells`, the first `while` loop will be executed once (if you pass the wrong index), leading to this:
```
    cell = getAvailableCell(index);
    setCellIndex(cell, index);
...
    positionCell(cell, offset);

```
Hence, the result is that the item at "index" 0 is positioned at the wrong offset, which is something you can detect with a test.

Maran23:
> Hence, the result is that the item at "index" 0 is positioned at the wrong offset, which is something you can detect with a test.

I will check it out. More tests are always better nonetheless.
I still think the existing test is good as a regression test. If the internal implementation changes in a way that this test fail, we can adjust it. Right now we would not expect that the first cell changes when scrolling a bit, but the first cell is still visible (and I dont see a reason this will change any time soon).

johanvos:
Thanks, those tests look good (although the assert in the positiontest could be more relaxed, we are ok if the position is -0.01 but not when it is -20). 
I agree the first test is a good regression test too. 

One of the difficulties in this area is that the position of the cell is not always the value we ask it to be. e.g. if we want it to be at -0.01, its layout will be at 0 (rounded in the layout phase) and subsequently the value of getPosition will be 0 as well, which makes it hard to be consistent during calculations (asking what cell is at -0.01 will give a different result than asking what cell is at 0 for example).
I agree that there are many optimizations possible, and some of them are very easy. It is hard though to prove that even the simplest optimization is not causing any regression at all. Therefore, I think there is a huge value in PR's like this because they add tests.

johanvos (approved):
This fixes the regression from JDK-8311983 and adds a few new regression tests.

Maran23:
> Therefore, I think there is a huge value in PR's like this because they add tests.

I agree, more tests also means we can more confidently implement potential performance improvements in the future, as well as catch any regression when changing the logic in the `VirtualFlow`.

karthikpandelu (approved):
The changes LGTM. It fixes the regression.

andy-goryachev-oracle (approved):
🕶 looking good

PR #1252, opened 2023-10-02
JDK-8284445: macOS 12 prints a warning when a function key shortcut is assigned to a menu

vlaaad:
This PR addresses [JDK-8284445](https://bugs.openjdk.org/browse/JDK-8284445).

Context: the warning might be more important than it seems. Our JavaFX-based desktop app has [an issue](https://github.com/defold/defold/issues/7845) — when the user opens 2 instances of the app, one of the instances hangs after a while. While debugging the issue, I found that it's somehow related to the system menu bar — not refreshing it fixes the issue. Further investigations revealed that the app hangs only if we show menus that use F1-F24 shortcuts. This reminded me of the warnings we get about `NSEventModifierFlagFunction specified to -setKeyEquivalentModifierMask` for such shortcuts. I experimented with removing the modifier for these shortcuts, and it helped with our issue!

---------
### Progress

### Issue
 * [JDK-8284445](https://bugs.openjdk.org/browse/JDK-8284445): macOS 12 prints a warning when a function key shortcut is assigned to a menu (**Bug** - P4)

### Reviewers
 * [Martin Fox](https://openjdk.org/census#mfox) (@beldenfox - Author)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1252/head:pull/1252` \
`$ git checkout pull/1252`

`$ git checkout pull/1252` \
`$ git pull https://git.openjdk.org/jfx.git pull/1252/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1252`

`$ git pr show -t 1252`

</details>

<a href="https://git.openjdk.org/jfx/pull/1252.diff">https://git.openjdk.org/jfx/pull/1252.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1252#issuecomment-1773016067)

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassMenu.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassMenu.m
@@ -236,16 +236,17 @@ static jfieldID  jPixelsScaleYField = 0;
     }
     if ((jmodifiers & com_sun_glass_events_KeyEvent_MODIFIER_FUNCTION) != 0)
     {
-        modifier = modifier | NSFunctionKeyMask;
-    }
-    if (jshortcut >= com_sun_glass_events_KeyEvent_VK_F1 &&
-        jshortcut <= com_sun_glass_events_KeyEvent_VK_F12) {
-        int delta = jshortcut - com_sun_glass_events_KeyEvent_VK_F1;
-        shortcut = [NSString stringWithFormat:@"%C", (unsigned short)(NSF1FunctionKey + delta)];
-    } else if (jshortcut >= com_sun_glass_events_KeyEvent_VK_F13 &&
-                jshortcut <= com_sun_glass_events_KeyEvent_VK_F24) {
-        int delta = jshortcut - com_sun_glass_events_KeyEvent_VK_F13;
-        shortcut = [NSString stringWithFormat:@"%C", (unsigned short)(NSF13FunctionKey + delta)];
+        if (jshortcut >= com_sun_glass_events_KeyEvent_VK_F1 &&
+            jshortcut <= com_sun_glass_events_KeyEvent_VK_F12) {
+            int delta = jshortcut - com_sun_glass_events_KeyEvent_VK_F1;
+            shortcut = [NSString stringWithFormat:@"%C", (unsigned short)(NSF1FunctionKey + delta)];
+        } else if (jshortcut >= com_sun_glass_events_KeyEvent_VK_F13 &&
+                   jshortcut <= com_sun_glass_events_KeyEvent_VK_F24) {
+            int delta = jshortcut - com_sun_glass_events_KeyEvent_VK_F13;
+            shortcut = [NSString stringWithFormat:@"%C", (unsigned short)(NSF13FunctionKey + delta)];
+        } else {
+            modifier = modifier | NSFunctionKeyMask;
+        }
     }
     [self->item setKeyEquivalent:shortcut];
     [self->item setKeyEquivalentModifierMask:modifier];
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassMenu.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassMenu.m
@@ -237,15 +237,15 @@ static jfieldID  jPixelsScaleYField = 0;
     if ((jmodifiers & com_sun_glass_events_KeyEvent_MODIFIER_FUNCTION) != 0)
     {
         modifier = modifier | NSFunctionKeyMask;
-        if (jshortcut >= com_sun_glass_events_KeyEvent_VK_F1 &&
-            jshortcut <= com_sun_glass_events_KeyEvent_VK_F12) {
-            int delta = jshortcut - com_sun_glass_events_KeyEvent_VK_F1;
-            shortcut = [NSString stringWithFormat:@"%C", (unsigned short)(NSF1FunctionKey + delta)];
-        } else if (jshortcut >= com_sun_glass_events_KeyEvent_VK_F13 &&
-                   jshortcut <= com_sun_glass_events_KeyEvent_VK_F24) {
-            int delta = jshortcut - com_sun_glass_events_KeyEvent_VK_F13;
-            shortcut = [NSString stringWithFormat:@"%C", (unsigned short)(NSF13FunctionKey + delta)];
-        }
+    }
+    if (jshortcut >= com_sun_glass_events_KeyEvent_VK_F1 &&
+        jshortcut <= com_sun_glass_events_KeyEvent_VK_F12) {
+        int delta = jshortcut - com_sun_glass_events_KeyEvent_VK_F1;
+        shortcut = [NSString stringWithFormat:@"%C", (unsigned short)(NSF1FunctionKey + delta)];
+    } else if (jshortcut >= com_sun_glass_events_KeyEvent_VK_F13 &&
+                jshortcut <= com_sun_glass_events_KeyEvent_VK_F24) {
+        int delta = jshortcut - com_sun_glass_events_KeyEvent_VK_F13;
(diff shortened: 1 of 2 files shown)


vlaaad:
Hey @kevinrushforth, it's been 2 weeks, and my OCA has still not been reviewed. Any chance you can check/affect this process? This is an important issue for our app 🙏

vlaaad:
Hello? @andy-goryachev-oracle maybe you can make the process going?

andy-goryachev-oracle:
@vlaaad it might just need more time, but I'll inquire.

beldenfox:
I will take a look.

beldenfox:
I can't officially approve this PR but I can save other reviewers a grueling five minutes of research.

MODIFIER_FUNCTION is barely mentioned in the code (and I mean the entire codebase, tests and all). When processing a key event the Mac code maps NSFunctionKeyMask to MODIFIER_FUNCTION even though it's not really a modifier (it is unconditionally set on certain keys). The Java side just ignores it.

The only other usage is in these two files. The Mac code shouldn't be setting NSFunctionKeyMask on the key equivalent because the function key is not a modifier. There's no point in the Java code setting this flag since the Mac is the only client. So all usage of MODIFIER_FUNCTION should just go away.

beldenfox:
With this change the Quit menu item shows up as Cmd+F2.

To fix the bug we just need to stop setting the NSFunctionKeyMask in the equivalent's modifier mask. The rest of the code removed in this PR needs to be restored.

When the core code is converting from Java internals to the data structure passed to Glass it only has one field to specify the shortcut. Normally this is a character but for keys like F2 it passes down a Java KeyCode. To signal the difference it sets the MODIFIER_FUNCTION flag. Otherwise it can't distinguish KeyCode F2 from character Q.

The naming of this flag is confusing since it's not a modifier but a bit used solely for this purpose. And I don't know why this flag is set only for the function keys, it should be set whenever the core code is passing down a KeyCode rather than a character. This is probably part of what's going wrong in [JDK-8088445](https://bugs.openjdk.org/browse/JDK-8088445).

vlaaad:
Thanks for reviewing the changes @beldenfox, I applied your suggestions (if I understood them correctly). Using the new changeset, I see the <kbd>Cmd+Q</kbd> shortcut displayed as <kbd>Cmd+Q</kbd>.

modules/javafx.graphics/src/main/native-glass/mac/GlassMenu.m
```diff
@@ -245,6 +244,8 @@ - (void)_setShortcut:(jchar)jshortcut modifiers:(jint)jmodifiers
                    jshortcut <= com_sun_glass_events_KeyEvent_VK_F24) {
             int delta = jshortcut - com_sun_glass_events_KeyEvent_VK_F13;
             shortcut = [NSString stringWithFormat:@"%C", (unsigned short)(NSF13FunctionKey + delta)];
+        } else {
```
beldenfox, line 247: This else clause can be removed. JavaFX should never set the NSFunctionKeyMask so this PR can be just a one-line change (remove line 239).

Sorry I wasn't clearer about this earlier. I was basically diagnosing a different bug in the same area of code. Not your concern.

vlaaad: A-ha, now I understand, thanks for the clarification!

vlaaad:
@beldenfox it's ready for review again :)

beldenfox (approved):
Verified that the OS warning message is gone and system menus still work as expected.

arapte:
When a function key is pressed [NSFunctionKeyMask](https://developer.apple.com/documentation/appkit/nsfunctionkeymask) is automatically included by the system.
Hence it looks not breaking behavior with this change.
As the warning is quite clear it looks logical to not include the modifier.
Tested on macOS 12.6.8, warning appears without this change and disappears with this change.

Can you please check _KeyCombinationTest.java_ test file and try to add a unit test, at least a mac specific test, which should press a Function key and check that modifiers in the received event contains Function modifier flag.

beldenfox:
The MODIFIER_FUNCTION constant is poorly named and in the wrong place since it has nothing to do with key events. There is no Function modifier in Java. The Mac glass code sets this flag on certain key events but Java just ignores it.

The only use for this flag is setting up the system menu bar. We don't have an existing test to verify that the system menu bar processes key events correctly (that would require a Robot and I don't see any Robot tests for this). I will enter a follow-up bug to fix that testing gap. For now I think it's enough that the warning is gone.

vlaaad:
At least in JavaFX it seems `Fn` modifier is ignored, e.g. pressing `Fn S` reports a key event indistinguishable from `S`.

vlaaad:
@arapte anything else I could do?

arapte:
@vlaaad 
Yes please, I think we should have a test along with this fix. A test is missing for this scenario.
If a unit test is not feasible, can you please check to add a system test ? ( may be something like setting F1 as accelerator key on a MenuItem results in MenuItem's action getting fired when Function+F1 is pressed )

beldenfox:
@arapte The best we can do when it comes to the Function keys is a manual test since JavaFX can't manipulate the fn key. I think I have a manual test I can clean up to use here.

beldenfox:
I added a manual test to the original bug report. @arapte I would prefer to approve this PR and submit the manual test under a separate PR since it's by a different author. FWIW it works for me but you may see different results, the function keys are overloaded and the OS or some other app you're running may grab one of these shortcuts. That's why this test avoids Cmd-F5, on Ventura that toggles Voice Over.

This variability is the primary reason I don't want to enter an automated Robot test. Perhaps with some experience we can find a subset of the function keys that work reliably for everyone but I don't know what that subset is now.

arapte:
@beldenfox 
Sounds good, I looked at a the test.
The test will need some changes to adapt to other platforms, otherwise looks good.

arapte (approved):

vlaaad:
Yay! Is there any ETA for a new javafx version with this changeset?

beldenfox:
@vlaaad The SKARA bot hasn't marked this as ready. I suspect you need to turn on pre-submit tests in your repo. See the Details link next to the line where it says pre-submit tests were skipped.

Next version of JavaFX is slated for release in March 2024.

kevinrushforth:
> YIs there any ETA for a new javafx version with this changeset?

First you need to integrate the fix so that it will get into the source repo. I see that Skara didn't "count" the review of @beldenfox (he is an Author in the OpenJFX project, although Skara doesn't recognize him as such.). Given this, I'll lower the reviewer criteria for this PR to 1 and attempt to manually add him as a reviewer so that it shows up in the commit metadata.

/reviewer credit mfox

kevinrushforth:
I should add that, once integrated it will be in the next weekly early access build of JavaFX 22.

kevinrushforth:
> @vlaaad The SKARA bot hasn't marked this as ready.

This is because Skara doesn't know that @beldenfox is the same as OpenJDK user `mfox`. To allow Skara to recognize you, follow the instruction on the Skara Wiki to [associate your GitHub account and OpenJDK username](https://wiki.openjdk.org/display/SKARA#Skara-AssociatingyourGitHubaccountandyourOpenJDKusername).

> I suspect you need to turn on pre-submit tests in your repo. See the Details link next to the line where it says pre-submit tests were skipped.

This is a good idea in any case. Skara doesn't care, but it is helpful for reviewers.

beldenfox:
I've filed a SKARA issue [SKARA-2100](https://bugs.openjdk.org/browse/SKARA-2100)

kevinrushforth:
> I've filed a SKARA issue [SKARA-2100](https://bugs.openjdk.org/browse/SKARA-2100)

This has been completed, so Skara will now recognize you. I don't know whether you will need to re-review it, but we'll know shortly.

/reviewers 2

kevinrushforth:
@vlaaad This PR is now ready for you to `/integrate`

PR #1253, opened 2023-10-02
8317370: JavaFX runtime version is wrong at runtime

jgneff:
Please review these changes to the build timestamp and to the JavaFX runtime version available in the system properties.

---------
### Progress

### Issue
 * [JDK-8317370](https://bugs.openjdk.org/browse/JDK-8317370): JavaFX runtime version is wrong at runtime (**Bug** - P3)

### Reviewers
 * [Joeri Sykora](https://openjdk.org/census#sykora) (@tiainen - Author)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1253/head:pull/1253` \
`$ git checkout pull/1253`

`$ git checkout pull/1253` \
`$ git pull https://git.openjdk.org/jfx.git pull/1253/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1253`

`$ git pr show -t 1253`

</details>

<a href="https://git.openjdk.org/jfx/pull/1253.diff">https://git.openjdk.org/jfx/pull/1253.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1253#issuecomment-1746945553)

### DIFF
--- a/UPDATING-VERSION.md
+++ b/UPDATING-VERSION.md
@@ -24,7 +24,7 @@ feature version number from `N` to `N+1`:
 
 * In
 `modules/javafx.base/src/test/java/test/com/sun/javafx/runtime/VersionInfoTest.java`,
-modify the testMajorVersion method to increment the feature version number
+modify the `FEATURE` variable to increment the feature version number
 from `N` to `N+1`.
 
 ## Incrementing the security version
--- a/build.gradle
+++ b/build.gradle
@@ -638,19 +638,20 @@ if (sourceDateEpoch != null) {
     def epochSeconds = Long.parseLong(sourceDateEpoch)
     buildInstant = Instant.ofEpochSecond(epochSeconds)
 }
-// Creates the timestamp in UTC using the ISO 8601 extended format
-def buildTimestamp = buildInstant.toString()
+// Creates the timestamp in UTC using the ISO 8601 extended format.
+def extendedTimestamp = buildInstant.toString()
+// Creates the timestamp in UTC using the historical ad hoc format, which is
+// valid for the OPT field of the version string. An alternative to the ad hoc
+// format is the ISO 8601 basic format with the pattern "yyyyMMdd'T'HHmmssX".
+def zonedTime = ZonedDateTime.ofInstant(buildInstant, ZoneOffset.UTC)
+def formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd-HHmmss")
+def buildTimestamp = zonedTime.format(formatter)
 defineProperty("BUILD_TIMESTAMP", buildTimestamp)
 def relSuffix = ""
 def relOpt = ""
 if (HUDSON_JOB_NAME == "not_hudson") {
-    // The version OPT field matches the regular expression "([-a-zA-Z0-9.]+)".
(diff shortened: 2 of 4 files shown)


kevinrushforth:
I'll take a closer look, and also test this, when I can. This looks like the right approach to me. It should not cause any problems. with either the earlier fix for [JDK-8317308](https://bugs.openjdk.org/browse/JDK-8317308) nor with a somewhat similar fix I did in our (outside the repo) CI build scripts a couple months ago, but it makes both of those fixes unnecessary.

/reviewers 2

jgneff:
This pull request makes three changes:

1. The ISO 8601 extended format of the build date and time is now used only in the `--date` option of the `jmod` command, as required. The historical custom format of `"yyyy-MM-dd-HHmmss"` is used everywhere else for the build timestamp.

2. The build file's short and long version strings, `RELEASE_VERSION_SHORT` and `RELEASE_VERSION_LONG`, are now added directly to the `VersionInfo.java` template file for use as the run-time values of the system properties `javafx.version` and `javafx.runtime.version`. This change prevents the values from getting out of sync again between their build-time and run-time values..

3. Four test cases are added to `VersionInfoTest.java`. Two of the tests verify that the JavaFX version and run-time version have the correct format. The other two tests verify that their values at run time are equal to their values in the static file `javafx.properties` created at build time and found in the JavaFX SDK and in the `javafx.base` modular JAR file. These test cases prevent this error from happening again.

jgneff:
@FlorianKirmaier Florian, I would appreciate any comments you might have, too, or even a formal review if you have the time. Don't feel obligated, though, as you've already helped enough by discovering and patching the bug. :trophy:

kevinrushforth:
@arapte or @johanvos Can one of you also review this?

johanvos:
@tiainen can you review?

jgneff:
Thanks, Kevin. I'll fix that.

To all reviewers: I added tests of the method [VersionInfo.setupSystemProperties][1] to the test cases in my [personal repository][2]. It's kind of overkill, and not necessary to prevent this bug, but let me know if I should add them anyway.

[1]: https://github.com/openjdk/jfx/blob/master/modules/javafx.base/src/main/version-info/VersionInfo.java#L152
[2]: https://github.com/jgneff/version-info-test/commit/573f559

kevinrushforth:
> To all reviewers: I added tests of the method [VersionInfo.setupSystemProperties](https://github.com/openjdk/jfx/blob/master/modules/javafx.base/src/main/version-info/VersionInfo.java#L152) to the test cases in my [personal repository](https://github.com/jgneff/version-info-test/commit/573f559). It's kind of overkill, and not necessary to prevent this bug, but let me know if I should add them anyway.

Yeah, I probably wouldn't add them.

kevinrushforth (approved):
Looks good.

FlorianKirmaier:
Based on the changes, it looks great - fewer variables, less code, more tests.
It's now also easier to follow, what the "no_hudson" flag does.

The ":" is now also removed from the version String - so my previous change would no longer be required.
But I guess it doesn't hurt either.

I tested making a build, and it works well!

So from my point of view, it looks quite great!

tiainen (approved):
Tested on our build infrastructure and build and tests all worked fine.

PR #1258, opened 2023-10-12
8316419: [macos] Setting X/Y makes Stage maximization not work before show

beldenfox:
When a window is visible the maximized, iconified, and fullscreen properties are two-way streets; changes made in Java are sent on to the platform window and changes in the platform window are sent back into Java.

When a window is hidden these properties (and others, like location and sizing information) are not sent on to the platform window until the window is made visible. In other words, the properties don't reflect the actual state of the window but the intended state after it's shown.

There's a period when the window is transitioning from hidden to shown where the core Stage code is using the stored properties to configure the platform window and the platform window is simultaneously calling back in to update the properties. This was causing the intended state to be wiped out before it could be sent on to the platform window.

The problem is specific to Mac because on that platform any change to the size or location of a window can cause it to enter or leave the maximized (zoomed) state. So just setting the location can cause the maximized flag to be altered.

The proposed solution is to copy the intended state bits to Stage local variables to be used later in the configuration.

---------
### Progress

### Issues
 * [JDK-8316419](https://bugs.openjdk.org/browse/JDK-8316419): [macos] Setting X/Y makes Stage maximization not work before show (**Bug** - P4)
 * [JDK-8255835](https://bugs.openjdk.org/browse/JDK-8255835): [macOS] Undecorated stage cannot be maximized (**Bug** - P4)
 * [JDK-8305675](https://bugs.openjdk.org/browse/JDK-8305675): [macos] Stage set to iconified before being shown is displayed on screen (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1258/head:pull/1258` \
`$ git checkout pull/1258`

`$ git checkout pull/1258` \
`$ git pull https://git.openjdk.org/jfx.git pull/1258/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1258`

`$ git pr show -t 1258`

</details>

<a href="https://git.openjdk.org/jfx/pull/1258.diff">https://git.openjdk.org/jfx/pull/1258.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1258#issuecomment-1758957846)

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
+++ b/modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
@@ -1138,19 +1138,9 @@ public class Stage extends Window {
      * This can be replaced by listening for the onShowing/onHiding events
      * Note: This method MUST only be called via its accessor method.
      */
-    private boolean postVisibleFullScreen;
-    private boolean postVisibleIconified;
-    private boolean postVisibleMaximized;
-
     private void doVisibleChanging(boolean value) {
         Toolkit toolkit = Toolkit.getToolkit();
         if (value && (getPeer() == null)) {
-            // In case the platform window alters these
-            // properties while we're finishing installation.
-            postVisibleFullScreen = isFullScreen();
-            postVisibleIconified = isIconified();
-            postVisibleMaximized = isMaximized();
-
             // Setup the peer
             Window window = getOwner();
             TKStage tkStage = (window == null ? null : window.getPeer());
@@ -1191,10 +1181,10 @@ public class Stage extends Window {
             TKStage peer = getPeer();
             peer.setImportant(isImportant());
             peer.setResizable(isResizable());
-            peer.setFullScreen(postVisibleFullScreen);
+            peer.setFullScreen(isFullScreen());
             peer.setAlwaysOnTop(isAlwaysOnTop());
-            peer.setIconified(postVisibleIconified);
-            peer.setMaximized(postVisibleMaximized);
+            peer.setIconified(isIconified());
+            peer.setMaximized(isMaximized());
             peer.setTitle(getTitle());
 
             if (!isIconified()) {
--- a/modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
+++ b/modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
@@ -1138,9 +1138,19 @@ public class Stage extends Window {
      * This can be replaced by listening for the onShowing/onHiding events
      * Note: This method MUST only be called via its accessor method.
      */
+    private boolean postVisibleFullScreen;
+    private boolean postVisibleIconified;
+    private boolean postVisibleMaximized;
+
     private void doVisibleChanging(boolean value) {
         Toolkit toolkit = Toolkit.getToolkit();
         if (value && (getPeer() == null)) {
+            // In case the platform window alters these
+            // properties while we're finishing installation.
+            postVisibleFullScreen = isFullScreen();
+            postVisibleIconified = isIconified();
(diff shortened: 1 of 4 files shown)


lukostyra (commented):
Looks good on first sight. I'll verify it soon on all platforms.

lukostyra (commented):
This change causes a regression on Windows, test `MaximizeUndecorated.testMaximized` starts failing:
```
MaximizeUndecorated > testMaximize FAILED
    java.lang.AssertionError: Stage has moved to desktop top corner
        at org.junit.Assert.fail(Assert.java:89)
        at test.javafx.stage.MaximizeUndecorated.testMaximize(MaximizeUndecorated.java:85)
```

beldenfox:
Fix is now localized to the Mac. When an NSView is being initialized the `isZoomed` call returned bogus results. New code checks to see if the view's screen is set. For undecorated windows the OS assumed the zoomed size should be the same as the unzoomed size, not the size of the screen. This could also lead `isZoomed` to return surprising results.

lukostyra (approved):
LGTM

lukostyra:
I actually wanted to take a second look at Linux too, since JDK-8255835 also mentioned the bug happening on Linux. Will get back to it tomorrow morning.

lukostyra:
/issue JDK-8255835

I can confirm this also solves JDK-8255835 - both MaximizeUndecorated test and the test program attached to the issue work as they are supposed to after these changes.

kevinrushforth:
@andy-goryachev-oracle Can you be the second reviewer?

andy-goryachev-oracle:
The test program StartIconified JDK-8305675 works the same regardless of the fix: the window flashes, then immediately gets minimized.  Is this expected?
macOS 13.5.2

andy-goryachev-oracle:
The test program MaximizedStage JDK-8255835 : very strange - the window appears to be maximized, but not in full screen more, the OS tool bar is on top and the dock is visible.  It is impossible to minimize or normalize or quit the app via its toolbar.

Is this is acceptable?

Screenshot 2023-10-31 at 13 58 57

andy-goryachev-oracle:
SetXYAndMaximize JDK-8316419 works the same with or without the fix, the window appears maximized to fill the visible area (excl. the dock and the tool bar):

Screenshot 2023-10-31 at 14 02 37

Is it expected?

beldenfox:
> The test program MaximizedStage JDK-8255835 : very strange - the window appears to be maximized beyond the visible area, the OS tool bar is on top and the dock is visible. It is impossible to minimize or normalize or quit the app via its toolbar.

The window is borderless. It's difficult to see what's going on with the original test file so I added a new one to the JBS. MaximizedBorderStage adds a thin red border to the edge of the scene so you can verify that the JavaFX window doesn't go under the toolbar, the top edge of the window is flush with the bottom edge of the toolbar.

> SetXYAndMaximize JDK-8316419 works the same with or without the fix, the window appears maximized to fill the visible area (excl. the dock and the tool bar):

> The test program StartIconified JDK-8305675 works the same regardless of the fix: the window flashes, then immediately gets minimized. Is this expected? macOS 13.5.2

When I run these two test programs using the current JavaFX master they both fail (the windows don't iconify or maximize) but they work correctly with this fix. This is on my Mac running 13.6 but the OS version shouldn't matter.

andy-goryachev-oracle:
MaximizedStage with a border is much cleaner, works as expected, thank you!

StartIconified_8305675 and SetXYAndMaximize_8316419 work the same with our without fix, I just double checked.
May be because I am running 13.5.2 on M1 silicon?

PR #1263, opened 2023-10-17
8205067: Resizing window with TextField hides text value

karthikpandelu:
Because of a missing conditional check in the `updateTextPos()`,  the `textTranslateX` value was not getting updated when TextField size was changed as a result of resizing window.

Updated the CENTER and LEFT cases in the `updateTextPos()` method to fix the issue.

The fix can be validated using MonkeyTester.
Steps to select TextField option in Monkey Tester.

- Open the MonkeyTester app and select TextField from the left option pane.
- Select Long from Text selection option and TOP_LEFT from Alignment dropdown.
- After above step, follow the steps given in the bug to validate the fix in monkey tester.

---------
### Progress

### Issue
 * [JDK-8205067](https://bugs.openjdk.org/browse/JDK-8205067): Resizing window with TextField hides text value (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1263/head:pull/1263` \
`$ git checkout pull/1263`

`$ git checkout pull/1263` \
`$ git pull https://git.openjdk.org/jfx.git pull/1263/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1263`

`$ git pr show -t 1263`

</details>

<a href="https://git.openjdk.org/jfx/pull/1263.diff">https://git.openjdk.org/jfx/pull/1263.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1263#issuecomment-1766307985)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
@@ -811,7 +811,8 @@ public class TextFieldSkin extends TextInputControlSkin<TextField> {
             } else {
                 newX = midPoint - textNodeWidth / 2;
                 // Update if there is space on the right
-                if (newX + textNodeWidth <= textRight.get() - caretWidth / 2) {
+                if (newX + textNodeWidth <= textRight.get() - caretWidth / 2 ||
+                        (textNodeWidth > textRight.get() && ((textNodeWidth - textRight.get() + caretWidth) < Math.abs(oldX)))) {
                     textTranslateX.set(newX);
                 } else if (newX < 0 && oldX > caretWidth / 2) {
                     textTranslateX.set(caretWidth / 2);
@@ -843,7 +844,8 @@ public class TextFieldSkin extends TextInputControlSkin<TextField> {
           default:
             newX = caretWidth / 2;
             // Update if there is space on either side.
-            if (newX < oldX || newX + textNodeWidth <= textRight.get()) {
+            if (newX < oldX || newX + textNodeWidth <= textRight.get() ||
+                    ((textNodeWidth - textRight.get() + caretWidth) < Math.abs(oldX))) {
                 textTranslateX.set(newX);
             }
             if (usePromptText.get()) {
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
@@ -811,9 +811,10 @@ public class TextFieldSkin extends TextInputControlSkin<TextField> {
             } else {
                 newX = midPoint - textNodeWidth / 2;
                 // Update if there is space on the right
-                if (newX + textNodeWidth <= textRight.get() - caretWidth / 2 ||
-                        (textNodeWidth > textRight.get() && ((textNodeWidth - textRight.get() + caretWidth) < Math.abs(oldX)))) {
+                if (newX + textNodeWidth <= textRight.get() - caretWidth / 2) {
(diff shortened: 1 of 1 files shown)


andy-goryachev-oracle:
@karthikpandelu there seems to much flicker.  Try placing the caret not at the end of the text, but a few characters before the end.
Could we make it more like JTextField behavior which tries to position the text such that it occupies all the available space?

karthikpandelu:
> @karthikpandelu there seems to much flicker. Try placing the caret not at the end of the text, but a few characters before the end. 

I see the issue. It always keeps the caret at the right edge if we increase the size. This creates flicker.

>Could we make it more like JTextField behavior which tries to position the text such that it occupies all the available space?

You mean we shall clip the text from right side as well similar to how it behaves if we decrease the window size from the left side?

andy-goryachev-oracle:
It looks like it tries to keep the caret visible.  There is no flicker if the caret at the end of text.  If you look at how JTextField behaves in a similar situation, there it tries to keep the end of text visible.
So perhaps if we change the logic to keep the end of text, instead of the caret, visible, the flicker might go away.

karthikpandelu:
> If you look at how JTextField behaves in a similar situation, there it tries to keep the end of text visible.

I see that in JTextField the end of the text gets clipped when we place the caret either at the end or in any other position and reduce the window size. This is different to what you suggested above. Attaching screenshot for reference.
image
Am I missing something here? Is there any changes I have to make to get the behavior you mentioned?

andy-goryachev-oracle:
The scenario I was referring to was when expanding the width of the TextField.  As it tries to reposition the text, the movement is jerky - try setting the cursor to left of the last character and compare it with when you set the cursor at the right of the last character (the end of the text).  The latter is smooth(er).

Then try to compare the same scenario with JTextField (Tools -> JTextArea/JTextField Embedded in SwingNode, you might need to git pull https://github.com/andy-goryachev-oracle/MonkeyTest)

Swing does it much smoother because I suspect it tries not to allow for blank space at the end of the field.

I am suggesting we might try doing something that would make the overall experience a bit smoother, if possible.  But if it is too much work, the fix in this PR can still be considered an improvement.

karthikpandelu:
I get it now. Thanks for the details @andy-goryachev-oracle . I'll work on it and see how I can improve the proposed solution.

karthikpandelu:
Modified the code to remove flickering when caret is placed few characters before the end.
@andy-goryachev-oracle please review.

andy-goryachev-oracle:
Works beautifully, even with the split caret:

Screenshot 2023-10-25 at 08 11 31

good job!

kevinrushforth:
Can you provide an automated test for this fix?

andy-goryachev-oracle:
> Can you provide an automated test for this fix?

I wonder.
1. this test ought to be headful, since the font metrics are involved
2. it might be better to develop a wider test that verifies that the caret (at least in the case of a single caret) is always visible, there is no dead space beyond the end/beginning of text when resized, using all node orientations, using a number of test strings, and so on.

There might be some limited value in testing the exact scenario (to avoid tripping over the same condition again), but I think it's a kind of situation the monkey tester was developed to address.

What do you think?

karthikpandelu:
> Can you provide an automated test for this fix?

I agree on the points shared by @andy-goryachev-oracle on this.
I also tried to write both unit test and system test but couldn't succeed because the issue could not be reproduced in the unit test framework and there is no convenient way to figure out the actually text position on changing the text field size in system test.
Monkey tester seems to test many scenarios related to this bug.

Created [JDK-8318860](https://bugs.openjdk.org/browse/JDK-8318860) to add wider range of tests as suggested above.

PR #1271, opened 2023-10-26
8317836: FX nodes embedded in JFXPanel need to track component orientation

prsadhuk:
FX Nodes embedded in a Swing JFXPanel does not track the component orientation and FX nodes remain unaffected when component orientation changes.
Fix made sure JavaFX scene embedded in a JFXPanel should inherit the value from the JFXPanel.

---------
### Progress

### Issue
 * [JDK-8317836](https://bugs.openjdk.org/browse/JDK-8317836): FX nodes embedded in JFXPanel need to track component orientation (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1271/head:pull/1271` \
`$ git checkout pull/1271`

`$ git checkout pull/1271` \
`$ git pull https://git.openjdk.org/jfx.git pull/1271/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1271`

`$ git pr show -t 1271`

</details>

<a href="https://git.openjdk.org/jfx/pull/1271.diff">https://git.openjdk.org/jfx/pull/1271.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1271#issuecomment-1780764719)


modules/javafx.swing/src/main/java/javafx/embed/swing/JFXPanel.java
```diff
@@ -810,6 +812,14 @@ protected void paintComponent(Graphics g) {
 
         Graphics gg = null;
         try {
+            ComponentOrientation cor = this.getComponentOrientation();
+            if (!cor.equals(ComponentOrientation.UNKNOWN)) {
+                String orient = cor.equals(ComponentOrientation.LEFT_TO_RIGHT)
```
andy-goryachev-oracle, line 817: Why a String??  Should it be a simple boolean?

prsadhuk: Yes, Optimized...

modules/javafx.swing/src/main/java/javafx/embed/swing/JFXPanel.java
```diff
@@ -810,6 +812,12 @@ protected void paintComponent(Graphics g) {
 
         Graphics gg = null;
         try {
+            ComponentOrientation cor = this.getComponentOrientation();
+            if (!cor.equals(ComponentOrientation.UNKNOWN)) {
+                getScene().setNodeOrientation(cor.equals(ComponentOrientation.RIGHT_TO_LEFT) ?
```
andy-goryachev-oracle, line 817: 1. I think getScene() might return null (sorry, did not see this immediately)
2. if possible, could we try limit code to one statement per line, i.e.

```
Scene scene = getScene(); // it's a volatile, so better assign to a variable
if(scene != null) {
  boolean rtl = cor.equals();
  scene.setNodeOrientation(rtl ? ...);
}
```

andy-goryachev-oracle:
Please merge in the master branch, as it has new changes in native:

Exception in thread "Thread-1" java.lang.UnsatisfiedLinkError: 'boolean com.sun.glass.ui.mac.MacApplication._isNormalTaskbarApp()'

modules/javafx.swing/src/main/java/javafx/embed/swing/JFXPanel.java
prsadhuk: ok..updated..

modules/javafx.swing/src/main/java/javafx/embed/swing/JFXPanel.java
```diff
@@ -810,6 +812,16 @@ protected void paintComponent(Graphics g) {
...
         Graphics gg = null;
         try {
+            ComponentOrientation cor = this.getComponentOrientation();
+            if (!cor.equals(ComponentOrientation.UNKNOWN)) {
+                Scene scene = getScene();
+                if (scene != null) {
+                    boolean rtl = cor.equals(ComponentOrientation.RIGHT_TO_LEFT);
+                    scene.setNodeOrientation(rtl ?
+                                             NodeOrientation.RIGHT_TO_LEFT :
+                                             NodeOrientation.LEFT_TO_RIGHT);
```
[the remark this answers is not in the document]
prsadhuk: I have tried to put in place what you suggested. Can you please take a look at the PR and see if it's in the right direction and suggest further.
I have few doubts

- It seems JFXPanel call EmbeddedStage interface methods and not the other way round, so I am not sure how to "call a method in embedded stage that would delegate to JFXPanel".
- It seems [Scene::calcEffectiveNodeOrientation](https://github.com/openjdk/jfx/blob/9b93c962f45e5cf0b3a9f1090f307603be130a0e/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java#L6343) is private and not part of swing-interop code/package so I am not sure how to leverage it from JFXPanel/EmbeddedStage....I have tried setting EmbeddedStage orientation from JFXPanel and query it from Scene but am not sure how to call `calcEffectiveNodeOrientation`

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/EmbeddedStage.java
```diff
@@ -280,6 +281,28 @@ public void setFocused(final boolean focused, final int focusCause) {
         }
     }
 
+    NodeOrientation orientation;
```
kevinrushforth, line 284: Move this to `com.sun.javafx.stage.EmbeddedWindow`.

kevinrushforth (commented):
I left some comments. This is heading in the right direction, but needs some adjustments.

modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
```diff
@@ -6348,6 +6348,8 @@ private NodeOrientation calcEffectiveNodeOrientation() {
                 Window parent = null;
                 if (window instanceof Stage) {
                     parent = ((Stage)window).getOwner();
+                } else if (window instanceof com.sun.javafx.embed.EmbeddedSceneInterface) {
+                    return ((com.sun.javafx.embed.EmbeddedStageInterface)window).getNodeOrientation();
```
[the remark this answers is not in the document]
kevinrushforth: I meant "...never an instance of `EmbeddedStageInterface`"

modules/javafx.swing/src/main/java/javafx/embed/swing/JFXPanel.java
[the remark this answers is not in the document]
prsadhuk: I have made the suggested changes and updated the PR.
But it seems the orientation of the text in the test attached in JBS is at the middle initially (and not at LTR or RTL) and it only begins to track the orientation after 1 iteration of changing JFrame orientation. Is it because of INHERIT default value which causes the effective orientation to not follow the changed orientation immediately? Does it need to be looked at or is it expected?

modules/javafx.graphics/src/main/java/com/sun/javafx/stage/EmbeddedWindow.java
```diff
@@ -85,4 +88,13 @@ private void doVisibleChanging(boolean visible) {
         }
     }
 
+    public void setNodeOrientation(NodeOrientation nor) {
+        orientation = nor;
+        final Scene sceneValue = getScene();
+        SceneHelper.parentEffectiveOrientationInvalidated(sceneValue);
```
kevinrushforth, line 94: You need to check whether the value has actually changed before calling invalidated or this will cause a performance regression.

Minor: I might not bother saving `getScene()` in a local var since you only use it once and `getScene()` is short.

modules/javafx.graphics/src/main/java/com/sun/javafx/stage/EmbeddedWindow.java
```diff
@@ -47,6 +49,7 @@ public class EmbeddedWindow extends Window {
      }
 
     private HostInterface host;
+    private NodeOrientation orientation;
```
kevinrushforth, line 52: Initialize this to LTR.

kevinrushforth (commented):
This looks good now with a couple comments. I'll so some testing.

And speaking of testing, my question about providing a regression test is still pending.

modules/javafx.swing/src/main/java/javafx/embed/swing/JFXPanel.java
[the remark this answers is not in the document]
kevinrushforth: This might be due to the initialization problem I pointed out or it might be for some other reason. Can you make the suggested change and then re-test?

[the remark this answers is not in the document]
kevinrushforth: OK. Please file a new testbug to create a regression test for this.

andy-goryachev-oracle:
tested with the following scenario (using the updated monkey tester https://github.com/andy-goryachev-oracle/MonkeyTest ):

1. Tools -> FX TextArea Embedded
2. change JFrame.componentOrientation to RTL
3. change JFrame.componentOrientation to INHERIT

I would expect the TextArea to go to LTR but it stays in RTL.

Screenshot 2023-10-30 at 09 53 29

prsadhuk:
> tested with the following scenario (using the updated monkey tester https://github.com/andy-goryachev-oracle/MonkeyTest ):
> 
>     1. Tools -> FX TextArea Embedded
> 
>     2. change JFrame.componentOrientation to RTL
> 
>     3. change JFrame.componentOrientation to INHERIT
> 
> 
> I would expect the TextArea to go to LTR but it stays in RTL.
> 
> Screenshot 2023-10-30 at 09 53 29

JFrame does not support "INHERIT" orientation as per spec https://docs.oracle.com/en/java/javase/21/docs/api/java.desktop/java/awt/ComponentOrientation.html#LEFT_TO_RIGHT
so it will effectively be ignored so I think it is expected to retain RTL in this case..

modules/javafx.swing/src/main/java/javafx/embed/swing/JFXPanel.java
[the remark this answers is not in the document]
prsadhuk: Yes, filed https://bugs.openjdk.org/browse/JDK-8319147

andy-goryachev-oracle:
> I think it is expected to retain RTL in this case..

Ok, INHERIT is not supported, but why do you expect RTL?  Shouldn't it go back to UNKNOWN and therefore LTR?

```
public static final [ComponentOrientation](https://docs.oracle.com/en/java/javase/21/docs/api/java.desktop/java/awt/ComponentOrientation.html) UNKNOWN

... To preserve the behavior of existing applications, isLeftToRight will return true for this value.
```

prsadhuk:
> > I think it is expected to retain RTL in this case..
> 
> Ok, INHERIT is not supported, but why do you expect RTL? Shouldn't it go back to UNKNOWN and therefore LTR?
> 
> ```
> public static final [ComponentOrientation](https://docs.oracle.com/en/java/javase/21/docs/api/java.desktop/java/awt/ComponentOrientation.html) UNKNOWN
> 
> ... To preserve the behavior of existing applications, isLeftToRight will return true for this value.
> ```

I expect RTL because INHERIT is unsupported so it will retain the existing orientation, which is what I thought..

andy-goryachev-oracle:
> I expect RTL because INHERIT is unsupported so it will retain the existing orientation,

Why existing and not the initial?

prsadhuk:
> > I expect RTL because INHERIT is unsupported so it will retain the existing orientation,
> 
> Why existing and not the initial?

Tha tis grey area which is not mentioned in the spec so I think retaining existing is the appropriate according to me..

prsadhuk:
I can add

if (!cor.equals(ComponentOrientation.UNKNOWN)) {
                boolean rtl = cor.equals(ComponentOrientation.RIGHT_TO_LEFT);
                stage.setNodeOrientation(rtl ? NodeOrientation.RIGHT_TO_LEFT :
                                               NodeOrientation.LEFT_TO_RIGH**T);
            } else {
stage.setNodeOrientation(NodeOrientation.LEFT_TO_RIGHT);**
}

if that is what your MonkeyTester expects..

andy-goryachev-oracle:
> Tha tis grey area which is not mentioned in the spec so I think retaining existing is the appropriate according to me..

We may need others to weigh in.  My understanding, UNKNOWN works exactly as INHERIT, behaving as LTR if nothing is set.  Setting the JFrame orientation to UNKNOWN (that's what happens when you choose INHERIT in the monkey tester) changes the Swing side to LTR and should set JFXPanel to INHERIT and therefore to LTR.

kevinrushforth:
What would the behavior of a JLabel in a JPanel be when the JFrame is set back to UNKNOWN? I would expect a Scene in a JFXPanel to work the same way.

andy-goryachev-oracle:
> What would the behavior of a JLabel in a JPanel be when the JFrame is set back to UNKNOWN? I would expect a Scene in a JFXPanel to work the same way.

exactly my point!  the orientation of JComboBoxes in the toolbar in the monkey tester switches to LTR per the spec (UNKNOWN == LTR).

(Please update the MT, I changed it so the choice correctly reflects the constant being set)

kevinrushforth:
> I can add in JFXPanel.java
> ```
> if (!cor.equals(ComponentOrientation.UNKNOWN)) {
> boolean rtl = cor.equals(ComponentOrientation.RIGHT_TO_LEFT);
> stage.setNodeOrientation(rtl ? NodeOrientation.RIGHT_TO_LEFT :
> NodeOrientation.LEFT_TO_RIGHT);
> } else {
> stage.setNodeOrientation(NodeOrientation.LEFT_TO_RIGHT);
> }
> ```
> 
> if that is what your MonkeyTester expects..

Presuming we want to treat UNKOWN the same as RIGHT_TO_LEFT, which I think we do, then it's even simpler than that. Just remove the check for UNKNOWN entirely, in which case `rtl` will be true if `cor` is RIGHT_TO_LEFT and false otherwise, meaning that UNKNOWN will be treated the same as LEFT_TO_RIGHT.

andy-goryachev-oracle:
> if (!cor.equals(ComponentOrientation.UNKNOWN)) {
> boolean rtl = cor.equals(ComponentOrientation.RIGHT_TO_LEFT);
> stage.setNodeOrientation(rtl ? NodeOrientation.RIGHT_TO_LEFT :
> NodeOrientation.LEFT_TO_RIGH**T); } else { stage.setNodeOrientation(NodeOrientation.LEFT_TO_RIGHT);**
> }

I think it should do a simple translation:
LTR -> LTR
RTL -> RTL
UNKNOWN -> INHERIT

andy-goryachev-oracle:
> Presuming we want to treat UNKOWN the same as RIGHT_TO_LEFT

no, we want to treat is as INHERIT, i.e. back to initial state.  INHERIT shows up as LTR.

kevinrushforth:
> > if (!cor.equals(ComponentOrientation.UNKNOWN)) {
> > boolean rtl = cor.equals(ComponentOrientation.RIGHT_TO_LEFT);
> > stage.setNodeOrientation(rtl ? NodeOrientation.RIGHT_TO_LEFT :
> > NodeOrientation.LEFT_TO_RIGH**T); } else { stage.setNodeOrientation(NodeOrientation.LEFT_TO_RIGHT);**
> > }
> 
> I think it should do a simple translation: LTR -> LTR RTL -> RTL UNKNOWN -> INHERIT

For the _effective_ NodeOrientation attribution in the embedded window, there's no such thing as INHERIT. It's either LTR (default) or RTL.

prsadhuk:
> What would the behavior of a JLabel in a JPanel be when the JFrame is set back to UNKNOWN?

It sets it back to LTR it is toggled from RIGHT_TO_LEFT to UNNOWN.
So, I updated PR to set JFXPanel to LTR for UNKNOWN..If set to INHERIT, then it renders the text in middle initially..

andy-goryachev-oracle:
> For the _effective_ NodeOrientation attribution in the embedded window, there's no such thing as INHERIT. It's either LTR (default) or RTL.

Correct.  The initial value for node orientation property is INHERIT which generates effective orientation LTR.  So UNKNOWN on JFrame should reset the value to INHERIT, making the UI turn LTR methinks.

kevinrushforth:
> The initial value for node orientation property is INHERIT

For the Scene, yes, but we aren't talking about that. The EmbeddedWindow currently never has the value of INHERIT. It is either LTR (default) or RTL. Maybe it _should_ be set to INHERIT by default and when the JFrame is UNKNOWN. We would need to check whether Scene's getEffectiveOrientation handles that correctly.

andy-goryachev-oracle:
> For the Scene, yes, but we aren't talking about that.

I got confused by the variable name "stage".  You are right!  

The last code https://github.com/openjdk/jfx/pull/1271/commits/3b7be5d9888a24a5d1db2a5abcad3c647533544f works as expected.

Perhaps we can clean up the logic a bit, something like

```
boolean rtl = ComponentOrientation.RIGHT_TO_LEFT.equals(cor);
stage.setNodeOrientation(rtl ? RTL : LTR)
```

kevinrushforth:
> 
The initial value for node orientation property is INHERIT

For the Scene, yes, but we aren't talking about that. The EmbeddedWindow currently never has the value of INHERIT. It is either LTR (default) or RTL. Maybe it should be set to INHERIT by default and when the JFrame is UNKNOWN. We would need to check whether Scene's getEffectiveOrientation handles that correctly.

I checked. It doesn't. effectiveOrientation must be either RTL or LTR (which makes sense), so without additional adjustments to the fix, I think what Prasanta has now is correct. I'll look more closely when I test it.

kevinrushforth:
> Perhaps we can clean up the logic a bit, something like
> 
> ```
> boolean rtl = ComponentOrientation.RIGHT_TO_LEFT.equals(cor);
> stage.setNodeOrientation(rtl ? RTL : LTR)
> ```

Yes, that's also what I trying to suggest in an earlier comment I made.

prsadhuk:
> > For the Scene, yes, but we aren't talking about that.
> 
> I got confused by the variable name "stage". You are right!
> 
> The last code [3b7be5d](https://github.com/openjdk/jfx/commit/3b7be5d9888a24a5d1db2a5abcad3c647533544f) works as expected.
> 
> Perhaps we can clean up the logic a bit, something like
> 
> ```
> boolean rtl = ComponentOrientation.RIGHT_TO_LEFT.equals(cor);
> stage.setNodeOrientation(rtl ? RTL : LTR)
> ```

Sure..done..

andy-goryachev-oracle (approved):
🕶 looks good, thanks!

works as expected when FX side is set to INHERIT
works as expected when FX is forced to LTR/RTL

kevinrushforth (approved):
All looks good now. Thanks.

PR #1274, opened 2023-10-31
8303478: DatePicker throws uncatchable exception on tab out from garbled text

brunesto:
The fix prevents the DatePicker from losing focus if the date is not parsable.

---------
### Progress

### Issue
 * [JDK-8303478](https://bugs.openjdk.org/browse/JDK-8303478): DatePicker throws uncatchable exception on tab out from garbled text (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1274/head:pull/1274` \
`$ git checkout pull/1274`

`$ git checkout pull/1274` \
`$ git pull https://git.openjdk.org/jfx.git pull/1274/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1274`

`$ git pr show -t 1274`

</details>

<a href="https://git.openjdk.org/jfx/pull/1274.diff">https://git.openjdk.org/jfx/pull/1274.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1274#issuecomment-1804292058)


modules/javafx.controls/src/main/java/javafx/scene/control/DatePicker.java
```diff
@@ -147,11 +148,23 @@ public DatePicker(LocalDate localDate) {
...
-                commitValue();
+                commitValueOnFocusLost();
             }
         });
     }
 
+    /**
+     * This is to address https://bugs.openjdk.org/browse/JDK-8303478
+     */
+    protected void commitValueOnFocusLost(){
```
mstr2, line 159: The `protected` modifier makes this method public API, which would most likely require a CSR to be approved. Since there doesn't seem to be a reason why this method should be public, I suggest making it private instead.

modules/javafx.controls/src/main/java/javafx/scene/control/DatePicker.java
```diff
@@ -147,11 +148,23 @@ public DatePicker(LocalDate localDate) {
...
+        try {
+            commitValue();
+        } catch (DateTimeParseException dtpe) {
+            this.requestFocus();
```
mstr2, line 163: I'm not sure whether moving focus to the control is a good idea, because this effectively captures focus and forces me to enter a valid date before I can do anything else. Why not just restore the previous value instead?

modules/javafx.controls/src/main/java/javafx/scene/control/DatePicker.java
brunesto: ok, I have changed it to private

modules/javafx.controls/src/main/java/javafx/scene/control/DatePicker.java
brunesto: I have changed the behavior as you suggested.
The downside is that user might not notice that his/her changes were replaced, but it is anyway consistent to what happens when typing ENTER.

Robot test is updated accordingly.

tests/system/src/test/java/test/robot/javafx/scene/DatePickerOnFocusLostTest.java
```diff
@@ -0,0 +1,214 @@
...
+
+        // 4 check that in case of typo the text value is reverted (fixed by patch for JDK-8303478)
+        Assert.assertEquals(typo,INITIAL_DATE_TEXT.equals(datePicker.getEditor().getText()));
+
+        // 5 check that in case of typo, the value has not changed (already working as expected before the patch)
+        Assert.assertEquals(typo, datePicker.valueProperty().get().equals(now));
+
+        // 6 check if onChangeListener was called (already working as expected before the patch)
+        if (typo)
+            Assert.assertEquals(onChangeListenerCalled, 0);
```
andy-goryachev-oracle, line 141: please always use { } in `if` statements

tests/system/src/test/java/test/robot/javafx/scene/DatePickerOnFocusLostTest.java
```diff
@@ -0,0 +1,214 @@
...
+            // Note that change listener was already beeing called as expected before the patch
+            ChangeListener chListener = (observable, oldValue, newValue) -> {
+                onChangeListenerCalled++;
+            };
+            datePicker.valueProperty().addListener(chListener);
+
+            root.getChildren().add(datePicker);
+            button = new Button("...");
+            root.getChildren().add(button);
+
```
andy-goryachev-oracle, line 185: extra newline (here and elsewhere)

tests/system/src/test/java/test/robot/javafx/scene/DatePickerOnFocusLostTest.java
```diff
@@ -0,0 +1,214 @@
...
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
+import java.util.concurrent.CountDownLatch;
+
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
```
andy-goryachev-oracle, line 36: could we switch to junit5 please?

tests/system/src/test/java/test/robot/javafx/scene/DatePickerOnFocusLostTest.java
```diff
@@ -0,0 +1,214 @@
...
+    public void testDatePickerCannotCommitTypo() throws Exception {
+        testDatePickerCommit(true);
+    }
+
+    // This test is for verifying a specific behavior.
+    // 1. Click on datePicker
+    // 2. Modify with keystrokes the day of month value, optionally with typo
+    // 3. Click on button to grab the focus and hence attempt to datePicker.commitValue()
+    // 4. Verify that in case of typo, the value was reverted, as well as the editor's text
+    public void testDatePickerCommit(boolean typo) throws Exception {
```
andy-goryachev-oracle, line 98: thank you for explaining what the test cases attempt to do!

tests/system/src/test/java/test/robot/javafx/scene/DatePickerOnFocusLostTest.java
```diff
@@ -0,0 +1,214 @@
+/*
+ * Copyright (c) 2018, 2023, Oracle and/or its affiliates. All rights reserved.
```
andy-goryachev-oracle, line 2: it's a brand new file, I wonder if it should have 2023 only

andy-goryachev-oracle (commented):
Just preliminary comments, will do a full review once OCA process is completed.

tests/system/src/test/java/test/robot/javafx/scene/DatePickerOnFocusLostTest.java
Maran23, line 98: you actually can also write a normal JUnit test as far as I get this. Then there is no need for `Util.runAndWait`, `sleep` and so on

modules/javafx.controls/src/main/java/javafx/scene/control/DatePicker.java
```diff
@@ -147,11 +148,23 @@ public DatePicker(LocalDate localDate) {
...
         focusedProperty().addListener(o -> {
             if (!isFocused()) {
-                commitValue();
+                commitValueOnFocusLost();
             }
         });
     }
 
+    /**
+     * This is to address https://bugs.openjdk.org/browse/JDK-8303478
```
Maran23, line 157: I don't think this javadoc is needed, since it just links the issue one can also lookup when checking the git history

kevinrushforth: Exactly. Remove this line and either add a comment describing what the method does, or since the name is self-explanatory, remove the comment block entirely.

modules/javafx.controls/src/main/java/javafx/scene/control/DatePicker.java
```diff
@@ -147,11 +148,23 @@ public DatePicker(LocalDate localDate) {
...
+     */
+    private void commitValueOnFocusLost(){
```
kevinrushforth, line 159: Minor: add a space before the `{`

tests/system/src/test/java/test/robot/javafx/scene/DatePickerOnFocusLostTest.java
kevinrushforth: Unless this file was substantially copied from another test, it should be 2023 only.

tests/system/src/test/java/test/robot/javafx/scene/DatePickerOnFocusLostTest.java
kevinrushforth: A headless test is the preferred way to do it, if feasible. If it is, move the test to the `javafx.controls` module.

kevinrushforth (commented):
I left a few inline comments. I'll leave it to others to formally review it.

kevinrushforth:
I should add that we won't formally review the fix until the OCA status is recorded.

brunesto:
FYI a company Agreement was sent for the OCA, it is under review.

tests/system/src/test/java/test/robot/javafx/scene/DatePickerOnFocusLostTest.java
brunesto: I have removed the robot test, and replaced it by a simple test case in the existing class `DatePickerTest` in `javafx.controls`

modules/javafx.controls/src/main/java/javafx/scene/control/DatePicker.java
```diff
@@ -147,11 +148,20 @@ public DatePicker(LocalDate localDate) {
...
 
+    private void commitValueOnFocusLost() {
+        try {
+            commitValue();
+        } catch (DateTimeParseException dtpe) {
+            cancelEdit();
+        }
+    }
+
+
```
andy-goryachev-oracle, line 164: minor: extra newline

modules/javafx.controls/src/main/java/javafx/scene/control/DatePicker.java
```diff
@@ -147,11 +148,20 @@ public DatePicker(LocalDate localDate) {
 
         focusedProperty().addListener(o -> {
             if (!isFocused()) {
-                commitValue();
+                commitValueOnFocusLost();
```
andy-goryachev-oracle, line 151: minor: I don't know if it's worth creating a new method here - try .. catch can simply be moved here.

andy-goryachev-oracle (commented):
thank you for adding a test case!

the testing looks good.  have a couple of minor suggestions.

andy-goryachev-oracle (approved):
Looks good, thank you for making the changes!

modules/javafx.controls/src/test/java/test/javafx/scene/control/DatePickerTest.java
```diff
@@ -737,6 +737,36 @@ public void testFocusLost() {
         stageLoader.dispose();
     }
 
+    /**
+     * When DatePicker looses focus with misformatted text in the editor,
+     * checks that the value is not changed, and the text is reverted to the value
+     */
+    @Test
+    public void testFocusLostWithTypo() {
+        //datePicker.setEditable(true); // other test cases have it .is this necessary?
```
Maran23, line 746: Looks like you can remove this

modules/javafx.controls/src/test/java/test/javafx/scene/control/DatePickerTest.java
```diff
@@ -737,6 +737,36 @@ public void testFocusLost() {
...
+
+        Button button = new Button();
+        StageLoader stageLoader = new StageLoader(new HBox(datePicker, button));
+
+        // initial value
+        datePicker.setValue(LocalDate.of(2015, 03, 25));
+        assertEquals("3/25/2015",datePicker.getEditor().getText());
+
+        // set misformatted text
+        //stageLoader.getStage().requestFocus(); // other test cases have it . is this necessary?
```
Maran23, line 756: same here, remove as it works without :)

modules/javafx.controls/src/test/java/test/javafx/scene/control/DatePickerTest.java
```diff
@@ -737,6 +737,36 @@ public void testFocusLost() {
...
+    @Test
+    public void testFocusLostWithTypo() {
+        //datePicker.setEditable(true); // other test cases have it .is this necessary?
+
+        Button button = new Button();
+        StageLoader stageLoader = new StageLoader(new HBox(datePicker, button));
+
+        // initial value
+        datePicker.setValue(LocalDate.of(2015, 03, 25));
+        assertEquals("3/25/2015",datePicker.getEditor().getText());
```
Maran23, line 753: Minor: A space after comma is missing

modules/javafx.controls/src/test/java/test/javafx/scene/control/DatePickerTest.java
```diff
@@ -737,6 +737,36 @@ public void testFocusLost() {
...
+        // set misformatted text
+        //stageLoader.getStage().requestFocus(); // other test cases have it . is this necessary?
+        datePicker.requestFocus();
+        datePicker.getEditor().setText("11/24/20xx");
+
+        // loosing focus triggers cancelEdit() because the text cannot be parsed
+        button.requestFocus();
+
+        // check that value remains unchanged, and text is reverted
+        assertEquals(LocalDate.of(2015, 03, 25),datePicker.getValue());
```
Maran23, line 764: Minor: A space after comma is missing

modules/javafx.controls/src/test/java/test/javafx/scene/control/DatePickerTest.java
```diff
@@ -737,6 +737,36 @@ public void testFocusLost() {
...
+        assertEquals("3/25/2015",datePicker.getEditor().getText());
```
Maran23, line 765: Minor: A space after comma is missing

Maran23 (commented):
Looks good to me, just some minor things.
I also dug into the code and found out how this worked in JavaFX 8:

```
...
try {
    value = c.fromString(text);
} catch (Exception ex) {
    // Most likely a parsing error, such as DateTimeParseException
}
...
```

So this approach looks correct to me!
I just wonder if we want to catch `Exception` here as well, as it was done in the old code.
The contract then would be: If a converter throws any exception, we cancel the edit.
@andy-goryachev-oracle any objections here from your side?

modules/javafx.controls/src/test/java/test/javafx/scene/control/DatePickerTest.java
brunesto: ok, I have committed the changes in the test class

andy-goryachev-oracle:
> I just wonder if we want to catch `Exception` here as well, as it was done in the old code.
> The contract then would be: If a converter throws any exception, we cancel the edit.

There might be some discussion on that, but yes, I'd rather catch(Exception) because there might unchecked RuntimeExceptions...

modules/javafx.controls/src/main/java/javafx/scene/control/DatePicker.java
```diff
@@ -29,6 +29,7 @@
 
 import java.time.LocalDate;
 import java.time.DateTimeException;
+import java.time.format.DateTimeParseException;
```
andy-goryachev-oracle, line 32: is this import still needed?

brunesto: probably not :) I will remove it...

brunesto: indeed it was not necessary, so it is removed

andy-goryachev-oracle (approved):
perfection achieved :-)

PR #1277, opened 2023-11-02
8318841: macOS: Memory leak with MenuItem when Menu.useSystemMenuBar(true) is used

johanvos:
When the Java layer removes a systemmenu, release the native resources related to this systemmenu.
This removes the strong JNI Global ref, which prevents its references from being gc'ed.

The current implementation for the mac-specific system menu creates a menu, but never releases its resources. In the `dealloc` of this menu, the strong jni refs are deleted.
With this PR, we now release the native resources associated with a menuItem when that one is removed from a menu. A consequence is that this menuItem should never be used after being removed from its current menu (e.g. it should not be re-added, or its text/shortcut should not be altered). 
The current implementation will create a new MacMenuDelegate every time a menuItem is inserted into a menu, so there should be no references to the native resources lingering.

---------
### Progress

### Issue
 * [JDK-8318841](https://bugs.openjdk.org/browse/JDK-8318841): macOS: Memory leak with MenuItem when Menu.useSystemMenuBar(true) is used (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1277/head:pull/1277` \
`$ git checkout pull/1277`

`$ git checkout pull/1277` \
`$ git pull https://git.openjdk.org/jfx.git pull/1277/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1277`

`$ git pr show -t 1277`

</details>

<a href="https://git.openjdk.org/jfx/pull/1277.diff">https://git.openjdk.org/jfx/pull/1277.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1277#issuecomment-1790572055)


kevinrushforth (commented):
The fix looks good aside from a missing null check on the Java side. I also left a question as to about two possible other place that might need to zero out a pointer.

This will need to be tested in a variety of scenarios, including menus with separators, nested menus, and turning the system menu on and off. What I tested so far looks good.

johanvos:
> Is it possible to add an automated regression test?

I'm working on that, but it is not trivial as this functionality requires the QuantumToolkit and can not use the StubToolkit which is used in most tests.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacMenuDelegate.java
```diff
@@ -90,6 +90,7 @@ private native long _createMenuItem(String title, char shortcut, int modifiers,
     @Override public boolean remove(MenuItemDelegate item, int pos) {
         MacMenuDelegate macMenu = (MacMenuDelegate)item;
         _remove(ptr, macMenu == null ? 0L : macMenu.ptr, pos);
+        macMenu.ptr = 0L;
```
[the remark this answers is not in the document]
johanvos: I added the null check on `macMenu`.
I also added the zeroing in the other `remove` method (which indeed is for adding a menu to a menu) and to the `remove` method in `MacMenuBarDelegate`.
Apart from that, I noticed that there were a number of possible scenario's where the Delegate objects would still be used. For example, after removing a menuItem from a menu, one could still invoke methods on that menuItem object. The invocations would propagate to the native layer, causing a crash because the pointer to the `GlassMenu` was already zero (and the native GlassMenu instance was released).

johanvos:
> > > Is it possible to add an automated regression test?
> > 
> > 
> > I'm working on that, but it is not trivial as this functionality requires the QuantumToolkit and can not use the StubToolkit which is used in most tests.
> 
> You could add a headful system test (those all use the QuantumToolkit). You may or may not want it to be a robot-based test, depending on what else you want the test to do (e.g., a robot-based test could click outside the window, which deactivates the window and calls the release methods).

I added a test to the existing SystemMenuBarTest in the system tests that fails before and succeeds after the patch. Note that the app needs focus for this test.

tests/system/src/test/java/test/javafx/stage/SystemMenuBarTest.java
```diff
@@ -108,4 +113,61 @@ public MenuBar createFailingMenuBar() {
 
         return menuBar;
     }
+
+    @Test
+    public void testMemoryLeak() throws InterruptedException {
+        Util.runAndWait(() -> {
+            Thread.currentThread().setUncaughtExceptionHandler((t,e) -> {
+                e.printStackTrace();
+                failed.set(true);
```
kevinrushforth, line 122: To avoid the test hanging indefinitely, I recommend adding the following after setting the `failed` flag:

```
    memoryLatch.countDown();
```

tests/system/src/test/java/test/javafx/stage/SystemMenuBarTest.java
```diff
@@ -108,4 +113,61 @@ public MenuBar createFailingMenuBar() {
...
+        final Menu menu = new Menu("MyMenu");
+        menuBar.getMenus().add(menu);
+        menuBar.setUseSystemMenuBar(true);
+        root.getChildren().add(menuBar);
+
+        Scene scene = new Scene(root);
+        stage.setScene(scene);
+        stage.show();
+        stage.requestFocus();
+        Thread t = new Thread(){
```
kevinrushforth, line 145: Question: Since you do all the work in a `Platform.runLater`, is this thread needed?

tests/system/src/test/java/test/javafx/stage/SystemMenuBarTest.java
```diff
@@ -108,4 +113,61 @@ public MenuBar createFailingMenuBar() {
...
+                        menu.getItems().clear();
+                        MenuItem menuItem = new MenuItem("MyItem");
+                        WeakReference wr = new WeakReference<>(menuItem);
+                        uncollectedMenuItems.add(wr);
+                        menu.getItems().add(menuItem);
+                    });
+                }
+                Platform.runLater( () -> {
+                    System.gc();
+                    uncollectedMenuItems.removeIf(ref -> ref.get() == null);
```
kevinrushforth, line 163: This assumes that the references will all be collected by a single call to `System.gc` with no delay. While it worked on my system, it might be fragile. Have you considered using JMemoryBuddy? Or if that isn't suitable for some reason, maybe consider looping a few times with a sleep in between each call to gc?

tests/system/src/test/java/test/javafx/stage/SystemMenuBarTest.java
```diff
@@ -108,4 +113,61 @@ public MenuBar createFailingMenuBar() {
...
+                    assertEquals(1, uncollectedMenuItems.size(), "Only the last menuItem should be alive");
+                    memoryLatch.countDown();
```
kevinrushforth, line 165: If there is a failure, this statement will be unreached. This is why I recommended also doing a countDown on this latch in the `UncaughtExceptionHandler`

kevinrushforth (commented):
The updated fix looks good. The test passes with the fix and hangs indefinitely without the fix. I left a comment about the fix for the hang and a couple other comments as well.

tests/system/src/test/java/test/javafx/stage/SystemMenuBarTest.java
johanvos: good idea.

tests/system/src/test/java/test/javafx/stage/SystemMenuBarTest.java
johanvos: That's a good question. I do this because we start the test on the FX Thread, but I want to add/remove the menuItems one by one, giving time for a `pulse` in between them. If we do the 50 runs inside a single Runnable, without the possibility for the FX Thread to do other work, the result can be different compared to what we do when doing this in small runnables. 
But I might be missing something here?

kevinrushforth: Oh, I see that now. I had missed that you were calling this method from within a `Util::runAndWait`. Never mind.

tests/system/src/test/java/test/javafx/stage/SystemMenuBarTest.java
johanvos: Done

kevinrushforth (approved):
The test changes look good. One thing I noticed is that it now takes ~ 50 seconds to time out if there is a failure (a call to checkCollectable that returns false takes ~ 1 second, which is needed to ensure that it is a robust check).

Maybe cut down the number of menu objects from 50 to, say, 10? I'll approve it as is, since the only way this will ever be a problem is if we reintroduce this bug, and it's still under 1 minute so OK.

kevinrushforth (approved):

johanvos:
> The test changes look good. One thing I noticed is that it now takes ~ 50 seconds to time out if there is a failure (a call to checkCollectable that returns false takes ~ 1 second, which is needed to ensure that it is a robust check).
> 
> Maybe cut down the number of menu objects from 50 to, say, 10? I'll approve it as is, since the only way this will ever be a problem is if we reintroduce this bug, and it's still under 1 minute so OK.

makes sense, done.

kevinrushforth:
@jperedadnr do you want to be the second reviewer on this?

jperedadnr:
Yes, I'll review it.

jperedadnr (commented):
Looks good, tested successfully before (fails) and after (passes).
I've left some comments, mostly minor remarks.

tests/system/src/test/java/test/javafx/stage/SystemMenuBarTest.java
```diff
@@ -108,4 +115,64 @@ public MenuBar createFailingMenuBar() {
...
+                    }
+                    Platform.runLater(() -> {
+                        menu.getItems().clear();
+                        MenuItem menuItem = new MenuItem("MyItem");
+                        WeakReference wr = new WeakReference<>(menuItem);
+                        uncollectedMenuItems.add(wr);
+                        menu.getItems().add(menuItem);
+                    });
+                }
+                Platform.runLater( () -> {
```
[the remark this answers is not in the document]
johanvos: No, it is needed because we want to give the JavaFX platform the opportunity to do "whatever is needed" after the last runLater() is done. Hence, we put what we want to do on the Runnable queue of the FX thread, so that we don't depend on internals.

tests/system/src/test/java/test/javafx/stage/SystemMenuBarTest.java
```diff
@@ -108,4 +115,64 @@ public MenuBar createFailingMenuBar() {
...
+        final Menu menu = new Menu("MyMenu");
+        menuBar.getMenus().add(menu);
+        menuBar.setUseSystemMenuBar(true);
+        root.getChildren().add(menuBar);
+
+        Scene scene = new Scene(root);
+        stage.setScene(scene);
+        stage.show();
+        stage.requestFocus();
+        Thread t = new Thread(){
```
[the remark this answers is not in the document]
johanvos: I'll go with the whitespace then, as I find it much more clear to explicitly construct, assign, and start a thread.

jperedadnr (approved):

kevinrushforth (approved):

tests/system/src/test/java/test/javafx/stage/SystemMenuBarTest.java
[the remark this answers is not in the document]
kevinrushforth: This makes sense. To answer one of Jose's questions:

> Is it guaranteed that it will be called only after the previous one ends?

Yes, this is specified behavior of `Platform::runLater`.

PR #1281, opened 2023-11-08
8092272: [D3D 3D] Need a robust 3D states management for texture

nlisker:
Moves the filter setting of the samplers from the device parameters configuration to the use-site, allowing for dynamic changes in the sampler. This PR does internal plumbing work only to bring it close to the ES2 pipeline. A followup PR will create the public API.

Summary of the changes:
* Created a new (internal for now) `TextureData` object that is intended to contain all the data of texture (map) of `PhongMaterial`, such as filters, addressing, wrapping mode, mipmaps etc. **This PR deals only with filters** as a starting point, more settings can be added later.
* Creates an update mechanism from the Java side material to the native D3D layer. The public API `PhoneMaterial` is *not* changed yet. The peer `NGPhongMaterial` is configured to receive update from the public `PhongMaterial` when the public API is created via new `ObjectProperty<TextureData>` properties.
* Small refactoring in the D3D layer with a new map types enum to control the texture settings more easily.

The JBS issue lists some regressions in a comment, but I couldn't reproduce them. It looks like the sampler settings needed to be added anywhere, and that was the easiest to do at the time. Now they were just moved.

---------
### Progress

### Issue
 * [JDK-8092272](https://bugs.openjdk.org/browse/JDK-8092272): [D3D 3D] Need a robust 3D states management for texture (**Enhancement** - P5)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1281/head:pull/1281` \
`$ git checkout pull/1281`

`$ git checkout pull/1281` \
`$ git pull https://git.openjdk.org/jfx.git pull/1281/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1281`

`$ git pr show -t 1281`

</details>

<a href="https://git.openjdk.org/jfx/pull/1281.diff">https://git.openjdk.org/jfx/pull/1281.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1281#issuecomment-1802921996)

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/scene/paint/TextureData.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/scene/paint/TextureData.java
@@ -1,0 +1,110 @@
+package com.sun.javafx.scene.paint;
+
+// When the public API is ready, move to the exported javafx.scene.paint
+/**
+ * Contains texture properties for a {@code Material}'s textures (maps).
+ * <p>
+ * Since {@code TextureData} does not depend on a {@code Material}, the same instance can be reused in multiple textures
+ * (in the same or in different materials). This is useful because many times a {@link javafx.scene.shape.Shape3D} will
+ * have all its textures use the same properties, or the same texture data will be used across the whole application.
+ */
+// Here we can support mipmaps, wrapping modes (which exists internally and can be pulled out), addressing modes etc.
+public class TextureData {
+
+    /**
+     * Filtering methods for sampling a magnified or minified texture. When a texture is used to color a surface that is
+     * larger or smaller than the texture size, the texture needs to be magnified or minified. Different algorithms
+     * exist for sampling the texture elements (texels) used to color the surface.
+     */
+    public enum MinMagFilterType {
+
+        /**
+         * Uses the texel closest to the sampling point. This sampling method is the fastest, but the results can be
+         * pixelated.
+         */
+        NEAREST_POINT,
+
+        /**
+         * Uses a weighted average of the 4 texels closest to the sampling point. This sampling method is slightly
+         * slower than {@code NEAREST_POINT}, but results in a smoother image.
+         */
+        BILINEAR;
+    }
+
+    /**
+     * Filtering methods for sampling between mipmaps. When a surface is at a distance that does not match a single
+     * mipmap different algorithms exist for interpolating between the closest mipmap levels. Mipmapping is only valid
+     * for minification since the base texture is the largest size image.
+     */
+    public enum MipmapFilterType {
+
+        /**
+         * No mimaping is used.
+         */
+        NONE,
+        /**
+         * Chooses the nearest mipmap level. This is a cheap method, but can result in abrupt changes when the mipmap
+         * level is switched.
+         */
+        NEAREST,
+        /**
+         * Linearly interpolates between the 2 closest mipmap levels. This method is slightly slower than
+         * {@code NEAREST}, but the result is a smooth transition between mipmap levels.
+         */
+        LINEAR;
+    }
+
+    private final MinMagFilterType minFilterType;
+    private final MinMagFilterType magFilterType;
+    private final MipmapFilterType mipmapFilterType;
+
+    private TextureData(MinMagFilterType minFilterType, MinMagFilterType magFilterType, MipmapFilterType mipmapFilterType) {
+        this.minFilterType = minFilterType;
+        this.magFilterType = magFilterType;
+        this.mipmapFilterType = mipmapFilterType;
+    }
+
+    public MinMagFilterType minFilterType() { return minFilterType; }
+    public MinMagFilterType magFilterType() { return magFilterType; }
+    public MipmapFilterType mipmapFilterType() { return mipmapFilterType; }
+
+    public static class Builder {
+
+        private MinMagFilterType minFilterType = MinMagFilterType.BILINEAR;
+        private MinMagFilterType magFilterType = MinMagFilterType.BILINEAR;
+        private MipmapFilterType mipmapFilterType = MipmapFilterType.NONE;
+
+        public Builder() {}
+
+        public Builder minFilterType(MinMagFilterType minFilterType) {
+            this.minFilterType = minFilterType;
+            return this;
+        }
+        
+        public Builder magFilterType(MinMagFilterType magFilterType) {
+            this.magFilterType = magFilterType;
+            return this;
+        }
+        
+        public Builder mipmapFilterType(MipmapFilterType mipmapFilterType) {
+            this.mipmapFilterType = mipmapFilterType;
+            return this;
+        }
+        
+        public TextureData build() {
+            return new TextureData(minFilterType, magFilterType, mipmapFilterType);
+        }
+    }
+
+    /**
+     * Mipmap images to be used when mipmapping is enabled. The images in this list must each be a power-of-2 smaller size
+     * than the previous one. The image used as the main texture (level 0) should not be included.
+     * To enable mipmapping, set {@link MipmapFilterType} to a value other than {@link MipmapFilterType#NONE NONE}.
+     */
+//    private ObservableList<Image> mipmaps;
+//
+//    private int mipmapLevels;
+
+//    enum WrapMode { // see com.sun.prism.Texture.WrapMode
+//    }
+}
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/scene/paint/TextureData.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/scene/paint/TextureData.java
@@ -80,17 +80,17 @@ public class TextureData {
             this.minFilterType = minFilterType;
             return this;
         }
-        
+
         public Builder magFilterType(MinMagFilterType magFilterType) {
             this.magFilterType = magFilterType;
             return this;
         }
-        
+
         public Builder mipmapFilterType(MipmapFilterType mipmapFilterType) {
             this.mipmapFilterType = mipmapFilterType;
             return this;
         }
-        
+
         public TextureData build() {
             return new TextureData(minFilterType, magFilterType, mipmapFilterType);
         }
--- a/modules/javafx.graphics/src/main/native-prism-d3d/D3DContext.cc
+++ b/modules/javafx.graphics/src/main/native-prism-d3d/D3DContext.cc
@@ -626,10 +626,10 @@ HRESULT D3DContext::setDeviceParametersFor3D() {
         SUCCEEDED(res = pd3dDevice->SetRenderState(D3DRS_ALPHABLENDENABLE, TRUE)) &&
         SUCCEEDED(res = pd3dDevice->SetRenderState(D3DRS_SCISSORTESTENABLE, FALSE)) &&
         SUCCEEDED(res = pd3dDevice->SetRenderState(D3DRS_LIGHTING, TRUE)) &&
-        // TODO: 3D - RT-34415: [D3D 3D] Need a robust 3D states management for texture
         // Set texture unit 0 to its default texture addressing mode for Prism
+        // TODO: Consider giving these the same treatment as setting the filters in D3DContext::nSetMap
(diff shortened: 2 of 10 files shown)


modules/javafx.graphics/src/main/native-prism-d3d/D3DPhongMaterial.cc
```diff
@@ -68,35 +67,41 @@ float * D3DPhongMaterial::getSpecularColor() {
 }
 
 bool D3DPhongMaterial::isBumpMap() {
-    return map[BUMP] ? true : false;
+    return map[map_type::bump] ? true : false;
```
mstr2, line 70: Since `map` contains pointers, you might consider `return map[map_type::bump] != NULL`.

modules/javafx.graphics/src/main/native-prism-d3d/D3DMeshView.cc
```diff
@@ -289,19 +289,22 @@ void D3DMeshView::render() {
...
     }
 
-    SUCCEEDED(device->SetTexture(SR_DIFFUSE_MAP, material->getMap(DIFFUSE)));
-    SUCCEEDED(device->SetTexture(SR_SPECULAR_MAP, material->getMap(SPECULAR)));
-    SUCCEEDED(device->SetTexture(SR_BUMPHEIGHT_MAP, material->getMap(BUMP)));
-    SUCCEEDED(device->SetTexture(SR_SELFILLUM_MAP, material->getMap(SELFILLUMINATION)));
+    // Set the textures and their filters in their corresponding samplers
+    for (int i = 0; i < map_type::num_map_types; i++) {
+        map_type type = static_cast<map_type>(i);
+        SUCCEEDED(device->SetTexture(type, material->getMap(type)));
```
mstr2, line 295: Wrapping the calls with the `SUCCEEDED` macro doesn't do anything useful.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/paint/TextureData.java
```diff
@@ -0,0 +1,110 @@
...
+// When the public API is ready, move to the exported javafx.scene.paint
+/**
+ * Contains texture properties for a {@code Material}'s textures (maps).
+ * <p>
+ * Since {@code TextureData} does not depend on a {@code Material}, the same instance can be reused in multiple textures
+ * (in the same or in different materials). This is useful because many times a {@link javafx.scene.shape.Shape3D} will
+ * have all its textures use the same properties, or the same texture data will be used across the whole application.
+ */
+// Here we can support mipmaps, wrapping modes (which exists internally and can be pulled out), addressing modes etc.
+public class TextureData {
```
mstr2, line 12: 1. This type could be a record.
2. There's no actual texture data (i.e. pixels) here, maybe a better name would be `SamplerParameters` or something like that.

modules/javafx.graphics/src/main/native-prism-d3d/D3DContext.cc
```diff
@@ -352,21 +352,29 @@ JNIEXPORT void JNICALL Java_com_sun_prism_d3d_D3DContext_nSetSpecularColor
...
  * Method:    nSetMap
  * Signature: (JJIJ)V
  */
 JNIEXPORT void JNICALL Java_com_sun_prism_d3d_D3DContext_nSetMap
   (JNIEnv *env, jclass, jlong ctx, jlong nativePhongMaterial,
-        jint mapType, jlong nativeTexture)
+        jint mapType, jint minFilter, jint magFilter, jint mipFilter, jlong nativeTexture)
 {
+    static const D3DTEXTUREFILTERTYPE minMagEnumMap[2] = { D3DTEXF_POINT, D3DTEXF_LINEAR };
+    static const D3DTEXTUREFILTERTYPE mipmapEnumMap[3] = { D3DTEXF_NONE, D3DTEXF_POINT, D3DTEXF_LINEAR };
```
mstr2, line 369: You could use `std::array` as a safer alternative for plain arrays.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/paint/TextureData.java
nlisker: 1. Not when it will be promoted to public API. Adding record components breaks backwards compatibility, so making this a record will not allow adding more configuration later on. What might be possible is making it an interface and using a record as an implementation.
2. I think that this class can go beyond sampler configurations. For example, wrapping is a render state, not a sampler state. Maybe something more general like "TextureParameters" can work.

modules/javafx.graphics/src/main/native-prism-d3d/D3DMeshView.cc
nlisker: Yes, but it's done everywhere else in the class, and was done for these calls before I changed this code. Maybe a better error handling design is needed.

modules/javafx.graphics/src/main/native-prism-d3d/D3DPhongMaterial.h
```diff
@@ -30,10 +30,15 @@
...
-#define SELFILLUMINATION 3
+// Map type numbered sequence used for sampling registers (we have 4 sampling registers for vs 3.0).
+// Order is defined by com.sun.prism.PhongMaterial's MapType enum
+enum map_type : unsigned int {
+    diffuse,
+    specular,
+    bump,
+    self_illumination,
+    num_map_types
+};
```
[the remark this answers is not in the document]
nlisker: About 2, I tried that initially and got into problems iterating over the enum in `D3DMeshView.cc` line 293. How would you perform the iteration?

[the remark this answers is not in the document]
mstr2: You're right, there's no implicit conversion from scoped enums to integers. While one could get very creative with C++, maybe just using an unscoped enum is easier. You might want to prefix the constants with `MT_` or something like that.

kevinrushforth:
Reviewers: @arapte @lukostyra

modules/javafx.graphics/src/main/native-prism-d3d/D3DMeshView.cc
```diff
@@ -289,19 +289,22 @@ void D3DMeshView::render() {
...
 
     if (context->state.cullMode != cullMode) {
         context->state.cullMode = cullMode;
         SUCCEEDED(device->SetRenderState(D3DRS_CULLMODE, D3DCULL(cullMode)));
     }
     if (context->state.wireframe != wireframe) {
         context->state.wireframe = wireframe;
-        SUCCEEDED(device->SetRenderState(D3DRS_FILLMODE,
-                wireframe ? D3DFILL_WIREFRAME : D3DFILL_SOLID));
+        SUCCEEDED(device->SetRenderState(D3DRS_FILLMODE, wireframe ? D3DFILL_WIREFRAME : D3DFILL_SOLID));
```
lukostyra, line 307: I think this formatting change should not be a part of this PR.

lukostyra (changes requested):

modules/javafx.graphics/src/main/native-prism-d3d/D3DPhongMaterial.cc
```diff
@@ -34,10 +34,9 @@ using std::endl;
 D3DPhongMaterial::~D3DPhongMaterial() {
     context = NULL;
     // The freeing of texture native resources is handled by its Java layer.
-    map[DIFFUSE] = NULL;
-    map[SPECULAR] = NULL;
-    map[BUMP] = NULL;
-    map[SELFILLUMINATION] = NULL;
+    for (int i = 0; i < map_type::num_map_types; i++) {
+        map[i] = NULL;
+    }
```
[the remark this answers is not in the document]
nlisker: There is all sorts of unused code in the native layer. I think that there are whole classes that aren't needed. It will require another PR.

modules/javafx.graphics/src/main/native-prism-d3d/D3DPhongMaterial.h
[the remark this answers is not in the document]
mstr2: The `BUMP` symbol is not scoped to `MapType`; prefixing with `MapType::` is neither required, nor changes its meaning. The symbol exists in the global namespace in all compilation units that include the header. Adding a prefix may help to avoid name collisions in the future.

modules/javafx.graphics/src/main/native-prism-d3d/D3DContext.cc
```diff
@@ -352,21 +353,26 @@ JNIEXPORT void JNICALL Java_com_sun_prism_d3d_D3DContext_nSetSpecularColor
...
  * Method:    nSetMap
  * Signature: (JJIJ)V
  */
 JNIEXPORT void JNICALL Java_com_sun_prism_d3d_D3DContext_nSetMap
   (JNIEnv *env, jclass, jlong ctx, jlong nativePhongMaterial,
-        jint mapType, jlong nativeTexture)
+        jint mapType, jint minFilter, jint magFilter, jint mipFilter, jlong nativeTexture)
 {
+    static const std::array<D3DTEXTUREFILTERTYPE, 2> minMagEnumMap = { D3DTEXF_POINT, D3DTEXF_LINEAR };
+    static const std::array<D3DTEXTUREFILTERTYPE, 3> mipmapEnumMap = { D3DTEXF_NONE, D3DTEXF_POINT, D3DTEXF_LINEAR };
```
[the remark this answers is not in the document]
nlisker: They are used only in this method. Why would like them to be moved out?

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/paint/TextureData.java
nlisker: `ImageSmoothing` is very specific. This class can include many configuration options, not only from the sampler, but also from the rendered. I'm not sure what you're suggesting.

Maybe I should try to list more of the future options for this class.

nlisker:
Keep open.

WasabiThumb:
Paying a visit here after writing [this SO answer](https://stackoverflow.com/a/79706844/10808904), any reason this PR died? Appreciate any and all responses.

nlisker:
> Paying a visit here after writing [this SO answer](https://stackoverflow.com/a/79706844/10808904), any reason this PR died? Appreciate any and all responses.

The PR is not really dead, it's hibernating :) The reason is that there's more research that needs to go into this. Initially, the texture data was only for the sampler state, but then it expanded to the render state, and it's not clear if this is correct and how inclusive this interface should be.

Add to it that Metal and D3D12 are in the works and we need to make sure JavaFX's abstraction fits those as well (and it's not clear now that it needs to fit D3D9). Writing an abstraction API over 4 or 5 different pipelines is something that takes a lot of consideration.

Note also that this PR is just for the implementation, and doesn't include exposing the API, which is planned for a followup PR.

If you'd like to help, you can research the sampler states and render states (and whatever else might fit) that the pipelines support and see if it makes sense to bundle all of them into a "texture parameters" type.

WasabiThumb:
The sampler option structures for [OpenGL](https://www.khronos.org/opengl/wiki/Sampler_Object#Sampling_parameters), [D3D9](https://learn.microsoft.com/en-us/windows/win32/direct3d9/d3dsamplerstatetype), [D3D12](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_sampler_desc) and [Metal](https://developer.apple.com/documentation/metal/mtlsamplerdescriptor) seem fairly congruent, albeit mangled. I agree that writing a comprehensive abstraction for even these few pipelines remains challenging. It's not clear whether the [D3D12 filter type](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_filter) is some kind of bitmask and how precisely it relates to the [D3D9 filter type](https://learn.microsoft.com/en-us/windows/win32/direct3d9/d3dtexturefiltertype), especially where mips are concerned. I'm also not sure whether information about sampler objects in OpenGL 3.3 is applicable to OpenGL ES 2.0. It does at least have [glTexParameter](https://registry.khronos.org/OpenGL-Refpages/es2.0/xhtml/glTexParameter.xml), the most limited of the APIs by far.

nlisker:
And there are also render states here to take into account.

PR #1283, opened 2023-11-10
8319779: SystemMenu: memory leak due to listener never being removed

johanvos:
A listener was added but never removed.
This patch removes the listener when the menu it links to is cleared. Fix for https://bugs.openjdk.org/browse/JDK-8319779

---------
### Progress

### Issues
 * [JDK-8319779](https://bugs.openjdk.org/browse/JDK-8319779): SystemMenu: memory leak due to listener never being removed (**Bug** - P4)
 * [JDK-8323787](https://bugs.openjdk.org/browse/JDK-8323787): Mac System MenuBar throws IOB exception (**Bug** - P3)

### Reviewers
 * [Jose Pereda](https://openjdk.org/census#jpereda) (@jperedadnr - **Reviewer**) 🔄 Re-review required (review applies to [55dc803f](https://git.openjdk.org/jfx/pull/1283/files/55dc803f0c54017fd3ac4086b99ef28c85e1a943))

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1283/head:pull/1283` \
`$ git checkout pull/1283`

`$ git checkout pull/1283` \
`$ git pull https://git.openjdk.org/jfx.git pull/1283/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1283`

`$ git pr show -t 1283`

</details>

<a href="https://git.openjdk.org/jfx/pull/1283.diff">https://git.openjdk.org/jfx/pull/1283.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1283#issuecomment-1805482784)


modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassSystemMenu.java
```diff
@@ -148,7 +160,33 @@ private void insertMenu(final Menu parent, final MenuBase mb, int pos) {
...
-        filteredItems.addListener((ListChangeListener.Change<? extends MenuItemBase> change) -> {
+        ListChangeListener menuItemListener = createListener(glassMenu);
+        filteredItems.addListener(menuItemListener);
+        menuListeners.put(glassMenu, menuItemListener);
+        listenerItems.put(menuItemListener, filteredItems);
+
+        for (MenuItemBase item : items) {
+            if (item instanceof MenuBase) {
+                // submenu
+                addMenu(glassMenu, (MenuBase)item);
```
hjohn, line 171:
```suggestion
            if (item instanceof MenuBase baseItem) {
                // submenu
                addMenu(glassMenu, baseItem);
```

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassSystemMenu.java
```diff
@@ -148,7 +160,33 @@ private void insertMenu(final Menu parent, final MenuBase mb, int pos) {
...
+
+        if (parent != null) {
+            parent.insert(glassMenu, pos);
+        } else {
+            glassSystemMenuBar.insert(glassMenu, pos);
+        }
+    }
+
+    private ListChangeListener createListener(final Menu glassMenu) {
+        ListChangeListener<MenuItemBase> answer = ((ListChangeListener.Change<? extends MenuItemBase> change) -> {
```
hjohn, line 189: minor: I see extra parenthesis around the whole, and I think you can just `return` this immediately (no need for `answer` local)

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassSystemMenu.java
```diff
@@ -57,6 +59,8 @@ class GlassSystemMenu implements TKSystemMenu {
 
     private List<MenuBase>      systemMenus = null;
     private MenuBar             glassSystemMenuBar = null;
+    private final Map <Menu, ListChangeListener> menuListeners = new HashMap<>();
+    private final Map <ListChangeListener, ObservableList> listenerItems  = new HashMap<>();
```
hjohn, line 63: minor:
```suggestion
    private final Map<Menu, ListChangeListener> menuListeners = new HashMap<>();
    private final Map<ListChangeListener, ObservableList> listenerItems = new HashMap<>();
```

hjohn (commented):
Should there be an updated test for this?

I see a lot of raw type use (`ListChangeListener`, `ObservableList`), any chance those can be avoided?

kevinrushforth:
@arapte can you be the primary reviewer for this?

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassSystemMenu.java
johanvos: I typically add a local variable for this (which is removed by javac) to make it easier to add debugging, but I'll remove it.

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassSystemMenu.java
johanvos: done

arapte:
Looks all good to me.
Is it possible to add an automated test ?

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassSystemMenu.java
```diff
@@ -148,7 +160,33 @@ private void insertMenu(final Menu parent, final MenuBase mb, int pos) {
...
+
+        if (parent != null) {
+            parent.insert(glassMenu, pos);
+        } else {
+            glassSystemMenuBar.insert(glassMenu, pos);
+        }
+    }
+
+    private ListChangeListener<MenuItemBase> createListener(final Menu glassMenu) {
+        return (ListChangeListener.Change<? extends MenuItemBase> change) -> {
```
jperedadnr, line 189: In this listener, when there are items removed that are `Menu`s, we should also call  `clearMenu`:
```
             if (i >= 0 && menuItemList.size() > i) {
                        glassMenu.remove(i);
+                        Object item = menuItemList.get(i);
+                        if (item instanceof Menu menu) {
+                                clearMenu(menu);
+                        }
                 }
```

Otherwise, they won't be removed from the new maps.

jperedadnr:
About adding an automated test, the leak that PR tries to fix happens in `com.sun.javafx.tk.quantum.GlassSystemMenu`, which is package private, and adding new tests to check `javafx.scene.control.Menu`, or related public classes, won't catch that leak, will it?

Unless we open it (and all the related classes/methods in `com.sun.glass.ui.*`), I don't think it can be done?

johanvos:
As for the testing issue: the memoryleak before this fix leads to an increasing number of `com.sun.glass.ui.MenuItem` instances.
We could test the previous memoryleak, as that was creating an increasing number of `javafx.scene.control.MenuItem` instances, which could be handled inside a JavaFX application and stored as WeakReferences in a list, that could then be required to be empty after GC.
Our systemtest does not have access to the `com.sun.glass.ui.MenuItem` instances, so we can't use the same approach. In theory, it should be possible to mock the whole path that leads to this leak, but it is complex.
I do understand the value of tests for this, though, as the fix is brittle. When a new InvalidationListener is added referencing the to-be-removed menuItems, the memoryleak will re-occur and go unnoticed.

johanvos:
Moving this to [WIP] while I'm working on automated tests.

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassSystemMenu.java
```diff
@@ -169,32 +208,14 @@ private void insertMenu(final Menu parent, final MenuBase mb, int pos) {
...
     }
 
+
     private void setMenuBindings(final Menu glassMenu, final MenuBase mb) {
-        mb.textProperty().addListener(valueModel -> glassMenu.setTitle(parseText(mb)));
-        mb.disableProperty().addListener(valueModel -> glassMenu.setEnabled(!mb.isDisable()));
-        mb.mnemonicParsingProperty().addListener(valueModel -> glassMenu.setTitle(parseText(mb)));
+        mb.textProperty().addListener(new WeakInvalidationListener(valueModel -> glassMenu.setTitle(parseText(mb))));
+        mb.disableProperty().addListener(new WeakInvalidationListener(valueModel -> glassMenu.setEnabled(!mb.isDisable())));
+        mb.mnemonicParsingProperty().addListener(new WeakInvalidationListener(valueModel -> glassMenu.setTitle(parseText(mb))));
```
hjohn, line 218: Isn't this the classic mistake being made with weak listeners where you are wrapping a listener, but then are not referring it?  The inner listener (which is created on the fly) is only weakly referenced, and so can go out of scope immediately.

hjohn:
Would it be an idea to do deterministic clean-up?

The JIRA tickets involved are light on details, and I don't have a Mac, so I can't test the leak (I think?), but it seems to me that you will want to remove the listeners when the menu bar isn't visible (and re-add them if it becomes visible again).

From what I understand, these are normal JavaFX controls wrapped into adapters, and so any listeners installed through the adapter interface will end up on controls.  The `GlobalMenuAdapter` seems to be tracking the `showing` property, and calls `show` / `hide` accordingly.

The ticket says:

> when an application loses focus, the SystemMenu is removed. When it gets focus again, the SystemMenu is "back".

My question is, is there any event that can be used to trigger clean-up when this happens?  Do the controls go invisible, are they no longer showing (is `hide` called?)  There must be something.  If there is something, then some kind of disposal can be triggered for deterministic clean-up (ie. add `dispose` method).

If you can go further and distill this information into a boolean property then a pattern like this can be used:

        mb.textProperty()
             .when(showingProperty)  // assumes showing property becomes false when focus is lost...
             .addListener(valueModel -> glassMenu.setTitle(parseText(mb)));

The `showingProperty` can be fed from the hide/show event handlers to set it to the right value, perhaps combined with focus information.

johanvos:
@hjohn You're absolutely right. It's another reason why I moved this PR to [WIP]. Instead of guessing and trying, I want to understand the deterministic path that leads to the issue with listeners being added and not removed. The one fixed in the first commit is clear, and well understood. Still, I saw uncollected listeners, with the GC root trace pointing to those other listeners (textProperty, mnemonicParsingProperty etc). Rather than trying to "fix" this with weak listeners, it is much better to find the real cause for this.
In the process of finding out, I'm adding a few shim classes and unit tests so that we can clearly test for regression.

johanvos:
As for the memory leak issue: there were several. The first commit in this PR fixed a clear memory leak, but the one that is still left is not described in the issue. 
It occurs because whenever the SystemMenuBar is shown after it was not shown, all MenuItems will be readded using `GlassSystemMenu.setMenu`. This will invoke `GlassSystemMenu.insertMenuItem(... MenuItemBase menuItem...)` 
Inside this method, platform-specific menuItems will be created, that will adapt when the `MenuItemBase` properties are changing. The annoying thing, as stated in the first comment of that method, is that we don't know if the passed `MenuItemBase` instance was previously used in this method (in which case the listeners are regisered to this instance) or not. That is why for example the `visibilityListener` is unconditionally removed (even if we don't know if it was ever added) and then added. 
We can do the same for the other properties (preventing the use of weak listeners), but it would be nice if we could address this case using the api introduced in the deterministic listeners management (I would very much prefer that over weak listeners).
I'm not too familiar with the new API though, but it seems difficult to do this _inside_ this method, as we would still add a listener each time the method is invoked with an item that already has a listener registered?

hjohn:
> As for the memory leak issue: there were several. The first commit in this PR fixed a clear memory leak, but the one that is still left is not described in the issue. It occurs because whenever the SystemMenuBar is shown after it was not shown, all MenuItems will be readded using `GlassSystemMenu.setMenu`. This will invoke `GlassSystemMenu.insertMenuItem(... MenuItemBase menuItem...)` Inside this method, platform-specific menuItems will be created, that will adapt when the `MenuItemBase` properties are changing. The annoying thing, as stated in the first comment of that method, is that we don't know if the passed `MenuItemBase` instance was previously used in this method (in which case the listeners are regisered to this instance) or not. That is why for example the `visibilityListener` is unconditionally removed (even if we don't know if it was ever added) and then added. We can do the same for the other properties (preventing the use of weak listeners), but it would be nice if we could address this case using the api introduced in the deterministic listeners management (I would very much prefer that over weak listeners). I'm not too familiar with the new API though, but it seems difficult to do this _inside_ this method, as we would still add a listener each time the method is invoked with an item that already has a listener registered?

It's still a bit unclear for me.  So what I write below may be off the mark.

From what I can see, `setMenus` will first remove all items, then re-add them.  The `clearMenu` function seems like a logical place to remove listeners that may be present.  The problem is where to keep track off the listeners.  I see a few options:

1. Track the listeners in a few `Map`s in `GlassSystemMenu`, something like `Map<MenuBase, InvalidationListener>`, with a different `Map` for each listener type (text, disabled, mnemonic, etc).  Or something more elaborate with a listener holder, nested maps, etc.

These maps can then iterated when `setMenus` is called to clear out ALL the listeners (or as part of `clearMenu`.

2. Similar to above but store the listener information in the properties of `Menu`

3. Using the new resource management system. Attach all listeners with a `when` condition based on a `SimpleBooleanProperty` that you **replace** each time `setMenus` is called (replacement is needed in this case as we don't know if the menus were used before or not, but it is still relatively clean).

The last one works like this:

Create a new non-final field: `BooleanProperty active`

In `setMenus` do:

      if (active != null) {
            active.set(false);  // this releases all listeners attached with `when(active)`
      }

      active = new SimpleBooleanProperty(true);  // replace active with a new property, as we can't reuse it

Then attach all listeners conditionally like so:

      mb.textProprty().when(active).addListener( ... );

The idea here is that because `active` is a different property each time `setMenus` is called, all old listeners will be removed when active goes to `false`, and will be eligible for GC because the `active` property was also replaced (and the old one, which we just set to `false`, is not referenced anywhere anymore).

Illustration:

image

johanvos:
Thanks @hjohn for the clear explanation. I replaced the Weak listeners with the new listener resource management approach, and that works fine.

I'm still looking into adding a test. The challenge is that the (previously) uncollected items are of type `com.sun.glass.ui.MenuItem`, and afaik we can't use JMemoryBuddy to check the collectable state of instances on the heap that we don't have a reference to ourselves. I see 2 options:
* a unit test, with new shim classes so that we get access to `com.sun.glass.ui.MenuItem` from the test. The problem here is that this quickly becomes a test for the shim class and not for the real classes
* a systemtest that can somehow inspect the heap and count the instances of `com.sun.glass.ui.MenuItem`

hjohn:
> I'm still looking into adding a test. The challenge is that the (previously) uncollected items are of type `com.sun.glass.ui.MenuItem`, and afaik we can't use JMemoryBuddy to check the collectable state of instances on the heap that we don't have a reference to ourselves. I see 2 options:
> 
> * a unit test, with new shim classes so that we get access to `com.sun.glass.ui.MenuItem` from the test. The problem here is that this quickly becomes a test for the shim class and not for the real classes
> * a systemtest that can somehow inspect the heap and count the instances of `com.sun.glass.ui.MenuItem`

Perhaps add a package private method to get access to `GlassSystemMenu#systemMenus` (or use reflection)?  I think you could then put a weak reference on the returned list (or items) and see if that disappears.

johanvos:
I added a systemtest using a new Shim (`GlassSystemMenuShim`) that keeps track of the `com.sun.glass.ui.Menu` instances that are created in the `GlassSystemMenu`. I had to do a few more hacks since you don't normally access `GlassSystemMenu` yourself (the MenuBarSkin is doing most of the work).
Note that this is very platform-dependent, as the creation of the `com.sun.glass.ui.Menu` is done by the `com.sun.glass.ui.Application`

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassSystemMenu.java
```diff
@@ -85,6 +92,10 @@ protected MenuBar getMenuBar() {
     }
 
     @Override public void setMenus(List<MenuBase> menus) {
+        if (active != null) {
+            active.set(false);
+        }
+        active = new SimpleBooleanProperty(true);
```
andy-goryachev-oracle, line 98: should this be
```
if(active == null) {
  active = new ...
} else {
 active.set(false);
}
```
instead?

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassSystemMenu.java
```diff
@@ -169,32 +214,14 @@ private void insertMenu(final Menu parent, final MenuBase mb, int pos) {
...
 
-    private void setMenuBindings(final Menu glassMenu, final MenuBase mb) {
-        mb.textProperty().addListener(valueModel -> glassMenu.setTitle(parseText(mb)));
-        mb.disableProperty().addListener(valueModel -> glassMenu.setEnabled(!mb.isDisable()));
-        mb.mnemonicParsingProperty().addListener(valueModel -> glassMenu.setTitle(parseText(mb)));
+
+    protected void setMenuBindings(final Menu glassMenu, final MenuBase mb) {
+        mb.textProperty().when(active).addListener(valueModel -> glassMenu.setTitle(parseText(mb)));
+        mb.disableProperty().when(active).addListener(valueModel -> glassMenu.setEnabled(!mb.isDisable()));
+        mb.mnemonicParsingProperty().when(active).addListener(valueModel -> glassMenu.setTitle(parseText(mb)));
```
andy-goryachev-oracle, line 224: is `active` guaranteed to be non-null at this point (and also on L279)?

assuming the code on L98 is fixed and does not clobber the old pointer.

tests/system/src/test/java/test/com/sun/javafx/tk/quantum/SystemMenuBarTest.java
```diff
@@ -172,4 +179,77 @@ private void createMenuBarWithItemsStage() {
...
+                gsmh.setMenus(List.of(menuBase));
+            });
+        }
+        Platform.runLater(() -> {
+            int strongCount = 0;
+            final List<WeakReference<com.sun.glass.ui.Menu>> u2 = gsmh.getWeakMenuReferences();
+            for (WeakReference<com.sun.glass.ui.Menu> wr : u2) {
+                if (!JMemoryBuddy.checkCollectable(wr)) {
+                    strongCount++;
+                    assertTrue("Too much refs", strongCount < 2);
```
andy-goryachev-oracle, line 227: "too many" ?

tests/system/src/test/java/test/com/sun/javafx/tk/quantum/SystemMenuBarTest.java
```diff
@@ -172,4 +179,77 @@ private void createMenuBarWithItemsStage() {
...
+        systemMenu.getItems().add(newItem);
+
+        Platform.runLater(() -> {
+            javafx.scene.control.Menu systemMenuContributions = new Menu("123");
+            systemMenu.getItems().add(systemMenuContributions);
+        });
+
+        return menuBar;
+    }
+
```
andy-goryachev-oracle, line 254: extra newline?

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassSystemMenu.java
johanvos: No, the explicit goal of this construction is to set active to false (in case it exists) so that existing listeners can be released; followed by creating a new active property that is by default set to true until `setMenus` is called again (e.g. after unfocus/focus events).
I noticed however that it is not as simple as that, and the current PR introduces regression, as the `SystemMenuBarTest.testMemoryLeak` test is now failing. Listeners are not only registered when `setMenus` is called, but also when menu(Item)s are added to menus. I'm not sure if a single `active` property can address this, as we don't want to remove all listeners for all menuitems in case a particular menuitem is added to a particular menu.
@hjohn does that make sense?

andy-goryachev-oracle: thank you for clarification!  

perhaps we ought to add a comment explaining why: it's one thing to create a single chain of conditions linked to a property, but here we keep creating new property instance, so the side effects are not that obvious.

hjohn: > I noticed however that it is not as simple as that, and the current PR introduces regression, as the SystemMenuBarTest.testMemoryLeak test is now failing. Listeners are not only registered when setMenus is called, but also when menu(Item)s are added to menus. I'm not sure if a single active property can address this, as we don't want to remove all listeners for all menuitems in case a particular menuitem is added to a particular menu.
@hjohn does that make sense?

So, if I understood correctly, that test is dealing with not with the entire menu bar disappearing (for which the `active` property ensures clean up), but with a submenu getting cleared, and then a new item added.  Any old items that were added should be garbage collectable after clearing.

This is a bit tricky.  Basically I think it means that the listeners on a certain `MenuBase` wrapper should be removed in two cases:

1. When the entire menu is no longer needed (what we have now)
2. When the `MenuBase` in question is removed from its parent

I think this may be resolved in one of two ways:

1. Create a specific active condition for `when` that includes a check on parent and on the top level `active` property
2. Gather all listeners added into a single `Subscription` (currently the `Subscription` that is returned is discarded, track this somewhere (probably easiest in a the user property map) taking care not to create a new hard reference (not sure if that will be an issue).  The `Subscription` can then be obtained and unsubscribed early.

Now I think option 1 can be achieved by doing something like the code below.

**Note:** completely untested, and did not check yet if we aren't creating any new problematic hard references; I can test this out over the weekend if you like.

     // given the global `active` property...
     // and a MenuBase mb, for which we can discover if it is a root menu or not...
     // and which exposes a new property hasParentProperty (you can derive this property
     // if you like by doing `parentMenuProperty.map(Objects::nonNull)`)

     BooleanProperty forWhen = active;

     if (mb "is not a root menu") {  // important we don't do this for root
           // include "and" condition that it also must have a parent to be considered "active"
           forWen = active.flatMap(b -> b ? mb.hasParentProperty() : active);
     }

     // then use `forWhen` in your `when` conditions instead of `active` as before.

What the above does is that using `flatMap` it creates a listener on `hasParentProperty`.  If it changes, the entire expression result may change.  In normal Java code it would simply be:

     boolean forWhen = active && mb.hasParent();

... except of course that the above code would not react to changes.

The logical and (`&&`) is achieved with the ternary expression.  If `b` is `true` then we need to check the second part of the `&&` which is `hasParentProperty`.  If `b` was already `false`, then the there is no need to check further.  In any case, a new **property** must be returned from `flatMap`, so that will either be `hasParentProperty` in case `active` is `true` or the `active` property (which then must be `false`).

Such scenario's are rare, but I've been considering adding specific support on `ObservableValue` for this.

There is also the `and` function on `ObservableBooleanValue` -- you can test this one as well, but I fear that it may be using a weak listener in there and that it may get collected early if you don't reference the intermediate expressions.  There is no such risk with `flatMap`.

I haven't looked further into option 2.

hjohn: > perhaps we ought to add a comment explaining why: it's one thing to create a single chain of conditions linked to a property, but here we keep creating new property instance, so the side effects are not that obvious.

Yeah, the replacing of the property used in `when` is not a normal scenario.  Normally, the property used in the `when` goes out of scope at the same time as the rest of the references, in cases like:

       label.textProperty().when(userInterfaceIsShowing).addListener( ... );

The `userInterfaceIsShowing` is then part of a `Node` based container, or is not referenced anywhere at all (it's just derived from say `node.sceneProperty().flatMap(Scene::windowProperty).flatMap(Window::isShowingProperty)`.

It is important to realize that `when` uses two (normal) listeners:
- One on its condition property (`active`), which is not weak and is **never** removed (iirc, it can't be weak as otherwise the observable created by `when` could get GC'd when the condition is `false`, but shouldn't have been as the condition may yet become `true` again).
- One on its parent property (x in `x.when( ... )`) -- this listener (and thus its hard reference) is only present when the condition evaluates to `true`

In this specific case there is nothing that tells us that the menu disappeared, so the best we can do is that when a **new** menu is created that at that time we discard the old one.  To be able to discard it, we unfortunately need to keep a hard reference to the `active` property we used (so it can be set to `false`).  However, the `when` has a listener on that `active` property (as it needs to know when the condition changes), and a listener points back to the observable value that `when` created (it is not, and can't be, a weak listener).  So in order to also cut that hard reference, we need to replace the `active` property as well after setting it to `false` (if you don't set it to `false`, there will be a listener on the parent of `when` that keeps a hard reference still... setting it to `false` removes that listener first).

So definitely not your run-of-the-mill `when` example, but it will at least always behave deterministically as there are no weak listeners used anywhere.  If there is a bug, it will be easy to reproduce on someone else's machine.

The illustration I added earlier above should hopefully make the above a bit more clear, but of course some documentation in the code to explain this may be good too.

hjohn: @johanvos I don't think what I wrote above (option 1) will work -- there will still be a reference from the `active` property that keeps the menu item alive; a reference is indeed removed when `hasParent` becomes `false`, but there is another (same reason we had to recreate the active property)... I will think on this some more over the weekend.

johanvos:
A few things about the latest commit:

1. The usage of the `active` property caused regression: the memoryLeak test that was introduced in the fix for JDK-8318841 now failed. A number of listeners were hard referenced from the `active` property.
The complexity is that there are different entry points by which a listener should be removed. What we try to do in this PR, is making sure we remove listeners after a defocus/focus operation that would otherwise stay referenced. A focus operation will result in GlassSystemMenu.setMenus() being called. 
I reverted some of the dependencies on the `active` property, and kept those that were directly created as a consequence of the setMenus call.

2. I added a test that demonstrates the issue when not removing menu's completely inside the listener, as reported by @jperedadnr and this issue is fixed as well now.

3. All tests now pass, but I noticed that in some cases, the systemtests do not correctly work with the application lifecycle management (see https://mail.openjdk.org/pipermail/openjfx-dev/2024-January/044516.html). For now, I consider this anomaly to be independent from JDK-8319779

FlorianKirmaier:
@johanvos 
**ticket**
We found a bug, which is also fixed by this PR. This is the ticket: https://bugs.openjdk.org/browse/JDK-8323787
It's an IndexOutOfBounds exception, related to the visible flag/listeners.

**unittest**
We wrote a unit test, which goes green with this PR.
Can you do us a favor, and add the test from the ticket to your PR?
Then we can be sure, that this issue will be solved and never reappears.

johanvos:
@FlorianKirmaier Thanks for the additional test. There was already a test in the SystemMenuBarTest, but it's really good to have an additional test for this -- I added it.

kevinrushforth:
> Thanks for the additional test. There was already a test in the SystemMenuBarTest, but it's really good to have an additional test for this -- I added it.

@johanvos [8323787](https://bugs.openjdk.org/browse/JDK-8323787) seems like a distinct bug, so I recommend adding it to this PR with `/issue add 8323787`.

kevinrushforth:
Getting back to this.... I ran the test on macOS and it now works as expected. However, one of the new tests fails on Linux:

```
SystemMenuBarTest > testFocusMemoryLeak FAILED
    org.opentest4j.AssertionFailedError: expected: <false> but was: <true>
        at app//org.junit.jupiter.api.AssertFalse.assertFalse(AssertFalse.java:40)
        at app//org.junit.jupiter.api.AssertFalse.assertFalse(AssertFalse.java:35)
        at app//org.junit.jupiter.api.Assertions.assertFalse(Assertions.java:227)
        at app//test.com.sun.javafx.tk.quantum.SystemMenuBarTest.testFocusMemoryLeak(SystemMenuBarTest.java:198)
```

Btw, all of my tests were done in a local branch with the latest upstream master merged in.

@johanvos since the source branch is now a bit out of date, can you merge in the latest upstream master?

Also, I think there are still some unaddressed questions from John and Andy.

Regarding the following:

> All tests now pass, but I noticed that in some cases, the systemtests do not correctly work with the application lifecycle management (see https://mail.openjdk.org/pipermail/openjfx-dev/2024-January/044516.html). For now, I consider this anomaly to be independent from JDK-8319779

Yes, this is unrelated to the memory leak being fixed by this PR. We should file a bug for the activation problem.

johanvos:
@kevinrushforth thanks for the feedback. I merged master and I see the failing test on Linux now as well. That is very interesting. I'll investigate and fix, and address the other comments as well.
Also, I got some offline feedback about edge cases that I'm still trying to reproduce.

johanvos:
The failure on Linux has an interesting cause. The GlassSystemMenu (in com.sun.javafx.tk.quantum) has this code:
```
            /*
             * Leave the Apple menu in place
             */
            for (int index = existingSize - 1; index >= 0; index--) {
                Menu menu = existingMenus.get(index);
                clearMenu(menu);
                glassSystemMenuBar.remove(index);
            }
```
As a consequence, the first menu is not cleared -- regardless of the presence of an apple or not.
There are plenty quick and dirty ways to fix this, but this platform-specific check probably does not belong in this package.
@kevinrushforth what do you prefer to do in cases like these?

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassSystemMenu.java
```diff
@@ -169,32 +216,14 @@ private void insertMenu(final Menu parent, final MenuBase mb, int pos) {
...
```
jperedadnr, line 226: Since you are not calling `get()` from the `when` resulting observable (to allow for invalidation), these need to use a `ChangeListener` instead of an `InvalidationListener`, otherwise, while `active` doesn't change, any change in those menuBase properties will trigger just one notification. 

Alternative, just replace `addListener` with  `subscribe` (since internally uses a `ChangeListener`).

@hjohn does this make sense to you?

hjohn: This could be a problem normally, but I think in this case you won't be able to get this to produce incorrect results.

`parseText` is accessing the property (and it calls `isMnemonicParsing`), but only if the text is not empty.  It's sort of "safe" as an empty text can't contain a mnemonic anyway, and you'd need to change the text at a minimum first to see the effect of turning mnemonic parsing on/off.

A change listener would be more obviously correct, but it is not strictly needed here.

(not all `subscribe`s use change listener, the one that takes a `Runnable` uses invalidation)

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassSystemMenu.java
hjohn: I think you already solved the issue.

tests/system/src/test/java/test/com/sun/javafx/tk/quantum/SystemMenuBarTest.java
johanvos: Not sure I understand where you want to add or remove a newline?

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassSystemMenu.java
jperedadnr: There is an issue using `when()`, because you add the listener to a new property, not to the original property. Therefore, using an invalidation listener doesn't trigger notifications if you changing values more than once to the original property.
Simply run this test:
```
        BooleanProperty b = new SimpleBooleanProperty();
        BooleanProperty active = new SimpleBooleanProperty(true);

        b.addListener(o -> System.out.println("b: " + b.get()));
        b.when(active).addListener(o -> System.out.println("when: " + b.get()));
        b.set(true);
        b.set(false);
```

It prints:

```
b: true
when: true
b: false
```

so with `when` we miss the changes! (I can go into details of why this happens, but just notice we have a new property after all, and not calling `when.get()` it doesn't get validated anymore).

This can happen in this SystemMenuBar case with the disabled property of any given `Menu` or `MenuItem`, if you simply change it twice, while the application is active. 

My proposal of using `subscribe` implied using the `Consumer` (therefore the "replacement").

tests/system/src/test/java/test/com/sun/javafx/tk/quantum/SystemMenuBarTest.java
andy-goryachev-oracle: please remove extra newline.

kevinrushforth:
> The failure on Linux has an interesting cause. The GlassSystemMenu (in com.sun.javafx.tk.quantum) has this code:
> 
> ```
>             /*
>              * Leave the Apple menu in place
>              */
>             for (int index = existingSize - 1; index >= 0; index--) {
>                 Menu menu = existingMenus.get(index);
>                 clearMenu(menu);
>                 glassSystemMenuBar.remove(index);
>             }
> ```
> 
> As a consequence, the first menu is not cleared -- regardless of the presence of an apple or not. There are plenty quick and dirty ways to fix this, but this platform-specific check probably does not belong in this package. @kevinrushforth what do you prefer to do in cases like these?

I missed responding to this. Presuming this really does need to be different on Mac than on other platforms, I guess the cleanest thing would be to move that logic to a method that can be overridden on Mac. If that's not feasible, then checking PlatformUtil.isMac() would be the other way.

One question, though: Why does this prevent the first menu from being cleared? Unless I'm missing something, doesn't that loop cover the whole range of menus? And since Linux doesn't even implement system menus, I wonder why this code is active for Linux at all?

kevinrushforth:
Answering my own questions...

The existing for loop quoted above actually loops while `>= 1` rather than `>= 0`:

```
            for (int index = existingSize - 1; index >= 1; index--) {
```

so indeed the last menu would be held.

And I think that the answer to my other question (why is this even being used on Linux) is that the `createAndRefocusMenuBarStage()` method in the newly added test is calling `GlassSystemMenuBar::createMenuBar` unconditionally (via the Shim), when the actual logic in WindowStage qualifies by whether or not system menu is supported (which it only is for Mac).

So I think this is a test bug. I might recommend adding something like the following to the test:

```
assumeTrue(glassSystemMenu.isSupported());
```

johanvos:
I think the logic in GlassSystemMenu should ultimately be platform-independent. Hence, moving the Apple-menu specific code to a Mac-specific class (in glass) sounds the right thing to do. However, it might be more pragmatic that we wait with this refactory until we have a second (or third) platform implementation. I can imagine there will need to be more changes if we deal with different linux window management systems.
Hence, for now I agree having an `assumeTrue(glassSystemMenu.isSupported())` is the easiest solution (and not an incorrect one).

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassSystemMenu.java
hjohn: Thanks, you are right.  The property generated by `when` is not getting revalidated, because its specific `getValue` is never called when using an invalidation listener on it. This seems to be a bit of pitfall that applies to all such mapped properties (not exclusive to `when`).

I'm fine with any of the options.  Option 3 certainly is the most convenient.

I'm more wondering if there should be some action taken in for the `map`, `flatMap`, `when`, `orElse` constructs to make them work better with invalidation listeners, or if that's not possible, a not explaining that adding an invalidation listener to such a construct probably is not doing what you think it is doing.

johanvos:
I think this PR is ready for review.

johanvos:
@jperedadnr @hjohn can you review this?
@kevinrushforth feel free to add other reviewers.

tests/system/src/test/java/test/com/sun/javafx/tk/quantum/SystemMenuBarTest.java
```diff
@@ -0,0 +1,351 @@
...
+        Platform.runLater(() -> {
+            mainMenu.getItems().clear();
+            mainMenu.getItems().add(subMenu);
+            subMenu.getItems().addAll(new MenuItem("new item 1"), new MenuItem("new item 2"));
+            removeMenuLatch.countDown();
+        });
+    }
+
+    @Test
+    public void testJDK8309935() {
```
hjohn, line 287: minor: Not sure what our policies are, but I find test names like this pretty useless, perhaps a short indication what this is doing?

tests/system/src/test/java/test/com/sun/javafx/tk/quantum/SystemMenuBarTest.java
```diff
@@ -0,0 +1,351 @@
...
+            Menu test4 = new Menu("test 4");
+            test4.getItems().add(new MenuItem("item 1"));
+            test4.getItems().add(new MenuItem("item 2"));
+            menuBar.getMenus().get(1).getItems().addAll(test3, test4);
+        });
+        // Some waiting is necessary. runAndWait twice makes it reliable.
+        Util.runAndWait(() -> {
+        });
+        Util.runAndWait(() -> {
+        });
```
hjohn, line 346: I checked the code that `runAndWait` would do when you pass in 0 runnables, and it basically does no waiting at all (less than a microsecond for sure).  If this is really essential (I removed it and the test still passed) then I think a `sleep` might be better.

hjohn (changes requested):
The new test passed with both the old and the new version (I reverted the code, aside from leaving `setMenuBindings` `protected`).  Is the test testing the right thing?

(If relevant, this was done on a Windows platform)

modules/javafx.graphics/src/shims/java/com/sun/javafx/tk/quantum/GlassSystemMenuShim.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public void createMenuBar() {
+        super.createMenuBar();
+    }
+
+    @Override
+    protected void setMenuBindings(final Menu glassMenu, final MenuBase mb) {
+        super.setMenuBindings(glassMenu, mb);
+        uncollectedMenus.add(new WeakReference(glassMenu));
```
jperedadnr, line 54: Add `<>` to avoid raw use of `WeakReference`

modules/javafx.graphics/src/shims/java/com/sun/javafx/tk/quantum/GlassSystemMenuShim.java
```diff
@@ -0,0 +1,60 @@
...
+import com.sun.glass.ui.Menu;
+import com.sun.javafx.menu.MenuBase;
+import java.lang.ref.WeakReference;
+import java.util.ArrayList;
+import java.util.List;
+
+public class GlassSystemMenuShim extends GlassSystemMenu {
+
+    private GlassSystemMenu gsm;
+    final ArrayList<WeakReference<Menu>> uncollectedMenus = new ArrayList<>();
```
jperedadnr, line 37: this can be private?

tests/system/src/test/java/test/com/sun/javafx/tk/quantum/SystemMenuBarTest.java
```diff
@@ -0,0 +1,351 @@
...
+            mainMenu.getItems().add(subMenu);
+            subMenu.getItems().addAll(new MenuItem("new item 1"), new MenuItem("new item 2"));
+            removeMenuLatch.countDown();
+        });
+    }
+
+    @Test
+    public void testJDK8309935() {
+        MenuBar menuBar = new MenuBar();
+        AtomicReference<Throwable> throwableRef = new AtomicReference();
```
jperedadnr, line 289: Add missing `<>` to avoid raw use of `AtomicReference`

jperedadnr (approved):
I've tested on macOS, and tests pass now, and fail without the changes from this PR.
I've left some comments, but once addressed, I'll re-approve.

kevinrushforth:
> The new test passed with both the old and the new version (I reverted the code, aside from leaving `setMenuBindings` `protected`). Is the test testing the right thing?
> 
> (If relevant, this was done on a Windows platform)

That's not surprising since only macOS provides the (optional) system menu capability, and the leak is in the glass system menu implementation. The tests will pass vacuously on other platforms, since the system menu code with the bug will never be reached.

modules/javafx.graphics/src/shims/java/com/sun/javafx/tk/quantum/GlassSystemMenuShim.java
johanvos: done

modules/javafx.graphics/src/shims/java/com/sun/javafx/tk/quantum/GlassSystemMenuShim.java
johanvos: done

tests/system/src/test/java/test/com/sun/javafx/tk/quantum/SystemMenuBarTest.java
johanvos: done

johanvos:
@arapte if possible, can you re-review this?

jperedadnr (approved):

tests/system/src/test/java/test/com/sun/javafx/tk/quantum/SystemMenuBarTest.java
```diff
@@ -0,0 +1,351 @@
...
+                menu2.getItems().removeIf(o -> Objects.equals(o.getText(), test1.getText()));
+                menu2.getItems().add(test1);
+                menu2.getItems().removeIf(o -> Objects.equals(o.getText(), test2.getText()));
+                menu2.getItems().add(test2);
+            });
+            BorderPane root = new BorderPane();
+            root.setTop(menuBar);
+            Stage stage = new Stage();
+            stage.setScene(new Scene(root));
+            stage.show();
```
arapte, line 321: I think, a CountDownLatch should be added to make sure that stage is shown before proceeding.

arapte (changes requested):
looks good, suggesting a few minor changes.

tests/system/src/test/java/test/com/sun/javafx/tk/quantum/SystemMenuBarTest.java
johanvos: It is a good question. We don't have a consistent approach yet. I personally like the link to the JBS issue as that has (or should have) all info. I added a short indication, but the real info should be on the issue, imho.
(this is something we should discuss on the mailing list)

tests/system/src/test/java/test/com/sun/javafx/tk/quantum/SystemMenuBarTest.java
johanvos: The Util.runAndWait was indeed not a good idea. The problem is not that the test might not pass, the problem is that the test might not fail if we don't wait for things that need to run to complete. But this is better done with a Platform.runLater() which will add a runnable to the queue, so when that is executed, we know that all previously scheduled runnables completed.
I changed the approach ehre.

tests/system/src/test/java/test/com/sun/javafx/tk/quantum/SystemMenuBarTest.java
```diff
@@ -0,0 +1,351 @@
+/*
```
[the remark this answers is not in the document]
johanvos: I manually reverted the add/remove part, and replaced it with `git mv`. I assume/hope that by squashing the commits, the add/remove will not be part of the change.

tests/system/src/test/java/test/com/sun/javafx/tk/quantum/SystemMenuBarTest.java
johanvos: I don't completely understand this. The stage.show() is a blocking call, and the runnable that it might schedule will be executed before other runnables (part of the processing in the rest of the method) are scheduled.
I added a CountDownLatch at the end of the processing, so that we allow for the processing to run into exceptions before we finalize the test.
Does that make sense?

tests/system/src/test/java/test/com/sun/javafx/tk/quantum/SystemMenuBarTest.java
hjohn: I see that you added a comment now, that helps a lot, thanks.

hjohn (approved):
LGTM, note that I couldn't test this as I don't have access to a Mac, but I didn't see a problem on Windows with these changes.

tests/system/src/test/java/test/com/sun/javafx/tk/quantum/SystemMenuBarTest.java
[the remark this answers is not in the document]
kevinrushforth: git doesn't actually track renames and copies, so there is ultimately no difference between a "git mv" and "git rm; git add", at least not when done as part of a single commit. What git does is a heuristic when presenting diffs based on how similar the two files are.

tests/system/src/test/java/test/com/sun/javafx/tk/quantum/SystemMenuBarTest.java
kevinrushforth: One correction: `Stage.show` is not a blocking call. I haven't looked at it closely, but perhaps what @arapte meant was to suggest that the test wait on a latch that is triggered when the Stage::showing event is delivered? Many of our headful tests do that, but I don't know whether that would be applicable here or not.

johanvos: Right, it wasn't correct to call it a "blocking call". However, before Stage.show() returns, the contents of the stage are created and their layoutChildren method is invoked (before the WINDOW_SHOWN is fired). 
All Runnables that are scheduled during this method will be executed before the content in the next Util.runAndWait() is executed, so I'm not sure what we should be waiting for?

tests/system/src/test/java/test/com/sun/javafx/tk/quantum/SystemMenuBarTest.java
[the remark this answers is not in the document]
arapte: My intention to use git mv was to retain history on the file. I tried below locally but it did not retain the history.

git mv <orig-path> <new-path>
< modify the file content >
git add <new-path>
git commit -m "commit message"
git log --follow <new-path>  -> expected to see old commits in log but it showed only one log.

What am I missing here !

arapte (approved):
lgtm

andy-goryachev-oracle:
After integrating with the master, I am getting these errors in Eclipse:

```
Description	Resource	Location
The method createMenuBar() of type GlassSystemMenuShim should be tagged with @Override since it actually overrides a superclass method	GlassSystemMenuShim.java	line 47
The method setMenus(List<MenuBase>) of type GlassSystemMenuShim should be tagged with @Override since it actually overrides a superclass method	GlassSystemMenuShim.java	line 43
GlassSystemMenuShim cannot be resolved to a type	SystemMenuBarTest.java	line 217
GlassSystemMenuShim cannot be resolved to a type	SystemMenuBarTest.java	line 217
MenuBase cannot be resolved to a type	SystemMenuBarTest.java	line 216
MenuBase cannot be resolved to a type	SystemMenuBarTest.java	line 221
MenuBase cannot be resolved to a type	SystemMenuBarTest.java	line 222
The import com.sun.javafx.tk.quantum cannot be resolved	SystemMenuBarTest.java	line 42
The type com.sun.javafx.menu.MenuBase is not accessible	SystemMenuBarTest.java	line 28

```

andy-goryachev-oracle:
Created https://bugs.openjdk.org/browse/JDK-8337228 and https://bugs.openjdk.org/browse/JDK-8337229

Dianne087:
hey, i just a little investigating about my work, and sometimes most probbly i found all of u in galaxy android after releasing we having a quest! have s good day

PR #1287, opened 2023-11-14
8282290: TextField Cursor Position one off

karthikpandelu:
The change listener on caretPositionProperty() was not getting invoked on replacing the content with same text as there was no change in caret position. Since the textProperty invalidation sets the forward bias to true by default, incorrect caret position was  calculated when the same text was replaced after clicking on the trailing edge of last character or empty space in the TextField.

Since caretShapeProperty invalidation listener gets invoked without changing the caret position, updating the caretBiasProperty on this listener solves the issue.

Since the caret position value will be same when the caret is present after the last character or before the last character, it can not be validated using unit test.
The fix can be validated using MonkeyTester.
Steps to select TextField option in Monkey Tester.

- Open the MonkeyTester app and select TextField from the left option pane.
- Select Short from Text selection option and click on the TextField to bring it into focus.
- Select all using cmd(ctrl) + a and copy using cmd(ctrl) + c
- Click on empty space in the TextField after the present content.
- Select all again using the keyboard shortcut and paste using cmd(ctrl) + v
- The caret should be displayed after the last character. Without the fix caret gets displayed before the last character.

---------
### Progress

### Issues
 * [JDK-8282290](https://bugs.openjdk.org/browse/JDK-8282290): TextField Cursor Position one off (**Bug** - P3)
 * [JDK-8248914](https://bugs.openjdk.org/browse/JDK-8248914): Javafx TextField positions the cursor incorrectly after pressing DEL key (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1287/head:pull/1287` \
`$ git checkout pull/1287`

`$ git checkout pull/1287` \
`$ git pull https://git.openjdk.org/jfx.git pull/1287/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1287`

`$ git pr show -t 1287`

</details>

<a href="https://git.openjdk.org/jfx/pull/1287.diff">https://git.openjdk.org/jfx/pull/1287.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1287#issuecomment-1809941852)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
@@ -245,7 +245,7 @@ public class TextFieldSkin extends TextInputControlSkin<TextField> {
         caretPath.layoutXProperty().bind(textTranslateX);
         textNode.caretShapeProperty().addListener(observable -> {
             caretPath.getElements().setAll(textNode.caretShapeProperty().get());
-            if (caretPath.getElements().size() == 0) {
+            if (caretPath.getElements().size() == 0 || caretPath.getElements().size() != 4) {
                 // The caret pos is invalid.
                 updateTextNodeCaretPos(control.getCaretPosition());
             } else if (caretPath.getElements().size() == 4) {
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
@@ -246,7 +246,10 @@ public class TextFieldSkin extends TextInputControlSkin<TextField> {
         textNode.caretShapeProperty().addListener(observable -> {
             caretPath.getElements().setAll(textNode.caretShapeProperty().get());
             if (caretPath.getElements().size() != 4) {
-                // The caret pos is invalid.
+                /* On replacing same text using keyboard shortcut,
+                 * caret position is not updated.
+                 * The caret pos is invalid in this case,
+                 * hence it should be updated when caret path size is not 4 */
                 updateTextNodeCaretPos(control.getCaretPosition());
             } else if (caretPath.getElements().size() == 4) {
                 // The caret is split. Ignore and keep the previous width value.
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
@@ -245,7 +245,7 @@ public class TextFieldSkin extends TextInputControlSkin<TextField> {
         caretPath.layoutXProperty().bind(textTranslateX);
         textNode.caretShapeProperty().addListener(observable -> {
             caretPath.getElements().setAll(textNode.caretShapeProperty().get());
-            if (caretPath.getElements().size() == 0 || caretPath.getElements().size() != 4) {
+            if (caretPath.getElements().size() != 4) {
                 // The caret pos is invalid.
                 updateTextNodeCaretPos(control.getCaretPosition());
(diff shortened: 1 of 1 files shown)


modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
```diff
@@ -245,7 +245,7 @@ public TextFieldSkin(final TextField control) {
         caretPath.layoutXProperty().bind(textTranslateX);
         textNode.caretShapeProperty().addListener(observable -> {
             caretPath.getElements().setAll(textNode.caretShapeProperty().get());
-            if (caretPath.getElements().size() == 0) {
+            if (caretPath.getElements().size() == 0 || caretPath.getElements().size() != 4) {
```
Maran23, line 248: I think this can be simplified to:

```
if (caretPath.getElements().size() != 4) {
    // The caret pos is invalid.
    updateTextNodeCaretPos(control.getCaretPosition());
} else {
    caretWidth = Math.round(caretPath.getLayoutBounds().getWidth());
}
```

andy-goryachev-oracle: this might be problematic, as it uses undocumented aspect of Text/Flow.caretShape().  the current implementation returns either a single line, or two lines (moveto,lineto,moveto,lineto) in the case of split/dual caret.  But this may not be true in the future!  If we add a directional indicator to the caret this logic will break.

What we may need is to invent a new API possibly that gives us more information about the caret - whether it's a single or split one, or whether it has a directionality indicator, information about the adjacent text segments (rtl,ltr), etc.

We can keep this logic for now, to be fixed once (and if) we add the more detailed caret API.

Maran23: Yes, right now it is a bit magic. At least the first if can be simplified still, as `!= 4` will also handle `== 0`.

karthikpandelu: I agree with Andy here. However the first condition can be simplified without any side effects but I had seen some side effects if we remove second condition. Hence simplified the first condition as suggested above and maintained second condition as it is.

andy-goryachev-oracle:
@karthikpandelu I think this change also fixes JDK-8248914 - could you please double check?  
I've tested it on macOS 13.5.2

andy-goryachev-oracle (approved):
looks good, with a minor comment.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
```diff
@@ -245,7 +245,7 @@ public TextFieldSkin(final TextField control) {
         caretPath.layoutXProperty().bind(textTranslateX);
         textNode.caretShapeProperty().addListener(observable -> {
             caretPath.getElements().setAll(textNode.caretShapeProperty().get());
-            if (caretPath.getElements().size() == 0) {
+            if (caretPath.getElements().size() != 4) {
```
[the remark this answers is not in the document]
karthikpandelu: Added comment

karthikpandelu:
> @karthikpandelu I think this change also fixes [JDK-8248914](https://bugs.openjdk.org/browse/JDK-8248914) - could you please double check? If so, we can add that ticket to this PR.
> 
Yes it fixes [JDK-8248914](https://bugs.openjdk.org/browse/JDK-8248914). I'll add this bug to this PR.

andy-goryachev-oracle (approved):
looks good, thanks!

karthikpandelu:
@Maran23 could you please re-review?

Maran23 (approved):
Looks good.

PR #1291, opened 2023-11-20
8284544: [Win] Name-Property of Spinner cannot be changed

arapte:
Currently we use the value of spinner as it's `UIA_NamePropertyId` when a11y client application requests for it.
Ideally we should use the text set by `Node.setAccessibleText()` as the `UIA_NamePropertyId`.
For other controls such as Slider, ListView we use the text set by setAccessibleText() API.

Fix:
Use the text set by `Node.setAccessibleText()` as the `UIA_NamePropertyId`.
This means, when a11y client requests `UIA_NamePropertyId`, which is mapped to AccessibleAttribute.TEXT attribute,  we shall return the accessible text.
So we need another way to read out the VALUE of the Spinner.
- For this we need to implement `IValueProvider` pattern for Spinner control
- IValueProvider requests the value of the control using it's API `get_ValueString()`
- It required to introduce a new AccessibleAttribute `VALUE_STRING`
- IValueProvider  also reads out if the control is editable or not, hence added `EDITABLE `case in `Spinner.queryAccessibleAttribute()`

Verification:
- Run any spinner app, with setAccessibleText set on spinner 
- Run Windows narrator and observe
- Without this fix:
- 1. Narrator does not read the text set by setAccessibleText
- 2. In application "Accessibility Insights for Windows", you can see the value of Spinner as the Name property
- After this fix:
- 1. Narrator reads the accessible text then value of Spinner and then if editable or not
- 2. In application "Accessibility Insights for Windows", you can see the text set by `setAccessibleText()` for Spinner as the Name property and the Spinner's value as value property

---------
### Progress
- [x] Change requires CSR request [JDK-8320505](https://bugs.openjdk.org/browse/JDK-8320505) to be approved

### Issues
 * [JDK-8284544](https://bugs.openjdk.org/browse/JDK-8284544): [Win] Name-Property of Spinner cannot be changed (**Bug** - P3)
 * [JDK-8320505](https://bugs.openjdk.org/browse/JDK-8320505): [Win] Name-Property of Spinner cannot be changed (**CSR**)

### Reviewers
 * [Ajit Ghaisas](https://openjdk.org/census#aghaisas) (@aghaisas - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1291/head:pull/1291` \
`$ git checkout pull/1291`

`$ git checkout pull/1291` \
`$ git pull https://git.openjdk.org/jfx.git pull/1291/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1291`

`$ git pr show -t 1291`

</details>

<a href="https://git.openjdk.org/jfx/pull/1291.diff">https://git.openjdk.org/jfx/pull/1291.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1291#issuecomment-1818264436)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/Spinner.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/Spinner.java
@@ -769,7 +769,7 @@ public class Spinner<T> extends Control {
             }
         }
 
-        notifyAccessibleAttributeChanged(AccessibleAttribute.TEXT);
+        notifyAccessibleAttributeChanged(AccessibleAttribute.VALUE_STRING);
         if (text == null) {
             if (value == null) {
                 getEditor().clear();
@@ -834,7 +834,7 @@ public class Spinner<T> extends Control {
     @Override
     public Object queryAccessibleAttribute(AccessibleAttribute attribute, Object... parameters) {
         switch (attribute) {
-            case TEXT: {
+            case VALUE_STRING: {
                 T value = getValue();
                 SpinnerValueFactory<T> factory = getValueFactory();
                 if (factory != null) {
@@ -845,6 +845,15 @@ public class Spinner<T> extends Control {
                 }
                 return value != null ? value.toString() : "";
             }
+
+            case TEXT: {
+                String accText = getAccessibleText();
+                return (accText != null) ? accText : "";
+            }
+
+            case EDITABLE:
+                return isEditable();
+
             default: return super.queryAccessibleAttribute(attribute, parameters);
         }
     }
--- a/modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacAccessible.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacAccessible.java
@@ -801,6 +801,11 @@ final class MacAccessible extends Accessible {
                     macNotification = MacNotification.NSAccessibilityValueChangedNotification;
                 }
                 break;
+            case VALUE_STRING:
+                if (getAttribute(ROLE) == AccessibleRole.SPINNER) {
+                    macNotification = MacNotification.NSAccessibilityValueChangedNotification;
+                }
+                break;
             case PARENT:
                 if (peer != 0L) {
                     _invalidateParent(peer);
@@ -1208,6 +1213,10 @@ final class MacAccessible extends Accessible {
                         case TEXT_AREA:
                         case COMBO_BOX:
                             jfxAttr = TEXT;
+                            map = MacVariant::createNSString;
+                            break;
+                        case SPINNER:
+                            jfxAttr = VALUE_STRING;
                             map = MacVariant::createNSString;
                             break;
                         case CHECK_BOX:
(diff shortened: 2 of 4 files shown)


kevinrushforth:
This should be tested in connection with PR #1226

modules/javafx.controls/src/main/java/javafx/scene/control/Spinner.java
```diff
@@ -845,6 +845,15 @@ public Object queryAccessibleAttribute(AccessibleAttribute attribute, Object...
                 }
                 return value != null ? value.toString() : "";
             }
+
+            case TEXT: {
+                String accText = getAccessibleText();
+                return (accText != null) ? accText : "";
```
kevinrushforth, line 851: Is there a better default than the empty string? What do we do for Slider and other similar controls?

modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
```diff
@@ -774,6 +774,17 @@ public enum AccessibleAttribute {
...
     VALUE(Double.class),
 
+    /**
+     * Returns the value as string for the node.
+     * <ul>
+     * <li>Used by: Spinner </li>
+     * <li>Needs notify: yes </li>
+     * <li>Return Type: {@link String} </li>
+     * <li>Parameters: </li>
+     * </ul>
```
kevinrushforth, line 784: This will need `@since 22`

kevinrushforth (commented):
I left quick comment and a question inline.

modules/javafx.controls/src/main/java/javafx/scene/control/Spinner.java
arapte: With this fix, a11y client app reads a spinner as,
- when accessible text is set: [Accessible Text] Spinner [Value] [Editable / read only]
- when accessible text is not set: Spinner [Value] [Editable / read only]

Other controls like ComboBox, ChoiceBox return the value of control when accessible text is not set.
If we do the same for spinner, then the Spinner gets read as:  [Value] Spinner [Value] [Editable / read only]
In this case the value gets read twice.

The way spinner is read when accessible text is not set, it seems Ok to skip accessible text when it is not set.

arapte:
Updated the PR with changes required for macOS, such that the Spinner is read same as before.

kevinrushforth (approved):
The changes look good and everything works as I would expect. I left one question, but it is fine as you have it.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacAccessible.java
```diff
@@ -801,6 +801,11 @@ public void sendNotification(AccessibleAttribute notification) {
                     macNotification = MacNotification.NSAccessibilityValueChangedNotification;
                 }
                 break;
+            case VALUE_STRING:
+                if (getAttribute(ROLE) == AccessibleRole.SPINNER) {
```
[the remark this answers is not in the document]
arapte: Yes Kevin, as of now VALUE_STRING is only used for Spinner, but it's a pattern that is followed in this file: If anything is specific to a control we check the role. It also helps in finding code specific to a control. Let's please keep the check.

aghaisas (approved):
The fix looks good!

PR #1292, opened 2023-11-20
JDK-8320444: Column drag header is positioned wrong for nested columns

Maran23:
When a nested column is dragged, the column drag header (blue) is positioned wrong.
It appears at the very left of the table and not where the column is.

image

The fix is to calculate the bound based off the `TableHeaderRow` (and not the parent `NestedTableColumnHeader`.

---------
### Progress

### Issue
 * [JDK-8320444](https://bugs.openjdk.org/browse/JDK-8320444): Column drag header is positioned wrong for nested columns (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1292/head:pull/1292` \
`$ git checkout pull/1292`

`$ git checkout pull/1292` \
`$ git pull https://git.openjdk.org/jfx.git pull/1292/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1292`

`$ git pr show -t 1292`

</details>

<a href="https://git.openjdk.org/jfx/pull/1292.diff">https://git.openjdk.org/jfx/pull/1292.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1292#issuecomment-1819908818)

### DIFF
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TableViewSkinTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TableViewSkinTest.java
@@ -32,6 +32,7 @@ import javafx.scene.Node;
 import javafx.scene.Scene;
 import javafx.scene.control.TableColumn;
 import javafx.scene.control.TableView;
+import javafx.scene.control.skin.NestedTableColumnHeader;
 import javafx.scene.control.skin.TableColumnHeader;
 import javafx.scene.control.skin.TableColumnHeaderShim;
 import javafx.scene.control.skin.TableHeaderRow;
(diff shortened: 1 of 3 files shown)


modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TableViewSkinTest.java
```diff
@@ -188,6 +189,43 @@ void testColumnHeaderReorderCorrectTranslateX() {
...
+        assertEquals(0, columnDragHeader.getTranslateX());
+
+        NestedTableColumnHeader nestedTableColumnHeader =
+                (NestedTableColumnHeader) header.getRootHeader().getColumnHeaders().get(2);
+        TableColumnHeader tableColumnHeader = nestedTableColumnHeader.getColumnHeaders().get(0);
+
+        Bounds bounds = tableColumnHeader.localToScene(tableColumnHeader.getLayoutBounds());
+        TableColumnHeaderShim.columnReordering(tableColumnHeader, bounds.getMinX() + 20, bounds.getMinY());
+
+        // 200, since we have 2 columns to the left with a size of 100.
```
karthikpandelu, line 225: Can you also add why assertion is done for 220 in this comment?

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableViewSkinTest.java
```diff
@@ -168,6 +169,43 @@ void testColumnHeaderReorderCorrectTranslateX() {
...
+        assertEquals(0, columnDragHeader.getTranslateX());
+
+        NestedTableColumnHeader nestedTableColumnHeader =
+                (NestedTableColumnHeader) header.getRootHeader().getColumnHeaders().get(2);
+        TableColumnHeader tableColumnHeader = nestedTableColumnHeader.getColumnHeaders().get(0);
+
+        Bounds bounds = tableColumnHeader.localToScene(tableColumnHeader.getLayoutBounds());
+        TableColumnHeaderShim.columnReordering(tableColumnHeader, bounds.getMinX() + 20, bounds.getMinY());
+
+        // 200, since we have 2 columns to the left with a size of 100.
```
karthikpandelu, line 205: Same as the previous comment. Can you also add why assertion is done for 220 in this comment?

karthikpandelu (commented):
The changes look good to me. Tested the changes in MacOS.
Added minor comments inline.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TableViewSkinTest.java
andy-goryachev-oracle: good point.

May I suggest to declare constants for column widths and column reordering coordinates so the origin of these numbers becomes clear?
Also, we are comparing floating point numbers, so perhaps it's better specify a small delta:
```
double CWIDTH = 100;
double EPSILON = 0.0001;
assertEquals(CWIDTH + CWIDTH + OFFSET, columnDragHeader.getTranslateX(), EPSILON);
```

andy-goryachev-oracle:
the fix looks good, tested with and without nested columns.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TableViewSkinTest.java
Maran23: Pushed. 
Since we compare very simple numbers (we even set them before as int), I personally would not expect a delta but an accurate number, therefore no delta added in this test.

andy-goryachev-oracle: Will it be a round integer in case of fractional scale (it might be in this case, since the test does not set the scale explicitly)?

Should we extends this test with fractional scale, let's say 1.25?

Have you tested this fix with fractional scale?

andy-goryachev-oracle (approved):
tested on macOS and windows (the latter with fractional scale).  looks good.

mstr2 (approved):
The fix does what I expect it to do.

karthikpandelu (approved):

PR #1293, opened 2023-11-21
8320359: ImageView: add styleable fitWidth, fitHeight, preserveRatio, smooth properties

andy-goryachev-oracle:
Adding missing styleable properties to ImageView:

```
-fx-preserve-ratio
-fx-smooth
-fx-fit-width
-fx-fit-height
```

Updated CSS Reference.

---------
### Progress
- [x] Change requires CSR request [JDK-8321126](https://bugs.openjdk.org/browse/JDK-8321126) to be approved

### Issues
 * [JDK-8320359](https://bugs.openjdk.org/browse/JDK-8320359): ImageView: add styleable fitWidth, fitHeight, preserveRatio, smooth properties (**Enhancement** - P4)
 * [JDK-8321126](https://bugs.openjdk.org/browse/JDK-8321126): ImageView: add styleable fitWidth, fitHeight, preserveRatio, smooth properties (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1293/head:pull/1293` \
`$ git checkout pull/1293`

`$ git checkout pull/1293` \
`$ git pull https://git.openjdk.org/jfx.git pull/1293/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1293`

`$ git pr show -t 1293`

</details>

<a href="https://git.openjdk.org/jfx/pull/1293.diff">https://git.openjdk.org/jfx/pull/1293.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1293#issuecomment-1834102789)


modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
```diff
@@ -327,5 +328,26 @@ public int hashCode() {
...
+     * Utility method which combines CssMetaData items in one unmodifiable list with the size equal to the number
+     * of items it holds (i.e. with no unnecessary overhead).
+     *
+     * @param list the css metadata items, usually from the parent, not nullable
+     * @param items the additional items
+     * @return the unmodifiable list containing all of the items
+     *
+     * @since 22
+     */
+    public static List<CssMetaData<? extends Styleable, ?>> initStyleables(
```
mstr2, line 341: I think this enhancement shouldn't be a part of this PR.

andy-goryachev-oracle: created  [JDK-8320796](https://bugs.openjdk.org/browse/JDK-8320796)

modules/javafx.graphics/src/main/java/com/sun/javafx/css/CssUtil.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.List;
+import javafx.css.CssMetaData;
+import javafx.css.Styleable;
+import javafx.scene.Node;
+import com.sun.javafx.UnmodifiableArrayList;
+
+/**
+ * Utility methods for dealing with CSS.
+ */
+public class CssUtil {
```
mstr2, line 36: I think it's best for utility classes to make it `final` and the constructor `private`.

mstr2 (approved):

modules/javafx.graphics/src/main/java/com/sun/javafx/css/CssUtil.java
andy-goryachev-oracle: this class is not accessible to the outside world, but sure, will do.

mstr2 (approved):

modules/javafx.graphics/src/main/java/com/sun/javafx/css/CssUtil.java
```diff
@@ -67,6 +68,25 @@ private CssUtil() {
         CssMetaData[] combined = new CssMetaData[inheritedFromParent.size() + items.length];
         inheritedFromParent.toArray(combined);
         System.arraycopy(items, 0, combined, inheritedFromParent.size(), items.length);
-        return new UnmodifiableArrayList<>(combined, combined.length);
+        return new ImmutableArrayList<>(combined);
+    }
+
+    /** immutable list with random access backed by an array */
+    private static class ImmutableArrayList<T> extends AbstractList<T> implements RandomAccess {
```
hjohn, line 75: I'd like to point that a skeletal implementation of `AbstractList` will perform worse than `ArrayList` for any method that needs to walk the list, as the `AbstractList` will use iterators for methods like `indexOf`, `hashCode` and `equals`.  Now that will probably be irrelevant, but as this seems to be a micro optimization, you should be aware of all the trade offs.

andy-goryachev-oracle: Good point, thanks!
This also applies to UnmodifiableArrayList.

For completeness sake, I wanted to mention a few issues (not in scope for this PR) that came out of the code review:

- could use `UnmodifiableArrayList` but it stores extra int size.  perhaps a factory method can be added to it for when `size != elements.length`.
- could improve UnmodifiableArrayList with fast(er) `indexOf`, `hashCode`, `equals` per @hjohn 's comment earlier
- `Control.getCssMetaData()` can be improved to skip merging of two lists if skinBase is null
- the same immutable list should be used in `Control.getCssMetaData()` instead of `ArrayList()`

hjohn: I wouldn't do any of the above unless there is a very good reason (and I'm not seeing one).  Just use standard `List.of` as the last step (or use the `Collections.unmodifableList` wrapper); you'll get the most optimized, automatically maintained, bugfree, immutable `List` implementation Java has to offer.  It means another copy will be made; that's fine, this only happens once -- it's not in a hot path.

If you feel like optimizing something, don't bother with `Control.getCssMetaData` either; instead, deduplicate the property lists so there is only one list per unique `Skin` + `Control` combo.  That saves a **complete** list per control **instance**.

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
```diff
@@ -1882,11 +1882,34 @@ <h4><a id="imageview">ImageView</a></h4>
       </thead>
       <tbody>
         <tr>
-        <th class="propertyname" scope="row">-fx-image</th>
+          <th class="propertyname" scope="row">-fx-fit-height</th>
+          <td class="value"><a href="#typenumber" class="typelink">&lt;number&gt;</a></td>
+          <td>0</td>
```
kevinrushforth, line 1887: Since you don't specify a style class for the newly added properties, they will all be left justified. The existing `null` value for `-fx-image` specifies `class="default"` which centers it. I recommend either adding the `class="default"` to all of the newly added properties, or removing it from the default value of the existing `fx-image` property.

I'll reapprove if you want to make the change.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/CssUtil.java
kevinrushforth: I will approve this as is, but I agree with John that it would be better to use one of the existing implementations of List: `new UnmodifiableArrayList`, `List.of`, or `Collections.unmodifiableList`. The cost of the extra copy for `List.of` is in the noise compared to the benefit of not having to maintain yet another special case List class. Similarly, the extra word of data storage in `UnmodifiableArrayList` is in the noise. This is a one-per-class not one-per instance list.

kevinrushforth (approved):
Looks good. I left one HTML formatting comment on the newly added `cssref.html` docs (I also left an editorial comment that you should feel free to ignore).

I'll reapprove if and when you make the suggested  doc formatting change.

kevinrushforth (approved):
Looks good.

Once you update the CSR to match the HTML changes, I'll Review that and you can Finalize it.

andy-goryachev-oracle:
> Once you update the CSR

updated already

mstr2 (approved):

hjohn (approved):

PR #1296, opened 2023-11-27
8320796: CssMetaData.combine()

andy-goryachev-oracle:
Provide a public utility method for use by custom and core Nodes to simplify boilerplate around styleable properties as well as to save some memory.

```
    /**
     * Utility method which combines {@code CssMetaData} items in one immutable list.
     * <p>
     * The intended usage is to combine the parent and the child CSS meta data for
     * the purposes of {@code getClassCssMetaData()} method, see for example {@link Node#getClassCssMetaData()}.
     * <p>
     * Example:
     * <pre>
     * private static final List&lt;CssMetaData&lt;? extends Styleable, ?>> STYLEABLES = CssMetaData.combine(
     *      &lt;Parent>.getClassCssMetaData(),
     *      STYLEABLE1,
     *      STYLEABLE2
     *  );
     * </pre>
     * This method returns an instance of {@link java.util.RandomAccess} interface.
     *
     * @param list the css metadata items, usually from the parent, must not be null
     * @param items the additional items
     * @return the immutable list containing all of the items
     *
     * @since 22
     */
    public static List<CssMetaData<? extends Styleable, ?>> combine(
        List<CssMetaData<? extends Styleable, ?>> list,
        CssMetaData<? extends Styleable, ?>... items)
```

Problem(s):

- the same less-than-optimal implementation is duplicated throughout the javafx code base which combines the parent and child styleable properties, using ArrayList wrapped in a  Collections.unmodifiableList()
- the capacity of underlying ArrayList might exceed the number of items, wasting memory
- a potential exists for the custom component developer to inadvertently create a sub-standard implementation (i.e. return a List which does not implement RandomAccess interface), or forget to include parent's styleables.

Non-Goals:

- not redesigning the lazy initialization of STYLEABLES list
- not changing the way styleables are enumerated in Nodes and Controls
- not adding any new methods to JDK (collections)

To Be Discussed:

- it is not clear whether the order of styleables returned by <N extends Node>.getClassCssMetaData() is of any importance
- the current spec for Node.getCssMetaData() specifies the return value as "The CssMetaData associated with this node, which may include the CssMetaData of its superclasses.", implying that it may or may not include.  It is unclear whether it must include the parent's styleables (well, except the Node whose superclass is Object and thus has no styleable properties).

---------
### Progress
- [ ] Change requires CSR request [JDK-8320936](https://bugs.openjdk.org/browse/JDK-8320936) to be approved

### Issues
 * [JDK-8320796](https://bugs.openjdk.org/browse/JDK-8320796): CssMetaData.combine() (**Enhancement** - P4)
 * [JDK-8320936](https://bugs.openjdk.org/browse/JDK-8320936): CssMetaData.combine() (**CSR**)

### Reviewers
 * [John Hendrikx](https://openjdk.org/census#jhendrikx) (@hjohn - Committer) 🔄 Re-review required (review applies to [8987a125](https://git.openjdk.org/jfx/pull/1296/files/8987a125dd256eece60f754af073fd2831ded3db))

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1296/head:pull/1296` \
`$ git checkout pull/1296`

`$ git checkout pull/1296` \
`$ git pull https://git.openjdk.org/jfx.git pull/1296/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1296`

`$ git pr show -t 1296`

</details>

<a href="https://git.openjdk.org/jfx/pull/1296.diff">https://git.openjdk.org/jfx/pull/1296.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1296#issuecomment-1828481139)

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
+++ b/modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
@@ -338,7 +338,7 @@ public abstract class CssMetaData<S extends Styleable, V> {
      *
      * @since 22
      */
-    public static List<CssMetaData<? extends Styleable, ?>> initStyleables(
+    public static List<CssMetaData<? extends Styleable, ?>> combine(
         List<CssMetaData<? extends Styleable, ?>> list,
         CssMetaData<? extends Styleable, ?>... items)
     {
--- a/modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
+++ b/modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -25,6 +25,7 @@
 
 package javafx.css;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import javafx.scene.Node;
@@ -327,5 +328,26 @@ public abstract class CssMetaData<S extends Styleable, V> {
             .append("}").toString();
     }
 
-
+    /**
+     * Utility method which combines CssMetaData items in one unmodifiable list with the size equal to the number
+     * of items it holds (i.e. with no unnecessary overhead).
+     *
+     * @param list the css metadata items, usually from the parent, not nullable
+     * @param items the additional items
+     * @return the unmodifiable list containing all of the items
+     *
+     * @since 22
+     */
+    public static List<CssMetaData<? extends Styleable, ?>> initStyleables(
+        List<CssMetaData<? extends Styleable, ?>> list,
+        CssMetaData<? extends Styleable, ?>... items)
+    {
+        int sz = list.size() + items.length;
+        ArrayList<CssMetaData<? extends Styleable, ?>> rv = new ArrayList<>(sz);
+        rv.addAll(list);
+        for (CssMetaData<? extends Styleable, ?> p: items) {
+            rv.add(p);
+        }
+        return Collections.unmodifiableList(rv);
+    }
 }
--- a/modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
+++ b/modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
@@ -329,12 +329,24 @@ public abstract class CssMetaData<S extends Styleable, V> {
     }
 
     /**
-     * Utility method which combines {@code CssMetaData} items in one unmodifiable list with size equal
-     * to the number of items it holds.
+     * Utility method which combines {@code CssMetaData} items in one immutable list.
+     * <p>
+     * The intended usage is to combine the parent and the child CSS meta data for
+     * the purposes of {@code getClassCssMetaData()} method, see for example {@link Node#getClassCssMetaData()}.
+     * <p>
+     * Example:
+     * <pre>
+     * private static final List&lt;CssMetaData&lt;? extends Styleable, ?>> STYLEABLES = CssMetaData.combine(
+     *      &lt;Parent>.getClassCssMetaData(),
+     *      STYLEABLE1,
+     *      STYLEABLE2
+     *  );
+     * </pre>
+     * This method returns an instance of {@link java.util.RandomAccess} interface.
      *
      * @param list the css metadata items, usually from the parent, must not be null
      * @param items the additional items
-     * @return the unmodifiable list containing all of the items
+     * @return the immutable list containing all of the items
      *
      * @since 22
      */
--- a/modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
+++ b/modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
@@ -329,10 +329,10 @@ public abstract class CssMetaData<S extends Styleable, V> {
     }
 
     /**
-     * Utility method which combines CssMetaData items in one unmodifiable list with the size equal to the number
-     * of items it holds (i.e. with no unnecessary overhead).
+     * Utility method which combines {@code CssMetaData} items in one unmodifiable list with size equal
+     * to the number of items it holds.
      *
-     * @param list the css metadata items, usually from the parent, not nullable
+     * @param list the css metadata items, usually from the parent, must not be null
      * @param items the additional items
      * @return the unmodifiable list containing all of the items
      *
--- a/modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
+++ b/modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
@@ -25,10 +25,10 @@
 
 package javafx.css;
 
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import javafx.scene.Node;
+import com.sun.javafx.UnmodifiableArrayList;
 
 /**
  * A CssMetaData instance provides information about the CSS style and
@@ -342,12 +342,9 @@ public abstract class CssMetaData<S extends Styleable, V> {
         List<CssMetaData<? extends Styleable, ?>> list,
         CssMetaData<? extends Styleable, ?>... items)
     {
-        int sz = list.size() + items.length;
-        ArrayList<CssMetaData<? extends Styleable, ?>> rv = new ArrayList<>(sz);
-        rv.addAll(list);
-        for (CssMetaData<? extends Styleable, ?> p: items) {
-            rv.add(p);
-        }
-        return Collections.unmodifiableList(rv);
+        CssMetaData[] combined = new CssMetaData[list.size() + items.length];
+        list.toArray(combined);
+        System.arraycopy(items, 0, combined, list.size(), items.length);
+        return new UnmodifiableArrayList<>(combined, combined.length);
     }
 }
--- a/modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
+++ b/modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
@@ -331,32 +331,33 @@ public abstract class CssMetaData<S extends Styleable, V> {
     /**
      * Utility method which combines {@code CssMetaData} items in one immutable list.
      * <p>
-     * The intended usage is to combine the parent and the child CSS meta data for
+     * The intended usage is to combine the parent and the child {@code CssMetaData} for
      * the purposes of {@code getClassCssMetaData()} method, see for example {@link Node#getClassCssMetaData()}.
      * <p>
      * Example:
-     * <pre>
-     * private static final List&lt;CssMetaData&lt;? extends Styleable, ?>> STYLEABLES = CssMetaData.combine(
-     *      &lt;Parent>.getClassCssMetaData(),
+     * <pre>{@code
+     * private static final List<CssMetaData<? extends Styleable, ?>> STYLEABLES = CssMetaData.combine(
+     *      <Parent>.getClassCssMetaData(),
      *      STYLEABLE1,
      *      STYLEABLE2
      *  );
-     * </pre>
-     * This method returns an instance of {@link java.util.RandomAccess} interface.
+     * }</pre>
+     * This method returns an instance of a {@code List} that implements
+     * {@link java.util.RandomAccess} interface.
      *
-     * @param list the css metadata items, usually from the parent, must not be null
+     * @param inheritedFromParent the {@code CssMetaData} items inherited from parent, must not be null
      * @param items the additional items
      * @return the immutable list containing all of the items
      *
      * @since 22
      */
     public static List<CssMetaData<? extends Styleable, ?>> combine(
-        List<CssMetaData<? extends Styleable, ?>> list,
+        List<CssMetaData<? extends Styleable, ?>> inheritedFromParent,
         CssMetaData<? extends Styleable, ?>... items)
     {
-        CssMetaData[] combined = new CssMetaData[list.size() + items.length];
-        list.toArray(combined);
-        System.arraycopy(items, 0, combined, list.size(), items.length);
+        CssMetaData[] combined = new CssMetaData[inheritedFromParent.size() + items.length];
+        inheritedFromParent.toArray(combined);
+        System.arraycopy(items, 0, combined, inheritedFromParent.size(), items.length);
         return new UnmodifiableArrayList<>(combined, combined.length);
     }
 }
--- a/modules/javafx.graphics/src/test/java/test/javafx/css/CssMetaDataTest.java
+++ b/modules/javafx.graphics/src/test/java/test/javafx/css/CssMetaDataTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -1329,4 +1329,25 @@ public class CssMetaDataTest {
                 };
     }
 
+    @Test
+    public void testInitStyleables() {
+        CssMetaData<? extends Styleable, ?> css0 = get(Node.getClassCssMetaData(), "-fx-opacity");
+        CssMetaData<? extends Styleable, ?> css1 = get(Node.getClassCssMetaData(), "-fx-rotate");
+        CssMetaData<? extends Styleable, ?> css2 = get(Node.getClassCssMetaData(), "-fx-focus-traversable");
+        CssMetaData<? extends Styleable, ?> css3 = get(Node.getClassCssMetaData(), "-fx-effect");
+
+        List<CssMetaData<? extends Styleable, ?>> list = List.of(css0, css1);
+
+        List<CssMetaData<? extends Styleable, ?>> rv = CssMetaData.initStyleables(list, css2, css3);
+        assertEquals(4, rv.size());
+        assertSame(css0, rv.get(0));
+        assertSame(css1, rv.get(1));
+        assertSame(css2, rv.get(2));
+        assertSame(css3, rv.get(3));
+
+        rv = CssMetaData.initStyleables(List.of(), css0, css1);
+        assertEquals(2, rv.size());
+        assertSame(css0, rv.get(0));
+        assertSame(css1, rv.get(1));
+    }
 }
--- a/modules/javafx.graphics/src/test/java/test/javafx/css/CssMetaDataTest.java
+++ b/modules/javafx.graphics/src/test/java/test/javafx/css/CssMetaDataTest.java
@@ -1330,7 +1330,7 @@ public class CssMetaDataTest {
     }
 
     @Test
-    public void testInitStyleables() {
+    public void testCombine() {
         CssMetaData<? extends Styleable, ?> css0 = get(Node.getClassCssMetaData(), "-fx-opacity");
         CssMetaData<? extends Styleable, ?> css1 = get(Node.getClassCssMetaData(), "-fx-rotate");
         CssMetaData<? extends Styleable, ?> css2 = get(Node.getClassCssMetaData(), "-fx-focus-traversable");
@@ -1338,14 +1338,14 @@ public class CssMetaDataTest {
 
         List<CssMetaData<? extends Styleable, ?>> list = List.of(css0, css1);
 
-        List<CssMetaData<? extends Styleable, ?>> rv = CssMetaData.initStyleables(list, css2, css3);
+        List<CssMetaData<? extends Styleable, ?>> rv = CssMetaData.combine(list, css2, css3);
         assertEquals(4, rv.size());
         assertSame(css0, rv.get(0));
         assertSame(css1, rv.get(1));
         assertSame(css2, rv.get(2));
         assertSame(css3, rv.get(3));
 
-        rv = CssMetaData.initStyleables(List.of(), css0, css1);
+        rv = CssMetaData.combine(List.of(), css0, css1);
         assertEquals(2, rv.size());
         assertSame(css0, rv.get(0));
         assertSame(css1, rv.get(1));


andy-goryachev-oracle:
On second thought, I don't want to craft another, array-based UnmodifiableRandomAccessList implementation to save one indirection.  Keep in mind that ArrayList is also array-based.

andy-goryachev-oracle:
Someone was here before: there is com.sun.javafx.UnmodifiableArrayList.  We'll use that.

kevinrushforth (commented):
This seems like a useful utility method. I left a few API doc comments.

andy-goryachev-oracle:
updated javadoc, thank you!

modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
```diff
@@ -327,5 +328,23 @@ public int hashCode() {
...
+     * Utility method which combines {@code CssMetaData} items in one unmodifiable list with size equal
+     * to the number of items it holds.
+     *
+     * @param list the css metadata items, usually from the parent, must not be null
+     * @param items the additional items
+     * @return the unmodifiable list containing all of the items
+     *
+     * @since 22
+     */
+    public static List<CssMetaData<? extends Styleable, ?>> initStyleables(
```
mstr2, line 341: Have you considered a method name such as `concat` or `combine`? This method doesn't initialize styleable properties (nor their metadata).

mstr2:
We should also consider changing the specification of `Node.getClassCssMetaData`. It currently says:
```java
    /**
     * Gets the {@code CssMetaData} associated with this class, which may include the
     * {@code CssMetaData} of its superclasses.
     */
```

That's strange. A subclass cannot elect to forgo its inherited properties; yet the specification allows it to pretend it doesn't have metadata for its inherited properties. This should be changed in the following ways:
1. A derived class **must** include the metadata of its base class.
2. The inherited metadata **must** come before the new metadata in the list.

The specification should also point developers to the new `CssMetaData.combine` API.

modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
```diff
@@ -327,5 +328,23 @@ public int hashCode() {
...
+     * @param list the css metadata items, usually from the parent, must not be null
+     * @param items the additional items
+     * @return the unmodifiable list containing all of the items
+     *
+     * @since 22
+     */
+    public static List<CssMetaData<? extends Styleable, ?>> combine(
+        List<CssMetaData<? extends Styleable, ?>> list,
+        CssMetaData<? extends Styleable, ?>... items)
+    {
```
hjohn, line 344: The method is not actually making use of `CssMetaData`, so a more generic utility method would have worked as well:

         public static <T> List<T> combine(List<T> list, T... items)

It could be placed in some kind of utility class (and in fact, many libs offer such a method already).

hjohn (approved):

andy-goryachev-oracle:
> 1. A derived class must include the metadata of its base class.
> 2. The inherited metadata must come before the new metadata in the list.

1. what if the intent is to disallow certain parent properties from being modified by CSS?  I know it's highly unlikely, but what is the reason we should impose this requirement?
2. What is the reason for this particular requirement?  Does SceneView or any other tools rely somehow on the order of properties?

Absent specific reasons, I'd rather leave these as is.

modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
andy-goryachev-oracle: Yeah, perhaps even in List?  
JavaFX does not provide a common utility package, and I am trying to limit the scope by narrowing the accepted data type here.

modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
```diff
@@ -327,5 +328,23 @@ public int hashCode() {
             .append("}").toString();
     }
 
-
+    /**
+     * Utility method which combines {@code CssMetaData} items in one unmodifiable list with size equal
```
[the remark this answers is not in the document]
andy-goryachev-oracle: Since this is not an observable list, the two are semantically equivalent, in my opinion.

@kevinrushforth should we say "immutable" here?

mstr2:
> > 1. A derived class must include the metadata of its base class.
> > 2. The inherited metadata must come before the new metadata in the list.
> 
> 1. what if the intent is to disallow certain parent properties from being modified by CSS?  I know it's highly unlikely, but what is the reason we should impose this requirement?
> 2. What is the reason for this particular requirement?  Does SceneView or any other tools rely somehow on the order of properties?
> 
> Absent specific reasons, I'd rather leave these as is.

1. If this is really a feature (hiding inherited properties from the CSS system), then it should be specified and tested. I don't think this was ever intended to be a feature, and if it was, it's a rather pointless feature. I wouldn't frame this as introducing a new requirement, but merely clarifying the specification.
2. Not that I know, but there's nothing to be gained from allowing controls to mix up the order of their properties. This is performance-critical code, and we might want to benefit from a stable and predictable order at some point in time.

I'd wager a bet that 99,999% of all controls ever created just copy the pattern of extending the list as shown in the core framework, so they're automatically compliant with this proposed spec.

modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
```diff
@@ -327,5 +328,23 @@ public int hashCode() {
...
+     * to the number of items it holds.
```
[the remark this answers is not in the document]
andy-goryachev-oracle: +1 for code sample.

andy-goryachev-oracle:
> 1. I wouldn't frame this as introducing a new requirement, but merely clarifying the specification.

I'd rather leave this as is.  If it is a requirement, then a) there should be a rationale and b) it has to be tested, at least in all of the core classes.  This, frankly, is beyond the scope of this PR, I'd think.

modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
```diff
@@ -327,5 +328,23 @@ public int hashCode() {
...
+     *
+     * @param list the css metadata items, usually from the parent, must not be null
+     * @param items the additional items
+     * @return the unmodifiable list containing all of the items
+     *
+     * @since 22
+     */
+    public static List<CssMetaData<? extends Styleable, ?>> combine(
+        List<CssMetaData<? extends Styleable, ?>> list,
```
nlisker, line 342: Any reason this is specifically a `List`? Can it not be a `Collection`?

mstr2:
> > 1. I wouldn't frame this as introducing a new requirement, but merely clarifying the specification.
> 
> I'd rather leave this as is. If it is a requirement, then a) there should be a rationale and b) it has to be tested, at least in all of the core classes. This, frankly, is beyond the scope of this PR, I'd think.

But I've already explained the rationale: a subclass **cannot** hide its inherited JavaFX properties. This is the way the Java language works. Allowing this method to circumvent basic language semantics isn't sensible API design.

Furthermore, it is already tested by the very fact that this is the current de-facto specification. There are no tests, however, that explicitly test "hidden" JavaFX properties. This is not a feature, this is a specification bug.

nlisker:
I don't see an explanation on why this is needed and what the use cases are (who is this intended for, how it's supposed to be used, and what's the current bad way of doing it). The JBS ticket doesn't mention anything either, and I haven't seen a discussion on the mailing list, which is required for new features.

modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
andy-goryachev-oracle: It should have been a Set maybe, or indeed a Collection.

But it is too late: Node codifies the List return values:

```
public static List<CssMetaData<? extends Styleable, ?>> getClassCssMetaData();

public List<CssMetaData<? extends Styleable, ?>> getCssMetaData();
```

mstr2: Nir was probably referring to the type of the `list` parameter.
As for the return value, not only must this be a `List`, it should also be mandated to be `RandomAccess`. Any implementation that doesn't support random access may cause severe performance degradation.

andy-goryachev-oracle:
> which is required for new features.

There will be a CSR filed since it's an addition to a public API.  It's a small change, and usually there are changes to javadoc and method signatures that make it more time consuming to keep two tickets in sync.

You are right - I'll update the JBS ticket with more information, and in the future I'll make sure to ask for a discussion prior to any feature, however small.

andy-goryachev-oracle:
> But I've already explained the rationale: a subclass **cannot** hide its inherited JavaFX properties.

It can, but only from the CSS subsystem.  Whether it makes sense or not is a different topic, but it can - by not reporting it via getClassCssMetaData();

(verified by removing FONT styleable from TextArea (TextInputControl:1649) - the user cannot change the font via CSS)

kevinrushforth:
> > But I've already explained the rationale: a subclass **cannot** hide its inherited JavaFX properties.
> 
> It can, but only from the CSS subsystem. Whether it makes sense or not is a different topic, but it can - by not reporting it via getClassCssMetaData();
> 
> (verified by removing FONT styleable from TextArea (TextInputControl:1649) - the user cannot change the font via CSS)

Just because you can do something, doesn't mean you should. I agree with @mstr2 that we should specify the intended and current behavior, but I also agree with @andy-goryachev-oracle that it is out of scope for this enhancement. Please file a follow-up bug to update the spec.

kevinrushforth (commented):
I left a few javadoc comments as well as replies to some of the questions.

modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
```diff
@@ -28,6 +28,7 @@
 import java.util.Collections;
 import java.util.List;
 import javafx.scene.Node;
+import com.sun.javafx.UnmodifiableArrayList;
```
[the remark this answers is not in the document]
kevinrushforth: Oops...forgot to delete this. :)

modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
```diff
@@ -327,5 +328,35 @@ public int hashCode() {
...
+     * <pre>
+     * private static final List&lt;CssMetaData&lt;? extends Styleable, ?>> STYLEABLES = CssMetaData.combine(
+     *      &lt;Parent>.getClassCssMetaData(),
+     *      STYLEABLE1,
+     *      STYLEABLE2
+     *  );
+     * </pre>
+     * This method returns an instance of {@link java.util.RandomAccess} interface.
+     *
+     * @param list the css metadata items, usually from the parent, must not be null
```
mstr2, line 347: Two minor things:
1. I think you should either spell out the actual class name `CssMetaData`, or capitalize the `CSS`.
2. Maybe change the parameter name to `inherited`, `inheritedMetaData`, or something like this. I'd also remove the `usually from the parent` part, as this leaves the option that it might come from somewhere else (which isn't intended).

modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
```diff
@@ -327,5 +328,35 @@ public int hashCode() {
             .append("}").toString();
     }
 
-
+    /**
+     * Utility method which combines {@code CssMetaData} items in one immutable list.
+     * <p>
+     * The intended usage is to combine the parent and the child CSS meta data for
```
mstr2, line 334: Very minor: I think `metadata` should be a single word.

modules/javafx.graphics/src/main/java/javafx/css/CssMetaData.java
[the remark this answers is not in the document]
andy-goryachev-oracle: what comment?  :-)

andy-goryachev-oracle:
Thank you for all the feedback and corrections!

nlisker:
I'm unconvinced that this is the way to solve the issue it describes in its doc. The problem is that each control wants to add its own css metadata to that of its parent (recursively), and that the list is immutable (or unmodifiable). The solution offered is a helper method for the general problem of creating a new immutable list from an existing one plus other elements (I agree with John that 
it's another copy of this kind of method, which would probably mean it's a good candidate for the JDK).

What each class brings with itself is a list of its own metadata properties. That's all it should declare. This list should automatically be added to the one of its parent, recursively, forming the final list. This part shouldn't be done by users as it's just another place to make a mistake in.

hjohn:
> What each class brings with itself is a list of its own metadata properties. That's all it should declare. This list should automatically be added to the one of its parent, recursively, forming the final list. This part shouldn't be done by users as it's just another place to make a mistake in.

I definitely always found the implementation for gathering CSS properties rather ugly. The problem however is in part that you can't define a method that only provides the CSS properties of the current class and have it be part of `Node` or some interface for the CSS subsystem to call.  Such a method implementation would have to call `super` itself, as an external caller can't call super methods to gather the full list.

Lacking that it would have to be a static method with a specific signature that needs to be discovered via reflection, and then go up the hierarchy to find all properties.  A static method however won't work for Skin based CSS properties.

Still, there are IMHO better ways to do this; IMHO `Node` should just have a dedicated list of all CSS properties (exposed via `getCssMetaData`) stored in a `CopyOnWriteArrayList` to which the subclasses contribute additional properties at construction time.   The only writes to this list occur at construction time and when Skin changes happen, so this would be an almost ideal usecase for `CopyOnWriteArrayList`.  The `getCssMetaData` method could then have been `final`:

Node:

         private final List<CssMetaData> cssProps = new CopyOnWriteArrayList<>();
         private final List<CssMetaData> cssPropsUnmodifiable = Collections.unmodifiableList(cssProps);
         
         protected final int contributeProperties(List<CssMetaData> props) { 
               cssProps.addAll(props); 
               return cssProps.size(); 
         }

         protected final void replaceProperties(int fromIndex, List<CssMetaData> props) { 
               cssProps.sublist(fromIndex, cssProps.size()).clear(); 
               cssProps.addAll(props); 
         }

         public final Node#getCssMetaData() { return cssPropsUnmodifiable; }
         
Subclasses simply call `contributeProperties` and they're part of the "immutable" list automatically.  `Control` subclasses are a bit more advanced as they first call `contributeProperties` with their fixed properties, store the returned index, and then call `replaceProperties` each time their `Skin` changes.  The `Skin` properties are always at the end, so clearing and re-adding them is easy.

No messing around with concatenations, static methods, overrides that are not calling `super`, etc, and just as performant as there are no object allocations involved when calling `getCssMetaData`.

andy-goryachev-oracle:
> I'm unconvinced that this is the way to solve the issue it describes in its doc. The problem is that each control wants to add its own css metadata to that of its parent (recursively), and that the list is immutable (or unmodifiable).

This is not the problem.

> The solution offered is a helper method for the general problem of creating a new immutable list from an existing one plus other elements (I agree with John that it's another copy of this kind of method, which would probably mean it's a good candidate for the JDK).

I would not go that far as to call it a candidate for JDK, the scope of this change is limited to CssMetaData and is sufficiently different from a general purpose entity.


> What each class brings with itself is a list of its own metadata properties. That's all it should declare. This list should automatically be added to the one of its parent, recursively, forming the final list. This part shouldn't be done by users as it's just another place to make a mistake in.

True.  The only way to avoid mistake is to have this data generated by reflection or pre-processing.  Neither is a good alternative, so I am not solving this particular problem.  The problem I am trying to solve is to provide a slightly more efficient utility method to replace repeating the same code over and over.

andy-goryachev-oracle:
> This introduces yet another moving part, namely that the list of styleable properties can change when the control skin is changed.

This functionality is outside of the scope of this PR, completely unrelated.  The part that is being changed is inside of per-class getClassCssMetaData(), the remaining code and how it's being used by getCssMetaData() is unchanged.

hjohn:
> > stored in a `CopyOnWriteArrayList`
> 
> I am sorry, are you proposing to add two pointers to each Node?

I was responding to Nir, seeing if there is an option to make the entire management of CSS properties simpler.

> > when Skin changes happen
> 
> can you please explain how Skins have managed to enter the discussion? Styleable properties are a part of the Control, they have nothing to do with Skins.

Unfortunately, they do, Skins can have styleable properties, although they seem to be documented as if they belong to the control.  See `TextInputControl` and properties like `-fx-prompt-text-fill`.  This has nothing to do with the PR, but complicates  the implementation (it means for example that the list of styleable properties is not static and can change each time a skin is changed).

kevinrushforth:
There are definitely some (preexisting) oddities of how CSS styleable properties are managed. And it's good to examine them in the context of this newly proposed utility method. Having said that, it still seems to me that this new utility method is useful in providing a better pattern for Node subclasses to create the immutable list of stylables. As an example, see the following from Andy's draft PR to initialize the `STYLEABLES` in Image:

```
    private static final List<CssMetaData<? extends Styleable, ?>> STYLEABLES = CssMetaData.combine(
        Node.getClassCssMetaData(),
        FIT_HEIGHT,
        FIT_WIDTH,
        IMAGE,
        PRESERVE_RATIO,
        SMOOTH
    );
```

Using the current pattern used by other classes, it would look more like this:

```
    private static final List<CssMetaData<? extends Styleable, ?>> STYLEABLES;
        static {
            final List<CssMetaData<? extends Styleable, ?>> styleables =
                new ArrayList<>(Node.getClassCssMetaData());
            styleables.add(FIT_HEIGHT);
            styleables.add(FIT_WIDTH);
            styleables.add(IMAGE);
            styleables.add(PRESERVE_RATIO);
            styleables.add(SMOOTH);
            STYLEABLES = Collections.unmodifiableList(styleables);
        }
    }
```

Seems a useful addition to me.

I do wonder if `combineStyleables` or `ofStyleables` might be a better name than just `combine`? I don't have a strong opinion.

andy-goryachev-oracle:
> I do wonder if `combineStyleables` or `ofStyleables` might be a better name than just `combine`? I don't have a strong opinion.

My initial name was a poor choice, I much prefer @mstr2 's suggestion `combine` now.  I think `CssMetaData.combine()` might be clear enough.

mstr2:
> I do wonder if `combineStyleables` or `ofStyleables` might be a better name than just `combine`? I don't have a strong opinion.

That seems like it would be a misnomer. The proposed method combines the metadata of styleable properties, but a `StyleableProperty` is not a `Styleable`.

kevinrushforth:
OK, let's stick with `combine` then.

nlisker:
> > I'm unconvinced that this is the way to solve the issue it describes in its doc. The problem is that each control wants to add its own css metadata to that of its parent (recursively), and that the list is immutable (or unmodifiable).
> 
> This is not the problem.

But it is according to the CSR: "the same less-than-optimal implementation is duplicated throughout the javafx code base which combines the parent and child styleable properties, using ArrayList wrapped in a Collections.unmodifiableList()".

> > The solution offered is a helper method for the general problem of creating a new immutable list from an existing one plus other elements (I agree with John that it's another copy of this kind of method, which would probably mean it's a good candidate for the JDK).
> 
> I would not go that far as to call it a candidate for JDK, the scope of this change is limited to CssMetaData and is sufficiently different from a general purpose entity.

But it's limited because it has been artificially limited. It's a special case of `<T> List<T> combine(List<T>, T... items)`, which is a general utility method. Just duplicating this method for specific `T`'s doesn't make it more fitting for that scope.

> > What each class brings with itself is a list of its own metadata properties. That's all it should declare. This list should automatically be added to the one of its parent, recursively, forming the final list. This part shouldn't be done by users as it's just another place to make a mistake in.
> 
> True. The only way to avoid mistake is to have this data generated by reflection or pre-processing. Neither is a good alternative, so I am not solving this particular problem. The problem I am trying to solve is to provide a slightly more efficient utility method to replace repeating the same code over and over.

I don't think this is the only way at all. John outlined one solution, I think there is at least one more that I can broadly think of that works in maybe a similar way to how dirty bits are implemented.

andy-goryachev-oracle:
1. the problem is sub-optimal existing code, replicated throughout.
2. yes, it is artificially limited for this particular case, to narrow the scope.  Here we are interested in a) reducing the memory wasted and b) providing a utility to simplify core Nodes and any custom nodes.  A generic List.combine() might be useful, but in this particular case we are interested in an immutable list for a very narrow case.  It's basically an equivalent of adding a different constructor, had we some kind of CssMetaDataList instead of List<CssMetaData>.
3. Adding two pointers to every Node is not a good solution, in my opinion.  Replacing lazy initialization with a busy one is not a good solution.  Am I missing something here?

mstr2:
> But it's limited because it has been artificially limited. It's a special case of `<T> List<T> combine(List<T>, T... items)`, which is a general utility method. Just duplicating this method for specific `T`'s doesn't make it more fitting for that scope.

Functional requirements aren't the only requirements that need to be considered. The CSS system only works well enough with some performance assumptions. As of now, performance requirements are not specified anywhere, but they are certainly implied by the implementation.

I think we have two options here:
1. Specify the performance requirements for `getCssMetaData` (and its associated `getClassCssMetaData`), and provide a utility method that makes it easy to create implementations that meet those requirements.
2. Enhance the CSS system (in this case, `CssStyleHelper`) such that the metadata gathered from public API is internally represented with data structures that decouple internal performance requirements from third-party implementations (basically, copy the metadata received from `getCssMetaData` into internal arrays or some other suitable data structure).

andy-goryachev-oracle:
I give up.

mstr2:
The downside of John's proposal is that the metadata list would now be duplicated in every node, while currently these lists are only duplicated in `Control` nodes. The size of these lists is typically around 20-40 elements.

In any case, expanding on what I wrote earlier, here is how the performance requirements could be specified for `getClassCssMetaData`, if we go for "option 1":
```java
/**
 * Gets an unmodifiable {@code CssMetaData} list for the styleable properties of this class.
 * <p>
 * If a derived class declares new styleable properties, it must redeclare this method to hide the
 * base method. From the redeclared method, it must return the concatenation of the base metadata
 * and the metadata of the newly-added styleable properties.
 * <p>
 * Implementations of this method must satisfy the following performance requirements to prevent
 * severe performance degradation in the CSS system:
 * <ol>
 *     <li>The returned list must support random access.
 *     <li>The method must not allocate a new list on each invocation.
 * </ol>
 * It is strongly advised to use {@link CssMetaData#combine} to create the concatenated list and
 * store it in a static field:
 * <pre>{@code
 *     private static final List<CssMetaData<? extends Styleable, ?>> CSS_METADATA =
 *         CssMetaData.combine(
 *             <Parent>.getClassCssMetaData(),
 *             CUSTOM_CSS_METADATA_1,
 *             CUSTOM_CSS_METADATA_2,
 *             CUSTOM_CSS_METADATA_3
 *         );
 *
 *     public static List<CssMetaData<? extends Styleable, ?>> getClassCssMetaData() {
 *         return CSS_METADATA;
 *     }
 * }</pre>
 * Note that in this example, {@code <Parent>} should be substituted with the name of the direct base
 * class, even if the direct base class does not have its own {@code getClassCssMetaData()} method.
 * This ensures that the code remains correct when such a method is added to the direct base class.
 *
 * @return the {@code CssMetaData} list for the styleable properties of this class
 * @see CssMetaData#combine
 * @since JavaFX 8.0
 */
public static List<CssMetaData<? extends Styleable, ?>> getClassCssMetaData()
```

nlisker:
> There are definitely some (preexisting) oddities of how CSS styleable properties are managed. And it's good to examine them in the context of this newly proposed utility method. Having said that, it still seems to me that this new utility method is useful in providing a better pattern for Node subclasses to create the immutable list of stylables. As an example, see the following from Andy's draft PR to initialize the `STYLEABLES` in Image:
> 
> ```java
>     private static final List<CssMetaData<? extends Styleable, ?>> STYLEABLES = CssMetaData.combine(
>         Node.getClassCssMetaData(),
>         FIT_HEIGHT,
>         FIT_WIDTH,
>         IMAGE,
>         PRESERVE_RATIO,
>         SMOOTH
>     );
> ```
> 
> Using the current pattern used by other classes, it would look more like this:
> 
> ```java
>     private static final List<CssMetaData<? extends Styleable, ?>> STYLEABLES;
>         static {
>             final List<CssMetaData<? extends Styleable, ?>> styleables =
>                 new ArrayList<>(Node.getClassCssMetaData());
>             styleables.add(FIT_HEIGHT);
>             styleables.add(FIT_WIDTH);
>             styleables.add(IMAGE);
>             styleables.add(PRESERVE_RATIO);
>             styleables.add(SMOOTH);
>             STYLEABLES = Collections.unmodifiableList(styleables);
>         }
>     }
> ```
> 
> Seems a useful addition to me.

I pondered that back when I was working on replacing these static initializers with the `.of` collection variants. It doesn't work here for problem stated above - we need to modify an unmodifiable list, which is why I didn't touch them in that pass. While the proposed method is useful for eliminating some ugly syntax, cementing a questionable API with more a public API doesn't seem to me like the right direction. If the method is made internal only, then that's fine. Alternatively, if the method is made useful outside of this specific context, then even if it won't be used here, it could be used in other places, and that's also fine.

hjohn:
> > There are definitely some (preexisting) oddities of how CSS styleable properties are managed. And it's good to examine them in the context of this newly proposed utility method. Having said that, it still seems to me that this new utility method is useful in providing a better pattern for Node subclasses to create the immutable list of stylables. As an example, see the following from Andy's draft PR to initialize the `STYLEABLES` in Image:
> > ```java
> >     private static final List<CssMetaData<? extends Styleable, ?>> STYLEABLES = CssMetaData.combine(
> >         Node.getClassCssMetaData(),
> >         FIT_HEIGHT,
> >         FIT_WIDTH,
> >         IMAGE,
> >         PRESERVE_RATIO,
> >         SMOOTH
> >     );
> > ```
Would this not still require a `static` `getClassCssMetaData` method so subclasses have a way of getting these? `ImageView` is not final, and I didn't see that method.

PR #1299, opened 2023-12-04
Reflective enumeration of styleable properties

mstr2:
---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1299/head:pull/1299` \
`$ git checkout pull/1299`

`$ git checkout pull/1299` \
`$ git pull https://git.openjdk.org/jfx.git pull/1299/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1299`

`$ git pr show -t 1299`

</details>

<a href="https://git.openjdk.org/jfx/pull/1299.diff">https://git.openjdk.org/jfx/pull/1299.diff</a>

</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/css/CssMetaDataCache.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/css/CssMetaDataCache.java
@@ -4,7 +4,6 @@ import javafx.beans.property.Property;
 import javafx.css.CssMetaData;
 import javafx.css.Styleable;
 import javafx.css.StyleableProperty;
-import javafx.scene.Node;
 
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
@@ -19,7 +18,7 @@ public final class CssMetaDataCache {
 
     private CssMetaDataCache() {}
 
-    public synchronized static <T extends Node> List<CssMetaData<? extends Styleable, ?>> getCssMetaData(T node) {
+    public synchronized static List<CssMetaData<? extends Styleable, ?>> getCssMetaData(Styleable node) {
         List<CssMetaData<? extends Styleable, ?>> list = cache.get(node.getClass());
         if (list == null) {
             cache.put(node.getClass(), list = CssMetaDataCache.reflectCssMetaData(node));
@@ -28,7 +27,7 @@ public final class CssMetaDataCache {
         return list;
     }
 
-    private static <T extends Node> List<CssMetaData<? extends Styleable, ?>> reflectCssMetaData(T node) {
+    private static List<CssMetaData<? extends Styleable, ?>> reflectCssMetaData(Styleable node) {
         List<CssMetaData<? extends Styleable, ?>> metadata = new ArrayList<>();
 
         for (Method method : node.getClass().getMethods()) {
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/css/CssMetaDataCache.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/css/CssMetaDataCache.java
@@ -1,0 +1,58 @@
+package com.sun.javafx.css;
+
+import javafx.beans.property.Property;
+import javafx.css.CssMetaData;
+import javafx.css.Styleable;
+import javafx.css.StyleableProperty;
+import javafx.scene.Node;
+
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public final class CssMetaDataCache {
+
+    private static final Map<Class<?>, List<CssMetaData<? extends Styleable, ?>>> cache = new HashMap<>();
+
+    private CssMetaDataCache() {}
+
+    public synchronized static <T extends Node> List<CssMetaData<? extends Styleable, ?>> getCssMetaData(T node) {
+        List<CssMetaData<? extends Styleable, ?>> list = cache.get(node.getClass());
+        if (list == null) {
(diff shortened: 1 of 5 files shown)


modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -8988,304 +9219,18 @@ protected Cursor getInitialCursor() {
...
-        // is referenced no earlier (and therefore loaded no earlier by the
-        // class loader) than the moment that  getClassCssMetaData() is called.
-        // This avoids loading the CssMetaData instances until the point at
-        // which CSS needs the data.
-        //
-        return StyleableProperties.STYLEABLES;
+        return classCssMetaData;
     }
 
+    private static final List<CssMetaData<? extends Styleable, ?>> classCssMetaData = CssMetaData.of(new Node() {});
```
andy-goryachev-oracle, line 9232: perhaps it should be `<S extends Styleable> CssMetaData.of(Class<S>)` ?

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -1512,13 +1547,26 @@ public final BlendMode getBlendMode() {
...
+                static final CssMetaData<Node, BlendMode> METADATA = new CssMetaData<>(
+                        "-fx-blend-mode", new EnumConverter<>(BlendMode.class)) {
+                    @Override
+                    public boolean isSettable(Node node) {
+                        return node.blendMode == null || !node.blendMode.isBound();
+                    }
+
+                    @Override
+                    public StyleableProperty<BlendMode> getStyleableProperty(Node node) {
+                        return (StyleableProperty<BlendMode>)node.blendModeProperty();
```
andy-goryachev-oracle, line 1559: these two methods can be implemented in the base class if we add a
`setGetter(Function<S, StyleableProperty<V>> getter)`
and also
`setIsSettable(Predicate<S>)`
?

This way we don't even need to declare a new class each time.

Just a thought.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/CssMetaDataCache.java
```diff
@@ -0,0 +1,57 @@
...
+import java.util.Map;
+
+public final class CssMetaDataCache {
+
+    private static final Map<Class<?>, List<CssMetaData<? extends Styleable, ?>>> cache = new HashMap<>();
+
+    private CssMetaDataCache() {}
+
+    public synchronized static List<CssMetaData<? extends Styleable, ?>> getCssMetaData(Styleable node) {
+        List<CssMetaData<? extends Styleable, ?>> list = cache.get(node.getClass());
```
andy-goryachev-oracle, line 22: wouldn't this fail in the case of a skin change?

andy-goryachev-oracle (commented):
I like the idea.
Could the amount of boilerplate code be reduced even further?

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -1364,7 +1364,30 @@ public final boolean isVisible() {
     public final BooleanProperty visibleProperty() {
         if (visible == null) {
             visible = new StyleableBooleanProperty(true) {
+                static final CssMetaData<Node, Boolean> METADATA = new CssMetaData<>(
```
andy-goryachev-oracle, line 1367: One problem: this idea requires creation of an instance, bc METADATA static objects get created when an instance is created.
I'd rather avoid this inversion...

mstr2: We're fully inflating a single instance of each particular class once to gather all metadata. After that, all further instances don't need to be fully inflated.

andy-goryachev-oracle: what if this is an abstract custom Node/Control?

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
mstr2: Yes, this can be further simplified.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
mstr2: We would never expect to encounter an instance of an abstract class. Generally, metadata reflection is done the first time we call `getCssMetaData` for any particular class, which we can only do if we have an instance of that class.

The only exceptions to this are the preexisting static `getClassCssMetaData` methods, which are initialized once with a throwaway instance of the class they are declared in, for example in the case of the `Node` class:
```java
private static final List<CssMetaData<? extends Styleable, ?>> classCssMetaData =
    CssMetaData.of(new Node() {});
```

New classes should not copy this pattern of having a static `getClassCssMetaData` method, so the problem shouldn't exist for these classes.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -6094,14 +6154,29 @@ public double getTranslateX() {
...
             if (translateX == null) {
                 translateX = new StyleableDoubleProperty(DEFAULT_TRANSLATE_X) {
+                    static final CssMetaData<Node, Number> METADATA = new CssMetaData<>(
+                            "-fx-translate-x", SizeConverter.getInstance(), 0.0) {
+                        @Override
+                        public boolean isSettable(Node node) {
+                            return node.nodeTransformation == null
+                                || node.nodeTransformation.translateX == null
+                                || node.nodeTransformation.canSetTranslateX();
+                        }
```
andy-goryachev-oracle, line 6164: this is an example of a more complex `isSettable() `logic.

PR #1305, opened 2023-12-12
8320965: Scrolling on a touch enabled display fails on Wayland

jperedadnr:
This PR replaces the deprecated `gdk_pointer_grab` with `gdk_seat_grab`, and `gdk_pointer_ungrab ` with `gdk_seat_ungrab`, using runtime checks and wrapped functions for GTK 3.20+ (so systems without it still run with GTK 3.8+), and fixes the dragging issue on Wayland.

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/gtk/glass_general.cpp
+++ b/modules/javafx.graphics/src/main/native-glass/gtk/glass_general.cpp
@@ -596,7 +596,13 @@ glass_gdk_mouse_devices_grab_with_cursor(GdkWindow *gdkWindow, GdkCursor *cursor
     if (disableGrab) {
         return TRUE;
     }
-
+#if GTK_CHECK_VERSION(3, 20, 0)
+    GdkGrabStatus status = wrapped_gdk_seat_grab(gdkWindow,
+            1 << 0 | 1 << 1 | 1 << 2 /* GDK_SEAT_CAPABILITY_ALL_POINTING */,
+            owner_events, cursor, NULL,
+            NULL /* GdkSeatGrabPrepareFunc */,
+            NULL);
+#else
     GdkGrabStatus status = gdk_pointer_grab(gdkWindow, owner_events, (GdkEventMask)
                                             (GDK_POINTER_MOTION_MASK
                                                 | GDK_POINTER_MOTION_HINT_MASK
@@ -607,13 +613,17 @@ glass_gdk_mouse_devices_grab_with_cursor(GdkWindow *gdkWindow, GdkCursor *cursor
                                                 | GDK_BUTTON_PRESS_MASK
                                                 | GDK_BUTTON_RELEASE_MASK),
                                             NULL, cursor, GDK_CURRENT_TIME);
-
+#endif
     return (status == GDK_GRAB_SUCCESS) ? TRUE : FALSE;
 }
 
 void
-glass_gdk_mouse_devices_ungrab() {
+glass_gdk_mouse_devices_ungrab(GdkWindow *gdkWindow) {
+#if GTK_CHECK_VERSION(3, 20, 0)
+    wrapped_gdk_seat_ungrab(gdkWindow);
+#else
     gdk_pointer_ungrab(GDK_CURRENT_TIME);
+#endif
 }
 
 void
--- a/modules/javafx.graphics/src/main/native-glass/gtk/wrapped.c
+++ b/modules/javafx.graphics/src/main/native-glass/gtk/wrapped.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2023, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -148,3 +148,73 @@ void wrapped_gdk_x11_display_set_window_scale (GdkDisplay *display,
     }
 }
 
+static GdkSeat * (*_gdk_display_get_default_seat) (GdkDisplay *display);
+
+// Note added in libgdk 3.20 which is > our OEL 7.0 version of 3.8
+GdkSeat * wrapped_gdk_display_get_default_seat (GdkDisplay *display)
+{
+#if GTK_CHECK_VERSION(3, 20, 0)
+    if (_gdk_display_get_default_seat == NULL) {
+        _gdk_display_get_default_seat = dlsym(RTLD_DEFAULT, "gdk_display_get_default_seat");
+        if (gtk_verbose && _gdk_display_get_default_seat) {
+            fprintf(stderr, "loaded gdk_display_get_default_seat\n"); fflush(stderr);
+        }
+    }
+#endif
+
+    if (_gdk_display_get_default_seat != NULL) {
+        return (*_gdk_display_get_default_seat)(display);
+    }
+
+    return NULL;
+}
+
+static GdkGrabStatus (*_gdk_seat_grab) (GdkSeat* seat, GdkWindow* window,
+                            int capabilities /* GdkSeatCapabilities capabilities */,
+                            gboolean owner_events, GdkCursor* cursor, const GdkEvent* event,
+                            void * prepare_func /* GdkSeatGrabPrepareFunc prepare_func */,
+                            gpointer prepare_func_data);
+
+// Note added in libgdk 3.20 which is > our OEL 7.0 version of 3.8
+GdkGrabStatus wrapped_gdk_seat_grab(GdkWindow* window,
+                            int capabilities /* GdkSeatCapabilities capabilities */,
+                            gboolean owner_events, GdkCursor* cursor, const GdkEvent* event,
+                            void * prepare_func /* GdkSeatGrabPrepareFunc prepare_func */,
+                            gpointer prepare_func_data)
+{
+#if GTK_CHECK_VERSION(3, 20, 0)
+    if(_gdk_seat_grab == NULL) {
+        _gdk_seat_grab = dlsym(RTLD_DEFAULT, "gdk_seat_grab");
+        if (gtk_verbose && _gdk_seat_grab) {
+            fprintf(stderr, "loaded gdk_seat_grab\n"); fflush(stderr);
+        }
+    }
+#endif
+
+    GdkSeat* seat = wrapped_gdk_display_get_default_seat(gdk_window_get_display(window));
+    if (seat != NULL && _gdk_seat_grab != NULL) {
+        return (*_gdk_seat_grab)(seat, window, capabilities,
+                                 owner_events, cursor, event, prepare_func, prepare_func_data);
+    }
+    return GDK_GRAB_FAILED;
+}
+
+static void (*_gdk_seat_ungrab) (GdkSeat* seat);
+
+// Note added in libgdk 3.20 which is > our OEL 7.0 version of 3.8
+void wrapped_gdk_seat_ungrab(GdkWindow* window)
+{
+#if GTK_CHECK_VERSION(3, 20, 0)
+    if(_gdk_seat_ungrab == NULL) {
+        _gdk_seat_ungrab = dlsym(RTLD_DEFAULT, "gdk_seat_ungrab");
+        if (gtk_verbose && _gdk_seat_ungrab) {
+            fprintf(stderr, "loaded gdk_seat_ungrab\n"); fflush(stderr);
+        }
+    }
+#endif
+
+    GdkSeat* seat = wrapped_gdk_display_get_default_seat(gdk_window_get_display(window));
+    if (seat != NULL && _gdk_seat_ungrab != NULL) {
+        (*_gdk_seat_ungrab)(seat);
+    }
+}
--- a/modules/javafx.graphics/src/main/native-glass/gtk/glass_general.h
+++ b/modules/javafx.graphics/src/main/native-glass/gtk/glass_general.h
@@ -317,7 +317,7 @@ gboolean
 glass_gdk_mouse_devices_grab_with_cursor(GdkWindow * gdkWindow, GdkCursor *cursor, gboolean owner_events);
 
 void
-glass_gdk_mouse_devices_ungrab();
+glass_gdk_mouse_devices_ungrab(GdkWindow *gdkWindow);
 
 void
 glass_gdk_master_pointer_grab(GdkEvent *event, GdkWindow *window, GdkCursor *cursor);
--- a/modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
+++ b/modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
@@ -610,7 +610,7 @@ bool WindowContextBase::grab_mouse_drag_focus() {
 
 void WindowContextBase::ungrab_mouse_drag_focus() {
     WindowContextBase::sm_mouse_drag_window = NULL;
-    glass_gdk_mouse_devices_ungrab();
+    glass_gdk_mouse_devices_ungrab(gdk_window);
     if (WindowContextBase::sm_grab_window) {
         WindowContextBase::sm_grab_window->grab_focus();
     }
@@ -628,7 +628,7 @@ bool WindowContextBase::grab_focus() {
 
 void WindowContextBase::ungrab_focus() {
     if (!WindowContextBase::sm_mouse_drag_window) {
-        glass_gdk_mouse_devices_ungrab();
+        glass_gdk_mouse_devices_ungrab(gdk_window);
     }
     WindowContextBase::sm_grab_window = NULL;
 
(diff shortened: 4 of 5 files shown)


kevinrushforth (commented):
We want the ability to build this on a system that may or may not have GTK 3.20, and then to have the binary produced by the build be able to run on a system that may or may not have GTK 3.20, using the new function if available.

This means that there should not be any compile-time `#if` checks. All checks should be runtime checks, for example, checking whether or not the function pointer obtained via dlsym is NULL vs non-NULL.

I think you already have the needed runtime checks. I was able to compile your PR branch, using headers and libraries from GTK > 3.20, and run it on a system with GTK < 3.20. I confirmed that the dlsym returned NULL on that system (as expected), and it didn't crash.

So it might be as simple as removing the `#if` checks along with the entire "else" blocks, letting the code currently in the `#if` checks be unconditionally compiled. As long as you aren't using anything from a system GTK header file that doesn't exist in 3.8, this should be fine.

modules/javafx.graphics/src/main/native-glass/gtk/glass_general.cpp
```diff
@@ -596,7 +596,13 @@ glass_gdk_mouse_devices_grab_with_cursor(GdkWindow *gdkWindow, GdkCursor *cursor
     if (disableGrab) {
         return TRUE;
     }
-
+#if GTK_CHECK_VERSION(3, 20, 0)
```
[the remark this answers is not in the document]
jperedadnr: Okay, that makes sense. 

I've removed the compile-time `#if-#else` from `wrapped.c`.

However, to do the same in `glass_general.cpp`, if the wrapped functions fail (when dlsym returns null), we still need to fall back to the old implementation, don't we? 

Therefore, I've changed the signature to return a `gboolean` (and removed the `GDK_GRAB_FAILED` enum value that was added in 3.16, as I just noticed).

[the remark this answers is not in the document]
kevinrushforth: This looks good. I'll review more closely and also test it.

kevinrushforth (commented):
This now looks like what I would expect. I'll test it, mainly to look for regressions, since I don't have a touch screen to test the actual fix.

@tsayao Can you be the second reviewer?

modules/javafx.graphics/src/main/native-glass/gtk/wrapped.c
```diff
@@ -148,3 +148,71 @@ void wrapped_gdk_x11_display_set_window_scale (GdkDisplay *display,
...
+        }
+    }
+
+    GdkSeat* seat = wrapped_gdk_display_get_default_seat(gdk_window_get_display(window));
+    if (seat != NULL && _gdk_seat_grab != NULL) {
+        *status = (*_gdk_seat_grab)(seat, window, capabilities,
+                                 owner_events, cursor, event, prepare_func, prepare_func_data);
+        return TRUE;
+    }
+    return FALSE;
```
tsayao, line 197: I did try to test on Ubuntu 16.04 and compilation failed (no surprise because `GdkSeat` does not exists there). Suggestion to keep  `#ifdef` here and `return FALSE` on `#else` so it would still compile on Ubuntu 16.04 and older systems. Will need to `#ifdef` all `GdkSeat` usage.

jperedadnr: I take `GdkSeat` is available since GTK 3.0? https://docs.gtk.org/gdk3/class.Seat.html

But don't we have a minimum set on 3.8.0?

Would this work?
```
#if GTK_CHECK_VERSION(3, 0, 0)
static GdkSeat * (*_gdk_display_get_default_seat) (GdkDisplay *display);
GdkSeat * wrapped_gdk_display_get_default_seat (GdkDisplay *display)
{...}
#endif

...

#if GTK_CHECK_VERSION(3, 0, 0)
    GdkSeat* seat = wrapped_gdk_display_get_default_seat(gdk_window_get_display(window));
    if (seat != NULL && _gdk_seat_grab != NULL) {
        *status = ...
        return TRUE;
    }
#endif
```

tsayao: You're right. GdkSeat should be available. Maybe It needs an include?

tsayao: I think the docs are wrong, I probably exists since 3.20, so maybe check for `GTK_CHECK_VERSION(3, 20, 0);`.

`/home/tsayao/jose/jfx/modules/javafx.graphics/src/main/native-glass/gtk/wrapped.c:200:34: error: unknown type name ‘GdkSeat’
`

jperedadnr: Okay, that is unfortunate (GTK docs inaccurate), but makes sense.

I'll add the compile-time checks to `wrapped.c`.

tsayao: The check is to allow compilation and test on Ubuntu 16.04 - When shipping the final build it should be built on gtk 3.20+, so both dlsym and the check makes sense. 

So when running on Ubuntu 16.04:
- For testing on the platform, the #ifdef will make it possible to compile;
- For running the released built (which should be built on gtk 3.20+) dlsym will fail and fallback;

tsayao: That's only if we want to keep building working on 16.04. I think it makes easier to test on it.
But, it's already failing for Platform preferences API code.

jperedadnr: In any case, if `GdkSeat` is available only since 3.20, then we need to add the compile-time checks anyway, since minimum supported is 3.8.

kevinrushforth: It seems like this is the best we can do for now. What is means is that we won't be able to build on a system with 3.8 - 3.19 and distribute a library that will use the new functionality when running on 3.20 or later. That's probably OK in this case, but is worth noting.

What is important, and why the runtime check is needed (as noted above) is that we can build a production release of JavaFX on a system that has a later GTK and it will run on a system with an older GTK. It won't use the feature on those older systems, but won't crash.

kevinrushforth: As an FYI, we use GTK 3.22 on our CI build machines so I wouldn't want to see the build-time dependency move any higher than it currently is.

tsayao: The suggestion was to allow building on 16.04 - so it's easier to test. But it currently does not build because of code introduced on Platform Preferences API. On the other hand, maybe it's best to fail so we make sure building requires 3.20.

kevinrushforth:
The addition of the compile-time flags looks OK.

I did a build with GTK 3.22 (so it compiles the new code, does the dlsym, and does the runtime check) and verified that there are no regressions when running on an older system (Ubuntu 16.04).

I then did a full test run on our headful test systems, and there is one new test failure -- it seems to be intermittent, although fails pretty consistently on our Ubuntu 22.04 and Ubuntu 20.04 test systems. I can reproduce it locally on a VM running Ubuntu 22.04, where it fails about 1/2 the time with this patch applied (it fails more often on our physical test systems).

```
DatePickerTest > testDatePickerSceneChange FAILED
    java.lang.AssertionError: Timeout: Failed to receive onAction call.
        at org.junit.Assert.fail(Assert.java:89)
        at test.util.Util.waitForLatch(Util.java:400)
        at test.robot.javafx.scene.DatePickerTest.clickDatePickerCalendarPopup(DatePickerTest.java:90)
        at test.robot.javafx.scene.DatePickerTest.testDatePickerSceneChange(DatePickerTest.java:123)
```

Not sure what to make of this. I am not aware of any problems with this test, but it's possible that your fix has exposed a latent issue either in the test or somewhere else.

modules/javafx.graphics/src/main/native-glass/gtk/wrapped.c
kevinrushforth: The native Platform Preferences code is failing for an entirely different reason (a too-old compiler), but you raise an interesting point.

If we really can't support building a distributable binary on a system with a too-old GTK library in a way that enables using a new feature like this, we might prefer to "fail fast" at build time, rather than produce a binary that doesn't support the new capability even when running on a newer system.

If we decide that is a reasonable thing to do, we might want the minimum _build-time_ version of GTK to be higher than the minimum version we support at _runtime_.

I think we shouldn't necessarily tie that decision to this PR. So I think the current approach in this PR: allow it to build on an older GTK without using the new functionality, is probably OK (although not ideal as evidenced by this discussion).

johanvos:
> I did a build with GTK 3.22 (so it compiles the new code, does the dlsym, and does the runtime check) and verified that there are no regressions when running on an older system (Ubuntu 16.04).

That sounds good.

> If we decide that is a reasonable thing to do, we might want the minimum _build-time_ version of GTK to be higher than the minimum version we support at _runtime_.

That makes sense. We already do this with libc. When building (using the devkit based on the OpenJDK devkit), we use later versions of certain libraries, but we make sure the binaries still work on systems with older versions of the libraries. We should clearly state the minimum runtime requirements for some libraries (rather than for distributions, as it is always possible to hack new versions of libraries into old releases) .
What I really want to avoid is having 2 (or more) versions of the binaries for the same platform/arch/os combination.

tsayao:
I did compile (updating gcc) and test on Ubuntu 16.04 and Ubuntu 23.10 with Xorg and Wayland. All OK.

tsayao (approved):
Looks good.

jperedadnr:
@kevinrushforth is there anything else to be done before we can move this PR forward?

kevinrushforth:
This dropped off my radar.

The only pending issue is the test failure I noted in an earlier comment:

> I then did a full test run on our headful test systems, and there is one new test failure -- it seems to be intermittent, although fails pretty consistently on our Ubuntu 22.04 and Ubuntu 20.04 test systems. I can reproduce it locally on a VM running Ubuntu 22.04, where it fails about 1/2 the time with this patch applied (it fails more often on our physical test systems).
> 
> ```
> DatePickerTest > testDatePickerSceneChange FAILED
>     java.lang.AssertionError: Timeout: Failed to receive onAction call.
>         at org.junit.Assert.fail(Assert.java:89)
>         at org.junit.Assert.assertTrue(Assert.java:42)
>         at test.util.Util.waitForLatch(Util.java:400)
>         at test.robot.javafx.scene.DatePickerTest.clickDatePickerCalendarPopup(DatePickerTest.java:90)
>         at test.robot.javafx.scene.DatePickerTest.testDatePickerSceneChange(DatePickerTest.java:123)
> ```
> 
> Not sure what to make of this. I am not aware of any problems with this test, but it's possible that your fix has exposed a latent issue either in the test or somewhere else.

@jperedadnr Can you try it on your system and see if you can reproduce this failure?

jperedadnr:
I'm testing on my local Linux Intel machine (Ubuntu 20.04), and running the test from head, it passes 5 out of 5 times.
After applying this patch, it fails with your same stacktrace 2 out of 5 times.

With some printouts, I can see that when the test fails, in `DatePickerTest::clickDatePickerCalendarPopup` the call to `mouseClick` processes correctly the mouse move at the correct coordinates and calls mouse press and mouse release, but what it should trigger the datePicker setOnAction event (line 168), fails to do so. For some reason, the event doesn't make it to the control.

jperedadnr:
So it seems definitely related to the changes in this PR. 

Running with `GDK_DEBUG=nograbs` works (10 out of 10).

Adding some debug to the native code, and running with grabs, I can see:

When it works:
```
DatePickerTest STANDARD_OUT
    Glass GTK library to load is glassgtk3

loaded gdk_seat_grab
loaded gdk_display_get_default_seat
grab
grab
found schema 'org.gnome.desktop.interface' and key 'scaling-factor'
loaded gdk_seat_ungrab
ungrab
grab
ungrab
found schema 'org.gnome.desktop.interface' and key 'scaling-factor'
ungrab
grab

> Task :systemTests:test
...
```
(so test starts after window is grabbed and focused, as it should be).

When it fails:
```
DatePickerTest STANDARD_OUT
    Glass GTK library to load is glassgtk3

loaded gdk_seat_grab
loaded gdk_display_get_default_seat
grab
grab
found schema 'org.gnome.desktop.interface' and key 'scaling-factor'
loaded gdk_seat_ungrab
ungrab

> Task :systemTests:test
...
```
so window is not grabbed and not focused, and therefore native mouse clicks go elsewhere.

tsayao:
For me it passes on XWayland and fails on Xorg (with Ubuntu 22.04).

tsayao:
It seems when it fails it's because `GtkWindow._ungrabFocus` is called before the mouse release event.

tsayao:
It seems when it fails, button press is reported, but button release is not.

tsayao:
Maybe this: https://docs.gtk.org/gdk3/enum.CrossingMode.html

tsayao:
A shot in the dark since I don't own a touch enabled monitor:

Test 1:

 Add `GDK_SCROLL_MASK` on the original `gdk_pointer_grab` function;

Test 2:

This PR uses `GDK_SEAT_CAPABILITY_ALL_POINTING` which includes the touch masks.

So the equivalent would include `GDK_TOUCH_MASK` on `gdk_pointer_grab`.


I would bet on option 2.

jperedadnr:
Thanks @tsayao, I can definitely test those suggestions.

If I get it right, you are proposing this change:

```
status = gdk_pointer_grab(gdkWindow, owner_events, (GdkEventMask)
                                            (GDK_POINTER_MOTION_MASK
                                                | GDK_POINTER_MOTION_HINT_MASK
+                                                | GDK_TOUCH_MASK // or GDK_SCROLL_MASK
                                                | GDK_BUTTON_MOTION_MASK
                                                | GDK_BUTTON1_MOTION_MASK
                                                | GDK_BUTTON2_MOTION_MASK
                                                | GDK_BUTTON3_MOTION_MASK
                                                | GDK_BUTTON_PRESS_MASK
                                                | GDK_BUTTON_RELEASE_MASK
                                            NULL, cursor, GDK_CURRENT_TIME);
```

But I don't see how this would fix the issue of the failing test on Linux that happens also without a touch enabled display. I tried both, the test still fails (two out of 5 times or so). Also, this is called only if there is no gdk_seat_grab function (which happens before GTK 3.20).

However, if I use `GDK_SEAT_CAPABILITY_NONE` instead of `GDK_SEAT_CAPABILITY_ALL_POINTING`, it passes 10 out of 10, but that would cancel pointer/touch events, wouldn't it?

Is there any chance that the robot implementation (via XTest, that uses `XTestGrabControl`) might not work fine with the new seat_grab approach after GTK 3.20+?

tsayao:
The rationale was:

This tells which events get delivered to the window while grabbing. XWayland might be sensitive to `GDK_TOUCH_MASK` while Xorg is not.

So the Idea was to keep the current way (with `gdk_pointer_grab` or `gdk_device_grab`, and adding the "deliver TOUCH events to me" might fix it.

Another place to investigate is:

```
#define GDK_FILTERED_EVENTS_MASK static_cast<GdkEventMask>(GDK_ALL_EVENTS_MASK \
                & ~GDK_TOUCH_MASK)

```

It seems that Xorg converts touch events to regular mouse events, but XWayland might be different.

jperedadnr:
I see... so removing all the changes from this PR, and using this diff from head instead:

```
diff --git a/modules/javafx.graphics/src/main/native-glass/gtk/glass_general.cpp b/modules/javafx.graphics/src/main/native-glass/gtk/glass_general.cpp
index e7a230b2b6..e89db2a55f 100644
--- a/modules/javafx.graphics/src/main/native-glass/gtk/glass_general.cpp
+++ b/modules/javafx.graphics/src/main/native-glass/gtk/glass_general.cpp
@@ -605,7 +605,8 @@ glass_gdk_mouse_devices_grab_with_cursor(GdkWindow *gdkWindow, GdkCursor *cursor
                                                 | GDK_BUTTON2_MOTION_MASK
                                                 | GDK_BUTTON3_MOTION_MASK
                                                 | GDK_BUTTON_PRESS_MASK
-                                                | GDK_BUTTON_RELEASE_MASK),
+                                                | GDK_BUTTON_RELEASE_MASK
+                                               | GDK_TOUCH_MASK),
                                             NULL, cursor, GDK_CURRENT_TIME);
 
     return (status == GDK_GRAB_SUCCESS) ? TRUE : FALSE;
```

it looks like dragging works on a touch enabled display and Wayland. (No changes needed in `GDK_FILTERED_EVENTS_MASK`).

Tested this change on X11 as well, I don't see any issue.

jperedadnr:
I've just reverted the previous changes, and just applied the touch mask to the `gdk_pointer_grab` function. Tests should be green now.

kevinrushforth (commented):
This is a much less intrusive fix. The tests are indeed green now with no regressions. I don't have access to a Lunux system with a touch enabled display to test it on, but as long as you've verified that it fixes the problem, it's a +1 from me.

kevinrushforth (approved):

kevinrushforth:
@tsayao can you re-review?

tsayao:
@jperedadnr Would you confirm that scroll on a touch screen still works on Xorg ?

My tests looks all good (manual and system/robot).

jperedadnr:
@tsayao yes, I've double checked again. On Wayland and Xorg, I can scroll with touch events on a touch screen with this fix.
The only issue I still see on Wayland: the mouse cursor location doesn't get updated from touch events, only from mouse events (I guess this can be a follow-up issue).

tsayao (approved):
All ok.

PR #1316, opened 2024-01-04
8322964: Optimize performance of CSS selector matching

hjohn:
Improves performance of selector matching in the CSS subsystem. This is done by using custom set implementation which are highly optimized for the most common cases where the number of selectors is small (most commonly 1 or 2). It also should be more memory efficient for medium sized and large applications which have many style names defined in various CSS files.

Due to the optimization, the concept of `StyleClass`, which was only introduced to assign a fixed bit index for each unique style class name encountered, is no longer needed. This is because style classes are no longer stored in a `BitSet` which required a fixed index per encountered style class.

The performance improvements are the result of several factors:
- Less memory use when only very few style class names are used in selectors and styles from a large pool of potential styles (a `BitSet` for potentially 1000 different style names needed 1000 bits (worst case)  as it was not sparse).
- Specialized sets for small number of elements (0, 1, 2, 3-9 and 10+)
- Specialized sets are append only (reduces code paths) and can be made read only without requiring a wrapper
- Iterator creation is avoided when doing `containsAll` check thanks to the inverse function `isSuperSetOf`
- Avoids making a copy of the list of style class names to compare (to convert them to `StyleClass` and put them into a `Set`) -- this copy could not be cached and was always discarded immediately after...

The overall performance was tested using the JFXCentral application which displays about 800 nodes on its start page with about 1000 styles in various style sheets (and which allows to refresh this page easily).  

On JavaFX 20, the fastest refresh speed was 121 ms on my machine.  With the improvements in this PR, the fastest refresh had become 89 ms.  The speed improvement is the result of a 30% faster `Scene#doCSSPass`, which takes up the bulk of the time to refresh the JFXCentral main page (about 100 ms before vs 70 ms after the change).

---------
### Progress

### Issue
 * [JDK-8322964](https://bugs.openjdk.org/browse/JDK-8322964): Optimize performance of CSS selector matching (**Enhancement** - P4)

### Reviewers
 * [Kevin Rushforth](https://openjdk.org/census#kcr) (@kevinrushforth - **Reviewer**) 🔄 Re-review required (review applies to [e4e3a7ce](https://git.openjdk.org/jfx/pull/1316/files/e4e3a7ce8b775232f9acbc4ede7740c319a30078))

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1316/head:pull/1316` \
`$ git checkout pull/1316`

`$ git checkout pull/1316` \
`$ git pull https://git.openjdk.org/jfx.git pull/1316/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1316`

`$ git pr show -t 1316`

</details>

<a href="https://git.openjdk.org/jfx/pull/1316.diff">https://git.openjdk.org/jfx/pull/1316.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1316#issuecomment-1877023611)


modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
```diff
@@ -321,8 +376,8 @@ private boolean matchStyleClasses(StyleClassSet otherStyleClasses) {
     @Override public int hashCode() {
         int hash = 7;
         hash = 31 * (hash + name.hashCode());
-        hash = 31 * (hash + styleClassSet.hashCode());
-        hash = 31 * (hash + styleClassSet.hashCode());
+        hash = 31 * (hash + selectorStyleClassNames.hashCode());
+        hash = 31 * (hash + selectorStyleClassNames.hashCode());
         hash = (id != null) ? 31 * (hash + id.hashCode()) : 0;
         hash = 31 * (hash + pseudoClassState.hashCode());
         return hash;
```
nlisker, line 383: Is it worth caching the hash code for this class?

modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
```diff
@@ -306,7 +361,7 @@ private boolean matchStyleClasses(StyleClassSet otherStyleClasses) {
         if ((this.id == null) ? (other.id != null) : !this.id.equals(other.id)) {
             return false;
         }
-        if (this.styleClassSet.equals(other.styleClassSet) == false) {
+        if (this.selectorStyleClassNames.equals(other.selectorStyleClassNames) == false) {
```
nlisker, line 364: Maybe replace the `==false` with `!`.

nlisker (commented):
I just scrolled quickly through the changes and left some comments,

If deprecation for removal is part of the PR the consequences will need to be discussed. Are the for-removal classes and methods not widely used?

kevinrushforth:
As @nlisker notes, the implications of any deprecation will need to be discussed. In particular, we need to consider what the consequences are for existing applications. If any of them might be used by apps, then simple deprecation (rather than deprecating for removal) might be best.

/reviewers 3
/csr

modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
```diff
@@ -73,31 +76,78 @@ public String getName() {
...
-     * @return the {@code Set} of {@code StyleClass}es
+     * Gets the immutable {@code Set} of style class names of this {@code Selector}.
+     *
+     * @return a set of style class names, never {@code null}, or contains {@code null}s but can be empty
      */
-    public Set<StyleClass> getStyleClassSet() {
-        return styleClassSet;
+    public Set<String> getStyleClassNames() {
+        return selectorStyleClassNames;
     }
```
kevinrushforth, line 140: This is new API, and as such needs an `@since 23`.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/FixedCapacitySet.java
```diff
@@ -0,0 +1,562 @@
+package com.sun.javafx.css;
```
kevinrushforth, line 1: This needs a standard copyright header block.

kevinrushforth (commented):
Left a couple inline comments.

modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
```diff
@@ -73,31 +76,78 @@ public String getName() {
     /**
      * Gets an immutable list of style-classes of the {@code Selector}.
      * @return an immutable list of style-classes of the {@code Selector}
+     * @deprecated for future removal, use {@link #getStyleClassNames()} instead
      */
     public List<String> getStyleClasses() {
```
mstr2, line 81: Have you considered keeping this method (instead of adding a new `getStyleClassNames`), and have `FixedCapacitySet` implement `List` to make it work?

modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
```diff
@@ -73,31 +76,78 @@ public String getName() {
...
+        return Collections.unmodifiableList(new ArrayList<>(selectorStyleClassNames));
```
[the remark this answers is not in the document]
hjohn: I will change this, I just simplified what was there (no need for the loop), and didn't look further.  `copyOf` would be better for sure.

modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
```diff
@@ -232,20 +288,9 @@ public NodeOrientation getNodeOrientation() {
...
-            final StyleClassSet otherStyleClassSet = new StyleClassSet();
-            final List<String> styleClasses = styleable.getStyleClass();
-            for(int n=0, nMax = styleClasses.size(); n<nMax; n++) {
-
-                final String styleClassName = styleClasses.get(n);
-                if (styleClassName == null || styleClassName.isEmpty()) continue;
-
-                final StyleClass styleClass = StyleClassSet.getStyleClass(styleClassName);
-                otherStyleClassSet.add(styleClass);
+            if (!matchesStyleClasses(styleable.getStyleClass())) {
```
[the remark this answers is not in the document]
hjohn: They absolutely can be, but I've left it in the style of the rest of the method (they do `if (matchOnXYZ)` and return `false` for failures, and fall through everywhere to reach the `return true` at the end).  So just trying to keep it consistent without cleaning up the whole method.

modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
hjohn: I put a debug print in to this method, and then started JFXCentral application (which is quite big) -- it's never called, so doesn't look like it is used anywhere.  Selectors are most often stored in `List`s, didn't find any `Map`s or `Set`s that use it as a key.

modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
hjohn: I've kept the style the same, there is another line like this just below it.  Don't like it either personally :)

hjohn:
> As @nlisker notes, the implications of any deprecation will need to be discussed. In particular, we need to consider what the consequences are for existing applications. If any of them might be used by apps, then simple deprecation (rather than deprecating for removal) might be best.

The methods cannot be reached easily even though they are public.  It would involve creating a `Selector` using `Selector#createSelector`, and then casting it to `SimpleSelector` to get access to its public methods that are not inherited from `Selector`.  `SimpleSelector` itself does not have a public constructor, and the type `SimpleSelector` is not returned anywhere directly (it is always just `Selector` in the public API).

The only use of one of the deprecated methods is to allow access for the internal class `SelectorPartitioning` which indeed does an `instanceof` to reach it.  The other method is not used anywhere, and seems to have been an attempt to make the selectors available in the same form as you'd find them on nodes `Styleable#getStyleClass` (which is a `List`), but is hard to reach (probably an oversight that it got exposed in the final API).

Google searches turn up no hits in 3rd party code for either of these.

I've posted on the mailing list to discuss it further.

modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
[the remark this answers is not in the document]
hjohn: Thanks, I've updated these.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/FixedCapacitySet.java
```diff
@@ -0,0 +1,562 @@
...
+
+    /**
+     * Freezes this collection, turning it permanently read-only. After freezing,
+     * any method that would modify the collection will instead throw
+     * {@link UnsupportedOperationException}.
+     *
+     * <p>This can be used to avoid wrapping the collection with an unmodifiable
+     * collection or making a read-only copy.
+     */
+    public abstract void freeze();
```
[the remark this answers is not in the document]
hjohn: I've pulled it up, it does make it a bit nicer.

modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
```diff
@@ -323,8 +378,8 @@ private boolean matchStyleClasses(StyleClassSet otherStyleClasses) {
     @Override public int hashCode() {
         int hash = 7;
         hash = 31 * (hash + name.hashCode());
-        hash = 31 * (hash + styleClassSet.hashCode());
-        hash = 31 * (hash + styleClassSet.hashCode());
+        hash = 31 * (hash + selectorStyleClassNames.hashCode());
+        hash = 31 * (hash + selectorStyleClassNames.hashCode());
```
andy-goryachev-oracle, line 382: this pattern destroys some information that otherwise should be used to produce a better hash code.
instead of 
`hash = 31 * (hash + newPart);`
it should be
`hash = 31 * hash + newPart;`

(also in CompoundSelector)

hjohn: I agree that it is not nice :)  However, changing this seems out of scope for this PR as my changes are unrelated to this, and I didn't even touch `CompoundSelector`.  As selectors are never stored in sets/maps, it is hard to justify changing this as part of this PR as it would not be related to improving performance.

andy-goryachev-oracle: these lines discard about 4.5 bits of information on each step, surely that will affect performance, no?

hjohn: I am confused, and maybe I am missing something. As far as I know, this method is not called anywhere. I put a breakpoint in it. JavaFX does not use this method anywhere, nor are Selectors ever used as keys in Sets or Maps.

What am I missing?

andy-goryachev-oracle: you are right, this code does not affect performance (I could not hit a break point here either).
still, since you are touching these lines, why not do it right?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/FixedCapacitySet.java
```diff
@@ -0,0 +1,569 @@
...
+     * throw an {@link IllegalStateException}.
+     *
+     * @param <T> the element type
+     * @param maximumCapacity the maximum possible number of elements the set can hold, cannot be negative
+     * @return a new empty set, never {@code null}
+     */
+    public static <T> FixedCapacitySet<T> of(int maximumCapacity) {
+        return maximumCapacity == 0 ? empty()
+             : maximumCapacity == 1 ? new Single<>()
+             : maximumCapacity == 2 ? new Duo<>()
```
Maran23, line 94: I can confirm that most of the code in my application uses 1-2 styleclasses.
Since all controls have one styleclass by default (some even more, think about `TextInputControl` with `text-input` -> `TextField` with `text-field`), does it make sense to also implement a specialized `Triple` class?

We also often add 2 more styleclasses, so 3 style classes seems like a usecase that happens quite often.
I also can confirm that 4 styleclasses or even more are very rare.
I only found one: (`.label` + 3 added from us)
image

hjohn: Thanks for having a look, it's good to have some more use cases from other applications!  The sets are used both for the style classes on `Node`s but also the set of style classes in CSS selectors (multiple style classes in  selectors are even more rare I think than multiple style classes on `Node`s).

I think from a performance perspective, `Duo` and `Hashless` are likely very closely matched, but I can do some testing in that area.  The reason I say this is that I only got a very minor boost from including `Duo`, but kept it because it is a bit more memory efficient.

If you want, I'm also curious about how many styles are used in your application (you can see this by looking at the size of `StyleClassSet#styleClasses` after all styles are loaded).

modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
hjohn: I can undo the rename so they're not touched.  My problem with fixing this is that I then also should do it for `CompoundSelector`, which is just completely unrelated to the intent of this PR.  I'm pretty sure that within JavaFX we don't change lines that are unrelated to the intent of the change (these lines were only hit because of a clarifying field rename related to this change).

Normally I would do such fixes in an extra commit, where the first commit would be something like "fix: Improve selector match performance" and the second is "fix: Fix bug in Selectors hashCode algorithm"...

... but since everything is squashed as the way of working in FX (instead of having stand alone commits that are fast forwarded when included in master), it would have to be done the official way, which is file a ticket, create a new PR, have it also reviewed, etc.  This puts an additional burden on the already long review queue, with well intentioned solid PR's that are abandoned because of lack of feedback and reviews... Even simple things like fixing warnings I've given up on as I see it taking time away from more pressing issues.

andy-goryachev-oracle: I feel the same kind of frustration sometimes.  The review process is indeed painfully slow.

A different JBS/PR would be fine, I think, something along the lines of "suboptimal hashCode() implementation in ..."

kevinrushforth: I agree with John that changing the existing hash algorithm it really out of scope for this PR (even if it is suboptimal, which it is).

Filing a separate bug he advantage of decoupling it in time as well as not burdening this review. It may or may not even be worth implementing the optimization if no one uses it, but in any case, it can be done later.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/FixedCapacitySet.java
Maran23: I see, thanks for checking that out.

I checked the application, here is the result:
After starting the app:
image

After using it a little bit:
image

---
I also checked out Scenebuilder after using it a little bit:
image

hjohn: Thanks, it looks like a bit less classes are in use when compared to JFXCentral (which has about 1000), but still quite a lot.  With `BitSet` that means if style classes are referenced that happen to have a high bit set, it creates an array to hold 400 bits (using 8 longs).  When few style classes are active, the sets provided by `FixedCapacitySet` will be often smaller and (due to the other optimizations as well) faster.  8 longs is sufficient to hold 8 or 16 style name references depending on the JVM (32/64 bit, compressed pointers y/n).

modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
mstr2: If this method is not used anywhere, why do we need to expose `getStyleClassNames()` as new API to replace this one? I'm a bit puzzled by that, especially since you're saying that the API shouldn't be used. Why create something that shouldn't be used?

I'd rather just document that you shouldn't expect great performance from this method, and be done with it. Changing API in a performance optimization PR seems out of scope.

hjohn: You were talking about `List<String> getStyleClasses()`.  This is not used by FX, and unlikely to be used anywhere by 3rd parties (see reasoning in other comment).

Then there is a 2nd method: `Set<StyleClass> getStyleClassSet()`.  This is/was called by `SelectorPartitioning`, an internal FX class. However, this method makes use of the `StyleClass` wrapper that is hurting performance.

To ensure `SelectorPartitioning` has a fast way to access the styles, I needed a new method: `Set<String> getStyleClassNames`.  As it is called from outside the `javafx.css` package, it has to be `public`.

**Now perhaps you were eluding to something else:** 

Would `SelectorPartitioning` mind if the styles were provided as `List` or `Collection`?  Looking at that code, I think it doesn't really care. It just uses the set as a key, and calls `containsAll` on it.  So perhaps we don't need to deprecate `List<String> getStyleClasses()` then, and we don't need a new API.  Instead I'll have `SelectorPartitioning` be happy with a `Collection` so I also don't have to make make a set implement a `List` (which you can't do without violating the contract of either `List` or `Set` as `hashCode` is defined differently).

The only thing that may land us into trouble here (reusing `List<String> getStyleClasses()`) is that there may be an expectation that it still contains duplicates, and retains the original order with which `SimpleSelector` was created.  This is already not the case (as it is based on the old `BitSet` that was created) -- it is deduplicated, and its order is changed (whether that was the intention is unclear, but that's what it does right now).  Then again, the method is currently not in use...

hjohn: I took another look, and I don't see a better option here than introducing a new method.  Things I looked at:

1. Reuse `List<String> getStyles()` by having the sets implement `List` as well as `Set`
    - this is impossible as the two specify different `equals` and `hashCode`
2. Reuse `List<String> getStyles()` by having the sets implement only `List` but still eliminate duplicates
    - this would violate `List` contract (can't eliminate duplicates)
    - still need fast `containsAll` performance in the rare case we may have a large number of styles in the set...
3. Moving `SimpleSelector` and `CompoundSelector` out of `javafx.css`:
    - arguably, they never should have been there (we ran into this before) but moving them is still an API break (albeit a minor one as they're only accessible by casting) -- we could start this process though, so any public methods we add now will be of no consequence once moved
4. Moving a ton of internal CSS classes to `javafx.css` (but keep most package private) and introduce a package private `getStyleClassNames`:
    - this runs into an issue that **something** eventually needs to be `public` to make use of the CSS engine, and that would be new API then
    
Of the above options **3** is by far the best.  I've looked further into how this can be achieved, and it requires the following:

**Minor API break**

- Move `SimpleSelector` and `CompoundSelector` out of `javafx.css` and into a non-public package -- this is a minor API break as these classes were accessible before if you cast `Selector` to either of these.  You could not construct them however.

This options was discussed before when it was determined these classes should probably never have been made public. They can't be easily moved though because they use package private methods of the abstract base class `Selector`.  However... nowadays we can **seal** classes, see below:

**Fully compatible changes**

- Make `Selector` `sealed`, only allowing `CompoundSelector` and `SimpleSelector` -- this is compatible because `Selector` does not allow arbitrary subclasses (it has a package private constructor)
- Change `Selector`s constructor to be `protected` and also one of its static methods needs to be `protected`.  This does not create new public API as `protected` methods are only accessible by subclasses, and it is now sealed.
- Implement `createMatch` directly in `Selector` (it is `abstract` now) and make it `final` for good measure.  All the information required is already available via `public` accessible methods of its sealed sub types.  This avoids having to access package private fields in `Match` by `CompoundSelector` which will no longer be in the same package.  This is fully compatible.
- Change some internal classes imports to the new location of `CompoundSelector` and `SimpleSelector`.

hjohn: @mstr2 I've created #1333 to show how it would look when we move `SimpleSelector` and `CompoundSelector` to internal packages.  I think that should alleviate most concerns, and we can either integrate this first with a new public method, which will later be internal anyway, or integrate the other change first.

mstr2: Thanks for the detailed explanation. I agree that moving `SimpleSelector` and `CompoundSelector` to an internal package would be the best choice, and would also make the question of new API a non-issue.

hjohn:
I think we can continue with this one, now that #1340 is done.

kevinrushforth:
@hjohn Please merge the latest upstream master to pick up the fix for [JDK-8324182](https://bugs.openjdk.org/browse/JDK-8324182).

modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
hjohn: A new method `getStyleClassNames` was now added as part of #1340 -- we can use that now for the optimization.  Eventually, the `SimpleSelector` and `CompoundSelector` classes will move to internal packages (probably in JFX 24) via #1333

Maran23:
Tested with a big application, did not found any CSS regressions. Did not dive into the code yet.

modules/javafx.graphics/src/test/java/test/com/sun/javafx/css/BitSetTest.java
```diff
@@ -141,45 +138,46 @@ void listenerManagementForSetChangeListenerShouldWorkCorrectly() {
...
+//        set1.add(new StyleClass("xyz", index));  // no idea why this is public API, but I'll take it
+//        set2.add(pseudoClass);
+//
+//        /*
+//         * The two sets above contain elements of different types (PseudoClass and StyleClass)
+//         * and therefore should never be equal, despite their bit pattern being the same:
+//         */
+//
+//        assertNotEquals(set1, set2);
+//    }
```
andy-goryachev-oracle, line 160: should we remove the dead code?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/FixedCapacitySet.java
```diff
@@ -0,0 +1,582 @@
...
+        this.frozen = true;
+    }
+
+    /**
+     * Checks if the set is allowed to be mutated, and throws an
+     * {@link UnsupportedOperationException} otherwise.
+     */
+    protected final void ensureNotFrozen() {
+        if (frozen) {
+            throw new UnsupportedOperationException();
```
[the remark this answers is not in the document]
hjohn: I'm not against adding an explanation, but it's not needed as this is part of the `Set` contract (sets that can't be modified are specified to throw `UnsupportedOperationException`), for example for `add` in `Set`:

     * @throws UnsupportedOperationException if the {@code add} operation
     *         is not supported by this set

The concept of "freezing" the set is not visible to any users, it's just there to avoid having to make a final copy of the set. It's modifiable by the code that created it, but if they freeze it before exposing it to the outside world, then for all intents and purposes it is a read only set for anyone else.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/FixedCapacitySet.java
```diff
@@ -0,0 +1,582 @@
...
+            for (T element : c) {
+                modified |= add(element);
+            }
+
+            return modified;
+        }
+
+        @Override
+        public int hashCode() {
+            return element == null ? 0 : element.hashCode();
```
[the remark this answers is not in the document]
hjohn: We can't, `hashCode` for `Set`s are specified:

> The hash code of a set is defined to be the sum of the hash codes of the elements in the set, where the hash code of a {@code null} element is defined to be zero

This is done so that using a set as a key would hash to the same value regardless of what type of set it is.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/FixedCapacitySet.java
```diff
@@ -0,0 +1,582 @@
...
+
+            return true;
+        }
+
+        @Override
+        public boolean addAll(Collection<? extends T> c) {
+            boolean modified = false;
+
+            for (T element : c) {
+                modified |= add(element);
```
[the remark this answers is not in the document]
hjohn: You are right, I hadn't noticed this, probably when I was splitting the sets in different capacities, but the above implementation is indeed the same as in `AbstractCollection#addAll`.  I will remove it.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/FixedCapacitySet.java
```diff
@@ -0,0 +1,582 @@
...
+        }
+
+        @Override
+        public int hashCode() {
+            int h = 0;
+
+            for (int i = 0; i < elements.length; i++) {
+                T element = elements[i];
+
+                h += element == null ? 0 : element.hashCode();
```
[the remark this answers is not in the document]
hjohn: This can't be changed, see other comment, but in short, `Set`s have a specified the implementation for `hashCode` to make them interoperable.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/FixedCapacitySet.java
```diff
@@ -0,0 +1,582 @@
...
+     * @param <T> the element type
+     * @param maximumCapacity the maximum possible number of elements the set can hold, cannot be negative
+     * @return a new empty set, never {@code null}
+     */
+    public static <T> FixedCapacitySet<T> of(int maximumCapacity) {
+        return maximumCapacity == 0 ? empty()
+             : maximumCapacity == 1 ? new Single<>()
+             : maximumCapacity == 2 ? new Duo<>()
+             : maximumCapacity < 10 ? new Hashless<>(maximumCapacity)  // will reject negative values
+                                    : new OpenAddressed<>(maximumCapacity);
```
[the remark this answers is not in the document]
hjohn: It's possible to use the standard `HashSet` as a final fallback (or even to replace `OpenAddressed`), although it would need to be wrapped so support the `freeze` method (or we need to implement this differently).

There are a few reasons however not to do so:
- `HashSet` doesn't support `freeze` so we'd need to copy it after it is created; we can't use `Set.of` as this would also require making a copy of the entries first (the entries are added one by one because empty values are filtered) -- the solution in this PR carefully avoids all unnecessary copying and allocations to achieve the better performance
- The sets are indeed in almost all cases very small (hence the optimizations for small sets) -- they contain style classes that apply to a single node or a single selector.  Nodes with many style classes applied to them (ie. 10 or more) I think rarely occurs in practice. Same goes for selectors; you'd have to have something like `a > b c d e f g h i > j` as selector -- remember that this doesn't apply to something like `a, b, c, d, e, f, g, h, i, j`, as they are actually 10 different selectors with 1 style each.
- `HashSet` is less memory efficient (between 21.5 to 26.7 bytes per entry with the load factor varying between 0.75 and 0.75 * 0.5 and a base cost of 16 bytes per entry) vs `OpenAddressed` which requires between 5.7 and 13.3 bytes per entry (due to its load factor variation between 0.3 and 0.7 with a base cost of 0 bytes per entry) -- why isn't the JDK using open addressed then you may wonder? Well, it actually does in its `Set.of` implementation, but not for `HashSet`, primarily because `HashSet` has less worse edge cases when hash codes are poor, and it has to be more generally usable (open addressed tables degrade easier and require more maintenance when under constant modification, which doesn't apply for us here).

modules/javafx.graphics/src/test/java/test/com/sun/javafx/css/BitSetTest.java
hjohn: Yeah, I guess we can -- it's unlikely we'll ever make another implementation of `BitSet` -- it's more likely that after this PR we may also do a similar change for `PseudoClassState` and remove `BitSet` completely :)

modules/javafx.graphics/src/main/java/javafx/css/StyleClass.java
```diff
@@ -27,6 +27,7 @@
 /**
  * A class that contains {@code StyleClass} information.
  * @since 9
+ * @deprecated for removal
  */
 public final class StyleClass {
```
nlisker, line 32: Needs the `@Deprecated(forRemoval = true)` annotation on the class, and the doc tag needs an explanation of why it's being removed ("erroneously exposed", "no longer needed", "flawed from the start"...) and what should be used instead, if at all.

hjohn: I added it.

Technically, we could leave this class in, but it would just be a dead class that serves no purpose within FX, at least, once the `SimpleSelector` and `CompoundSelector` classes have been moved to internal packages.

I'm wondering if I should add any deprecations in this PR at all, as now it requires a CSR.  This could be all handled in the PR that moves these classes to internal packages.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/FixedCapacitySet.java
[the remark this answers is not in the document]
andy-goryachev-oracle: But it is visible: **public** `void freeze();`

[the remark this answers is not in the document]
kevinrushforth: I'm not sure there is a need, but if text is added, it should be something simple like "Unmodifiable set". I would _not_ use the word "freeze" since there is no such concept in the API. From the POV of the application, the set is unmodifiable.

modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
```diff
@@ -76,32 +77,82 @@ public String getName() {
...
+        return List.copyOf(selectorStyleClassNames);
+    }
+
+    /**
+     * Gets the immutable {@code Set} of {@code StyleClass}es of the {@code Selector}.
+     * @return the {@code Set} of {@code StyleClass}es
+     * @deprecated {@code StyleClass} wrapper has become unnecessary, use {@link #getStyleClassNames()} instead
+     */
+    @Deprecated(forRemoval = true)
+    public Set<StyleClass> getStyleClassSet() {
```
kevinrushforth, line 93: Given that the class itself is deprecated for removal, marking this method as deprecated is unnecessary.

modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
```diff
@@ -76,32 +77,82 @@ public String getName() {
     /**
      * Gets an immutable list of style-classes of the {@code Selector}.
      * @return an immutable list of style-classes of the {@code Selector}
+     * @deprecated use {@link #getStyleClassNames()} instead
      */
+    @Deprecated(forRemoval = true)
```
kevinrushforth, line 82: Given that the class itself is deprecated for removal, marking this method as deprecated is unnecessary.

modules/javafx.graphics/src/main/java/javafx/css/StyleClass.java
kevinrushforth: Ideally, you would not need any deprecations in this PR.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/FixedCapacitySet.java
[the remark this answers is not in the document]
andy-goryachev-oracle: makes sense, thanks!

hjohn:
I think this PR now no longer needs a CSR.

andy-goryachev-oracle (approved):
looks good to me.

how do we undo the csr command?

hjohn:
I think its /csroff or something...

kevinrushforth:
I was about to reply, but I see that the Skara bot beat me to it. :)

I'll double-check later (or Andy can), and if there are no public API implications, issue the `/csr unneeded` command.

andy-goryachev-oracle:
@hjohn is right - no public API changes remain in this PR.

hjohn:
I think this is ready to be integrated. Let me know if there is anything still missing.

mstr2 (approved):
Looks good to me.

modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
```diff
@@ -78,30 +79,65 @@ public String getName() {
...
+        if (cachedStyleClasses == null) {
+            cachedStyleClasses = getStyleClassNames().stream().map(SimpleSelector::getStyleClass).collect(Collectors.toUnmodifiableSet());
+        }
+
+        return cachedStyleClasses;
+    }
+
+    /*
+     * Copied from removed StyleClassSet to give StyleClasses a fixed index when
+     * first encountered. No longer needed once StyleClass is removed.
```
Maran23, line 99: Is this a possible idea for the future?

Maran23 (approved):
Code looks good to me. As mentioned above, I tested it already and everything was good, so I'm certain that there is no regression here.

I'm generally not a big fan of 'reimplementing' Collections, but I can see why it is needed here and it also makes sense, especially for something like CSS which needs to be fast.
Something I saw as well in Swing (just check the `ArrayTable` class 😄 ).

I wonder if we may want to add some tests for the `FixedCapacitySet`?

kevinrushforth:
I'm in the process of testing this. I'll do that and report back.

kevinrushforth (approved):
Code looks good. Testing looks good.

modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
hjohn: Yes, once this class is moved to private API (see https://github.com/openjdk/jfx/pull/1333) then this method and `StyleClass` can be removed altogether.  The `SimpleSelector` class is already deprecated for removal since 23.

The method is almost impossible to reach (you'd have to cast to `SimpleSelector`), but removing it as part of this PR would have required a CSR making this harder to integrate.

hjohn:
> Code looks good to me. As mentioned above, I tested it already and everything was good, so I'm certain that there is no regression here.
> 
> I'm generally not a big fan of 'reimplementing' Collections, but I can see why it is needed here and it also makes sense, especially for something like CSS which needs to be fast. Something I saw as well in Swing (just check the `ArrayTable` class 😄 ).

I'm not a fan either, especially because custom collection implementations when referred to be their interface may sneak their way through the various classes and eventually be returned to users.  Any deviation from the collection contracts then suddenly is a bug users may encounter.  I would have done this with plain `HashSet` were (or `Set.of`) were it not for the fact that a lot of performance gains were possible due to the limited way FX is using these sets.

> I wonder if we may want to add some tests for the `FixedCapacitySet`?

Yeah, now that it is more likely that this will make it into FX, I will add a small set of unit tests for this class.

kevinrushforth:
> > I wonder if we may want to add some tests for the `FixedCapacitySet`?
> 
> Yeah, now that it is more likely that this will make it into FX, I will add a small set of unit tests for this class.

Since this PR is ready to integrate, I think it would be fine to file a new test bug for the additional tests if you like. If you prefer to add the new tests now, that's fine, too (we can re-review it).

hjohn:
> > > I wonder if we may want to add some tests for the `FixedCapacitySet`?
> > 
> > 
> > Yeah, now that it is more likely that this will make it into FX, I will add a small set of unit tests for this class.
> 
> Since this PR is ready to integrate, I think it would be fine to file a new test bug for the additional tests if you like. If you prefer to add the new tests now, that's fine, too (we can re-review it).

I'm fine integrating this as-is and adding a test soon after. I will leave this over the weekend to give others time to review.  

Also some clarification on the contributing rules: "all Reviewers who have requested the chance to review have done so" -- does the indication at the top right of the PR count towards this or should it be a comment? :)  In the first case, @nlisker and @arapte, please indicate if you wish to review this still.

nlisker:
The code looks good. I didn't test it, but I'm fine with integrating.

kevinrushforth:
> Also some clarification on the contributing rules: "all Reviewers who have requested the chance to review have done so" -- does the indication at the top right of the PR count towards this or should it be a comment? :) In the first case, @nlisker and @arapte, please indicate if you wish to review this still.

If someone wants you to wait for them, they should make it clear by adding a comment. Also if someone has given substantive feedback, but hasn't (re)approved, it's good to give them a change to review.

@arapte can add a comment if he wants to review, otherwise go ahead and integrate on Monday.

hjohn:
@kevinrushforth @andy-goryachev-oracle @Maran23 @mstr2 

I couldn't leave this without a proper test, so I created a test for `FixedCapacitySet`.

I didn't discover any issues, but did discover there was a line of code I couldn't cover in the `OpenAddressed` variant (not even with a specially crafted test case).  I checked the code, and realized the situation can never occur, so I've removed those lines and added a comment and `assert` instead.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/FixedCapacitySet.java
```diff
@@ -0,0 +1,547 @@
...
+            this.elements = (T[]) new Object[1 << shift];
+            this.mask = (1 << shift) - 1;
+
+            /*
+             * Note: the size of the elements array MUST always be greater than the
+             * requested capacity as the contains check relies on there always being
+             * at least one empty bucket.
+             */
+
+            assert elements.length > requestedCapacity : "must have more buckets than capacity";
```
mstr2, line 419: This just asserts that the previous three lines of code were correct. Seems a bit excessive to me, if you really need this tested, I'd prefer a unit test.

hjohn: True, but I think it's a lot easier to follow than the logic before it. I see it a bit more as a hint to future maintainers (which might be me) that there are assumptions being made here that code relies on.

mstr2: I understand that, my objection is only with the `assert` keyword. I don't like test code in production code (which this is, it verifies that an assumption holds). I think only unit tests should contain test code, but it's a minor issue and I'll reapprove as-is.

mstr2 (approved):

modules/javafx.graphics/src/main/java/com/sun/javafx/css/FixedCapacitySet.java
andy-goryachev-oracle: I think (and please correct me if I am wrong) that by default java assertions are disabled.  They are only enabled for the tests, see build.gradle :2106 :2628

andy-goryachev-oracle (approved):

modules/javafx.graphics/src/test/java/test/com/sun/javafx/css/FixedCapacitySetTest.java
```diff
@@ -0,0 +1,225 @@
+package test.com.sun.javafx.css;
```
Maran23, line 1: The Copyright is missing from this class and the `if` and `for` clauses have no space after.
I think we usually do that, still minor of course. Tests are looking good to me!

Maran23 (approved):

modules/javafx.graphics/src/test/java/test/com/sun/javafx/css/FixedCapacitySetTest.java
andy-goryachev-oracle: oops I missed that!  I think we must include the standard copyright header per OCA agreement.

andy-goryachev-oracle (changes requested):
please add copyright header

modules/javafx.graphics/src/test/java/test/com/sun/javafx/css/FixedCapacitySetTest.java
hjohn: I always forget the header, thanks, I added it!

andy-goryachev-oracle (approved):
thank you!  sorry, I missed that earlier.

mstr2 (approved):

PR #1325, opened 2024-01-10
8324797: Code example in JavaDoc of ObservableValue#when doesn't compile

paltherr:
---------
### Progress

### Issue
 * [JDK-8324797](https://bugs.openjdk.org/browse/JDK-8324797): Code example in JavaDoc of ObservableValue#when doesn't compile (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1325/head:pull/1325` \
`$ git checkout pull/1325`

`$ git checkout pull/1325` \
`$ git pull https://git.openjdk.org/jfx.git pull/1325/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1325`

`$ git pr show -t 1325`

</details>

<a href="https://git.openjdk.org/jfx/pull/1325.diff">https://git.openjdk.org/jfx/pull/1325.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1325#issuecomment-1913265117)

### DIFF
--- a/modules/javafx.base/src/main/java/javafx/beans/value/ObservableValue.java
+++ b/modules/javafx.base/src/main/java/javafx/beans/value/ObservableValue.java
@@ -272,8 +272,8 @@ public interface ObservableValue<T> extends Observable {
      * <p>
      * For example:
      * <pre>{@code
-     * ObservableValue<Boolean> condition = new SimpleBooleanProperty(true);
-     * ObservableValue<String> longLivedProperty = new SimpleStringProperty("A");
+     * Property<Boolean> condition = new SimpleBooleanProperty(true);
+     * Property<String> longLivedProperty = new SimpleStringProperty("A");
      * ObservableValue<String> whenProperty = longLivedProperty.when(condition);
      *
      * // observe whenProperty, which will in turn observe longLivedProperty


kevinrushforth:
@paltherr Good  catch. This needs a JBS bug ID. Maybe @hjohn would help sponsor this for you (in which case he could create the doc bug)?

paltherr:
If I can file the bug, I'm glad to do it.

It's a tiny bug. I noticed it while looking into Java FX for a new project. As I had no experience of FX, it wasn't immediately clear to me how to fix the example. Once I got the answer, I thought that I could as well quickly fix the documentation...

kevinrushforth:
See the [CONTRIBUTING guidelines](https://github.com/openjdk/jfx/blob/master/CONTRIBUTING.md) for information about how to file a bug (along with other information about pull requests). As this is a simple doc bug, you won't need a very detailed description in the bug report.

hjohn:
I don't mind the change, but there is nothing wrong with it as far as I can see.  `ObservableValue` is used for the types as we're documenting `ObservableValue`.

kevinrushforth:
The existing example code doesn't compile, since there is no `ObservableValue::setValue` method.

paltherr:
I did a quick search to see if there other instances of the same problem under [jfx/modules...](https://github.com/openjdk/jfx/blob/master/modules/). I didn't find any but I noticed that [here](https://github.com/openjdk/jfx/blob/37d7561c390f706b9e711dd535580fe5ae76e879/modules/javafx.base/src/main/java/javafx/beans/value/ObservableValue.java#L160) and [here](https://github.com/openjdk/jfx/blob/37d7561c390f706b9e711dd535580fe5ae76e879/modules/javafx.base/src/main/java/javafx/beans/value/ObservableValue.java#L191) in the same file you have the same issue and avoid it by using a `var`, which looks cleaner. My fix should probably do the same.

Below is the current code of the example. None of the calls to `setValue` compile.

```
     ObservableValue<Boolean> condition = new SimpleBooleanProperty(true);
     ObservableValue<String> longLivedProperty = new SimpleStringProperty("A");
     ObservableValue<String> whenProperty = longLivedProperty.when(condition);

     // observe whenProperty, which will in turn observe longLivedProperty
     whenProperty.addListener((ov, old, current) -> System.out.println(current));

     longLivedProperty.setValue("B");  // "B" is printed

     condition.setValue(false);

     // After condition becomes false, whenProperty stops observing longLivedProperty; condition
     // and whenProperty may now be eligible for GC despite being observed by the ChangeListener

     longLivedProperty.setValue("C");  // nothing is printed
     longLivedProperty.setValue("D");  // nothing is printed

     condition.setValue(true);  // longLivedProperty is observed again, and "D" is printed
```

kevinrushforth:
> I noticed that here and here in the same file you have the same issue and avoid it by using a var, which looks cleaner. My fix should probably do the same.

Either is fine with me. I'll let @hjohn decide, since he wrote all three examples.

kevinrushforth:
@paltherr Can you enable GitHub actions for your repo so that the next time you push a commit GitHub will run them? To do this, go to the "Actions" page in your repo and press the green "I understand my workflows..." button. It's very unlikely to be a problem for this sort of PR, but we ask everyone to do that.

paltherr:
I have enabled workflows. I have also filed an issue but it looks like this will go through a review of its own: review ID : 9076496 :-( In retrospect, I should have added a link to this PR.

kevinrushforth:
> I have enabled workflows. I have also filed an issue but it looks like this will go through a review of its own: review ID : 9076496 :-( In retrospect, I should have added a link to this PR.

I took care of the bug. It is now visible as [JDK-8324797](https://bugs.openjdk.org/browse/JDK-8324797). You can update this PR title with the bug ID and move this PR out of draft.

kevinrushforth:
> You can update this PR title with the bug ID

And change the title to match. So:

```
8324797: Code example in JavaDoc of ObservableValue#when doesn't compile
```

modules/javafx.base/src/main/java/javafx/beans/value/ObservableValue.java
```diff
@@ -272,8 +272,8 @@ default <U> ObservableValue<U> flatMap(Function<? super T, ? extends ObservableV
      * <p>
      * For example:
      * <pre>{@code
-     * ObservableValue<Boolean> condition = new SimpleBooleanProperty(true);
-     * ObservableValue<String> longLivedProperty = new SimpleStringProperty("A");
+     * Property<Boolean> condition = new SimpleBooleanProperty(true);
```
paltherr, line 275: A better alternative might be to use a `var` like at line 160 and 191.

hjohn:
@paltherr thanks for the change, I hadn't checked the rest of the code so missed the compile error.

hjohn (approved):

modules/javafx.base/src/main/java/javafx/beans/value/ObservableValue.java
hjohn: I think it is fine as is, some extra type information, especially in examples, never hurts.

kevinrushforth (approved):

nlisker:
By the way, [code snippets](https://openjdk.org/jeps/413) brought a solution to this with external snippets. The example code exists in a standard compiled java file, whose contents can be referenced in the docs.

kevinrushforth:
> By the way, [code snippets](https://openjdk.org/jeps/413) brought a solution to this with external snippets. The example code exists in a standard compiled java file, whose contents can be referenced in the docs.

True, but we can't use them yet. (the minimum JDK needed to compile JavaFX is still 17).

PR #1326, opened 2024-01-10
8323511: Scrollbar Click jumps inconsistent amount of pixels

FlorianKirmaier:
As seen in the unit test of the PR, when we click on the area above/below the scrollbar the position jumps - but the jump is now not always consistent.
In the current version on the last cell - the UI always jumps to the top. In the other cases, the assumed default cell height is used.

With this PR, always the default cell height is used, to determine how much is scrolled.
This makes the behavior more consistent.

Especially from the unit-test, it's clear that with this PR the behavior is much more consistent.

This is also related to the following PR: https://github.com/openjdk/jfx/pull/1194

---------
### Progress
- [x] Change requires CSR request [JDK-8332041](https://bugs.openjdk.org/browse/JDK-8332041) to be approved

### Issues
 * [JDK-8323511](https://bugs.openjdk.org/browse/JDK-8323511): Scrollbar Click jumps inconsistent amount of pixels (**Bug** - P4)
 * [JDK-8332041](https://bugs.openjdk.org/browse/JDK-8332041): Scrollbar Click jumps inconsistent amount of pixels (**CSR**)

### Reviewers

### Contributors
 * Eduard Sedov `<eduard.sdv@web.de>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1326/head:pull/1326` \
`$ git checkout pull/1326`

`$ git checkout pull/1326` \
`$ git pull https://git.openjdk.org/jfx.git pull/1326/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1326`

`$ git pr show -t 1326`

</details>

<a href="https://git.openjdk.org/jfx/pull/1326.diff">https://git.openjdk.org/jfx/pull/1326.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1326#issuecomment-1884770531)


kevinrushforth:
/reviewers 2

Reviewers: @johanvos @andy-goryachev-oracle

kevinrushforth:
Reviewers: @johanvos @andy-goryachev-oracle

/reviewers 2

johanvos:
Thanks for the PR!
Quick observations:

* The unit test is indeed clear and at first sight, it looks to me that this is reasonable behavior. One of the hard parts with the VirtualFlow and related concepts, is that different use case scenarios assume different "reasonable behavior". Since we now have more tests than before, I am increasingly confident that a new change to the behavior is not causing regression.

* Until we have the exact heights of all cells, the different measurable values (thumb location of scrollbar, position of the cell) can not be set simultaneously to desired values. (That is one of the other hard parts). Hence, changing the behavior of the scrollbar may affect the accuracy of the cell positioning in other scenario's. Again, since we have a growing amount of tests, I'm less worried about this.

I'll look into this more detailed.

Maran23:
Agree, with all the tests added, especially in this area in https://github.com/openjdk/jfx/pull/1194 and in https://github.com/openjdk/jfx/pull/1246, it is much easier for us to catch regression. I will also have a look in the next days. I also noted that I got weird scrolling behaviour once, but could never reproduce it.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1545,25 +1545,37 @@ public void scrollTo(int index) {
     }
 
     // will return true if scroll is successful
-    private boolean tryScrollOneCell(int targetIndex, boolean downOrRight) {
+private boolean tryScrollOneCell(int targetIndex, boolean downOrRight) {
```
andy-goryachev-oracle, line 1548: please revert indentation change

andy-goryachev-oracle:
Tested on macOS 14.1.2 with ListView, fixed and variable cell height, using MonkeyTester 

I do see two problems:

1. With variable height cells: if I click below the scrollbar thumb, followed by a click above the thumb, I expect the view to go back to the same position exactly, but it does not (see the screenshots).  I think I've mentioned this problem before in some other PR.

initial condition:
Screenshot 2024-01-12 at 15 39 13

clicked below the thumb:
Screenshot 2024-01-12 at 15 39 17

clicked above the thumb:
Screenshot 2024-01-12 at 15 39 22


2. this one is more serious.  sometimes it gets into state when clicking below the thumb does not move the scroll bar at all, here is the screenshot:
Screenshot 2024-01-12 at 15 38 44

It seems to happen when the cell height exceeds the viewport height AND the cell is positioned exactly at the top of the viewport.

Can you see it?

johanvos:
> 1. With variable height cells: if I click below the scrollbar thumb, followed by a click above the thumb, I expect the view to go back to the same position exactly, but it does not (see the screenshots).  I think I've mentioned this problem before in some other PR.

This is one of the major difficulties with the VirtualFlow. The word "expect" is very context-dependent. Different scenario's have different expectations. As long as there is nothing specified in the JavaDoc, I don't think one can expect anything.
In this particular case, for example, I can think of a number of reasons why you will not go back to the exact same position.
* a cell has been added (outside the viewport).
* the contents of a cell (outside the viewport) have changed, leading to different sizes
* the estimates for cell dimensions have been updated

The location of the scrollbar thumb does not depend on previous actions, but on the current estimated position in the total sheet, which may vary for a number of reasons. We have the implicit restriction that even though the estimated position may change intermediately, we will not reposition the scrollbar thumb unless there is an explicit user action (scrolling or jumping etc).

For some time, we have been going back and forth to patch VirtualFlow in order to match particular expectations. By making a change, it is often very likely that the expectations in some scenario are violated -- although no contract is violated, hence no bug is introduced.
With the changes in VirtualFlow over the past years, we implicitly specify the expectations by adding regression tests.
In the end, it would be good to somehow have an exhaustive specification document.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1552,18 +1552,30 @@ private boolean tryScrollOneCell(int targetIndex, boolean downOrRight) {
...
-            T cell = getAvailableCell(targetIndex);
-            setCellIndex(cell, targetIndex);
-            resizeCell(cell);
-            setMaxPrefBreadth(Math.max(getMaxPrefBreadth(), getCellBreadth(cell)));
-            cell.setVisible(true);
-            if (downOrRight) {
-                cells.addLast(cell);
-                scrollPixels(getCellLength(cell));
+            if (targetIndex < 0) {
+                T cell = getCell(targetIndex);
```
Maran23, line 1556: Is there a reason you call `getCell` here, but `getAvailableCell` below?

johanvos:
@FlorianKirmaier I still think this would be a good addition. I believe there is only one open question (from Marius) so it would be great if you can answer that.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
eduardsdv: As I was also involved in fixing this error, I can answer.

The `getAvailableCell()` method creates a new cell if none exists for the specified index (-1). 
The `getCell()` returns the `accumCell` in this case, which is completely sufficient for our case, because we only need the dimensions of the cell here. The cell itself will not be added to the scene-graph.

FlorianKirmaier:
@johanvos 
Great to see you would like to see it merged!
I've worked on this topic together with @eduardsdv , 
So consider his response to be my answer to the question.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
johanvos: That sounds right.
@Maran23 does this answer your question?

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/VirtualScrollBar.java
```diff
@@ -139,11 +139,10 @@ public final BooleanProperty virtualProperty() {
             if (firstVisibleCell != null && firstVisibleCell == lastVisibleCell) {
                 int index = firstVisibleCell.getIndex();
                 if (newValue < oldValue) {
-                    index = Math.max(0, index - 1);
```
johanvos, line 142: Why are the boundary checks (Math.max/min) removed here?

eduardsdv: The ``Math.max(0, index - 1)`` was introduced to fix an IndesOutOfBoundsException ([JDK-8311983](https://bugs.openjdk.org/browse/JDK-8311983)) for mouse clicks above the thumb, but it also introduced this bug (different scroll amount for clicks above and below thumb).
This change fixes the handling of -1 cell indexes in a different way, but for that we have to pass it to ``VirtualFlow.scrollTo(int index)`` and then to  ``VirtualFlow.tryScrollOneCell(int targetIndex, boolean downOrRight)``.
Therefore it is necessary to remove this ``Math.max(..)`` call.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
Maran23: Yes. But is it possible to scroll to an index that is negative? (<0) And what do we estimate then?

eduardsdv: > But is it possible to scroll to an index that is negative?

Yes, both the negative value (``-1``) and the positive value, which is greater than the items count (to be precise ``last item index + 1``), are possible and indicate the direction in which the scrolling takes place.

The ``last item index + 1`` is already handled correctly. We only need to fix the handling of ``-1`` index.

Can you explain your second question in more detail?

eduardsdv:
Are there any open questions or objections? Can this pull request be merged?

johanvos:
I plan to do an in-depth review later this week.

johanvos:
Intuitively, the added test seems to be the right thing to do. It indeed fails before and succeeds after the patch.
However, I'm not sure about the implementation.

The suggested patch changes the conceptual idea of `VirtualFlow.scrollTo(int index)` where a negative index is not specified (this is probably what @Maran23 asked at https://github.com/openjdk/jfx/pull/1326#discussion_r1530233902 .

The way the `scrollTo(int index)` is modified doesn't sound right to me. In case of the test, where the function is called with `index = -1`, it will first try to obtain the cell at index -2 (which will always return false), and then it will try to obtain the cell at index 0 (the one we need to have indeed), and then scroll X pixels where X is the height of the cell at index -1. In the test, all cells except the one at 0 have height 100, so the cell at -1 will have 100 as well, so it will use that. 
This seems a complex way to deal with the issue. It would need a clear definition of "default size" and it changes the implicit assumption that scrollTo(index) should be called with a valid index (which does not have to be visible in the current viewport).

I believe the first thing that needs to be done, is to extend the "specification" that is currently in comments in the VirtualScrollBar:
```
            /*
             * Scroll one cell further in the direction the user has clicked if only one cell is shown.
             * Otherwise, a click on the trough would have no effect when cell height > viewport height.
             */
```
What does this mean if there is no cell further in the direction the user has clicked? Should we them scroll to the top of the current cell (current behavior), or should we use a more gradual scroll (which is the new behavior, which I believe is better indeed)? If the latter is the preferred case, this looks to a behavior that is more similar to the Event that is received when the mousewheel is used (and which invokes `VirtualFlow.scrollPixels(double delta)`)

Maran23:
> The suggested patch changes the conceptual idea of `VirtualFlow.scrollTo(int index)` where a negative index is not specified (this is probably what @Maran23 asked at [#1326 (comment)](https://github.com/openjdk/jfx/pull/1326#discussion_r1530233902) .

Yes, this is exactly what I mean and where I do not know if this is the right approach.

> The way the scrollTo(int index) is modified doesn't sound right to me

Agree, it sounds somewhat weird to me that when `scrollTo` is called with index = -1, that means we just scroll up more gradually (not to the top of the cell).

> If the latter is the preferred case, this looks to a behavior that is more similar to the Event that is received when the mousewheel is used (and which invokes VirtualFlow.scrollPixels(double delta))

I completely agree. This sounds like we may should call `scrollPixels`  directly instead.
As @johanvos mentioned, we also need to change the "specification" in the comment at least.

eduardsdv:
I understand your objections. I will create a new pull request. We can then discuss which solution we use.

But I already have a few questions about it.

In the new pull request, the ``VirtualScrollBar.adjustValue(double pos)`` method will then use the ``VirtualFlow.scrollPixels(double delta)`` method as follows.

````java
public class VirtualScrollBar extends ScrollBar {

    @Override
    public void adjustValue(double pos) {
        if (isVirtual()) {
            IndexedCell<?> cell = flow.getCell(-1);
            double pixels = flow.getFixedCellSize() > 0
                            ? flow.getFixedCellSize()
                            : flow.isVertical()
                              ? cell.getLayoutBounds().getHeight()
                              : cell.getLayoutBounds().getWidth();
            if (pos > getValue()) {
                flow.scrollPixels(pixels);
            } else {
                flow.scrollPixels(-pixels);
            }
        } else {
            super.adjustValue(pos);
        }
    }
}
````

The main question is how far should we scroll when the user clicks on the thumb?

For backwards compatibility reasons, I have used the same logic to calculate the delta pixels as is currently used.
Therefore, to calculate the delta pixels, we probably need to change the visibility of the method ``VirtualFlow.getCellLength(T cell)`` to ``public``. Currently I copied the content of it.
Furthermore, the method ``VirtualFlow.getCell(int index)`` is always called with the index ``-1`` that always returns the ``accumCell``. 
Can we somehow shortcut this or should we use a different logic here?

Should we move the entire delta pixel calculation logic into a new method ``VirtualFlow.getBlockIncrement()``?

**Comment by johanvos** [comment]
That's a good question. Since there are a number of scenario's that change the current position (offset) of the VirtualFlow, we have to be careful that those scenario's are not conflicting, and are using the same code when appropriate. Hence, duplicating code between VirtualFlow and VirtualScrollBar sounds dangerous (what if code is changed on one place but not on the other?)

Rather than shifting some logic to VirtualScrollBar, I believe it would be safer to have VirtualScrollBar calling into VirtualFlow. For mouse-scrollwheel events, this is already the case so it might be a good idea to do something similar here?

**Comment by eduardsdv** [comment]
I added a new method ``VirtualFlow.getBlockIncrement()``, which returns the amount of pixels by which the position of the VirtualFlow should be adjusted when clicking on the scrollbar track.

Now, if only one cell is visible, the ``VirtualScrollBar`` changes the position of the ``VirtualFlow`` simply by calling the VirtualFlow.scrollPixels(double)``.

File: modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java [hunk 4]
```diff
@@ -1800,6 +1800,16 @@ public T getCell(int index) {
         return accumCell;
     }
 
+    /**
+     * The amount of pixels by which to adjust the position when the the scroll bar track is clicked.
```
andy-goryachev-oracle, line 1804: should it be the height of the viewport, so to speak?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1800,6 +1800,16 @@ public T getCell(int index) {
...
+     *
+     * @return the value in pixels
+     */
+    public double getBlockIncrement() {
+        // For reasons of backward compatibility, we use the cell length of the empty cell (cell index -1)
+        return getCellLength(-1);
```
andy-goryachev-oracle, line 1810: How would that work when cell heights are different for each cell?

What if the cell height is much larger than the viewport height (as in, 10 times larger?)

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
eduardsdv: Yes, that would also be my favorite choice for the click on the scrollbar track as well as for the PageUp/PageDown.
But for reasons of backwards compatibility, I didn't want to change it here.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
eduardsdv: I don't like that either, but as I said I didn't want to change it because of backwards compatibility. 

----
I would like to make it configurable or even make the complete scrolling behavior interchangeable.
Because, the actual scrolling behavior is still not consistent. 

If multiple cells are visible, the cells are currently aligned (when clicking on the scrollbar) so that they are either exactly at the top or exactly at the bottom. If only one cell is visible, the scrolling behavior changes to the applying of pixel delta.

For example, if you have two large cells and a small ViewPort so that only one cell is visible.
1) If you click on the thumb below the scroll bar, the ListView is scrolled by a pixel delta so that two cells can become visible. 
2) If you click below scroll thumb again, the ListView scrolls so that the second cell appears at the top. 
3) If you click above the scroll thumb, the ListView scrolls by pixel delta, as the second cell is also larger than the viewport. 

Now the ListView is in a different state than before the click (2).

These two behaviors are each useful for their own situation and should not be mixed.
1) Align the cell at the top or bottom if you have cells of the same size that are smaller than the viewport.
2) Scroll by pixel delta if you have cells of different sizes or if the cells are larger than the viewport.

andy-goryachev-oracle:
I might have raised this question before.  In my opinion, the scrolling in these virtualized views should be more or less equivalent to the scrolling in the ScrollPane.  In other words, using thumb track to page down and then page up (or using PgUp/PgDn keys) should get the control to its original state, unless it hit the rail.

This is how it's currently works in MS Word (which isn't exactly a pinnacle of usability, but something a lot of people across the globe use routinely).  See here using Design -> Paragraph Spacing popup, notice how the initial state is not aligned exactly at the top of the cell, clicking on the track below the thumb to page down, then above the thumb to page up, we arrive at the same exact position:

 
Screenshot 2024-05-06 at 10 29 00

Screenshot 2024-05-06 at 10 29 08

Screenshot 2024-05-06 at 10 29 15

diff 1 and 3:

Screenshot 2024-05-06 at 10 30 41

eduardsdv:
I agree with you, at least with scrolling when clicking on the scrollbar track. 
I would like to change the amount of scroll pixels to the viewport length.

PageUp/PageDown not only scrolls the view, but also moves the selection/focus, so it makes sense to align the cells up/down.

andy-goryachev-oracle:
> PageUp/PageDown not only scrolls the view, but also moves the selection/focus, so it makes sense to align the cells up/down.

You are right about focus/selection!

This brings up another question: for accessibility reasons, do we want to add key bindings for the vertical scrolling in addition to the horizontal ones in #1393 ?  For example, when the cell height exceeds the viewport height, there should be a way to scroll the content vertically using unit increment via keyboard only.

What do  you think?

eduardsdv:
It would be nice if it possible to scroll with the keyboard without moving the selection/focus.

eduardsdv:
I followed @andy-goryachev-oracle's suggestion and changed the block increment to the viewport length.
I also fixed the scroll behavior for the case described in my comment https://github.com/openjdk/jfx/pull/1326#discussion_r1591327052.

eduardsdv:
In my opinion, this PR is now ready.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1800,6 +1800,15 @@ public T getCell(int index) {
         return accumCell;
     }
 
+    /**
+     * The amount of pixels by which to adjust the position when the the scroll bar track is clicked.
+     *
+     * @return the value in pixels
+     */
+    public double getBlockIncrement() {
```
andy-goryachev-oracle, line 1808: Adding a new public method, this PR requires CSR.

andy-goryachev-oracle (approved):
The scrolling finally works as I would expect.  Tested with the MonkeyTester using ListView/TableView pages with fixed and variable cell sizes.

andy-goryachev-oracle:
should the contributor info be updated to include @eduardsdv ?

johanvos:
Good to see this is progressing, I'll do a detailed review later this week.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
johanvos, line 1808: I'd rather not add that method and make `getViewportLength()` public. That would benefit other use cases, and it is more behavior-agnostic. The "viewportLength" is easier to define in a contract than a "blockIncrement" and since this is exactly what you return, it sounds like a reasonable thing to do.

eduardsdv: Good suggestion. I have made the necessary changes.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1911,7 +1910,15 @@ void setViewportLength(double value) {
         this.absoluteOffset = getPosition() * (estimatedSize - viewportLength);
         recalculateEstimatedSize();
     }
-    double getViewportLength() {
+
+    /**
+     * The length of the viewport portion of the VirtualFlow as computed during the layout pass.
```
kevinrushforth, line 1915: Suggestion: use code case, `{@code VirtualFlow}`

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1911,7 +1910,15 @@ void setViewportLength(double value) {
...
+     * For a vertical flow, this corresponds to the height and for a horizontal flow to the width of the clip view,
+     * but it does not necessarily have to be the same value.
+     *
+     * @return the viewport length in pixel
```
kevinrushforth, line 1919: That should be "in pixels" (plural).

Also, please add an `@since 23` tag after the return tag.

kevinrushforth (commented):
@eduardsdv I left a couple comments on the API docs for the new public method.

Regarding the CSR, if you don't have a JBS account, perhaps you can ask @FlorianKirmaier or @johanvos to create the CSR for you.

And I agree with @andy-goryachev-oracle that you should be added a contributor. @FlorianKirmaier will need to do that, since only the author of the PR can add contributors.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
eduardsdv: Done.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
eduardsdv: Done.

eduardsdv:
I ask @FlorianKirmaier to create a CSR ticket.

johanvos:
I did a number of tests with the new approach. It works as expected, and I don't see scenarios that could lead to a performance issue.
FWIW, this is a rather drastic change to the behavior, but there are no technical reasons to not do it. There are currently no tests that assert the current behavior, and I didn't find it in the public docs either (the idea about mapping scroll events to the top/bottom of a cell is hinted in the code though).
 It is very likely that existing application will behave differently after this PR. But I think that is more a UX discussion than a code discussion.

andy-goryachev-oracle:
> the idea about mapping scroll events to the top/bottom of a cell is hinted in the code though

could you please point to these places?

I think this change is inline with the modern UIs, at least macOS and Windows.  As Eduard correctly pointed out, it makes sense to align on the cell top boundary when changing selection, but mouse navigation via scroll bar should use the view port height.  In other words, going back and forward with the mouse should revert the control to the same state (unless it hit the rail).

FlorianKirmaier:
I've added eduard as a contributor!

**CSR**
I've created a CSR.
https://bugs.openjdk.org/browse/JDK-8332041
If i remember correctly, the name of the ticket and the CSR must be the same?
Otherwise i could change it to something like "JavaFX, add the method double VirtualFlow.getViewportLength()"

kevinrushforth:
> I've added eduard as a contributor!
> 
> **CSR** I've created a CSR. https://bugs.openjdk.org/browse/JDK-8332041

Thanks!

> If i remember correctly, the name of the ticket and the CSR must be the same? Otherwise i could change it to something like "JavaFX, add the method double VirtualFlow.getViewportLength()"

Yes, the titles of the JBS bug, the associated CSR, and the PR must all match.

And speaking of the CSR, based on Johan's comment about the behavior change, that should also be documented in the CSR. We don't currently document the behavior of most controls in the spec, but significant behavioral changes should still be listed in the CSR when they change.

FlorianKirmaier:
Ok, I've updated the CSR accrodingly together with Eduard.

andy-goryachev-oracle:
> I'm not sure I understand the question

I thought there were specific comments somewhere in the code.  Never mind then.  Thank you for responding!

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1911,7 +1910,16 @@ void setViewportLength(double value) {
         this.absoluteOffset = getPosition() * (estimatedSize - viewportLength);
         recalculateEstimatedSize();
     }
-    double getViewportLength() {
+
+    /**
+     * The length of the viewport portion of the {@code VirtualFlow} as computed during the layout pass.
```
kevinrushforth, line 1915: Suggested change:

"Returns the length of the viewport ..."

kevinrushforth (commented):
The API changes look good. I left one minor wording suggesting on the API docs.

The new behavior feels more consistent to me, so I have no concerns. I'll leave it to @johanvos and @andy-goryachev-oracle to formally review it.

kevinrushforth:
@FlorianKirmaier or @eduardsdv can one of you merge in the latest upstream master? The source branch for this PR is a few months (and 134 commits) behind master.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
eduardsdv: Done

eduardsdv:
I have merged the master into this branch.

andy-goryachev-oracle (approved):

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1912,7 +1911,16 @@ void setViewportLength(double value) {
         this.absoluteOffset = getPosition() * (estimatedSize - viewportLength);
         recalculateEstimatedSize();
     }
-    double getViewportLength() {
+
+    /**
+     * Returns the length of the viewport portion of the {@code VirtualFlow} as computed during the layout pass.
+     * For a vertical flow, this corresponds to the height and for a horizontal flow to the width of the clip view,
+     * but it does not necessarily have to be the same value.
```
kevinrushforth, line 1918: @johanvos added a question in the CSR about this last  part:

> I understand and agree with the goal behind this.
> 
> I'm a bit confused though about the following: "...but it does not necessarily have to be the same value." -> can you elaborate a bit about this?

I share this concern. I think that removing that last clause and putting a period after "clip view" is probably the best.

Johan: what do you think?

johanvos: > @johanvos added a question in the CSR about this last part:
> 
> > I understand and agree with the goal behind this.
> > I'm a bit confused though about the following: "...but it does not necessarily have to be the same value." -> can you elaborate a bit about this?
> 
> I share this concern. I think that removing that last clause and putting a period after "clip view" is probably the best.
> 
> Johan: what do you think?

I agree that removing that clause is probably best to avoid confusion. Having open-ended suggestions in javadoc can lead to broad speculation, so I think it either should be explained (like you did in the CSR issue with the rubberband effect example) or removed.

eduardsdv: We should probably also avoid the word "corresponds".

````java
/**
 * Returns the length of the viewport portion of the {@code VirtualFlow} as computed during the layout pass.
 * For a vertical flow this is based on the height and for a horizontal flow on the width of the clip view.
 *
 * @return the viewport length in pixels
 * @since 23
 */
````
The text explains that depending on the orientation of the view height or width is used in the calculation and the word "based" makes it clear that the value can differ from the respective size of the view.

This version looks good to me. If it is fine for you too, I would check it in.

kevinrushforth: I'm happy with this wording.

kevinrushforth (approved):
The latest version looks good.

andy-goryachev-oracle (approved):

kevinrushforth:
@FlorianKirmaier This is ready for you to `/integrate`

PR #1321, opened 2024-01-05
8322703: Intermittent crash in WebView in a JFXPanel from IME calls on macOS

kevinrushforth:
As noted in the JBS bug, this is a follow-on to [JDK-8196011](https://bugs.openjdk.org/browse/JDK-8196011) that I discovered while testing the fix for [JDK-8221261](https://bugs.openjdk.org/browse/JDK-8221261) (a deadlock in the IME code when using WebView in a JFXPanel on macOS).

I have tested this in connection with with the proposed fix for JDK-8221261, although it is a valid fix regardless.

This expands the fix done in [JDK-8221261](https://bugs.openjdk.org/browse/JDK-8221261) to call all of the WebKit methods on the right thread. Additionally, we sometimes see spurious exceptions where the committed text is coming back as null, so I changed the log level to "fine" rather than "severe" for those exceptions. I'll file a follow-up bug  to see if any of these are real problems or not.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issue
 * [JDK-8322703](https://bugs.openjdk.org/browse/JDK-8322703): Intermittent crash in WebView in a JFXPanel from IME calls on macOS (**Bug** - P2)

### Reviewers
 * [Jay Bhaskar](https://openjdk.org/census#jbhaskar) (@jaybhaskar - Committer)
 * [Ambarish Rapte](https://openjdk.org/census#arapte) (@arapte - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1321/head:pull/1321` \
`$ git checkout pull/1321`

Update a local copy of the PR: \
`$ git checkout pull/1321` \
`$ git pull https://git.openjdk.org/jfx.git pull/1321/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 1321`

View PR using the GUI difftool: \
`$ git pr show -t 1321`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jfx/pull/1321.diff">https://git.openjdk.org/jfx/pull/1321.diff</a>

</details>

### Webrev
[...]

PR #1327, opened 2024-01-10
8221261: Deadlock on macOS in JFXPanel app when handling IME calls

kevinrushforth:
As described in the JBS bug, there is a long-standing deadlock that happens on macOS between the AWT EDT and the JavaFX Application thread (which on macOS is the AppKit thread) when processing Input Method Events (IME) in a WebView node in a JFXPanel.

This PR fixes the deadlock by adding `"AWTRunLoopMode"` to the list of modes that will accept the calls to `performSelectorOnMainThread` used by `_submitForLaterInvocation` and `_invokeAndWait`. These two native methods are the macOS implemention `runLater` and `runAndWait`, respectively, and are used to run Java Runnables on the JavaFX Application Thread.

This deadlock is happening much more often on recent macOS versions, and in macOS 14, pressing CAPS LOCK is sufficient to trigger IME calls.

The OS calls the AWT IME methods on the AppKit thread, which is also the JavaFX Application Thread. The AWT IME callback methods, for example `characterIndexForPoint`, call invokeAndWait to run the IME handler on the AWT Event thread. In the case of JFXPanel, the registered IME handler is in JavaFX code and we often need to run something on the JavaFX Application Thread to avoid threading problems or crashes. See [JDK-8196011](https://bugs.openjdk.org/browse/JDK-8196011) and [JDK-8322703](https://bugs.openjdk.org/browse/JDK-8322703).

A similar deadlock was observed while trying to fix a threading problem in JFXPanel's handling of  the IME calls as described in [JDK-8322784](https://bugs.openjdk.org/browse/JDK-8322784), so it isn't necessarily limited to using WebView.

Anton Tarasov @forantar proposed a fix (in a comment in the JBS bug) in AWT that would change the invokeAndWait calls used by the IME methods in CInputMethod to call `doAWTRunLoop` with the `processEvents` flag set to true, which will allow all events to be processed while in `doAWTRunLoop`.

NOTE: I had created Draft PR openjdk/jdk#17290 with Anton's proposed fix, expanded to cover all of the cases, but during the discussion it was pointed out that `doAWTRunLoop` should already be able to run selectors via `performSelectorOnMainThread` -- see [this comment](https://github.com/openjdk/jdk/pull/17290#issuecomment-1880001813). The reason `doAWTRunLoop` doesn't process our messages is that it runs the loop in a mode that only handles messages that include a custom `"AWTRunLoopMode"` mode.    
          
Adding `"AWTRunLoopMode"` to the `performSelectorOnMainThread` calls used by `_submitForLaterInvocation` and `_invokeAndWait` allows them to run when in `doAWTRunLoop` so there is no deadlock.

This is a better fix than the proposed fix in PR openjdk/jdk#17290 for two reasons. First, it is a more targeted fix. For example, it doesn't open the door for processing mouse events while in `doAWTRunLoop`. Second, by fixing it in JavaFX, we can fix additional threading bugs like [JDK-8322703](https://bugs.openjdk.org/browse/JDK-8322703) and [JDK-8322784](https://bugs.openjdk.org/browse/JDK-8322784) without worrying about introducing deadlocks if run on a JDK without the fix.

---------
### Progress

### Issue
 * [JDK-8221261](https://bugs.openjdk.org/browse/JDK-8221261): Deadlock on macOS in JFXPanel app when handling IME calls (**Bug** - P2)

### Reviewers
 * [Prasanta Sadhukhan](https://openjdk.org/census#psadhukhan) (@prsadhuk - **Reviewer**)
 * [Sergey Bylokhov](https://openjdk.org/census#serb) (@mrserb - no project role)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1327/head:pull/1327` \
`$ git checkout pull/1327`

`$ git checkout pull/1327` \
`$ git pull https://git.openjdk.org/jfx.git pull/1327/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1327`

`$ git pr show -t 1327`

</details>

<a href="https://git.openjdk.org/jfx/pull/1327.diff">https://git.openjdk.org/jfx/pull/1327.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1327#issuecomment-1885308253)

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
@@ -61,6 +61,21 @@ static BOOL disableSyncRendering = NO;
 static BOOL firstActivation = YES;
 static BOOL shouldReactivate = NO;
 
+// Custom NSRunLoopMode constant that matches the one used by AWT in its
+// doAWTRunLoopImpl method. This is not formally documented, but there is
+// an open bug to do so. See https://bugs.openjdk.org/browse/JDK-8270211
+static NSString* JavaRunLoopMode = @"AWTRunLoopMode";
+
+// List of allowable runLoop modes that Java runnables can run in.
+// This is used when calling performSelectorOnMainThread from
+// the JNI _invokeAndWait and _submitForLaterInvocation methods.
+// We include JavaRunLoopMode in the list of allowable run modes in case
+// we are running on the AWT event thread. This will allow JavaFX
+// runnables to be scheduled when AWT is running doAWTRunLoopImpl
+// on the AppKit thread (which it does for IME callbacks) so that we
+// don't deadlock.
+static NSArray<NSString*> *runLoopModes = nil;
+
 #ifdef STATIC_BUILD
 jint JNICALL JNI_OnLoad_glass(JavaVM *vm, void *reserved)
 #else
@@ -568,6 +583,19 @@ jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved)
                 GLASS_CHECK_EXCEPTION(jEnv);
             }
         }
+
+        // List of RunLoopModes in which we will run runnables passed
+        // to _invokeAndWait and _invokeAndExit. This includes
+        // JavaRunLoopMode to avoid a possible deadlock with AWT.
+        // There is no harm always adding it, since it will have no
+        // effect if the run loop that receives this message does not
+        // specify it.
+        runLoopModes = [[NSArray alloc] initWithObjects:
+            NSDefaultRunLoopMode,
+            NSModalPanelRunLoopMode,
+            NSEventTrackingRunLoopMode,
+            JavaRunLoopMode,
+            nil];
 
         // Determine if we're running embedded (in AWT, SWT, elsewhere)
         NSApplication *app = [NSApplicationFX sharedApplication];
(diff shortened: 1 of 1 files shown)


kevinrushforth:
Reviewers: @prrace @jaybhaskar 

/reviewers 2

kevinrushforth:
NOTE to reviewers: If you want to test this with WebView in a JFXPanel, you will also need the fix for [JDK-8322703](https://bugs.openjdk.org/browse/JDK-8322703), which is under review in PR #1321 to avoid a crash (that crash was effectively hidden by this deadlock so it was missed earlier).

I have created a test branch that has both fixes:

https://github.com/kevinrushforth/jfx/tree/test-8221261-8322703

You can either fetch that branch or fetch both PR branches and build. It is not necessary to build the native WebKit if you have an up-to-date library in your cache.

prrace:
I think there's a bit of additional explanation joining the dots that would be helpful here

> The AWT IME callback methods, for example characterIndexForPoint, call invokeAndWait to run the IME handler on the AWT Event thread.

And that means specifically call AWT's LWCToolkit.invokeAndWait() and THAT uses doAWTRunLoop(), which when entered uses the flag 'inAWT'
(which will be true) to invoke NSRunLoop (in a loop) in a way that it should ONLY process requests that specify the special mode

[[NSRunLoop currentRunLoop] runMode:(inAWT ? [ThreadUtilities javaRunLoopMode] : NSDefaultRunLoopMode)

It stays in this loop until it is informed that the thing it was waiting for is done.

This won't happen unless FX specifies that mode so its call is eligible to be processed.

prrace (approved):

prsadhuk (approved):
Looks good to me..

mrserb (approved):

Eng-Fouad:
Hi @kevinrushforth, we are trying to run JavaFX on GraalVM native-image on MacOS. When we start the app, `com.sun.glass.ui.mac.MacApplication#runLoop` gets stuck on the native side. We tried everything and couldn't figure out how to solve it. We would really appreciate you help to find the real cause and fix this issue. Thanks.

https://github.com/quarkiverse/quarkus-fx/issues/205

PR #557, opened 2021-06-30
8089398: [ChoiceBox, ComboBox] throws NPE on setting value on null selectionModel

Maran23:
This PR fixes multiple NPEs in Choice-and ComboBox, when the selection model is null.

ChoiceBox: 
- Null check in **valueProperty()** listener

ComboBox:
- Null check in **editableProperty()** listener
- Null check in **valueProperty()** listener
- Null check in **ComboBoxListViewSkin#updateValue()**
- Null check in **ComboBoxListViewSkin#layoutChildren()**
- 2x Null check in **ComboBoxListViewSkin#createListView()**

~~The tests checks, that no NPE is printed to the console.
Some checks, that the set value is still displayed (either in the ComboBox button cell or the ChoiceBox display label)~~

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed

### Issue
 * [JDK-8089398](https://bugs.openjdk.java.net/browse/JDK-8089398): [ChoiceBox, ComboBox] throws NPE on setting value on null selectionModel

### Reviewers
 * [Jeanette Winzenburg](https://openjdk.java.net/census#fastegal) (@kleopatra - **Reviewer**)
 * [Ajit Ghaisas](https://openjdk.java.net/census#aghaisas) (@aghaisas - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jfx pull/557/head:pull/557` \
`$ git checkout pull/557`

Update a local copy of the PR: \
`$ git checkout pull/557` \
`$ git pull https://git.openjdk.java.net/jfx pull/557/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 557`

View PR using the GUI difftool: \
`$ git pr show -t 557`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.java.net/jfx/pull/557.diff">https://git.openjdk.java.net/jfx/pull/557.diff</a>

</details>

PR #1329, opened 2024-01-10
JDK-8323543: NPE when table items are set to null

EstelonAgarwaen:
This PR fixes a nullpointer in TableSkinUtils that occured when the Tables items were null.

---------
### Progress

### Issue
 * [JDK-8323543](https://bugs.openjdk.org/browse/JDK-8323543): NPE when table items are set to null (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1329/head:pull/1329` \
`$ git checkout pull/1329`

`$ git checkout pull/1329` \
`$ git pull https://git.openjdk.org/jfx.git pull/1329/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1329`

`$ git pr show -t 1329`

</details>

<a href="https://git.openjdk.org/jfx/pull/1329.diff">https://git.openjdk.org/jfx/pull/1329.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1329#issuecomment-1885444337)

### DIFF
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewTest.java
@@ -25,6 +25,7 @@
 
 package test.javafx.scene.control;
 
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static test.com.sun.javafx.scene.control.infrastructure.ControlTestUtils.assertStyleClassContains;
 import static javafx.scene.control.TableColumn.SortType.ASCENDING;
 import static javafx.scene.control.TableColumn.SortType.DESCENDING;
@@ -6063,5 +6064,20 @@ public class TableViewTest {
 
         table.getSelectionModel().selectIndices(1, new int[]{1, 2});
         assertEquals(2, table.getSelectionModel().getSelectedIndex());
+    }
+
+    @Test
+    public void testTableItemsNullShouldNotThrow() {
+        final TableColumn<String, String> c = new TableColumn<>("C");
+        c.setCellValueFactory(value -> new SimpleStringProperty(value.getValue()));
+        table.getColumns().add(c);
+
+        table.getItems().addAll("1", "2", "3");
+
+        stageLoader = new StageLoader(table);
+        table.setItems(null);
+
+        assertDoesNotThrow(() -> Toolkit.getToolkit().firePulse());
+
     }
 }
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewTest.java
@@ -6082,9 +6082,6 @@ public class TableViewTest {
 
     @Test
     public void testTableItemsNullQueryAcceessibleAttributeRowCountShouldNotThrow() {
-        table.getItems().addAll("1", "2", "3");
-
-        stageLoader = new StageLoader(table);
         table.setItems(null);
 
         assertDoesNotThrow(() -> table.queryAccessibleAttribute(AccessibleAttribute.ROW_COUNT));
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewTest.java
@@ -25,7 +25,6 @@
 
 package test.javafx.scene.control;
 
-import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static test.com.sun.javafx.scene.control.infrastructure.ControlTestUtils.assertStyleClassContains;
 import static javafx.scene.control.TableColumn.SortType.ASCENDING;
 import static javafx.scene.control.TableColumn.SortType.DESCENDING;
@@ -6076,14 +6075,14 @@ public class TableViewTest {
 
         stageLoader = new StageLoader(table);
         table.setItems(null);
-
-        assertDoesNotThrow(() -> Toolkit.getToolkit().firePulse());
+        // Should not throw an NPE.
+        Toolkit.getToolkit().firePulse();
     }
 
     @Test
     public void testTableItemsNullQueryAcceessibleAttributeRowCountShouldNotThrow() {
         table.setItems(null);
-
-        assertDoesNotThrow(() -> table.queryAccessibleAttribute(AccessibleAttribute.ROW_COUNT));
+        // Should not throw an NPE.
+        table.queryAccessibleAttribute(AccessibleAttribute.ROW_COUNT);
     }
 }
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewTest.java
@@ -6078,6 +6078,15 @@ public class TableViewTest {
         table.setItems(null);
 
         assertDoesNotThrow(() -> Toolkit.getToolkit().firePulse());
+    }
 
+    @Test
+    public void testTableItemsNullQueryAcceessibleAttributeRowCountShouldNotThrow() {
+        table.getItems().addAll("1", "2", "3");
+
+        stageLoader = new StageLoader(table);
+        table.setItems(null);
+
+        assertDoesNotThrow(() -> table.queryAccessibleAttribute(AccessibleAttribute.ROW_COUNT));
     }
 }
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeTableViewTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeTableViewTest.java
@@ -7269,6 +7269,5 @@ public class TreeTableViewTest {
         treeTableView.setRoot(null);
 
         assertDoesNotThrow(() -> Toolkit.getToolkit().firePulse());
-
     }
 }
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeTableViewTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeTableViewTest.java
@@ -36,7 +36,6 @@ import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
-import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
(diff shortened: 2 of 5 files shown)


modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeTableViewTest.java
```diff
@@ -7250,4 +7253,26 @@ public void testFirstRowSelectionWithEmptyArrayAsParameter() {
...
+        for (int i = 0; i < 4; i++) {
+            TreeItem<String> parent = new TreeItem<String>("item - " + i);
+            treeTableView.getRoot().getChildren().add(parent);
+        }
+
+        stageLoader = new StageLoader(treeTableView);
+        treeTableView.setRoot(null);
+
+        assertDoesNotThrow(() -> Toolkit.getToolkit().firePulse());
+
```
andy-goryachev-oracle, line 7272: please remove extra newline

andy-goryachev-oracle:
FYI: there is also a possibility of an NPE in 

TableCellBehavior:70
TableView:2198
TableView:1922

I think we should add null checks there as well.

EstelonAgarwaen:
should i add the null checks in this PR?

andy-goryachev-oracle:
I think it makes sense to add the checks in this PR.  We are addressing the issue and not introducing new code paths.
I wonder if we should also add the unit tests to cover the newly added checks.  What do you think?

modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewTest.java
```diff
@@ -6064,4 +6065,19 @@ public void testFirstRowSelectionWithEmptyArrayAsParameter() {
...
+        c.setCellValueFactory(value -> new SimpleStringProperty(value.getValue()));
+        table.getColumns().add(c);
+
+        table.getItems().addAll("1", "2", "3");
+
+        stageLoader = new StageLoader(table);
+        table.setItems(null);
+
+        assertDoesNotThrow(() -> Toolkit.getToolkit().firePulse());
+
```
andy-goryachev-oracle, line 6081: extra newline?

EstelonAgarwaen:
i will look into whether it is possible to generate a nullpointer with the current code, and add a test to check that they are not thrown anymore.

andy-goryachev-oracle:
Thank you @EstelonAgarwaen and @Maran23

EstelonAgarwaen:
Regarding the NPEs: 
- i was able to trigger an exception in TableView:1922
- the code in TableCellBehaviour:70 is only reachable by user interaction with a cell. A cell only exists when a TableView has items. The interaction will never happen while `items` is null, because the code is unreachable in that case

andy-goryachev-oracle:
I might suggest to add a null check to TableCellBehaviour:70 regardless, just to be on the safe side.

EstelonAgarwaen:
since the conditions needed for TableCellBehaiviour:70 to be execute require items to not be null for that code to even be triggered, i think it does not make sense to check whether the items are null.

andy-goryachev-oracle (approved):
Looks good, thank you for finding and fixing the issue!

This PR has a very narrow focus to be a candidate for a back port to jfx22 @kevinrushforth

andy-goryachev-oracle (approved):
FYI, pushing another commit requires re-review (easy in this particular case).

EstelonAgarwaen:
dont worry, im in a call with @Maran23 rn

andy-goryachev-oracle:
@Maran23 would you like to double check and sponsor?

Maran23:
> @Maran23 would you like to double check and sponsor?

sure!

kevinrushforth:
> This PR has a very narrow focus to be a candidate for a back port to jfx22 @kevinrushforth

This seems a safe and targeted fix. I have no objection.

kevinrushforth:
Although... I just spotted something. The updated tests use JUnit5 assertions into what is otherwise a JUnit4 test. We generally avoid doing this.

Maran23:
> Although... I just spotted something. The updated tests use JUnit5 assertions into what is otherwise a JUnit4 test. We generally avoid doing this.

Should we use the JUnit5 `@Test` annotation, or migrate the class to JUnit5? The migration is also fine, although some assert need to be switched from `assertEquals(text, expected, actual)` -> `assertEquals(expected, actual, text)`

modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewTest.java
```diff
@@ -3871,22 +3929,22 @@ private void test_rt_37538(boolean callCNextOnce, boolean callCNextInLoop) {
         sl.dispose();
     }
 
-    @Ignore("Fix not yet developed for TableView")
+    @Disabled("Fix not yet developed for TableView")
```
andy-goryachev-oracle, line 3932: `@Test` was placed on its own line elsewhere but not in this file?

modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewTest.java
```diff
@@ -5334,8 +5411,10 @@ public void test_rt_40012_selectedItemNotificationOnDisjointRemovesAbove() {
 
     @Test public void test_rt_40319_toRight_toBottom()          { test_rt_40319(true, true, false);   }
     @Test public void test_rt_40319_toRight_toTop()             { test_rt_40319(true, false, false);  }
-    @Test public void test_rt_40319_toLeft_toBottom()           { test_rt_40319(false, true, false);  }
-    @Test public void test_rt_40319_toLeft_toTop()              { test_rt_40319(false, false, false); }
+    @Test
```
andy-goryachev-oracle, line 5414: inconsistent formatting?

modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewTest.java
```diff
@@ -5695,7 +5780,8 @@ public void test_clearAndSelectChangeMultipleSelectionCellMode() {
     @Test
     public void test_ChangeToStringKeyboardMultipleSelectionCellMode() {
         final Thread.UncaughtExceptionHandler exceptionHandler = Thread.currentThread().getUncaughtExceptionHandler();
-        Thread.currentThread().setUncaughtExceptionHandler((t, e) -> fail("We don't expect any exceptions in this test!"));
+        Thread.currentThread().setUncaughtExceptionHandler((t, e) -> fail(
+                "We don't expect any exceptions in this test!"));
```
andy-goryachev-oracle, line 5784: unrelated change?

modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewTest.java
Maran23: Hmm, I suggested @EstelonAgarwaen to use the IntelliJ Junit 5 Migration functionality. So looks like it is not always putting `@Test` in a new line. Only when nothing is above it seems?

andy-goryachev-oracle:
I wonder if it was a mistake to migrate to junit5 in this PR, it should have been a separate task: the scope of the change has exploded, it might cause issues with backporting, etc.

Would it be possible to separate the migration from the fix?

Maran23:
> Would it be possible to separate the migration from the fix?

Another idea could be to use the full qualified path: `@org.junit.jupiter.api.Test`. But `@Before` and `@After` are still JUnit 4.
Or we can reduce the change by at least reverting the new line changes? Then the diff should be much smaller.

andy-goryachev-oracle:
I think the problem is that we are trying to address two issues in one PR - the fix and migration to junit5.
It might be better to go back one commit (but keep the most recent changes on a separate branch), convert the fix to use junit4 and submit the fix in this PR.

Then, create a separate JBS ticket for junit5 migration.  This way the scope of the fix is reduced to a minimum, and it can be easily backported and not cause many issues.

Maran23:
> I think the problem is that we are trying to address two issues in one PR - the fix and migration to junit5. It might be better to go back one commit (but keep the most recent changes on a separate branch), convert the fix to use junit4 and submit the fix in this PR.

Sure, just FYI that there is nothing like `assertDoesNotThrow` for JUnit4.
What I did in the past is to at least write a comment, so we can easily add/convert that later on.
See also my PR in the past:
https://github.com/openjdk/jfx/pull/557
More specific:
https://github.com/openjdk/jfx/pull/557/files#diff-ea6f98c080f3abde83c8de64cd1b026b727bd5998cf6e0e57dde857f0e3e541eR353
cc @EstelonAgarwaen, I would recommend doing it the same way I did above and stash the JUnit 5 migration, as @andy-goryachev-oracle mentioned.

Maran23:
> looks good now, thanks! should the fix be backported to jfx22?

Sounds good. I can do that, just need to find the instructions again.

andy-goryachev-oracle:
backport instructions:
https://openjdk.org/guide/#working-with-backports-in-jbs

@kevinrushforth there is still time to backport, right?

kevinrushforth:
> backport instructions: https://openjdk.org/guide/#working-with-backports-in-jbs

See also https://mail.openjdk.org/pipermail/openjfx-dev/2024-January/044529.html which has additional information.

> @kevinrushforth there is still time to backport, right?

Yes. JavaFX 22 is in RDP1 for another 3 weeks.

kevinrushforth:
> /backport jfx22u

@Maran23 That will target jfx22u for JavaFX 22.0.1, which would probably be OK, but as I mentioned above, since we are still in RDP1 for JavaFX 22, this is a good candidate for JavaFX 22. So I recommend abandoning the above PR and instead target the jfx22 branch of the jfx repo: `/backport jfx jfx22` instead.

Maran23:
> So I recommend abandoning the above PR and instead target the jfx22 branch of the jfx repo: `/backport jfx jfx22` instead.

Ahh, makes sense, thanks. First time backporting. :)
I deleted the backport branch in jfx22u.

kevinrushforth:
> Ahh, makes sense, thanks. First time backporting. :)

Yeah, and it's somewhat confusing to have jfx22u active during RDP1 of JavaFX 22 where most fixes that go back into the 22 code line will want to go into 22 GA rather than 22.0.1. I had considered delaying the announcement of jfx22u being open, but we need it to backport the WebKit update (which we explicitly do not want in JavaFX 22 GA, but in 22.0.1).

Maran23:
> Yeah, and it's somewhat confusing to have jfx22u ..., but we need it to backport the WebKit update (which we explicitly do not want in JavaFX 22 GA, but in 22.0.1).

Thanks for the clarification!
I totally forgot/missed that we have a branch for every GA release in this repository, while every backport after is done in the jfx..u repository.

PR #1330, opened 2024-01-10
8218745: TableView: visual glitch at borders on horizontal scrolling

Maran23:
This PR fixes the border glitch/gap as explained in both linked tickets.

I noted that the `ScrollPane` control does not suffer from this problem, although the code is mostly the same as in `VirtualFlow`. The `ScrollPane` snaps the content correctly, no matter which scale. I carefully checked the code and it seems that the container structure in `ScrollPane` was explicitly written to handle this perfectly. There was definitely some thought on that.

So to finally fix this issue, I rewrote the `VirtualFlow` container/scrolling code to be **exactly** the same as in `ScrollPane`(Skin).
And this also fixes the issue, while behaving the same as before.

In the future it may makes sense to try to somewhat unify the code from `ScrollPane` and `VirtualFlow`. I already have some ideas.

Unfortunately though, one more container is introduced to `VirtualFlow`, so the css needs to be changed since it is very strictly written in the first place:
Before: `.list-view:focused > .virtual-flow > .clipped-container > .sheet > .list-cell`
After: `.list-view:focused > .virtual-flow > .viewport > .clipped-container > .sheet > .list-cell`

To better understand this, the container structure (tree) is shown below:

- ScrollPane
  - viewRect ->  `setClip` -> clipRect (viewContent size)
    - viewContent -> `setLayoutX`
      - Content
  - vsb
  - hsb
  - corner

---
- VirtualFlow
  - viewRect **(->NEW IN THIS PR<-)** -> `setClip` -> clipRect (clippedContainer size)
    - clippedContainer/clipView -> `setLayoutX` (when scrolling)
      - sheet
        - Cell
  - vsb
  - hsb
  - corner

/issue add JDK-8146406

---------
### Progress
- [ ] Change requires a CSR request matching fixVersion jfx24 to be approved (needs to be created)

### Issues
 * [JDK-8218745](https://bugs.openjdk.org/browse/JDK-8218745): TableView: visual glitch at borders on horizontal scrolling (**Bug** - P4)
 * [JDK-8146406](https://bugs.openjdk.org/browse/JDK-8146406): Unexplainable gap between TableCell inside a TableView (**Bug** - P4)

### Reviewers
 * [Andy Goryachev](https://openjdk.org/census#angorya) (@andy-goryachev-oracle - **Reviewer**) 🔄 Re-review required (review applies to [c3594907](https://git.openjdk.org/jfx/pull/1330/files/c3594907b8cdb72f8610512ab6388351fa00b3a4))

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1330/head:pull/1330` \
`$ git checkout pull/1330`

`$ git checkout pull/1330` \
`$ git pull https://git.openjdk.org/jfx.git pull/1330/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1330`

`$ git pr show -t 1330`

</details>

<a href="https://git.openjdk.org/jfx/pull/1330.diff">https://git.openjdk.org/jfx/pull/1330.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1330#issuecomment-1885563403)
</details>


kevinrushforth:
This will need careful review and testing.

/reviewers 2

kevinrushforth:
@johanvos : I presume you will want to review this?

modules/javafx.controls/src/main/resources/com/sun/javafx/scene/control/skin/caspian/caspian.css
```diff
@@ -1505,32 +1505,32 @@
     -fx-background-color: derive(-fx-control-inner-background,-5%);
 }
 
-.list-view:focused > .virtual-flow > .clipped-container > .sheet > .list-cell:focused {
+.list-view:focused > .virtual-flow > .viewport > .clipped-container > .sheet > .list-cell:focused {
```
andy-goryachev-oracle, line 1508: This might require a CSR as it might break custom stylesheets that modify list/tree/table/views.

andy-goryachev-oracle:
The scrolling works correctly with the fix (tested TableView on macOS (scale=1 and 2) and win11 (scale=200%, 225%).  I'll continue testing tomorrow.

It might be a dumb question: would it be possible to avoid creating an intermediate container and keep the existing .css files?

Maran23:
> It might be a dumb question: would it be possible to avoid creating an intermediate container and keep the existing .css files?

Not a dumb question, that is what I tried at first. See also my old PR: https://github.com/openjdk/jfx/pull/630. I tried multiple things with the container and snapping, but never managed to fix all the issues with all scales.
So I didn't found a way to fix this issue other than having the same structure as `ScrollPane`, which works very well with whatever you put into it.

modules/javafx.controls/src/main/resources/com/sun/javafx/scene/control/skin/caspian/caspian.css
Maran23: Probably. Was not sure since this is a very specific case which we never had before (as far as I can tell).

andy-goryachev-oracle (approved):
looks good, tested list/tree/table views on macOS 14.1.2 in light mode

modules/javafx.controls/src/main/resources/com/sun/javafx/scene/control/skin/caspian/caspian.css
kevinrushforth: Yes, I think a CSR would be in order. Thanks for bringing this up @andy-goryachev-oracle

Maran23:
Note that I found out something interesting which can MAY help us here to solve the problem in another way.

What I found out:
If I set the the opacity of the clip rect to something under 1, this bug disappears.
image

So something in the low level clip rendering is different when an opacity is set. Will investigate if there could be another solution.

Maran23:
Ok, so I found out the following: 
When a `Rectangle` is used as clip without any effect or opacity modification, the rendering goes another (probably faster) route with rendering the clip. That's why setting the opacity to `0.99` fixes the issue - another route will be used for the rendering.
This happens at the low level (`NGNode`) side of JavaFX. 

See here `NGNode`:
image
`renderRectClip` seems to be doing something wrong.

I see that the clip rect is floored there, which seems to be the cause. Rounding it is fixing this for me.
The bug always appears when I scroll and the clip `RectBounds` are something like:
`RectBounds { minX:6.999996, minY:37.0, maxX:289.00003, maxY:194.0} (w:282.00003, h:157.0)`

while this does not happen when the value is something like:
`RectBounds { minX:7.000004, minY:37.0, maxX:289.0, maxY:194.0} (w:282.0, h:157.0)`

Note that the `minX` is floored down, that explains the 1px gap.

And now I fully understand why this only happens for the `VirtualFlow`:
This is the only container and location where the clip is layouted as well! (`setLayoutX`/`setLayoutY`).
In `ScrollPaneSkin`, there was the decision, that the clip is set and the content is layouted only, not the clip.

It may still make sense to me to actually imrpove the `VirtualFlow` here - it seems like a good choice to just set the clip and never touch it again and only change the underlying content layout, might be even a bit faster.

But there is obviously a bug on the low level `NGNode` side as well. Not sure where to go from here.

Maran23:
open for discussion still, but there will be a counter PR from me at one point.

andy-goryachev-oracle:
> But there is obviously a bug on the low level NGNode side as well. Not sure where to go from here.

does it mean this PR is not ready, or we have other issues?

Also, a CSR is needed for this PR to go forward.

andy-goryachev-oracle:
should this PR be closed in favor of #1462 ?

Maran23:
Moving to draft.

Maran23:
Closing as https://github.com/openjdk/jfx/pull/1462 turned out to be the better choice (and is merged now).

PR #806, opened 2022-06-29
8276056: Control.skin.setSkin(Skin) fails to call dispose() on discarded Skin

Maran23:
For some reason the `skinProperty` did not allow to set a new skin which is the same class as the previous one.
This leads to multiple issues:
1. When creating a new skin (same class as previous), the skin will likely install some children and listener but is then rejected when setting it due to the `skinProperty` class constraint
-> Control is in a weird state as the current skin was not disposed since it is still set, although we already and 'set' a new skin
2. A skin of the same class can behave differently, so it is not really valid to reject a skin just because it is the same class as the previous
-> Just imagine we have the following skin class
```
class MyCustomButtonSkin extends ButtonSkin {
    public MyCustomButtonSkin(Button button, boolean someFlag) { super(button); ... }
}
```
Now if we would change the skin of the Button two times like this:
```
button.setSkin(new MyCustomButtonSkin(button, true));
button.setSkin(new MyCustomButtonSkin(button, false));
```
The second time the skin will be rejected as it is the same class, although I may changed the skin behaviour, in this case demonstrated by a boolean flag for showing purposes.

PR #1331, opened 2024-01-11
JDK-8323615: PopupControl.skin.setSkin(Skin) fails to call dispose() on discarded Skin

Maran23:
For some reason the `skinProperty` did not allow to set a new skin when it is the same class as the previous one.
This leads to multiple issues:
1. When creating a new skin (same class as previous), the skin will likely install listener but is then rejected when setting it due to the `skinProperty` class constraint
-> `PopupControl` is in a weird state as the current skin was not disposed since it is still set, although we already created and 'set' a new skin
2. A skin of the same class can behave differently, so it is not really valid to reject a skin just because it is the same class as the previous
-> Just imagine we have the following skin class
```
class MyCustomSkin<C extends PopupControl> extends Skin<C> {
    public MyCustomSkin(C skinnable, boolean someFlag) { ... }
}
```
Now if we would change the skin of the `PopupControl` two times like this:
```
popup.setSkin(new MyCustomSkin(popup, true));
popup.setSkin(new MyCustomSkin(popup, false));
```
The second time the skin will be rejected as it is the same class, although I may changed the skin behaviour, in this case demonstrated by a boolean flag for showing purposes.

This is the same issue and fix as done for `Control` in [JDK-8276056](https://bugs.openjdk.org/browse/JDK-8276056) (PR: https://github.com/openjdk/jfx/pull/806)

---------
### Progress

### Issue
 * [JDK-8323615](https://bugs.openjdk.org/browse/JDK-8323615): PopupControl.skin.setSkin(Skin) fails to call dispose() on discarded Skin (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1331/head:pull/1331` \
`$ git checkout pull/1331`

`$ git checkout pull/1331` \
`$ git pull https://git.openjdk.org/jfx.git pull/1331/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1331`

`$ git pr show -t 1331`

</details>

<a href="https://git.openjdk.org/jfx/pull/1331.diff">https://git.openjdk.org/jfx/pull/1331.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1331#issuecomment-1887911803)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/PopupControl.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/PopupControl.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -212,18 +212,6 @@ public class PopupControl extends PopupWindow implements Skinnable, Styleable {
         // only needed because invalidated() does not currently take
         // a reference to the old value.
         private Skin<?> oldValue;
-
-        @Override
-        public void set(Skin<?> v) {
-
-            if (v == null
-                    ? oldValue == null
-                    : oldValue != null && v.getClass().equals(oldValue.getClass()))
-                return;
-
-            super.set(v);
-
-        }
 
         @Override protected void invalidated() {
             Skin<?> skin = get();
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/PopupControlTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/PopupControlTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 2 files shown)


modules/javafx.controls/src/main/java/javafx/scene/control/PopupControl.java
```diff
@@ -213,18 +213,6 @@ public PopupControl() {
         // a reference to the old value.
         private Skin<?> oldValue;
 
-        @Override
```
andy-goryachev-oracle, line 216: This logic was introduce to deal with skins set from the css JDK-8096194

Are you sure it will still work?  Do we need to test this scenario explicitly?

Maran23: No, the `skin`/`skinProperty` logic was there before, just moved.
It was added in https://bugs.openjdk.org/browse/JDK-8127070, 
with the comment from `David Grieve`: `The code for handling the skin property in PopupControl needs to look like the code in Control.`

So they basically synchronized the whole skin/skin name API to `PopupControl`.

modules/javafx.controls/src/test/java/test/javafx/scene/control/PopupControlTest.java
```diff
@@ -650,4 +653,29 @@ private static final class PopupControlSkin<C extends PopupControl> implements S
...
+    @Test
+    public void testSetSkinOfSameClass() {
+        popup.setSkin(new PopupControlSkin<>());
+        Skin<?> oldSkin = popup.getSkin();
+        popup.setSkin(new PopupControlSkin<>());
+        Skin<?> newSkin = popup.getSkin();
+
+        assertNotEquals("New skin was not set", oldSkin, newSkin);
+    }
+
```
andy-goryachev-oracle, line 680: please remove extra newline

andy-goryachev-oracle:
I think the proposed fix is correct.  I've tried to re-load skin via CSS (see https://github.com/andy-goryachev-oracle/Test/blob/main/src/goryachev/bugs/PopupControl_SetSkin_8323615.java ) and it does seem to reload correctly, as far as I can tell.

Loading fails without the fix, as expected.

Still, i would like a second pair of eyes to look at this, if possible.

andy-goryachev-oracle (approved):

kevinrushforth:

mstr2 (approved):
Looks good, and works as I expect it.

arapte:
@Maran23 
This is ready for integration. If you were waiting for me. Thanks, It looks good, I don't have any comments.

Maran23:
> @Maran23 This is ready for integration. If you were waiting for me. Thanks, It looks good, I don't have any comments.

Thank you for checking as well!

/integrate

PR #1332, opened 2024-01-12
8260013: Snapshot does not work for nodes in a subscene

lukostyra:
Originally this issue showed the problem of Node being incorrectly rendered (clipped) when snapshotting, compared to a snapshot of the whole Scene. Later on there was another problem added - lights not being taken into account if they are added to a SubScene.

As it later turned out, the original problem from this bug report is a problem with ParallelCamera incorrectly estimating near/far clipping planes, which just happened to reveal itself while snapshotting a Node. During testing I found out you can make the Node clip regardless of snapshot mechanism. Clipping issue was moved to a separate bug report and this PR only fixes the inconsistency in lights being gathered for a snapshot.

`Scene.doSnapshot()` was expanded to also check if SubScene provided to it is non-null and to fetch lights assigned to it. Scenario was tested with added SnapshotLightsTest.

Rest of the tests were checked and don't produce any noticeable regressions.

---------
### Progress

### Issue
 * [JDK-8260013](https://bugs.openjdk.org/browse/JDK-8260013): Snapshot does not work for nodes in a subscene (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1332/head:pull/1332` \
`$ git checkout pull/1332`

`$ git checkout pull/1332` \
`$ git pull https://git.openjdk.org/jfx.git pull/1332/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1332`

`$ git pr show -t 1332`

</details>

<a href="https://git.openjdk.org/jfx/pull/1332.diff">https://git.openjdk.org/jfx/pull/1332.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1332#issuecomment-1889323348)

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
@@ -1298,9 +1298,22 @@ public class Scene implements EventTarget {
 
     }
 
+    /*
+     * Shared between Scene and SubScene
+     */
+    static NGLightBase[] accumulateLightsForSnapshot(List<LightBase> lights) {
+        if (lights == null || lights.isEmpty()) return null;
+
+        NGLightBase[] l = new NGLightBase[lights.size()];
+        for (int i = 0; i < lights.size(); i++) {
+            l[i] = lights.get(i).getPeer();
+        }
+        return l;
+    }
+
     // Shared method for Scene.snapshot and Node.snapshot. It is static because
     // we might be doing a Node snapshot with a null scene
-    static WritableImage doSnapshot(Scene scene,
+    static WritableImage doSnapshot(Scene scene, SubScene subScene,
             double x, double y, double w, double h,
             Node root, BaseTransform transform, boolean depthBuffer,
             Paint fill, Camera camera, WritableImage wimg) {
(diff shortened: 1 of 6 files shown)


modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
```diff
@@ -1346,12 +1359,33 @@ static WritableImage doSnapshot(Scene scene,
...
         }
 
-        // Grab the lights from the scene
+        // Grab the lights from the scene and/or subscene
         context.lights = null;
-        if (scene != null && !scene.lights.isEmpty()) {
-            context.lights = new NGLightBase[scene.lights.size()];
-            for (int i = 0; i < scene.lights.size(); i++) {
-                context.lights[i] = scene.lights.get(i).getPeer();
+        int totalLightCount = 0;
```
mstr2, line 1364: Here's a suggestion: you might be able to replace all of the new code, including the `accumulateLightsForSnapshot` method, with the following shorter code:

```java
Stream<NGLightBase> lights = Stream.concat(
    Optional.ofNullable(scene).stream().flatMap(s -> s.lights.stream()).map(LightBase::getPeer),
    Optional.ofNullable(subScene).stream().flatMap(s -> s.getLights().stream()).map(LightBase::getPeer));

context.lights = lights.toArray(NGLightBase[]::new);
```

modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
```diff
@@ -1346,14 +1347,13 @@ static WritableImage doSnapshot(Scene scene,
...
-        if (scene != null && !scene.lights.isEmpty()) {
-            context.lights = new NGLightBase[scene.lights.size()];
-            for (int i = 0; i < scene.lights.size(); i++) {
-                context.lights[i] = scene.lights.get(i).getPeer();
-            }
-        }
+        // Grab the lights from the scene and/or subscene
+        Stream<NGLightBase> lights = Stream.concat(
+            Optional.ofNullable(scene).stream().flatMap(s -> s.lights.stream()).map(LightBase::getPeer),
+            Optional.ofNullable(subScene).stream().flatMap(s -> s.getLights().stream()).map(LightBase::getPeer)
```
kevinrushforth, line 1353: For a node in a SubScene this will apply the lights from both the Scene and SubScene, which is not the right behavior. It should apply the lights from `scene` only if `subScene` is null.

kevinrushforth (commented):
This looks mostly correct except for the case where both the Scene and the SubScene have lights. It seems worth adding a test for this case as well.

lukostyra:
Corrected the behavior and added a test as well

kevinrushforth (approved):
Looks good. I left a few minor cleanup comments. I'll reapprove when/if you fix them.

lukostyra:
Cleanup done, should be all OK now

kevinrushforth (approved):

mstr2 (approved):
Looks good!

PR #1333, opened 2024-01-14
8323706: Remove SimpleSelector and CompoundSelector classes

hjohn:
Moves `SimpleSelector` and `CompoundSelector` to internal packages.

This can be done with only a minor API break, as `SimpleSelector` and `CompoundSelector` were public before.  However, these classes could not be constructed by 3rd parties.  The only way to access them was by doing a cast (generally they're accessed via `Selector` not by their sub types).  The reason they were public at all was because the CSS engine needs to be able to access them from internal packages.

This change fixes a mistake (or possibly something that couldn't be modelled at the time) when the CSS API was first made public. The intention was always to have a `Selector` interface/abstract class, with private implementations (`SimpleSelector` and `CompoundSelector`).

This PR as said has a small API break.  The other changes are (AFAICS) source and binary compatible:

- Made `Selector` `sealed` only permitting `SimpleSelector` and `CompoundSelector` -- as `Selector` had a package private constructor, there are no concerns with pre-existing subclasses
- `Selector` has a few more methods that are now `protected` -- given that the class is now sealed, these modified methods are not accessible (they may still require rudimentary documentation I suppose)
- `Selector` now has a `public` default constructor -- as the class is sealed, it is inaccessible
- `SimpleSelector` and `CompoundSelector` have a few more `public` methods, but they're internal now, so it is irrelevant
- `createMatch` was implemented directly in `Selector` to avoid having to expose package private fields in `Match` for use by `CompoundSelector`
- No need anymore for the `SimpleSelectorShim`

---------
### Progress
- [x] Change requires CSR request [JDK-8339385](https://bugs.openjdk.org/browse/JDK-8339385) to be approved

### Issues
 * [JDK-8323706](https://bugs.openjdk.org/browse/JDK-8323706): Remove SimpleSelector and CompoundSelector classes (**Enhancement** - P4)
 * [JDK-8339385](https://bugs.openjdk.org/browse/JDK-8339385): Remove SimpleSelector and CompoundSelector classes (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1333/head:pull/1333` \
`$ git checkout pull/1333`

`$ git checkout pull/1333` \
`$ git pull https://git.openjdk.org/jfx.git pull/1333/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1333`

`$ git pr show -t 1333`

</details>

<a href="https://git.openjdk.org/jfx/pull/1333.diff">https://git.openjdk.org/jfx/pull/1333.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1333#issuecomment-1890979717)


kevinrushforth:
I'll take a closer look tomorrow, but this will very likely need to be done in two steps. Step 1 would be to terminally deprecate the existing public classes in an exported package. Step 2 would then be to remove them from the public API by moving them to a non-exported package. Given that we are still in RDP1 for JavaFX 22, we could do the first step (terminal deprecation) in 22 paving the way for removal in 23, so the net result would be the same, but it would give a clear warning to anyone using JavaFX 22 that this removal is pending.

/reviewers 3
/csr

modules/javafx.graphics/src/main/java/javafx/css/Selector.java
```diff
@@ -150,9 +173,10 @@ protected void writeBinary(DataOutputStream os, StyleConverter.StringStore strin
      * @param bssVersion bss version identifier
      * @param is {@code DataInputStream} to read {@code Selector} data from
      * @param strings string array containing selector details
+     * @return a selector, never {@code null}
      * @throws IOException if reading from {@code DataInputStream} fails
      */
-    static Selector readBinary(int bssVersion, DataInputStream is, String[] strings)
+    protected static Selector readBinary(int bssVersion, DataInputStream is, String[] strings)
```
kevinrushforth, line 179: This is an API addition, so will need an `@since 23`. Since you are adding new API here, should it be an instance method rather than a static method? Or is is called without having an instance of Selector in some cases?

kevinrushforth (commented):
This seems like a reasonable thing to do. As I mentioned above, the first step is to file a new enhancement to deprecate these two classes for removal, with the idea being to get them into jfx22 via a late enhancement request. This will need to be done quickly, since we have only 2 weeks before RDP2.

In parallel with that, discussion on openjfx-dev with a new subject indicating that you propose to remove these two classes from the public API by deprecating them for removal in JavaFX 22 and removing them in 23.

I left a few inline comments, but nothing that would call the overall proposal into question.

andy-goryachev-oracle:
Maybe @jperedadnr could check whether the ScenicView tools uses either SimpleSelector or CompoundSelector?

modules/javafx.graphics/src/main/java/javafx/css/Selector.java
```diff
@@ -96,7 +93,33 @@ public int getOrdinal() {
...
      * @return a match, never {@code null}
      */
-    public abstract Match createMatch();
+    public final Match createMatch() {
+        if (this instanceof SimpleSelector s) {
+            int idCount = s.getId().isEmpty() ? 0 : 1;
+            int styleClassCount = s.getStyleClassSet().size();
+
+            return new Match(this, s.getPseudoClassStates(), idCount, styleClassCount);
+        }
```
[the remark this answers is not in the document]
hjohn: Yes, correct, `CompoundSelector` accesses the package private fields `idCount` and `styleClassCount` of `Match` directly, which it can't do anymore after being moved to a different package; these lines:

                idCount += match.idCount;
                styleClassCount += match.styleClassCount;

I'm aware of the Helper classes, but I feel that they are a much more invasive concept (and also harder to follow) to achieve this than doing a pattern match with `instanceof` (which can be replaced with pattern matches for switch once we can use Java 21).  However, if you think this is a requirement, I'm happy to change it -- that said, we're not locked in either choice as far as I can see.

Alternatively, with everything needed in `Selector` being publicly accessible, I'm not sure if the match creation really needed to be in `Selector` or its subtypes at all.  It feels like more a job for an external type to handle (like how you don't write serialization logic for JSON or XML in each subtype).  If it were up to me, I'd probably create a static method in `Match` which given a `Selector` creates the match.  That way, no `Match` internals need exposing at all.  I could still do this, as the method needed could be package private, and then all the match fields can be made fully private.

jperedadnr:
@andy-goryachev-oracle Scenic View source code doesn't use neither of them.

modules/javafx.graphics/src/main/java/javafx/css/Selector.java
hjohn: I'll add the `@since 23`, however it can't be called by anyone except FX itself.

Some background: the readBinary/writeBinary methods are ultimately called via the publicly accessible methods `loadBinary` and `saveBinary` in `javafx.css.Stylesheet`.  

The reason it needs to be `protected` now is because `CompoundSelector` is using this (but IMHO, it shouldn't have).  Why I say it shouldn't?  That's because it already knows what it will be reading will all be `SimpleSelector`s, as it stores a counter (a `short`) and then loads that many `SimpleSelector`s.  However, by not taking the direct route of using `SimpleSelector#readBinary` (and the corresponding `SimpleSelector#writeBinary`) there is an additional `byte` prefix indicating the type of selector -- this isn't needed and wastes some space in the binary format.

Changing that now however would make the binary format incompatible with earlier versions, so I think making this `protected` is a reasonable solution.  It also mirrors the `writeBinary` method then, which was already `protected`.

hjohn: I missed something in my above evaluation.  The counterpart method `writeBinary` is not `static`.  Although that makes sense as in that case you do have an instance you wish to write, it does make the read/write API asymmetric.

It's not possible to make `readBinary` an instance method though as it is the method that is creating those instances.

The other way around is possible (make `writeBinary` static), but it would then again need a pattern match to determine the correct static `writeBinary` to call when writing an arbitrary `Selector`.  However, this would have allowed `CompoundSelector` to call a static version of `SimpleSelector#writeBinary` and `readBinary`, avoiding the extra identifying byte in the binary format, and avoiding the cast when reading it back.

The read/write loops below:

+
        final int nSelectors = is.readShort();
        final List<SimpleSelector> selectors = new ArrayList<>();
        for (int n=0; n<nSelectors; n++) {
            selectors.add((SimpleSelector)Selector.readBinary(bssVersion, is,strings));
        }
+
        os.writeShort(selectors.size());   // writes the number of SimpleSelectors to the binary stream
        for (int n=0; n< selectors.size(); n++) selectors.get(n).writeBinary(os,stringStore);  // writes each individually

Would then have become:

+
        final int nSelectors = is.readShort();
        final List<SimpleSelector> selectors = new ArrayList<>();
        for (int n=0; n<nSelectors; n++) {
            selectors.add(SimpleSelector.readBinary(bssVersion, is,strings));  // cast is gone
        }
+
        os.writeShort(selectors.size());   // writes the number of SimpleSelectors to the binary stream
        for (int n=0; n< selectors.size(); n++) SimpleSelector.writeBinaryStatic(selectors.get(n), os, stringStore);  // writes each individually

andy-goryachev-oracle:
> However, Scene Builder does:

Thank you, @jperedadnr !
Does this mean that we ought to add a public method to Selector?

hjohn:
> @andy-goryachev-oracle Scenic View source code doesn't use neither of them.
> 
> However, Scene Builder does: https://github.com/gluonhq/scenebuilder/blob/master/kit/src/main/java/com/oracle/javafx/scenebuilder/kit/util/CssInternal.java#L251
> 
> I guess it can be used a quick test of this PR?

Well, because those classes are moved, SceneBuilder would no longer work (probably something like a class not found exception will occur).

Is SceneBuilder tied to JavaFX releases?  Otherwise we may need to take more action.

I see that SceneBuilder is using these classes to obtain a list of all style classes that are part of a style sheet.  Such functionality could be offered in several places, either on `Stylesheet`, which gathers all the style classes, a public method on `Rule` which gets all style classes that the rule uses, or a public method on `Selector` which returns all style classes that the selector uses.

andy-goryachev-oracle:
In my opinion, Selector might be the most logical place to add the new methods.

This is a breaking change, so  I think we should talk about adding these methods at the same time as deprecating the classes, to give the tools developers time to switch to the new implementation and possibly re-packaging their app in a multi-release jar (it's always fun to make changes to the build scripts!).  

Will there be enough time for a complete review?

kevinrushforth:
Hmm. This suggests taking a step back. I think there are only two choices that make sense (there are a couple others, but I don't think they are useful).

1. Abandon the notion of making an incompatible change to `SimpleSelector` and `CompoundSelector`. This means that PR #1316 will need to preserve the existing `SimpleSelector` method `List<String> getStyleClasses()` and `Set<StyleClass> getStyleClassSet()`, deprecating them with simple deprecation (not for removal).
2. Make an incompatible change, meaning that at some point, the existing SceneBuilder binaries will break. We could make the transition a little less painful by providing replacement API in a release that still has the existing API (in fact, I think if we go with this option, then we _must_ do that), but the fact remains that there will then be no way to have a single binary that runs on JavaFX 21 and JavaFX 23 short of a multi-release jar file, which really isn't a good solution for this sort of problem.

Preserving the existing `SimpleSelector` methods is fairly easy and already done. So, is there any real drawback to doing this? I can't think of any.

kevinrushforth:
I note that we could still consider deprecating the `SimpleSelector` and `CompoundSelector` classes for removal, but it wouldn't have to be done in such a rush.

andy-goryachev-oracle:
> I note that we could still consider deprecating the `SimpleSelector` and `CompoundSelector` classes for _eventual_ removal

I like the idea of deprecating with the following removal, coupled with adding the "missing" APIs to Selector.  I think the lack of this method in the base class was an unfortunate oversight, and we should fix that.

hjohn:
> Hmm. This suggests taking a step back. I think there are only two choices that make sense (there are a couple others, but I don't think they are useful).
> 
> 1. Abandon the notion of making an incompatible change to `SimpleSelector` and `CompoundSelector`. This means that PR [JDK-8322964 Optimize performance of CSS selector matching #1316](https://github.com/openjdk/jfx/pull/1316) will need to preserve the existing `SimpleSelector` method `List<String> getStyleClasses()` and `Set<StyleClass> getStyleClassSet()`, deprecating them with simple deprecation (not for removal).
> 2. Make an incompatible change, meaning that at some point, the existing SceneBuilder binaries will break. We could make the transition a little less painful by providing replacement API in a release that still has the existing API (in fact, I think if we go with this option, then we _must_ do that), but the fact remains that there will then be no way to have a single binary that runs on JavaFX 21 and JavaFX 23 short of a multi-release jar file, which really isn't a good solution for this sort of problem.
> 
> Preserving the existing `SimpleSelector` methods is fairly easy and already done. So, is there any real drawback to doing this? I can't think of any.

Does the comment from José Pereda here https://github.com/openjdk/jfx/pull/1340#issuecomment-1900843857 make any difference in that regard?  If scene builder is always tied to a specific FX release, then I would think we could still go ahead.

The drawback might be that more code may start relying on these classes, or that it will be harder to evolve the CSS API if more than just Simple/CompoundSelector are needed at some point.  The API still seems to have been designed with the idea that `Selector` is the only public class.

I think regardless adding a public API to get the style classes is a good idea, it seems to round out the API a bit, and it would make Scene Builder less dependent on what we consider internals in the future.

hjohn:
> I note that we could still consider deprecating the `SimpleSelector` and `CompoundSelector` classes for _eventual_ removal, but it wouldn't have to be done in such a rush.

That's certainly true, the only reason we wanted to do a quick deprecation is to avoid adding a new public method to `SimpleSelector` for the performance improvement PR.  However, I think everyone will be able to live with adding such a method for now if that method will eventually become part of a non-exported package.

kevinrushforth:
This PR should probably be moved to draft. Now that the deprecation for removal is targeted for 23, this won't happen until JavaFX 24.

hjohn:
@kevinrushforth I think this is ready for a closer inspection again; in a strict sense there are API additions in this PR, but because of how `sealed` works, they're not accessible except to permitted sub types (and none of those permitted types are public API).

kevinrushforth:
> I think this is ready for a closer inspection again

I'll put it on my review queue for after Thursday's fork.

> in a strict sense there are API additions in this PR, but because of how `sealed` works, they're not accessible except to permitted sub types (and none of those permitted types are public API).

Right, the API additions are the package-scope constructor and the `readBinary` method, although as you say, they are only accessible to internal JavaFX classes. Since `Selector` will be a sealed class with all permitted classes being internal, no package-scope methods can be accessed by an application.

The main API change is the removal of the terminally-deprecated `SimpleSelector` and `CompoundSelector` classes. With that in mind, a better title for this bug would be:

"Remove SimpleSelector and CompoundSelector classes"

You could add the existing title as a summary:

"/summary Move SimpleSelector and CompoundSelector to internal packages"

modules/javafx.graphics/src/main/java/javafx/css/Selector.java
```diff
@@ -40,12 +43,14 @@
...
  * @since 9
  */
-abstract public class Selector {
+public abstract sealed class Selector permits SimpleSelector, CompoundSelector {
 
     /**
-     * Package scoped constructor.
+     * Explicit constructor for subtype use.
+     *
+     * @since 23
```
kevinrushforth, line 51: The pattern we use elsewhere is:

"Constructor for subclasses to call."

Also, that should be `@since 24`.

modules/javafx.graphics/src/main/java/javafx/css/Selector.java
```diff
@@ -161,10 +192,12 @@ protected void writeBinary(DataOutputStream os, StyleConverter.StringStore strin
      * @param bssVersion bss version identifier
      * @param is {@code DataInputStream} to read {@code Selector} data from
      * @param strings string array containing selector details
+     * @return a selector, never {@code null}
      * @throws IOException if reading from {@code DataInputStream} fails
+     * @since 23
```
kevinrushforth, line 197: That should be `@since 24`.

kevinrushforth (commented):
I left a couple quick comments on the API changes. I'll review the rest after the fork.

modules/javafx.graphics/src/main/java/javafx/css/Selector.java
```diff
@@ -96,7 +93,33 @@ public int getOrdinal() {
      *
      * @return a match, never {@code null}
      */
-    public abstract Match createMatch();
+    public final Match createMatch() {
```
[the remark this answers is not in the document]
andy-goryachev-oracle: I agree with Kevin here: the implementation should be moved to respective child classes.

If Match constructor is not public, then we should create a factory method in a helper, for example.

modules/javafx.graphics/src/main/java/javafx/css/Selector.java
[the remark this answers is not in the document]
andy-goryachev-oracle: what do you think of

Match MatchHelper.create(SimpleSelector)
Match MatchHelper.create(CompoundSelector)

?

modules/javafx.graphics/src/main/java/javafx/css/Selector.java
andy-goryachev-oracle: it looks to me readBinary should be static, and writeBinary an instance method - this is a normal pattern.  the asymmetry is dictated by the fact that we don't have an instance when reading, so typically read() methods read the stream and create an instance at the last moment with the specific constructor.  unless, of course, the design allows for mutation and the fields can be set().

Alternatively, readBinary() could be moved to another class (helper? reader?) to avoid making this an API change.

what do you think?

hjohn:
/summary Move SimpleSelector and CompoundSelector to internal packages

modules/javafx.graphics/src/main/java/javafx/css/Selector.java
hjohn: I can see if I can externalize this or if that would run into issues.  Also please note, although technically an API change, it is NOT an accessible API (and so can be removed at any time) because only the permitted types can access this API.

In other words, this could wait and be done separately or never.

andy-goryachev-oracle: moving it to an internal class would eliminate the public API.

I agree with you that the risk is rather low because because the probability of people ever touching these classes is nearly zero.

kevinrushforth: I am fine with either approach. If it is easy, you can move it now, but there are no compatibility concerns with doing it later since a protected method of a sealed class without any exported subclasses is not accessible.

andy-goryachev-oracle: what is the resolution, @hjohn ?

hjohn: - I've moved the serialization logic to `BinarySerializer` and removed the two methods from `Selector`.  `writeBinary` was `protected` but couldn't be accessed before anyway as it was impossible to subclass `Selector` (as it had a package private constructor)
- I've moved most of the code to create a `Match` to the `Match` class so the relevant fields can now be `private`. 

Let me know if this is an improvement.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/BinarySerializer.java
```diff
@@ -0,0 +1,186 @@
...
+        for (int n = 0; n < nSelectors; n++) {
+
+            /*
+             * An extra byte is part of the binary format containing TYPE_SIMPLE that
+             * isn't strictly needed. However, to remain backwards compatible with the
+             * first (and only) supported binary format, this byte must be skipped over
+             * before attempting to read what is known to be a simple selector.
+             */
+
+            is.readByte();
```
andy-goryachev-oracle, line 106: should we still check the value and throw an IOE if it is wrong _for security reasons_?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/BinarySerializer.java
```diff
@@ -0,0 +1,186 @@
...
+
+            selectors.add(readSimpleSelector(is, strings));
+        }
+
+        int nRelationships = is.readShort();
```
andy-goryachev-oracle, line 111: same here: should we check for a positive value?

as a general rule, we should be validating the input as it might come from untrusted sources, right?  L79 and other places?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/BinarySerializer.java
```diff
@@ -0,0 +1,186 @@
...
+
+            if (ordinal == Combinator.CHILD.ordinal()) {
+                relationships.add(Combinator.CHILD);
+            }
+            else if (ordinal == Combinator.DESCENDANT.ordinal()) {
+                relationships.add(Combinator.DESCENDANT);
+            }
+            else {
+                assert false : "error deserializing CompoundSelector: Combinator = " + ordinal;
+                relationships.add(Combinator.DESCENDANT);
```
andy-goryachev-oracle, line 126: throw an exception instead?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/CompoundSelector.java
```diff
@@ -78,20 +69,21 @@ public List<SimpleSelector> getSelectors() {
...
-    //  * The relationships between the selectors
-    //  * @return Immutable List&lt;Combinator&gt;
-    //  */
-    // public List<Combinator> getRelationships() {
-    //     return relationships;
-    // }
+
+    /**
+     * The relationships between the selectors
+     * @return Immutable List&lt;Combinator&gt;
```
andy-goryachev-oracle, line 75: minor: would {@code List<Combinator>} be a better choice here, if it works?

andy-goryachev-oracle (commented):
this PR has been out for too long, could you please merge the latest master branch in?

hjohn:
> this PR has been out for too long, could you please merge the latest master branch in?

I did yesterday, or did something go wrong?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/BinarySerializer.java
hjohn: This is original code, just moved.  Do you want me to change it still?

andy-goryachev-oracle:
oops, yes, you did, my mistake.  the master progressed since then, got me confused.

I got a bunch of errors again after switching:

```
Cannot infer type arguments for ParsedValueImpl<>	CssParser.java	Java Problem	/graphics/src/main/java/javafx/css	line 3024
Cannot infer type arguments for ParsedValueImpl<>	CssParser.java	Java Problem	/graphics/src/main/java/javafx/css	line 3214
Cannot infer type arguments for ParsedValueImpl<>	CssParser.java	Java Problem	/graphics/src/main/java/javafx/css	line 3267
Cannot infer type arguments for ParsedValueImpl<>	CssParser.java	Java Problem	/graphics/src/main/java/javafx/css	line 3282
Cannot infer type arguments for ParsedValueImpl<>	CssParser.java	Java Problem	/graphics/src/main/java/javafx/css	line 3607
Cannot infer type arguments for ParsedValueImpl<>	CssParser.java	Java Problem	/graphics/src/main/java/javafx/css	line 3651

```

modules/javafx.graphics/src/main/java/com/sun/javafx/css/BinarySerializer.java
hjohn: I considered doing more here, but as this is all just moved code, I'm hesitant to change it as part of this PR.  For example, if there is a faulty binary CSS file which has a negative value for the short, then the original code will just skip the loop.  If I add a check, it will change the behavior.

andy-goryachev-oracle: Well, we could create a separate ticket.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/BinarySerializer.java
hjohn: I'm not entirely sure what security reasons you may be referring to; we're not instantiating classes by string for example, and Java doesn't suffer from out of bounds memory accesses due to a bad interpretation of a binary format.

However, this code is a bit more lenient than the original.  The original code would deserialize this as a `CompoundSelector` but then subsequently fail with a `ClassCastException` as it assumes it will be a `SimpleSelector`. I could make it fully compatible in this regard, although it is extremely unlikely that there are any bad files out there where this would apply (as the old implementation would never load it).

Anyway, I'm fine with rejecting values that are not `TYPE_SIMPLE`.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/BinarySerializer.java
hjohn: I already cleaned it up without altering the behavior.  It probably isn't worth investing more time in.  The whole binary CSS code is AFAICS basically only used to create binary versions of the FX standard css files, like `modena.css`, by calling the main method of `Css2Bin` during the gradle build.  This probably saves some time for JavaFX startup as the binary version will load faster than the text versions (or so I hope, I didn't check if that's true).  I have doubts it is used anywhere else, so although this has creeped into a few public API's, it's practically for internal use only.

andy-goryachev-oracle (approved):
You are right, @hjohn , we probably should not alter the behavior in this PR.

The rest looks good.  If you decide to make the minor changes I've mentioned, I'll re-approve.

modules/javafx.graphics/src/main/java/javafx/css/Match.java
```diff
@@ -42,15 +46,46 @@
...
-    final int idCount;
+    private final int styleClassCount;
+    private final int idCount;
 
     // CSS3 spec gives weight to id count, then style class count,
     // then pseudoclass count, and finally matching types (i.e., java name count)
-    final int specificity;
+    private final int specificity;
 
     @SuppressWarnings("removal")
```
nlisker, line 90: What is this removal warning suppression? I know it was there before these changes.

nlisker:
I see a lot of aversion from for-each loops in favor of indexed loops in the classes this PR touches. While I don't suggest to change it here, any idea why?

hjohn:
> I see a lot of aversion from for-each loops in favor of indexed loops in the classes this PR touches. While I don't suggest to change it here, any idea why?

It's just the style the original code was written with.  There is no reason to write the binary load/save code like this (as IO will be a far greater bottleneck).

There are/were other parts (in `Match` and the selector subtypes) however that are performance sensitive, so perhaps the entire classes were just written in a similar style "just in case".

modules/javafx.graphics/src/main/java/javafx/css/Match.java
hjohn: `SimpleSelector` was deprecated for removal (from the public javafx.css package). Now that it is moved to the internal package, the deprecation on that was also removed.  The `SuppressWarnings` is therefore no longer needed here as the deprecation is no longer present on the now internal `SimpleSelector` type.

hjohn: The diff seems to show I did not remove that line, but it should no longer be needed.  I'll check.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/CompoundSelector.java
hjohn: Fixed

nlisker:
> oops, yes, you did, my mistake. the master progressed since then, got me confused.
> 
> I got a bunch of errors again after switching:
> 
> ```
> Description	Resource	Type	Path	Location
> Cannot infer type arguments for ParsedValueImpl<>	CssParser.java	Java Problem	/graphics/src/main/java/javafx/css	line 3024
> Cannot infer type arguments for ParsedValueImpl<>	CssParser.java	Java Problem	/graphics/src/main/java/javafx/css	line 3214
> Cannot infer type arguments for ParsedValueImpl<>	CssParser.java	Java Problem	/graphics/src/main/java/javafx/css	line 3267
> Cannot infer type arguments for ParsedValueImpl<>	CssParser.java	Java Problem	/graphics/src/main/java/javafx/css	line 3282
> Cannot infer type arguments for ParsedValueImpl<>	CssParser.java	Java Problem	/graphics/src/main/java/javafx/css	line 3607
> Cannot infer type arguments for ParsedValueImpl<>	CssParser.java	Java Problem	/graphics/src/main/java/javafx/css	line 3651
> ```

I got these too, as well as many
```
The method createShader(String, InputStream, Map<String,Integer>, Map<String,Integer>, int, boolean, boolean) in the type ShaderFactory is not applicable for the arguments (InputStream, HashMap<String,Integer>, HashMap<String,Integer>, int, boolean, boolean)
```
I haven't worked on this repo for 3 weeks, what changed?

hjohn:
> > oops, yes, you did, my mistake. the master progressed since then, got me confused.
> > I got a bunch of errors again after switching:
> > ```
> > Description	Resource	Type	Path	Location
> > Cannot infer type arguments for ParsedValueImpl<>	CssParser.java	Java Problem	/graphics/src/main/java/javafx/css	line 3024
> > Cannot infer type arguments for ParsedValueImpl<>	CssParser.java	Java Problem	/graphics/src/main/java/javafx/css	line 3214
> > Cannot infer type arguments for ParsedValueImpl<>	CssParser.java	Java Problem	/graphics/src/main/java/javafx/css	line 3267
> > Cannot infer type arguments for ParsedValueImpl<>	CssParser.java	Java Problem	/graphics/src/main/java/javafx/css	line 3282
> > Cannot infer type arguments for ParsedValueImpl<>	CssParser.java	Java Problem	/graphics/src/main/java/javafx/css	line 3607
> > Cannot infer type arguments for ParsedValueImpl<>	CssParser.java	Java Problem	/graphics/src/main/java/javafx/css	line 3651
> > ```
> 
> I got these too, as well as many
> 
> ```
> The method createShader(String, InputStream, Map<String,Integer>, Map<String,Integer>, int, boolean, boolean) in the type ShaderFactory is not applicable for the arguments (InputStream, HashMap<String,Integer>, HashMap<String,Integer>, int, boolean, boolean)
> ```
> 
> I haven't worked on this repo for 3 weeks, what changed?

The ones mentioned by Andy I also see, and are just a difference in opinion between Javac and ECJ, or perhaps a bug in the incremental compilation that Eclipse does (as they usually only appear when it does an incremental compile, not a full one).  When those occur, I just select them and remove them from the errors list.

The others I haven't seen before, and I'm not seeing them now.  I did a search for `createShader` but didn't see anything unusual.

I'm not aware of any special changes in the last 3 weeks.  I did a fresh import recently, and nothing strange popped up.

kevinrushforth:
> ```
> The method createShader(String, InputStream, Map<String,Integer>, Map<String,Integer>, int, boolean, boolean) in the type ShaderFactory is not applicable for the arguments (InputStream, HashMap<String,Integer>, HashMap<String,Integer>, int, boolean, boolean)
> ```
> 
> I haven't worked on this repo for 3 weeks, what changed?

You need to do a clean build. This method was changed by [JDK-8336624](https://bugs.openjdk.org/browse/JDK-8336624) / PR #1530

andy-goryachev-oracle (approved):

modules/javafx.graphics/src/main/java/com/sun/javafx/css/BinarySerializer.java
kevinrushforth: > Anyway, I'm fine with rejecting values that are not `TYPE_SIMPLE`.

That seems like a good idea for maximum compatibility.

kevinrushforth (approved):
Doc changes and implementation both look good. I left one comment and will reapprove if you decide to make that change.

Please create the CSR.

kevinrushforth:
This PR is more limited in scope than when I initially requested 3 reviewers. I think 2 reviewers are sufficient. If someone else wants to review it as well, that would be fine. It will take a few more days to get the CSR reviewed and approved, so there will be time for others to weigh in if desired.

/reviewers 2 reviewers

andy-goryachev-oracle:
I agree - two reviewers should be fine.

kevinrushforth:
@hjohn This PR is approved and just waiting for the CSR. Please create one when you have a chance.

modules/javafx.graphics/src/main/java/javafx/css/Match.java
hjohn: The line was correctly removed.

hjohn:
I made the small change suggested in `BinarySerializer` and created a CSR.  I wasn't entirely sure on the "compatability" part and whether or not it should take into account removal of deprecated classes.  With the removal of `SimpleSelector` and `CompoundSelector` code that managed to use them (by casting) will no longer compile for sure.  I'm not 100% sure if it could possibly be binary compatible (ie. if a previously compiled class casts to these types) but I suspect it won't be.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/BinarySerializer.java
```diff
@@ -103,7 +103,11 @@ private static CompoundSelector readCompoundSelector(DataInputStream is, String[
              * before attempting to read what is known to be a simple selector.
              */
 
-            is.readByte();
+            byte type = is.readByte();
+
+            if (type != TYPE_SIMPLE) {
+                throw new IllegalStateException("Expected compound selector to consist of simple selectors only, but found type: " + type);
```
andy-goryachev-oracle, line 109: minor: the output will be a signed decimal.  would it be better if this can be formatted as hex?

`String.format("Expected compound selector to consist of TYPE_SIMPLE only, but found type: 0x%02X", type)`

kevinrushforth (approved):
Looks good.

And I think the CSR is ready to Finalize.

andy-goryachev-oracle:
can't review if my questions are not answered.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/BinarySerializer.java
hjohn: I don't think that would matter much.  This error will only ever be interpreted by programmers, I'm sure they'll be able to understand what's going on.

kevinrushforth: Agreed. This should never happen, and if it does, the difference between a hex and decimal printout for the erroneous value wouldn't matter.

andy-goryachev-oracle: I disagree, especially since the target audience is developers.  When this happens and the dev needs to diagnose, hex is better: some things become immediately obvious, and some can be found in the binary viewer which is always in hex.  I don't understand why you resist.

Will re-approve if you decide to fix it.

andy-goryachev-oracle (approved):
minor fix suggested (hex).

PR #1169, opened 2023-07-04
8090267: JFXPanel Input Problem

prsadhuk:
When Japanse (IME on) is inputted to the TextFIeld, which is on JFXPanel, 
small window for inputting appears on top-left side of screen

image

For swing-interop case, WmImeStartComposition starts composition in native ImmSetCompositionWindow window as "m_useNativeCompWindow" below is true for FX
https://github.com/openjdk/jdk/blob/514816ed7d7dea1fb13d32b80aef89774bee13d3/src/java.desktop/windows/native/libawt/windows/awt_Component.cpp#L3957

m_useNativeCompWindow is true because during sun.awt.im.InputContext#focusGained() calls activateInputMethod which calls WInputMethod.activate() which calls haveActiveClient() which checks for
clientComponent.getInputMethodRequests().
Now, in JFXPanel, getInputMethodRequests() returns null as setEmbeddedScene() is not called yet.
Since getInputMethodRequests() returns null, haveActiveClient() is false which calls enableNativeIME() with 1 [thereby native composition window is enabled]
https://github.com/openjdk/jdk/blob/514816ed7d7dea1fb13d32b80aef89774bee13d3/src/java.desktop/windows/classes/sun/awt/windows/WInputMethod.java#L316

Proposed fix is to ensure there is an active client "initially" so that enableNativeIME() is called with 0 and no native compostion window is shown.
getInputMethodRequests() is called in setEmbeddedScene() so as to make sure getInputMethodRequest() is initialised to correct "InputMethodSupport.InputMethodRequestsAdapter.fxRequests" object and not NULL.

AFter fix
image

---------
### Progress
- [ ] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8090267](https://bugs.openjdk.org/browse/JDK-8090267): JFXPanel Input Problem (**Bug** - P3)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1169/head:pull/1169` \
`$ git checkout pull/1169`

Update a local copy of the PR: \
`$ git checkout pull/1169` \
`$ git pull https://git.openjdk.org/jfx.git pull/1169/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
[...]

PR #1337, opened 2024-01-16
8322784: JFXPanel calls InputMethodRequests on wrong thread

beldenfox:
On Windows we need to ensure InputMethodRequests coming from JFXPanel are processed on the JavaFX application thread instead of the AWT EventQueue thread. This PR adds the runAndWait() calls to do that.

This would be difficult to test on Windows without a fix for [JDK-8090267](https://bugs.openjdk.org/browse/JDK-8090267) so I've included the fix first proposed by @prsadhuk in PR #1169. If a developer uses the sample code provided in the JavaDoc to create and show a JFXPanel there's a good chance the JFXPanel will get focus before the scene has been set. To ensure AWT always treats the JFXPanel as an active IME client we return a stub version of the InputMethodRequests object if there's no scene. AWT will continue to ask for the InputMethodRequests and once the scene has been set the panel will return a non-stub version.

---------
### Progress

### Issues
 * [JDK-8322784](https://bugs.openjdk.org/browse/JDK-8322784): JFXPanel calls InputMethodRequests on wrong thread (**Bug** - P3)
 * [JDK-8090267](https://bugs.openjdk.org/browse/JDK-8090267): JFXPanel Input Problem (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1337/head:pull/1337` \
`$ git checkout pull/1337`

`$ git checkout pull/1337` \
`$ git pull https://git.openjdk.org/jfx.git pull/1337/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1337`

`$ git pr show -t 1337`

</details>

<a href="https://git.openjdk.org/jfx/pull/1337.diff">https://git.openjdk.org/jfx/pull/1337.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1337#issuecomment-1894257134)

### DIFF
--- a/modules/javafx.swing/src/main/java/javafx/embed/swing/InputMethodSupport.java
+++ b/modules/javafx.swing/src/main/java/javafx/embed/swing/InputMethodSupport.java
@@ -43,6 +43,8 @@ import java.text.AttributedString;
 import java.text.CharacterIterator;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 
 /**
  * A utility class containing the functions to support Input Methods
@@ -60,58 +62,59 @@ class InputMethodSupport {
 
         @Override
         public Rectangle getTextLocation(TextHitInfo offset) {
-            Point2D[] location = { new Point2D(0.0, 0.0) };
+            AtomicReference<Point2D> location = new AtomicReference<>(new Point2D(0.0, 0.0));
             if (fxRequests != null) {
                 PlatformImpl.runAndWait(() -> {
-                    location[0] = fxRequests.getTextLocation(offset.getInsertionIndex());
+                    location.set(fxRequests.getTextLocation(offset.getInsertionIndex()));
                 });
             }
-            return new Rectangle((int)location[0].getX(), (int)location[0].getY(), 0, 0);
+            return new Rectangle((int)location.get().getX(), (int)location.get().getY(), 0, 0);
         }
 
         @Override
         public TextHitInfo getLocationOffset(int x, int y) {
-            int[] offset = { 0 };
+            AtomicInteger offset = new AtomicInteger(0);
             if (fxRequests != null) {
                 PlatformImpl.runAndWait(() -> {
-                    offset[0] = fxRequests.getLocationOffset(x, y);
+                    offset.set(fxRequests.getLocationOffset(x, y));
                 });
             }
-            return TextHitInfo.afterOffset(offset[0]);
+            return TextHitInfo.afterOffset(offset.get());
         }
 
         @Override
         public int getInsertPositionOffset() {
-            int[] offset = { 0 };
+            AtomicInteger offset = new AtomicInteger(0);
             if (fxRequests instanceof ExtendedInputMethodRequests) {
                 PlatformImpl.runAndWait(() -> {
-                    offset[0] = ((ExtendedInputMethodRequests)fxRequests).getInsertPositionOffset();
+                    offset.set(((ExtendedInputMethodRequests)fxRequests).getInsertPositionOffset());
                 });
             }
-            return offset[0];
+            return offset.get();
         }
 
         @Override
         public AttributedCharacterIterator getCommittedText(int beginIndex, int endIndex, AttributedCharacterIterator.Attribute[] attributes) {
-            String[] comitted = { null };
+            AtomicReference<String> committed = new AtomicReference<>(null);
             if (fxRequests instanceof ExtendedInputMethodRequests) {
                 PlatformImpl.runAndWait(() -> {
-                    comitted[0] = ((ExtendedInputMethodRequests)fxRequests).getCommittedText(beginIndex, endIndex);
+                    committed.set(((ExtendedInputMethodRequests)fxRequests).getCommittedText(beginIndex, endIndex));
                 });
             }
-            if (comitted[0] == null) comitted[0] = "";
-            return new AttributedString(comitted[0]).getIterator();
+            String text = committed.get();
+            if (text == null) text = "";
+            return new AttributedString(text).getIterator();
         }
 
         @Override
         public int getCommittedTextLength() {
-            int[] length = { 0 };
+            AtomicInteger length = new AtomicInteger(0);
             if (fxRequests instanceof ExtendedInputMethodRequests) {
                 PlatformImpl.runAndWait(() -> {
-                    length[0] = ((ExtendedInputMethodRequests)fxRequests).getCommittedTextLength();
+                    length.set(((ExtendedInputMethodRequests)fxRequests).getCommittedTextLength());
                 });
             }
-            return length[0];
+            return length.get();
         }
 
         @Override
@@ -122,14 +125,15 @@ class InputMethodSupport {
 
         @Override
         public AttributedCharacterIterator getSelectedText(AttributedCharacterIterator.Attribute[] attributes) {
-            String[] selected = { null };
+            AtomicReference<String> selected = new AtomicReference<>(null);
             if (fxRequests != null) {
                 PlatformImpl.runAndWait(() -> {
-                    selected[0] = fxRequests.getSelectedText();
+                    selected.set(fxRequests.getSelectedText());
                 });
             }
-            if (selected[0] == null) selected[0] = "";
-            return new AttributedString(selected[0]).getIterator();
+            String text = selected.get();
+            if (text == null) text = "";
+            return new AttributedString(text).getIterator();
         }
     }
 
--- a/modules/javafx.swing/src/main/java/javafx/embed/swing/InputMethodSupport.java
+++ b/modules/javafx.swing/src/main/java/javafx/embed/swing/InputMethodSupport.java
@@ -25,8 +25,10 @@
 
 package javafx.embed.swing;
 
+import com.sun.javafx.application.PlatformImpl;
 import com.sun.javafx.collections.ObservableListWrapper;
 import com.sun.javafx.scene.input.ExtendedInputMethodRequests;
+import javafx.application.Platform;
 import javafx.collections.ObservableList;
 import javafx.geometry.Point2D;
 import javafx.scene.input.InputMethodHighlight;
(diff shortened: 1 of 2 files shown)


prsadhuk:
Works ok in windows11 as per my testing...BTW, did your testing on mac works as deadlock was seen as per JBS comment? probably should have been solved by [JDK-8221261](https://bugs.openjdk.org/browse/JDK-8221261)

beldenfox:
Tested on Mac, worked fine with no deadlock.

modules/javafx.swing/src/main/java/javafx/embed/swing/InputMethodSupport.java
```diff
@@ -56,42 +58,64 @@ public InputMethodRequestsAdapter(javafx.scene.input.InputMethodRequests fxReque
...
+            stringValue = null;
             if (fxRequests instanceof ExtendedInputMethodRequests) {
-                result = ((ExtendedInputMethodRequests)fxRequests).getCommittedText(beginIndex, endIndex);
+                PlatformImpl.runAndWait(() -> {
+                    stringValue = ((ExtendedInputMethodRequests)fxRequests).getCommittedText(beginIndex, endIndex);
+                });
             }
-            if (result == null) result = "";
-            return new AttributedString(result).getIterator();
+            if (stringValue == null) stringValue = "";
```
prsadhuk, line 106: Not sure of FX coding guidelines but in JDK, even for a single statement, we need to put in braces

prsadhuk: I guess this should add [JDK-8090267](https://bugs.openjdk.org/browse/JDK-8090267) through "/issue add" command

modules/javafx.swing/src/main/java/javafx/embed/swing/InputMethodSupport.java
```diff
@@ -56,42 +58,64 @@ public InputMethodRequestsAdapter(javafx.scene.input.InputMethodRequests fxReque
             this.fxRequests = fxRequests;
         }
 
+        private Point2D pointValue;
+        private int     intValue;
+        private String  stringValue;
```
kevinrushforth, line 63: Using instance fields like this is not thread-safe in general (although unlikely to cause problems in practice for this specific case). I recommend either creating a utility method that returns the result of a `Future` (see PR #1321 ) or changing these to local `AtomicObject<Type>` / `AtomicInteger` (or `Type[]` `int[]`) variables.

modules/javafx.swing/src/main/java/javafx/embed/swing/InputMethodSupport.java
kevinrushforth: Regarding FX coding style, the one place where braces are not needed is when everything is all on one line.

Regarding [JDK-8090267](https://bugs.openjdk.org/browse/JDK-8090267), should the call to  `getInputMethodRequests` added to [JFXPanel line 1028](https://github.com/openjdk/jfx/pull/1169/files#diff-e6f5782dba36f256cceb649ce526ed1f7a8f6704d3a32bd8bd17cef42621d99cR1028) by PR #1169 be added to this PR as well?

modules/javafx.swing/src/main/java/javafx/embed/swing/InputMethodSupport.java
beldenfox: Using instance fields here does look clunky. There's no thread-safety issue because the JFXPanel code creates a new adapter for every request so I decided to avoid the overhead of creating local arrays (apologies for the micro-optimizing). I will switch to local array variables if you want.

For the same reason I don't think the `getInputMethodRequests` call added to line 1028 in PR #1169 is necessary. The adapter it creates is discarded so the only side-effect is calling the scene's `getInputMethodRequests` before the requests are needed. I can't think of any reason this front-loading would be necessary.

modules/javafx.swing/src/main/java/javafx/embed/swing/InputMethodSupport.java
prsadhuk: /issue add JDK-8090267

kevinrushforth:
> /issue add JDK-8090267

Alternatively, you could close JDK-8090267 as a duplicate. Either is fine with me.

modules/javafx.swing/src/main/java/javafx/embed/swing/InputMethodSupport.java
```diff
@@ -102,9 +122,14 @@ public AttributedCharacterIterator cancelLatestCommittedText(AttributedCharacter
 
         @Override
         public AttributedCharacterIterator getSelectedText(AttributedCharacterIterator.Attribute[] attributes) {
-            String text = fxRequests.getSelectedText();
-            if (text == null) text = "";
-            return new AttributedString(text).getIterator();
+            String[] selected = { null };
```
andy-goryachev-oracle, line 125: Would AtomicReference<String> be a better choice here?

modules/javafx.swing/src/main/java/javafx/embed/swing/InputMethodSupport.java
```diff
@@ -102,9 +122,14 @@ public AttributedCharacterIterator cancelLatestCommittedText(AttributedCharacter
...
+            if (fxRequests != null) {
+                PlatformImpl.runAndWait(() -> {
+                    selected[0] = fxRequests.getSelectedText();
+                });
+            }
+            if (selected[0] == null) selected[0] = "";
+            return new AttributedString(selected[0]).getIterator();
```
andy-goryachev-oracle, line 132:
```
AtomicReference<String> ref = ...
...
String s = ref.get();
return new AttributedString(s == null ? "" : s).getIterator();
```

here and L102?

andy-goryachev-oracle:
Noticed a (possibly unrelated) problem -
Upon first launch , the IME window appears in the bottom right corner instead of near the text field.  Dismiss, type - the windows appears where it's supposed to be:


Screenshot 2024-01-19 145505

andy-goryachev-oracle:
the fix seems to work on window 11 (with one abovementioned problem).

andy-goryachev-oracle:
I think this code crashes on macOS with Arabic input...

modules/javafx.swing/src/main/java/javafx/embed/swing/InputMethodSupport.java
beldenfox: I'm not familiar enough with AtomicReference to have an opinion on this. Someone else will have to weigh in on this.

beldenfox:
@andy-goryachev-oracle I can reproduce the misplaced input window problem on Win 11. I added some println statements and verified that JavaFX is returning correct coordinates. My guess is that this is an OS issue though there's the small chance that the problem is in Swing.

I'll update JDK-8324232 with some notes. As you've figured out it's an existing problem that predates this PR.

modules/javafx.swing/src/main/java/javafx/embed/swing/InputMethodSupport.java
kevinrushforth: Either an array or an AtomicReference is OK in this case. AtomicReference is thread-safe, but that's not a concern here, since using `runAndWait` ensures that writing to the array happens before reading it.

kevinrushforth (commented):
The latest code changes look good to me. I'll do some additional testing, and I'm sure Andy plans to do the same.

modules/javafx.swing/src/main/java/javafx/embed/swing/InputMethodSupport.java
andy-goryachev-oracle: I might be wrong, but array would need to store its size (an int, 8 bytes on 64 bit machines) in addition to the pointers themselves, while AtomicReference needs just the pointer.
Plus, its semantics is much cleaner.

beldenfox: Again, thread-safety isn't an issue here. Even if the OS were to invoke the IM methods on different threads (which is a terrifying thought) these adapters are created on a per-call basis so each thread would get its own instance. This also means those extra eight bytes are transitory.

If there's an established JavaFX pattern of using AtomicReference instead of an array I will gladly make the changes. But this does require manual testing (I still haven't figured out how to automate this) so I'm reluctant to do this based on personal style preferences.

andy-goryachev-oracle: Here is more info [0].

We **are** dealing with multiple threads.

I've seen weird cases where modifying memory from different threads without explicit synchronization resulted in subtle bugs - we don't know all the details of cache coherence [1] implementation in multi-core systems, so I would strongly suggest to use AtomicReference.

[0] https://stackoverflow.com/questions/70381029/java-atomincinteger-vs-one-element-array-for-streams

[1] https://en.wikipedia.org/wiki/Cache_coherence

beldenfox: Sorry, I was still thinking about the original thread-safety concern which involved simultaneous access to the fields in this adapter object. Not the issue here.

I agree that explicit synchronization needs to happen but I thought that was what `runAndWait` was for. @kevinrushforth also states that `runAndWait` should be sufficient. If you think there's a cache coherence problem that's much, much bigger than this PR since this pattern is used throughout the code base.

andy-goryachev-oracle: > this pattern is used throughout the code base.

that might be the case, but I think we should use the right pattern for the new code.
single-element array is an anachronism, especially in the case of multi-threaded code as here.

kevinrushforth (approved):
Looks good.

I tested it on both macOS (to confirm no deadlock) and on Windows (to test the fix). All good.

andy-goryachev-oracle:
For the misplaced IME window, I could only find this - 
[JDK-8088485](https://bugs.openjdk.org/browse/JDK-8088485) Windows 8, Chinese input method cadidate window is always placed at the bottom of the screen.

Not sure if that's the same issue or not, as it is pretty old.

andy-goryachev-oracle (approved):
verified on windows 11 with atomic references, thank you for making the change!

I'll probably create a new bug for the misplaced IME window issue.

PR #1339, opened 2024-01-18
8323879: constructor Path(Path) which takes another Path object fail to draw on canvas html

jaybhaskar:
Issue: constructor Path(Path) which takes another Path object fails to draw on canvas html.
Solution: copy the old path while creating a new Path object from the existing Path that is already created with the same canvas rendering context.

---------
### Progress

### Issue
 * [JDK-8323879](https://bugs.openjdk.org/browse/JDK-8323879): constructor Path(Path) which takes another Path object fail to draw on canvas html (**Bug** - P2)

### Reviewers
 * [Hima Bindu Meda](https://openjdk.org/census#hmeda) (@HimaBinduMeda - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1339/head:pull/1339` \
`$ git checkout pull/1339`

`$ git checkout pull/1339` \
`$ git pull https://git.openjdk.org/jfx.git pull/1339/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1339`

`$ git pr show -t 1339`

</details>

<a href="https://git.openjdk.org/jfx/pull/1339.diff">https://git.openjdk.org/jfx/pull/1339.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1339#issuecomment-1897747831)

### DIFF
--- a/modules/javafx.web/src/main/native/Source/WebCore/platform/graphics/java/PathJava.cpp
+++ b/modules/javafx.web/src/main/native/Source/WebCore/platform/graphics/java/PathJava.cpp
@@ -118,13 +118,11 @@ PathJava::PathJava(RefPtr<RQRef>&& platformPath, std::unique_ptr<PathStream>&& e
 
 UniqueRef<PathImpl> PathJava::clone() const
 {
-    auto platformPathCopy = createEmptyPath();
-
-    RefPtr<RQRef> pathCopy(copyPath(platformPath()));
+    RefPtr<RQRef> platformPathCopy(copyPath(platformPath()));
 
     auto elementsStream = m_elementsStream ? m_elementsStream->clone().moveToUniquePtr() : nullptr;
 
-    return PathJava::create(WTFMove(pathCopy), std::unique_ptr<PathStream> { downcast<PathStream>(elementsStream.release()) });
+    return PathJava::create(WTFMove(platformPathCopy), std::unique_ptr<PathStream> { downcast<PathStream>(elementsStream.release()) });
 }
 
 PlatformPathPtr PathJava::platformPath() const
--- a/modules/javafx.web/src/main/native/Source/WebCore/platform/graphics/java/PathJava.cpp
+++ b/modules/javafx.web/src/main/native/Source/WebCore/platform/graphics/java/PathJava.cpp
@@ -120,9 +120,11 @@ UniqueRef<PathImpl> PathJava::clone() const
 {
     auto platformPathCopy = createEmptyPath();
 
+    RefPtr<RQRef> pathCopy(copyPath(platformPath()));
+
     auto elementsStream = m_elementsStream ? m_elementsStream->clone().moveToUniquePtr() : nullptr;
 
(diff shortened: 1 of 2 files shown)


kevinrushforth:
Reviewers: @HimaBinduMeda @kevinrushforth 

/reviewers 2

modules/javafx.web/src/main/native/Source/WebCore/platform/graphics/java/PathJava.cpp
```diff
@@ -120,9 +120,11 @@ UniqueRef<PathImpl> PathJava::clone() const
 {
     auto platformPathCopy = createEmptyPath();
 
+    RefPtr<RQRef> pathCopy(copyPath(platformPath()));
```
kevinrushforth, line 123: The `platformPathCopy` local variable is now unused and can be removed. Better still, you could make the following change and then revert the name change in the `create` method where it is used.

```
    RefPtr<RQRef> platformPathCopy(copyPath(platformPath()));
```

kevinrushforth (commented):
The fix looks reasonable, although there is now an unused variable, and I have a suggestion for making it clearer.

I verified that the test fails on Windows without your fix. I will rebuild WebKit and verify that it passes with your fix.

modules/javafx.web/src/test/java/test/javafx/scene/web/PathContructorTest.java
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+package test.javafx.scene.web;
+
+import com.sun.webkit.WebPage;
+import com.sun.webkit.WebPageShim;
+import java.awt.Color;
+import java.awt.image.BufferedImage;
+import java.util.Base64;
+import javafx.scene.web.WebEngineShim;
```
kevinrushforth, line 33: All of these imports are unused and can be removed.

modules/javafx.web/src/test/java/test/javafx/scene/web/PathContructorTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+
+import org.junit.After;
```
kevinrushforth, line 36: Unused import.

kevinrushforth (commented):
The test passes with your fix.

There are a few unused imports in the test that you can remove.

modules/javafx.web/src/test/java/test/javafx/scene/web/PathContructorTest.java
jaybhaskar: done

modules/javafx.web/src/test/java/test/javafx/scene/web/PathContructorTest.java
jaybhaskar: done

kevinrushforth (approved):

HimaBinduMeda (approved):
Verified the changes. Testing is green

PR #1340, opened 2024-01-19
JDK-8324182 Deprecate for removal SimpleSelector and CompoundSelector classes

hjohn:
The SimpleSelector and CompoundSelector classes are public classes in an exported package, javafx.css, but they are not intended to be used by applications. They are implementation details. They cannot be constructed directly and no other JavaFX API accepts or returns a SimpleSelector or CompoundSelector.

We should deprecate them for removal so we can move them to a non-exported package, removing them from the public API.

---------
### Progress
- [x] Change requires CSR request [JDK-8325704](https://bugs.openjdk.org/browse/JDK-8325704) to be approved

### Issues
 * [JDK-8324182](https://bugs.openjdk.org/browse/JDK-8324182): Deprecate for removal SimpleSelector and CompoundSelector classes (**Enhancement** - P4)
 * [JDK-8325704](https://bugs.openjdk.org/browse/JDK-8325704): Deprecate for removal SimpleSelector and CompoundSelector classes (**CSR**)

### Reviewers
 * [Nir Lisker](https://openjdk.org/census#nlisker) (@nlisker - **Reviewer**) 🔄 Re-review required (review applies to [6c2a9265](https://git.openjdk.org/jfx/pull/1340/files/6c2a926519a40165e6ec09aa9a7a55ba32671ceb))

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1340/head:pull/1340` \
`$ git checkout pull/1340`

`$ git checkout pull/1340` \
`$ git pull https://git.openjdk.org/jfx.git pull/1340/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1340`

`$ git pr show -t 1340`

</details>

<a href="https://git.openjdk.org/jfx/pull/1340.diff">https://git.openjdk.org/jfx/pull/1340.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1340#issuecomment-1900115054)

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
+++ b/modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
@@ -36,6 +36,7 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import com.sun.javafx.css.ImmutablePseudoClassSetsCache;
 import com.sun.javafx.css.PseudoClassState;
@@ -51,7 +52,7 @@ import static javafx.geometry.NodeOrientation.RIGHT_TO_LEFT;
  * @since 9
  * @deprecated This class was exposed erroneously and will be removed in a future version
  */
-@Deprecated(since = "22", forRemoval = true)
+@Deprecated(since = "23", forRemoval = true)
 final public class SimpleSelector extends Selector {
 
     /**
@@ -192,6 +193,11 @@ final public class SimpleSelector extends Selector {
         // if id is not null and not empty, then match needs to check id
         this.matchOnId = (id != null && !("".equals(id)));
 
+    }
+
+    @Override
+    public Set<String> getClasses() {
+        return styleClassSet.stream().map(StyleClass::getStyleClassName).collect(Collectors.toUnmodifiableSet());
     }
 
     @Override public Match createMatch() {
--- a/modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
+++ b/modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
@@ -49,7 +49,9 @@ import static javafx.geometry.NodeOrientation.RIGHT_TO_LEFT;
  * A simple selector which behaves according to the CSS standard.
  *
  * @since 9
+ * @deprecated This class was exposed erroneously and will be removed in a future version
  */
+@Deprecated(forRemoval = true)
 final public class SimpleSelector extends Selector {
 
     /**
--- a/modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
+++ b/modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
@@ -51,7 +51,7 @@ import static javafx.geometry.NodeOrientation.RIGHT_TO_LEFT;
  * @since 9
  * @deprecated This class was exposed erroneously and will be removed in a future version
  */
-@Deprecated(forRemoval = true)
+@Deprecated(since = "22", forRemoval = true)
 final public class SimpleSelector extends Selector {
 
     /**
--- a/modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
+++ b/modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
@@ -196,7 +196,7 @@ final public class SimpleSelector extends Selector {
     }
 
     @Override
-    public Set<String> getClasses() {
+    public Set<String> getStyleClassNames() {
         return styleClassSet.stream()
             .map(StyleClass::getStyleClassName)
             .collect(Collectors.toUnmodifiableSet());
--- a/modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
+++ b/modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
@@ -197,7 +197,9 @@ final public class SimpleSelector extends Selector {
 
     @Override
     public Set<String> getClasses() {
-        return styleClassSet.stream().map(StyleClass::getStyleClassName).collect(Collectors.toUnmodifiableSet());
+        return styleClassSet.stream()
+            .map(StyleClass::getStyleClassName)
+            .collect(Collectors.toUnmodifiableSet());
     }
 
     @Override public Match createMatch() {
--- a/modules/javafx.graphics/src/main/java/javafx/css/CompoundSelector.java
+++ b/modules/javafx.graphics/src/main/java/javafx/css/CompoundSelector.java
@@ -62,7 +62,7 @@ import java.util.Set;
  * @since 9
  * @deprecated This class was exposed erroneously and will be removed in a future version
  */
-@Deprecated(forRemoval = true)
+@Deprecated(since = "22", forRemoval = true)
 final public class CompoundSelector extends Selector {
 
     private final List<SimpleSelector> selectors;
--- a/modules/javafx.graphics/src/main/java/javafx/css/CompoundSelector.java
+++ b/modules/javafx.graphics/src/main/java/javafx/css/CompoundSelector.java
@@ -60,7 +60,9 @@ import java.util.Set;
  * <code>Combinator.DESCENDANT</code>.
  *
  * @since 9
+ * @deprecated This class was exposed erroneously and will be removed in a future version
  */
+@Deprecated(forRemoval = true)
 final public class CompoundSelector extends Selector {
 
     private final List<SimpleSelector> selectors;
--- a/modules/javafx.graphics/src/main/java/javafx/css/CompoundSelector.java
+++ b/modules/javafx.graphics/src/main/java/javafx/css/CompoundSelector.java
@@ -102,9 +102,9 @@ final public class CompoundSelector extends Selector {
     }
 
     @Override
-    public Set<String> getClasses() {
+    public Set<String> getStyleClassNames() {
         return selectors.stream()
-            .map(Selector::getClasses)
+            .map(Selector::getStyleClassNames)
             .flatMap(Collection::stream)
             .collect(Collectors.toUnmodifiableSet());
     }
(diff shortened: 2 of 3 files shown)


nlisker (approved):

nlisker:
Should this deprecation happen in 22 already so that they can be removed in 23?

hjohn:
> Should this deprecation happen in 22 already so that they can be removed in 23?

That's the idea, but it needs to go in 23 first so it can be backported to 22 (Kevin mentioned as much in the linked ticket)

modules/javafx.graphics/src/main/java/javafx/css/SimpleSelector.java
```diff
@@ -49,7 +49,9 @@
  * A simple selector which behaves according to the CSS standard.
  *
  * @since 9
+ * @deprecated This class was exposed erroneously and will be removed in a future version
  */
+@Deprecated(forRemoval = true)
```
kevinrushforth, line 54:
```
    @Deprecated(since = "22", forRemoval = true)
```

kevinrushforth (commented):
Add the `since` parameter to the Deprecation annotation, but otherwise good.

andy-goryachev-oracle:
Should this change also introduce a public method 
`Selector.getStyleClasses()` and/or `Selector.getStyleClassesSet()`
due to
https://github.com/openjdk/jfx/pull/1333#issuecomment-1900085510 ?

hjohn:
> Should this change also introduce a public method `Selector.getStyleClasses()` and/or `Selector.getStyleClassesSet()` due to [#1333 (comment)](https://github.com/openjdk/jfx/pull/1333#issuecomment-1900085510) ?

Maybe yes, it depends on how important it is that SceneBuilder keeps working with newer and older FX releases

1. If SceneBuilder is tied to the same FX release, it just needs to be updated at the same time
2. If an older SceneBuilder must keep working with JavaFX 23, then we can't move those classes ever
3. If the requirement is only that it is compatible with one older version, then I guess we can introduce a new public method in 22.  It would mean that the update SceneBuilder would no longer work with JavaFX 21 or older versions.

andy-goryachev-oracle:
> 3\. the updated SceneBuilder would no longer work with JavaFX 21 or older versions.

or they (and other tools' developers) could use a multi-release jar https://openjdk.org/jeps/238
In any case, this change seems to have a greater impact than we thought originally.

jperedadnr:
Scene Builder releases are always tight to the JavaFX version (i.e, SB 21 uses JavaFX 21), so once JavaFX 23 is out with this change (or if we start testing 23-ea before the release of SB 23), we'll just have to modify the `CSSInternal` class as needed.

andy-goryachev-oracle:
> Scene Builder releases are always tight to the JavaFX version

Thank you for clarification!
Still, there might be other tools out there that we don't know of.

hjohn:
> > Scene Builder releases are always tight to the JavaFX version
> 
> Thank you for clarification! Still, there might be other tools out there that we don't know of.

That's always possible, but I can't find any. I overlooked SceneBuilder because it 

> > Scene Builder releases are always tight to the JavaFX version
> 
> Thank you for clarification! Still, there might be other tools out there that we don't know of.

That's always possible.  I couldn't find anything with an internet search, but closed source tools will not be found.  I missed Scene Builder, but only because I dismissed too quickly since it had `com.oracle.*` packages and thought it was part of FX.  I can't find anything else.  These classes are not accepted as parameters anywhere, or returned as types anywhere, nor can they be constructed by 3rd parties.  The only way to get at their methods is with an explicit cast, like Scene Builder does.  At a minimum we need to warn people not to use these.

nlisker:
Maybe *Scenic View*, as it does a lot of introspecting?

hjohn:
> Maybe _Scenic View_, as it does a lot of introspecting?

I just checked, no mention of `Selector` or `SimpleSelector` or `CompoundSelector` :)

kevinrushforth:
Since there is now no hurry to do this, I think we should proceed as follows:

1. Target this deprecation-for-removal enhancement, including adding the needed base class methods from the two subclasses that are being deprecated, to JavaFX 23. This would be done first, as soon as we settle on the new base class methods.
2. Proceed with [JDK-8322964](https://bugs.openjdk.org/browse/JDK-8322964) / PR #1316 .
3. Later, probably in JavaFX 24, proceed with [JDK-8323706](https://bugs.openjdk.org/browse/JDK-8323706) / PR #1333 to remove SimpleSelector and CompoundSelector from the public API (by moving them to an internal package).

hjohn:
Okay, so I added the method `Set<String> getClasses()`:

The reasoning is as follows:
- The ideal name `getStyleClasses` is currently in use by `List<String> getStyleClasses()`
   - Internally (for selectors), style classes are not a `List`, duplicates are removed, and the order is lost.  Exposing it as a `List` seems to offer guarantees we can't offer
- Perhaps the next best name `getStyleClassSet` is taken up by the old `Set<StyleClass> getStyleClassSet()` -- I can't use this as `StyleClass` is about to be phased out for performance reasons
- Mirroring what `Styleable` does with `ObservableList<String> getStyleClass()` is less than ideal (it is a `List`), however it does set some precedence for using a name like `getStyleClass` even though that is not a best practice for collections (it should be plural)
- CSS reference never really uses the term "style classes", just classes: https://www.w3schools.com/cssref/css_selectors.php

My conclusion is that we either need two rounds to get this right (to first free up our first choice `getStyleClasses`), or we can settle on one of: `getStyleClassNames`, `getStyles`, `getStyleNames`, `getClasses`, `getClassNames`.

kevinrushforth:
> My conclusion is that we either need two rounds to get this right (to first free up our first choice getStyleClasses), or we can settle on one of: getStyleClassNames, getStyles, getStyleNames, getClasses, getClassNames.

Let's settle on one of these names. I think `getClasses` is fine. My second choice would be `getClassNames`.

There is one other method used by SceneBuilder's `CssInternal` class, `CompoundSelector::getSelectors`. I think you need to add a method to the `Selector` base class that returns a `List<Selector>` instead of the `List<SimpleSelector>` returned by the existing `CompoundSelector` method. Maybe `getChildSelectors` or similar? That method would be overridden in the subclasses with `CompoundSelector` wrapping the existing selectors list in an unmodifiable list and `SimpleSelectors` returning an empty List.

hjohn:
> There is one other method used by SceneBuilder's `CssInternal` class, `CompoundSelector::getSelectors`. I think you need to add a method to the `Selector` base class that returns a `List<Selector>` instead of the `List<SimpleSelector>` returned by the existing `CompoundSelector` method. Maybe `getChildSelectors` or similar? That method would be overridden in the subclasses with `CompoundSelector` wrapping the existing selectors list in an unmodifiable list and `SimpleSelectors` returning an empty List.

@kevinrushforth if I'm correct, SceneBuilder won't need that anymore.  Since it shouldn't need to do an `instanceof` check (which will break once they're moved to internal packages), the fragment that gets all the selectors should be changed to:

        for (Rule r : s.getRules()) {
            for (Selector ss : r.getSelectors()) {
                styleClasses.addAll(ss.getClasses());   // provisional name getClasses
            }
        }

Old SceneBuilder code fragment:

        for (Rule r : s.getRules()) {
            for (Selector ss : r.getSelectors()) {
                if (ss instanceof SimpleSelector) {
                    SimpleSelector simple = (SimpleSelector) ss;
                    styleClasses.addAll(simple.getStyleClasses());
                } else {
                    if (ss instanceof CompoundSelector) {
                        CompoundSelector cs = (CompoundSelector) ss;
                        for (Selector selector : cs.getSelectors()) {
                            if (selector instanceof SimpleSelector) {
                                SimpleSelector simple = (SimpleSelector) selector;
                                styleClasses.addAll(simple.getStyleClasses());
                            }
                        }
                    }
                }
            }
        }

kevinrushforth:
@hjohn since the implementation of `getClasses` in `CompoundSelector` aggregates the style classes for all of its `SimpleSelectors`, then I think you are right that a `getChildSelectors` is unnecessary.

@jperedadnr Will this work for you?

jperedadnr:
Yes, I think so, simplifying the code without breaking the existing functionality should be fine. 
I guess it should be easy to do a build with this PR and test Scene Builder before and after the proposed change.

kevinrushforth:
> Yes, I think so, simplifying the code without breaking the existing functionality should be fine. I guess it should be easy to do a build with this PR and test Scene Builder before and after the proposed change.

In order to fully test it, you would need to locally make the change to `SceneBuilder` that John suggested.

jperedadnr:
Sure, that was what I meant. I'll give it a try over the weekend.

jperedadnr:
I've tested the change proposed in https://github.com/openjdk/jfx/pull/1340#issuecomment-1924508539 successfully: 
I've built locally this PR and Scene Builder with the proposed change. Then, showing the CSS analyzer, it calls `CSSInternal::getStyleClasses` using `Selector::getClasses`, which produces the same result as using the old implementation.

andy-goryachev-oracle:
> I think `getClasses` is fine. My second choice would be `getClassNames`.

Perhaps it might be `getStyleClasses()` to parallel one in `Node`?
(I don't know why Node's method is called `getStyleClass` (singular) when it clearly returns multiple Strings.)

andy-goryachev-oracle:
> need 2 passes, because there already is a method named like that:

oh yeah, sorry.
`getStyleClassesSet()` ?

hjohn:
> > need 2 passes, because there already is a method named like that:
> 
> oh yeah, sorry. `getStyleClassesSet()` ?

That one is still available yes :)

andy-goryachev-oracle (approved):
Looks good, with minor changes.
This also needs a CSR (which has already been requested).  
I'll re-approve once we all agree on the new method name.

kevinrushforth:
So we can add `getStyleClassesSet` to the list of possible method names. I think the word "class" should be in the name of the method, which excludes `getStyles` and `getStyleNames`, leaving the following:

```
getStyleClassNames
getClasses
getClassNames
getStyleClassesSet
```

Any of them seem OK to me. I don't have a strong preference.

kevinrushforth:
@hjohn I'm now leaning towards either `getStyleClassNames` or `getNames`, but it isn't a strong preference. Why don't you pick on, update the PR, and create a Draft CSR.

kevinrushforth (approved):

andy-goryachev-oracle (approved):
lgtm, thank you.

hjohn:
@kevinrushforth CSR was created

Maran23 (approved):
Looks good and reasonable to me as well. 
I can also confirm that I have never used both `Selector` classes or seen any use for them.

Also +1 for `getStyleClassNames`, feels the most natural to me.

PR #1342, opened 2024-01-19
8324219: Remove incorrect documentation from Animation methods

mstr2:
The `Animation` class states in the documentation of various methods that the method call would be asynchronous, using language similar to:

```
{@code stop()} is an asynchronous call, the {@code Animation} may not stop immediately.
```

This is factually wrong, there are no asynchronous calls. In fact, the methods in question can only be called synchronously on the JavaFX Application thread, and the state of the animation is changed immediately. For example, when `stop()` is called, the animation transitions to the stopped state instantly, without waiting for the next animation pulse.

---------
### Progress
- [ ] Change requires a CSR request matching fixVersion jfx23 to be approved (needs to be created)

### Issue
 * [JDK-8324219](https://bugs.openjdk.org/browse/JDK-8324219): Remove incorrect documentation from Animation methods (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1342/head:pull/1342` \
`$ git checkout pull/1342`

`$ git checkout pull/1342` \
`$ git pull https://git.openjdk.org/jfx.git pull/1342/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1342`

`$ git pr show -t 1342`

</details>

<a href="https://git.openjdk.org/jfx/pull/1342.diff">https://git.openjdk.org/jfx/pull/1342.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1342#issuecomment-1900701819)

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/animation/Animation.java
+++ b/modules/javafx.graphics/src/main/java/javafx/animation/Animation.java
@@ -883,14 +883,16 @@ public abstract class Animation {
      *
      * Note that unlike {@link #playFromStart()} calling this method will not
      * change the playing direction of this {@code Animation}.
+     * <p>
+     * This method must be called on the JavaFX Application thread.
      *
      * @param cuePoint
      *            name of the cue point
      * @throws NullPointerException
      *             if {@code cuePoint} is {@code null}
-     * @throws IllegalStateException
-     *             if embedded in another animation,
-     *                such as {@link SequentialTransition} or {@link ParallelTransition}
+     * @throws IllegalStateException if this method is called on a thread
+     *             other than the JavaFX Application Thread, or if embedded in another animation,
+     *             such as {@link SequentialTransition} or {@link ParallelTransition}
      * @see #getCuePoints()
      */
     public void playFrom(String cuePoint) {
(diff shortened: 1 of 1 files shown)


mstr2:
Since this is a documentation-only change, it might also go into jfx22.

kevinrushforth:
We need to resolve the mailing list discussion thread [X](URL) before deciding whether to proceed with this or not. If we keep the change to require animation to be done on the JavaFX application thread (meaning we reject the proposal put forward in that mailing list thread), then I agree that this doc change should be done, and backported to JavaFX 22.

/reviewers 2
/csr

kevinrushforth:
We need to resolve [this mailing list discussion thread](https://mail.openjdk.org/pipermail/openjfx-dev/2024-January/044734.html) before deciding whether to proceed with this or not. If we keep the change to require animation to be done on the JavaFX application thread (meaning we reject the proposal put forward in that mailing list thread), then I agree that this doc change should be done, and backported to JavaFX 22.

/reviewers 2
/csr

hjohn (approved):
I think this is a good change. All animation code always runs on the FX thread. Calling start/stop etc on a different thread would mess with flags like `aborted` and `inTimePulse` which are not marked volatile.  This change should probably have been made when the FX thread enforcing was added.

nlisker (approved):
Looks fine, but I'd like to see what Jurgen's case is on the mailing list.

PR #1351, opened 2024-01-26
8301900: TextArea: Committing text with ENTER in an IME window inserts newline

beldenfox:
In the Mac glass code the presence of "marked" text (which is tracked in the nsAttrBuffer) signals that an IME is active. In this state the current code assumes that when NSTextInputContext handles a `keyDown:` it will either generate a call to `insertText:replacementRange:` or one of the routines that manipulates the marked (composed) text. But this bug shows that sometimes the IME acts on the event without generating any calls back into glass at all.

In this PR the logic is simplified: if the NSTextInputContext handles the `keyDown:` and there's marked (composed) text we don't generate a KeyEvent. Otherwise we do. This PR removes the `shouldProcessKeyEvent` flag since it no longer assumes we can catch callbacks to update it correctly.

The existing code also assumes that the composition phase ends when the NSTextInputContext calls `insertText:replacementRange` to commit the text. This is true but if the user tries to use a dead-key sequence to generate a non-existent character (like an accented 'q') the context will call `insertText` twice while handling one key down event. In that case we can't exit the composition mode upon seeing the first `insertText` call since it will cause us to mis-handle the second one. This PR defers exiting composition mode until the end of `keyDown:`.

I also updated a few deprecated constants so this file no longer generates compiler warnings.


kevinrushforth:
Reviewers: @andy-goryachev-oracle @kevinrushforth 

/reviewers 2

kevinrushforth (approved):
Code changes look good. All my testing looks good.

beldenfox:
I need to amend this PR. If the IM is disabled in the middle of composition we don't dismiss the IM window. In the past that was sort of OK because we ignored the enabled stated and kept sending events to the NSTextInputContext. With this PR we honor the state and stop talking to the context so the IM window just freezes and never goes away. I'll be adding a line of code to ensure we get rid of the IM window when the enabled state changes (which we should have been doing all along).

This condition can arise if the user starts composing and then clicks on a control that doesn't accept IM input (like a button). It can also happen with [JDK-8087477](https://bugs.openjdk.org/browse/JDK-8087477) which is a bug I'm still investigating.

andy-goryachev-oracle:
... and scenarios described in JDK-8088172 and JDK-8089803 seem to have been fixed.

modules/javafx.graphics/src/main/native-glass/mac/GlassView3D.m
```diff
@@ -511,11 +509,20 @@ - (void)keyDown:(NSEvent *)theEvent
...
+    didCommitText = NO;
+    BOOL inputContextHandledEvent = (imEnabled && [self.inputContext handleEvent:theEvent]);
+    handlingKeyEvent = NO;
+
+    if (didCommitText) {
+        // Exit composition mode
+        didCommitText = NO;
+        nsAttrBuffer = [nsAttrBuffer initWithString: @""];
+    }
+    else if (!inputContextHandledEvent || (nsAttrBuffer.length == 0)) {
```
andy-goryachev-oracle, line 522: `} else if {` ?

beldenfox:
@andy-goryachev-oracle 

> @beldenfox are you going to update this PR with a fix for dismissing the IME window you alluded in the previous comment?

I did that with the most recent commit (Jan 27th).

> * click elsewhere.  the IME window disappears

That indicates that you got my most recent commit, otherwise the IME window would not disappear immediately.

I'm really glad you caught this. The old code for disabling the IME just called `[self unmarkText]`. I added the code for dismissing the IME window but didn't notice that `unmarkText` doesn't seem to work. Currently it commits a zero-length string (which appears to be a no-op) and leaves the composition buffer untouched. That combination doesn't make much sense and doesn't match Apple's spec. I need to dig around in the code history to see why it's behaving this way. This also explains some of the bad behavior you noticed in JDK-8320912.

beldenfox:
> Currently it commits a zero-length string (which appears to be a no-op) and leaves the composition buffer untouched.

Correcting myself: it does clear the composition buffer. But the zero-length commit doesn't seem to be doing what it needs to.

beldenfox:
@andy-goryachev-oracle The behavior you're seeing when clicking on another control is a manifestation of https://bugs.openjdk.org/browse/JDK-8320912. I put together a fix for that bug but decided not to roll it into this PR. If it would help testing I could do that.

The current code to disable the IME happens *after* focus has already moved to the node that doesn't want IME input. So the commit that's fired off by `unmarkText` targets the wrong node. The fix is to implement finishInputMethodComposition so we do the cleanup before the focus changes.

andy-goryachev-oracle:
> [JDK-8320912](https://bugs.openjdk.org/browse/JDK-8320912). I put together a fix for that bug but decided not to roll it into this PR.

Are these two bugs independent?  Should JDK-8320912 be resolved first?

beldenfox:
> > [JDK-8320912](https://bugs.openjdk.org/browse/JDK-8320912). I put together a fix for that bug but decided not to roll it into this PR.
> 
> Are these two bugs independent? Should JDK-8320912 be resolved first?

They are independent. This bug has to do with how we divide keystrokes between KeyEvents and InputMethodEvents. JDK-8320912 is about cleaning up during focus changes. The order in which they get resolved doesn't matter.

andy-goryachev-oracle:
> The order in which they get resolved doesn't matter.

would it be possible to get [JDK-8320912](https://bugs.openjdk.org/browse/JDK-8320912) integrated first then?  It seems the  scenario it addresses is narrower, and it will be easier to review this bug.

beldenfox:
@andy-goryachev-oracle I submitted PR #1356.

andy-goryachev-oracle:
> I submitted PR

thank you!

kevinrushforth:
Once #1356 is integrated I'll re-review.

kevinrushforth:
I'm surprised Skara hasn't noticed that there is a merge conflict -- when it does, it should add a `merge-conflict` label. Maybe it needs something other than a PR comment to wake it up?

beldenfox:
Skara has noted the merge conflict. I'll get that patched up. I also need to tweak the implementation of setEnabled, it's trying to handle focus moving from node to node but the logic is wrong.

kevinrushforth:
Skara hasn't labeled it, so I'm restating the requirement of 2 reviewers to see if that wakes up the bot to check for the merge conflict.

/reviewers 2

beldenfox:
Two additions to the code. I was not handling the case where the keystroke didn't commit text but did clear the marked text (getting us out of composition mode). You can do this by repeatedly pressing ESC.

When the enabled state changes we shouldn't have any marked text but if we do it should just be discarded. The old code was trying to deal with focusOwner changes but that cleanup has to happen before the enabled state changes instead of after (which is why finishInputMethodComposition exists).

andy-goryachev-oracle (approved):
As far as I can tell, the new behavior matches the native app one.

有難う

kevinrushforth (approved):

PR #1352, opened 2024-01-26
8324658: Allow animation play/start/stop/pause methods to be called on any thread

nlisker:
Added a utility method to run code on the FX thread if it's not already, and changed the animation methods to use it.

---------
### Progress
- [x] Change requires CSR request [JDK-8324796](https://bugs.openjdk.org/browse/JDK-8324796) to be approved

### Issues
 * [JDK-8324658](https://bugs.openjdk.org/browse/JDK-8324658): Allow animation play/start/stop/pause methods to be called on any thread (**Enhancement** - P3)
 * [JDK-8324796](https://bugs.openjdk.org/browse/JDK-8324796): Allow animation play/start/stop/pause methods to be called on any thread (**CSR**)

### Reviewers
 * [John Hendrikx](https://openjdk.org/census#jhendrikx) (@hjohn - Committer) 🔄 Re-review required (review applies to [62716493](https://git.openjdk.org/jfx/pull/1352/files/62716493a8d7daa5bd9140470d0eb7e8336dbf78))

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1352/head:pull/1352` \
`$ git checkout pull/1352`

`$ git checkout pull/1352` \
`$ git pull https://git.openjdk.org/jfx.git pull/1352/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1352`

`$ git pr show -t 1352`

</details>

<a href="https://git.openjdk.org/jfx/pull/1352.diff">https://git.openjdk.org/jfx/pull/1352.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1352#issuecomment-1912853936)


nlisker:
The tests from [JDK-8159048](https://bugs.openjdk.org/browse/JDK-8159048) should either be removed or reworked. @kevinrushforth Should I just check that calling the specified methods both from the FX thread and from a background thread end up executing on the FX thread?

Running
`./gradlew :systemTests:test --tests test.com.sun.javafx.animation.AnimationTest` still succeeds for me for some reason even after a clean. Is there something missing here?

kevinrushforth:
> The tests from [JDK-8159048](https://bugs.openjdk.org/browse/JDK-8159048) should either be removed or reworked. @kevinrushforth Should I just check that calling the specified methods both from the FX thread and from a background thread end up executing on the FX thread?

Checking that the animation always runs on the FX thread seems useful. It might also be worth adding a test based on the original reproducer in [JDK-8159048](https://bugs.openjdk.org/browse/JDK-8159048).
 
> Running `./gradlew :systemTests:test --tests test.com.sun.javafx.animation.AnimationTest` still succeeds for me for some reason even after a clean. Is there something missing here?

The system tests are only run if `-PFULL_TEST=true` is passed to gradle.

kevinrushforth:
@jperedadnr Do you want to be the second reviewer?

kevinrushforth:
I did a first pass over the doc and implementation changes, and they look like what I would expect. I'll create a placeholder CSR and you can fill it in. Because of the short time, let's try to get it moved to Finalized on Monday if possible.

modules/javafx.graphics/src/main/java/javafx/animation/Timeline.java
```diff
@@ -191,11 +191,9 @@ void doStart(boolean forceSync) {
         clipCore.start(forceSync);
     }
 
-    /**
-     * {@inheritDoc}
-     */
     @Override
-    public void stop() {
+    void stopOnFxThread() {
+        // The parent check is redone in the super method. Consider refactoring.
```
kevinrushforth, line 196: As you've probably already seen, this causes infinite recursion and an eventual StackOverflow. The easiest (and probably safest) fix is to revert this part of the change, keeping the overridden `stop` method, and have it call a private `stopOnFxThread` method (and make the one in the parent class also private). You can file a bug to refactor it later.

nlisker: I just pushed a fix that calls the correct super method. The implementation isn't fully tested yet because I wanted to get the CSR moving in parallel.

I'm still thinking on the least intrusive way to write the tests. I seems that the only way to check if a method runs on the FX thread is to override a method on its stack and check there. That means that I would need to expose (package-private) the `play/start/stop/pauseOnFxThread` methods.

kevinrushforth: It works now with change to call the right super method.

As for the tests, I'd sooner not do anything intrusive in this PR, especially given the tight window to get this in. I'm not even sure whether it's worth inspecting the internals to make sure it is on the right thread in all cases, but if so, a follow-up testbug could be filed for that.

modules/javafx.graphics/src/main/java/javafx/animation/Animation.java
```diff
@@ -976,18 +976,16 @@ public void playFromStart() {
      *  animation.play();<br>
      * </code>
      * <p>
-     * Note: <ul>
-     * <li>{@code play()} is an asynchronous call, the {@code Animation} may not
-     * start immediately. </ul>
-     * <p>
-     * This method must be called on the JavaFX Application thread.
+     * Note: if this method is not called on the JavaFX Application Thread, it is delegated to it automatically.
```
kevinrushforth, line 979: We need to leave in the note about it being asynchronous and that that that animation may not start immediately. I recommend to revert that part of the change here and elsewhere.

kevinrushforth (commented):
I think the three `playFrom*` methods need the same doc and implementation change to delegate to the FX app thread to avoid any potential problems.

I also recommend reverting the doc changes around not executing immediately (it is even more likely to be true in some cases with this change).

modules/javafx.graphics/src/main/java/javafx/animation/Animation.java
nlisker: The call is asynchronous (and the animation might not start immediately) only when called from a background thread, so I propose:

```
Note: if this method is not called on the JavaFX Application Thread, it is delegated to it automatically.
In this case, the call is asynchronous and may not happen immediately.
```

kevinrushforth: That seems fine.

hjohn (approved):
This looks good. It seems most non-property methods (aside from `jumpTo`) have this functionality now.

modules/javafx.graphics/src/main/java/javafx/animation/Animation.java
```diff
@@ -976,18 +992,17 @@ public void playFromStart() {
...
      */
     public void play() {
-        Toolkit.getToolkit().checkFxUserThread();
+        Utils.runOnFxThread(this::playOnFxThread);
+    }
+
+    private void playOnFxThread() {
         if (parent != null) {
             throw new IllegalStateException("Cannot start when embedded in another animation");
         }
```
hjohn, line 1008: Hm, this is going to look really ugly if the caller wasn't on the FX thread.  The exception won't be thrown, it seems to work, but once run the FX thread, it will just log this instead of breaking the callers flow.

hjohn: I suppose you can use this as a "trick" to test if the code to shunt it to the FX thread works (if you don't get the exception immediately, it worked...)

kevinrushforth: Perhaps a better solution is to move this (and similar) checks back to the `play` method, add the checks to the `playFrom*` methods, and do the same for `pause` and `stop`.

kevinrushforth (commented):
The API looks good now, so you should be able to update the CSR.

nlisker:
> The API looks good now, so you should be able to update the CSR.

I think that the update to the CSR from an hour ago included all the changes already.

kevinrushforth:
> I think that the update to the CSR from an hour ago included all the changes already.

I missed that. Thanks.

kevinrushforth:
> The API looks good now...

As for the implementation, the only two things I see that are needed are:
1. Move the checks for `parent != null ` back to the original methods (i.e., before delegating). Without this, we are not conforming the spec of the `@throws` clause, since the exception isn't thrown back to the caller.
2. Fix or remove the failing system test. Unless you add it as part of this PR, I recommend filing a follow-up issue to add a system test based on the failing test program that is attached to [JDK-8159048](https://bugs.openjdk.org/browse/JDK-8159048).

jperedadnr (commented):
Changes look fine. I have two minor comments.

modules/javafx.graphics/src/main/java/com/sun/javafx/util/Utils.java
```diff
@@ -997,4 +999,17 @@ public static synchronized void loadNativeSwingLibrary() {
             return null;
         });
     }
+
+    /**
+     * Ensures that a code segment is run on the FX thread.
+     *
+     * @param runnable a {@code Runnable} encapsulating the code
+     */
+    public static void runOnFxThread(Runnable runnable) {
```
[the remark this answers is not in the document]
nlisker: I thought about it when I wrote this method, but I don't see much value in it for an internal utility method as it doesn't serve as a boundary. In addition, there was a discussion about future work that will eliminate the need for this method by dealing with the threading issue at a deeper level, so it is somewhat temporary.

modules/javafx.graphics/src/main/java/javafx/animation/Animation.java
```diff
@@ -894,6 +897,10 @@ public void jumpTo(String cuePoint) {
      * @see #getCuePoints()
      */
     public void playFrom(String cuePoint) {
+        Utils.runOnFxThread(() -> playFromOnFxThread(cuePoint));
+    }
+
+    private void playFromOnFxThread(String cuePoint) {
```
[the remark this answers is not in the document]
nlisker: > However, I'm not sure about the method naming `OnFxThread`, as it might imply that what it does is already done on the FX thread, therefore not needing to be wrapped up by `runOnFxThread`.

I understand what you mean, but the way I see it is that they do already run on the FX thread because that's the only reason they exist. I could use something like `runMustBeOnFxThread` or `runOnlyFromFxThread` if that's better.
Note that there are already `do___` methods (bad names IMO), and adding `run___` might be confusing.

tests/system/src/test/java/test/com/sun/javafx/animation/SynchronisityTest.java
```diff
@@ -0,0 +1,90 @@
+package test.com.sun.javafx.animation;
```
jperedadnr, line 1: Add license

tests/system/src/test/java/test/com/sun/javafx/animation/AnimationTimerTest.java
```diff
@@ -1,112 +1,34 @@
-/*
```
jperedadnr, line 1: Don't remove license

tests/system/src/test/java/test/com/sun/javafx/animation/AnimationTest.java
```diff
@@ -1,116 +1,30 @@
-/*
```
jperedadnr, line 1: Don't remove license

modules/javafx.graphics/src/main/java/javafx/animation/Animation.java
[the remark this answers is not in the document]
jperedadnr: Probably you won't like `xxImpl` either?
I'm not sure we should start adding `runMustBeOnFxThread` to every method name out there that should run on the FX thread. In this case, maybe we could simply add a small javadoc with a comment, and that's it. Neither will enforce a real check about that.

tests/system/src/test/java/test/com/sun/javafx/animation/SynchronisityTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+    /**
+     * Number of seconds to wait for a failure. If an exception is not thrown in this time, it's assumed it won't be
+     * thrown later too.
+     */
+    private static final int GRACE_PERIOD = 15;
+
+    private AtomicBoolean failed = new AtomicBoolean(false);
+    private CountDownLatch waiter = new CountDownLatch(1);
+    private ExecutorService executor = Executors.newCachedThreadPool();
```
jperedadnr, line 53: These can be `final`

tests/system/src/test/java/test/com/sun/javafx/animation/SynchronisityTest.java
```diff
@@ -0,0 +1,90 @@
...
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+
+import javafx.application.Application;
+import javafx.application.Platform;
+import javafx.stage.Stage;
+import test.util.Util;
+
+// Based on https://bugs.openjdk.org/browse/JDK-8159048
+public class SynchronisityTest extends Application {
```
jperedadnr, line 20: shouldn't it be `SynchronicityTest`?

tests/system/src/test/java/test/com/sun/javafx/animation/AnimationTest.java
```diff
@@ -1,116 +1,30 @@
...
+        while (true) {
+            // Initialize the concurrent starts and stops
+            anim.play();
+            sleep(10);
+            anim.stop();
+            sleep(10);
+        }
     }
-}
+}
```
kevinrushforth, line 30: Minor: add newline at end of file.

tests/system/src/test/java/test/com/sun/javafx/animation/SynchronisityTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+    public static void main(String[] args) {
+        Application.launch(args);
+    }
```
kevinrushforth, line 24: The main method is not needed.

tests/system/src/test/java/test/com/sun/javafx/animation/SynchronisityTest.java
```diff
@@ -0,0 +1,90 @@
...
+    private Throwable throwable;
+
+    protected void runTest(Runnable runnable) throws InterruptedException {
+        Platform.runLater(() -> registerExceptionHandler());
+
+        for (int i = 0; i < 10; i++) {
+            executor.submit(runnable);
+        }
+
+        waiter.await(GRACE_PERIOD, TimeUnit.SECONDS);
```
kevinrushforth, line 65: I don't think this will do what you want. This will return almost immediately, before any of the animation has run (as soon as the runnable that sets the uncaught exception handler finished). What I think you need is something more like this:

```
        Platform.runLater(() -> registerExceptionHandler());
        assertTrue(waiter.await(GRACE_PERIOD, TimeUnit.SECONDS));

        // start the 10 threads...

        Util.sleep(GRACE_PERIOD);
```

The await ensures that the uncaught exception handler has been registered before starting the threads. The sleep then runs the threads for a fixed amount of time.

kevinrushforth (commented):
I left a few more inline comments, mostly on the test, but I also noted some missing `parent != null` checks.

modules/javafx.graphics/src/main/java/javafx/animation/Animation.java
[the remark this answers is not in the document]
nlisker: Changed to `xxImpl` with a doc comment.

tests/system/src/test/java/test/com/sun/javafx/animation/SynchronisityTest.java
nlisker: > I don't think this will do what you want. This will return almost immediately, before any of the animation has run (as soon as the runnable that sets the uncaught exception handler finished).

I observed that this setup works correctly for the `AnimationTimer` test. The test does wait for `GRACE_PERIOD` seconds when there is no exception and then succeeds, or catches and fails immediately when there onw is thrown. The `Animation` test was more finicky, but I think that is because I'm not simulating a processing load properly.

> The await ensures that the uncaught exception handler has been registered before starting the threads. 

While technically true, the handler register in a negligible amount of time compared to the grace period. Because exceptions are thrown continuously )it's not a one-time event), the handler will never miss a failure.

I will change it for correctness sake.

modules/javafx.graphics/src/main/java/javafx/animation/Animation.java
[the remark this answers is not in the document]
kevinrushforth: Thanks. You missed the one in Timeline (which is why there is a compilation error).

modules/javafx.graphics/src/main/java/javafx/animation/Animation.java
```diff
@@ -938,17 +958,22 @@ public void playFrom(Duration time) {
...
      *                such as {@link SequentialTransition} or {@link ParallelTransition}
      */
     public void playFromStart() {
+        if (parent != null) {
+            throw new IllegalStateException("Cannot start when embedded in another animation");
+        }
+        Utils.runOnFxThread(this::playFromStartOnFxThread);
+    }
+
+    private void playFromStartOnFxThread() {
```
kevinrushforth, line 976: Maybe change this (and the other `playFrom* methods`) to `xxxImpl` as well?

nlisker:
The tests are a best-effort to reproduce the issue and may pass even before the fix.

I found that the `AnimationTimer` test cases can fail in 2 ways: one is the reported NPE on the FX thread, but another is an AIIOB on the background thread that originates in (or close to) the master timer. For this reason, I added exception handlers on the FX and on each background thread.

The `Animation` test is more unstable, maybe because it doesn't simulate a load on every tick (like in `AnimationTimer::handle()`). Maybe using a `Timer` implementation instead of a `Transition` will allow to reproduce the issue better.

tests/system/src/test/java/test/com/sun/javafx/animation/AnimationTimerTest.java
```diff
@@ -25,88 +25,36 @@
...
-        assertFalse(Platform.isFxApplicationThread());
-        AnimationTimer timer = new AnimationTimer() {
-            @Override public void handle(long l) {
-                assertThrows(IllegalStateException.class, () -> stop());
+        var timer = new AnimationTimer() {
+
+            @Override
+            public void handle(long now) {
+                // Simulate intensive processing
+                Util.sleep(10);
```
kevinrushforth, line 48: Would this be more likely to catch a problem if this sleep (which is on the FX thread) were reduced to 1 msec? Not sure, but since the timer is started and stopped every 10 msec, it might be worth looking at. On the other hand, it does detect the bug on my Mac, so it might be fine as is.

kevinrushforth (approved):
The updated test looks fine (I did leave one question, but it's not important).

I modified the `runOnFxAppThread` method to always run on the calling thread (this reintroducing the original problem), and confirmed that AnimationTimerTest fails (thus catching the bug), but `AnimationTest` still passes, which matches what you are seeing. That might be OK, since the original bug report was on AnimationTimer.

kevinrushforth:
One more thing I tried: I reverted the changes to the animation implementation, going back to the current JavaFX 22 behavior of throwing an exception if start, stop, etc is called on a background thread, and the tests still pass.

It looks like the executor service is catching the exception, which is interfering with using an uncaught exception to flag an error. Instead of registering an uncaught exception handler for the test thread, you might wrap the runnable in another runnable with a try / catch and set the "failed" flag, etc., from there?

tests/system/src/test/java/test/com/sun/javafx/animation/SynchronizationTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private Thread thread;
+    private Throwable throwable;
+
+    protected void runTest(Runnable runnable) throws InterruptedException {
+        Platform.runLater(() -> registerExceptionHandler());
+
+        for (int i = 0; i < 10; i++) {
+            executor.submit(runnable);
+        }
```
kevinrushforth, line 87: I did a quick prototype of the idea I mentioned in my last comment, and it looks like it works:

```
        Runnable wrappedRunnable = () -> {
            try {
                runnable.run();
            } catch (Throwable e) {
                thread = Thread.currentThread();
                throwable = e;
                failed.set(true);
                waiter.countDown();
            }
        };

        for (int i = 0; i < 10; i++) {
            executor.submit(wrappedRunnable);
        }
```

This obviates the need for the `registerExceptionHandler` calls in the two tests themselves (this one in _this_ method is still needed).

kevinrushforth (changes requested):

tests/system/src/test/java/test/com/sun/javafx/animation/SynchronizationTest.java
```diff
@@ -0,0 +1,104 @@
...
+     * thrown later too.
+     */
+    private static final int GRACE_PERIOD = 15;
+
+    final private AtomicBoolean failed = new AtomicBoolean(false);
+    final private CountDownLatch waiter = new CountDownLatch(1);
+    final private ExecutorService executor = Executors.newCachedThreadPool();
+
+    private Thread thread;
+    private Throwable throwable;
```
kevinrushforth, line 80: Since these can be set from any thread, maybe use AtomicObject?

tests/system/src/test/java/test/com/sun/javafx/animation/AnimationTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2024, Oracle and/or its affiliates. All rights reserved.
```
jperedadnr, line 2: Copyright header should keep 2023, even if it is almost completely changed

tests/system/src/test/java/test/com/sun/javafx/animation/AnimationTimerTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2024, Oracle and/or its affiliates. All rights reserved.
```
jperedadnr, line 2: Same, copyright header should keep 2023, even if it is almost completely changed

modules/javafx.graphics/src/main/java/javafx/animation/Animation.java
```diff
@@ -938,17 +968,27 @@ public void playFrom(Duration time) {
...
+        if (parent != null) {
+            throw new IllegalStateException("Cannot start when embedded in another animation");
+        }
+        Utils.runOnFxThread(this::playFromStartImpl);
+    }
+
+    /**
+     * This method must be run on the JavaFX Application Thread.
+     *
+     * @see #playFromStartImpl()
```
jperedadnr, line 989: Do you mean `playFromStart()`?

modules/javafx.graphics/src/main/java/javafx/animation/Animation.java
```diff
@@ -894,6 +897,18 @@ public void jumpTo(String cuePoint) {
...
+        if (parent != null) {
+            throw new IllegalStateException("Cannot start when embedded in another animation");
+        }
+        Utils.runOnFxThread(() -> playFromImpl(cuePoint));
+    }
+
+    /**
+     * This method must be run on the JavaFX Application Thread.
+     *
+     * @see #playFromStartImpl(String)
```
jperedadnr, line 909: there is no such symbol. Do you mean `playFrom(String)`?

modules/javafx.graphics/src/main/java/javafx/animation/Animation.java
```diff
@@ -923,6 +941,18 @@ public void playFrom(String cuePoint) {
...
+        if (parent != null) {
+            throw new IllegalStateException("Cannot start when embedded in another animation");
+        }
+        Utils.runOnFxThread(() -> playFromImpl(time));
+    }
+
+    /**
+     * This method must be run on the JavaFX Application Thread.
+     *
+     * @see #playFromStartImpl(Duration)
```
jperedadnr, line 953: there is no such symbol. Do you mean `playFrom(Duration)`?

modules/javafx.graphics/src/main/java/javafx/animation/Animation.java
nlisker: Yes, there are copy-paste mistakes.

tests/system/src/test/java/test/com/sun/javafx/animation/AnimationTest.java
nlisker: Even if it wasn't released?

tests/system/src/test/java/test/com/sun/javafx/animation/SynchronizationTest.java
nlisker: Alright, but because of the latch I don't think it will ever matter in practice.

tests/system/src/test/java/test/com/sun/javafx/animation/SynchronizationTest.java
nlisker: I don't think the executor service catches the exception because if it did then the `AnimationTimer` test would also always pass. I do get the AIIOB exceptions from the background thread caught in the handler (which fails the test). This test is rather stable for me. The `Animation` one is more problematic, sometimes failing and sometimes not when it should fail.

tests/system/src/test/java/test/com/sun/javafx/animation/AnimationTest.java
kevinrushforth: Copyrights in source code aren't about whether it is released, so probably best in this case to keep the 2023.

tests/system/src/test/java/test/com/sun/javafx/animation/SynchronizationTest.java
kevinrushforth: Based on the testing I did, the executor service does prevent at least some exceptions that happen on the test thread from reaching the uncaught exception handler. An explicit try/catch will catch those.

nlisker: Alright, will implement the change.

tests/system/src/test/java/test/com/sun/javafx/animation/SynchronizationTest.java
kevinrushforth: Thanks. I agree that in practice it won't matter.

kevinrushforth (approved):
The updated test looks good. I verified that it will correctly fail if [JDK-8159048](https://bugs.openjdk.org/browse/JDK-8159048) is reintroduced, and will also fail without the fix for _this_ issue (that is, with the current jfx master).

jperedadnr (approved):
Looks good!

PR #1359, opened 2024-02-07
8349373: Support JavaFX preview features

mstr2:
This PR contains a definition of preview features for JavaFX, as well as a helper class to verify that an application has opted into preview features.

---------
### Progress
- [x] Change requires CSR request [JDK-8352083](https://bugs.openjdk.org/browse/JDK-8352083) to be approved

### Issues
 * [JDK-8349373](https://bugs.openjdk.org/browse/JDK-8349373): Support JavaFX preview features (**Enhancement** - P4)
 * [JDK-8352083](https://bugs.openjdk.org/browse/JDK-8352083): Support JavaFX preview features (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1359/head:pull/1359` \
`$ git checkout pull/1359`

`$ git checkout pull/1359` \
`$ git pull https://git.openjdk.org/jfx.git pull/1359/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1359`

`$ git pr show -t 1359`

</details>

<a href="https://git.openjdk.org/jfx/pull/1359.diff">https://git.openjdk.org/jfx/pull/1359.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1359#issuecomment-2634423447)
</details>


PREVIEW-FEATURES.md
```diff
@@ -0,0 +1,45 @@
+# Preview features
```
andy-goryachev-oracle, line 1: is this a good location of this document?
should it be in /doc-files somewhere?

PREVIEW-FEATURES.md
```diff
@@ -0,0 +1,45 @@
...
+
+1. Identify all relevant API elements of the feature, and annotate them with `@Deprecated`.
+   The `since` element of the annotation must be set to the same value as the corresponding `@since`
+   javadoc tag of the API element.
+2. Add the following javadoc tag to each of the previously deprecated API elements:<p>
+   `@deprecated This is a preview feature which may be changed or removed in a future release.`
+3. Add a new constant to the `com.sun.javafx.PreviewFeature` enumeration, and choose a human-readable name
+   for the preview feature. This name will be used in warning and error messages when the preview feature
+   is used by application developers.
+4. Add runtime checks in appropriate places by invoking `com.sun.javafx.PreviewFeature.<FEATURE>.checkEnabled()`.
```
andy-goryachev-oracle, line 45: what do you think about adding this checklist to the PreviewFeature class?
possibly following by an example (to allow for copy-paste)?

PREVIEW-FEATURES.md
```diff
@@ -0,0 +1,45 @@
...
+
+2. **Not experimental**: A preview feature must not be experimental, risky, incomplete, or unstable.
+   For the purpose of comparison, if an experimental feature is considered 25% "done", then a preview
+   feature should be at least 90% "done".
+
+3. **Opt-in:** The use of preview features must be explicitly enabled by application developers by
+   setting a system property. The implementation must detect the opt-in, and fail at runtime when the
+   application has not opted into the use of preview features. All preview features have equal status
+   in any given JavaFX release and can not be enabled individually.
+
```
andy-goryachev-oracle, line 32: do you want to talk about similarities and differences between this and jdk preview features https://openjdk.org/jeps/12 ?

modules/javafx.base/src/main/java/com/sun/javafx/PreviewFeature.java
```diff
@@ -0,0 +1,79 @@
...
+import com.sun.javafx.runtime.VersionInfo;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Using a preview feature requires an opt-in from application developers by specifying the
+ * {@code javafx.enablePreview=true} system property. This class verifies that the application
+ * has opted into preview features.
+ */
+public enum PreviewFeature {
```
andy-goryachev-oracle, line 37: Why enum?
Maybe use a regular class with a public constructor should do it.
What do you think?

modules/javafx.base/src/main/java/com/sun/javafx/PreviewFeature.java
```diff
@@ -0,0 +1,79 @@
...
+        } else if (!suppressWarning && enabledFeatures.add(this)) {
+            System.err.printf("""
+                Note: This program uses the following preview feature of JavaFX %s: %s
+                      Preview features may be removed in a future release, or upgraded to permanent features of JavaFX.
+                      This warning can be disabled with the following system property: -D%s=true
+                """, VersionInfo.getVersion(), featureName, SUPPRESS_WARNING_PROPERTY);
+        }
+    }
+
+    private final String featureName;
```
andy-goryachev-oracle, line 78: would it be possible to move field to the beginning of the class please?

PREVIEW-FEATURES.md
mstr2: This definition is, as you point out, largely lifted from JEP 12. Aside from wording changes, here are some key differences:

* JEP 12 assumes that a feature should be finished within two releases. I think for JavaFX, being just a framework and not a programming language, it is sensible to assume that we can finish features in the next release; that is, a preview feature that is introduced in JavaFX N should be finished in N+1.
* JEP 12 says that a preview feature should be at least 95% "done". I think we can be a little bit more risk-tolerant here, which is why I lowered this number to 90% "done".

The general idea is that preview features in JavaFX should be faster and accept a little bit more risk than preview features in the JDK.

andy-goryachev-oracle: The space of people who know what a JDK preview is is larger than that of JFX people.  I think we might want to point to JEP12 and explicitly explain the difference, that's all.

mstr2: We could also do that, but the problem is that the contents of JEP 12 only map to JavaFX in some parts, but not in others. So we would need to include weasel language like "where applicable" (which always leaves a bit of room for interpretation), or provide what would basically be a detailed sentence-by-sentence diff.

So in the end, it might still be easier to point at our own (complete) definition instead of a partially applicable definition somewhere else.

PREVIEW-FEATURES.md
mstr2: I think it's not a documentation of JavaFX itself, but more like a meta-documentation like our contribution guidelines.

modules/javafx.base/src/main/java/com/sun/javafx/PreviewFeature.java
mstr2: The idea is to make it easy and obvious to use from preview code:
```java
PreviewFeature.MY_PREVIEW_FEATURE.checkEnabled();
```
In addition to that, any IDE can easily list all uses of an enum constant, which gives you an overview where APIs of a particular feature are located. After all, enums are just regular classes with singleton instances.

By the way, preview features can't be individually enabled or disabled. The point of checking on a enum constant is to generate an appropriate error message (including the name of the feature) in case the user didn't enable preview features.

andy-goryachev-oracle: my only objection is that enum implies the use of a switch pattern, which is not really applicable here.

kevinrushforth (commented):
Overall, I like the direction of this. This isn't a review (yet), but just a few high-level comments (I also left a couple inline comments).

As with Incubating Modules, the goal is to provide features that are not yet final, and are able to evolve before being finalized. To achieve this without surprising application developers and end users, we need to provide tools to allow developers and end users to know that they are using preview features that might work differently or not be available in subsequent versions of JavaFX. I think it generally does that, although I'll review it in more detail later.

* I recommend a Draft PR with an example preview feature. That would also be a good place to host the JEP document (for now). You can see what I did in PR #1617 for Incubator Modules.
* One of the important tenets of this is that the application developer needs to "opt in" in an obvious way. Since we can't have javac and java launcher support for this, your idea of a system property seems a good one. The system property should be read as early as possible to encourage passing it on the command line to make it more obvious (similarly our docs should show examples of running with `java -Djavafx.enablePreview=...` rather than a call to `System.setProperty`). Initializing the `PreviewFeature` class from the static initializer of `PlatformImpl` and `LauncherImpl` (as I did when adding the security manager check) might be helpful to this end.
* Should the property be an integer (feature release) rather than boolean? Without some sort of annotation, I'm not sure it would help, but I'll throw it out there
* I don't think the warning should be suppressible; the equivalent JDK warning isn't
* Is there anything we could do with annotations that might be useful? Not sure on this one

PREVIEW-FEATURES.md
kevinrushforth: > So in the end, it might still be easier to point at our own (complete) definition instead of a partially applicable definition somewhere else.

Even if this ends up being the way we go, a pointer to JEP 12 and list of key differences would be helpful.

modules/javafx.base/src/main/java/com/sun/javafx/PreviewFeature.java
```diff
@@ -0,0 +1,79 @@
...
+    PreviewFeature(String featureName) {
+        this.featureName = featureName;
+    }
+
+    private final String featureName;
+
+    private static final String ENABLE_PREVIEW_PROPERTY = "javafx.enablePreview";
+    private static final String SUPPRESS_WARNING_PROPERTY = "javafx.suppressPreviewBanner";
+
+    private static final boolean enabled = Boolean.getBoolean(ENABLE_PREVIEW_PROPERTY);
```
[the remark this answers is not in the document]
mstr2: This is only done at compilation time, not at runtime. JEP 12 elaborates on this:

> --enable-preview itself does not take a version number because it would be easy to misinterpret. For example, on JDK 18, the (hypothetical) flag --enable-preview 19 would appear to suggest support for the preview features of JDK 19, but those features are not known at the time of JDK 18's release.

kevinrushforth (commented):
The changes look good. I thought of one more option we might consider to check that the version of the preview API that the app compiled against matches the runtime version of the JavaFX runtime. Whether it is worth doing is something we can discuss.

modules/javafx.base/src/main/java/com/sun/javafx/PreviewFeature.java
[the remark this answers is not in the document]
mstr2: I wouldn't be in favor of requiring application developers to call a method to unlock a preview API. It seems a bit too cumbersome and intrusive to me, since it requires you to embed build information into your source code. I've also never seen this in other libraries or frameworks.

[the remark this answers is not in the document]
kevinrushforth: > I wouldn't be in favor ... since it requires you to embed build information into your source code

That seems reason enough to abandon this idea.

> I've also never seen this in other libraries or frameworks.

True. Significantly, I didn't propose anything like this for the incubator modules, which can have the same problem.

So I think this is a good minimal solution that provide a clue to the developer that they are relying on API that is unstable and will change in the future (meaning that using such API is a risk they need to be willing to take).

andy-goryachev-oracle:
> That method would unlock preview features only if the version passed in matches the runtime feature version.

what would happen if we don't add this logic?  the app will exit with `ClassNotFoundException` or some runtime error?

kevinrushforth (approved):
Looks good. I tested it and it does what I expected. I did some testing with the JDK and I was wrong about one thing:

> I don't think the warning should be suppressible; the equivalent JDK warning isn't

The key difference between the JDK and JavaFX is that the JDK only produces warnings (which are not suppressible) at compile time. Because of the compiler and JVM support, they didn't feel the need for a runtime warning.

Given that ours is a runtime warning, it might be worth restoring the system property to allow the warnings to be suppressed at runtime (default is false). What do you think?

Regarding compile time warnings for deprecated API: Should we specify that the `@Deprecated` annotation on preview features be marked as `forRemoval=true`? I can see arguments for both choices.

kevinrushforth:
> > Should we specify that the `@Deprecated` annotation on preview features be marked as `forRemoval=true`
> 
> wouldn't this send a wrong message?

Yes, that's the main counter-argument. Unlike incubator modules, which necessarily need to move elsewhere when finalized, there is a very good chance that a preview feature API will not change. So let's shelf the notion of `forRemoval=true`.

> could we invent a new annotation `@Preview` ?

That's an interesting idea, but probably not, at least not without a lot more effort. We would need to ship the annotation as part of `javafx.base`, since it would need runtime retention (so that javac and IDEs would be able to display it). We would also need to add a similar javadoc tag.

@mstr2 I presume you initially considered something like this?

mstr2:
> > Should we specify that the `@Deprecated` annotation on preview features be marked as `forRemoval=true`
> 
> wouldn't this send a wrong message? could we invent a new annotation `@Preview` ?

We could, but then we'd lose the integration that the `@Deprecated` annotation has in IDEs. For example, IntelliJ shows you visually that you're using a deprecated API, which helps with discoverability.

> @mstr2 I presume you initially considered something like this?

I've chosen the `@Deprecated` annotation because, according to its documentation, it's supposed to be used for this very situation:

`An element may be deprecated for any of several reasons, for example, [...] it may be changed incompatibly or removed in a future version [...]`

Notable, the annotation was used for preview APIs in JDK 12 and 13, so there is precedent.

kevinrushforth:
Thanks for the explanation. That all seems good to me. To summarize, we'll use the `@Deprecated` annotation (without the `forRemoval` parameter), and also use the `@deprecated` javadoc tag.

modules/javafx.base/src/main/java/com/sun/javafx/PreviewFeature.java
```diff
@@ -47,13 +47,17 @@ public enum PreviewFeature {
     private final String featureName;
 
     private static final String ENABLE_PREVIEW_PROPERTY = "javafx.enablePreview";
+    private static final String SUPPRESS_WARNING_PROPERTY = "javafx.suppressPreviewBanner";
```
kevinrushforth, line 50: Perhaps `javafx.suppressPreviewWarning` would be a better name?

kevinrushforth (commented):
Looks good, with one naming suggestion.

andy-goryachev-oracle (approved):
needs a CSR now.

andy-goryachev-oracle (approved):

modules/javafx.base/src/main/java/com/sun/javafx/PreviewFeature.java
```diff
@@ -0,0 +1,79 @@
...
+    private static final String SUPPRESS_WARNING_PROPERTY = "javafx.suppressPreviewWarning";
+
+    private static final boolean enabled = Boolean.getBoolean(ENABLE_PREVIEW_PROPERTY);
+    private static final boolean suppressWarning = Boolean.getBoolean(SUPPRESS_WARNING_PROPERTY);
+    private static final Set<PreviewFeature> enabledFeatures = new HashSet<>();
+
+    /**
+     * Verifies that preview features are enabled, and throws an exception otherwise.
+     * <p>
+     * Unless suppressed with the {@code javafx.suppressPreviewBanner=true} system property, this method
```
kevinrushforth, line 59: Banner --> Warning

kevinrushforth (approved):
I note one more place (in the docs for an internal method) where it still says "Banner".

kevinrushforth (approved):

andy-goryachev-oracle (approved):

PR #863, opened 2022-08-04
8291908: VirtualFlow creates unneeded empty cells

johanvos:
When calculating the viewportOffset, we now already take into account that cells will have to shift.

This prevents the creation of temporary empty cells in the layout phase.
One test needed to be fixed, since the number of invocations to updateItem() was hardcoded (and it
is decreased by this commit).

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issue
 * [JDK-8291908](https://bugs.openjdk.org/browse/JDK-8291908): VirtualFlow creates unneeded empty cells

### Reviewers
 * [Kevin Rushforth](https://openjdk.org/census#kcr) (@kevinrushforth - **Reviewer**)
 * [Ajit Ghaisas](https://openjdk.org/census#aghaisas) (@aghaisas - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx pull/863/head:pull/863` \
`$ git checkout pull/863`

Update a local copy of the PR: \
`$ git checkout pull/863` \
`$ git pull https://git.openjdk.org/jfx pull/863/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 863`

View PR using the GUI difftool: \
`$ git pr show -t 863`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jfx/pull/863.diff">https://git.openjdk.org/jfx/pull/863.diff</a>

</details>

PR #1360, opened 2024-02-08
JDK-8325402: TreeTableRow updateItem() does not check item with isItemChanged(..) unlike all other cell implementations

Maran23:
`TreeTableRow` does not check the item with `isItemChanged(..)`, unlike all other implementations of the cell.

This also means that the `TreeTableRow` always updates the item, although it should not, resulting in a performance loss as a `TreeTableRow` will always call `updateItem(..)`.

It looks like that this was forgotten in [JDK-8092593](https://bugs.openjdk.org/browse/JDK-8092593).

Checking the whole history, it looks like the following was happening:
1. There was a check if the item is the same in all cell implementations (with `.equals(..)`)
2. Check was removed as it caused bugs
3. Check was readded, but instead we first check the index (same index) and then if we also have the same item (this time with `.isItemChanged(..)`, to allow developers to subclass it)
4. Forgotten for `TreeTableRow` inbetween this chaos.

Added many tests for the case where an item should be changed and where it should not.
Improved existing tests as well. Using `stageLoader`, as before the tests only created a stage when doing the assert.

NOTE: The update logic is now the same as for all other 5 cell implementations. Especially `TreeCell` (since it is for a tree structure as well).

---------
### Progress

### Issue
 * [JDK-8325402](https://bugs.openjdk.org/browse/JDK-8325402): TreeTableRow updateItem() does not check item with isItemChanged(..) unlike all other cell implementations (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1360/head:pull/1360` \
`$ git checkout pull/1360`

`$ git checkout pull/1360` \
`$ git pull https://git.openjdk.org/jfx.git pull/1360/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1360`

`$ git pr show -t 1360`

</details>

<a href="https://git.openjdk.org/jfx/pull/1360.diff">https://git.openjdk.org/jfx/pull/1360.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1360#issuecomment-1934757834)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/TreeTableRow.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/TreeTableRow.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -258,7 +258,7 @@ public class TreeTableRow<T> extends IndexedCell<T> {
                 weakTreeTableViewRef = new WeakReference<>(get());
             }
 
-            updateItem();
+            updateItem(-1);
             requestLayout();
         }
     };
@@ -291,14 +291,13 @@ public class TreeTableRow<T> extends IndexedCell<T> {
 
 
     @Override void indexChanged(int oldIndex, int newIndex) {
-        index = getIndex();
+        super.indexChanged(oldIndex, newIndex);
 
         // when the cell index changes, this may result in the cell
         // changing state to be selected and/or focused.
-        updateItem();
+        updateItem(oldIndex);
         updateSelection();
         updateFocus();
-//        oldIndex = index;
     }
 
 
@@ -388,25 +387,26 @@ public class TreeTableRow<T> extends IndexedCell<T> {
      *                                                                         *
      **************************************************************************/
 
-    private int index = -1;
     private boolean isFirstRun = true;
 
-    private void updateItem() {
+    private void updateItem(int oldIndex) {
         TreeTableView<T> tv = getTreeTableView();
         if (tv == null) return;
 
         // Compute whether the index for this cell is for a real item
-        boolean valid = index >=0 && index < tv.getExpandedItemCount();
+        final int newIndex = getIndex();
+        boolean valid = newIndex >= 0 && newIndex < tv.getExpandedItemCount();
 
         final TreeItem<T> oldTreeItem = getTreeItem();
         final boolean isEmpty = isEmpty();
 
         // Cause the cell to update itself
-        if (valid) {
+        outer: if (valid) {
             // update the TreeCell state.
             // get the new treeItem that is about to go in to the TreeCell
-            final TreeItem<T> newTreeItem = tv.getTreeItem(index);
+            final TreeItem<T> newTreeItem = tv.getTreeItem(newIndex);
             final T newValue = newTreeItem == null ? null : newTreeItem.getValue();
+            final T oldValue = oldTreeItem == null ? null : oldTreeItem.getValue();
 
             // For the sake of RT-14279, it is important that the order of these
             // method calls is as shown below. If the order is switched, it is
@@ -414,11 +414,17 @@ public class TreeTableRow<T> extends IndexedCell<T> {
             // though calling cell.getTreeItem().getValue() returns the value
             // as expected
 
-            // There used to be conditional code here to prevent updateItem from
-            // being called when the value didn't change, but that led us to
-            // issues such as RT-33108, where the value didn't change but the item
-            // we needed to be listening to did. Without calling updateItem we
-            // were breaking things, so once again the conditionals are gone.
+            // RT-35864 - if the index didn't change, then avoid calling updateItem
+            // unless the item has changed.
+            if (oldIndex == newIndex) {
+                if (!isItemChanged(oldValue, newValue)) {
+                    // RT-37054:  we break out of the if/else code here and
+                    // proceed with the code following this, so that we may
+                    // still update references, listeners, etc as required.
+                    break outer;
+                }
+            }
+
             updateTreeItem(newTreeItem);
             updateItem(newValue, false);
         } else {
@@ -438,7 +444,7 @@ public class TreeTableRow<T> extends IndexedCell<T> {
 
     private void updateSelection() {
         if (isEmpty()) return;
-        if (index == -1 || getTreeTableView() == null) return;
+        if (getIndex() == -1 || getTreeTableView() == null) return;
 
         TreeTableViewSelectionModel<T> sm = getTreeTableView().getSelectionModel();
         if (sm == null) {
@@ -448,7 +454,7 @@ public class TreeTableRow<T> extends IndexedCell<T> {
             return;
         }
 
-        boolean isSelected = !sm.isCellSelectionEnabled() && sm.isSelected(index);
+        boolean isSelected = !sm.isCellSelectionEnabled() && sm.isSelected(getIndex());
         if (isSelected() != isSelected) {
             updateSelected(isSelected);
         }
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/TreeTableRow.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/TreeTableRow.java
@@ -401,7 +401,7 @@ public class TreeTableRow<T> extends IndexedCell<T> {
         final boolean isEmpty = isEmpty();
 
         // Cause the cell to update itself
-        outer: if (valid) {
+        if (valid) {
             // update the TreeCell state.
             // get the new treeItem that is about to go in to the TreeCell
             final TreeItem<T> newTreeItem = tv.getTreeItem(newIndex);
@@ -421,7 +421,7 @@ public class TreeTableRow<T> extends IndexedCell<T> {
                     // RT-37054:  we break out of the if/else code here and
                     // proceed with the code following this, so that we may
                     // still update references, listeners, etc as required.
-                    break outer;
+                    return;
                 }
             }
 
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/ListCellTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/ListCellTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -25,6 +25,7 @@
 
 package test.javafx.scene.control;
 
+import javafx.scene.control.IndexedCell;
 import javafx.scene.control.skin.ListCellSkin;
 import test.com.sun.javafx.scene.control.infrastructure.StageLoader;
 import javafx.beans.InvalidationListener;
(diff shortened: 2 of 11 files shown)


modules/javafx.controls/src/main/java/javafx/scene/control/TreeTableRow.java
```diff
@@ -388,37 +387,44 @@ private void setTreeItem(TreeItem<T> value) {
...
-            // There used to be conditional code here to prevent updateItem from
-            // being called when the value didn't change, but that led us to
-            // issues such as RT-33108, where the value didn't change but the item
-            // we needed to be listening to did. Without calling updateItem we
-            // were breaking things, so once again the conditionals are gone.
+            // RT-35864 - if the index didn't change, then avoid calling updateItem
+            // unless the item has changed.
+            if (oldIndex == newIndex) {
+                if (!isItemChanged(oldValue, newValue)) {
+                    // RT-37054:  we break out of the if/else code here and
```
andy-goryachev-oracle, line 421: could we also add (or replace?) the new reference for RT-37054: JDK-8096969

modules/javafx.controls/src/main/java/javafx/scene/control/TreeTableRow.java
```diff
@@ -388,37 +387,44 @@ private void setTreeItem(TreeItem<T> value) {
...
+                    // proceed with the code following this, so that we may
+                    // still update references, listeners, etc as required.
+                    break outer;
```
andy-goryachev-oracle, line 424: I understand neither the comment nor this `break outer;`: there is no code that follows `outer: if`, so we might as well `return`, right?

modules/javafx.controls/src/test/java/test/javafx/scene/control/ListCellTest.java
```diff
@@ -1130,6 +1135,56 @@ public void testEditCommitValueChangeIsReflectedInCell() {
...
+    @Test
+    public void testSameIndexAndItemShouldNotUpdateItem() {
+        AtomicInteger counter = new AtomicInteger();
+
+        list.setCellFactory(e -> new ListCell<>() {
+            @Override
+            protected void updateItem(String item, boolean empty) {
+                super.updateItem(item, empty);
+
+                counter.incrementAndGet();
```
andy-goryachev-oracle, line 1150: very, very minor, and probably not important: shouldn't 'incrementAndGet()` be called before `super.updateItem()`?

modules/javafx.controls/src/test/java/test/javafx/scene/control/TableCellTest.java
```diff
@@ -925,6 +928,58 @@ public void testEditCommitValueChangeIsReflectedInCell() {
...
+    @Test
+    public void testSameIndexAndItemShouldNotUpdateItem() {
+        AtomicInteger counter = new AtomicInteger();
+
+        editingColumn.setCellFactory(view -> new TableCell<>() {
+            @Override
+            protected void updateItem(String item, boolean empty) {
+                super.updateItem(item, empty);
+
+                counter.incrementAndGet();
```
andy-goryachev-oracle, line 943: same here: move before super.updateItem?

modules/javafx.controls/src/test/java/test/javafx/scene/control/TableCellTest.java
```diff
@@ -925,6 +928,58 @@ public void testEditCommitValueChangeIsReflectedInCell() {
...
+        cell.updateIndex(0);
+
+        assertEquals(0, counter.get());
+    }
+
+    /**
+     * The contract of a {@link TableCell} is that isItemChanged(..)
+     * is called when the index is 'changed' to the same number as the old one, to evaluate if we need to call
+     * updateItem(..).
+     */
```
andy-goryachev-oracle, line 961: thank you so much for providing clear explanations of what all these tests are trying to accomplish!

modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewRowTest.java
```diff
@@ -154,4 +159,56 @@ public void test_TableView_select_all_but_one() {
...
+    public void testSameIndexAndItemShouldNotUpdateItem() {
+        AtomicInteger counter = new AtomicInteger();
+
+        TableView<String> table = ControlUtils.createTableView();
+        table.setRowFactory(view -> new TableRow<>() {
+            @Override
+            protected void updateItem(String item, boolean empty) {
+                super.updateItem(item, empty);
+
+                counter.incrementAndGet();
```
andy-goryachev-oracle, line 176: and here.  the reason is that if super.updateItem() throws an exception which gets eaten (it shouldn't but we should not be assuming that) we still get the counter incremented.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeCellTest.java
```diff
@@ -1060,4 +1063,54 @@ public void testEditCommitValueChangeIsReflectedInCell() {
...
+    @Test
+    public void testSameIndexAndItemShouldNotUpdateItem() {
+        AtomicInteger counter = new AtomicInteger();
+
+        tree.setCellFactory(e -> new TreeCell<>() {
+            @Override
+            protected void updateItem(String item, boolean empty) {
+                super.updateItem(item, empty);
+
+                counter.incrementAndGet();
```
andy-goryachev-oracle, line 1078: ...

modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeTableCellTest.java
```diff
@@ -1237,6 +1240,58 @@ public void testEditCommitValueChangeIsReflectedInCell() {
...
+    @Test
+    public void testSameIndexAndItemShouldNotUpdateItem() {
+        AtomicInteger counter = new AtomicInteger();
+
+        editingColumn.setCellFactory(view -> new TreeTableCell<>() {
+            @Override
+            protected void updateItem(String item, boolean empty) {
+                super.updateItem(item, empty);
+
+                counter.incrementAndGet();
```
andy-goryachev-oracle, line 1255: ...

modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeTableRowTest.java
```diff
@@ -939,4 +943,56 @@ public void test_TreeTableView_select_all_but_one() {
...
+    public void testSameIndexAndItemShouldNotUpdateItem() {
+        AtomicInteger counter = new AtomicInteger();
+
+        TreeTableView<String> tree = ControlUtils.createTreeTableView();
+        tree.setRowFactory(view -> new TreeTableRow<>() {
+            @Override
+            protected void updateItem(String item, boolean empty) {
+                super.updateItem(item, empty);
+
+                counter.incrementAndGet();
```
andy-goryachev-oracle, line 960: ...

modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeTableViewTest.java
```diff
@@ -1772,6 +1774,49 @@ private void addChildren(TreeItem parent, String name) {
         VirtualFlowTestUtils.assertCellTextEquals(table, 2, "2", "updated name2");
     }
 
+    @Test public void testSetChildrenShouldUpdateTheCells() {
```
andy-goryachev-oracle, line 1777: we probably should have `@Test` on its own line

modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeTableViewTest.java
```diff
@@ -2591,14 +2636,14 @@ public void updateItem(String item, boolean empty) {
 
         StageLoader sl = new StageLoader(treeTableView);
 
-        assertEquals(22, rt_31200_count);
+        assertEquals(18, rt_31200_count);
```
andy-goryachev-oracle, line 2639: magic number... should we explain why exactly 18 ?

andy-goryachev-oracle (commented):
Tried the path with the TreeTableView page in the MonkeyTester on macOS 14.3, see no ill effects - though the page is somewhat limited to what it can exercise.

The SCCE generates the following output:
```
updateItem
updateItem
updateItem
updateItem
updateItem
updateItem
updateItem
updateItem
updateItem
updateItem
updateItem
updateItem
updateItem
updateItem
updateItem
updateItem
updateItem
updateItem
updateItem
updateItem
updateItem
updateItem
updateItem
isItemChanged
<clicked to expand>
isItemChanged
isItemChanged
updateItem
isItemChanged
isItemChanged
<clicked to collapse>
isItemChanged
updateItem
isItemChanged
```

is this expected output?

modules/javafx.controls/src/main/java/javafx/scene/control/TreeTableRow.java
Maran23: Yes! I did it this way to be consistent with all other cell implementations. I really think that this should be refactored, I can also create a ticket for this. My preference would be to be consistent now, but change it for all cell implementations in one go.
What is your opinion?

modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeTableViewTest.java
Maran23: sure, just copied from above.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeTableViewTest.java
Maran23: It is the count, how often update item was called. 
That it got smaller shows, that we call it less often, as the item was not changed (which is a good thing).
That also matches with your observation, which is the expected output here.

I saw that this was changed for some tests in https://github.com/openjdk/jfx/pull/863, we also can do something similar here as well.

modules/javafx.controls/src/main/java/javafx/scene/control/TreeTableRow.java
Maran23: same as the other comment, I would prefer to do that for all cell implementations in one go. If that is okay for you as well. :)

modules/javafx.controls/src/test/java/test/javafx/scene/control/ListCellTest.java
Maran23: Hm, I see no reason for either one or the other. So not strong preference from my side here.

modules/javafx.controls/src/main/java/javafx/scene/control/TreeTableRow.java
andy-goryachev-oracle: my opinion: the code is not broken per se.  Since this block is edited, I would have changed it just in this file to make the logic clearer (in other places, like TreeTableCell:685 it is functional because we have more code in line 692 and on).
I don't think another ticket is necessary because, technically speaking, the code is not broken.
(just my opinion)

modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeTableViewTest.java
andy-goryachev-oracle: on second thought, this is ok as is - I asked because the code is not testing a specific requirement.
The second `assertEquals(18)` might be testing a specific requirement - that the count has not changed, right?

Maran23: Yes, as far as I can see.

modules/javafx.controls/src/main/java/javafx/scene/control/TreeTableRow.java
Maran23: yes, but other cells have the exact same pattern, although it is not needed. And technically speaking, this is not even needed in `TreeTableCell`, tbh I never even needed a labeled if ever. 😄 
But sure, can also change.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeTableViewTest.java
andy-goryachev-oracle: so in theory, we could just test the fact that the count has not changed:
```
int ct = rt_31200_count;
...
assertEquals(ct, rt_31200_count);
```

this will make the test independent of the first number (which might change) and only test what's important (that the count does not change later).

what do you think?

modules/javafx.controls/src/main/java/javafx/scene/control/TreeTableRow.java
andy-goryachev-oracle: okay

andy-goryachev-oracle (approved):
Tested in MT on macOS 14.3.1
Looks good, thank you for making the changes!

karthikpandelu (approved):
LGTM, tested in Mac 14.3.1.

PR #1361, opened 2024-02-09
8311492: FontSmoothingType LCD produces wrong color when transparency is used

kevinrushforth:
JavaFX LCD text rendering (aka sub-pixel antialiasing) uses a pixel shader and alpha blending. The alpha channel is used is ways that interfere with its use for transparency. The existing logic checks that the current blend equation is SRC_OVER and that the surface is opaque, and that we are rendering using a Paint of type Color. It fails to check that the text color is opaque. When it isn't, the resulting alpha value is not preserved, even in the middle of the filled portion of the text, resulting in a visually noticeable difference in color.

transparent-lcd-text

The solution is to add the missing check for alpha == 1 to the test that checks whether we can use LCD text rendering. I note that Java2D falls back to gray scale when the text color is transparent for a similar reason.

I added a robot test that checks the color in the middle of the filled portion of a rendered text character and also checks that we use LCD for opaque colors and GRAY scale for transparent colors.

---------
### Progress

### Issue
 * [JDK-8311492](https://bugs.openjdk.org/browse/JDK-8311492): FontSmoothingType LCD produces wrong color when transparency is used (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1361/head:pull/1361` \
`$ git checkout pull/1361`

`$ git checkout pull/1361` \
`$ git pull https://git.openjdk.org/jfx.git pull/1361/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1361`

`$ git pr show -t 1361`

</details>

<a href="https://git.openjdk.org/jfx/pull/1361.diff">https://git.openjdk.org/jfx/pull/1361.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1361#issuecomment-1935892929)

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWGraphics.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWGraphics.java
@@ -660,8 +660,8 @@ final class SWGraphics implements ReadbackGraphics {
         final boolean doLCDText = drawAsMasks &&
                 (strike.getAAMode() == FontResource.AA_LCD) &&
                 getRenderTarget().isOpaque() &&
-                (this.paint.getType() == Paint.Type.COLOR) &&
-                ((Color)this.paint).getAlpha() == 1.0f &&
+                this.paint instanceof Color c &&
+                c.getAlpha() == 1.0f &&
                 tx.is2D();
         BaseTransform glyphTx = null;
 
--- a/modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWGraphics.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWGraphics.java
@@ -661,6 +661,7 @@ final class SWGraphics implements ReadbackGraphics {
                 (strike.getAAMode() == FontResource.AA_LCD) &&
                 getRenderTarget().isOpaque() &&
                 (this.paint.getType() == Paint.Type.COLOR) &&
+                ((Color)this.paint).getAlpha() == 1.0f &&
                 tx.is2D();
         BaseTransform glyphTx = null;
 
--- a/tests/system/src/test/java/test/robot/javafx/scene/TransparentLCDTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/scene/TransparentLCDTest.java
@@ -78,7 +78,7 @@ public class TransparentLCDTest {
     // and only sample outside the text, where we expect to read the background
     // color, or in the middle of the text fill area, where we expect to find
     // the unadjusted text color.
-    private static final double TOLERANCE = 2.0 / 255.0;
+    private static final double TOLERANCE = 2.5 / 255.0;
 
     private static final int TEXT_X_LEFT = 5;
     private static final int TEXT_Y_BOTTOM = 30;
@@ -187,7 +187,7 @@ public class TransparentLCDTest {
 
     @AfterEach
     public void doTeardown() {
-        Platform.runLater(() -> {
+        Util.runAndWait(() -> {
             if (testStage != null) {
                 testStage.hide();
             }
@@ -269,5 +269,4 @@ public class TransparentLCDTest {
     public void testOpaqueLCD() {
         runTest(true);
     }
-
 }
--- a/tests/system/src/test/java/test/robot/javafx/scene/TransparentLCDTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/scene/TransparentLCDTest.java
@@ -92,8 +92,8 @@ public class TransparentLCDTest {
     private static final int TEST_X_RIGHT = TEST_WIDTH - 1;
     private static final int TEST_X_MID = TEST_WIDTH / 2;
 
-    private static final Color transpColor = Color.color(0.5, 0.5, 0.5, 0.6);
-    private static final Color opaqueColor = makeOpaque(transpColor);
+    private static final Color TRANSP_COLOR = Color.color(0.5, 0.5, 0.5, 0.6);
+    private static final Color OPAQUE_COLOR = makeOpaque(TRANSP_COLOR);
 
     private Robot robot;
     private Stage testStage;
@@ -124,8 +124,7 @@ public class TransparentLCDTest {
 
    private void assertColorEquals(Color expected, Color actual) {
         if (!testColorEquals(expected, actual, TOLERANCE)) {
-            fail("expected:" + colorToString(expected) +
-                    " but was:" + colorToString(actual));
+            fail("expected:" + expected + " but was:" + actual);
         }
     }
 
@@ -135,14 +134,6 @@ public class TransparentLCDTest {
         double deltaBlue = Math.abs(expected.getBlue() - actual.getBlue());
         double deltaOpacity = Math.abs(expected.getOpacity() - actual.getOpacity());
         return (deltaRed <= delta && deltaGreen <= delta && deltaBlue <= delta && deltaOpacity <= delta);
-    }
-
-    private static String colorToString(Color c) {
-        int r = (int)(c.getRed() * 255.0);
-        int g = (int)(c.getGreen() * 255.0);
-        int b = (int)(c.getBlue() * 255.0);
-        int a = (int)(c.getOpacity() * 255.0);
-        return "rgba(" + r + "," + g + "," + b + "," + a + ")";
     }
 
     private List<Color> getColors(Scene scene, int x, int y, int width) {
@@ -206,7 +197,7 @@ public class TransparentLCDTest {
     // Called by the test methods to run the test either using an opaque color,
     // which should use LCD, or a transparent color, which should not.
     private void runTest(boolean opaque) {
-        final Color textColor = opaque ? opaqueColor : transpColor;
+        final Color textColor = opaque ? OPAQUE_COLOR : TRANSP_COLOR;
 
         Font font = Font.font("System", FontWeight.BOLD, 36);
 
@@ -239,11 +230,11 @@ public class TransparentLCDTest {
             List<Color> colors = getColors(testScene, SAMPLE_X_START, SAMPLE_Y, TEST_WIDTH);
 
             if (DEBUG) {
-                System.err.println("transpColor = " + colorToString(transpColor));
-                System.err.println("opaqueColor = " + colorToString(opaqueColor));
+                System.err.println("TRANSP_COLOR = " + TRANSP_COLOR);
+                System.err.println("OPAQUE_COLOR = " + OPAQUE_COLOR);
                 System.err.println("");
                 colors.stream()
-                        .map(TransparentLCDTest::colorToString)
+                        .map(Color::toString)
                         .forEach(System.err::println);
             }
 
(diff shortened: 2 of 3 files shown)


modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWGraphics.java
```diff
@@ -661,6 +661,7 @@ public void drawString(GlyphList gl, FontStrike strike, float x, float y,
                 (strike.getAAMode() == FontResource.AA_LCD) &&
                 getRenderTarget().isOpaque() &&
                 (this.paint.getType() == Paint.Type.COLOR) &&
+                ((Color)this.paint).getAlpha() == 1.0f &&
```
hjohn, line 664: Could avoid the cast here if you want:

```suggestion
                this.paint instanceof Color c && 
                c.getAlpha() == 1.0f &&
```

tests/system/src/test/java/test/robot/javafx/scene/TransparentLCDTest.java
```diff
@@ -0,0 +1,282 @@
...
+    // the first part of the 'V', and in the middle of the first part of the 'V'
+    private static final int SAMPLE_X_START = TEXT_X_LEFT;
+    private static final int SAMPLE_Y = 11;
+    private static final int TEST_WIDTH = 12;
+    private static final int TEST_X_LEFT = 0;
+    private static final int TEST_X_RIGHT = TEST_WIDTH - 1;
+    private static final int TEST_X_MID = TEST_WIDTH / 2;
+
+    private static final Color transpColor = Color.color(0.5, 0.5, 0.5, 0.6);
+    private static final Color opaqueColor = makeOpaque(transpColor);
```
hjohn, line 96: minor: These look like constants to me, but are not in caps.

tests/system/src/test/java/test/robot/javafx/scene/TransparentLCDTest.java
```diff
@@ -0,0 +1,282 @@
...
+        return (deltaRed <= delta && deltaGreen <= delta && deltaBlue <= delta && deltaOpacity <= delta);
+    }
+
+    private static String colorToString(Color c) {
+        int r = (int)(c.getRed() * 255.0);
+        int g = (int)(c.getGreen() * 255.0);
+        int b = (int)(c.getBlue() * 255.0);
+        int a = (int)(c.getOpacity() * 255.0);
+        return "rgba(" + r + "," + g + "," + b + "," + a + ")";
+    }
```
hjohn, line 146: Color has a decent `toString` already

    @Override
    public String toString() {
        return "Color[" +
            "r=" + r +
            ", g=" + g +
            ", b=" + b +
            ", a=" + a +
            "]";
    }

hjohn (approved):
Looks good to me

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWGraphics.java
kevinrushforth: I had thought of it earlier, but didn't do it at the time. I made this change along with the suggested test changes. Thanks for the suggestions.

hjohn (approved):

tests/system/src/test/java/test/robot/javafx/scene/TransparentLCDTest.java
```diff
@@ -0,0 +1,273 @@
...
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+/**
+ * Test that renders transparent text with LDC rendering selected.
+ * Prior to the fix for JDK-8311492 the color of the text was wrong.
+ * This test verifies that the color is correct and that we fall back to
+ * gray scale AA.
+ *
+ * @test
+ * @bug 8311492
```
andy-goryachev-oracle, line 67: for my education: what is the meaning of these tags?  how are they used?

kevinrushforth: These are jtreg tags. They aren't used now, since we don't use jtreg to run our tests, but might in the future.

tests/system/src/test/java/test/robot/javafx/scene/TransparentLCDTest.java
```diff
@@ -0,0 +1,273 @@
...
+        // check the GraphicsPipeline, so we check for 3D support instead.
+        assumeTrue(Platform.isSupported(ConditionalFeature.SCENE3D));
+
+        Util.runAndWait(() -> robot = new Robot());
+        Util.parkCursor(robot);
+    }
+
+    @AfterEach
+    public void doTeardown() {
+        Platform.runLater(() -> {
```
andy-goryachev-oracle, line 190: should this be Util.runAndWait() ?

tests/system/src/test/java/test/robot/javafx/scene/TransparentLCDTest.java
```diff
@@ -0,0 +1,273 @@
...
+
+    private static final int WIDTH = 400;
+    private static final int HEIGHT = 300;
+
+    // We use a very tight color tolerance, which requires some extra care.
+    // We force a screen scale of 1, stay away from the edges of the window,
+    // and only sample outside the text, where we expect to read the background
+    // color, or in the middle of the text fill area, where we expect to find
+    // the unadjusted text color.
+    private static final double TOLERANCE = 2.0 / 255.0;
```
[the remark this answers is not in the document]
kevinrushforth: Maybe. Or maybe even `2.5 / 255.0`. I like the latter idea better, so will do that.

tests/system/src/test/java/test/robot/javafx/scene/TransparentLCDTest.java
```diff
@@ -0,0 +1,273 @@
...
+    public void testTransparentLCD() {
+        runTest(false);
+    }
+
+    @Test
+    @Timeout(15)
+    public void testOpaqueLCD() {
+        runTest(true);
+    }
+
```
andy-goryachev-oracle, line 272: extra newline?

tests/system/src/test/java/test/robot/javafx/scene/TransparentLCDTest.java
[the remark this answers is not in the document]
andy-goryachev-oracle: yes, much better, thank you.

tests/system/src/test/java/test/robot/javafx/scene/TransparentLCDTest.java
```diff
@@ -0,0 +1,273 @@
...
+/**
+ * Test that renders transparent text with LDC rendering selected.
+ * Prior to the fix for JDK-8311492 the color of the text was wrong.
+ * This test verifies that the color is correct and that we fall back to
+ * gray scale AA.
+ *
+ * @test
+ * @bug 8311492
+ */
+public class TransparentLCDTest {
```
andy-goryachev-oracle, line 69: Was there a reason you did not use `VisualTestBase` class?
The code is fine the way it is, I am just curious.

kevinrushforth: I started out using `VisualTestBase`, but I needed to set system properties before launching the FX runtime, and there is no way that I could find to insert any code into the subclass that would get executed before  the superclass.

The only thing I ended up using from `VisualTestBase` was the `assertColorEquals` method, which could be moved to Util anyway. There is other cleanup that could be done to `VisualTestBase`, so I'll file a follow-on bug.

tests/system/src/test/java/test/robot/javafx/scene/TransparentLCDTest.java
kevinrushforth: It doesn't really matter in this case since Runnables are run in order, meaning that the creation and showing of the stage for the next text is guaranteed to happen after the stage is hidden. I can change it, though.

tests/system/src/test/java/test/robot/javafx/scene/TransparentLCDTest.java
```diff
@@ -0,0 +1,272 @@
...
+
+        Platform.setImplicitExit(false);
+        final CountDownLatch launchLatch = new CountDownLatch(1);
+        Util.startup(launchLatch, launchLatch::countDown);
+        assertEquals(0, launchLatch.getCount());
+    }
+
+    @AfterAll
+    public static void doTeardownOnce() {
+        Util.shutdown();
```
andy-goryachev-oracle, line 172: I am sorry, Kevin, you just walked into a trap carefully laid by me before.
Util.shutdown() needs a Stage argument, otherwise it does nothing.
Perhaps we could change the impl. to shutdown all the stages regardless instead of passing an array of them.

And, of course, there is more than one place where we do it.  Good job, me!

andy-goryachev-oracle: We can leave this code as is and fix it during the test sprint...  I'll file a bug.

andy-goryachev-oracle (approved):
LgTm

prrace (approved):
Didn't look at the test, but the code change seems right.
I'm surprised this didn't get reported a long time ago.

tests/system/src/test/java/test/robot/javafx/scene/TransparentLCDTest.java
kevinrushforth: One could argue that the method is fine as it is. It will first hide the stages that are passed in (possibly zero of them), and then shutdown the platform. I guess I don't see the problem.

Note that by the time this method is called, there are no visible stages.

andy-goryachev-oracle: I guess the reason for explicit hiding() of the stages passed to it was done for the sake of headless tests?

kevinrushforth: It's not really a headless vs headful distinction (the system tests are all headful). It's just that some tests have a global (static) Stage that should be hidden at shutdown time. Others, like the test in this PR, create, hide, and show a Stage in each test instance. Both are valid use cases.

kevinrushforth:
> I'm surprised this didn't get reported a long time ago.

I'm also surprised.

tests/system/src/test/java/test/robot/javafx/scene/TransparentLCDTest.java
andy-goryachev-oracle: here it's under @AfterAll, not after each test.

But still, what's the point of hiding the stage(s) if we follow it with Platform.exit() ?
Then we should get rid of the stage(s) argument and simply hide all the current stages, regardless.

hjohn:
> > I'm surprised this didn't get reported a long time ago.
> 
> I'm also surprised.

I do a lot of transparent text, cross fades, animation and other stuff, but first thing I always do is turn lcd rendering off...

hjohn (approved):

prrace:
> > > I'm surprised this didn't get reported a long time ago.
> > 
> > 
> > I'm also surprised.
> 
> I do a lot of transparent text, cross fades, animation and other stuff, but first thing I always do is turn lcd rendering off...

Which is a reasonable thing to do for all of that. LCD text is for the standard case of labels, textfields, menu items etc, and so only UI controls make it the default (where available). But I am STILL surprised

tests/system/src/test/java/test/robot/javafx/scene/TransparentLCDTest.java
kevinrushforth: For this test, the call to shutdown is in an `@AfterAll` method. The call to `Stage.hide` is in an `@AfterEach` method.

PR #1368, opened 2024-02-14
8322748: Caret blinking in JavaFX should only stop when caret moves

andy-goryachev-oracle:
Move caret animation handling due to keyboard input to the Skin, by registering a listener on the caretPosition property.  This will restart animation only when the caret position changes instead of every key press.

This also simplifies internal behaviors of TextArea, TextField, and PasswordField in light of the future InputMap RFE [JDK-8314968](https://bugs.openjdk.org/browse/JDK-8314968)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextAreaSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextAreaSkin.java
@@ -175,10 +175,12 @@ public class TextAreaSkin extends TextInputControlSkin<TextArea> {
             }
         };
         caretPosition.addListener((observable, oldValue, newValue) -> {
+            setCaretAnimating(false);
             targetCaretX = -1;
             if (control.getWidth() > 0) {
                 setForwardBias(true);
             }
+            setCaretAnimating(true);
         });
 
         forwardBiasProperty().addListener(observable -> {
--- a/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/TextInputControlBehavior.java
+++ b/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/TextInputControlBehavior.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -320,14 +320,8 @@ public abstract class TextInputControlBehavior<T extends TextInputControl> exten
         for (Object o : map.getChildInputMaps()) {
             addKeyPadMappings((InputMap<T>)o);
         }
-
     }
 
-
-    /**
-     * Wraps the event handler to pause caret blinking when
-     * processing the key event.
-     */
     protected KeyMapping keyMapping(final KeyCode keyCode, final EventHandler<KeyEvent> eventHandler) {
         return keyMapping(new KeyBinding(keyCode), eventHandler);
     }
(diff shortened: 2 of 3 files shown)


andy-goryachev-oracle:
@hjohn would you like to review this change?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextAreaSkin.java
```diff
@@ -175,10 +175,12 @@ public TextAreaSkin(final TextArea control) {
             }
         };
         caretPosition.addListener((observable, oldValue, newValue) -> {
+            setCaretAnimating(false);
             targetCaretX = -1;
             if (control.getWidth() > 0) {
                 setForwardBias(true);
             }
+            setCaretAnimating(true);
```
hjohn, line 183: I find this a bit confusing.  By putting the two `setCaretAnimating` separate, you give the impression it must stop blinking for a while before continuing because of the call to `setForwardBias`.

However, all we want to is call it with `false` then `true` to restart the animation (for a lack of a `playFromStart` call).  Perhaps it may be good to introduce a method that captures this more clearly as I find it confusing (`restartCaretAnimation`)

Same comment for `TextFieldSkin`.

hjohn (changes requested):
I think if we're busy here, it would be worthwhile to also remove the other calls to the skin to change the caret animating. I think it would require a focus listener in Skin.  Also code like this in `TextInputControlSkin` makes me believe that it may work out of the box correctly, even with selections, and all the code in the behavior is unnecessary (or can at least be fully handled by the `Skin` with perhaps another listener:

From `TextInputControlSkin`:

        /**
         * The caret is visible when the text box is focused AND when the selection
         * is empty. If the selection is non empty or the text box is not focused
         * then we don't want to show the caret. Also, we show the caret while
         * performing some operations such as most key strokes. In that case we
         * simply toggle its opacity.
         * <p>
         */
        caretVisible = new BooleanBinding() {
            { bind(control.focusedProperty(), control.anchorProperty(), control.caretPositionProperty(),
                    control.disabledProperty(), control.editableProperty(), displayCaret, blinkProperty());}
            @Override protected boolean computeValue() {
                // RT-10682: On Windows, we show the caret during selection, but on others we hide it
                return !blinkProperty().get() && displayCaret.get() && control.isFocused() &&
                        (isWindows() || (control.getCaretPosition() == control.getAnchor())) &&
                        !control.isDisabled() &&
                        control.isEditable();
            }
        };

At a minimum it seems there is some duplication going on...

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextAreaSkin.java
andy-goryachev-oracle: good point

andy-goryachev-oracle:
> I think if we're busy here, it would be worthwhile to also remove the other calls to the skin to change the caret animating.

I would rather not: it would require a much larger effort.  Avoiding the caret flickering when its position does not change is an incremental improvement from which we all benefit.

hjohn (approved):

hjohn:
> > I think if we're busy here, it would be worthwhile to also remove the other calls to the skin to change the caret animating.
> 
> I would rather not: it would require a much larger effort. Avoiding the caret flickering when its position does not change is an incremental improvement from which we all benefit.

Alright, the original issue did envision this though, specifically to start get rid of some of the blockers for a Behavior API.  The issue isn't fully addressed then, so I guess I'll file another one.

andy-goryachev-oracle:
> so I guess I'll file another one.

Thank you, please do.

As a side note, before we can start re-writing the controls' behaviors, we should:

a) identify all the behavioral aspects of each control and
b) develop a set of tests to exercise each one, along the lines of [JDK-8314906](https://bugs.openjdk.org/browse/JDK-8314906)

without these steps, we are guaranteed to create regression.

andy-goryachev-oracle:
@aghaisas or @karthikpandelu could you please take a look at this PR?

hjohn:
> As a side note, before we can start re-writing the controls' behaviors, we should:
> 
> a) identify all the behavioral aspects of each control and b) develop a set of tests to exercise each one, along the lines of [JDK-8314906](https://bugs.openjdk.org/browse/JDK-8314906)
> 
> without these steps, we are guaranteed to create regression.

I think those tests should be unit tests, not via Robot, to make this a feasible endeavor. This is the reason why I asked about a mocking framework to be included in FX test utilities. We should look into making it much easier to set up a Behavior or Control or Skin, without having to fire up a complete FX.  Things like animations and timers should be mocked out, so there can be precise control during the tests.

andy-goryachev-oracle:
> I think those tests should be unit tests, not via Robot

Preferably, yes.  If something can be tested via headless unit tests, it should be done so, you are absolutely right.

Some tests, however, require a fully rendered skin - for example, PgUp in a TextArea.  I suspect only 10-20% of all the tests will require a headful environment.

karthikpandelu (approved):
LGTM

kevinrushforth (approved):

PR #1373, opened 2024-02-20
8273743: KeyCharacterCombination for "+" does not work on US QWERTY keyboard layout

beldenfox:
On Linux getKeyCodeForChar does not consult the current keyboard layout. For example, it assumes the “+” character is on KeyCode.PLUS even on layouts which don’t generate KeyCode.PLUS. The result is that most KeyCharacterCombinations don’t work.

This PR fixes this using the same approach that Mac glass uses. When the user types a key we lookup all the characters that key might generate and put them in a map. getKeyCodeForChar then consults the map to find the Java key code. This ensures that we only pay attention to keys that are on the user’s physical keyboard.

(Some Linux layout maps are expansive and include entries for keys that may be on the user’s keyboard but probably aren’t, like “(“ and “)” on the numeric keypad. If we simply ask for all entries that generate a given character we can get multiple hits with no good way to determine which ones are physically present.)

This PR also contains fixes to the Robot to enable testing. When Glass consults the GdkKeymap to determine which keys might generate a keyval GDK returns a list covering every installed layout and shift level. The old Glass code simply picked the first entry in the list. This PR iterates through the list looking for one that works for the current layout and correct shift level.

---------
### Progress

### Issue
 * [JDK-8273743](https://bugs.openjdk.org/browse/JDK-8273743): KeyCharacterCombination for "+" does not work on US QWERTY keyboard layout (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1373/head:pull/1373` \
`$ git checkout pull/1373`

`$ git checkout pull/1373` \
`$ git pull https://git.openjdk.org/jfx.git pull/1373/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1373`

`$ git pr show -t 1373`

</details>

<a href="https://git.openjdk.org/jfx/pull/1373.diff">https://git.openjdk.org/jfx/pull/1373.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1373#issuecomment-1954795166)


beldenfox:
Comment added to keep the bots at bay.

beldenfox:
I've updated to the latest master. This will probably be the last time I refresh this pull request; I can only hold off the bots for so long.

andy-goryachev-oracle:
In the context of the recent discussion on the mailing list, would it be possible to actually detect the keyboard layout?  There is a limited number of layouts, so in theory that should be possible.

Another question - I recall on Windows there is a way to enter certain key using Alt codes.  Does this method exist on Linux?

beldenfox:
> In the context of the recent discussion on the mailing list, would it be possible to actually detect the keyboard layout? 

I answered this question in full on the mailing list but will provide a summary here. There's no documented way of doing this on Windows. The Mac organizes layouts based on language (for example, English), Linux organizes them based on country (so Australia, Great Britain, and the U.S. are all distinct). On all platforms there are more layouts than you think (the Mac has 5 German layouts and Linux has 20 Germany layouts) and there's no standard on how the variants are designated.

> Another question - I recall on Windows there is a way to enter certain key using Alt codes. Does this method exist on Linux?

I'm not that familiar with this feature but here's my understanding. The feature isn't universal but tied to the input method system you're using. With IBus you can press Shift+Ctrl+U followed by the Unicode value followed by Enter. But if you want to use an IME with JavaFX you have to use the obsolete XIM framework instead which doesn't support this.

andy-goryachev-oracle (approved):
Tested on linux mint, the app in the ticket fails without the fix and works with the fix.

The code looks good to me, though I'd like to have a second pair of eyes on it.

andy-goryachev-oracle:
I think this PR should be reopened.

andy-goryachev-oracle (approved):
@kevinrushforth please review

andy-goryachev-oracle:
oh no it got auto closed again... sorry.
I think this PR should be re-opened.

kevinrushforth (approved):
The code changes look good, and all my tests pass.

kevinrushforth:
@beldenfox This is ready to integrate. Sorry for the long delay.

beldenfox:
> @beldenfox This is ready to integrate. Sorry for the long delay.

Thanks for letting me know. I haven't been getting e-mails when my PR's are ready to integrate. And thanks for reviewing this, KeyCharacterCombinations now work reliably on all the desktop platforms.

PR #1379, opened 2024-02-23
8325550: Grammatical error in AnchorPane.setLeftAnchor (and other setters) javadoc

nlisker:
/solves JDK-8318624

Fixes for the `AnchorPane` docs, as well as for the `NodeOrientation` docs in `Node` and `Scene`.

Note that the default value for a `Scene`'s `NodeOrientation` depends on a system property, while for `Node` it isn't (which means `SubScene` will be different from `Scene`). Not sure if this is intended.

---------
### Progress

### Issues
 * [JDK-8325550](https://bugs.openjdk.org/browse/JDK-8325550): Grammatical error in AnchorPane.setLeftAnchor (and other setters) javadoc (**Bug** - P4)
 * [JDK-8318624](https://bugs.openjdk.org/browse/JDK-8318624): API docs specify incorrect default value for nodeOrientation property (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1379/head:pull/1379` \
`$ git checkout pull/1379`

`$ git checkout pull/1379` \
`$ git pull https://git.openjdk.org/jfx.git pull/1379/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1379`

`$ git pr show -t 1379`

</details>

<a href="https://git.openjdk.org/jfx/pull/1379.diff">https://git.openjdk.org/jfx/pull/1379.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1379#issuecomment-1961758432)

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -6446,29 +6446,28 @@ public abstract class Node implements EventTarget, Styleable {
     private byte resolvedNodeOrientation =
             EFFECTIVE_ORIENTATION_LTR | AUTOMATIC_ORIENTATION_LTR;
 
+    private static final NodeOrientation DEFAULT_NODE_ORIENTATION = NodeOrientation.INHERIT;
+
     public final void setNodeOrientation(NodeOrientation orientation) {
         nodeOrientationProperty().set(orientation);
     }
 
     public final NodeOrientation getNodeOrientation() {
-        return nodeOrientation == null ? NodeOrientation.INHERIT : nodeOrientation.get();
+        return nodeOrientation == null ? DEFAULT_NODE_ORIENTATION : nodeOrientation.get();
     }
     /**
-     * Property holding NodeOrientation.
-     * <p>
      * Node orientation describes the flow of visual data within a node.
      * In the English speaking world, visual data normally flows from
      * left-to-right. In an Arabic or Hebrew world, visual data flows
-     * from right-to-left.  This is consistent with the reading order
-     * of text in both worlds.  The default value is left-to-right.
-     * </p>
+     * from right-to-left. This is consistent with the reading order
+     * of text in both worlds.
      *
-     * @return NodeOrientation
+     * @defaultValue {@code NodeOrientation.INHERIT}
      * @since JavaFX 8.0
      */
     public final ObjectProperty<NodeOrientation> nodeOrientationProperty() {
         if (nodeOrientation == null) {
-            nodeOrientation = new StyleableObjectProperty<NodeOrientation>(NodeOrientation.INHERIT) {
+            nodeOrientation = new StyleableObjectProperty<NodeOrientation>(DEFAULT_NODE_ORIENTATION) {
                 @Override
                 protected void invalidated() {
                     nodeResolvedOrientationInvalidated();
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -6433,6 +6433,16 @@ public abstract class Node implements EventTarget, Styleable {
      *                                                                         *
      **************************************************************************/
 
+    /**
+     * Node orientation describes the flow of visual data within a node.
+     * In the English speaking world, visual data normally flows from
+     * left-to-right. In an Arabic or Hebrew world, visual data flows
+     * from right-to-left. This is consistent with the reading order
+     * of text in both worlds.
+     *
+     * @defaultValue {@code NodeOrientation.INHERIT}
+     * @since JavaFX 8.0
+     */
     private ObjectProperty<NodeOrientation> nodeOrientation;
     private EffectiveOrientationProperty effectiveNodeOrientationProperty;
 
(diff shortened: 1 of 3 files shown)


modules/javafx.graphics/src/main/java/javafx/scene/layout/AnchorPane.java
```diff
@@ -182,9 +182,9 @@ public static void setBottomAnchor(Node child, Double value) {
     }
 
     /**
-     * Returns the child's bottom anchor constraint if set.
+     * Returns the child's bottom anchor constraint, if set.
      * @param child the child node of an anchor pane
-     * @return the offset from the bottom of the anchor pane or null if no bottom anchor was set
+     * @return the offset from the bottom of the anchor pane, or {@code null} if no bottom anchor was set
```
andy-goryachev-oracle, line 187: minor suggestion: move explanation of when it returns null to the description,  i.e.:
```

Returns the child's bottom anchor constraint, if set, otherwise returns {@code null}.

@return the offset from the bottom of the anchor pane, or {@code null}
```

modules/javafx.graphics/src/main/java/javafx/scene/layout/AnchorPane.java
```diff
@@ -204,9 +204,9 @@ public static void setRightAnchor(Node child, Double value) {
     }
 
     /**
-     * Returns the child's right anchor constraint if set.
+     * Returns the child's right anchor constraint, if set.
      * @param child the child node of an anchor pane
-     * @return the offset from the right of the anchor pane or null if no right anchor was set
+     * @return the offset from the right of the anchor pane, or {@code null} if no right anchor was set
```
andy-goryachev-oracle, line 209: and here

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -6446,29 +6446,28 @@ private EventHandlerProperties getEventHandlerProperties() {
...
      * In the English speaking world, visual data normally flows from
      * left-to-right. In an Arabic or Hebrew world, visual data flows
-     * from right-to-left.  This is consistent with the reading order
-     * of text in both worlds.  The default value is left-to-right.
-     * </p>
+     * from right-to-left. This is consistent with the reading order
+     * of text in both worlds.
      *
-     * @return NodeOrientation
+     * @defaultValue {@code NodeOrientation.INHERIT}
```
andy-goryachev-oracle, line 6465: adds a javadoc warning:

```
/Users/angorya/Projects/jfx3/jfx/rt/modules/javafx.graphics/src/main/java/javafx/scene/Node.java:6468: warning: no @return
    public final ObjectProperty<NodeOrientation> nodeOrientationProperty() {
                                                 ^
```

modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
```diff
@@ -6258,25 +6259,23 @@ public final void setNodeOrientation(NodeOrientation orientation) {
...
      * left-to-right. In an Arabic or Hebrew world, visual data flows
-     * from right-to-left.  This is consistent with the reading order
-     * of text in both worlds.  The default value is left-to-right.
-     * </p>
+     * from right-to-left. This is consistent with the reading order
+     * of text in both worlds.
      *
-     * @return NodeOrientation
+     * @defaultValue if the system property {@code javafx.scene.nodeOrientation.RTL} is {@code true},
+     *     {@code NodeOrientation.RIGHT_TO_LEFT}, otherwise {@code NodeOrientation.INHERIT}
```
andy-goryachev-oracle, line 6273: adds a javadoc warning:

```
/Users/angorya/Projects/jfx3/jfx/rt/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java:6276: warning: no @return
    public final ObjectProperty<NodeOrientation> nodeOrientationProperty() {
                                                 ^
```

andy-goryachev-oracle (changes requested):
looks good (with minor suggestion)
however, this change adds 2 javadoc warnings which need to be addressed,  I would think .

modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
```diff
@@ -6258,25 +6259,23 @@ public final void setNodeOrientation(NodeOrientation orientation) {
...
      * In the English speaking world, visual data normally flows from
      * left-to-right. In an Arabic or Hebrew world, visual data flows
-     * from right-to-left.  This is consistent with the reading order
-     * of text in both worlds.  The default value is left-to-right.
-     * </p>
+     * from right-to-left. This is consistent with the reading order
+     * of text in both worlds.
      *
-     * @return NodeOrientation
+     * @defaultValue if the system property {@code javafx.scene.nodeOrientation.RTL} is {@code true},
```
kevinrushforth, line 6272: Restore the `@return` to avoid a warning.

modules/javafx.graphics/src/main/java/javafx/scene/layout/AnchorPane.java
kevinrushforth: I don't mind either way.

kevinrushforth (commented):
The changes to the docs look fine, although I note the same additional warnings that Andy saw (restoring the `@return` will fix it; the text after the `@return` doesn't matter, since it is unused).

Also, I think it would be better to revert all code changes, even though they look harmless.

modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
andy-goryachev-oracle: it seems that placing javadoc at the property declaration instead of the method which returns said property eliminates the warning. 
See for example Stage:1329

kevinrushforth: > I suggest to move the doc to the field.

Yeah that's even better.

modules/javafx.graphics/src/main/java/javafx/scene/layout/AnchorPane.java
nlisker: `@return` statements usually state this condition explicitly, like `Map#get`: "the value to which the specified key is mapped, or `null` if this map contains no mapping for the key".

nlisker:
By the way, is the difference in the default values between `Node` and `Scene` intentional? Only `Scene` has the RTL system property checked.

modules/javafx.graphics/src/main/java/javafx/scene/layout/AnchorPane.java
andy-goryachev-oracle: My point is that it should be mentioned in the first paragraph then.  It can be repeated in @return later, if you prefer.  Map.get() does indeed mention this in the first paragraph:

Screenshot 2024-02-23 at 14 30 02

andy-goryachev-oracle:
> Only `Scene` has the RTL system property checked.

I suppose because Scene must INHERIT this value from the system?

modules/javafx.graphics/src/main/java/javafx/scene/layout/AnchorPane.java
kevinrushforth: Given that the changes are so far limited to spelling and formatting (code style for null), I think it would be fine to limit the changes in this PR to what he has already done.

kevinrushforth (approved):

andy-goryachev-oracle (approved):

PR #1387, opened 2024-03-03
8327179: Update the 3D lighting application

nlisker:
Update for the 3D lighting test tool as described in the JBS issue.

---------
### Progress

### Issue
 * [JDK-8327179](https://bugs.openjdk.org/browse/JDK-8327179): Update the 3D lighting application (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1387/head:pull/1387` \
`$ git checkout pull/1387`

`$ git checkout pull/1387` \
`$ git pull https://git.openjdk.org/jfx.git pull/1387/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1387`

`$ git pr show -t 1387`

</details>

<a href="https://git.openjdk.org/jfx/pull/1387.diff">https://git.openjdk.org/jfx/pull/1387.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1387#issuecomment-1975386277)
</details>


kevinrushforth:
Reviewer: @arapte 

@jayathirthrao you might also want to take a look?

arapte:
Couple quick comments:
- Executing the LightingSample manually fails with an exception
```
Caused by: java.lang.NullPointerException: Input stream must not be null
	at javafx.graphics@23-internal/javafx.scene.image.Image.validateInputStream(Image.java:1140)
	at javafx.graphics@23-internal/javafx.scene.image.Image.<init>(Image.java:707)
	at lighting3D.Environment.<clinit>(Environment.java:64)
```
Steps: 
cd tests/performance/3DLighting/src/main/java
javac <compile.args> lighting3D/*.java
java <run.args> lighting3D.LightingApplication

- The copyright year new files should be 2024 only. Some new files have more than one copyright year. If you are moving any existing files then please use `git mv` so that the files are shown as moved.

arapte (changes requested):
providing few more comments

arapte:
An observation out of scope of this PR.
The effect of lighting on mesh vs a box is quite different. 
Notice the screenshot, left side is a Box and right side is Mesh, both have Magenta SpotLight.
I think the effect of light should be same on a 2D surface, and the effect on Box seems correct to me.
Another observation: The effect of light on Mesh is different with 100 divisions than 200+ divisions.

Screenshot 2024-03-06 at 3 36 09 PM

tests/performance/3DLighting/src/main/java/lighting3D/Benchmark.java
```diff
@@ -0,0 +1,182 @@
...
+        anim.setFromZ(150);
+        anim.setToZ(0);
+        return anim;
+    }
+
+    Button createStopButton() {
+        var stopGraphic = new Text("⏹");
+        stopGraphic.setBoundsType(TextBoundsType.VISUAL);
+        stopGraphic.setFill(Color.RED);
+        stopGraphic.setFont(Font.font(20));
```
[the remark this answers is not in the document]
nlisker: As you will see, the font sizes and visual sizes don't render as you would expect. Not sure why, but this way they look better. Perhaps small adjustments are in order, but if you match the sizes it comes out off.

nlisker:
> Executing the LightingSample manually fails with an exception

I think the package names of the main and resources weren't aligned.

> The copyright year new files should be 2024 only. Some new files have more than one copyright year. If you are moving any existing files then please use git mv so that the files are shown as moved.

For now, for the ease of reviewing, I reverted the package name change.

> The effect of lighting on mesh vs a box is quite different.

I think this is the same issue you submitted in the past: https://bugs.openjdk.org/browse/JDK-8269133

nlisker:
I can't reproduce the exception. The path to the image is correct. Both the class and the image are under the `attenuation` package. Here is my launch command:

```
C:\Program Files\Java\jdk-21\bin\javaw.exe "-Djava.library.path=C:\Users\Nir\git\jfx\modules\javafx.graphics\build\module-lib"
--add-modules=javafx.controls,javafx.swing
-Dfile.encoding=UTF-8
-Dstdout.encoding=UTF-8
-Dstderr.encoding=UTF-8
-p "C:\Users\Nir\git\jfx\modules\javafx.controls\bin;C:\Users\Nir\git\jfx\modules\javafx.graphics\bin;C:\Users\Nir\git\jfx\modules\javafx.base\bin;C:\Users\Nir\git\jfx\modules\javafx.swing\bin"
-classpath "C:\Users\Nir\git\jfx\tests\performance\3DLighting\bin"
-XX:+ShowCodeDetailsInExceptionMessages
--add-exports javafx.graphics/test.com.sun.javafx.pgstub=javafx.controls
--add-exports javafx.base/test.com.sun.javafx.binding=javafx.controls
--add-exports javafx.base/test.util.memory=javafx.controls
--add-exports javafx.base/test.javafx.collections=javafx.controls
--add-exports javafx.base/com.sun.javafx.property=javafx.graphics
--add-exports javafx.base/test.javafx.collections=javafx.graphics
--add-exports javafx.base/test.util.memory=javafx.graphics
--add-exports java.base/sun.security.util=javafx.graphics
--add-reads javafx.base=java.management
--add-reads javafx.base=jdk.management attenuation.LightingApplication
```

Perhaps it's a platform-dependent path issue?

kevinrushforth:
I can reproduce the exception if I compile the test program from the command line in the `.../src/main/java` dir using javac and run the program from there. The reason that happens is the separation of the resources into a separate directory tree.

One way to solve this, so we can still easily compile and run it from the command line, is to provide an "ant" or "gradle" script as was done for `MonkeyTester` and `FXMediaPlayer`

nlisker:
Not sure what jcheck wants. It's complaining about empty spaces that have a `}` appear after them. This is the error: https://github.com/openjdk/jfx/pull/1387/files#annotation_19798318258.

nlisker:
Looks like the line specified by jcheck was off.

build.gradle
```diff
@@ -4040,6 +4044,27 @@ project(":systemTests") {
     addValidateSourceSets(project, nonModSrcSets, modSrcSets)
 }
 
+project(":3DLighting") {
+
+    apply plugin: "application"
```
kevinrushforth, line 4049: I hadn't meant to suggest adding it to the main build.gradle. I think this is not the best place for it. I was thinking of something along the lines of a build.gradle or build.xml in the`tests/performance/3DLighting` dir which is what the other manual tests do -- at least those that are more complicated than just a collection of files in a single directory.

tests/system/src/test/.classpath
```diff
@@ -40,24 +40,24 @@
 			<attribute name="add-exports" value="java.desktop/sun.awt.datatransfer=ALL-UNNAMED"/>
 		</attributes>
 	</classpathentry>
-    <classpathentry kind="con" path="org.eclipse.jdt.junit.JUNIT_CONTAINER/5">
-        <attributes>
-            <attribute name="test" value="true"/>
-        </attributes>
-    </classpathentry>
+	<classpathentry kind="con" path="org.eclipse.jdt.junit.JUNIT_CONTAINER/5">
```
kevinrushforth, line 43: The changes in this file seem unrelated to your PR.

nlisker:
As part of investigating performance issues, I updated this application for my own use.

I think that this is ready for re-review. @lukostyra You might be interested in this too. I did a lot of tests on the D3D9 pipeline and I hope to see how D3D12 compares soon.

andy-goryachev-oracle:
Looks nice!

1. is `src/main/resources` dir still needed?
2. the play and stop icons in the Performance tab have different size
3. how do you control the scene with the mouse?  I can't figure it out - right-drag in any direction seems to rotate the objects around the y axis only
4. do you think it may be better to make this app a part of the monkey tester?


Screenshot 2026-09-18 at 15 09 19

nlisker:
> Looks nice!
> 
> 1. is `src/main/resources` dir still needed?

No, but git doesn't track dirs, so I can't remove it.

> 2. the play and stop icons in the Performance tab have different size

Ambarish [noted it too](https://github.com/openjdk/jfx/pull/1387#discussion_r1513164981), but I thought that he was looking at the code rather than the GUI. For me, they are rendered (just about) the same size even though the the code uses different sizes. I noticed it with other unicode characters too. Might this be a bug only in Windows regarding font rendering?

image

Notice the camera icon too.

> 3. how do you control the scene with the mouse?  I can't figure it out - right-drag in any direction seems to rotate the objects around the y axis only

Left click drag pans, right click drag rotates, middle click drag swivels, scrollwheel zooms.

> 4. do you think it may be better to make this app a part of the monkey tester?

I'm not sure. This app focuses on 3D behavior and performance, it's very specific.

andy-goryachev-oracle:
> right click drag rotates

it only rotates around y axis, very unintuitive.  you can't rotate it around the x axis at all.  I think typically the rotation is done like a track ball (remember those?)

> I'm not sure. This app focuses on 3D behavior and performance, it's very specific.

I think it might fit perfectly as a page in the monkey tester.  What do other people think?

nlisker:
> > right click drag rotates
> 
> it only rotates around y axis, very unintuitive. you can't rotate it around the x axis at all. I think typically the rotation is done like a track ball (remember those?)

Feel free to add an option for the bindings that are intuitive for you, either as followup or suggest the code to add; I can add you as an author.
I prefer precision: If I want to rotate on one axis only I need a single mouse button for it, otherwise very slight movements can creep in rotation on other axes and corrupt the visual examination. I need to be able to debug edge cases with this app, it's not a 3D model studio.

nlisker:
> I think it might fit perfectly as a page in the monkey tester.

If you want monkey tester to be able to load it as a page, I don't mind it, including moving it to a different location within the jfx repository. What I need is to be able to launch it as a standalone with the specified VM options (see the gradle file for defaults) and build it with gradle since it will be bound to the graphics native tasks later on when the build file starts making sense.

andy-goryachev-oracle:
> I prefer precision:

perhaps a modifier key (shift?) should do the trick, otherwise rotation around the other axis is _impossible_.

nlisker:
Added swivel with secondary mouse button + shift modifier.

tests/performance/3DLighting/src/main/java/app/CaptureUtils.java
```diff
@@ -0,0 +1,77 @@
...
+
+    enum Format {
+        BMP,
+        GIF,
+        JPG,
+        PNG,
+        TIF
+    }
+
+    private static final Path DIRECTORY = Path.of("screenshots");
```
andy-goryachev-oracle, line 52: This creates a directory in the project tree.  I would suggest either to move the directory outside (user home?), or making sure it's `.gitignore`'d

this file is probably a bad place to declare it - maybe in LightingApplication itself?  also, make sure to explain in the javadoc there about files/dirs it creates.

tests/performance/3DLighting/src/main/java/app/CaptureUtils.java
```diff
@@ -0,0 +1,77 @@
...
+        rgbImage.getGraphics().drawImage(image, 0, 0, null);
+
+        String formatName = extension.name().toLowerCase();
+        var file = DIRECTORY.resolve(Path.of("screenshot" + imageNum + "." + formatName)).toAbsolutePath().toFile();
+        try {
+            Files.createDirectories(DIRECTORY);
+            if (!ImageIO.write(rgbImage, formatName, file)) {
+                throw new IOException("No writer found for " + formatName);
+            }
+            Desktop.getDesktop().open(DIRECTORY.toAbsolutePath().toFile());
```
andy-goryachev-oracle, line 68: seems backwards - first it writes the screenshot, then opens a file chooser, why?
also, if I cancel the file chooser I expect nothing to be written out.

I think it either needs to write silently (or maybe with a message saying "saved in XXX"), or follow the standard procedure to let the user select the folder to write the screenshot or cancel.  and remember that choice (ideally, between the sessions).

andy-goryachev-oracle:
May I ask you to remove
```
	<classpathentry kind="src" path="src/main/resources"/>
```
(line 4) from the `.classpath` file, or create a 0 bytes `.gitkeep` file in that directory?

tests/performance/3DLighting/src/main/java/app/Benchmark.java
```diff
@@ -0,0 +1,274 @@
...
+            elapsedFrames += 1;
+            totalElapsedTime += currTime - lastTime;
+            totalElapsedFrames += 1;
+
+            double elapsedSeconds = elapsedTime / 1e9;
+            double totalElapsedSeconds = totalElapsedTime / 1e9;
+            if (elapsedSeconds >= 5.0) {
+                double fps = elapsedFrames / elapsedSeconds;
+                instantFps.set(fps);
+                System.out.println("\ninstant fps: " + fps);
```
andy-goryachev-oracle, line 242: these stats are printed to stdout - would it make more sense to shows them in let's say a status bar at the bottom?

tests/performance/3DLighting/src/main/java/app/Benchmark.java
```diff
@@ -0,0 +1,274 @@
...
+    final DoubleProperty averageFps = new SimpleDoubleProperty();
+
+    Benchmark(Environment environment) {
+        this.environment = environment;
+    }
+
+    Button createStopButton() {
+        var stopGraphic = createGraphic("⏹");
+        stopGraphic.setFill(Color.RED);
+        stopGraphic.setFont(Font.font(20));
```
andy-goryachev-oracle: this sets graphic size to 20, but L182 it's 40 - that explains why they look different.  use one size for all perhaps?

tests/performance/3DLighting/src/main/java/app/Environment.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    Node createScreenshotControls() {
+        var formats = new ChoiceBox<Format>();
+        formats.getItems().addAll(Format.values());
+        formats.setValue(Format.PNG);
+
+        var graphic = new Text("📷");
+        graphic.setBoundsType(TextBoundsType.VISUAL);
+        graphic.setFont(Font.font(32));
```
andy-goryachev-oracle: can we use the same size for all the icons?

tests/performance/3DLighting/src/main/java/app/Controls.java
```diff
@@ -0,0 +1,194 @@
...
+        light.lightOnProperty().bind(lightOn.selectedProperty());
+        var colorPicker = new ColorPicker(light.getColor());
+        light.colorProperty().bind(colorPicker.valueProperty());
+        var titleControls = new HBox(5, lightOn, colorPicker);
+        titleControls.setAlignment(Pos.CENTER_LEFT);
+
+        var titlePane = new TitledPane("", content);
+        titlePane.setGraphic(titleControls);
+        titlePane.setExpanded(false);
+        return titlePane;
```
andy-goryachev-oracle, line 166: are "AmbientLights" sections supposed to expand?  if not, it's probably wrong control for the puspose

Image

nlisker:
I searched JBS for a bug regarding the difference in rendering of font sizes on Windows and Mac (and Linux?), but couldn't find anything. If others haven't come across it, I'll file the issue.

tests/performance/3DLighting/src/main/java/app/CameraScene3D.java
```diff
@@ -130,16 +120,21 @@ private final void setUIBindings() {
...
-                boolean positiveY = curY > getHeight() / 2;
-                deltaX = positiveY ? -deltaX : deltaX;
-                deltaY = positiveX ? deltaY : -deltaY;
-                rotate((deltaX + deltaY)/2);
-            } else if (e.getButton() == MouseButton.MIDDLE) {
-                swivle(deltaY);
+            switch (e.getButton()) {
+                case PRIMARY -> pan(deltaX, deltaY);
+                case SECONDARY -> {
+                    if (e.isShiftDown()) {
```
andy-goryachev-oracle, line 126: thank you.  the rotation is still counter-intuitive, but at least it can be done in both axes.

how does the user know about controls?  should the control gestures be described in the main app javadoc or a `README.md` ?

tests/performance/3DLighting/src/main/java/app/Environment.java
```diff
@@ -0,0 +1,197 @@
...
+import javafx.scene.text.Font;
+import javafx.scene.text.Text;
+import javafx.scene.text.TextBoundsType;
+
+/// The 3D environment. Includes the lights and shapes.
+class Environment extends CameraScene3D {
+
+    // Reuse the background from the [PhongMaterial] docs instead of copying it to the resources of this project.
+    static {
+        Path path = Path.of("").toAbsolutePath().getParent().getParent().getParent() // jfx root
```
andy-goryachev-oracle, line 64: should this dependency be documented?  possibly in a readme.md?

andy-goryachev-oracle (commented):
This is a test app, so perhaps a good README.md is in order, describing:
 
- features
- dependencies
- how to control

tests/performance/3DLighting/src/main/java/app/CaptureUtils.java
nlisker: That line opens the folder where the screenshot it after it was taken, it's not a file chooser. It's for convenience. The screenshots you take in this app are for demonstration of something, so you always want to do something with it. It makes it easier if you don't need to dig around for the path.

tests/performance/3DLighting/src/main/java/app/Benchmark.java
nlisker: They are already shown in the GUI at the Performance pane (that's what `instantFps.set(fps);` does).

The printing is done to keep the previous results in a log-like fashion. They are also legacy (Kevin wrote the fps counter, as well as the original quad).

tests/performance/3DLighting/src/main/java/app/CaptureUtils.java
andy-goryachev-oracle: oh right, it is just a small Finder shaped like a chooser.  still, a but unexpected (opens each time I make a screenshot), but it's a test program so probably ok.

tests/performance/3DLighting/src/main/java/app/Benchmark.java
andy-goryachev-oracle: ah, ok, so make it look like a log: timestamp, value1, value2; one line per event?

tests/performance/3DLighting/src/main/java/app/Benchmark.java
nlisker: And then Windows is broken:
image
This is with size 40 on all 3 icons.

Someone who knows fonts needs to look at this.

tests/performance/3DLighting/src/main/java/app/Controls.java
nlisker: It's not supposed to expand, there are no properties for this light. The control was reused for ease. I removed the arrow.

tests/performance/3DLighting/src/main/java/app/Benchmark.java
andy-goryachev-oracle: maybe use a regular `Label` with inline style?

tests/performance/3DLighting/src/main/java/app/CaptureUtils.java
nlisker: Added a `.gitignore`. I prefer to keep it local and not "pollute" places outside of this application.

I think that the `DIRECTORY` field is in the right place. It's a local constant. Had it been configurable it could have moved to some settings class.

tests/performance/3DLighting/src/main/java/app/Benchmark.java
nlisker: It's not a log per-se and a timestamp doesn't add anything because the duration is constant. In what way does it not do its function when used?

tests/performance/3DLighting/src/main/java/app/Benchmark.java
nlisker: Still different render sizes for the same (default) font:
image

tests/performance/3DLighting/src/main/java/app/Benchmark.java
andy-goryachev-oracle: it's just weird and a lot of whitespace/scrolling.
it's ok to leave as is since it's not essential.

nlisker:
I could add a readme, but reading the code would probably be faster. It's a test application for developers who're already looking at minute details and will need to read the code to see what the performance tests actually do.
Also, a readme can easily drift as the program updates, so I prefer javadoc.

tests/performance/3DLighting/src/main/java/app/Benchmark.java
andy-goryachev-oracle: it's weird: why do play/stop buttons have different height?

andy-goryachev-oracle:
javadoc is fine.  
reading a bunch of files is probably not a good user experience, assuming you are writing for the users (and not just yourself).

tests/performance/3DLighting/src/main/java/app/Benchmark.java
nlisker: No idea. It's why I suggested that someone who knows about fonts should look at this. Looks like each unicode symbol has it's own default size.

andy-goryachev-oracle: how about using text then?  Play/Stop?

nlisker: Was thinking about it, or svg's.

andy-goryachev-oracle: I've been trying to create icons programmatically but it's a slow process

https://github.com/andy-goryachev/AppFramework/tree/main/src/goryachev/fx/icon

nlisker: An svg creator like InkScape or any of the online ones can be used to create all the symbols. The problem is when you need complex shapes like the camera.

tests/performance/3DLighting/.gitignore
```diff
@@ -0,0 +1 @@
+screenshots/
```
andy-goryachev-oracle: missing newline?  not an issue if git understands it, but github shows a marker.

andy-goryachev-oracle (approved):
thank you for making the changes!
one minor comment about .gitignore file

andy-goryachev-oracle:
forgot to ask - are you going to switch to the text buttons?

kevinrushforth:
@arapte @jayathirthrao @lukostyra Can one of you be the second reviewer?

nlisker:
> forgot to ask - are you going to switch to the text buttons?

I used a uniform size compromise for now. It's not too bad on Windows and should be correct on the other OS's.

andy-goryachev-oracle (commented):
We had a brief conversation about using text as icons https://bugs.openjdk.org/browse/JDK-8392843 - basically, no amount of trickery can produce the desired result.  we should be using svg or generated Paths.  Having said that, for a test applications the text-based icons should be fine.

While playing with the app, I loaded background from a large image (a screenshot).  Selected box shape, moved it around, then switched to null background.  The shape disappeared - all I can see is a blank area.  Where did it go?  Is there a way to reset the viewport parameters to the initial values (maybe they should get rest when changing the model?)

nlisker:
> We had a brief conversation about using text as icons https://bugs.openjdk.org/browse/JDK-8392843 - basically, no amount of trickery can produce the desired result.

Is the analysis available anywhere? I wonder why it's impossible to get uniform sizes for unicode characters.

> While playing with the app, I loaded background from a large image (a screenshot). Selected box shape, moved it around, then switched to null background. The shape disappeared - all I can see is a blank area. Where did it go?

No idea, didn't go anywhere for me:
demo2

> Is there a way to reset the viewport parameters to the initial values (maybe they should get rest when changing the model?)

Never needed that functionality. It's not a 3D studio like Maya or Blender. I do very specific tests with it, then close it so I can recompile the native sources, than open it again. It's rare that I need to do more than a handful of operations with it. What visuals are you testing with it?

andy-goryachev-oracle:
I'll let Phil write the answer in the ticket.  From what I understood, the it depends on the particular font's (or fallback font's) glyphs, and they can vary - there is no guarantee of consistency.

For disappearing shapes - I figured it out.  If you use the mouse wheel to slide the shape all the way back, they disappear.  Why?  The background is None, why should they disappear?

example:
Screenshot 2026-09-23 at 11 51 27

nlisker:
> For disappearing shapes - I figured it out. If you use the mouse wheel to slide the shape all the way back, they disappear. Why? The background is None, why should they disappear?

Ah, it's because they are not in the [frustum](https://en.wikipedia.org/wiki/Viewing_frustum) anymore. Increase the `farClip` size (in the code) and then try, but the size is always finite so at some point your shape will always disappear.

andy-goryachev-oracle:
> [frustum](https://en.wikipedia.org/wiki/Viewing_frustum)

it should have a larger depth maybe - what is the point of having it too shallow?  is this a meaningful test?

nlisker:
> > [frustum](https://en.wikipedia.org/wiki/Viewing_frustum)
> 
> it should have a larger depth maybe - what is the point of having it too shallow? is this a meaningful test?

For what purpose do you need a larger depth? All the tests I've ever done didn't need a larger far clip. What are you testing that requires it?

andy-goryachev-oracle:
I asked first :-)
The setting is invisible, maybe it should be configurable via UI?

nlisker:
Why would you want to set it dynamically here? Like I said, all the tests I've done didn't need such a large frustum.

andy-goryachev-oracle (approved):
lgtm

PR #1388, opened 2024-03-04
8327177: Wrong GlobalRef deleted

johanvos:
Explicitly remove old global ref value from menu->jCallback.
This avoids the possibility of that value (which can be reused for other refs) to be deleted twice.


kevinrushforth (approved):
This is a safe fix that is clearly the right thing to do.

I recommend changing the Bug title (and then the PR title to match) to add "macOS" and "GlassMenu" somewhere in the title to make it a little more informative, but I'll leave it up to you.

This is a simple enough fix that a single reviewer is sufficient. Once integrated, I recommend backporting this to jfx22u for 22.0.1 (and possibly to earlier releases as you see fit).

kevinrushforth:
Updated title looks good. Thanks.

PR #1389, opened 2024-03-04
8092102: Labeled: textTruncated property

andy-goryachev-oracle:
Adds **Labeled.textTruncated** property which indicates when the text is visually truncated (and the ellipsis string is inserted) in order to fit the available width.

The new property is being set by the code which computes the actual text string to be displayed (and which inserts the ellipsis string) in `LabeledSkinBase.updateDisplayedText(double,double)`.

**Alternative**

None exists as this requires changes to the core (Utils).

**See Also**

* [JDK-8327483](https://bugs.openjdk.org/browse/JDK-8327483) TreeView: Allow for tooltip when cell text is truncated
* [JDK-8205211](https://bugs.openjdk.org/browse/JDK-8205211) Ability to show Tooltip only when text is shown with ellipsis (...)

---------
### Progress
- [x] Change requires CSR request [JDK-8329293](https://bugs.openjdk.org/browse/JDK-8329293) to be approved

### Issues
 * [JDK-8092102](https://bugs.openjdk.org/browse/JDK-8092102): Labeled: textTruncated property (**Enhancement** - P3)
 * [JDK-8329293](https://bugs.openjdk.org/browse/JDK-8329293): Labeled: textTruncated property (**CSR**)

### Reviewers
 * [Ambarish Rapte](https://openjdk.org/census#arapte) (@arapte - **Reviewer**) 🔄 Re-review required (review applies to [aa28eb4e](https://git.openjdk.org/jfx/pull/1389/files/aa28eb4ef2e62c5968f45676bcdc5077cea2393e))

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1389/head:pull/1389` \
`$ git checkout pull/1389`

`$ git checkout pull/1389` \
`$ git pull https://git.openjdk.org/jfx.git pull/1389/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1389`

`$ git pr show -t 1389`

</details>

<a href="https://git.openjdk.org/jfx/pull/1389.diff">https://git.openjdk.org/jfx/pull/1389.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1389#issuecomment-1977641676)


karthikpandelu (commented):
Overall change looks good to me.
Do you think change in `labelPaddingProperty`or `graphicProperty` should also trigger check in newly added truncated property? I believe the properties which are already bound should take care of this. Just wanted to confirm here.

modules/javafx.controls/src/main/java/javafx/scene/control/Labeled.java
```diff
@@ -818,6 +818,48 @@ public final BooleanProperty mnemonicParsingProperty() {
...
+                        textProperty(),
+                        widthProperty(),
+                        wrapTextProperty()
+                    );
+                }
+
+                @Override
+                protected boolean computeValue() {
+                    if (isWrapText()) {
+                        return false;
```
nlisker, line 850: Are you sure that allowing text to wrap necessarily means it won't be truncated? What if the max height doesn't allow another line?

modules/javafx.controls/src/main/java/javafx/scene/control/Labeled.java
```diff
@@ -818,6 +818,48 @@ public final BooleanProperty mnemonicParsingProperty() {
...
+    /**
+     * Truncated read-only property indicates whether the text has been truncated
+     * when it cannot fit into the available width.
+     * <p>
+     * When truncated, the {@link #ellipsisStringProperty() ellipsis string}
+     * gets inserted in the place dictated by the
+     * {@link #textOverrun} property.
+     *
+     * @since 23
+     * @defaultValue false
```
nlisker, line 830: I don't think that a default value is applicable here since it's completely dependent on other properties and can't be set. It's possible to call the `get` of this property for the first time and get `true`. The value here is the value used in the get method:
```java
return property == null ? DEFAULT_VALUE : property.get();
```
at least in all the cases I've seen.

modules/javafx.controls/src/main/java/javafx/scene/control/Labeled.java
```diff
@@ -818,6 +818,48 @@ public final BooleanProperty mnemonicParsingProperty() {
...
+     */
+    private ObservableBooleanValue truncated;
+
+    public final ObservableBooleanValue truncatedProperty() {
```
nlisker, line 834: `ObservableBooleanValue` is not a property, so it's odd to use it as the return type of one. Usually `ReadOnlyBooleanProperty` is used. This implementation through binding is also unique for read-only properties. A look in `Node` and `Labeled` show a different way of implementing read-only properties.

modules/javafx.controls/src/main/java/javafx/scene/control/Labeled.java
```diff
@@ -818,6 +818,48 @@ public final BooleanProperty mnemonicParsingProperty() {
         return mnemonicParsing;
     }
 
+    /**
+     * Truncated read-only property indicates whether the text has been truncated
+     * when it cannot fit into the available width.
```
nlisker, line 823: No need to repeat the name and type in the first sentence as it's used in the summary table. Most phrasings opt (inconsistently for): "Indicates whether..." or just "Whether..." (I prefer the former).

I'm also not sure about the sole role of the width. Can the text not be truncated for a reason other than not fitting in the width, like not fitting in the height?

modules/javafx.controls/src/main/java/javafx/scene/control/Labeled.java
```diff
@@ -818,6 +818,48 @@ public final BooleanProperty mnemonicParsingProperty() {
...
+     * <p>
+     * When truncated, the {@link #ellipsisStringProperty() ellipsis string}
+     * gets inserted in the place dictated by the
+     * {@link #textOverrun} property.
+     *
+     * @since 23
+     * @defaultValue false
+     */
+    private ObservableBooleanValue truncated;
```
nlisker, line 832: The property name should probably be `textTruncated` to be in line with `textOverrun`, `textFill`, `wrapText`...

Otherwise it could look like the labeled is truncated.

nlisker (commented):
This enhancement should be discussed in the mailing list. I've never needed this myself, but it already had an old JBS ticket, so I can see that others did and for a legitimate reason (adding tooltips). I don't mind if it's added or not.

andy-goryachev-oracle:
> Do you think change in `labelPaddingProperty`or `graphicProperty` should also trigger check

good question!  a change in these properties gets handled correctly - no need to add explicit binding.

kevinrushforth:
> This enhancement should be discussed in the mailing list. I've never needed this myself, but it already had an old JBS ticket, so I can see that others did and for a legitimate reason (adding tooltips). I don't mind if it's added or not.

Yes, this warrants a mailing list discussion, especially given the following comment in the PR Description:

> **Alternative**
> 
> The desired functionality can be just as easily achieved on an application level, by adding a similar property to a subclass.  What is the benefit of adding this functionality to the core?

I also don't mind one way or the other. If it is felt to be a useful convenience, especially to make it easier to optionally enable Tooltip generation, it seems OK to me.

/reviewers 2

kevinrushforth (commented):
Presuming we go forward with this, I left a couple comments on the API.

modules/javafx.controls/src/main/java/javafx/scene/control/Labeled.java
andy-goryachev-oracle: Good point.  So the label will have its text truncated (by inserting the ellipsis string) when wrapText is on and the size is constrained by setting maxHeight.

Interestingly, it will not truncate the text (again, ellipsis string) if the label is resized by the layout:

Screenshot 2024-03-05 at 12 54 14

Using the latest MonkeyTester to test
uncomment the listener in LabelPage

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableCellSkin.java
```diff
@@ -123,7 +130,7 @@ protected void layoutChildren(double x, double y, double w, double h) {
     @Override
     protected double computePrefWidth(double height, double topInset, double rightInset, double bottomInset,
             double leftInset) {
-        if (isDeferToParentForPrefWidth) {
+        if (LabeledHelper.isUseContentWidth() || isDeferToParentForPrefWidth) {
```
Maran23, line 133: I'm not conviced that this is the correct solution here. 
First of all I would keep this PR simple and ONLY add the truncated Property.
Second, I would rather want to see if this can be changed without adding some flags which once again will not help application developers that may want to extend something like this.

IMO, the separation of concern is wrong. A Cell should always give back the prefWidth, other callers should think if they want to use the pref width or the table column width.

In order to move this forward, I would like to see only the necessary changes, which is the truncated property. Some tests for it would be good as well.

Maran23 (changes requested):

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableCellSkin.java
andy-goryachev-oracle: Thank you for the feedback! 

The property is being added to Labeled, which happens to be a base class for the cells' skins.  The truncated property therefore must work in all the subclasses, even those where computePrefWidth is hacked via `Properties.DEFER_TO_PARENT_PREF_WIDTH`.

> Second, I would rather want to see if this can be changed without adding some flags which once again will not help application developers that may want to extend something like this.

This is a much more difficult task:  First, it needs a careful discussion on which APIs to add, if any (there is a possibility that some people might say that no new APIs are needed as one can always create an instance of Text or Label and use it to get the sizing).  There might be a need to undo the hacks made by the founding father such as `Properties.DEFER_TO_PARENT_PREF_WIDTH`.  I would say all this is out of scope for this PR, as it achieves its goal without introducing the new APIs.

> IMO, the separation of concern is wrong. A Cell should always give back the prefWidth, other callers should think if they want to use the pref width or the table column width.

Why cells are controls is slightly beyond me, but who am I to ask?  I would say there should be one control in this picture, and that is TableView (or any other cell-based control) whose job is to size and lay out its various parts.  The current design is different, and it is out of scope for this PR to change the design.

> In order to move this forward, I would like to see only the necessary changes

And that 's exactly what you see here - only the necessary changes (that unfortunately include undoing the earlier hacks in cells).

As for tests - I certainly agree, but that would be a manual test.  The behavior can be verified with the examples in JDK-8205211, I just felt the value of a manual test is rather low.  What I would rather see is an automated test that actually verifies the new property behavior down to a pixel.  Any ideas how to make one?

Maran23: Okay.

Tests: Well, there are some properties we can test here:
- wrapText = true -> height should be used -> check truncated 
- width changed -> check truncated 
- text changed -> check truncated

Maran23: You can check here how I manipulated the size of the stage of the `StageLoader`.
https://github.com/openjdk/jfx/pull/1396
Than we can check and set some properties on any `Labeled`, probably the `Label` is good here. And on the stage if needed (width, height).

andy-goryachev-oracle: thanks, tests are coming.

modules/javafx.controls/src/test/java/test/javafx/scene/control/LabeledTruncatedTest.java
```diff
@@ -0,0 +1,168 @@
...
+        truncated.set(null);
+
+        System.out.println("set 20"); // TODO remove
+        container.setPrefWidth(20);
+        container.setMaxWidth(20);
+        container.setMinWidth(20);
+        firePulse();
+
+        // FIX fails
+        //assertEquals(Boolean.TRUE, truncated.get());
```
andy-goryachev-oracle, line 153: still WIP, I can't figure out why the table cells have width 0

andy-goryachev-oracle:
unit tests added, please re-review.

karthikpandelu (commented):
Overall this looks good to me. Left couple of inline comments.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/LabeledHelper.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Labeled Helper.
+ */
+public class LabeledHelper {
+    /** Accessor */
+    public interface Accessor {
+        /**
+         * Returns true when the Labeled must compute the actual content width in computePrefWidth().
+         * @return whether computePrefWidth() must compute the actual content width
```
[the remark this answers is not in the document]
andy-goryachev-oracle: is the new version clearer?

karthikpandelu (approved):
LGTM

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/LabeledHelper.java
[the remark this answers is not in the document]
karthikpandelu: Yes. Thanks for updating.

andy-goryachev-oracle:
@aghaisas could you take a look at this PR please?

victordyakov:
@azuev-java please review

arapte:
Testing looks good. Works as expected.
The change looks good. Please allow me a little more time to take a more closer look.

arapte (approved):
Looks good. I tested a few scenarios with TableView, Button, Label. It behaved as expected

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/LabeledHelper.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+package com.sun.javafx.scene.control;
+
+import javafx.scene.control.Labeled;
+import com.sun.javafx.util.Utils;
+
+/**
+ * Labeled Helper.
+ */
+public class LabeledHelper {
```
kevinrushforth, line 33: You might consider making `LabeledHelper` a subclass of `ControlHelper`, which is the usual pattern for helper classes of nodes, although it may not matter here (and could be done later if there was a need).

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/LabeledHelper.java
```diff
@@ -0,0 +1,65 @@
...
+    public static void setAccessor(Accessor a) {
+        accessor = a;
+    }
+
+    /**
+     * Determines whether {@code computePrefWidth()} must compute the actual content width,
+     * or the width of the table header.
+     * @return true to use the actual content width, false to delegate to the parent
+     */
+    public static boolean isUseActualContentWidth() {
```
kevinrushforth, line 62: If you make the `useActualContentWidth` flag an instance variable of `Labeled` (see my comment in that class), then you would need to add the `Labeled` as an argument to this method (here and in the accessor method).

modules/javafx.controls/src/test/java/test/javafx/scene/control/LabeledTruncatedTest.java
```diff
@@ -0,0 +1,95 @@
...
+import org.junit.jupiter.api.Test;
+import com.sun.javafx.scene.layout.RegionHelper;
+import com.sun.javafx.tk.Toolkit;
+
+/**
+ * Tests textTruncated property of Labeled, using Label, TableCell, and TreeTableCell controls
+ * (the last two contain conditional code that redirects the execution of computePrefWidth())
+ * in their skins to different code paths.
+ */
+public class LabeledTruncatedTest {
```
kevinrushforth, line 45: It might be worth adding a test for `Button`.

modules/javafx.controls/src/test/java/test/javafx/scene/control/LabeledTruncatedTest.java
```diff
@@ -0,0 +1,95 @@
...
+        RegionHelper.setWidth(control, 10);
+        RegionHelper.setHeight(control, h);
+        firePulse();
+
+        assertTrue(control.isTextTruncated());
+
+        control.setWrapText(true);
+        firePulse();
+
+        assertFalse(control.isTextTruncated());
```
kevinrushforth, line 93: Can you add a test for the case where we are wrapping and the preferred height is exceeded?

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/LabeledHelper.java
kevinrushforth: Thinking on it further, let's not do this unless / until there is a need (there likely won't be).

modules/javafx.controls/src/main/java/javafx/scene/control/Labeled.java
```diff
@@ -96,6 +98,29 @@ public abstract class Labeled extends Control {
...
+     * to compute the actual content width rather than table column width.
+     * (We should have never used such a logic without exposing the way to obtain
+     * the content width!).
+     * It is safe to use a public global flag because:
+     * a) it's an implementation detail and
+     * b) we are always in the content of the FX app thread
+     * This functionality is made separate from Properties.DEFER_TO_PARENT_PREF_WIDTH which
+     * by itself looks rather weird.
+     */
+    private static boolean useActualContentWidth;
```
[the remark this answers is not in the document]
andy-goryachev-oracle: re-entrancy should not be an issue: the flag is used in the context of a single method which is always being called from an fx application thread.

I do not want to increase the memory footprint by making it an instance variable.

modules/javafx.controls/src/test/java/test/javafx/scene/control/LabeledTruncatedTest.java
andy-goryachev-oracle: test added

modules/javafx.controls/src/main/java/javafx/scene/control/Labeled.java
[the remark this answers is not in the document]
andy-goryachev-oracle: on second thought, you are right.  we cannot guarantee that someone won't stick a TableView as a graphic into another TableView cell.

Switched to use an ephemeral entry in `Node.getProperties()`, same as `Properties.DEFER_TO_PARENT_PREF_WIDTH`.

kevinrushforth:
The latest code changes look good. I need to do some more testing tomorrow. I have what I think is a simple test program that isn't behaving like I expected. I'll double-check and let you know.

kevinrushforth (changes requested):
I found at least one case where this fails to detect truncation. There might be others. I will attach a test case that shows the failure.

Additionally, I left a comment about the modifications to Region. Since they are only needed for testing, you can instead add getWidth / getHeight to RegionShim, allowing you to revert all changes to Region (and RegionHelper).

kevinrushforth:
Here is a test showing the failure to detect truncated text.

```
import javafx.scene.Scene;
import javafx.stage.Stage;

public class TruncatedTest extends Application {

    private static final String clickMe = "Click me";
    private static final String clickMeAgain = "Click me again, please";

    @Override public void start(Stage stage) {
        stage.setTitle("Truncated Button Text (fails)");

        var root = new HBox(10);
        Scene scene = new Scene(root, 600, 450);

        var button = new Button(clickMe);
        button.setPrefWidth(100);
        button.setOnAction(e -> {
            button.setText(clickMe.equals(button.getText()) ? clickMeAgain : clickMe);
            //System.err.println("truncated: " + button.isTextTruncated());
        });
        root.getChildren().add(button);

        var label = new Label();
        label.textProperty().bind(button.textTruncatedProperty().asString());
        root.getChildren().add(label);

        stage.setScene(scene);
        stage.show();
    }

        Application.launch(args);
    }
}
```

modules/javafx.controls/src/main/java/javafx/scene/control/Labeled.java
```diff
@@ -818,6 +829,63 @@ public final BooleanProperty mnemonicParsingProperty() {
...
+                Bindings.createBooleanBinding(() -> {
+                    // make sure prefWidth always returns the actual content width
+                    // rather than the column width if inside a table
+                    setUseActualContentWidth(true);
+                    try {
+                        if (isWrapText()) {
+                            return (getHeight() < prefHeight(getWidth()));
+                        }
+
+                        return (getWidth() < prefWidth(getHeight()));
```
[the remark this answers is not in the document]
andy-goryachev-oracle: > Would it be possible to set the value of the truncated property at the point where truncation occurs?

yes, LabeledSkinBase.updateDisplayedText() determines whether the text is truncated or not.

andy-goryachev-oracle:
Changed the way the new property is being set.  Instead of listening for a bunch of existing properties (that's bad), hooked up directly into `LabeledSkinBase.updateDisplayedText()` which does the actual manipulation with the text.

In addition, modified the Monkey Tester

to add ability to see the value of various properties in real time.

For example, to verify that the new property is being updated in the context of a table, select TableView page, set Control -> Context Menu -> Show Properties Monitor, select a cell and right click to open the Property Monitor window.  Find a `TextFieldTableCell` in the list and observe the value changing when resizing the column:

Screenshot 2024-05-03 at 15 27 43

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/skin/Utils.java
```diff
@@ -213,8 +212,14 @@ public static int computeTruncationIndex(Font font, String text, double width) {
...
 
-    public static String computeClippedText(Font font, String text, double width,
-                                     OverrunStyle type, String ellipsisString) {
+    public static String computeClippedText(
+        Font font,
+        String text,
+        double width,
+        OverrunStyle type,
+        String ellipsisString,
+        AtomicBoolean textTruncated
```
kevinrushforth, line 221: I recommend setting `textTruncated` to false as the first statement (alternatively, add a comment that the caller is expected to initialize it to false).

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/skin/Utils.java
```diff
@@ -414,9 +425,17 @@ public static String computeClippedText(Font font, String text, double width,
...
-                                            String ellipsisString, TextBoundsType boundsType) {
+    public static String computeClippedWrappedText(
+        Font font,
+        String text,
+        double width,
+        double height,
+        double lineSpacing,
+        OverrunStyle truncationStyle,
+        String ellipsisString,
+        AtomicBoolean textTruncated,
```
kevinrushforth, line 436: Similarly, I recommend setting it to false as the first statement. In this case, I see that it will be set to false in the case where it makes it to the return statement at the end of the method without being truncated, but there is one early return where this isn't the case.

modules/javafx.controls/src/main/java/javafx/scene/control/Labeled.java
```diff
@@ -818,6 +828,37 @@ public final BooleanProperty mnemonicParsingProperty() {
...
+
+    public final boolean isTextTruncated() {
+        return textTruncated().get();
+    }
+
+    private final void setTextTruncated(boolean on) {
+        textTruncated().set(on);
+    }
+
+    private ReadOnlyBooleanWrapper textTruncated() {
```
kevinrushforth, line 855: Suggestion: rename this method to something like `textTruncatedImpl()` for clarity (as it is, the method, which is a writable property, shares the same name as the read-only property field, which could be confusing).

kevinrushforth (commented):
Yes, this looks like a better approach for the implementation. I did a quick test and it now works as expected.

I left a couple comments inline.

modules/javafx.controls/src/main/java/javafx/scene/control/Labeled.java
kevinrushforth: Never mind. I missed that the types of the field and method are identical, so I withdraw my suggestion.

modules/javafx.controls/src/main/java/javafx/scene/control/Labeled.java
```diff
@@ -818,6 +828,37 @@ public final BooleanProperty mnemonicParsingProperty() {
         return mnemonicParsing;
     }
 
+    /**
+     * Indicates whether the text has been truncated
+     * because it cannot fit into the available width.
+     * <p>
+     * When truncated, the {@link #ellipsisStringProperty() ellipsisString}
+     * gets inserted in the place dictated by the
+     * {@link #textOverrun} property.
```
kevinrushforth, line 837: I just noticed that this link doesn't work. Looks like you'll need to change it to point to the `textOverrunProperty()` method.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/skin/Utils.java
andy-goryachev-oracle: setting to false would be incorrect in the case of multi-line text.

I can add a comment, though it's not a public API and there was no comment/javadoc before.

kevinrushforth: Yes, I see. In that case, it's up to you whether to add a simple comment that this should be initialized by the caller (but maybe it doesn't need to be?)

kevinrushforth (approved):
Looks good. You'll need to change the PR title to match the updated JBS title.

Also, you can finalize the CSR.

andy-goryachev-oracle:
> Looks good. You'll need to change the PR title to match the updated JBS title.
> Also, you can finalize the CSR.

Thank you for your help, Kevin!

kevinrushforth:
@karthikpandelu @arapte Can you re-review?

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/skin/Utils.java
```diff
@@ -213,8 +212,29 @@ public static int computeTruncationIndex(Font font, String text, double width) {
...
+     * @param textTruncated the reference to a flag indicating the truncation
+     * @return the actual text to be shown, with the ellipsis string inserted when needed
+     */
+    public static String computeClippedText(
+        Font font,
+        String text,
+        double width,
+        OverrunStyle type,
+        String ellipsisString,
+        AtomicBoolean textTruncated
```
mstr2, line 236: Instead of returning the clipped text with the return value, and the truncation flag with an argument, have you considered just using a record to return both at the same time? That would be a cleaner solution.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/skin/Utils.java
```diff
@@ -213,8 +212,29 @@ public static int computeTruncationIndex(Font font, String text, double width) {
         return index;
     }
 
-    public static String computeClippedText(Font font, String text, double width,
-                                     OverrunStyle type, String ellipsisString) {
+    /**
+     * Computes the actual text to be shown in the Labeled with the text wrapping disabled:
+     * unmodified if it fits into available area,
+     * or with the ellipsis string inserted into strategic place(s) if it does not.
```
karthikpandelu, line 218: ellipsis string can be referenced similar to textTruncated?

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/skin/Utils.java
```diff
@@ -414,9 +440,35 @@ public static String computeClippedText(Font font, String text, double width,
         }
     }
 
-    public static String computeClippedWrappedText(Font font, String text, double width,
-                                            double height, double lineSpacing, OverrunStyle truncationStyle,
-                                            String ellipsisString, TextBoundsType boundsType) {
+    /**
+     * Computes the actual text to be shown in the Labeled with the text wrapping enabled:
+     * unmodified if it fits into available area,
+     * or with the ellipsis string inserted into strategic place(s) if it does not.
```
karthikpandelu, line 446: Similar to previous comment. ellipsis string can be referenced similar to textTruncated?

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/skin/Utils.java
andy-goryachev-oracle: Not really a cleaner solution: `computeClippedWrappedText()` might get called multiple times (for each line of text), and doing an extra allocation per call is less optimal than passing a reference to an AtomicBoolean.

karthikpandelu (commented):
Looks good to me overall. Added a minor comment.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/skin/Utils.java
andy-goryachev-oracle: Can you please clarify your question?  I don't understand.

kevinrushforth: While that would be an OK change, this is an internal method, so it really doesn't much matter -- the docs are only seen by those looking at the code.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/skin/Utils.java
andy-goryachev-oracle: oh, you want to use `{@code ellipsisString}` instead of the words "ellipsis string", got it.
I thought it was clear enough, but yes, can be easily changed.

kevinrushforth (approved):

karthikpandelu (approved):

PR #1394, opened 2024-03-08
8296387: [Tooltip, CSS] -fx-show-delay is only applied to the first tooltip that is shown before it is displayed

Maran23:
This PR fixes a long standing issue where the `Tooltip` will always wait one second until it appears the very first time, even if the 
`-fx-show-delay` was set to another value.

The culprit is, that the `cssForced` flag is not inside `Tooltip`, but inside the `TooltipBehaviour`. So the very first `Tooltip` gets processed correctly, but after no `Tooltip` will be processed by CSS before showing, resulting in the set `-fx-show-delay` to not be applied immediately.

Added a bunch of headful and headless tests for the behaviour since there were none before.

---------
### Progress

### Issue
 * [JDK-8296387](https://bugs.openjdk.org/browse/JDK-8296387): [Tooltip, CSS] -fx-show-delay is only applied to the first tooltip that is shown before it is displayed (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1394/head:pull/1394` \
`$ git checkout pull/1394`

`$ git checkout pull/1394` \
`$ git pull https://git.openjdk.org/jfx.git pull/1394/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1394`

`$ git pr show -t 1394`

</details>

<a href="https://git.openjdk.org/jfx/pull/1394.diff">https://git.openjdk.org/jfx/pull/1394.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1394#issuecomment-1985956774)


andy-goryachev-oracle:
Testing the combined (this PR merged with the latest master) on macOS 14.3.1 M1 Pro (though this is probably not important), using the latest MonkeyTester

- select Tooltip page
- open the Tools -> CSS Playground tool
- type `.tooltip { -fx-show-delay:1ms; }` into Custom CSS text area, press Update button.  (this updates the stylesheet in all the open windows)
- hover over the tooltip page

What I see is the very first time the tooltip appears it's using the old show-delay value of 1000ms.  Any subsequent attempts the new delay is in effect. 

Then, change back to 1000ms.  Same thing happens: the first time the old timeout value is used instead of the new.

andy-goryachev-oracle:
There is also another problem (though it is a separate issue as it can be seen with the master branch):

- set the Graphic: to "Image"

Notice how sometimes, when slowly entering the tooltip page from the left side, the tooltip appears and then disappears.  Async code is hard!

andy-goryachev-oracle:
video:


https://github.com/openjdk/jfx/assets/107069028/02589ed0-d359-4811-8843-de1109fd23c9

andy-goryachev-oracle:
This is another, unrelated issue: If I specify a number instead of duration like `-fx-show-delay: 1;` we'll get an exception every time the tooltip is about to be shown.

```
Exception in thread "JavaFX Application Thread" java.lang.ClassCastException: class java.lang.Double cannot be cast to class javafx.util.Duration (java.lang.Double is in module java.base of loader 'bootstrap'; javafx.util.Duration is in module javafx.base of loader 'app')
	at javafx.controls/javafx.scene.control.Tooltip.getShowDelay(Tooltip.java:334)
	at javafx.controls/javafx.scene.control.Tooltip$TooltipBehavior.lambda$0(Tooltip.java:1015)
	at javafx.base/com.sun.javafx.event.CompositeEventHandler$NormalEventHandlerRecord.handleBubblingEvent(CompositeEventHandler.java:247)
	at javafx.base/com.sun.javafx.event.CompositeEventHandler.dispatchBubblingEvent(CompositeEventHandler.java:80)
	at javafx.base/com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:232)
	at javafx.base/com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:189)
	at javafx.base/com.sun.javafx.event.CompositeEventDispatcher.dispatchBubblingEvent(CompositeEventDispatcher.java:59)
	at javafx.base/com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:58)
	at javafx.base/com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
	at javafx.base/com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:56)
	at javafx.base/com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
	at javafx.base/com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:56)
	at javafx.base/com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
	at javafx.base/com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:56)
	at javafx.base/com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
	at javafx.base/com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:56)
	at javafx.base/com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
	at javafx.base/com.sun.javafx.event.BasicEventDispatcher.dispatchEvent(BasicEventDispatcher.java:56)
	at javafx.base/com.sun.javafx.event.EventDispatchChainImpl.dispatchEvent(EventDispatchChainImpl.java:114)
	at javafx.base/com.sun.javafx.event.EventUtil.fireEventImpl(EventUtil.java:74)
	at javafx.base/com.sun.javafx.event.EventUtil.fireEvent(EventUtil.java:54)
	at javafx.base/javafx.event.Event.fireEvent(Event.java:199)
	at javafx.graphics/javafx.scene.Scene$MouseHandler.process(Scene.java:3987)
	at javafx.graphics/javafx.scene.Scene.processMouseEvent(Scene.java:1893)
	at javafx.graphics/javafx.scene.Scene$ScenePeerListener.mouseEvent(Scene.java:2711)
	at javafx.graphics/com.sun.javafx.tk.quantum.GlassViewEventHandler$MouseEventNotification.run(GlassViewEventHandler.java:411)
	at javafx.graphics/com.sun.javafx.tk.quantum.GlassViewEventHandler$MouseEventNotification.run(GlassViewEventHandler.java:1)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:400)
	at javafx.graphics/com.sun.javafx.tk.quantum.GlassViewEventHandler.lambda$2(GlassViewEventHandler.java:450)
	at javafx.graphics/com.sun.javafx.tk.quantum.QuantumToolkit.runWithoutRenderLock(QuantumToolkit.java:430)
	at javafx.graphics/com.sun.javafx.tk.quantum.GlassViewEventHandler.handleMouseEvent(GlassViewEventHandler.java:449)
	at javafx.graphics/com.sun.glass.ui.View.handleMouseEvent(View.java:551)
	at javafx.graphics/com.sun.glass.ui.View.notifyMouse(View.java:937)
	at javafx.graphics/com.sun.glass.ui.mac.MacView.notifyMouse(MacView.java:128)
```

Once that happens, the exception will get thrown every time, even after the correct CSS is entered.  The app must be restarted.

But we don't get an exception for a boolean value: `-fx-wrap-text:"yo";` or a double value like `.tooltip {  -fx-graphic-text-gap:"true"; }`.  The latter will correctly issue a warning
```
WARNING: Failed to set css [-fx-graphic-text-gap] on [DoubleProperty [bean: javafx.scene.control.Tooltip@468235b3, name: graphicTextGap, value: 4.0]] due to 'class java.lang.Boolean cannot be cast to class java.lang.Number (java.lang.Boolean and java.lang.Number are in module java.base of loader 'bootstrap')'
```

hjohn (commented):
Although the change is an improvement, I think a similar problem will still be present when CSS changes.  The `cssForced` flag I think needs to reset to `false` whenever a CSS styleable property of the tool tip is modified.  That would include all such properties...

This is a bit of a general issue in JavaFX (it really shouldn't be possible for any `Node` to be shown that hasn't had CSS processed for it), yet this can easily happen, and depending on how different the default style is versus the intended style, this can lead to a visual artifact (most noticable is for example a white background, when it is supposed to be black or transparent).

Perhaps we should open an issue to solve this for all controls in one go? A solution that would prevent **any** `Node` from being displayed without CSS applied?

Maran23:
> What I see is the very first time the tooltip appears it's using the old show-delay value of 1000ms. Any subsequent attempts the new delay is in effect.
> 
> Then, change back to 1000ms. Same thing happens: the first time the old timeout value is used instead of the new.

There seems to be a difference when using `setStyle` and a global stylesheet. I'm investigating.

Maran23:
> Perhaps we should open an issue to solve this for all controls in one go? A solution that would prevent **any** `Node` from being displayed without CSS applied?

Yes, that sounds like a good idea. Checking the calls to `applyCss`, it is mostly in table related classes and I remember that I already asked myself the question whether they are really needed there.

> Although the change is an improvement, I think a similar problem will still be present when CSS changes. The cssForced flag I think needs to reset to false whenever a CSS styleable property of the tool tip is modified. That would include all such properties...

Will check. `Tooltip` is unfortunately a corner case where we to get some CSS properties BEFORE we show it in order to correctly setup the animation timeline.

modules/javafx.controls/src/main/java/javafx/scene/control/Tooltip.java
```diff
@@ -170,6 +171,16 @@ public Tooltip(String text) {
         getStyleClass().setAll("tooltip");
     }
 
+    @Override
+    protected void show() {
```
Maran23, line 175: Another idea I was trying that worked fine is to have a new method inside `PopupWindow`, which we can call rather than `show()`:

```
protected final void applySceneStylesFromOwner(Window owner) {
    Scene sceneValue = getScene();
    final Scene ownerScene = getRootWindow(owner).getScene();
    if (ownerScene != null) {
        if (ownerScene.getUserAgentStylesheet() != null) {
            sceneValue.setUserAgentStylesheet(ownerScene.getUserAgentStylesheet());
        }
        sceneValue.getStylesheets().setAll(ownerScene.getStylesheets());
        if (sceneValue.getCursor() == null) {
            sceneValue.setCursor(ownerScene.getCursor());
        }
    }
}
```

The `show()` method inside `PopupWindow` is also doing excactly the code above, thats why I ended up calling it. 
But we can also think about just calling that code directly and not `show`, we just need to extract that part of the code and create a `protected` method with it.

Maran23:
keep it open.

andy-goryachev-oracle:
Please sync up with the latest master (should be a SOP for any long outstanding PRs - sorry for long delay!)

```
The method shutdown() in the type Util is not applicable for the arguments (Stage)	TooltipTest.java	Java Problem	/systemTests-test/java/test/robot/javafx/scene	line 238

```

modules/javafx.controls/src/main/java/javafx/scene/control/Tooltip.java
```diff
@@ -170,6 +171,16 @@ public Tooltip(String text) {
...
+        // The very first show call is just for us to do the correct CSS processing, so we ignore the request here.
+        if (!cssForced) {
```
andy-goryachev-oracle, line 177: I am *very* suspicious of this change.  Yes, it sort of works, but I wonder if it might backfire as it breaks the contract of show() and also depends on some assumptions ("The very first show call is just for us to do the correct CSS processing").

Would it be possible to try the applySceneStylesFromOwner() approach, maybe by moving the method to Utils to avoid changing the API?  This issue does not seem to be limited to Tooltip - there is a similar code in CustomColorDialog:88 for example.

What do you think?

Maran23: As written above, that would also be a possibility I explorer and that worked as well. Would be fine for me and I agree that this is rather unexpected. I will explore the other solution soon, when time permits.

andy-goryachev-oracle:
Strange: I still see the same issue described in
https://github.com/openjdk/jfx/pull/1394#issuecomment-1986520680
(the very first tooltip after updating stylesheet uses the old value)

Maran23:
> Strange: I still see the same issue described in [#1394 (comment)](https://github.com/openjdk/jfx/pull/1394#issuecomment-1986520680) (the very first tooltip after updating stylesheet uses the old value)

I don't know what is happening there, are you sure that there isn't a problem on your side? Testing `Tooltip` with a normal stylesheet added to the scene works perfectly fine for me.

andy-goryachev-oracle:
I am not sure what the difference might be.  The code in the MonkeyTester generates a stylesheet, encodes it in base64 data url and adds it to all scenes in the application (it also removes the old version, if any).

The stylesheet gets loaded - I can see it by slightly changed colors and the fact that the tooltip show delay gets changed on the second and any subsequent invocations.

Can you try it?

andy-goryachev-oracle:
with the latest monkey tester I see that updating the stylesheet does not update the showingDelay property immediately.  When the tooltip gets shown, I see the following output from the change listener added to this property:

```
showDelay=1000.0 ms
showDelay=100.0 ms
```

I suppose the second setting of 100ms (the value I actually set) happens too late or simply is ignored.

Maran23:
> with the latest monkey tester I see that updating the stylesheet does not update the showingDelay property immediately. When the tooltip gets shown, I see the following output from the change listener added to this property:
> 
> ```
> showDelay=1000.0 ms
> showDelay=100.0 ms
> ```
> 
> I suppose the second setting of 100ms (the value I actually set) happens too late or simply is ignored.

Ah, I think I know why. Same problem, we copy the stylesheet right before showing, so this is updated to late once again. We may need to do the update stuff always before showing.

modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
```diff
@@ -479,6 +477,23 @@ private void showImpl(final Window owner) {
...
+     * @param owner the owner {@link Window}
+     */
+    protected void applyStylesheetFromOwner(Window owner) {
+        Scene scene = getScene();
+        final Scene ownerScene = getRootWindow(owner).getScene();
+        if (ownerScene != null) {
+            if (ownerScene.getUserAgentStylesheet() != null) {
+                scene.setUserAgentStylesheet(ownerScene.getUserAgentStylesheet());
+            }
+            scene.getStylesheets().setAll(ownerScene.getStylesheets());
```
andy-goryachev-oracle, line 493: so if the application decides to set a particular style sheet on the tooltip scene, it will be ignored?  Should it be addAll()?  And if yes, should it check against duplicate entries?

modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
```diff
@@ -458,12 +458,10 @@ private void showImpl(final Window owner) {
         SceneHelper.parentEffectiveOrientationInvalidated(sceneValue);
 
         // RT-28447
+        applyStylesheetFromOwner(owner);
```
andy-goryachev-oracle, line 461: minor: Does this comment RT-28477 (https://bugs.openjdk.org/browse/JDK-8116444) applies to the change of to the code that follows?

tests/system/src/test/java/test/robot/javafx/scene/TooltipTest.java
```diff
@@ -71,19 +93,45 @@ public void testDefaultTooltip() {
...
+        tooltipShowTime = waitForTooltip();
+
+        assertTrue(tooltip.isShowing());
+
+        assertTooltipShowDelay(tooltipShowTime, 200, 100);
     }
 
     @Test
-    public void testSmallShowDelayTooltip() {
+    void testSmallShowDelayTooltip() {
```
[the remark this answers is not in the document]
Maran23: JUnit 5 does not require you to keep the methods public.
It was mostly done for better encapsulation, you can read more here, where also one of the JUnit Team responded. :)
https://stackoverflow.com/questions/55215949/why-junit-5-default-access-modifier-changed-to-package-private

modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
Maran23: Yes, but I did not change this behaviour, as this method is always called right before showing.
While this a bit weird, this is out of scope for at least this PR.

modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
Maran23: Either here or in the new method I think. What is your opinion?

andy-goryachev-oracle: I wasn't sure; I suppose you have it right.

I would have added the new JBS ticket id since RT- prefix is no more, but it is ok as it is.

andy-goryachev-oracle (approved):
LGTM

modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
Maran23: Changed to the JDK number, in case this lookup is needed in the future

andy-goryachev-oracle: thank you

andy-goryachev-oracle (approved):

Maran23:
Question: Since I add a new protected method to `PopupWindow` (`applyStylesheetFromOwner`), this needs a CSR, right?

Another approach would be to access this method via a Helper class, in this case maybe: `PopupWindowHelper`.

andy-goryachev-oracle:
> in this case maybe: `PopupWindowHelper`.

good point!  using a helper class would avoid CSR.

question is - do we really need to expose this method (for custom popup windows, for example), or not?

kevinrushforth:
I just took a quick look and yes, without changes, this would need a CSR. Using a Helper method would be one way to avoid it.

Also, I think a second reviewer would be good here.

/reviewers 2

Maran23:
> question is - do we really need to expose this method (for custom popup windows, for example), or not?

I can not really see why this is needed, other than maybe to not copy any style from the owner. 
It is hard to guess a usecase, so hiding this for now is better.

Maran23:
TIL how we can test things that involves animations like `Timeline`. Wrote a lot of unit tests for `Tooltip` as well now.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TooltipTest.java
```diff
@@ -542,16 +555,155 @@ public class TooltipTest {
...
+    }
+
+    @Test
+    public void testTooltipShouldBeShownAfterSetCssShowDelay() {
+        Rectangle rect = new Rectangle(0, 0, 100, 100);
+
+        stageLoader = new StageLoader(rect);
+        // Style: .tooltip { -fx-show-delay: 200ms; }
+        stageLoader.getStage().getScene().getStylesheets()
+                .add("data:base64,LnRvb2x0aXAgeyAtZngtc2hvdy1kZWxheTogMjAwbXM7IH0=");
```
andy-goryachev-oracle, line 657: this is clever.  can we generate the base64-encoded string on the fly?  something along the lines

```
return "data:text/css;base64," + Base64.getEncoder().encodeToString(b);
```

(here and elsewhere?)

tests/system/src/test/java/test/robot/javafx/scene/TooltipTest.java
```diff
@@ -73,12 +73,12 @@ private static void assertTooltipShowDelay(long tooltipShowTime, long expectedTi
...
 
     @BeforeAll
-    public static void initFX() {
+    static void initFX() {
         Util.launch(startupLatch, TestApp.class);
     }
 
     @AfterAll
-    public static void exit() {
+    static void exit() {
```
andy-goryachev-oracle, line 81: I still not convinced I like the concept of non-public methods in tests, since they are annotated.

Why do we need to do this?

modules/javafx.controls/src/test/java/test/javafx/scene/control/TooltipTest.java
Maran23: Thought about this well, didn't test yet

tests/system/src/test/java/test/robot/javafx/scene/TooltipTest.java
Maran23: Please check: https://stackoverflow.com/questions/55215949/why-junit-5-default-access-modifier-changed-to-package-private

It is essentially the good old 'why make something public when you do not need it?'. 
It follows the way encapsulation is meant to be, by lowering the visibility as far as you can for something that is not meant to be called from the outside.

Also static code analysis will check that and issue a code smell, which makes sense. Quoting Sonarlint here:

"
JUnit5 is more tolerant regarding the visibility of test classes and methods than JUnit4, which required everything to be public. Test classes and methods can have any visibility except private. It is however recommended to use the default package visibility to improve readability.

Test classes, test methods, and lifecycle methods are not required to be public, but they must not be private.
It is generally recommended to omit the public modifier for test classes, test methods, and lifecycle methods unless there is a technical reason for doing so – for example, when a test class is extended by a test class in another package. Another technical reason for making classes and methods public is to simplify testing on the module path when using the Java Module System.
— JUnit5 User Guide
"

modules/javafx.controls/src/test/java/test/javafx/scene/control/TooltipTest.java
andy-goryachev-oracle: see for example 
https://github.com/openjdk/jfx/blob/72701e6cb4095b8f5042f54ae6bb2c0cec446bcf/modules/javafx.graphics/src/test/java/test/javafx/scene/Node_transition_Test.java#L142

tests/system/src/test/java/test/robot/javafx/scene/TooltipTest.java
andy-goryachev-oracle: thank you for clarification!

there seems to be no downside to making the test methods non public, and some upside.
we might suggest doing so for every new test we write, similarly to the requirement to use junit5 instead of junit4.

andy-goryachev-oracle (approved):
LGTM.  If you want to change to base64 encoding on the fly, I'll re-approve (this PR still needs a second reviewer).

tests/system/src/test/java/test/robot/javafx/scene/TooltipTest.java
kevinrushforth: This seems like a good suggestion to me.

kevinrushforth:
The  new system test isn't stable on Windows. I ran it several times, and most of the time I see one or two failures. Here are the results from a couple of the test runs:

Test run 1:

```
TooltipTest > testShowDelayCssShowTooltipTwice() FAILED
    org.opentest4j.AssertionFailedError: 193 <= 150 ==> expected: <true> but was: <false>

TooltipTest > testSmallShowDelayCssTooltip() FAILED
    org.opentest4j.AssertionFailedError: 160 <= 150 ==> expected: <true> but was: <false>
```

Test run 2:

```
TooltipTest > testShowDelayCssShowTooltipTwice() FAILED
    org.opentest4j.AssertionFailedError: 165 <= 150 ==> expected: <true> but was: <false>

TooltipTest > testChangeShowDelayTooltip() FAILED
    org.opentest4j.AssertionFailedError: 151 <= 150 ==> expected: <true> but was: <false>
```

This is likely due to either inaccuracies in the timer itself or in measuring the elapsed time.

andy-goryachev-oracle:
> org.opentest4j.AssertionFailedError: 193 <= 150 ==> expected: <true> but was: <false>

I don't see `testShowDelayCssShowTooltipTwice()` checking for the value of 150...
Where did it come from?

kevinrushforth:
> > org.opentest4j.AssertionFailedError: 193 <= 150 ==> expected:  but was:
> 
> I don't see `testShowDelayCssShowTooltipTwice()` checking for the value of 150... Where did it come from?

It's 100 (expected value) plus 50 (maximum delta). Here is the call stack:

```
L72:         assertTrue(tooltipShowTime <= maximumTime, tooltipShowTime + " <= " + maximumTime);
L64:         assertTooltipShowDelay(tooltipShowTime, expectedTime, 50);
L209:        assertTooltipShowDelay(tooltipShowTime, 100);
```

tests/system/src/test/java/test/robot/javafx/scene/TooltipTest.java
```diff
@@ -0,0 +1,295 @@
...
+    static CountDownLatch tooltipShownLatch;
+    static Robot robot;
+    static Button button;
+    static Tooltip tooltip;
+
+    static volatile Stage stage;
+    static volatile Scene scene;
+
+    private static void assertTooltipShowDelay(long tooltipShowTime, long expectedTime) {
+        assertTooltipShowDelay(tooltipShowTime, expectedTime, 50);
```
kevinrushforth, line 64: If I change this to 100, then all tests pass most of the time. So maybe 150 or 200 would be a better maximum delta?

If you do make this change, then you will need to find all of the three-arg calls with a maxDifference less than that and change them to the two-arg version.

tests/system/src/test/java/test/robot/javafx/scene/TooltipTest.java
```diff
@@ -0,0 +1,295 @@
...
+
+        long tooltipShowTime = waitForTooltip();
+
+        assertTrue(tooltip.isShowing());
+
+        // Since the stylesheet needs to be processed we need a bigger timeout here.
+        long expectedTime = 30;
+        long maximumTime = expectedTime + 100;
+
+        assertTooltipShowDelay(tooltipShowTime, expectedTime, maximumTime);
```
kevinrushforth, line 109: The third argument is a maximum difference, not a max time. So unless there is some reason to add the expected time twice, I would just use a larger constant delta.. You might even be able to get rid of the third arg depending on what we eventually go with for the default max delta.

tests/system/src/test/java/test/robot/javafx/scene/TooltipTest.java
```diff
@@ -0,0 +1,295 @@
...
+    void testCssStylesheetChangeTooltip() {
+        scene.getStylesheets().add(getClass().getResource("tooltip.css").toExternalForm());
+
+        assertFalse(tooltip.isShowing());
+
+        long tooltipShowTime = waitForTooltip();
+
+        assertTrue(tooltip.isShowing());
+
+        assertTooltipShowDelay(tooltipShowTime, 30, 100);
```
kevinrushforth, line 122: This is an example of the three-arg call where you can eliminate the third arg if you make the default > 100.

kevinrushforth (commented):
I left a few inline comments, including a possible suggestion to fix the intermittent test failures.

tests/system/src/test/java/test/robot/javafx/scene/TooltipTest.java
andy-goryachev-oracle: even with max=200 we might see occasional failures due to load or other circumstances.

I wonder a better approach would be to check whether the measurement of the delay falls sufficiently outside of the default 1000 ms (which in itself might be more than that)?

kevinrushforth: We have many tests that will occasionally fail with a heavily loaded system.

The question is whether we can come up with a max delta that we can reliably use that is less than the difference between the default value and the test value. Otherwise, we can't distinguish them.

On two different test systems, both of which fail pretty consistently with  50 msec, I see a 100% pass rate over several tries with 150 msec. More testing is needed.

andy-goryachev-oracle: Another possibility is the code that measures the time it takes to show the tooltip.

The current code uses Util.waitForLatch(), L244 which was written for a different use case and actually introduces a small measurement error as it includes the time needed for the context switch.  Perhaps a better solution would be to note the timestamp in L271 where the tooltip listener gets called (similar to the way the start moment is captured in L241)

kevinrushforth: Good catch. Capturing the current time in the listener will be more accurate than doing it after the latch wait.

Maran23: Good catch indeed. Will test and report back accordingly.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TooltipTest.java
Maran23: Ah, looks good. Maybe in the future we can create a Utility function for this.

Maran23:
The Tooltip timing looks much better now in the tests. I decided to use 80ms as a threshhold. In my testing, I got a maximum difference of ~35 ms.

Also merged in master.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TooltipTest.java
```diff
@@ -706,4 +706,8 @@ private void assertTooltipHiddenAfter(Rectangle rect, int millis) {
         assertFalse(toolTip.isShowing());
     }
 
+    private String toBase64(String css) {
```
andy-goryachev-oracle, line 709: suggestion:
```
private static String encode(String css) {
  return "data:base64," + Base64.getUrlEncoder().encodeToString(css.getBytes(StandardCharsets.UTF_8));
}
```

tests/system/src/test/java/test/robot/javafx/scene/TooltipTest.java
```diff
@@ -232,18 +225,17 @@ private long waitForTooltip() {
...
         assertFalse(tooltip.isShowing());
 
-        AtomicLong time = new AtomicLong();
         Util.runAndWait(() -> {
             Window window = scene.getWindow();
             robot.mouseMove(
                     window.getX() + scene.getX() + button.getLayoutX() + button.getLayoutBounds().getWidth() / 2,
                     window.getY() + scene.getY() + button.getLayoutY() + button.getLayoutBounds().getHeight() / 2);
-            time.set(System.currentTimeMillis());
+            tooltipStartTime = System.currentTimeMillis();
```
andy-goryachev-oracle, line 233: it is better to use `System.nanoTime()` instead of `currentTimeMillis()` for this kind of measurements because it is not affected by the updates to the current time (such as time sync).

(since there are plenty of other places where we use `currentTimeMillis()` it's probably ok - very unlikely for the test to hit this scenario)

modules/javafx.controls/src/test/java/test/javafx/scene/control/TooltipTest.java
kevinrushforth: I agree that it could be static, although I prefer the name `toBase64`.

andy-goryachev-oracle: my comment was to avoid replicating "data:base64,", and with that, it is not base64 anymore, but base-64-encoded url.

the best place for this method is probably in some utility class so it can be reused.

tests/system/src/test/java/test/robot/javafx/scene/TooltipTest.java
kevinrushforth: That seems like a good overall test cleanup that could be done in a follow-up issue.

kevinrushforth:
> The Tooltip timing looks much better now in the tests. I decided to use 80ms as a threshhold. In my testing, I got a maximum difference of ~35 ms.
> 
> Also merged in master.

Thanks. I'll test and re-review.

kevinrushforth:
I ran it on our CI Linux and macOS systems and it passed on all of them. It also passed on my local Linux system.

On my Windows machine I still get the following test failure almost every time I run it:

```
TooltipTest > testShowDelayCssShowTooltipTwice() FAILED
    org.opentest4j.AssertionFailedError: 260 <= 180 ==> expected: <true> but was: <false>
        at app//org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:40)
        at app//org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:210)
        at app//test.robot.javafx.scene.TooltipTest.assertTooltipShowDelay(TooltipTest.java:69)
        at app//test.robot.javafx.scene.TooltipTest.testShowDelayCssShowTooltipTwice(TooltipTest.java:202)
```

I've seen the actual result as high as 390 one time. The other times were in the 200 - 260 range.

Have you run it on Windows? If it's only my system, we could file a bug and look at it later. We might consider skipping that test on Windows using `@Ignore` and referencing that follow-on bug.

Maran23:
> Have you run it on Windows?

Yes, on Windows 10 and Windows 11. Executed the test multiple time now again, always around 110 - 140 ms for me. Weird.

kevinrushforth:
With your latest change, the test now passes on my Windows 11 system. Maybe the resetting of `tooltipStartTime and `tooltipShownTime` fixed it? I'll look more closely in a bit.

Maran23:
> With your latest change, the test now passes on my Windows 11 system. Maybe the resetting of `tooltipStartTime` and `tooltipShownTime` fixed it? I'll look more closely in a bit.

I would not expect that, but maybe it has something to do with it. 
I did this change so that the test is more safe in case something is not triggered the way it should. But maybe it had a positive side effect, although not 100% sure.

kevinrushforth:
I looked at the code, and you're right, it doesn't seem like it would have mattered. Regardless, it now passes on my system reliably. I also reverted your fix and verified that the new test catches the failures. So all good from my point of view.

kevinrushforth (approved):

andy-goryachev-oracle (approved):
lgtm.
I'll create the tickets for creating utility methods that use base64 urls and system nano time.

kevinrushforth:
@Maran23 This is ready to integrate. If you can do it in the next 30 minutes, it will make JavaFX 23 before the RDP1 fork. If it is after the fork, it might be a candidate to backport to jfx23.

Maran23:
> If you can do it in the next 30 minutes, it will make JavaFX 23 before the RDP1 fork. If it is after the fork, it might be a candidate to backport to jfx23.

Damn, missed the time. I will keep it open until tomorrow, in case someone else want to review or just check the PR. 
If we consider this a candidate for backporting, then I can certainly do the request after merging as well! :)

modules/javafx.controls/src/test/java/test/javafx/scene/control/TooltipTest.java
andy-goryachev-oracle: created https://bugs.openjdk.org/browse/JDK-8336334

PR #1395, opened 2024-03-08
8314968: Public InputMap

andy-goryachev-oracle:
## Summary

Add the **InputMap** property in the Control class which enables customization of controls with the application key mappings and event handlers, and to guarantee specific order of event handlers invocation and priority of application mappings over those created by the skin.

Add **SkinInputMap** and optional **BehaviorBase** classes to simplify creation of skins for existing and new controls.

## Goals

The goals of this proposal are:

- allow for customization of a control behavior by changing the existing or adding new key mappings
- support dynamic modification of key mappings
- allow for accessing the default functionality even when it was overwritten by the application
- allow for reverting customization to the default implementation
- guarantee priorities between the application and the skin event handlers and key mappings
- allow for *gradual* migration of the existing controls to use the new InputMap
- support stateful and stateless (fully static) behavior implementations

## Non-Goals

It is not the goal of this proposal:

- to require specific base class for the behavior implementations
- to allow for complete decoupling of the skin from the behavior
- to make the legacy internal InputMap class public
- to make existing core behaviors public

Complete proposal:
https://github.com/andy-goryachev-oracle/Test/blob/ag.jep.input.map.v2/doc/InputMap/InputMapV2.md

### DIFF
--- a/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/ColorPickerBehavior.java
+++ b/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/ColorPickerBehavior.java
@@ -27,7 +27,6 @@ package com.sun.javafx.scene.control.behavior;
 
 import javafx.scene.control.ColorPicker;
 import javafx.scene.control.PopupControl;
-
 import javafx.scene.paint.Color;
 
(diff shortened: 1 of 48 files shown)


andy-goryachev-oracle:
keep alive

PR #1403, opened 2024-03-14
8207379: Robot screen capture test fails with HiDPI at some screen locations

lukostyra:
There were two different problems with this test's stability and HiDPI and both were fixed with this change.

The whole reason for this tests lack of stability comes with how Windows calculates window position when using HiDPI. When window's X/Y coordinates are not provided by the application, they are (generally speaking) randomly selected by Windows. X/Y are picked as two integers, and afterwards the HiDPI scaling is applied. With fractional scaling like 125% this can cause X/Y coordinates to have fractional values.

When performing a screen capture, Robot takes a generous estimate of window's position and dimensions. This is done by taking provided X/Y coordinates (in case of this test, stage's X/Y coords), multiplying them by scaling (ex. 1.25) and then flooring the result. Similar estimation is done to opposite coordinates - we take X/Y coords, add width/height to them respectively, multiply them by scaling and ceiling them. As a result, we have minX/Y and maxX/Y coords of capture region, which is used to calculate capture region's width and height. The first problem with test's stability is right here - the test did not take this calculation and floor/ceil operations into account, which with appropriately randomized window coordinates could cause discrepancy between expected width/height and actual. This logic that is used by `getScreenCapture()` seems to be necessary and good, so to remedy the problem I replicated those calculations on test side.

Another problem came after that - fractional X/Y values caused capture to not always line up with pixel layout, which made the 1-pixel border of capture sometimes incorrect (an average of stage's MAGENTA color and whatever background happened to be under the displayed stage). This means we cannot exactly expect those pixels to have exactly MAGENTA color when the system has fractional HiDPI enabled. To help that, the test now doesn't check for correctness of the 1-pixel border of captured image. I feel this still gives us enough confidence that the Robot screen capture works properly, while stabilizing the test on all systems.

I briefly considered an alternative approach to hard-set X/Y values but doing it this way gives us a bit more robustness, makes the test independent of Stage's X/Y coordinates which matter quite a lot.

Verified this change on macOS and especially on Windows - ran the test on Windows 50 times each on 125% and 100% scaling and noticed no intermittent failures.

---------
### Progress

### Issue
 * [JDK-8207379](https://bugs.openjdk.org/browse/JDK-8207379): Robot screen capture test fails with HiDPI at some screen locations (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1403/head:pull/1403` \
`$ git checkout pull/1403`

`$ git checkout pull/1403` \
`$ git pull https://git.openjdk.org/jfx.git pull/1403/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1403`

`$ git pr show -t 1403`

</details>

<a href="https://git.openjdk.org/jfx/pull/1403.diff">https://git.openjdk.org/jfx/pull/1403.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1403#issuecomment-1997819939)

### DIFF
--- a/tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
@@ -755,12 +755,12 @@ public class RobotTest {
         // this is redundant.
         // Below calculations follow how getScreenCapture should calculate screen capture dimensions. This
         // is to make this code consistent and stable on HiDPI systems.
-        int stageX = (int)stage.getX();
-        int stageY = (int)stage.getY();
-        int shouldBeMinX = (int)Math.floor(stageX * screenScaleX);
-        int shouldBeMinY = (int)Math.floor(stageY * screenScaleY);
-        int shouldBeMaxX = (int)Math.ceil((stageX + WIDTH) * screenScaleX);
-        int shouldBeMaxY = (int)Math.ceil((stageY + HEIGHT) * screenScaleY);
+        int stageX = (int) stage.getX();
+        int stageY = (int) stage.getY();
+        int shouldBeMinX = (int) Math.floor(stageX * screenScaleX);
+        int shouldBeMinY = (int) Math.floor(stageY * screenScaleY);
+        int shouldBeMaxX = (int) Math.ceil((stageX + WIDTH) * screenScaleX);
+        int shouldBeMaxY = (int) Math.ceil((stageY + HEIGHT) * screenScaleY);
         int shouldBeWidth = shouldBeMaxX - shouldBeMinX;
         int shouldBeHeight = shouldBeMaxY - shouldBeMinY;
         Assert.assertEquals((double) shouldBeWidth, screenCaptureNotScaledToFit.get().getWidth(), 0.0001);
--- a/tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
@@ -725,10 +725,6 @@ public class RobotTest {
 
     @Test
     public void testScreenCapture() throws Exception {
-        if (PlatformUtil.isWindows() && Screen.getPrimary().getOutputScaleX() > 1) {
-            // Mark this test as unstable on Windows when HiDPI scale is more than 100%
-            Assume.assumeTrue(Boolean.getBoolean("unstable.test")); // JDK-8207379
-        }
         CountDownLatch setSceneLatch = new CountDownLatch(1);
         Pane pane = new StackPane();
         InvalidationListener invalidationListener = observable -> setSceneLatch.countDown();
--- a/tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
@@ -753,10 +753,24 @@ public class RobotTest {
         // Should be scaled to the primary screen x and y scales. Note that screenCaptureScaledToFit and
         // screenCaptureNotScaledToFit will be the same if screenScaleX = screenScaleY = 1.0 and in that case
         // this is redundant.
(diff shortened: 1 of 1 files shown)


tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
```diff
@@ -757,19 +753,34 @@ public void testScreenCapture() throws Exception {
         // Should be scaled to the primary screen x and y scales. Note that screenCaptureScaledToFit and
         // screenCaptureNotScaledToFit will be the same if screenScaleX = screenScaleY = 1.0 and in that case
         // this is redundant.
-        Assert.assertEquals((double) WIDTH * screenScaleX, screenCaptureNotScaledToFit.get().getWidth(), 0.0001);
-        Assert.assertEquals((double) HEIGHT * screenScaleY, screenCaptureNotScaledToFit.get().getHeight(), 0.0001);
-        for (int x = 0; x < WIDTH * screenScaleX; x++) {
-            for (int y = 0; y < HEIGHT * screenScaleY; y++) {
+        // Below calculations follow how getScreenCapture should calculate screen capture dimensions. This
+        // is to make this code consistent and stable on HiDPI systems.
+        int stageX = (int)stage.getX();
```
karthikpandelu, line 758: Minor: Do you think space should be added after `(int)` similar to `(double)`?

karthikpandelu (commented):
Tested the changes in Mac, Windows and Linux. The changes fixes the issues in Mac and Windows.
I ran the test in Linux VM and it failed with and without the fix with error:
```
RobotTest > testScreenCapture FAILED
    junit.framework.AssertionFailedError: expected: rgba(255,0,255,255) but was: rgba(0,0,0,255)
        at test.robot.javafx.scene.RobotTest.assertColorEquals(RobotTest.java:848)
        at test.robot.javafx.scene.RobotTest.testScreenCapture(RobotTest.java:774)

```
Changing the scale also did not make any difference. Not sure if it is related to anything in VM.
It is a Ubuntu 22.04 VM

Added a minor comment inline.

tests/system/src/test/java/test/robot/javafx/scene/RobotTest.java
lukostyra: It would be good for consistent formatting. I'll add those.

lukostyra:
@karthikpandelu I just tried this on Ubuntu 22.04 VM and everything seems to work fine. (0,0,0,255) returned from capture suggests capture failed - could it be your VM is running on Wayland instead of X11?

brcolow:
Thanks for fixing this - was a thorn in my side when developing the Robot API.

kevinrushforth:
> @karthikpandelu I just tried this on Ubuntu 22.04 VM and everything seems to work fine. (0,0,0,255) returned from capture suggests capture failed - could it be your VM is running on Wayland instead of X11?

I also just ran this, and it runs fine for me on an Ubuntu 22.04 VM. Since Wayland is the default in 22.04, I also suspect that Karthik was running in that mode. You can check the value of the `XDG_SESSION_TYPE` env var to see. All robot tests that read the screen will fail on Wayland until [JDK-8326712](https://bugs.openjdk.org/browse/JDK-8326712) is fixed (should be done reasonably soon, sometime in the JavaFX 23 time frame).

kevinrushforth (approved):
Fix looks good to me. I confirm that the test no longer fails on my Windows system (running at 1.25x scale).

karthikpandelu:
> I also suspect that Karthik was running in that mode. You can check the value of the `XDG_SESSION_TYPE` env var to see. All robot tests that read the screen will fail on Wayland until [JDK-8326712](https://bugs.openjdk.org/browse/JDK-8326712) is fixed (should be done reasonably soon, sometime in the JavaFX 23 time frame).

Yes it was because of Wayland. The test looks good now.

karthikpandelu (approved):

PR #1411, opened 2024-03-20
8324939: Editable TableView loses focus after commit

jperedadnr:
This PR fixes the issue that after committing an edit on a ListView/TreeView/TableView/TreeTableView control, the control might lose the focus unexpectedly.

For that, it refactors the `ControlUtils::requestFocusOnControlOnlyIfCurrentFocusOwnerIsChild` method, in order to check if the control (`ListView`, `TreeView`, `TableView`, `TreeTableView`) should request the focus _before_ the actual focus owner (which could be the control added to the cell to edit its content, like a `TextField`) is removed from the cell, so the `Control::requestFocus` call, if needed, can be still invoked after the edit commit is done (as it was done before).

By adding `ControlUtils::controlShouldRequestFocusIfCurrentFocusOwnerIsChild` the `Cell::commitEdit` implementations can now query if the control should have the focus, after `super.commitEdit(newValue);` but before firing the `CellEditEvent` and calling `updateItem()`, and if the result is true, then request focus after the edit commit ends (like it was done before).

Two new tests per control have been included, to verify that the focus remains at the control, one for edit cancel (this passes before and after the proposed changes), one for edit commit (this fails before and passes after including the proposed fix).

---------
### Progress

### Issue
 * [JDK-8324939](https://bugs.openjdk.org/browse/JDK-8324939): Editable TableView loses focus after commit (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1411/head:pull/1411` \
`$ git checkout pull/1411`

`$ git checkout pull/1411` \
`$ git pull https://git.openjdk.org/jfx.git pull/1411/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1411`

`$ git pr show -t 1411`

</details>

<a href="https://git.openjdk.org/jfx/pull/1411.diff">https://git.openjdk.org/jfx/pull/1411.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1411#issuecomment-2009304221)

### DIFF
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeViewTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeViewTest.java
@@ -2359,9 +2359,7 @@ public class TreeViewTest {
         treeView.requestFocus();
         assertTrue(treeView.isFocused());
 
-        VirtualFlowTestUtils.BLOCK_STAGE_LOADER_DISPOSE = true;
         TreeCell cell = (TreeCell) VirtualFlowTestUtils.getCell(treeView, 0);
-        VirtualFlowTestUtils.BLOCK_STAGE_LOADER_DISPOSE = false;
         assertTrue(cell.getSkin() instanceof TreeCellSkin);
         assertNull(cell.getGraphic());
         assertEquals("John", cell.getText());
@@ -2405,9 +2403,7 @@ public class TreeViewTest {
         treeView.requestFocus();
         assertTrue(treeView.isFocused());
 
-        VirtualFlowTestUtils.BLOCK_STAGE_LOADER_DISPOSE = true;
         TreeCell cell = (TreeCell) VirtualFlowTestUtils.getCell(treeView, 0);
-        VirtualFlowTestUtils.BLOCK_STAGE_LOADER_DISPOSE = false;
         assertTrue(cell.getSkin() instanceof TreeCellSkin);
         assertNull(cell.getGraphic());
         assertEquals("John", cell.getText());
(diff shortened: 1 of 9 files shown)


andy-goryachev-oracle:
> Can you review this?

in the queue; yes.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeViewTest.java
```diff
@@ -2343,6 +2344,98 @@ private void test_rt_37853(boolean replaceRoot) {
...
+        treeView.setShowRoot(false);
+
+        TreeItem<String> item = new TreeItem<>("John");
+        treeView.getRoot().getChildren().add(item);
+
+        StageLoader sl = new StageLoader(new Button(), treeView);
+        treeView.requestFocus();
+        assertTrue(treeView.isFocused());
+
+        VirtualFlowTestUtils.BLOCK_STAGE_LOADER_DISPOSE = true;
```
Maran23, line 2362: This should not be needed since you are creating a `stageLoader` above. This seems to be only used when no `stageLoader` was created before

jperedadnr: Thanks @Maran23, I've removed it now from the added tests.

andy-goryachev-oracle (approved):

modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeViewTest.java
Maran23: > I wish there was a more reliable way of implementing getCell() without resorting to undocumented flags like BLOCK_STAGE_LOADER_DISPOSE.

There is, just add the virtual container to a Stage. Since the cells are only created inside the skin, which is only created when the component is inside the scene tree.

Maran23 (approved):
Looks good to me too. 
Fix and the rationale behind makes sense. 
Something I also noticed on virtual containers.

PR #1413, opened 2024-03-20
8319555: [TestBug] Utility for creating instruction window for manual tests

andy-goryachev-oracle:
## ManualTestWindow

This facility provides the base class for manual tests which displays the test instructions,
the UI under test, and the Pass/Fail buttons.

Example:

```java
public class ManualTestExample extends ManualTestWindow {
    public ManualTestExample() {
        super(
            "Manual Test Example",
            """
            Instructions:
            1. you will see a button named "Test"
            2. press the button
            3. verify that the button can be pressed""",
            400, 250
        );
    }

    public static void main(String[] args) throws Exception {
        launch(args);
    }

    @Override
    protected Node createContent() {
        return new Button("Test");
    }
}
```

Resulting application window:

ManualTestWindow

Readme:

https://github.com/andy-goryachev-oracle/jfx/blob/8319555.manual/tests/manual/util/README.md

@prrace 's test EmojiTest has been converted to use the new test window as a demonstration (also fixed the Eclipse project so it works now).

Q: What other features can be added to the test window?

Edit: the sources are left in their original place at the root of the project.

---------
### Progress

### Warning
&nbsp;⚠️ Patch contains a binary file (tests/manual/util/doc/ManualTestWindow.png)

### Issue
 * [JDK-8319555](https://bugs.openjdk.org/browse/JDK-8319555): [TestBug] Utility for creating instruction window for manual tests (**Enhancement** - P4)

### Reviewers
 * [Alexander Matveev](https://openjdk.org/census#almatvee) (@sashamatveev - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1413/head:pull/1413` \
`$ git checkout pull/1413`

`$ git checkout pull/1413` \
`$ git pull https://git.openjdk.org/jfx.git pull/1413/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1413`

`$ git pr show -t 1413`

</details>

<a href="https://git.openjdk.org/jfx/pull/1413.diff">https://git.openjdk.org/jfx/pull/1413.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1413#issuecomment-2012828693)


andy-goryachev-oracle:
> I'm not sure how much value there is in moving the files under a "src" directory

Thank you Kevin for the feedback!  I had problems configuring the project with sources being in the root, but will take another look.

andy-goryachev-oracle:
@mstr2 @jperedadnr @sashamatveev @aghaisas 
would you gentlemen be interested in reviewing this?

kevinrushforth:
This could use a second pair of eyes. I don't have time right now, but I'll add a couple comments.

/reviewers 2

tests/manual/util/README.md
```diff
@@ -0,0 +1,33 @@
...
+public class SampleManualTest {
+     public static void main(String[] args) throws Exception {
+         ManualTestWindow.builder().
+             title("Sample Manual Test").
+             instructions(
+                 """
+                 Provide
+                 multi-line instructions here.
+                 """
+             ).
```
sashamatveev, line 19: Add `size(1200, 800).` to example. I think adjusting size will be common enough and no need to look at `ManualTestWindow` to figure it out.

tests/manual/util/README.md
```diff
@@ -0,0 +1,33 @@
...
+             buildAndRun();
+     }
+
+     private static Node createTestUI() {
+         return new Label("Test UI");
+     }
+}
+```
+
+Resulting application window:
```
sashamatveev, line 30: Resulting application window does not look like it is from example above. Not sure if it is important.

tests/manual/text/EmojiTest.java
```diff
@@ -23,64 +23,42 @@
...
-        transparent background pixels.
-        There are 3 different ways it is displayed to verify
-        1) Text node. 2) Label control, 3) TextField Control
-        Press the Pass or Fail button as appropriate and the test will exit.
-        If what you see is not explained here, ask before filing a bug.
-
-
-        """;
-
+public class EmojiTest {
```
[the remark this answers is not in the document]
andy-goryachev-oracle: Everything, including creating an instance of Application, is taken care of by the `ManualTestWindow`.  This simplifies the test code greatly.

What is the reason that the test must extend `Application`?

tests/manual/util/src/com/oracle/util/testing/ManualTestWindow.java
```diff
@@ -0,0 +1,279 @@
...
+ * JavaFX application thread.
+ * <p>
+ * The {@link Builder#ui(Supplier)} method accepts interfaces which create the test Node.
+ * For tests that require multiple windows, these windows can be created by the code specified
+ * with {@link Builder#runAfter(Runnable)}.
+ */
+// TODO timeout
+// TODO log area
+// TODO screenshots on failure?
+// TODO initial position?
```
sashamatveev, line 89: In general I do not like TODO in comments. You can file follow up issue to improve/extend `ManualTestWindow` if needed.

tests/manual/util/src/com/oracle/util/testing/ManualTestWindow.java
```diff
@@ -0,0 +1,279 @@
...
+                instructionField,
+                cp,
+                buttons
+            );
+            vb.setPadding(new Insets(10));
+            VBox.setVgrow(cp, Priority.ALWAYS);
+
+            Scene scene = new Scene(vb);
+            stage.setWidth(width == 0 ? 800 : width);
+            stage.setHeight(height == 0 ? 600 : height);
```
sashamatveev, line 250: I think `width` and `height` is `double` and not `int`. Also, lets not change it. If test wants 0, then pass 0.

tests/manual/util/README.md
andy-goryachev-oracle: good point!

tests/manual/text/EmojiTest.java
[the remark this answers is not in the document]
sashamatveev: For example if test needs to access `getParameters()` or `getHostServices ()` from `Application` how it can be done? I think it might be better to extend `ManualTestWindow` from `Application` and require test itself to extend `ManualTestWindow`. In this case tests will extend `Application`.

tests/manual/util/README.md
andy-goryachev-oracle: Not really important.

Apart from being resized to decrease the size of the PNG image, it looks exactly the same.  Are you running it on a different platform?  How does it look, anything appears missing (can you post a screenshot)?

sashamatveev: If you run `EmojiTest`, then yes it will look correct, but if you run `SampleManualTest` it will not. Since you added this image to `SampleManualTest` as resulting image of `SampleManualTest`. Look at [ReadMe.md](https://github.com/openjdk/jfx/blob/a106587a69859966e668233a9d89539c3438fed7/tests/manual/util/README.md).

andy-goryachev-oracle: indeed!  thanks!

andy-goryachev-oracle:
Thank you all for comments and suggestions!
Changed to extend Application.  See if you like this better.

tests/manual/util/src/com/oracle/util/testing/ManualTestWindow.java
```diff
@@ -0,0 +1,203 @@
...
+
+        Region fill = new Region();
+
+        /*
+        Button screenshotButton = new Button("Screenshot");
+        screenshotButton.setMinWidth(100);
+        screenshotButton.setOnAction((ev) -> {
+        });
+        screenshotButton.setDisable(true);
+        */
```
sashamatveev, line 130: I will suggest to delete unused code.

tests/manual/util/src/com/oracle/util/testing/ManualTestWindow.java
```diff
@@ -0,0 +1,203 @@
...
+
+        Button passButton = new Button("✔ Pass");
+        passButton.setMinWidth(100);
+        passButton.setOnAction((ev) -> {
+            Platform.exit();
+        });
+
+        HBox buttons = new HBox(
+            10,
+            //screenshotButton,
```
sashamatveev, line 148: Same here.

sashamatveev (commented):
Only minor comment about unused code. Otherwise looks fine to me.

andy-goryachev-oracle:
it's in the test code, but ok - removed commented out code.  thanks!

sashamatveev (approved):

andy-goryachev-oracle:
keep-alive

mstr2:
I run manual tests on the command line like so:
```
java --module-path=$env:JFX_SDK --add-modules=javafx.controls ./tests/manual/events/PlatformPreferencesChangedTest.java
```

This doesn't work with the new utility class, since it is outside of the class file hierarchy.
Short of using a build system, how would I run a manual test with this new class?

andy-goryachev-oracle:
But this is a good question: is there a better way to organize the tests to make it easier to use common code?

mstr2:
> > Short of using a build system, how would I run a manual test with this new class?
> 
> I am running from within Eclipse. Here is the command line (remove newlines) it launches the EmojiTest. I think the key is to add the location of compiled classes via `-classpath`:

Yes this will work, but it requires me to compile `ManualTestWindow.java` before. That's quite a lot of heavy-lifting without tooling support (like you get from Eclipse). Maybe we should integrate the manual tests into the Gradle build, so that running a manual test might be as easy as invoking the application plugin's `:run` task for the manual test.

kevinrushforth:
We need to provide a build script for any manual test that isn't self-contained in a single directory (see my earlier comment). For example, we have a build script for MonkeyTester and FXMediaPlayer.

As a separate step, we could make it available as part of the gradle build, as long as we can do it without adding individual tests to build.gradle (sort of like we do for apps), but the previous is the minimum.

This should probably be put on the back burner until the next test sprint.

andy-goryachev-oracle:
moving to DRAFT until the next test sprint (after jfx23 ships).

andy-goryachev-oracle:
keep-alive

andy-goryachev-oracle:
this PR will have to wait until the next test sprint.

PR #1420, opened 2024-03-21
8328746: Remove unused imports in demo apps

andy-goryachev-oracle:
Using Eclipse IDE to remove unused imports in **demo apps** (3D, Ensemble, etc.) and update the copyright year to 2024. Using wildcard for more than 10 static imports.

**PROBLEM**
The code in `graphics/build/gensrc/jsl-decora/` is generated and cannot be fixed.
This means this warning **must be suppressed** in IDEs until the generator script is changed not to emit unused imports.

--

This is a trivial change (though fairly large), 1 reviewer is probably enough.

### DIFF
--- a/apps/samples/Ensemble8/src/app/java/ensemble/samplepage/Description.java
+++ b/apps/samples/Ensemble8/src/app/java/ensemble/samplepage/Description.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008, 2014, Oracle and/or its affiliates.
+ * Copyright (c) 2008, 2024, Oracle and/or its affiliates.
  * All rights reserved. Use is subject to license terms.
  *
  * This file is available and licensed under the following license:
@@ -31,26 +31,28 @@
  */
 package ensemble.samplepage;
 
-import ensemble.EnsembleApp;
-import ensemble.PlatformFeatures;
-import ensemble.SampleInfo;
-import ensemble.SampleInfo.URL;
-import ensemble.generated.Samples;
+import static ensemble.samplepage.SamplePage.INDENT;
+import static ensemble.samplepage.SamplePageContent.title;
+import javafx.application.ConditionalFeature;
+import javafx.application.Platform;
 import javafx.event.ActionEvent;
-import javafx.event.EventHandler;
 import javafx.geometry.Insets;
-import javafx.scene.control.*;
+import javafx.scene.control.ContentDisplay;
+import javafx.scene.control.Hyperlink;
+import javafx.scene.control.Label;
+import javafx.scene.control.ScrollPane;
+import javafx.scene.control.Tooltip;
 import javafx.scene.image.Image;
 import javafx.scene.image.ImageView;
 import javafx.scene.layout.ColumnConstraints;
 import javafx.scene.layout.GridPane;
 import javafx.scene.layout.VBox;
 import javafx.scene.text.Text;
-import javafx.util.Callback;
-import static ensemble.samplepage.SamplePage.INDENT;
-import static ensemble.samplepage.SamplePageContent.title;
-import javafx.application.ConditionalFeature;
-import javafx.application.Platform;
+import ensemble.EnsembleApp;
+import ensemble.PlatformFeatures;
+import ensemble.SampleInfo;
+import ensemble.SampleInfo.URL;
+import ensemble.generated.Samples;
 
 /**
  * Description Section on Sample Page
--- a/apps/samples/Ensemble8/src/samples/java/ensemble/samples/graphics2d/puzzle/Piece.java
+++ b/apps/samples/Ensemble8/src/samples/java/ensemble/samples/graphics2d/puzzle/Piece.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008, 2014, Oracle and/or its affiliates.
+ * Copyright (c) 2008, 2024, Oracle and/or its affiliates.
  * All rights reserved. Use is subject to license terms.
  *
  * This file is available and licensed under the following license:
@@ -31,7 +31,6 @@
  */
 package ensemble.samples.graphics2d.puzzle;
 
-import javafx.event.EventHandler;
 import javafx.geometry.Point2D;
 import javafx.scene.Parent;
 import javafx.scene.effect.DropShadow;
@@ -43,9 +42,10 @@ import javafx.scene.shape.Circle;
 import javafx.scene.shape.Ellipse;
 import javafx.scene.shape.Rectangle;
 import javafx.scene.shape.Shape;
-   /**
-     * Node that represents a puzzle piece
-     */
+
+/**
+ * Node that represents a puzzle piece
+ */
 public class Piece extends Parent{
         public static final int SIZE = 100;
         private final double correctX;
--- a/apps/samples/3DViewer/src/main/java/com/javafx/experiments/importers/obj/MtlReader.java
+++ b/apps/samples/3DViewer/src/main/java/com/javafx/experiments/importers/obj/MtlReader.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2014, Oracle and/or its affiliates.
+ * Copyright (c) 2010, 2024, Oracle and/or its affiliates.
  * All rights reserved. Use is subject to license terms.
  *
  * This file is available and licensed under the following license:
@@ -31,9 +31,8 @@
  */
 package com.javafx.experiments.importers.obj;
 
+import static com.javafx.experiments.importers.obj.ObjImporter.log;
 import java.io.BufferedReader;
-import java.io.File;
-import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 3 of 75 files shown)


apps/samples/Ensemble8/src/app/java/ensemble/samplepage/Description.java
```diff
@@ -31,26 +31,28 @@
  */
 package ensemble.samplepage;
 
-import ensemble.EnsembleApp;
-import ensemble.PlatformFeatures;
-import ensemble.SampleInfo;
-import ensemble.SampleInfo.URL;
-import ensemble.generated.Samples;
+import static ensemble.samplepage.SamplePage.INDENT;
+import static ensemble.samplepage.SamplePageContent.title;
```
kevinrushforth, line 35: I see that the static imports (for ensemble.xxx) were moved up to the top. Is that deliberate?

apps/samples/Ensemble8/src/app/java/ensemble/samplepage/Description.java
```diff
@@ -31,26 +31,28 @@
...
-import javafx.util.Callback;
-import static ensemble.samplepage.SamplePage.INDENT;
-import static ensemble.samplepage.SamplePageContent.title;
-import javafx.application.ConditionalFeature;
-import javafx.application.Platform;
+import ensemble.EnsembleApp;
+import ensemble.PlatformFeatures;
+import ensemble.SampleInfo;
+import ensemble.SampleInfo.URL;
+import ensemble.generated.Samples;
```
kevinrushforth, line 55: I see that `ensemble.xxx` is moved after `javafx.xxx`; similarly, elsewhere I see `com.sun.xxx` imports moved after. What's the algorithm your IDE uses to sort imports? sort stuff that starts with "java" first and then everything else alphabetically?

apps/samples/Ensemble8/src/samples/java/ensemble/samples/graphics2d/puzzle/Piece.java
```diff
@@ -43,9 +42,10 @@
 import javafx.scene.shape.Ellipse;
 import javafx.scene.shape.Rectangle;
 import javafx.scene.shape.Shape;
-   /**
-     * Node that represents a puzzle piece
-     */
+
+/**
+ * Node that represents a puzzle piece
+ */
```
kevinrushforth, line 48: Hmm. This is more than just fixing imports. Were these lines reformatted because they were touching the modified import lines? Anything other than changes to the import statements will require extra time / mental energy to review.

kevinrushforth (commented):
I spot-checked a few files and left a couple inline questions. They apply to all of the cleanup PRs.

apps/samples/Ensemble8/src/app/java/ensemble/samplepage/Description.java
andy-goryachev-oracle: it's how the formatter is currently configured.
do we have a preferred order?

apps/samples/Ensemble8/src/app/java/ensemble/samplepage/Description.java
andy-goryachev-oracle: this is how the IDE formatter is currently configured - it should not matter, especially since it dos not use wildcards (except for static imports).

apps/samples/Ensemble8/src/samples/java/ensemble/samples/graphics2d/puzzle/Piece.java
andy-goryachev-oracle: this is just one spot that bothered me.  reformatted since I have to delete a blank line.

won't happen again, I promise.

kevinrushforth:
Here is my $0.02 on the sorting of imports. I haven't cared enough before now to raise the issue, but since we are now going to be setting a precedent with these PRs, I want to codify the style guidelines. We should update the CONTRIBUTING guidelines to reflect whatever we decide. The current guidelines say this:

> Don't worry too much about import order. Try not to change it but don't worry about fighting your IDE to stop it from doing so.

We should replace this with the recommended sort order.

As for what that order should be, I have never much liked the idea of sorting or organizing imports into multiple groups with a set of rules that you have to remember / describe / teach your IDE about. The various IDEs or code-style guidelines with such conventions are often at odds with each other. Google used to have multiple groups (with rules about sorting the groups), but finally tired of this and changed to a simpler convention of having two groups: one for static imports (if any) and one for ordinary imports:

static imports (sorted alphabetically)
one blank line
ordinary imports (sorted alphabetically)

No wildcard imports (I favor an exception for static imports).

I propose that we do the same, but if someone has strong opinions otherwise, we can discuss.

apps/samples/Ensemble8/src/app/java/ensemble/samplepage/Description.java
andy-goryachev-oracle: It looks like Eclipse uses the following order **by default**: [java,javax,org,com].

andy-goryachev-oracle:
> We should replace this with the recommended sort order.

My 2 cents would be on the other side: I would strongly recommend **against** enforcing a specific order, especially since it causes no issues.  We should avoid wildcards, yes (except for static imports), but the order - we should follow the current recommendation of ignoring the order.  The main reason it is adding zero value and costs some time (just like whitespace warnings in jcheck).

kevinrushforth:
> > We should replace this with the recommended sort order.
> 
> My 2 cents would be on the other side: I would strongly recommend **against** enforcing a specific order, especially since it causes no issues. We should avoid wildcards, yes (except for static imports), but the order - we should follow the current recommendation of ignoring the order. The main reason it is adding zero value and costs some time (just like whitespace warnings in jcheck).

I agree that we shouldn't _enforce_ the order (and didn't mean to imply otherwise). I do see some value in having a "soft" recommendation, though, and encouraging IDE users to configure their IDEs with those recommendations. Otherwise, it can be a source of merge conflicts if one IDE sorts it one way, and another IDE sorts it a different way, and two different people are proposing changes to the same file. Maybe unlikely, but it has happened.

So the questions are:
1. Should we have at least a soft recommendation or leave it unspecified?
2. If the answer to 1 is "yes", what should that recommendation be?

kevinrushforth:
The other thing that a reordering of imports does is add noise to a review, so there is some advantage in having it be at least reasonably consistent. The current advice is already "try not to reorder imports", so as long as people follow that, it might be good enough. What I don't want to see is a lot of back and forth reordering, but maybe it won't be a problem (it typically hasn't been up to now).

I don't have a strong opinion on this, other than the "no wildcard imports" rule, which is separate and already documented. I do like consistency, but as long there is some reasonable ordering, I can let it go if no one else feels the need for a gudeline.

Maran23:
> static imports (sorted alphabetically)
> one blank line
> ordinary imports (sorted alphabetically)
>
> No wildcard imports (I favor an exception for static imports).

+1 for something like this. It is a simple rule and pretty sure that it works for all IDEs.
Swing is also ordered alphabetically, but with groups usually.
Otherwise it looks the order in Swing is the other way around:

```
ordinary imports (sorted alphabetically)
one blank line
static imports (sorted alphabetically)
```
For me, both are good and I understand and agree to have some king of guideline. 
That will make reviewing all the import PRs easier as well.

Also +1 for no wildcard imports other than static imports.

nlisker:
I find it helpful to separate the imports by their high-level domain name, "java.", "javafx.", "com." etc. It makes it easier to see the "span" or requirements of the class.

andy-goryachev-oracle:
I would highly suggest to use the current guidelines in CONTRIBUTING.md

```
* Don't worry too much about import order. Try not to change it but don't worry about fighting your IDE to stop it from doing so.
```

as not doing so generates unnecessary (in my opinion) discussion.

karthikpandelu:
Leaving it unspecified would be ok but some kind of soft recommendation could also help in my opinion. So there will be a direction for someone who is looking for a recommended way to add imports and have consistency class files.

> > static imports (sorted alphabetically)
> > one blank line
> > ordinary imports (sorted alphabetically)
> > No wildcard imports (I favor an exception for static imports).
> 
> +1 for something like this. 

+1 on this and no wildcard imports other than static imports.

kevinrushforth:
> I would highly suggest to use the current guidelines in CONTRIBUTING.md
> 
> ```
> * Don't worry too much about import order. Try not to change it but don't worry about fighting your IDE to stop it from doing so.
> ```

Since this is the existing guideline, and there isn't clear agreement on a new guideline, this seems like the best approach. We can revisit this after the current set of PRs are integrated if there is a desire to do so.

> as not doing so generates unnecessary (in my opinion) discussion.

And I think that it is the _absence_ of at least a soft guideline that generates the discussion, but we can table the discussion for now.

kevinrushforth:
> Leaving it unspecified would be ok but some kind of soft recommendation could also help in my opinion. So there will be a direction for someone who is looking for a recommended way to add imports and have consistency class files.

This might be the best approach. A soft recommendation in addition to the current advice to try not to reorder the imports in connection with bug fixes. We can discuss on the list (decoupled from this PR).

kevinrushforth (approved):
This PR looks good.

PR #1405, opened 2024-03-16
8186188: TableColumHeader: initial auto-size broken if has graphic

Maran23:
This PR fixes the issue that the initial column autosizing is wrong under some circumstances.
The following things will break the initial autosizing:
- Bold Column text (that is where I initially found this problem)
- Another font / font size
- Graphic
- Graphic Content Display (which can be styled via CSS)

The reason is actually quite simple: The CSS is not (yet) applied initially, we therefore ALWAYS take the default font into account + the graphic is not yet layouted as well.

~It was not so easy to write tests for this, also for me the `test_resizeColumnToFitContentHeader` is always failing locally. I don't know what happens here, but he seems to not find a (Stub?) `Font` for me.~
**EDIT: Found out the cause and fixed it. I will check if I can write more tests since it works now. :)**

The test I wrote now is checking if the css is applied after we triggered the autosize, which is what we would expect here since we measure text.

I also copied the `TableColumnHeaderTest` and rewrote the tests for `TreeTableView` as well, so we can catch any errors here as well since they both use different code (although it is technically the same - C&P errors can happen very easy).

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issue
 * [JDK-8186188](https://bugs.openjdk.org/browse/JDK-8186188): TableColumHeader: initial auto-size broken if has graphic (**Bug** - P4)

### Reviewers
 * [Andy Goryachev](https://openjdk.org/census#angorya) (@andy-goryachev-oracle - **Reviewer**)
 * [Robert Lichtenberger](https://openjdk.org/census#rlichten) (@effad - Author)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1405/head:pull/1405` \
`$ git checkout pull/1405`

Update a local copy of the PR: \
`$ git checkout pull/1405` \
`$ git pull https://git.openjdk.org/jfx.git pull/1405/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 1405`

View PR using the GUI difftool: \
`$ git pr show -t 1405`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
[...]

PR #1422, opened 2024-03-21
JDK-8328750: [TestBug] Improve Stub Font Support

Maran23:
In https://github.com/openjdk/jfx/pull/1405, I identified some shortcomings of the stub font implementation. As I don't want to clutter the PR with that, I decided to cherrypick the improvements I did to a new ticket and PR.

The current implementation has the following shortcomings:
- It does not reliable detect the System Font, as a consequence, tests in TableColumnHeaderTest.java are failing on my local machine
- Another consequence of this is, that the font size is always estimated with 0, as it is not detected
- One shortcoming currently is, that the stub font siie estimate is not considering bold fonts. That would improve writing tests for some scenarios, e.g. for TableColumnHeader, where we would expect that the size of the header is bigger since it is bold

Some tests were failing for the following reasons:
- `AreaChartTest.java` - `expected -30.0, was -30.00000000004` - I added rounding to the data.
- `StackedBarChartTest.java` - since we now calculate correctly, the path changed
- A test tried to load `Helvetica`, which is not supported in the stub font loader. I changed it
- The default System font is considered a `Regular` one (style) - just like in JavaFX

I wrote tests and documented the stub behaviour.
I did some minor changes here:
- System font is now detected, also in bold and italic
- A bold font will be calculated with a little bit more width (1px). Checkout the test as well for that

Note: This only changes test setup, no 'production' code.

---------
### Progress

### Issue
 * [JDK-8328750](https://bugs.openjdk.org/browse/JDK-8328750): [TestBug] Improve Stub Font Support (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1422/head:pull/1422` \
`$ git checkout pull/1422`

`$ git checkout pull/1422` \
`$ git pull https://git.openjdk.org/jfx.git pull/1422/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1422`

`$ git pr show -t 1422`

</details>

<a href="https://git.openjdk.org/jfx/pull/1422.diff">https://git.openjdk.org/jfx/pull/1422.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1422#issuecomment-2013935734)

### DIFF
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/chart/AreaChartTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/chart/AreaChartTest.java
@@ -534,8 +534,8 @@ public class AreaChartTest extends XYChartTestBase {
                 .filter(pathElement -> pathElement instanceof LineTo)
                 .map(pathElement -> (LineTo) pathElement)
                 .map(lineTo -> new Point2D(
-                        Math.ceil(xAxis.getValueForDisplay(lineTo.getX()).doubleValue()),
-                        Math.ceil(yAxis.getValueForDisplay(lineTo.getY()).doubleValue()))
+                        Math.round(xAxis.getValueForDisplay(lineTo.getX()).doubleValue()),
+                        Math.round(yAxis.getValueForDisplay(lineTo.getY()).doubleValue()))
                 )
                 .collect(Collectors.toList());
         // Due to fillPath, one additional LineTo element is added to close the loop
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/chart/AreaChartTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/chart/AreaChartTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -534,8 +534,8 @@ public class AreaChartTest extends XYChartTestBase {
                 .filter(pathElement -> pathElement instanceof LineTo)
                 .map(pathElement -> (LineTo) pathElement)
                 .map(lineTo -> new Point2D(
-                        xAxis.getValueForDisplay(lineTo.getX()).doubleValue(),
-                        yAxis.getValueForDisplay(lineTo.getY()).doubleValue())
+                        Math.ceil(xAxis.getValueForDisplay(lineTo.getX()).doubleValue()),
+                        Math.ceil(yAxis.getValueForDisplay(lineTo.getY()).doubleValue()))
                 )
                 .collect(Collectors.toList());
         // Due to fillPath, one additional LineTo element is added to close the loop
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/StubFontContractTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/StubFontContractTest.java
@@ -1,0 +1,175 @@
+package test.javafx.scene.control.skin;
+
+import com.sun.javafx.scene.control.skin.Utils;
+import javafx.scene.control.Label;
+import javafx.scene.text.Font;
+import javafx.scene.text.FontPosture;
+import javafx.scene.text.FontWeight;
+import org.junit.jupiter.api.Test;
+import test.com.sun.javafx.scene.control.infrastructure.StageLoader;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+/**
+ * This is a test for fonts, font loading and their computed sizes.
+ * Note that while we only test the stub font loading here, we still want to verify some basic rules that apply
+ * for all headless tests.
+ * <p>
+ * See the javadoc for every method to get more details.
+ * @see test.com.sun.javafx.pgstub.StubFontLoader
+ * @see test.com.sun.javafx.pgstub.StubTextLayout
+ */
+class StubFontContractTest {
+
+    /**
+     * An unknown font has no text width.
+     */
+    @Test
+    public void testUnknownFont() {
+        Font font = new Font("bla", 12);
+
+        assertEquals(0, Utils.computeTextWidth(font, "ABCDEFGHIJ", -1));
+    }
+
+    /**
+     * Even if the font name is equal with a family name, we do not know the font by name.
+     */
+    @Test
+    public void testFamilyAsFontName() {
+        Font font = new Font("System", 12);
+
+        assertEquals(0, Utils.computeTextWidth(font, "ABCDEFGHIJ", -1));
+    }
+
+    /**
+     * System Regular is the default font for headless testing but also for JavaFX itself.
+     * 12 is the default font size for headless testing. This is different from JavaFX, as size from the OS will be used.
+     */
+    @Test
+    public void testDefaultFont() {
+        Label lbl = new Label();
+        Font defaultFont = lbl.getFont();
+
+        assertEquals("System Regular", defaultFont.getName());
+        assertEquals("System", defaultFont.getFamily());
+        assertEquals("Regular", defaultFont.getStyle());
+        assertEquals(12, defaultFont.getSize());
+
+        assertEquals(120, Utils.computeTextWidth(defaultFont, "ABCDEFGHIJ", -1));
+    }
+
+    @Test
+    public void testDefaultFontSet() {
+        Label lbl = new Label();
+        Font font = lbl.getFont();
+
+        assertEquals("System Regular", font.getName());
+        assertEquals(12, font.getSize());
+
+        lbl.setFont(Font.font("system", FontWeight.BOLD, 20));
+
+        font = lbl.getFont();
+        assertEquals("System Bold", font.getName());
+        assertEquals(20, font.getSize());
+
+        assertEquals(210, Utils.computeTextWidth(font, "ABCDEFGHIJ", -1));
+    }
+
+    @Test
+    public void testDefaultFontCssSet() {
+        Label lbl = new Label();
+        Font font = lbl.getFont();
+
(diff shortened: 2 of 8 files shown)


modules/javafx.controls/src/test/java/test/javafx/scene/chart/AreaChartTest.java
```diff
@@ -534,8 +534,8 @@ private List<Point2D> findDataPointsFromPathLine(AreaChart<Number, Number> areaC
                 .filter(pathElement -> pathElement instanceof LineTo)
                 .map(pathElement -> (LineTo) pathElement)
                 .map(lineTo -> new Point2D(
-                        xAxis.getValueForDisplay(lineTo.getX()).doubleValue(),
-                        yAxis.getValueForDisplay(lineTo.getY()).doubleValue())
+                        Math.ceil(xAxis.getValueForDisplay(lineTo.getX()).doubleValue()),
+                        Math.ceil(yAxis.getValueForDisplay(lineTo.getY()).doubleValue()))
```
andy-goryachev-oracle, line 538: would it make more sense to either do a Math.round(), or better yet - write a utility that computes array equality of Point2D's with some non-zero tolerance?

Do we have more tests like this that might warrant a new utility?

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/StubFontContractTest.java
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+ * This is a test for fonts, font loading and their computed sizes.
+ * Note that while we only test the stub font loading here, we still want to verify some basic rules that apply
+ * for all headless tests.
+ * <p>
+ * See the javadoc for every method to get more details.
+ * @see test.com.sun.javafx.pgstub.StubFontLoader
+ * @see test.com.sun.javafx.pgstub.StubTextLayout
+ */
```
andy-goryachev-oracle, line 21: thank you for explaining what the test is trying to do (and not trying to do)!

modules/javafx.graphics/src/test/java/test/com/sun/javafx/pgstub/StubFontLoader.java
```diff
@@ -47,123 +47,127 @@ public class StubFontLoader extends FontLoader {
...
-                    "Regular");
-        } else if (name.equals("amble light condensed italic")) {
-            FontHelper.setNativeFont(font, nativeFont, font.getName(), "Amble LtCn",
-                    "Italic");
+        switch (name) {
+            case "system regular" -> FontHelper.setNativeFont(font, nativeFont, font.getName(), "System", "Regular");
+            case "system bold" -> FontHelper.setNativeFont(font, nativeFont, font.getName(), "System", "Bold");
+            case "system italic" -> FontHelper.setNativeFont(font, nativeFont, font.getName(), "System", "Italic");
+            case "system bold italic" ->
+                    FontHelper.setNativeFont(font, nativeFont, font.getName(), "System", "Bold Italic");
```
andy-goryachev-oracle, line 57: minor suggestions:
font -> f
nativeFont -> nf
this will nicely fit each case on a single line (and reduce clutter)

modules/javafx.graphics/src/test/java/test/com/sun/javafx/pgstub/StubFontLoader.java
```diff
@@ -47,123 +47,127 @@ public class StubFontLoader extends FontLoader {
...
+                    FontHelper.setNativeFont(font, nativeFont, font.getName(), "Amble Cn", "Italic");
+            case "amble bold condensed italic" ->
+                    FontHelper.setNativeFont(font, nativeFont, font.getName(), "Amble Cn", "Bold Italic");
+            case "amble light" -> FontHelper.setNativeFont(font, nativeFont, font.getName(), "Amble Lt", "Regular");
+            case "amble light italic" ->
+                    FontHelper.setNativeFont(font, nativeFont, font.getName(), "Amble Lt", "Italic");
+            case "amble light condensed" ->
+                    FontHelper.setNativeFont(font, nativeFont, font.getName(), "Amble LtCn", "Regular");
+            case "amble light condensed italic" ->
+                    FontHelper.setNativeFont(font, nativeFont, font.getName(), "Amble LtCn", "Italic");
```
andy-goryachev-oracle, line 76: does it make sense to handle/log the default case, or this is indeed the desired code (i.e. a no-op)?

modules/javafx.graphics/src/test/java/test/com/sun/javafx/pgstub/StubTextLayout.java
```diff
@@ -92,7 +92,17 @@ public BaseBounds getBounds() {
...
     public BaseBounds getBounds(TextSpan filter, BaseBounds bounds) {
-        final double fontSize = (font == null ? nullFontSize : font.getSize());
+        double fontSize = nullFontSize;
+        if (font != null) {
+            fontSize = font.getSize();
+
+            // For better testing, we make bold text a little bit bigger.
+            boolean bold = font.getStyle().toLowerCase().contains("bold");
+            if (bold) {
+                fontSize++;
```
andy-goryachev-oracle, line 102: should this only apply to width computation (line 128)?

in other words, the height of for bold text of size 12 should have the same height as plain text size 12, shouldn't it?

modules/javafx.controls/src/test/java/test/javafx/scene/chart/AreaChartTest.java
Maran23: AFAIK, we do not have other tests with that problem.
I tried to keep the diff small, but nothing against writing a better method to compare the points here with a delta.

modules/javafx.graphics/src/test/java/test/com/sun/javafx/pgstub/StubTextLayout.java
Maran23: good question! I checked some fonts and they do look to me that they have the same height regular as well as bold.
So it's probably better to use the normal size for the height here?

modules/javafx.graphics/src/test/java/test/com/sun/javafx/pgstub/StubFontLoader.java
Maran23: Right now, if we try to load a font that is not known by the test setup, the font will not have a 'native font' since no branch will get executed. With that, the width calculation will return 0.

Just checked the Prism font code, there the code will fallback to the system font if it can not load the specified font (family). 
That would also be an idea here, not sure if it will have consequences on some tests. Also okay for me. What do you think?

modules/javafx.controls/src/test/java/test/javafx/scene/chart/AreaChartTest.java
andy-goryachev-oracle: I am not sure why this is needed here, so my concern stems from the fact that `ceil()` might introduce a large bias (0.000001 -> 1.0), and `rnd()` _seems_ to be a better choice.

But then again, ceil() or rnd() might be a wrong thing to do if we are dealing with scaled coordinates and we don't expect the whole integers, but instead are dealing with small differences between expected and obtained values (in which case a better comparison is required).

Maran23: I can also use `round` here. Seems to be a bit safer, I agree.

andy-goryachev-oracle: thank you.

one more question - is it possible to set fractional scale with the StubToolkit, and how would that work in the StubFontLoader?  Does it mean we'll be rounding the scaled values?

andy-goryachev-oracle:
One more question, for my education, since you are now the expert :-)

I am trying to decide whether I should write a headful test for Labeled.truncated, or a headless one will suffice.

If I write a test that uses the StubToolkit, one that requires some kind of font metrics - would that work?  Will I get some kind of semi-valid values for preferred width of a text (Text, Labeled, etc.), or I cannot really rely on the correctness of the StubTextLayout?

modules/javafx.graphics/src/test/java/test/com/sun/javafx/pgstub/StubFontLoader.java
Maran23: I gave the fallback font a shot, since this is closer to the `PrismFontLoader`, which is good when we try to be as close a possible in the tests.

modules/javafx.graphics/src/test/java/test/com/sun/javafx/pgstub/StubTextLayout.java
```diff
@@ -33,6 +33,12 @@
 import javafx.scene.shape.PathElement;
 import javafx.scene.text.Font;
 
+/**
+ * Stub implementation of the {@link TextLayout} for testing purposes.
+ * <br>
+ * Can calculate the bounds of text by simply using the size of the font.
+ * If the text is bold, the font will be 1 pixel bigger.
```
andy-goryachev-oracle, line 40: maybe change to "somewhat wider" instead, since the height is the same?

andy-goryachev-oracle:
> Yes, it should!

Thank you so much for the code and explanations!
I'll try to do a formal review tomorrow, using Labeled.truncated as a test bed.

modules/javafx.controls/src/test/java/test/javafx/scene/chart/AreaChartTest.java
Maran23: You mean for the font size? That will as far as I can see just work, although not 100% sure what the rendering, e.g. a Label will do with it

andy-goryachev-oracle (approved):
looks good, one minor javadoc comment.
if you decide to update, I'll re-approve.

modules/javafx.graphics/src/test/java/test/com/sun/javafx/pgstub/StubTextLayout.java
Maran23: sure!
Fun fact: I also was not happy with the javadoc but couldn't come up with a better wording. Fixed now. :)

andy-goryachev-oracle (approved):

PR #1430, opened 2024-03-24
8273349: Check uses of Stream::peek in controls and replace as needed

drmarmac:
This PR removes potentially incorrect usages of Stream.peek().
The changed code should be covered by the tests that are already present.

---------
### Progress

### Issue
 * [JDK-8273349](https://bugs.openjdk.org/browse/JDK-8273349): Check uses of Stream::peek in controls and replace as needed (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1430/head:pull/1430` \
`$ git checkout pull/1430`

`$ git checkout pull/1430` \
`$ git pull https://git.openjdk.org/jfx.git pull/1430/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1430`

`$ git pr show -t 1430`

</details>

<a href="https://git.openjdk.org/jfx/pull/1430.diff">https://git.openjdk.org/jfx/pull/1430.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1430#issuecomment-2016841067)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/ControlUtils.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/ControlUtils.java
@@ -38,7 +38,6 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
 import java.util.function.IntPredicate;
-import java.util.stream.Collectors;
 
 class ControlUtils {
     private ControlUtils() { }
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/ControlUtils.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/ControlUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2013, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -33,6 +33,7 @@ import javafx.scene.Node;
 import javafx.scene.Parent;
 import javafx.scene.Scene;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
@@ -166,19 +167,26 @@ class ControlUtils {
             // ensuring that the selectedIndices bitset is correctly updated.
 
             sm.startAtomic();
-            final List<Integer> removed = c.getRemoved().stream()
+
+            final List<Integer> removed = new ArrayList<>(c.getRemovedSize());
+            c.getRemoved().stream()
                     .mapToInt(TablePositionBase::getRow)
                     .distinct()
                     .filter(removeRowFilter)
-                    .boxed()
-                    .peek(sm.selectedIndices::clear)
-                    .collect(Collectors.toList());
+                    .forEach(row -> {
+                        removed.add(row);
+                        sm.selectedIndices.clear(row);
+                    });
 
-            final int addedSize = (int)c.getAddedSubList().stream()
+            final int[] addedSize = new int[1];
+            c.getAddedSubList().stream()
                     .mapToInt(TablePositionBase::getRow)
                     .distinct()
-                    .peek(sm.selectedIndices::set)
-                    .count();
+                    .forEach(row -> {
+                        addedSize[0]++;
+                        sm.selectedIndices.set(row);
+                    });
+
             sm.stopAtomic();
 
             int from = c.getFrom();
@@ -188,7 +196,7 @@ class ControlUtils {
                 int tpRow = c.getList().get(from).getRow();
                 from = sm.selectedIndices.indexOf(tpRow);
             }
-            final int to = from + addedSize;
+            final int to = from + addedSize[0];
 
             if (c.wasReplaced()) {
                 sm.selectedIndices._nextReplace(from, to, removed);
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/ControlUtils.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/ControlUtils.java
@@ -162,10 +162,6 @@ class ControlUtils {
         sm.selectedIndices._beginChange();
 
         while (c.next()) {
-            // it may look like all we are doing here is collecting the removed elements (and
-            // counting the added elements), but the call to 'peek' is also crucial - it is
-            // ensuring that the selectedIndices bitset is correctly updated.
-
             sm.startAtomic();
 
             final List<Integer> removed = new ArrayList<>(c.getRemovedSize());
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/MultipleSelectionModelBase.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/MultipleSelectionModelBase.java
@@ -761,16 +761,18 @@ abstract class MultipleSelectionModelBase<T> extends MultipleSelectionModel<T> {
                 // by finding all contiguous indices, of all indices that are
                 // not already selected, and which are in the valid range
                 startAtomic();
-                List<Integer> sortedNewIndices =
-                        IntStream.concat(IntStream.of(index), IntStream.of(indices))
+
+                List<Integer> sortedNewIndices = new ArrayList<>(indices.length + 1);
+                IntStream.concat(IntStream.of(index), IntStream.of(indices))
                         .distinct()
                         .filter(this::isValidIndex)
                         .filter(this::isNotSelected)
                         .sorted()
-                        .boxed()
-                        .collect(Collectors.toList());
+                        .forEach(i -> {
+                            sortedNewIndices.add(i);
+                            set(i);
+                        });
 
-                sortedNewIndices.forEach(this::set);
                 stopAtomic();
 
                 final int size = sortedNewIndices.size();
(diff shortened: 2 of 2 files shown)


kevinrushforth:
Reviewers: @andy-goryachev-oracle @karthikpandelu 

/reviewers 2

modules/javafx.controls/src/main/java/javafx/scene/control/ControlUtils.java
```diff
@@ -166,19 +167,26 @@ public static <S> void updateSelectedIndices(MultipleSelectionModelBase<S> sm, b
...
+
+            final List<Integer> removed = new ArrayList<>(c.getRemovedSize());
+            c.getRemoved().stream()
                     .mapToInt(TablePositionBase::getRow)
                     .distinct()
                     .filter(removeRowFilter)
-                    .boxed()
-                    .peek(sm.selectedIndices::clear)
-                    .collect(Collectors.toList());
+                    .forEach(row -> {
```
karthikpandelu, line 176: In the java.util.stream package [docs](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/util/stream/package-summary.html#SideEffects) it is mentioned that `forEach()` method operates only via side-effects. So do you think we should avoid using `forEach()` here and iterate the generated list separately to clear selected index?

modules/javafx.controls/src/main/java/javafx/scene/control/ControlUtils.java
```diff
@@ -166,19 +167,26 @@ public static <S> void updateSelectedIndices(MultipleSelectionModelBase<S> sm, b
...
+                    });
 
-            final int addedSize = (int)c.getAddedSubList().stream()
+            final int[] addedSize = new int[1];
+            c.getAddedSubList().stream()
                     .mapToInt(TablePositionBase::getRow)
                     .distinct()
-                    .peek(sm.selectedIndices::set)
-                    .count();
+                    .forEach(row -> {
```
karthikpandelu, line 185: Similar comment as above. Here if we do not use `forEach()` on streams, we can also avoid using array of size one for keeping count as well right?

modules/javafx.controls/src/main/java/javafx/scene/control/ControlUtils.java
drmarmac: I'd say .forEach() is used correctly here, according to docs, it guarantees execution of the side-effects (add to removed list & clear index), just not in any particular order. This way we avoid multiple iteration.

modules/javafx.controls/src/main/java/javafx/scene/control/ControlUtils.java
drmarmac: We'd need to iterate twice as well (select index & count), with forEach it's just once.

drmarmac:
I also removed a code comment that explained the usage of peek() here, which would be outdated now.

modules/javafx.controls/src/main/java/javafx/scene/control/ControlUtils.java
Maran23: Another idea is to use `toList()`, which is a very efficient operation and then iterate over it.

nlisker: > In the java.util.stream package [docs](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/util/stream/package-summary.html#SideEffects) it is mentioned that `forEach()` method operates only via side-effects. So do you think we should avoid using `forEach()` here and iterate the generated list separately to clear selected index?

`forEach` is used correctly here. From the docs:
> With the exception of terminal operations [forEach](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/stream/Stream.html#forEach(java.util.function.Consumer)) and [forEachOrdered](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/stream/Stream.html#forEachOrdered(java.util.function.Consumer)), side-effects of behavioral parameters may not always be executed when the stream implementation can optimize away the execution of behavioral parameters without affecting the result of the computation.

> Another idea is to use `toList()`, which is a very efficient operation and then iterate over it.

That's still 2 iterations. If the code is not performance-critical it doesn't matter.

modules/javafx.controls/src/main/java/javafx/scene/control/MultipleSelectionModelBase.java
```diff
@@ -768,8 +768,9 @@ public void set(int index, int... indices) {
                         .filter(this::isNotSelected)
                         .sorted()
                         .boxed()
-                        .peek(this::set) // we also set here, but it's atomic!
                         .collect(Collectors.toList());
+
+                sortedNewIndices.forEach(this::set);
```
nlisker, line 773: Why do the double-iteration pattern here and not do the `peek` operation in a `forEach` like in the other 2 places?

Maran23: `forEach` is void, so we can not return a list afterwards.

nlisker: You don't need to return a list, you create it ahead of time like was done in line 167
```java
List<Integer> indices = new ArrayList<>();
```
and the add the elements in `forEach`.

modules/javafx.controls/src/main/java/javafx/scene/control/ControlUtils.java
Maran23: > That's still 2 iterations.

Yes, but one advantage here:
We currently do `final List<Integer> removed = new ArrayList<>(c.getRemovedSize());`, 
where we allocate a list with a size, that is probably too big since we filter the removed items.
So with `toList`, we at least get back a list with the correct size.
But true, that we technically iterate twice then. It probably does not matter too much.

modules/javafx.controls/src/main/java/javafx/scene/control/MultipleSelectionModelBase.java
karthikpandelu: > Why do the double-iteration pattern here and not do the `peek` operation in a `forEach` like in the other 2 places?

Yes, if the usage of `forEach` in previous 2 places are correct then I would like to see similar change here as well.

modules/javafx.controls/src/main/java/javafx/scene/control/ControlUtils.java
karthikpandelu: > `forEach` is used correctly here. From the docs:
> 
> > With the exception of terminal operations [forEach](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/stream/Stream.html#forEach(java.util.function.Consumer)) and [forEachOrdered](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/stream/Stream.html#forEachOrdered(java.util.function.Consumer)), side-effects of behavioral parameters may not always be executed when the stream implementation can optimize away the execution of behavioral parameters without affecting the result of the computation.
> 

Thanks for pointing this out.

modules/javafx.controls/src/main/java/javafx/scene/control/MultipleSelectionModelBase.java
drmarmac: I changed this to the pre-allocated pattern like in the other places. An initial capacity of `indices.length + 1` optimizes the worst case (all given indices need to be set) which probably occurs quite often in real-world usage.

drmarmac:
I also removed a now unused import in ControlUtils.java.

karthikpandelu (approved):

andy-goryachev-oracle:
a minor question: would it be much easier to understand if the code was written in conventional procedural style?   a straightforward `for` loop?

Maran23 (approved):
Looks good. Should be good regarding performance as well.

kevinrushforth:
> a minor question: would it be much easier to understand if the code was written in conventional procedural style? a straightforward `for` loop?

Even if it were (and I'm not sure it would be), it would be a more intrusive change than is required by the fix, so I wouldn't recommend it.

modules/javafx.controls/src/main/java/javafx/scene/control/ControlUtils.java
```diff
@@ -161,24 +161,27 @@ public static <S> void updateSelectedIndices(MultipleSelectionModelBase<S> sm, b
...
 
         while (c.next()) {
-            // it may look like all we are doing here is collecting the removed elements (and
-            // counting the added elements), but the call to 'peek' is also crucial - it is
-            // ensuring that the selectedIndices bitset is correctly updated.
-
             sm.startAtomic();
-            final List<Integer> removed = c.getRemoved().stream()
+
+            final List<Integer> removed = new ArrayList<>(c.getRemovedSize());
```
andy-goryachev-oracle, line 166: I wonder if we should add a check for 0 size here to bypass all this code if nothing is removed (same for added)

drmarmac: We certainly could, or maybe use wasRemoved(), but I doubt there will be much impact. I guess it's preferred to keep changes unrelated to the issue minimal, so I'd leave it as it is if everyone's ok with that.

nlisker: These short circuiting operations tend to be worth it only if it's a critical path. The GC will collect the allocations very efficiently these days. I didn't check what this code segment is used for, but unless it's looped somewhere, I doubt there will by any change.

andy-goryachev-oracle (approved):
Thank you for clarifications!

Does not look like this is called from a loop, so whatever inefficiency we have is likely to be ok.

PR #1433, opened 2024-03-25
8306322: JDK8130122Test fails intermittently

jayathirthrao:
This test has failed once and we are not seeing its failure after that instance in our test systems.

This test verifies whether bounds of GridPane gets updated properly on adding an invisible node.
Initial test has 8 nodes in GridPane and then we update it with another node with larger bounds without making it visible. After adding additional node we make the scenegraph visible and check for colors of the newly added node.

We are making this test robust to make sure we don't see any of these single instance failures.
Test is updated to:
1) To always show on top, so that we pick proper color.
2) Add additional sleep so that we give more time for test to be visible.
3) Pick color exactly at mid point in y axis, so that we pick the green color properly.

---------
### Progress

### Issue
 * [JDK-8306322](https://bugs.openjdk.org/browse/JDK-8306322): JDK8130122Test fails intermittently (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1433/head:pull/1433` \
`$ git checkout pull/1433`

`$ git checkout pull/1433` \
`$ git pull https://git.openjdk.org/jfx.git pull/1433/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1433`

`$ git pr show -t 1433`

</details>

<a href="https://git.openjdk.org/jfx/pull/1433.diff">https://git.openjdk.org/jfx/pull/1433.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1433#issuecomment-2017893789)

### DIFF
--- a/tests/system/src/test/java/test/robot/scenegraph/JDK8130122Test.java
+++ b/tests/system/src/test/java/test/robot/scenegraph/JDK8130122Test.java
@@ -84,7 +84,6 @@ public class JDK8130122Test extends VisualTestBase {
             testScene.setCamera(new PerspectiveCamera());
             testScene.setFill(Color.WHITE);
             testStage.setScene(testScene);
-            testStage.setAlwaysOnTop(true);
             testStage.show();
         });
         waitFirstFrame();
@@ -99,8 +98,8 @@ public class JDK8130122Test extends VisualTestBase {
             assertColorEquals(Color.WHITE, color, TOLERANCE);
             horizontalListView.setVisible(true);
         });
-        sleep(1000);
-        waitNextFrame();
+        // Give more time after setVisible(true) is called for frame update
+        waitFirstFrame();
         runAndWait(() -> {
             Color color = getColor(testScene, 200, 200);
             assertColorEquals(Color.GREEN, color, TOLERANCE);
--- a/tests/system/src/test/java/test/robot/scenegraph/JDK8130122Test.java
+++ b/tests/system/src/test/java/test/robot/scenegraph/JDK8130122Test.java
@@ -50,7 +50,7 @@ public class JDK8130122Test extends VisualTestBase {
 
     private static final double TOLERANCE = 0.07;
 
-    @Test(timeout = 15000)
+    @Test(timeout = 20000)
     public void testEmptyShapes() {
         final int WIDTH = 800;
         final int HEIGHT = 400;
@@ -84,6 +84,7 @@ public class JDK8130122Test extends VisualTestBase {
             testScene.setCamera(new PerspectiveCamera());
             testScene.setFill(Color.WHITE);
             testStage.setScene(testScene);
+            testStage.setAlwaysOnTop(true);
             testStage.show();
         });
         waitFirstFrame();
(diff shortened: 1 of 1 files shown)


tests/system/src/test/java/test/robot/scenegraph/JDK8130122Test.java
```diff
@@ -84,6 +84,7 @@ public void testEmptyShapes() {
             testScene.setCamera(new PerspectiveCamera());
             testScene.setFill(Color.WHITE);
             testStage.setScene(testScene);
+            testStage.setAlwaysOnTop(true);
```
kevinrushforth, line 87: This is redundant. `getStage()` already sets `alwaysOnTop`. If the window currently isn't always on top then we have a bigger problem.

kevinrushforth (commented):
The changes look fine, except for the redundant call to `setAlwaysOnTop` (it already is).

tests/system/src/test/java/test/robot/scenegraph/JDK8130122Test.java
```diff
@@ -98,9 +99,10 @@ public void testEmptyShapes() {
             assertColorEquals(Color.WHITE, color, TOLERANCE);
             horizontalListView.setVisible(true);
         });
+        sleep(1000);
```
andy-goryachev-oracle, line 102: we still have JDK-8176902 open, but perhaps here we could add a new method to VisualTestBase similar to waitNextFrame(), with an appropriate parameter?

Something like
```
protected void waitForIdle() {
  frameWait(100);
}
```

instead of 1 second sleep() ?

tests/system/src/test/java/test/robot/scenegraph/JDK8130122Test.java
jayathirthrao: I see that thanks, i didn't notice it. I will remove the change.

I took https://github.com/openjdk/jfx/commit/56f2e17f#diff-43fd8d27727f352ddb3526031bd32e5d30aefa2bad722d1f3d21c4a0ca8a5a1f as an example and made this change.

tests/system/src/test/java/test/robot/scenegraph/JDK8130122Test.java
jayathirthrao: I think waitForIdle() would not be a proper name for this use case.
And we already have waitFirstFrame() which does frameWait(100).

What i will do is, use waitFirstFrame() and add comment in test case that we are waiting after setVisible(true).

kevinrushforth (approved):

tests/system/src/test/java/test/robot/scenegraph/JDK8130122Test.java
```diff
@@ -98,9 +98,10 @@ public void testEmptyShapes() {
             assertColorEquals(Color.WHITE, color, TOLERANCE);
             horizontalListView.setVisible(true);
         });
-        waitNextFrame();
+        // Give more time after setVisible(true) is called for frame update
+        waitFirstFrame();
```
andy-goryachev-oracle, line 102: this is fine, though I'd rather have a dedicated method - waitFirstFrame() implies that it should only be used to wait for the first frame, that it might have a different purpose and implementation.

andy-goryachev-oracle (approved):
anyway, even if we decide to add another method that will be a separate PR.

PR #1434, opened 2024-03-28
8328577: Toolbar's overflow button overlaps the items

eduardsdv:
This change fixes the calculation of which nodes go to the toolbar and which go to the overflow menu.
It is now determined before the nodes are removed from the scene graph.
This is important because the values returned by ``Node.prefWidth(..)``/``Node.prefHeight(..)`` may depend on whether the node is added to the scene graph.

Furthermore I corrected the ``hasOveflow`` value passed to the ``organizeOverflow(double, boolean)`` in ``correctOverflow(double)``.

---------
### Progress

### Issue
 * [JDK-8328577](https://bugs.openjdk.org/browse/JDK-8328577): Toolbar's overflow button overlaps the items (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1434/head:pull/1434` \
`$ git checkout pull/1434`

`$ git checkout pull/1434` \
`$ git pull https://git.openjdk.org/jfx.git pull/1434/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1434`

`$ git pr show -t 1434`

</details>

<a href="https://git.openjdk.org/jfx/pull/1434.diff">https://git.openjdk.org/jfx/pull/1434.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1434#issuecomment-2024971783)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
@@ -39,7 +39,6 @@ import com.sun.javafx.scene.traversal.TraversalContext;
 import javafx.beans.property.ObjectProperty;
 import javafx.beans.property.DoubleProperty;
 import javafx.beans.value.WritableValue;
-import javafx.collections.FXCollections;
 import javafx.collections.ListChangeListener;
 import javafx.collections.ObservableList;
 import javafx.geometry.HPos;
@@ -76,6 +75,7 @@ import com.sun.javafx.scene.control.behavior.ToolBarBehavior;
 import com.sun.javafx.scene.traversal.Direction;
 
 import javafx.css.Styleable;
+import javafx.stage.WindowEvent;
 
 import static com.sun.javafx.scene.control.skin.resources.ControlResources.getString;
 
@@ -94,6 +94,12 @@ public class ToolBarSkin extends SkinBase<ToolBar> {
      **************************************************************************/
 
     private Pane box;
+    /**
+     * The overflow logic needs properly calculated prefWidth(..)/prefHeight(..) values.
+     * These values are valid if the elements have been added to the scene and the CSS has been applied properly.
+     * To ensure this, we add the overflow items to this pane if they are not currently visible in the overflow menu.
+     */
+    private Pane overflowBox;
     private ToolBarOverflowMenu overflowMenu;
     private boolean overflow = false;
     private int overflowNodeIndex = Integer.MAX_VALUE;
@@ -101,7 +107,6 @@ public class ToolBarSkin extends SkinBase<ToolBar> {
     private double previousHeight = 0;
     private double savedPrefWidth = 0;
     private double savedPrefHeight = 0;
-    private ObservableList<MenuItem> overflowMenuItems;
     private boolean needsUpdate = false;
     private final ParentTraversalEngine engine;
     private final BehaviorBase<ToolBar> behavior;
@@ -128,7 +133,6 @@ public class ToolBarSkin extends SkinBase<ToolBar> {
         behavior = new ToolBarBehavior(control);
 //        control.setInputMap(behavior.getInputMap());
 
-        overflowMenuItems = FXCollections.observableArrayList();
         initialize();
         registerChangeListener(control.orientationProperty(), e -> initialize());
 
@@ -246,6 +250,7 @@ public class ToolBarSkin extends SkinBase<ToolBar> {
             while (c.next()) {
                 for (Node n: c.getRemoved()) {
                     box.getChildren().remove(n);
+                    overflowBox.getChildren().remove(n);
                 }
                 box.getChildren().addAll(c.getAddedSubList());
             }
@@ -553,17 +558,25 @@ public class ToolBarSkin extends SkinBase<ToolBar> {
     private void initialize() {
         if (getSkinnable().getOrientation() == Orientation.VERTICAL) {
             box = new VBox();
+            overflowBox = new VBox();
         } else {
             box = new HBox();
+            overflowBox = new HBox();
         }
         box.getStyleClass().add("container");
         box.getChildren().addAll(getSkinnable().getItems());
-        overflowMenu = new ToolBarOverflowMenu(overflowMenuItems);
+        // The overflowBox must have the same style classes, otherwise the overflow items may get wrong values.
+        overflowBox.setId(box.getId());
+        overflowBox.getStyleClass().addAll(box.getStyleClass());
+        overflowBox.setManaged(false);
+        overflowBox.setVisible(false);
+        overflowMenu = new ToolBarOverflowMenu(overflowBox.getChildren());
         overflowMenu.setVisible(false);
         overflowMenu.setManaged(false);
 
         getChildren().clear();
         getChildren().add(box);
+        getChildren().add(overflowBox);
         getChildren().add(overflowMenu);
 
         previousWidth = 0;
@@ -599,8 +612,8 @@ public class ToolBarSkin extends SkinBase<ToolBar> {
         // Determine which node goes to the toolbar and which goes to the overflow.
         ObservableList<Node> nodes = getSkinnable().getItems();
 
-        overflowMenuItems.clear();
         box.getChildren().clear();
+        overflowBox.getChildren().clear();
         for (int i = 0; i < nodes.size(); i++) {
             Node node = nodes.get(i);
             node.getStyleClass().remove("menu-item");
@@ -609,6 +622,7 @@ public class ToolBarSkin extends SkinBase<ToolBar> {
             if (i < newOverflowNodeIndex) {
                 box.getChildren().add(node);
             } else {
+                overflowBox.getChildren().add(node);
                 if (node.isFocused()) {
                     if (!box.getChildren().isEmpty()) {
                         Node last = engine.selectLast();
@@ -619,63 +633,13 @@ public class ToolBarSkin extends SkinBase<ToolBar> {
                         overflowMenu.requestFocus();
                     }
                 }
-                if (node instanceof Separator) {
-                    overflowMenuItems.add(new SeparatorMenuItem());
-                } else {
-                    CustomMenuItem customMenuItem = new CustomMenuItem(node);
-
-                    // RT-36455:
-                    // We can't be totally certain of all nodes, but for the
-                    // most common nodes we can check to see whether we should
-                    // hide the menu when the node is clicked on. The common
-                    // case is for TextField or Slider.
-                    // This list won't be exhaustive (there is no point really
-                    // considering the ListView case), but it should try to
-                    // include most common control types that find themselves
-                    // placed in menus.
-                    final String nodeType = node.getTypeSelector();
-                    switch (nodeType) {
-                        case "Button":
-                        case "Hyperlink":
-                        case "Label":
-                            customMenuItem.setHideOnClick(true);
-                            break;
-                        case "CheckBox":
-                        case "ChoiceBox":
-                        case "ColorPicker":
-                        case "ComboBox":
-                        case "DatePicker":
-                        case "MenuButton":
-                        case "PasswordField":
-                        case "RadioButton":
-                        case "ScrollBar":
-                        case "ScrollPane":
-                        case "Slider":
-                        case "SplitMenuButton":
-                        case "SplitPane":
-                        case "TextArea":
-                        case "TextField":
-                        case "ToggleButton":
-                        case "ToolBar":
-                        default:
-                            customMenuItem.setHideOnClick(false);
-                            break;
-                    }
-
-                    overflowMenuItems.add(customMenuItem);
-                }
             }
         }
 
         // Check if we overflowed.
-        overflow = overflowMenuItems.size() > 0;
+        overflow = !overflowBox.getChildren().isEmpty();
         overflowNodeIndex = newOverflowNodeIndex;
-        if (overflow) {
-            // The nodes needs to be added to the scene and the css needs to be applied,
-            // otherwise subsequent prefWidth(..)/prefHeight(..) may return wrong values.
-            overflowMenu.popup.getItems().setAll(overflowMenuItems);
-            overflowMenu.popup.getScene().getRoot().applyCss();
-        } else if(overflowMenu.isFocused()) {
+        if (!overflow && overflowMenu.isFocused()) {
             Node last = engine.selectLast();
             if (last != null) {
                 last.requestFocus();
@@ -742,14 +706,14 @@ public class ToolBarSkin extends SkinBase<ToolBar> {
     class ToolBarOverflowMenu extends StackPane {
         private StackPane downArrow;
         private ContextMenu popup;
-        private ObservableList<MenuItem> menuItems;
+        private ObservableList<Node> overflowItems;
 
-        public ToolBarOverflowMenu(ObservableList<MenuItem> items) {
+        public ToolBarOverflowMenu(ObservableList<Node> items) {
             getStyleClass().setAll("tool-bar-overflow-button");
             setAccessibleRole(AccessibleRole.BUTTON);
             setAccessibleText(getString("Accessibility.title.ToolBar.OverflowButton"));
             setFocusTraversable(true);
-            this.menuItems = items;
+            this.overflowItems = items;
             downArrow = new StackPane();
             downArrow.getStyleClass().setAll("arrow");
             downArrow.setOnMousePressed(me -> {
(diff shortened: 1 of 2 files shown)


andy-goryachev-oracle:
The proposed solution still does not work - using the sample code in the ticket, try resizing back and forth around the button with -xxxx- text:

Screenshot 2024-03-28 at 08 30 37

andy-goryachev-oracle:
a couple of notes:

1. do you think we should add a test?
2. should there be a call to applyCss() somewhere?

eduardsdv:
Thanks for finding it. I have reworked my solution. 

It is indeed necessary to add the overflow nodes to the scene of the popup and reapply the CSS so that the ``prefWidth(..)``/``prefHeight(..)`` return correct values, even if the nodes are no longer in the toolbar.

I will try to create a unit test that ensures the bug does not occur again.

andy-goryachev-oracle:
Thank you for adding a unit test!

I think you might have introduced an issue.  I am using the Monkey Tester
 to dynamically change the application style sheet by adding the following line:
```
.button { -fx-padding: 100 100 100 100; }
```
it's a bit extreme, yes, but once that update is made, I can get it to generate continuous flicker (see the recording).

https://github.com/openjdk/jfx/assets/107069028/b1a2d8ac-3d14-45b3-adff-b3893cd0ff90

steps to reproduce (with the MT):
- select ToolBar page from the list on the left
- set HORIZONTAL orientation
- open Tools -> CSS Playground tool, enter the CSS string above, press Update button
- try resizing the split pane containing the tool bar.  at some point the flicker starts.
- I also see the small buttons in the area that flickers, suggesting that CSS might not be applied correctly

For reference, I am running this on a secondary monitor with 100% scale with macOS 14.4.1

eduardsdv:
I solved the issue with applying of CSS by introducing a 'shadow' pane (``overflowBox``).
This pane contains the overflowed items all the time unless overflow menu is open.
After the overflow menu is closed the items are moved back the to the pane.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
```diff
@@ -552,17 +558,25 @@ public CssMetaData<ToolBar,Pos> getCssMetaData() {
...
         } else {
             box = new HBox();
+            overflowBox = new HBox();
         }
         box.getStyleClass().add("container");
         box.getChildren().addAll(getSkinnable().getItems());
-        overflowMenu = new ToolBarOverflowMenu(overflowMenuItems);
+        // The overflowBox must have the same style classes, otherwise the overflow items may get wrong values.
+        overflowBox.setId(box.getId());
+        overflowBox.getStyleClass().addAll(box.getStyleClass());
```
andy-goryachev-oracle, line 570: I think this solution might not be the best one.  One reason is that both the `id` property and the style class list can change at runtime, and this code does not handle that (using `bind()` and `bindContent()` for property and the observable list correspondingly).

On a higher level, do you think it is possible to simply lay out hidden buttons using width/height of 0 instead?  This way there will be no need for the `overflowBox` and its maintenance.

What do you think?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
```diff
@@ -778,11 +767,64 @@ private void fire() {
...
 
+        private List<MenuItem> createMenuItems() {
+            List<MenuItem> menuItems = new ArrayList<>();
+            for (Node node : overflowItems) {
+                if (node instanceof Separator) {
+                    menuItems.add(new SeparatorMenuItem());
+                } else {
+                    CustomMenuItem customMenuItem = new CustomMenuItem(node);
+
+                    // RT-36455:
```
andy-goryachev-oracle, line 783: since we moved this code, could we change this line to
`     // RT-36455 (JDK-8096292):`

and also make sure that this change introduced no regression w.r.t. JDK-8096292 ?
(My limited testing indicates that it still works as expected, but please double check)

andy-goryachev-oracle (commented):
It looks like the flicker issue has been resolved.

There might be an issue with runtime style update which the proposed solution may not handle (I can't readily test it right now, but I think it is a valid concern).

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
eduardsdv: I tested it with the Modena application. The bug is not reproducible.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
eduardsdv: I agree with the first point. The style related properties are now bound. See the commit [92921e3](https://github.com/openjdk/jfx/pull/1434/commits/92921e36987e3cd1cfe78c17b464547aa73d241e)

As for the second point, this does work in a quick test. But I have some concerns.
The nodes can define minWidth/minHeight, so 0 is not a valid value and can lead to errors in app code.

You could move the nodes outside the visible area instead, but the nodes can have effects and margins, these can still be visible if they are large enough. Furthermore, in order for the toolbar to continue to return correct values for prefWidth(-1)/prefHeight(-1), we need to use a clip node.

I honestly can't validate that this doesn't lead to any errors.

E.g. shadow of the button, that is moved outside visible area:
image

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
eduardsdv: I changed the comment. See [e63b0b9](https://github.com/openjdk/jfx/pull/1434/commits/e63b0b9b313131460fbe154ce8af164a3a97672f)

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
```diff
@@ -566,8 +568,18 @@ private void initialize() {
         box.getStyleClass().add("container");
         box.getChildren().addAll(getSkinnable().getItems());
         // The overflowBox must have the same style classes, otherwise the overflow items may get wrong values.
-        overflowBox.setId(box.getId());
-        overflowBox.getStyleClass().addAll(box.getStyleClass());
+        overflowBox.idProperty().bind(box.idProperty());
+        overflowBox.getStyleClass().setAll(box.getStyleClass());
+        box.getStyleClass().addListener((ListChangeListener<? super String>) change -> overflowBox.getStyleClass().setAll(change.getList()));
```
andy-goryachev-oracle, line 573: I think what you need here (and below) is `Bindings.bindContent(List, ObservableList)`

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
```diff
@@ -552,17 +560,35 @@ public CssMetaData<ToolBar,Pos> getCssMetaData() {
...
         }
         box.getStyleClass().add("container");
         box.getChildren().addAll(getSkinnable().getItems());
-        overflowMenu = new ToolBarOverflowMenu(overflowMenuItems);
+        // The overflowBox must have the same style classes, otherwise the overflow items may get wrong values.
+        overflowBox.idProperty().bind(box.idProperty());
+        overflowBox.getStyleClass().setAll(box.getStyleClass());
+        box.getStyleClass().addListener((ListChangeListener<? super String>) change -> overflowBox.getStyleClass().setAll(change.getList()));
+        overflowBox.getStylesheets().setAll(box.getStylesheets());
+        box.getStylesheets().addListener((ListChangeListener<? super String>) change -> overflowBox.getStylesheets().setAll(change.getList()));
```
andy-goryachev-oracle, line 575: this is interesting.  

isn't `overflowBox` a sibling of `box`, having the same parent, and therefore inheriting the same set of stylesheets from the parent `Scene`?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
eduardsdv: Good suggestion

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
eduardsdv: In addition to those from the scene, each ``Parent`` can have its own stylesheets.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
```diff
@@ -552,17 +561,33 @@ public CssMetaData<ToolBar,Pos> getCssMetaData() {
...
+            overflowBox = new HBox();
         }
         box.getStyleClass().add("container");
         box.getChildren().addAll(getSkinnable().getItems());
-        overflowMenu = new ToolBarOverflowMenu(overflowMenuItems);
+        // The overflowBox must have the same style classes, otherwise the overflow items may get wrong values.
+        overflowBox.idProperty().bind(box.idProperty());
+        Bindings.bindContent(overflowBox.getStyleClass(), box.getStyleClass());
+        Bindings.bindContent(overflowBox.getStylesheets(), box.getStylesheets());
+        box.getPseudoClassStates().addListener((SetChangeListener<? super PseudoClass>) change -> {
```
andy-goryachev-oracle, line 575: you can also use Bindings.bindContent(Set,ObservableSet)

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
andy-goryachev-oracle: OK, you are correct - the `box` is technically visible via node lookup by the "content" style, so the app developer can attach a stylesheet to it (why they would do it is another question, I am sure there are many other places where doing so would break the code).
And we are not creating leaks because both nodes are siblings.

andy-goryachev-oracle:
> Can this pull request be merged?

Almost there: please address the remaining review comments, and we need one more reviewer.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
andy-goryachev-oracle: line 575 still uses a listener instead of Bindings.bindContent(**Set,ObservableSet**)

andy-goryachev-oracle:
@karthikpandelu or @aghaisas could one of you be the second reviewer?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
eduardsdv: The ``Node.getPseudoClassStates()`` returns an unmodifiable collection. The bindings would not work. Therefore we need to use a listener and change the pseudo-class-state by using the ``Node.pseudoClassStateChanged(PseudoClass, boolean)`` method.

andy-goryachev-oracle: how silly of me!  you are right, of course.  thank you for clarification.

eduardsdv: You are welcome, partner :-). 
It was a surprise for me too. 
It's strange, why it needs to be an unmodifiable collection.

andy-goryachev-oracle: good question!  admittedly, it might be easier to control what happens as a result of a change if the only mutator is `pseudoClassStateChanged()` method.

andy-goryachev-oracle (approved):
Looks good on macOS 14.4.1 M1, works as expected, including the case when the application stylesheet gets reloaded at run time.

karthikpandelu (commented):
I see few of issue with the item selection when toolbar items are overflown.

1. If a visible toolbar item is selected and we select an item from overflown list, both the items get marked as selected with blue outline. But overflow icon remains grey.
2. If an item is selected and then change window size so that selected item moves to overflow list, overflow icon becomes blue. If we change the selection to another item in the overflow list, then overflow icon becomes grey. But on hovering anywhere in the window, overflow icon becomes blue again.
3. If an item is selected in the overflow list and we select an item from the visible list, selection in the overflow list remains on opening the overflow list for the first time but overflow icon becomes grey. If we open overflow list again, selection will get removed.

I'm not sure whether this issue is in the scope of this PR.
Other than these, this PR looks good to me.

eduardsdv:
I can reproduce the behavior #1 in both versions before PR and after PR.

But I cannot reproduce the #2 and the #3 (may be I do not the necessary steps). Can you make a video or try to reproduce it the version before PR?

karthikpandelu:
Here is the video of issue #2. Strangely I couldn't reproduce the issue when I tried to record it using Mac's inbuilt tool. But using zoom, I could reproduce the issue alternatively.

https://github.com/openjdk/jfx/assets/26969459/60bd48a9-4239-4a16-9c92-8894363b159d

This is the video of issue #3. In this case I was able to reproduce issue #2 as well.

https://github.com/openjdk/jfx/assets/26969459/3ad8b4f2-0b97-4d8e-b80b-64883a9e83e2

eduardsdv:
So, none of them can be reproduced on windows, only on mac.

The #2 can be reproduced in the version before PR and after PR.

The #3 is reproducible only with PR changes and in my opinion an other bug in JavaFX, which is made to appear by this PR.

It can be quick-fixed by adding a ``Platform.runLater(..)`` into WINDO_HIDDEN event handler:
````java
            popup.addEventHandler(WindowEvent.WINDOW_HIDDEN, e -> {
                // Quick-fix for MacOs:
                // If an item is selected in the overflow list and we select an item from the visible list,
                // selection in the overflow list remains on opening the overflow list for the first time but overflow icon becomes grey.
                // If we open overflow list again, selection will get removed.
                Platform.runLater(() -> {
                    // Put the overflowed items back to the list,
                    // otherwise subsequent prefWidth(..)/prefHeight(..) may return wrong values.
                    overflowItems.clear();
                    for (Node item : getSkinnable().getItems()) {
                        if (!box.getChildren().contains(item)) {
                            overflowItems.add(item);
                        }
                    }
                });
            });
````

What is your suggestion? Should it be **quick**-fixed in this PR or should a new issue be created for it?

andy-goryachev-oracle:
I think these three issues are not related to this PR (even though the behavior seems to have changed a bit).

I would look at this from the perspective of the keyboard-only user.  Can they do everything they need to do, such as selecting/focusing/editing of the items in the toolbar as well as the overflow popup?  (We might need to add a TextField and a CheckBox and maybe a ListView to the toolbar to verify that).

If no function is impaired, we are ok.  If the focus indicator (blue outlines etc) is wrong, or an item cannot be focused or some other function is broken, then we have an issue.

What do you think?

andy-goryachev-oracle:
> Should it be **quick**-fixed in this PR

I would rather not.

We *could* create a ticket named "improve focus handling in ToolBar" but then again, since no functionality is disabled, it will be a P5 (very low priority).

karthikpandelu:
What @andy-goryachev-oracle  mentioned above seems ok to me. If focus indicator function is not broken, we can go ahead with this PR and create separate ticket to fix other issues.

eduardsdv:
I tested the focus handling with the MonkeyTest app. I also tested it with a ListView in the toolbar (see PR: https://github.com/andy-goryachev-oracle/MonkeyTest/pull/2).
All components are reachable with the mouse and the keyboard.
I found only one issue with closing the overflow-menu with the ESC key when the currently focused Node (e.g. ListView) processes the ESC itself. This issue is also reproducible in the version before PR.

andy-goryachev-oracle:
> closing the overflow-menu with the ESC key when the currently focused Node (e.g. ListView) processes the ESC itself.

On macOS, option-ESC closes the overflow menu.

andy-goryachev-oracle:
let's try this one:

Reviewer: @karthikpandelu

kevinrushforth:
> let's try this one:
> 
> Reviewer: @karthikpandelu

Yep.

There is no Skara command to request a review from someone. The `/reviewer` is use to credit someone who has not done a review in GitHub (and is not expected to), but who has reviewed it via some other means. Such an extra reviewer will be credited in the meta-data, but their review will not "count" for the minimum number of reviewers.

karthikpandelu:
Created [JDK-8330859](https://bugs.openjdk.org/browse/JDK-8330859) for tracking focus handling issue in toolbar

karthikpandelu (approved):

PR #1442, opened 2024-04-18
8330462: StringIndexOutOfBoundException when typing anything into TextField

koppor:
Fixes https://bugs.openjdk.org/browse/JDK-8330462.

If the parameter `maxLength` is larger than `Integer.MAX_VALUE - start`, then an addition of `start` to it leads to a negative value. This is "fixed" by using `Math.max` comparing the `maxLength` and `maxLength + start`.

---------
### Progress

### Issue
 * [JDK-8330462](https://bugs.openjdk.org/browse/JDK-8330462): StringIndexOutOfBoundException when typing anything into TextField (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1442/head:pull/1442` \
`$ git checkout pull/1442`

`$ git checkout pull/1442` \
`$ git pull https://git.openjdk.org/jfx.git pull/1442/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1442`

`$ git pr show -t 1442`

</details>

<a href="https://git.openjdk.org/jfx/pull/1442.diff">https://git.openjdk.org/jfx/pull/1442.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1442#issuecomment-2063711432)


andy-goryachev-oracle:
would this fix need to be backported?  how far back?

andy-goryachev-oracle (approved):
The code looks right.  I can't verify the fix with the deepl app.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinTextRangeProvider.java
```diff
@@ -362,7 +362,7 @@ private String GetText(int maxLength) {
         String text = (String)getAttribute(TEXT);
         if (text == null) return null;
         validateRange(text);
-        int endOffset = maxLength != -1 ? Math.min(end, start + maxLength) : end;
+        int endOffset = maxLength >= 0 ? Math.min(end, Math.max(start + maxLength, maxLength)) : end;
```
andy-goryachev-oracle, line 365: there is a similar code on LL 101,102,381,485
would we need to apply the same logic there?

kevinrushforth:
In addition to the comment Andy made about possible other places for the check, I'd like to see a test, if possible. Also, this doesn't follow the usual pattern of checking for integer overflow, so that could be done in the utility method that Andy suggested.

/reviewers 2

koppor:
> I'd like to see a test, if possible.

Where can I find hints on this? I would use TestFX, fire up some minimal app and try to hit the branches of WinTextRangeProvider. However, I did not find TestFX as used. And there is no `javafx.graphics/src/test/java/test/com/sun/glass/ui/win`. There is also no  `javafx.graphics/src/test/java/test/com/sun/javafx/scene/control` for testing a TextField.

I assume, I need to create some class in `test.javafx.scene.input`?

Based on my experience, I would create a test for `WinTextRangeProvider` directly and supply some values to it to check that the "right" branches are covered.

koppor:
> would this fix need to be backported? how far back?

Yes - JavaFX 22 would be helpful for me.

`8u20-b12` was the first tag appearing with that code. -- I checked "git blame" https://github.com/openjdk/jfx/blame/master/modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinTextRangeProvider.java. The commit https://github.com/openjdk/jfx/commit/d66a6dd06eeb199e2e1078f562a8d51de67e4510#diff-17176994e11e44400ad30a7092a1fd738b268811b275884b946e11ce2448c379 changed the implementation from a dummy (`return null;`) to something more meaningful. - I did not find the exact releases.

Since [EOL of JavaFX in Java 8](https://www.oracle.com/uk/java/technologies/java-se-support-roadmap.html) is less than 12 months from now, I would not do it.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinTextRangeProvider.java
koppor: Done! Hopefully, not too overengeered - https://github.com/openjdk/jfx/pull/1442/files#diff-cf00c7b3f2ccd166df5df3aabcd0940072739356910a2c2107a9442aec6010dfR105

(The JDK code seems to be very compact)

kevinrushforth (commented):
I left one inline comment.

I'd like @arapte to review this, since he is familiar with this code, having fixed other IOOBE bugs in it.

kevinrushforth:

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinTextRangeProvider.java
koppor: I implemented it for L101, felt a bit bad and at https://github.com/openjdk/jfx/pull/1442/files/e7502d674c579667545b97ab944e82f2b093a470#diff-cf00c7b3f2ccd166df5df3aabcd0940072739356910a2c2107a9442aec6010df, Kevin also noted it could not be right there. Thus, I reverted to the original.

For the others, I used the new utility method `getValidStringIndex`

koppor:
> This can easily be done with a unit test,

How to create an instance of `WinTextRangeProvider` with a fake `WinAccessible`? I need `WinAccessible.getAttribute(...)`. Normally, I would have used Mockito... `WinAccessible` is `final`. -- I do not dare to remove `final` just for test subclassing ^^.

koppor:
I am new to testing in the JFX project. It seems that `test.` is required as package prefix. Thus, I did not use the approach of package-private methods and classes, but needed to make the class and the tested method public.

However, now I get the (expected) error that the module `javafx.graphics` does not export `com.sun.glass.ui.win` to unnamed module @0x1e6454ec.

```
class test.com.sun.glass.ui.win.WinTextRangeProviderTest (in unnamed module @0x1e6454ec) cannot access class com.sun.glass.ui.win.WinTextRangeProvider (in module javafx.graphics) because module javafx.graphics does not export com.sun.glass.ui.win to unnamed module @0x1e6454ec
java.lang.IllegalAccessError: class test.com.sun.glass.ui.win.WinTextRangeProviderTest (in unnamed module @0x1e6454ec) cannot access class com.sun.glass.ui.win.WinTextRangeProvider (in module javafx.graphics) because module javafx.graphics does not export com.sun.glass.ui.win to unnamed module @0x1e6454ec
	at test.com.sun.glass.ui.win.WinTextRangeProviderTest.getValidStringIndex
```

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinTextRangeProvider.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2014, 2022, 2024, Oracle and/or its affiliates. All rights reserved.
```
andy-goryachev-oracle, line 2: this needs the following form:
```
Copyright (c) 2014, 2024, Oracle and/or its affiliates. All rights reserved.
```

modules/javafx.graphics/src/test/java/test/com/sun/glass/ui/win/WinTextRangeProviderTest.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package test.com.sun.glass.ui.win;
+
+import org.junit.Test;
+import static org.junit.Assert.*;
+import com.sun.glass.ui.win.WinTextRangeProvider;
```
andy-goryachev-oracle, line 29: since it's a new test, would it be possible to convert it to junit5?

andy-goryachev-oracle:
> module `javafx.graphics` does not export `com.sun.glass.ui.win` to unnamed module @0x1e6454ec.

I think you need to add
```
--add-exports javafx.graphics/com.sun.glass.ui.win=ALL-UNNAMED
```
to graphics/src/test/addExports
(see line 7)

@kevinrushforth I wonder if there was a way to auto-generate addExports somehow, at least the part needed for the tests.

kevinrushforth:
> > module `javafx.graphics` does not export `com.sun.glass.ui.win` to unnamed module @0x1e6454ec.
> 
> I think you need to add
> 
> ```
> --add-exports javafx.graphics/com.sun.glass.ui.win=ALL-UNNAMED
> ```

Correct.

> @kevinrushforth I wonder if there was a way to auto-generate addExports somehow, at least the part needed for the tests.

I'd rather have it be explicit. It changes rarely.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinTextRangeProvider.java
```diff
@@ -37,7 +37,7 @@
  * This class is the Java peer for GlassTextRangeProvider.
  * GlassTextRangeProvider implements ITextRangeProvider.
  */
-class WinTextRangeProvider {
+public final class WinTextRangeProvider {
```
kevinrushforth, line 40: Making both this class and the newly added method public just for the purpose of a unit test is not a pattern we want to use. Instead, leave them as default (package) scope and create a public `WinTextRangeProvideShim` class underneath the shims.

kevinrushforth (commented):
I left a comment regarding the unit test. Please use a shim class instead of making `WinTextRangeProvider` public.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinTextRangeProvider.java
```diff
@@ -102,6 +102,19 @@ private void validateRange(String text) {
...
+        int fixedStart = Math.max(0, Math.min(start, fixedMaxEnd));
+        if (requestedSteps < 0) {
+            return fixedMaxEnd;
+        }
+        try {
+            return Math.min(fixedMaxEnd, Math.addExact(fixedStart, requestedSteps));
+        } catch (ArithmeticException e) {
+            return fixedMaxEnd;
+        }
+    }
```
andy-goryachev-oracle, line 116: Frankly, I have hard time understanding this code (maybe a comment describing what the method does and why might help).

It looks to me that all we need to do is to guard against a very large maxLength (which for some reason called here 'requestedSteps' which does not seem right - should the last two arguments be swapped?)

```
public static int getEndIndex(int start, int length, int maxLength) {
  if(length > maxLength) {
    length = maxLength;
  }
  return start + length;
}
```

That is, I assume we don't have to worry about start + fixedLength overflowing, we just need to make sure we don't go beyond maxLength.  Or is my assumption wrong and start can be negative, or start+fixedLength might overflow?

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinTextRangeProvider.java
```diff
@@ -362,7 +375,7 @@ private String GetText(int maxLength) {
         String text = (String)getAttribute(TEXT);
         if (text == null) return null;
         validateRange(text);
-        int endOffset = maxLength != -1 ? Math.min(end, start + maxLength) : end;
+        int endOffset = getValidStringIndex(start, maxLength, end);
```
andy-goryachev-oracle, line 378: If I presume the arguments are (start, length, maxLength), the last two arguments on this line only need to be swapped.  They seem to be correct on LL 394, 498.

koppor:
> I think you need to add

The mentioned issue is fixed. Now, I get 

```
java.lang.UnsatisfiedLinkError: 'void com.sun.glass.ui.win.WinTextRangeProvider._initIDs()'
	at javafx.graphics@23-internal/com.sun.glass.ui.win.WinTextRangeProvider._initIDs(Native Method)
```

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinTextRangeProvider.java
andy-goryachev-oracle: I think the comment in https://github.com/openjdk/jdk/pull/10704 referred to an external project, so it was deemed as unrelated.

but this method, I feel, is rather convoluted.  why do we need to catch exceptions (malloc!) when we should be able to implement the required functionality with a few `if`s.

So I really have two comments: a) the code looks way too complex and b) if an average developer (like me) cannot figure out what it does within 10 seconds, a good comment would help.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinTextRangeProvider.java
koppor: Smalltalk 1: I thought this was easier to understand than some byte code generation in the JVM. 😅 

Smalltalk 2: Sometimes, the code for "Calculate a + b, but return c at most" is pretty hard to craft.

Smalltalk 3: The whole checks stem from possible "out of range" values, especially from the other functions mentioned at https://github.com/openjdk/jfx/pull/1442/#discussion_r1570948582. That was too defensive, as only `requestedSteps` AKA `length` can be out of range.

OK, I seem to have understood "Also, this doesn't follow the usual pattern of checking for integer overflow" by Kevin wrong. I googled the Java way of the usual pattern for Integer overflow. Since Java 8, there is `Math.addExact`. I thought, that this was meant. -- I found it from https://stackoverflow.com/a/3001879/873282. (Inlining the code of `Math.addExact` seems to have negative performance impact.)

The proposed code works OKish if strings are not in length area of Integer.MAX_VALUE. I think, we can safely assume that. - It however returns more characters if start is greater then 0. Example: I request start 2, length of 5, but maximum end index  of 3. Then 3 should be returned, not 7.

---

I changed the code accordingly. Also added a comment when an overflow might happen. From the discussion here, it seems, we can ignore these cases.

Note that the old code returned `0` if `start` was negative. New might return some negative value if `start` is negative enough. However, did not seem to happen, because otherwise, IndexOutOfBounds exceptions might have been seen.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinTextRangeProvider.java
koppor: > So I really have two comments: a) the code looks way too complex and b) if an average developer (like me) cannot figure out what it does within 10 seconds, a good comment would help.

Sure thing.

I applied the one "usual pattern" of Integer overflow at a1c135abf7. Maybe, you like it.

koppor:
There is use of native code in the `static` block

```java
    private native static void _initIDs();
    static {
        _initIDs();
    }
```

Currently, all tests fail because that native library is not found

    java.lang.UnsatisfiedLinkError: 'void com.sun.glass.ui.win.WinTextRangeProvider._initIDs()'

I don't know how to load native libraries here. - I could also change that code part of WinTextRangeProvider to behave different in test mode. But that is a bad pattern, isn't it?

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinTextRangeProvider.java
andy-goryachev-oracle: Thank you!  Somehow the code looks much cleaner and clearer now.

kevinrushforth:
> > I think you need to add
> 
> The mentioned issue is fixed. Now, I get
> 
> ```
> java.lang.UnsatisfiedLinkError: 'void com.sun.glass.ui.win.WinTextRangeProvider._initIDs()'
> 	at javafx.graphics@23-internal/com.sun.glass.ui.win.WinTextRangeProvider._initIDs(Native Method)
> ```

Ah, right, sorry for not noticing this earlier. You cannot access glass from unit tests in the `javafx.graphics` module. Those tests use the StubToolkit which bypasses glass. You will need to move the test to the system tests project under `tests/system/src/test/java/`. Those will use the real toolkit and the glass libraries will be loaded for you. You will need to explicitly start the toolkit (via `Platform.startup` or a call to a utility method to do the same) in an `@BeforeAll` method.

@andy-goryachev-oracle or @arapte should be able to help you with the transition. You would still use the same shim class, which should remain under `javafx.graphics`.

koppor:
I was thinking about diving into this accessible area. Did not find any existing test and wanted to start with something small (in `test.javafx.scene.control.TextFieldTest`).

```
    @Test public void setTextAndSeeValueAccessible() throws Exception {
        StackPane stackPane = new StackPane();
        Scene scene = new Scene(stackPane);
        Accessible accessible = SceneHelper.getAccessible(scene);
        TextField textField = new TextField();
        stackPane.getChildren().add(textField);
        textField.setText("This is a test");
        assertEquals("This is a test", accessible.getAttribute(AccessibleAttribute.TEXT));
    }
```

Fails, because `SceneHelper.getAccessible(scene)` returns `null`. I am not that much into the accessibility activation things - and did not find any accessibility tests.

tests/system/src/test/java/test/com/sun/glass/ui/win/WinTextRangeProviderTest.java
```diff
@@ -0,0 +1,99 @@
...
+import org.junit.jupiter.api.condition.EnabledOnOs;
+import org.junit.jupiter.api.condition.OS;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import static org.junit.Assert.fail;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@EnabledOnOs({OS.WINDOWS})
```
koppor, line 48: I did not use `Platform`s OS detection, because this seems to be more resource friendly and native JUnit5. I can switch to the style of `test.com.sun.glass.ui.gtk.Gtk2RemovalCommon`, i.e., use `if (!PlatformUtil.isWindows()) return;` in all test methdos.

modules/javafx.graphics/src/test/addExports
```diff
@@ -5,6 +5,7 @@
 --add-exports javafx.base/com.sun.javafx.logging=ALL-UNNAMED
 #
 --add-exports javafx.graphics/com.sun.glass.ui=ALL-UNNAMED
+--add-exports javafx.graphics/com.sun.glass.ui.win=ALL-UNNAMED
```
koppor, line 8: This is needed because of the Shim (IMHO)

koppor:
@andy-goryachev-oracle I think, I'm finished. (The "tests" requested at https://github.com/openjdk/jfx/pull/1442#issuecomment-2064196119 were IMHO the "simple" unit tests I provided - not some accessibility engine tests)

tests/system/src/test/java/test/com/sun/glass/ui/win/WinTextRangeProviderTest.java
andy-goryachev-oracle: Interesting = @kevinrushforth what do you think?

tests/system/src/test/java/test/com/sun/glass/ui/win/WinTextRangeProviderTest.java
```diff
@@ -0,0 +1,99 @@
...
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package test.com.sun.glass.ui.win;
+
+import com.sun.javafx.PlatformUtil;
+import com.sun.glass.ui.win.WinTextRangeProviderShim;
```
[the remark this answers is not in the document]
andy-goryachev-oracle: Please replace tests/system/src/test/.classpath with this:

```
<classpath>
	<classpathentry combineaccessrules="false" kind="src" path="/swing">
		<attributes>
			<attribute name="module" value="true"/>
		</attributes>
	</classpathentry>
	<classpathentry combineaccessrules="false" kind="src" path="/base">
		<attributes>
			<attribute name="module" value="true"/>
			<attribute name="add-exports" value="javafx.base/com.sun.javafx=ALL-UNNAMED:javafx.base/test.util.memory=ALL-UNNAMED"/>
		</attributes>
	</classpathentry>
	<classpathentry combineaccessrules="false" kind="src" path="/graphics">
		<attributes>
			<attribute name="module" value="true"/>
			<attribute name="add-exports" value="javafx.graphics/com.sun.glass.ui=ALL-UNNAMED:javafx.graphics/com.sun.glass.ui.monocle=ALL-UNNAMED:javafx.graphics/com.sun.javafx.sg.prism=ALL-UNNAMED:javafx.graphics/com.sun.prism.impl=ALL-UNNAMED:javafx.graphics/com.sun.javafx.image.impl=ALL-UNNAMED:javafx.graphics/com.sun.glass.events=ALL-UNNAMED:javafx.graphics/com.sun.javafx.application=ALL-UNNAMED:javafx.graphics/com.sun.javafx.css=ALL-UNNAMED:javafx.graphics/com.sun.javafx.geom=ALL-UNNAMED:javafx.graphics/com.sun.javafx.tk=ALL-UNNAMED:javafx.graphics/com.sun.glass.ui.mac=ALL-UNNAMED:javafx.graphics/com.sun.glass.ui.win=ALL-UNNAMED:javafx.graphics/com.sun.javafx.scene.text=ALL-UNNAMED:javafx.graphics/com.sun.javafx.text=ALL-UNNAMED:javafx.graphics/com.sun.javafx.font=ALL-UNNAMED"/>
		</attributes>
	</classpathentry>
	<classpathentry combineaccessrules="false" kind="src" path="/controls">
		<attributes>
			<attribute name="module" value="true"/>
			<attribute name="add-exports" value="javafx.controls/test.com.sun.javafx.scene.control.infrastructure=ALL-UNNAMED:javafx.controls/com.sun.javafx.scene.control=ALL-UNNAMED"/>
		</attributes>
	</classpathentry>
	<classpathentry combineaccessrules="false" kind="src" path="/fxml">
		<attributes>
			<attribute name="module" value="true"/>
		</attributes>
	</classpathentry>
	<classpathentry combineaccessrules="false" kind="src" path="/web">
		<attributes>
			<attribute name="module" value="true"/>
			<attribute name="add-exports" value="javafx.web/com.sun.webkit=ALL-UNNAMED"/>
		</attributes>
	</classpathentry>
	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER">
		<attributes>
			<attribute name="module" value="true"/>
			<attribute name="add-exports" value="java.desktop/sun.awt.datatransfer=ALL-UNNAMED"/>
		</attributes>
	</classpathentry>
    <classpathentry kind="con" path="org.eclipse.jdt.junit.JUNIT_CONTAINER/5">
        <attributes>
            <attribute name="test" value="true"/>
        </attributes>
    </classpathentry>
	<classpathentry kind="src" output="bin" path="java">
	    <attributes>
            <attribute name="gradle_scope" value="test"/>
            <attribute name="gradle_used_by_scope" value="test"/>
            <attribute name="test" value="true"/>
        </attributes>
	</classpathentry>
	<classpathentry kind="src" output="bin" path="resources">
        <attributes>
            <attribute name="gradle_scope" value="test"/>
            <attribute name="gradle_used_by_scope" value="test"/>
            <attribute name="test" value="true"/>
        </attributes>
    </classpathentry>
	<classpathentry kind="output" path="bin"/>
</classpath>
```

(the only change is on line 17)

andy-goryachev-oracle (commented):
Turns out I cannot reproduce the issue on my windows 11 setup.
Installed deepl app, ctrl-c-c pops up the translation window, translation works, close the window, start the test app - can type into a TextField just fine.

the code looks ok to me, one change must be made to a .classpath file - take a look at
https://github.com/andy-goryachev-oracle/jfx/pull/new/review.1442
branch (I think you can just merge it into yours)

tests/system/src/test/java/test/com/sun/glass/ui/win/WinTextRangeProviderTest.java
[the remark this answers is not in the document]
koppor: Done.

I opened the project in Eclipse (instead of IntelliJ). How can I run the test in Eclipse?

A simple "run" does not work, because the output is then:

```
WARNING: Unknown module: javafx.base specified to --patch-module
WARNING: Unknown module: javafx.web specified to --patch-module
WARNING: Unknown module: javafx.controls specified to --patch-module
WARNING: Unknown module: javafx.graphics specified to --patch-module
```

How do I need to modify the JUnit "run configuration" inside Eclipse?

arapte:
Sorry for jumping in little late.
I am reviewing the changes, shall test the fix with some usecases and update.

tests/system/src/test/java/test/com/sun/glass/ui/win/WinTextRangeProviderTest.java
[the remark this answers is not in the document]
andy-goryachev-oracle: Yes, running tests in eclipse requires a bit of set up.  First run the (new) test with  the usual 'Run', this creates a run configuration (should appear under 'JUnit').  Then open that via main menu Run -> Run Configurations... , select "Dependencies" tab, click on "Override Dependencies" button and then the pain starts.

Depending on the type of test, you'd need to either replace or append various `--add-exports`, `--add-modules`, `-Djava.library.path=`, and `-Djavafx.toolkit= `

tests/system/src/test/java/test/com/sun/glass/ui/win/WinTextRangeProviderTest.java
kevinrushforth: An excellent question. If it is robust, then it seems OK. Using Junit5 Assumptions is more flexible, though, so might lean toward wanting to use that. Either way, test this on the other platforms to ensure that the test is skipped (and marked as skipped in the report).

arapte:
I tested the fix with a simple TextArea sample and observed an issue that: Windows Narrator reads only the last character of the Text in a TextArea, when moving the cursor, and the focus rect does not move with cursor.

Steps:
1. Launch Windows Narrator (shortcut key combination: Windows + Ctrl + Enter)
2. Run a simple TextArea sample with fix 
```
public class TextAreaTest extends Application  {
    @Override
    public void start(Stage primaryStage) throws Exception {
        primaryStage.setScene(new Scene(new VBox(new TextArea("javafx test"))));
        primaryStage.show();
    }
}

```
-> Observe that a Fcous Rect is drawn around all text in TextArea.

3. Press right arrow key to move cursor
-> Observe that:
i. Focus rect is drawn around the last character, even though the cursor is elsewhere
ii. With every right arrow key press, Narrator always read 't'

tests/system/src/test/java/test/com/sun/glass/ui/win/WinTextRangeProviderTest.java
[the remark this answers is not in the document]
kevinrushforth: @koppor Running the test in Eclipse is optional, so maybe Andy can test that. As long as the test runs in gradle, that's sufficient.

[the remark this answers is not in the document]
andy-goryachev-oracle: optional, yes, I verified that the eclipse config change is good.

tests/system/src/test/java/test/com/sun/glass/ui/win/WinTextRangeProviderTest.java
koppor: +1 for flexibility. Switched in [`ade506c` (#1442)](https://github.com/openjdk/jfx/pull/1442/commits/ade506c0e44544f3d61a164765e19b983a99373d).

Note that `assumeTrue` also works in `@BeforeAll` to skip the whole class. (I wonder whether I should rewrite the other tests following that pattern, but that is maybe a job for a follow-up issue)

koppor:
> Windows Narrator reads only the last character of the Text in a TextArea, when moving the cursor, and the focus rect does not move with cursor.

I tested it with a JFX distribution without the fix. Also happens there. Can you check, too? - If yes, would it be possible to handle this in a separate issue?

andy-goryachev-oracle:
> I tested it with a JFX distribution without the fix. Also happens there.

Nope, this fix breaks Narrator.  The window cursor is moving but the narrator outlines the trailing 't' as Ambarish described.

(every time I change branches for the purposes of review, I do a complete gradle build preceded by nuking the build directory:
```
rm -rf ./build ; gradle clean sdk apps javadoc
```)

**Comment by koppor** [comment]
> Nope, this fix breaks Narrator.

I think, I do not get what Narrator is doing.

If I type "Testx" into a Text field, what should be highlighted?

image

If moving the cursor in the tool [javafxreproducer](https://github.com/Siedlerchr/javafxreproducer) also always the last letter is highlighted. Does my patch sneakly sneak in in all JavaFX code I am running on my machine?

**Comment by koppor** [comment]
> focus rectangle follows the windows cursor and announces the letter at the cursor.

I did not change branches, but downloaded a binary from the net.

I type "abcdef" in the text field and then press cursor left - no change of the letter.

Maybe, this behavior was fixed in the main branch meanwhile. The binary uses JavaFX 21.

**Comment by koppor** [comment]
... and no luck with Eclipse ...

```
WARNING: package test.com.sun.javafx.scene.control.test not in javafx.controls
Graphics Device initialization failed for :  d3d, sw
Error initializing QuantumRenderer: no suitable pipeline found
```

**Comment by calixtus** [comment]
Tried on my local machine w/o patch with narrator:
grafik

**Comment by andy-goryachev-oracle** [comment]
> ... and no luck with Eclipse ...

I assume you imported the whole jfx repository into Eclipse as a gradle build as described here
https://wiki.openjdk.org/display/OpenJFX/Using+an+IDE#UsinganIDE-UsingEclipse

If not, you could still use command line to build and execute tests, and that is described here 
https://wiki.openjdk.org/display/OpenJFX/Building+OpenJFX

I admit, there is a bit of pain configuring Eclipse - the main build uses gradle and Eclipse has other ideas when it sees a gradle project, but once you figure it out (I can help) you can debug in eclipse and it is much easier, in my opinion.

On the other hand, we could completely take the IDE out of the picture and just run the command line to build and test.

**Comment by andy-goryachev-oracle** [comment]
> Tried on my local machine (win10) w/o patch with narrator:

Thank you for checking this out!

so you see the windows cursor decouple from the narrator focus rectangle?  And when you press left/right, the narrator focus does not change?

just to confirm, you are running the latest master branch, or if now, which version of jfx/jdk?

**Comment by calixtus** [comment]
I tested with Oracle OpenJDK 22.0.1 and JavaFX 22.0.1+7

**Comment by koppor** [comment]
(I switched to TextField; is a bit easier for me to debug)

Intermedidate testing result:

When pressing "space", the rectangle moves correctly. When moving the cursor "inside" these much spaces, it also moves.

image

---

Ctrl+A leads to the box borders be of length of the complete text, not the visible text

image

---

I thought, following would help, but does not:

```diff
     */
     public void selectRange(int anchor, int caretPosition) {
         caretPosition = Utils.clamp(0, caretPosition, getLength());
         anchor = Utils.clamp(0, anchor, getLength());
 
+        notifyAccessibleAttributeChanged(AccessibleAttribute.SELECTION_START);
+        notifyAccessibleAttributeChanged(AccessibleAttribute.SELECTION_END);
+        notifyAccessibleAttributeChanged(AccessibleAttribute.CARET_OFFSET);
+
         TextFormatter.Change change = new TextFormatter.Change(this, getFormatterAccessor(), anchor, caretPosition);
         TextFormatter<
```

koppor:
On branch `main`

`rm -rf build && ./gradlew sdk && ./gradlew  --info -PFULL_TEST=true systemTests:test --tests "test.com.sun.glass.ui.win.TextAreaTest_Narrator_8330462" || ./gradlew  --info -PFULL_TEST=true -PTEST_ONLY=true systemTests:test --tests "test.com.sun.glass.ui.win.TextAreaTest_Narrator_8330462"`

With [Carnac](https://github.com/Code52/carnac/) and [ShareX](https://getsharex.com/) I tried to capture what happens on **`main`**. Showing especially interactions with <kbd>Space</kbd> and <kbd>Cursor-left</kbd>.

https://github.com/openjdk/jfx/assets/1366654/f25bb801-565c-4d50-9471-c4391962049d

arapte:
Here is a capture from my machine.
I have two copies of source code: One with fix and other without fix.
Please do unmute the video playback, you can listen to Narrator voice output.

https://github.com/openjdk/jfx/assets/11330676/bd3efcae-c661-4906-8388-fcb373e7e5c7

kevinrushforth:
> Here is a capture from my machine.
> I have two copies of source code: One with fix and other without fix.
> Please do unmute the video playback, you can listen to Narrator voice output.

@arapte Thank you for doing this and attaching it. This matches what I see and is, I think, what Andy is seeing as well.

@koppor This clearly shows that the current version of the fix in this PR introduces a regression. You will need to fix it before we can continue the review. While testing and debugging this problem, use a manual program rather that a Robot-based automated test.

Speaking of tests, we cannot reliably test a11y using automated tests, so I wouldn't spend any more time trying to get your new `TextAreaTest_Narrator_8330462` test working. The automated test in this PR that verifies the correctness of the new `getEndIndex` method will have to do (along with manual testing of the a11y functionality). Once you figure out what the problem is with the `getEndIndex` method, you might want to add additional cases to your test.

koppor:
Sure thing! I will work on that!

And I will forget about the TextField. TextArea FTW!

koppor:
@arapte Please share your concrete Windows setup with me. I cannot reproduce here. I bet, you are on Windows 11. Reason: Here on my Windows 10 machine (and on @calixtus and on @siedlerchr's machine), the Narrator does not work, but the Narrator works on Windows 11. (from `master` branch)

Currently, I can only move on forward blindly. I hope, I find someone with a Windows 11 machine willing to try out the `base` branch of my fork and help me with the values... https://github.com/koppor/jfx/pull/4

---

I could get JFX to get compile on Windows 10, but I have issues to get it to run on Windows 11. You know, this things "java.util.concurrent.ExecutionException: org.gradle.process.internal.ExecException: A problem occurred starting process 'command '/VC/BIN/amd64/cl.exe'' and `rc.exe` not being found etc. (Yes, I followed the instructions, but the binaries are in `Hostx64/x64` instead of `x64`; some tools are only in `Windows Kits\10`. The installation instructions tell about Visual Studio 2022, but the example is windows_tools.properties is for 2019 (https://wiki.openjdk.org/display/OpenJFX/Building+OpenJFX).

koppor:
@arapte OK, I have one more idea: The usage of the utility method in the non-issue related methods. I reverted the change at that place. Could you please recheck if the Narrator works on Windows 11? (I am still unable to compile JavaFX. Now, it's `stdio.h` missing)

andy-goryachev-oracle (approved):
Regression fixed.  No differences in behavior doing a quick test with with TextField/TextArea and Narrator on Windows 11.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinTextRangeProvider.java
```diff
@@ -98,8 +100,28 @@ private void validateRange(String text) {
         }
 
         int length = text.length();
-        start = Math.max(0, Math.min(start, length));
-        end = Math.max(start, Math.min(end, length));
+        start = Utils.clamp(0, start, length);
+        end = Utils.clamp(start, end, length);
```
arapte, line 104: This is only cleanup and not required for this fix.

tests/system/src/test/java/test/com/sun/glass/ui/win/WinTextRangeProviderTest.java
```diff
@@ -0,0 +1,104 @@
...
+                Arguments.of(50, 50, Integer.MAX_VALUE, 50),
+                Arguments.of(Integer.MAX_VALUE, 0, Integer.MAX_VALUE, Integer.MAX_VALUE),
+                Arguments.of(60, 50, -1, 60),
+                Arguments.of(60, 50, Integer.MIN_VALUE, 60),
+
+                // No range check for startIndex
+                Arguments.of(20, -10, 9, 20),
+
+                // No range check for maxEndIndex
+                Arguments.of(-1, -1, 2, -1),
```
arapte, line 93: negative valued tests for `start` or `end` are not required for this particular scenario as `0 <= start <= end <= total length of text` is guaranteed.

arapte (changes requested):
In addition to the evaluation added by @jperedadnr to [JDK-8330462](https://bugs.openjdk.org/browse/JDK-8330462)
- The UI Automation framework offers two sets of APIs.
    1. Provider APIs : to be implemented by UI providers ( JavaFX )
    2. Client APIs : to be implemented by the Client applications like (Screen readers, DeepL)
- Both providers(JavaFX) and client application should follow their respective specification.
- For this scenario
    1. Client Application should follow : [IUIAutomationTextRange::GetText method](https://learn.microsoft.com/en-us/windows/win32/api/uiautomationclient/nf-uiautomationclient-iuiautomationtextrange-gettext#parameters)
    2. JavaFX should follow : [ITextRangeProvider::GetText](https://learn.microsoft.com/en-us/windows/win32/api/uiautomationcore/nf-uiautomationcore-itextrangeprovider-gettext#parameters)
- As per both above spec, the length variable could be either -1 or a valid value.
- The check we had earlier: `length != -1` was based on the spec, but seems like we need to validate against out of range check i.e. `length <= end - start`.
- It seems in this scenario DeepL invokes the GetText call with INT_MAX value, which is the root cause. The check `length <= end - start` should handle that scenario.
- and `length < 0` would future protect us from any negative value though as per spec `length != -1` should suffice

Testing:
- I could reproduce the issue with DeepL
- Issue gets fixed with this fix and also with the change that I suggesting
- A couple tests of negative values for start and end should be removed.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinTextRangeProvider.java
koppor: I applied the software engineering principle to leave the code cleaner than seen. (Martin Fowler et all) 

Should I revert this?

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinTextRangeProvider.java
```diff
@@ -98,8 +100,28 @@ private void validateRange(String text) {
...
+     *                    (not checked in the code).
+     */
+    static int getEndIndex(int startIndex, int length, int maxEndIndex) {
+        if (length < 0 || length > maxEndIndex) {
+            return maxEndIndex;
+        }
+        int res = Math.min(startIndex + length, maxEndIndex);
+        // In case there was an overflow, return the maximum end index
+        if (res < 0) return maxEndIndex;
+        return res;
```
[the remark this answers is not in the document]
koppor: Smalltalk: I could go back to my first commit https://github.com/openjdk/jfx/commit/e81712ca12035e6607c9a31379fa0180be8be7fd to be consistent with the other style in the class 🤣🤣🤣. Only joking, the new code is more readable to Java newcomers IMHO. 😅✌️

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinTextRangeProvider.java
arapte: Sorry for not being clear. I just wanted to point out that it is not required for the issue fix.
Please don't revert. This looks better.

koppor:
> In addition to the [evaluation](https://bugs.openjdk.org/browse/JDK-8330462?focusedId=14666181&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14666181) added by `@jperedadnr` to [JDK-8330462](https://bugs.openjdk.org/browse/JDK-8330462)

Thank you for the long investigation!

I incorporated your updates, renamed the end variable to `endIndex`, and tried to add some more JavaDoc mirroring your input. - Reads good for me now :)

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinTextRangeProvider.java
kevinrushforth: This is the sort of tangentially-related change that should usually be avoided. In this specific case, `clamp(min, val, max)` will produce the same results as `max(start, min(val, end))`, but it isn't guaranteed to do so for malformed input where `max < min`. Given that `text.length()` always returns a non-negative number, the first call to `clamp` is identical to the existing code. And since the resulting` start` value is guaranteed to be `<= length`, the second call to `clamp` is also equivalent.

But... reviewers shouldn't have to spend time verifying this to ensure no regression. Since Ambarish is the primary reviewer of this, and he is OK with this change (and I already did spend the time to verify that it is equivalent), then I don't insist that you revert it.

kevinrushforth (commented):
The latest version of the fix and the test looks good to me, although I'll let @arapte and @andy-goryachev-oracle formally review it.

arapte (changes requested):
The fix looks good, I have few comments about the test.
1. The test does not compile without fix, hence it won't fail without fix as we are only testing the newly added helper method.
2. It is not required to change the error stream, as the test does/need not inspect the error output.
3. The test method and parameter source method names can be changed a little.
4. We should use the Util class for standard setup like initializing JavaFX/ shutting it down.


I tried above changes to test locally, attaching the file for ease.
[WinTextRangeProviderTest.zip](https://github.com/openjdk/jfx/files/15186626/WinTextRangeProviderTest.zip)

koppor:
>     1. The test does not compile without fix, hence it won't fail without fix as we are only testing the newly added helper method.

Yes, we could not build on existing a11y test.

>     2. It is not required to change the error stream, as the test does/need not inspect the error output.

Nice!

>     3. The test method and parameter source method names can be changed a little.

Note that `@MethodSoruce` can be used without any parameter - then the same name is used. OK, this is not the style of JFX.

For JUnit5, the `test` prefix is not necessary any more (and can be removed - see https://docs.openrewrite.org/recipes/java/testing/cleanup/removetestprefix), but I think, because of consistency, it is kept.

>     4. We should use the Util class for standard setup like initializing JavaFX/ shutting it down.

Nice!

> I tried above changes to test locally, attaching the file for ease. [WinTextRangeProviderTest.zip](https://github.com/openjdk/jfx/files/15186626/WinTextRangeProviderTest.zip)

Thank you. I committed at d03bdd40a3340bd85397f (with you as author, hope this is in line with the policies?!).

arapte:
>Thank you. I committed at https://github.com/openjdk/jfx/commit/d03bdd40a3340bd85397fe2ceaa7f053a58da6c5 (with you as author, hope this is in line with the policies?!).

Thanks @koppor, the fix looks good.
That's a great gesture, thanks, but I was just reviewing the changes, you don't have to add me as a contributor. 
Thanks for your contribution, Approving.!

arapte (approved):

tests/system/src/test/java/test/com/sun/glass/ui/win/WinTextRangeProviderTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @BeforeAll
+    static void initFX() throws Exception {
+        assumeTrue(PlatformUtil.isWindows());
+        Util.startup(startupLatch, () -> startupLatch.countDown());
+    }
+
+    @AfterAll
+    static void shutdown() {
+        Util.shutdown();
```
kevinrushforth, line 56: This fails on Mac and Linux because the initialization (and consequently all tests) are skipped, but the call to `Util.shutdown` is not skipped. You will either need to also add `assumeTrue(PlatformUtil.isWindows());` here or else move the `assumeTrue` to the test method.

kevinrushforth (changes requested):
The test gets an error on Mac and Linux. See inline comments.

tests/system/src/test/java/test/com/sun/glass/ui/win/WinTextRangeProviderTest.java
andy-goryachev-oracle: is there a good way to skip the whole file entirely, e.g. via junit annotation?

andy-goryachev-oracle: maybe
https://stackoverflow.com/questions/26115139/junit-ignore-or-skip-entire-test-class
?

kevinrushforth: We still want to use `assumeTrue` as the pattern (for flexibility), so the question is whether there is a better place besides a method annotated with `@BeforeAll` to put it (which also means duplicating the assumption in the `@AfterAll` method). I'm not aware of one, but maybe someone else is?

andy-goryachev-oracle (approved):
The code changes look good to me; don't see any problems testing with Narrator on win11.  I do have Deepl installed (though it is not connected due to proxy).  See no issues in TextField, PasswordField, TextArea in the test app as well as the Monkey Tester.

(once the tests are fixed, I'll re-approve)

tests/system/src/test/java/test/com/sun/glass/ui/win/WinTextRangeProviderTest.java
Siedlerchr: Junit 5 has an EnabledOnOs annotation https://junit.org/junit5/docs/5.2.0/api/org/junit/jupiter/api/condition/EnabledOnOs.html

andy-goryachev-oracle:
please do not integrate before tests are fixed...

tests/system/src/test/java/test/com/sun/glass/ui/win/WinTextRangeProviderTest.java
andy-goryachev-oracle: > EnabledOnOs

thank you @Siedlerchr for this info.
I wonder what criteria are set for determining the OS value, and do they match jfx ones?

Siedlerchr: They just read the `os.name` with `System.getProperty("os.name")`
https://github.com/junit-team/junit5/blob/db47616ab4ccf38ff63e8bff41050d5102c9ff15/junit-jupiter-api/src/main/java/org/junit/jupiter/api/condition/OS.java#L110

andy-goryachev-oracle: the logic **is** different, but both variants should be ok for testing purposes.

for example,
junit: mac = toLowercase(ENGLISH).contains("mac")
jfx: mac = startsWith("Mac")

kevinrushforth: I prefer to stick with assumptions since that's what we use everywhere else. Also, it gives us more control.

So while we _could_ do something else in the future, this PR isn't the place to do it.

kevinrushforth (approved):
Looks good. I confirm that the test passes on Windows and is skipped on macOS.

andy-goryachev-oracle (approved):
LGTM

arapte (approved):

kevinrushforth:
@koppor This is now ready for you to `/integrate`.

koppor:
@arapte @andy-goryachev-oracle @kevinrushforth Thank you for your ongoing support, deep analysis (both the issue and the algorithms needed), and code pointers and guidance. Looking forward to the next JavaFX release where this fix is included!

andy-goryachev-oracle:
Thank you @koppor for fixing the issue!

kevinrushforth:
> Looking forward to the next JavaFX release where this fix is included!

It should be available in the next EA build of JavaFX 23. @arapte will backport it to jfx22u for JavaFX 22.0.2.

PR #1445, opened 2024-04-23
8320563: Remove D3D9 code paths in favor of D3D9Ex

lukostyra:
JFX minimum requirements guarantee 9Ex availability, so old non-Ex paths are no longer needed.

In multiple parts (ex. Mesh, Graphics, etc.) where the Device is acquired I changed the type to explicitly use `IDirect3DDevice9Ex`. Technically it doesn't matter much (`IDirect3DDevice9Ex` inherits `IDirect3DDevice` - it was leveraged to transparently use the Ex device in the backend) but now we don't have the non-Ex device, so that keeps it a bit more consistent and clear IMO.

Verified by running tests on Windows 11, did not notice any regressions. Unfortunately I have no way to test this on older systems.

---------
### Progress

### Issue
 * [JDK-8320563](https://bugs.openjdk.org/browse/JDK-8320563): Remove D3D9 code paths in favor of D3D9Ex (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1445/head:pull/1445` \
`$ git checkout pull/1445`

`$ git checkout pull/1445` \
`$ git pull https://git.openjdk.org/jfx.git pull/1445/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1445`

`$ git pr show -t 1445`

</details>

<a href="https://git.openjdk.org/jfx/pull/1445.diff">https://git.openjdk.org/jfx/pull/1445.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1445#issuecomment-2071974710)

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/prism/d3d/D3DRTTexture.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/prism/d3d/D3DRTTexture.java
@@ -172,7 +172,7 @@ class D3DRTTexture extends D3DTexture
 
     @Override
     public boolean isVolatile() {
-        return getContext().isRTTVolatile();
+        return false;
     }
 
     @Override
--- a/modules/javafx.graphics/src/main/native-prism-d3d/D3DPipeline.cc
+++ b/modules/javafx.graphics/src/main/native-prism-d3d/D3DPipeline.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2007, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2007, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -30,10 +30,8 @@
 
 // d3d9.dll library dynamic load
 HMODULE hLibD3D9 = 0;
-typedef IDirect3D9 * WINAPI FnDirect3DCreate9(UINT SDKVersion);
 typedef HRESULT WINAPI FnDirect3DCreate9Ex(UINT SDKVersion, IDirect3D9Ex**);
 
-FnDirect3DCreate9 * pD3D9FactoryFunction = 0;
 FnDirect3DCreate9Ex * pD3D9FactoryExFunction = 0;
 
 static jboolean checkAndClearException(JNIEnv *env) {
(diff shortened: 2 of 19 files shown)


modules/javafx.graphics/src/main/java/com/sun/prism/d3d/D3DRTTexture.java
```diff
@@ -172,7 +172,7 @@ public boolean isOpaque() {
 
     @Override
     public boolean isVolatile() {
-        return getContext().isRTTVolatile();
+        return false;
```
lukostyra, line 175: Quick note - this was checking on native side if we have non-Ex device and returning `true` if we did, `false` if D3D9 was initialized with Ex.

So far I only kept these changes within D3D backend. However, I checked the rest of the code and it seems that only D3D9 non-Ex ever returned anything else than `false`. Right now all the backends (d3d, egl, sw) return `false` in this method, which technically leaves a bit of dead code inside common part of Prism. Should we consider removing this method completely?

kevinrushforth: > Should we consider removing this method completely?

Not as part of this PR. Maybe in a follow-up if you think it is worth doing as a (low priority) cleanup.

kevinrushforth:
I skimmed the PR and it looks good to me, but I'll let @arapte be the primary reviewer.

/reviewers 2

kevinrushforth:
Good to hear.

modules/javafx.graphics/src/main/native-prism-d3d/D3DPipeline.cc
```diff
@@ -237,12 +229,12 @@ void fillOsVersionInformation(JNIEnv *env, jobject object, jclass clazz) {
...
 }
 
-inline IDirect3D9* addRef(IDirect3D9* i) {
+inline IDirect3D9Ex* addRef(IDirect3D9Ex* i) {
     i->AddRef();
     return i;
 }
 
-int getMaxSampleSupport(IDirect3D9 *d3d9, UINT adapter) {
+int getMaxSampleSupport(IDirect3D9Ex *d3d9, UINT adapter) {
```
nlisker, line 237: Minor: In some cases you also change the name of the variable to add the "Ex" suffix., like in
```cpp
D3DContext::D3DContext(IDirect3D9Ex *pd3dEx, UINT adapter)
                                         ^
```
Here and In `PiplineManager.h` it's left as `IDirect3D9Ex * pd3d9;` without "Ex".
I don't mind it either way (I would probably not bother changing them myself), but perhaps we should remain consistent.

nlisker (approved):
Tested the functionality with the 3DLighting app, things look the same as before the patch. Left a minor comment.

modules/javafx.graphics/src/main/native-prism-d3d/D3DPipeline.cc
lukostyra: You're right. Most places use it as `pd3d9` so I'll leave it this way (remove the `Ex` suffix from other spots).

lukostyra:
I also noticed there's a property in Prism settings called `disableD3D9Ex` - I will also remove it as it is now redundant.

nlisker (approved):

mstr2 (approved):
LGTM

PR #1447, opened 2024-04-30
8331319: IME Window breaks after popup menu

beldenfox:
When focus moves away from a node JavaFX calls `finishInputMethodComposition` so glass can clean up any in-progress IME editing. On Mac we call `discardMarkedText` on the view's NSTextInputContext to dismiss the IME.

It appears that the OS can get confused if `discardMarkedText` is called on an NSTextInputContext that is not the current active context. JavaFX popups are displayed in windows that don't have OS focus and therefore do not have the active input context but the JavaFX scene associated with the popup doesn't know this and still makes calls to manipulate the IME. This seems to be triggering a bug in the OS that leads to bad behavior which persists until the user moves focus away from the main JavaFX stage altogether and then brings it back.

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassView3D.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassView3D.m
@@ -742,7 +742,12 @@
 {
     IMLOG("finishInputMethodComposition called");
     [self unmarkText];
-    [self.inputContext discardMarkedText];
+    // If we call discardMarkedText on an input context that is not
+    // the current one the IM will get into a persistent state where
+    // it will not call setMarkedText or firstRectForCharacterRange.
+    if (self.inputContext == NSTextInputContext.currentInputContext) {
+        [self.inputContext discardMarkedText];
+    }
 }
 
 /*


kevinrushforth:
Reviewers: @andy-goryachev-oracle @kevinrushforth 

/reviewers 2

andy-goryachev-oracle:
Seems to work with the little test program.

Observed a weird behavior in the Monkey Tester

- open TextArea page
- set font to System Regular
- select text: "Writing Systems"
- click on any English word, invoke IME (I use Japanese), observe candidate inserted correctly
- click on non-English word, invoke IME, the candidate is still shows up correctly
- click on an English word, invoke IME and the candidate shows garbage.  

It seems to me it might be using font from the previous invocation?
This could be a different issue though.

Screenshot 2024-04-30 at 08 18 55

andy-goryachev-oracle:
same issue without the fix.
I won't be able to try with an earlier jdk in the next couple of hours, sorry.

andy-goryachev-oracle:
With jfx21-ea+12, I see the same incorrect font, so it must be a different issue.

```
java --module-path "REDACTED/Java/javafx-sdk-21/lib" --add-modules javafx.base,javafx.graphics,javafx.controls,javafx.swing,javafx.web -jar "REDACTED/MonkeyTest2/MonkeyTest/dist/MonkeyTester.jar"
```

andy-goryachev-oracle (approved):
ship it!

PR #1081, opened 2023-04-04
8290310: ChangeListener events are incorrect or misleading when a nested change occurs

hjohn:
This provides and uses a new implementation of `ExpressionHelper`, called `ListenerManager` with improved semantics.

See also #837 for a previous attempt which instead of triggering nested emissions immediately (like this PR and `ExpressionHelper`) would wait until the current emission finishes and then start a new (non-nested) emission.

# Behavior

|Listener...|ExpressionHelper|ListenerManager|
|---|---|---|
|Invocation Order|In order they were registered, invalidation listeners always before change listeners|(unchanged)|
|Removal during Notification|All listeners present when notification started are notified, but excluded for any nested changes|Listeners are removed immediately regardless of nesting|
|Addition during Notification|Only listeners present when notification started are notified, but included for any nested changes|New listeners are never called during the current notification regardless of nesting|

## Nested notifications:

| |ExpressionHelper|ListenerManager|
|---|---|---|
|Type|Depth first (call stack increases for each nested level)|(same)|
|# of Calls|Listeners * Depth (using incorrect old values)|Collapses nested changes, skipping non-changes|
|Vetoing Possible?|No|Yes|
|Old Value correctness|Only for listeners called before listeners making nested changes|Always|

# Performance

|Listener|ExpressionHelper|ListenerManager|
|---|---|---|
|Addition|Array based, append in empty slot, resize as needed|(same)|
|Removal|Array based, shift array, resize as needed|(same)|
|Addition during notification|Array is copied, removing collected WeakListeners in the process|Appended when notification finishes|
|Removal during notification|As above|Entry is `null`ed (to avoid moving elements in array that is being iterated)|
|Notification completion with changes|-|Null entries (and collected WeakListeners) are removed|
|Notifying Invalidation Listeners|1 ns each|(same)|
|Notifying Change Listeners|1 ns each (*)|2-3 ns each|

(*) a simple for loop is close to optimal, but unfortunately does not provide correct old values

# Memory Use 

Does not include alignment, and assumes a 32-bit VM or one that is using compressed oops.

|Listener|ExpressionHelper|ListenerManager|OldValueCaching ListenerManager|
|---|---|---|---|
|No Listeners|none|none|none|
|Single InvalidationListener|16 bytes overhead|none|none|
|Single ChangeListener|20 bytes overhead|none|16 bytes overhead|
[...]

PR #1448, opened 2024-05-03
8331616: Fix problem in ExpressionHelper that could copy the wrong current value

hjohn:
This PR provides a fix for the linked issue.

The issue was that when an invalidation listener is removed, and the `ExpressionHelper` type changes from `Generic` to `SingleChange` that it would copy the current value of the `Generic` instance before it was updated (this is because invalidation listeners trigger before change listeners and the current value would only be updated **after** invalidation listeners notifications were completed).

The code now will update the current value before sending out any notifications **if** there are change listeners present to head off this problem.

Added a few test cases to verify this.

Note: the PR which replaces `ExpressionHelper` does not have this problem: https://github.com/openjdk/jfx/pull/1081

### DIFF
--- a/modules/javafx.base/src/main/java/com/sun/javafx/binding/ExpressionHelper.java
+++ b/modules/javafx.base/src/main/java/com/sun/javafx/binding/ExpressionHelper.java
@@ -355,6 +355,22 @@ public abstract class ExpressionHelper<T> extends ExpressionHelperBase {
 
             try {
                 locked = true;
+
+                final T oldValue = currentValue;
+
+                if (curChangeSize > 0) {
+
+                    /*
+                     * Because invalidation listeners may get removed during notification, this may
+                     * change the Helper type from Generic to SingleChange. When this transition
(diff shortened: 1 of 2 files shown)


kevinrushforth:
The fix looks good, but I'll formally review it early next week. It could use a second reviewer.

/reviewers 2

mstr2 (approved):
The fix works as expected. However, I hope that we can get #1081 in as a replacement.

kevinrushforth (approved):
Looks good.

hjohn:
@johanvos @kevinrushforth This is quite a fundamental bug introduced in 21 with properties; should we perhaps backport this somewhere?  The current 21 and 22 releases are both bugged, and 21 seems to be receiving updates still (this was fixed in may 2024 and I see that the latest release of 21 is from October).

If we're doing any kind of LTS support on 21, I really think this needs a backport.

johanvos:
I agree, this fixes a rather serious bug that went into 21.

PR #1450, opened 2024-05-06
8331214: Doc: update spec for SpinnerFactory classes

drmarmac:
This PR updates the javadoc for the SpinnerFactory wrap-around behavior introduced in #1431.

---------
### Progress

### Issue
 * [JDK-8331214](https://bugs.openjdk.org/browse/JDK-8331214): Doc: update spec for SpinnerFactory classes (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1450/head:pull/1450` \
`$ git checkout pull/1450`

`$ git checkout pull/1450` \
`$ git pull https://git.openjdk.org/jfx.git pull/1450/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1450`

`$ git pr show -t 1450`

</details>

<a href="https://git.openjdk.org/jfx/pull/1450.diff">https://git.openjdk.org/jfx/pull/1450.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1450#issuecomment-2095446703)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/SpinnerValueFactory.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/SpinnerValueFactory.java
@@ -414,11 +414,11 @@ public abstract class SpinnerValueFactory<T> {
      * A {@link javafx.scene.control.SpinnerValueFactory} implementation designed to iterate through
      * integer values.
      * <p>
-     * If {@link #wrapAroundProperty wrapAround} is {@code true}, the {@code IntegerSpinnerValueFactory} will step from
-     * the minimum value to the maximum value (and vice versa). As a consequence of that, the number
-     * of steps required to wrap around to the same value is {@code N+1}, where {@code N} is the number of steps between
-     * {@link #minProperty min} (inclusive) and {@link #maxProperty max} (inclusive). The new value after a step is
-     * {@code val = (val + amountToStepBy) % (max - min + 1)}.
+     * If {@link SpinnerValueFactory#wrapAroundProperty wrapAround} is {@code true}, the
+     * {@code IntegerSpinnerValueFactory} will step from the minimum value to the maximum value (and vice versa).
+     * As a consequence of that, the number of steps required to wrap around to the same value is {@code N+1}, where
+     * {@code N} is the number of steps between {@link #minProperty min} (inclusive) and {@link #maxProperty max}
+     * (inclusive). The new value after a step is {@code val = (val + amountToStepBy) % (max - min + 1)}.
      * <p>
      * Note that the default {@link #converterProperty() converter} is implemented
      * as an {@link javafx.util.converter.IntegerStringConverter} instance.
@@ -617,10 +617,10 @@ public abstract class SpinnerValueFactory<T> {
      * A {@link javafx.scene.control.SpinnerValueFactory} implementation designed to iterate through
      * double values.
      * <p>
-     * If {@link #wrapAroundProperty wrapAround} is {@code true}, the {@code DoubleSpinnerValueFactory} will step
-     * through from the maximum value to the minimum value seamlessly; that is, any step up from the maximum value
-     * is equal to the same step up from the minimum value (and vice versa). The new value after a step is
-     * {@code val = (val + amountToStepBy) % (max - min)}.
+     * If {@link SpinnerValueFactory#wrapAroundProperty wrapAround} is {@code true}, the
+     * {@code DoubleSpinnerValueFactory} will step through from the maximum value to the minimum value seamlessly; that
+     * is, any step up from the maximum value is equal to the same step up from the minimum value (and vice versa).
+     * The new value after a step is {@code val = (val + amountToStepBy) % (max - min)}.
      * <p>
      * Note that the default {@link #converterProperty() converter} is implemented
      * simply as shown below, which may be adequate in many cases, but it is important
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/SpinnerValueFactory.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/SpinnerValueFactory.java
@@ -115,7 +115,7 @@ public abstract class SpinnerValueFactory<T> {
      *
      * @param steps The number of decrements that should be performed on the value.
      *              If the number is negative, the call is equivalent to calling
-     *              {@link #increment(int)} with the absolute number of steps.
+     *              {@link #increment(int)} with the absolute value of {@code steps}.
      */
     public abstract void decrement(int steps);
 
@@ -126,7 +126,7 @@ public abstract class SpinnerValueFactory<T> {
      *
      * @param steps The number of increments that should be performed on the value.
      *              If the number is negative, the call is equivalent to calling
-     *              {@link #decrement(int)} with the absolute number of steps.
+     *              {@link #decrement(int)} with the absolute number of {@code steps}.
      */
     public abstract void increment(int steps);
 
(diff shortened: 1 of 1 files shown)


kevinrushforth:
@andy-goryachev-oracle Do you also want to take a look?

modules/javafx.controls/src/main/java/javafx/scene/control/SpinnerValueFactory.java
```diff
@@ -409,8 +413,13 @@ private T _getValue(int index) {
     /**
      * A {@link javafx.scene.control.SpinnerValueFactory} implementation designed to iterate through
      * integer values.
-     *
-     * <p>Note that the default {@link #converterProperty() converter} is implemented
+     * <p>
+     * If {@link #wrapAround} is {@code true}, the {@code IntegerSpinnerValueFactory} will step from
```
andy-goryachev-oracle, line 417: The link for wrapAround does not resolve in Eclipse.
I'd suggest to change it to
```
* If {@link SpinnerValueFactory#wrapAround wrapAround}
```

Same comment applies to the min and max in line 420.

modules/javafx.controls/src/main/java/javafx/scene/control/SpinnerValueFactory.java
```diff
@@ -409,8 +413,13 @@ private T _getValue(int index) {
...
+     * the minimum value to the maximum value (and vice versa). As a consequence of that, the number
+     * of steps required to wrap around to the same value is N+1, where N is the number of steps between
+     * {@link #min} (inclusive) and {@link #max} (inclusive).
```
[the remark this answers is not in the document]
andy-goryachev-oracle: a good idea

modules/javafx.controls/src/main/java/javafx/scene/control/SpinnerValueFactory.java
```diff
@@ -606,8 +615,12 @@ public final IntegerProperty amountToStepByProperty() {
     /**
      * A {@link javafx.scene.control.SpinnerValueFactory} implementation designed to iterate through
      * double values.
-     *
-     * <p>Note that the default {@link #converterProperty() converter} is implemented
+     * <p>
+     * If {@link #wrapAround} is {@code true}, the {@code DoubleSpinnerValueFactory} will step through
```
andy-goryachev-oracle, line 619: wrapAround link does not resolve in Eclipse, see my earlier comment

modules/javafx.controls/src/main/java/javafx/scene/control/SpinnerValueFactory.java
kevinrushforth: I don't think this will help. This is the same issue I mentioned in PR #1389 -- you need to link to the property _method_, so I think changing to the following will do it:

```
* If {@link #wrapAroundProperty wrapAround}
```

(it shouldn't be necessary to qualify it with the enclosing class, but it wouldn't hurt to do so)

andy-goryachev-oracle: `{@link #wrapAroundProperty wrapAround}` does not resolve **in Eclipse** (it does in javadoc output).

This might be Eclipse limitation, but it would be awfully nice to have it resolved in Eclipse.

modules/javafx.controls/src/main/java/javafx/scene/control/SpinnerValueFactory.java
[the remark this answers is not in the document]
drmarmac: agreed, I used "amountToStepBy" instead of "delta" here.

modules/javafx.controls/src/main/java/javafx/scene/control/SpinnerValueFactory.java
drmarmac: I also fixed the min/max links to refer to minProperty/maxProperty. IntelliJ IDEA has the same problem as eclipse. I guess this has to do with the "javafx" option to the javadoc command (build.gradle:4286 `options.addBooleanOption("javafx").setValue(true);`) that seems to allow to use the private fields' docs for the properties, by naming convention? Maybe we need to file issues in Eclipse/IntelliJ for this so they add the option to their javadoc renderer.

modules/javafx.controls/src/main/java/javafx/scene/control/SpinnerValueFactory.java
```diff
@@ -409,8 +413,14 @@ private T _getValue(int index) {
     /**
      * A {@link javafx.scene.control.SpinnerValueFactory} implementation designed to iterate through
      * integer values.
-     *
-     * <p>Note that the default {@link #converterProperty() converter} is implemented
+     * <p>
+     * If {@link #wrapAroundProperty wrapAround} is {@code true}, the {@code IntegerSpinnerValueFactory} will step from
```
andy-goryachev-oracle, line 417: wrapAround still does not work in Eclipse, but this does, while providing the same visuals:

```
{@link SpinnerValueFactory#wrapAroundProperty wrapAround}
```

does it work in intellij?

modules/javafx.controls/src/main/java/javafx/scene/control/SpinnerValueFactory.java
```diff
@@ -606,8 +616,13 @@ public final IntegerProperty amountToStepByProperty() {
     /**
      * A {@link javafx.scene.control.SpinnerValueFactory} implementation designed to iterate through
      * double values.
-     *
-     * <p>Note that the default {@link #converterProperty() converter} is implemented
+     * <p>
+     * If {@link #wrapAroundProperty wrapAround} is {@code true}, the {@code DoubleSpinnerValueFactory} will step
```
andy-goryachev-oracle, line 620: same here

modules/javafx.controls/src/main/java/javafx/scene/control/SpinnerValueFactory.java
kevinrushforth: There are two issues here. The first is that you cannot just link to a `xxx` property by name -- you have to link to the `xxxProperty` method. The former doesn't work in the generated javadocs either (so it's not just an IDE problem). The second issue seems specific to Eclipse, in that if you don't prefix the method name with the name of the enclosing (outer) class, Eclipse doesn't generate the link. That seems like an Eclipse bug to me, but as Andy points out, it's a trivial thing to add the class name (and won't affect the output).

modules/javafx.controls/src/main/java/javafx/scene/control/SpinnerValueFactory.java
kevinrushforth: That one should work everywhere, since it is equivalent. Worth double-checking, though.

drmarmac: Updated. I checked this variant with IntelliJ: While I can follow the link, it doesn't show the wrapAround field's javadoc, just the wrapAroundProperty's javadoc which is empty.

andy-goryachev-oracle (approved):
Looks good in Eclipse and in the browser.  Thank you for making the doc clearer and also for fixing the issue!

kevinrushforth (approved):
Looks good to me, too.

modules/javafx.controls/src/main/java/javafx/scene/control/SpinnerValueFactory.java
kevinrushforth: I guess that's because IntelliJ doesn't process the javafx properties.

andy-goryachev-oracle: that's why _some say_ eclipse is better:

Screenshot 2024-05-07 at 08 37 11
Screenshot 2024-05-07 at 08 37 19

drmarmac: I upvoted a matching [issue in the IntelliJ bugtracker](https://youtrack.jetbrains.com/issue/IDEA-273592/JavaFX-incorrectly-navigates-to-property-field-ignores-JavaDoc). As this affects quite a few properties in javafx it would be good to see this fixed at some point.

PR #1451, opened 2024-05-06
8322619: Parts of SG no longer update during rendering - overlapping - culling - dirty

eduardsdv:
This is an alternative solution to the PR: https://github.com/openjdk/jfx/pull/1310.

This solution is based on the invariant that if a node is marked as dirty, all ancestors must also be marked as dirty and that if an ancestor is marked as clean, all descendants must also be marked as clean. 
Therefore I removed the ``clearDirtyTree()`` method and put its content to the ``clearDirty()`` method.

Furthermore, since dirty flag is only used when rendering by ``ViewPainter``, it should also be deleted by ``ViewPainter`` only. 
This guarantees:
1. that all dirty flags are removed after rendering, and 
2. that no dirty flags are removed when a node is rendered, e.g. by creating a snapshot or printing.
Therefore I removed all calls of the methods ``clearDirty()`` and ``clearDirtyTree()`` from all other classes except the ``ViewerPainter``.

The new version of the ``clearDirty()`` method together with calling it from the ``ViewerPainter`` needs to visit far fewer nodes compared to the version prior this PR.

The supplied test checks that the nodes are updated even if they are partially covered, which led to the error in the version before the PR. The test can be started with:
``gradlew -PFULL_TEST=true -PUSE_ROBOT=true :systemTests:test --tests NGNodeDirtyFlagTest``

---------
### Progress

### Issues
 * [JDK-8322619](https://bugs.openjdk.org/browse/JDK-8322619): Parts of SG no longer update during rendering - overlapping - culling - dirty (**Bug** - P4)
 * [JDK-8332040](https://bugs.openjdk.org/browse/JDK-8332040): Parts of SG no longer update during rendering - overlapping - culling - dirty (**CSR**) (Withdrawn)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1451/head:pull/1451` \
`$ git checkout pull/1451`

`$ git checkout pull/1451` \
`$ git pull https://git.openjdk.org/jfx.git pull/1451/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1451`

`$ git pr show -t 1451`

</details>

<a href="https://git.openjdk.org/jfx/pull/1451.diff">https://git.openjdk.org/jfx/pull/1451.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1451#issuecomment-2096141074)

### DIFF
--- a/tests/system/src/test/java/test/robot/com/sun/prism/NGNodeDirtyFlagTest.java
+++ b/tests/system/src/test/java/test/robot/com/sun/prism/NGNodeDirtyFlagTest.java
@@ -25,10 +25,6 @@
 
 package test.robot.com.sun.prism;
 
-import com.sun.javafx.tk.RenderJob;
-import com.sun.javafx.tk.Toolkit;
-import java.util.concurrent.CountDownLatch;
-import javafx.application.Platform;
 import javafx.beans.property.ObjectProperty;
 import javafx.beans.property.SimpleObjectProperty;
 import javafx.geometry.Bounds;
@@ -88,7 +84,6 @@ public class NGNodeDirtyFlagTest extends VisualTestBase {
             Util.runAndWait(() -> circleColor.set(Color.LIGHTGREEN));
             Util.waitForIdle(root.getScene());
 
-            waitForRenderer();
             checkLineColor(root, lineColor.get());
 
             Util.runAndWait(() -> lineColor.set(Color.DARKGREEN));
@@ -96,15 +91,8 @@ public class NGNodeDirtyFlagTest extends VisualTestBase {
             Util.runAndWait(() -> circleColor.set(Color.DARKGREEN));
             Util.waitForIdle(root.getScene());
 
-            waitForRenderer();
             checkLineColor(root, lineColor.get());
         }
-    }
-
-    private void waitForRenderer() {
-        CountDownLatch latch = new CountDownLatch(1);
-        Platform.runLater(() -> Toolkit.getToolkit().addRenderJob(new RenderJob(latch::countDown)));
-        Util.await(latch);
     }
 
     private void checkLineColor(StackPane root, Color expected) {


kevinrushforth:
Reviewers: @lukostyra @arapte 

This is a more intrusive fix than #1310 so we would only want to go this route if we can show that it is a correct fix, introduces no regressions (in either correctness or performance), and makes it easier to maintain in the future.

/reviewers 2

hjohn:
Even though this fix is more intrusive, it does seem to address the root cause.  Instead of cleaning dirty bits in many places where it is easy to miss one, it always occurs only exactly where needed (after painting completes) using a method that can't accidentally miss parts.

At a minimum I think the tests that are part of this PR should be included in FX whichever fix ends up being integrated.

A quick glance doesn't turn up anything unexpected.  I'm only wondering if the code in `paintImpl` should always clear the dirty bits even if an exception occurs during painting, to harden it against potential bugs and not end up trying to repaint again and again likely getting the same exception again and again.

eduardsdv:
> At a minimum I think the tests that are part of this PR should be included in FX whichever fix ends up being integrated.

The test has already been added to both PRs.

> I'm only wondering if the code in ``paintImpl`` should always clear the dirty bits even if an exception occurs during painting, to harden it against potential bugs and not end up trying to repaint again and again likely getting the same exception again and again.

Hm, that can indeed happen. On the other hand, if the dirty flag is reset even in the case of an exception, parts of the UI may not be updated for a long time until a node in that area receives a change. The question is which of these two options is the least harmful.

I'm fine with each of them. What do the others think?

kevinrushforth:
> > I'm only wondering if the code in `paintImpl` should always clear the dirty bits even if an exception occurs during painting, to harden it against potential bugs and not end up trying to repaint again and again likely getting the same exception again and again.
> 
> Hm, that can indeed happen. On the other hand, if the dirty flag is reset even in the case of an exception, parts of the UI may not be updated for a long time until a node in that area receives a change. The question is which of these two options is the least harmful.
> 
> I'm fine with each of them. What do the others think?

I'd probably lean towards addressing this in a follow-up issue, if there turns out to be a need.

@arapte can you look at this when you review it?

eduardsdv:
keep alive

eduardsdv:
@arapte Please review.

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
```diff
@@ -0,0 +1,188 @@
...
+            myApp = this;
+        }
+
+        @Override
+        public void start(Stage primaryStage) throws Exception {
+            root = new StackPane();
+            primaryStage.setScene(new Scene(root, 500, 400));
+            primaryStage.show();
+
+            launchLatch.countDown();
```
arapte, line 68: This countDown() should be called after stage is shown to ensure the stage shown before proceeding.
`primaryStage.setOnShown(e -> Platform.runLater(launchLatch::countDown));`

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
```diff
@@ -0,0 +1,188 @@
...
+        }
+    }
+
+    @BeforeClass
+    public static void setupOnce() {
+        Util.launch(launchLatch, MyApp.class);
+        assertEquals(0, launchLatch.getCount());
```
arapte, line 75: Util.launch() asserts if launchLatch.countDown() does not occur in 15 seconds, so this line can be removed.

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
```diff
@@ -0,0 +1,188 @@
...
+
+    private Pane createSideArea() {
+        VBox result = new VBox();
+        result.setPrefSize(150, 9900);
+        result.setMaxWidth(200);
+        result.setStyle("-fx-background-color: lightblue;");
+        result.getChildren().add(new Label("SideArea"));
+        return result;
+    }
+}
```
arapte, line 188: Please add an empty line at the end.

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
```diff
@@ -0,0 +1,188 @@
...
+        } catch (ExecutionException e) {
+            if (e.getCause() instanceof AssertionError error) {
+                throw error;
+            } else {
+                throw new RuntimeException(e);
+            }
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
```
arapte, line 149: Please use the `Util.runAndWait()` instead of adding this method. So this method can be removed and all calls can be changed to Util.runAndWait().
Please do remove any imports that become unused afterwards.

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
```diff
@@ -0,0 +1,188 @@
...
+
+    @Test
+    public void testNGNodesNotDirty() throws InterruptedException {
+        ObjectProperty<Color> lineColor = new SimpleObjectProperty<>(Color.DARKGREEN);
+        ObjectProperty<Color> circleColor = new SimpleObjectProperty<>(Color.DARKGREEN);
+
+        StackPane root = myApp.root;
+
+        runAndWait(() -> {
```
arapte, line 90: All such calls can be changed to use `Util.runAndWait()`

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
```diff
@@ -0,0 +1,188 @@
...
+            contents.getChildren().add(contentElement("L", lineColor, circleColor));
+            contents.getChildren().add(contentElement("R", lineColor, circleColor));
+            root.getChildren().add(contents);
+
+            Pane sideArea = createSideArea();
+            StackPane.setAlignment(sideArea, Pos.CENTER_RIGHT);
+            root.getChildren().add(sideArea);
+        });
+
+        Thread.sleep(500);
```
arapte, line 103: Could use `Util.waitForIdle(Scene)` instead of sleep. It would reduce the test time and should work. We can consider a sleep if `Util.waitForIdle(Scene)` does not work. Similarly please see if other sleep calls can be removed.

arapte (changes requested):
Providing a few comments on test. 
I might need a day or two more for testing/reviewing the fix.

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
eduardsdv: Done

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
eduardsdv: Done

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
eduardsdv: Done

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
eduardsdv: Done

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
eduardsdv: Done

eduardsdv:
@kevinrushforth The build fails on Linux. Can you please take a look?

eduardsdv:
Ok. I solved the build by merging from the master branch.

arapte:
I have tested the patch with several samples. Fix looks good.
Though the test fails on my Mac machine with below error.
Please verify..

```
NGNodeDirtyFlagTest > testNGNodesNotDirty FAILED
    org.junit.ComparisonFailure: A node was not rendered properly. Wrong color found expected:<[LIGHTGREEN]> but was:<[0x81ee7eff]>
        at org.junit.Assert.assertEquals(Assert.java:117)
        at test.com.sun.prism.impl.NGNodeDirtyFlagTest.checkColor(NGNodeDirtyFlagTest.java:135)
        at test.com.sun.prism.impl.NGNodeDirtyFlagTest.lambda$checkLineColor$5(NGNodeDirtyFlagTest.java:126)

1 test completed, 1 failed

> Task :systemTests:test FAILED
```

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
```diff
@@ -0,0 +1,190 @@
...
+        @Override
+        public void init() {
+            myApp = this;
+        }
+
+        @Override
+        public void start(Stage primaryStage) throws Exception {
+            root = new StackPane();
+            primaryStage.setScene(new Scene(root, 500, 400));
+
```
arapte, line 68: Please add `primaryStage.setAlwaysOnTop(true);` call here, without it the test can fail fail intermittently if the test window gets behind any other window.

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
```diff
@@ -0,0 +1,190 @@
...
+            checkLineColor(root, lineColor.get());
+
+            Util.runAndWait(() -> lineColor.set(Color.DARKGREEN));
+            Util.waitForIdle(root.getScene());
+            Util.runAndWait(() -> circleColor.set(Color.DARKGREEN));
+            Util.waitForIdle(root.getScene());
+
+            checkLineColor(root, lineColor.get());
+        }
+
```
arapte, line 121: minor: can remove the empty line.

arapte (commented):
Few minor comments on the tests, apart from the failure that I see on Mac.
Otherwise I could not find any issues.

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
```diff
@@ -0,0 +1,190 @@
+package test.com.sun.prism.impl;
```
kevinrushforth, line 1: This needs a standard Copyright header.

Also, all tests that use Robot `must` be under the `test.robot.**` package hierarchy. Please move the test  and change the package.

kevinrushforth (dismissed):

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
```diff
@@ -0,0 +1,190 @@
...
+            checkColor(root.lookup("#Line-L"), expected);
+            checkColor(root.lookup("#Line-R"), expected);
+        });
+    }
+
+    private void checkColor(Node node, Color expected) {
+        Robot robot = new Robot();
+        Bounds screenBounds = node.localToScreen(node.getBoundsInLocal());
+        WritableImage image = robot.getScreenCapture(null, screenBounds.getMinX(), screenBounds.getMinY(), 100, 100);
+        Assert.assertEquals("A node was not rendered properly. Wrong color found", name(expected), name(image.getPixelReader().getColor(1, 1)));
```
kevinrushforth, line 135: This method of color comparison seems convoluted and error prone. I recommend using one of the existing color comparison methods with toloerance. See VisualTestBase, for example.

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
kevinrushforth: Also, we generally recommend an undecorated stage for any tests that do a screen capture.

kevinrushforth (commented):
A couple more test comments.

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
```diff
@@ -0,0 +1,190 @@
...
+
+public class NGNodeDirtyFlagTest {
+
+    // Used to launch the application before running any test
+    private static final CountDownLatch launchLatch = new CountDownLatch(1);
+
+    // Singleton Application instance
+    static MyApp myApp;
+
+    private CountDownLatch latch = new CountDownLatch(1);
```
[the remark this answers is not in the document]
eduardsdv: Done

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
eduardsdv: Done

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
eduardsdv: I moved it to the ``test.robot.com.sun.prism`` package and added the copyright header.

tests/system/src/test/java/test/com/sun/prism/impl/NGNodeDirtyFlagTest.java
eduardsdv: The test now uses ``VisualTestBase.assertColorEquals(Color expected, Color actual, double delta)`` to compare colors.

eduardsdv:
@arapte: The failed test on the Mac is probably caused by the lack of synchronization between the JavaFX and QuantumRenderer threads. I added the ``waitForRenderer()`` method to wait for the stage to render before checking the colors..

arapte:
The newer version still fails on my Mac machine.

Attaching modified test that executes as expected on my Windows and Mac machines.
The main change that makes it work on Mac is the change in `checkColor()` method. 
There are some additional minor changes, and I recommend to keep those.
Please do test on your machines.

[NGNodeDirtyFlagTest.java.zip](https://github.com/user-attachments/files/16090444/NGNodeDirtyFlagTest.java.zip)

eduardsdv:
Yes, the test with your suggestion also works under Windows. Thank you, @arapte.

arapte:
> Yes, the test with your suggestion also works under Windows. Thank you, @arapte.

Thanks for verifying. In that case, let's remove the `waitForRenderer()` method that was added to solve the test failure on Mac but is not required anymore.

eduardsdv:
>> Yes, the test with your suggestion also works under Windows. Thank you, @arapte.
>
>Thanks for verifying. In that case, let's remove the waitForRenderer() method that was added to solve the test failure on Mac but is not required anymore.

I removed the method.

arapte (approved):

kevinrushforth:
@lukostyra Can you be the second reviewer on this?

lukostyra (approved):
Changes look good. Run all tests on all platforms I have available for me (Win 11, macOS 14.5, Ubuntu 22.04) and noticed no regressions or issues. Ensemble also works fine.

PR #1452, opened 2024-05-07
8279140: ComboBox can lose selected value on item change via setAll

karthikpandelu:
The `ComboBox` value was not set to previously selected value in the item list change listener when `setAll` method is used to change the items. Fixed the issue by restoring the selection in this case.

Added a unit test to validate the fix.
[JDK-8279139](https://bugs.openjdk.org/browse/JDK-8279139) is also fixed with this change and added a unit test for the same.

---------
### Progress

### Issue
 * [JDK-8279140](https://bugs.openjdk.org/browse/JDK-8279140): ComboBox can lose selected value on item change via setAll (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1452/head:pull/1452` \
`$ git checkout pull/1452`

`$ git checkout pull/1452` \
`$ git pull https://git.openjdk.org/jfx.git pull/1452/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1452`

`$ git pr show -t 1452`

</details>

<a href="https://git.openjdk.org/jfx/pull/1452.diff">https://git.openjdk.org/jfx/pull/1452.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1452#issuecomment-2097951529)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/ComboBox.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/ComboBox.java
@@ -618,17 +618,11 @@ public class ComboBox<T> extends ComboBoxBase<T> {
                     T selectedItem = getSelectedItem();
                     for (int i = 0; i < comboBox.getItems().size(); i++) {
                         if (selectedItem.equals(comboBox.getItems().get(i))) {
-                            comboBox.setValue(null);
-                            setSelectedItem(null);
-
-                            //Restore the previous selection
-                            comboBox.setValue(selectedItem);
-                            setSelectedIndex(i);
+                            clearAndSelect(i);
                             break;
                         }
                     }
                 }
-
                 comboBox.previousItemCount = getItemCount();
             }
         };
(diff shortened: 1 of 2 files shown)


karthikpandelu:
I missed to validate the test case present in [JDK-8279139](https://bugs.openjdk.org/browse/JDK-8279139) and looks like it is not yet fixed. Hence I will change this PR to draft for now

andy-goryachev-oracle:
It looks like we have quite a few bugs related to different aspects of ComboBox.  This change is definitely an improvement, as it fixes the scenario described in JDK-8279140.

This PR changes parts of the code that fixed JDK-8087838.  While I don't see a regression in respect to JDK-8087838, I do see a different behavior with and without this fix in JDK-8090221, using the test app there (to be specific, with the fix in this PR, the tester in JDK-8090221 continues to display the selected item in the combo box even though the output shows `Selected Item After: null`

Do you think that JDK-8090221 scenario is a different case and should be addressed separately?

karthikpandelu:
Yes there are few bugs related to ComboBox.
I think we can fix JDK-8090221 separately. The issue JDK-8279139, which is closed as duplicate of this issue is fixed, but changing the test slightly shows another issue which is mentioned in JDK-8263942. Since we have different scenarios like these, we can fix these issue with these PRs one by one. So I believe other issues can be fixed separately.

andy-goryachev-oracle (approved):
looks good.  thank you for adding the tests!

karthikpandelu:
Thanks for reviewing @andy-goryachev-oracle

PR #1458, opened 2024-05-16
8324327: ColorPicker shows a white rectangle on clicking on picker

andy-goryachev-oracle:
Root cause: premature 'return' from initialization code leaves the hover square visible, even when it should be hidden, as the case is when the color value is not present in the palette.

- added unit tests

---------
### Progress

### Issue
 * [JDK-8324327](https://bugs.openjdk.org/browse/JDK-8324327): ColorPicker shows a white rectangle on clicking on picker (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1458/head:pull/1458` \
`$ git checkout pull/1458`

`$ git checkout pull/1458` \
`$ git pull https://git.openjdk.org/jfx.git pull/1458/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1458`

`$ git pr show -t 1458`

</details>

<a href="https://git.openjdk.org/jfx/pull/1458.diff">https://git.openjdk.org/jfx/pull/1458.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1458#issuecomment-2115943666)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ColorPalette.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ColorPalette.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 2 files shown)


arapte (approved):
Fix looks good, I think a single reviewer is sufficient for this. But please do wait for a while, may be for a day before integrating.

I do have an observation to share:
Another way to look at the root cause of the issue is that the `hoverSquare`'s  visible property is not initialized in the `ColorPalette` constructor.  
The `hoverSquare` should **not** be visible by default, so adding this `hoverSquare.setVisible(false);` to `ColorPalette` constructor also solves the problem.
Compared to PR change, this would avoid setting the visible property of focusSquare to false repeatedly in scenario when user keeps selecting custom color repeatedly.

Anyway the fix is good, approving.
Shall re-approve if you make any change.

andy-goryachev-oracle:
> `hoverSquare`'s visible property is not initialized in the `ColorPalette` constructor.

thank you for reviewing, @arapte !

effectively, yes, and that's the issue I have fixed, by explicitly setting visibility in `setFocusedSquare()` (which is called from the constructor).  I think this is a better fix because it fixes the logic in this method which may be applicable in other cases.  Setting `false` to a boolean property which is already `false` has very little overhead.

PR #1461, opened 2024-05-23
8332732: Clean up non-standard use of /// comments in JavaFX

kevinrushforth:
[JDK-8298405](https://bugs.openjdk.org/browse/JDK-8298405) added markdown support to javadoc, using `///` to indicate markdown documentation comments. As a result, building JavaFX docs with JDK 23 produces new warnings, which causes the build to fail (since we treat warnings as errors).

This PR was generated by running a program to automate the modification of each line that contains three or more slash chars in a row, `///` (except those in String literals). The replacement algorithm is as follows:

1. Replace each occurrence of exactly 3 slash chars `///` with 2 slash chars `//`
2. Replace each occurrence of 4 or more slash chars with an equal number of `-` chars, leaving the first two `//` chars of the first run of `////` as is.

This is similar to the proposed fix in java.base for [JDK-8331879](https://bugs.openjdk.org/browse/JDK-8331879) / PR openjdk/jdk#19130 (except for how I propose to handle the case of exactly three slashes).

As an alternative approach, I could fix just the 5 javadoc comments causing the error, and leave the rest alone, but this will prevent problems from cropping up in the future, and is in keeping with changes going into the JDK source base.


modules/javafx.media/src/main/java/com/sun/media/jfxmedia/locator/Locator.java
```diff
@@ -349,7 +349,7 @@ public void init() throws URISyntaxException, IOException, FileNotFoundException
             if (firstSlash != -1 && uriString.charAt(firstSlash + 1) != '/') {
                 // Only one '/' after the ':'.
                 if (protocol.equals("file")) {
-                    // Map file:/somepath to file:///somepath
+                    // Map "file:/somepath" to "file:///somepath"
```
kevinrushforth, line 352: I did this one by hand, adding the `"` chars so that the automated program wouldn't touch the `///` in `file:///`.

andy-goryachev-oracle (approved):
went through all 50 files, all is good.  ship it!

andy-goryachev-oracle:
Are these all the places where javadoc fails, or more files are coming separately?
For example, see AccordionBehavior:151

kevinrushforth:
> Are these all the places where javadoc fails, or more files are coming separately?

There are only 5 places where javadoc fails with the latest JDK 23. By eliminating all cases where we have more than three slashes, this PR fixes those 5 as well as avoids a pattern that could cause problems in the future.

> For example, see AccordionBehavior:151, ChoiceBoxSkin:327

Not sure what you mean here. Those lines are modified by this PR such that they no longer have more than 2 slashes.

andy-goryachev-oracle:
> Those lines are modified by this PR

You are right!  I had cached java files in build/modulal-sdk/modules_src which Eclipse search picked up.

Apparently, there is a way to mark the folder as "derived" so it will be ignored by the Search, but that hits an 18-year old Eclipse bug...
https://bugs.eclipse.org/bugs/show_bug.cgi?id=154089

kevinrushforth:
Thanks for checking. After this PR there should be no instances of `///` outside of String literals any `.java` file in the repo.

PR #1465, opened 2024-05-28
8087444: CornerRadii with different horizontal and vertical values treated as uniform

mstr2:
This PR fixes the incorrect computation of the `CornerRadii.uniform` flag for the 16-argument constructor.

I've also added tests for all other constructors.

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/layout/CornerRadii.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/layout/CornerRadii.java
@@ -312,18 +312,20 @@ public class CornerRadii {
                 topRightVerticalRadiusAsPercent || topRightHorizontalRadiusAsPercent ||
                 bottomRightHorizontalRadiusAsPercent || bottomRightVerticalRadiusAsPercent ||
                 bottomLeftVerticalRadiusAsPercent || bottomLeftHorizontalRadiusAsPercent;
-        uniform = topLeftHorizontalRadius == topRightHorizontalRadius &&
-                topLeftVerticalRadius == topRightVerticalRadius &&
+        uniform = topLeftHorizontalRadius == topLeftVerticalRadius &&
+                topLeftHorizontalRadius == topRightVerticalRadius &&
+                topLeftHorizontalRadius == topRightHorizontalRadius &&
                 topLeftHorizontalRadius == bottomRightHorizontalRadius &&
-                topLeftVerticalRadius == bottomRightVerticalRadius &&
+                topLeftHorizontalRadius == bottomRightVerticalRadius &&
+                topLeftHorizontalRadius == bottomLeftVerticalRadius &&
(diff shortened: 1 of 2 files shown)


modules/javafx.graphics/src/test/java/test/javafx/scene/layout/CornerRadiiTest.java
```diff
@@ -53,8 +56,71 @@ public class CornerRadiiTest {
...
 
-    @Test(expected = IllegalArgumentException.class)
+    @Test
     public void negativeRadiusNotAllowed_singleConstructor() {
-        new CornerRadii(-1);
+        assertThrows(IllegalArgumentException.class, () -> new CornerRadii(-1));
+    }
+
+    @Nested
+    class IsUniformTests {
```
andy-goryachev-oracle, line 65: just curious - what was the rationale for `@Nested` here, as the number of the test cases is fairly small.

mstr2: I will add some more tests with [JDK-8332895](https://bugs.openjdk.org/browse/JDK-8332895) (adding interpolation support), and the `@Nested` nicely groups all tests for a given functionality.

modules/javafx.graphics/src/main/java/javafx/scene/layout/CornerRadii.java
```diff
@@ -312,18 +312,20 @@ public CornerRadii(
                 topRightVerticalRadiusAsPercent || topRightHorizontalRadiusAsPercent ||
                 bottomRightHorizontalRadiusAsPercent || bottomRightVerticalRadiusAsPercent ||
                 bottomLeftVerticalRadiusAsPercent || bottomLeftHorizontalRadiusAsPercent;
-        uniform = topLeftHorizontalRadius == topRightHorizontalRadius &&
-                topLeftVerticalRadius == topRightVerticalRadius &&
+        uniform = topLeftHorizontalRadius == topLeftVerticalRadius &&
```
hjohn, line 315: I found the diff a bit troubling to read, but from what I gather, all the radii need to equal to qualify for uniformity, and what you did here is add a `topLeftHorizontalRadius == topLeftVerticalRadius` (and percent version) to this check.

You also already improved the readability by using `topLeftHorizontalRadius` everywhere.  It might be even better to extract this one to a new variable, so it is even more clear that you're comparing all the other radii with the same value.

hjohn (approved):
LGTM

modules/javafx.graphics/src/main/java/javafx/scene/layout/CornerRadii.java
mstr2: I tried that, but in my eyes it didn't clearly improve readability beyond the current version.

PR #1467, opened 2024-05-30
8330304: MenuBar: Invisible Menu works incorrectly with keyboard arrows

andy-goryachev-oracle:
The root cause is that the skin used two fields to store one entity (`focusedMenu` and `focusedMenuIndex`), causing mismatch when invisible menu(s) are present.

The fix involves using a single index variable.

Also wanted to note that in theory, `openMenu` and `openMenuButton` can also be replaced with a single boolean, but I decided not to change these in order to keep the amount of diffs to a minimum.

---------
### Progress

### Issue
 * [JDK-8330304](https://bugs.openjdk.org/browse/JDK-8330304): MenuBar: Invisible Menu works incorrectly with keyboard arrows (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1467/head:pull/1467` \
`$ git checkout pull/1467`

`$ git checkout pull/1467` \
`$ git pull https://git.openjdk.org/jfx.git pull/1467/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1467`

`$ git pr show -t 1467`

</details>

<a href="https://git.openjdk.org/jfx/pull/1467.diff">https://git.openjdk.org/jfx/pull/1467.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1467#issuecomment-2140989050)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/MenuBarSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/MenuBarSkin.java
@@ -121,16 +121,14 @@ public class MenuBarSkin extends SkinBase<MenuBar> {
      *                                                                         *
      **************************************************************************/
 
+    /** contains MenuBarButton's children only */
     private final HBox container;
+    /** index of *focused* MenuBarButton */
+    private int focusedMenuIndex = -1;
 
     // represents the currently _open_ menu
     private Menu openMenu;
     private MenuBarButton openMenuButton;
-
-    // represents the currently _focused_ menu. If openMenu is non-null, this should equal
-    // openMenu. If openMenu is null, this can be any menu in the menu bar.
-    private Menu focusedMenu;
-    private int focusedMenuIndex = -1;
 
     private static WeakHashMap<Stage, Reference<MenuBarSkin>> systemMenuMap;
     private static List<MenuBase> wrappedDefaultMenus = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


modules/javafx.controls/src/main/java/javafx/scene/control/skin/MenuBarSkin.java
```diff
@@ -733,8 +730,8 @@ public void dispose() {
      *                                                                         *
      **************************************************************************/
 
-    // For testing purpose only.
-    MenuButton getNodeForMenu(int i) {
+    // package protected for testing purposes
+    MenuBarButton menuBarButtonAt(int i) {
         if (i < container.getChildren().size()) {
```
mstr2, line 735: This method throws `IndexOutOfBoundsException` if the index is negative, but returns `null` if the index is larger than the collection size. None of the callers account for this, so we'd just get `NullPointerException` at the call sites.

Since you touched this method, I suggest replacing the implementation with
```java
MenuBarButton menuBarButtonAt(int i) {
    return (MenuBarButton)container.getChildren().get(i);
}
```

andy-goryachev-oracle: yes, the suggested change is safe: the product code always guards against invalid indexes, while the test code always expects a non-null return value.

the only time where it might backfire is if a (new) test sets up an empty menu bar.  but we don't have such a test.

mstr2 (approved):
LGTM

andy-goryachev-oracle:
@aghaisas or @arapte could you please take a look at this PR?

kevinrushforth (approved):
The fix looks good. I also verified that it fixes the problem, and that the new test passes with the fix and fails without the fix.

PR #870, opened 2022-08-16
8311895: CSS Transitions

mstr2:
Implementation of [CSS Transitions](https://www.w3.org/TR/css-transitions-1/).

### Future enhancements
CSS transition support for backgrounds and borders: #1471 

### Limitations
This implementation supports both shorthand and longhand notations for the `transition` property. However, due to limitations of JavaFX CSS, mixing both notations doesn't work:
```css
.button {
    transition: -fx-background-color 1s;
    transition-easing-function: linear;
}
```
This issue should be addressed in a follow-up enhancement.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (3 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 2 [Authors](https://openjdk.org/bylaws#author))
- [x] Change requires CSR request [JDK-8313383](https://bugs.openjdk.org/browse/JDK-8313383) to be approved

### Issues
 * [JDK-8311895](https://bugs.openjdk.org/browse/JDK-8311895): CSS Transitions (**Enhancement** - P4)
 * [JDK-8313383](https://bugs.openjdk.org/browse/JDK-8313383): CSS Transitions (**CSR**)

### Reviewers
 * [Markus Mack](https://openjdk.org/census#mmack) (@drmarmac - Author)
 * [Nir Lisker](https://openjdk.org/census#nlisker) (@nlisker - **Reviewer**)
 * [Kevin Rushforth](https://openjdk.org/census#kcr) (@kevinrushforth - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/870/head:pull/870` \
`$ git checkout pull/870`

Update a local copy of the PR: \
`$ git checkout pull/870` \
`$ git pull https://git.openjdk.org/jfx.git pull/870/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 870`

View PR using the GUI difftool: \
`$ git pr show -t 870`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jfx/pull/870.diff">https://git.openjdk.org/jfx/pull/870.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/870#issuecomment-1650938857)

PR #1471, opened 2024-06-02
8332895: Support interpolation for backgrounds and borders

mstr2:
This PR completes the CSS Transitions story (see #870) by adding interpolation support for backgrounds and borders, making them targetable by transitions.

`Background` and `Border` objects are deeply immutable, but not interpolatable. Consider the following `Background`, which describes the background of a `Region`:

```
Background {
    fills = [
        BackgroundFill {
            fill = Color.RED
        }
    ]
}
```

Since backgrounds are deeply immutable, changing the region's background to another color requires the construction of a new `Background`, containing a new `BackgroundFill`, containing the new `Color`.

Animating the background color using a CSS transition therefore requires the entire Background object graph to be interpolatable in order to generate intermediate backgrounds.

More specifically, the following types will now implement `Interpolatable`.

- `Insets`
- `Background`
- `BackgroundFill`
- `BackgroundImage`
- `BackgroundPosition`
- `BackgroundSize`
- `Border`
- `BorderImage`
- `BorderStroke`
- `BorderWidths`
- `CornerRadii`
- `ImagePattern`
- `LinearGradient`
- `RadialGradient`
- `Stop`

## Interpolation of composite objects

As of now, only `Color`, `Point2D`, and `Point3D` are interpolatable. Each of these classes is an aggregate of `double` values, which are combined using linear interpolation. However, many of the new interpolatable classes comprise of not only `double` values, but a whole range of other types. This requires us to more precisely define what we mean by "interpolation".

Mirroring the CSS specification, the `Interpolatable` interface defines several types of component interpolation:

| Interpolation type | Description |
|---|---|
| default | Component types that implement `Interpolatable` are interpolated by calling the `interpolate(Object, double)}` method. |
| linear | Two components are combined by linear interpolation such that `t = 0` produces the start value, and `t = 1` produces the end value. This interpolation type is usually applicable for numeric components. |
| discrete | If two components cannot be meaningfully combined, the intermediate component value is equal to the start value for `t < 0.5` and equal to the end value for `t >= 0.5`. |
| pairwise | Two lists are combined by pairwise interpolation. If the start list has fewer elements than the target list, the missing elements are copied from the target list. If the start list has more elements than the target list, the excess elements are discarded. |
| (see prose) | Some component types are interpolated in specific ways not covered here. Refer to their respective documentation for more information. |

Every component of an interpolatable class specifies its interpolation type with the new `@interpolationType` javadoc tag. For example, this is the specification of the `CornerRadii.topLeftHorizontalRadius` component:
```java
    /**
     * The length of the horizontal radii of the top-left corner.
     *
     * @return the length of the horizontal radii of the top-left corner
     * @interpolationType <a href="../../animation/Interpolatable.html#linear">linear</a> if both values are
     *                    absolute or both values are {@link #isTopLeftHorizontalRadiusAsPercentage() percentages},
     *                    <a href="../../animation/Interpolatable.html#discrete">discrete</a> otherwise
     */
    public double getTopLeftHorizontalRadius()
```

In the generated documentation, this javadoc tag shows up as a new entry in the specification list:

## Limitations

Implementations usually fall back to discrete interpolation when the start value is an absolute value, and the end value is a percentage (see the example of `CornerRadii.topLeftHorizontalRadius` above). However, we can often solve these scenarios by first canonicalizing the values before interpolation (i.e. converting percentages to absolute values). This will be a future enhancement.

/reviewers 2
/csr

---------
### Progress
- [ ] Change requires CSR request [JDK-8333455](https://bugs.openjdk.org/browse/JDK-8333455) to be approved

### Issues
 * [JDK-8332895](https://bugs.openjdk.org/browse/JDK-8332895): Support interpolation for backgrounds and borders (**Enhancement** - P4)
 * [JDK-8226911](https://bugs.openjdk.org/browse/JDK-8226911): Interpolatable's contract should be reexamined (**Enhancement** - P4)
 * [JDK-8333455](https://bugs.openjdk.org/browse/JDK-8333455): Support interpolation for backgrounds and borders (**CSR**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1471/head:pull/1471` \
`$ git checkout pull/1471`

`$ git checkout pull/1471` \
`$ git pull https://git.openjdk.org/jfx.git pull/1471/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1471`

`$ git pr show -t 1471`

</details>

<a href="https://git.openjdk.org/jfx/pull/1471.diff">https://git.openjdk.org/jfx/pull/1471.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1471#issuecomment-2143985295)


FlorianKirmaier:
First, Thank you for working on these transitions!

> Note that this PR also changes the specification of Interpolatable to make users aware that they shouldn't assume any particular identity of the object returned from the interpolate() method. This allows the implementation to re-use objects and reduce the number of object allocations.

Can you elaborate on this?
Are changes in the Region.background still trigger change events?
If not, is there a mechanism to get them? Like Transform.onTransformChanged?
Are the previous immutable objects like Background and Border now mutable?

FlorianKirmaier:
Ok, great! Thank you for the feedback.
I feared some magic was happening, but based on your response, this looks good!
As far as I understand this, this is just command sense and, in my opinion, doesn't have to be documented.
It's somewhat confusing and irrelevant.

kevinrushforth:
This seems like a reasonable enhancement. I'll review it (not immediately, though).

@nlisker would you be willing to be the second reviewer?

A couple general comments:

1. Since you propose interpolating objects that aren't simple sets of floating-point fields, the overridden interpolate method should specify the behavior of the fields that are not. For example, different types of Paint are sometimes interpolatable and sometimes return either the start or end; some fields of BackgroundImage are interpolatable and others (notably the image itself) return either the start or the end; and so forth.
2. I spotted at least one place where you override equals, but not hashCode, so you'll need to provide an override for that.

nlisker:
> @nlisker would you be willing to be the second reviewer?

I could do some of the review, but probably no time for a full one.

> Since you propose interpolating objects that aren't simple sets of floating-point fields, the overridden interpolate method should specify the behavior of the fields that are not. For example, different types of Paint are sometimes interpolatable and sometimes return either the start or end; some fields of BackgroundImage are interpolatable and others (notably the image itself) return either the start or the end; and so forth.

When I did the (simple) interpolation work on [Point2D/3D](https://bugs.openjdk.org/browse/JDK-8226454) I also looked at `Border` and `Background` because animating the color of these is rather common. The problem is that they are rather complex and don't interpolate trivially as a whole. As a user, if I want to interpolate between images, I would think of a smooth transition between them, not a jump cut. There are other oddities as well.

nlisker:
> Note that this PR also changes the specification of Interpolatable to make users aware that they shouldn't assume any particular identity of the object returned from the interpolate() method. This allows the implementation to re-use objects and reduce the number of object allocations.

There is an issue for that already: https://bugs.openjdk.org/browse/JDK-8226911. You can take it.

mstr2:
I've added a new "interpolation type" specification for all components of interpolatable classes. This mirrors the "animation type" of the CSS specification, which is listed as an entry in the property table.

@kevinrushforth This might be a candidate for a late enhancement, because the risk is relatively low (it's mostly the addition of the `interpolate()` method to various types), and it seems to be important to make the CSS transitions feature useful (backgrounds and borders will probably be one of the most obvious things developers will want to animate).

modules/javafx.graphics/src/main/java/javafx/scene/paint/Stop.java
```diff
@@ -184,7 +264,44 @@ public final Color getColor() {
      */
     public Stop(@NamedArg("offset") double offset, @NamedArg(value="color", defaultValue="BLACK") Color color) {
         this.offset = offset;
-        this.color = color;
+        this.color = Objects.requireNonNullElse(color, Color.TRANSPARENT);
```
mstr2, line 267: Note that a `null` color is now treated as `TRANSPARENT` for the following reasons:
1. The previous implementation was broken: if a `Stop` is constructed with a `null` color, the `hashCode()` method throws a NPE because it doesn't check for `null`.
2. It doesn't make sense to have a stop with `null` color. What does it even mean?
3. When the stop list is normalized, an empty or null list is treated as a two-stop list with `TRANSPARENT` color (see `normalize()`). So we already have a scenario where `null` is treated as `TRANSPARENT`.

mstr2:
As proposed, this enhancement doesn't cover some edge cases. Back to the drawing board.

PR #1475, opened 2024-06-12
8334161: Enable -Werror for javac tasks to fail on any warnings

kevinrushforth:
This PR is a follow-on to #1474.

Enable `javac -Werror` so that javac warnings, including but not limited to javac lint warnings, will cause the build to fail. As with the fix for [JDK-8327255](https://bugs.openjdk.org/browse/JDK-8327255), we define three new properties to enable `-Werror`, one for each type of java compilation task: sdk classes, test classes (including shims), and tool classes (including JSLC). The defaults for `-Werror` for these three groups are:

* `JAVAC_WERROR` : `true`
* `TOOL_JAVAC_WERROR` : `false`
* `TEST_JAVAC_WERROR` : `false`

They can be overridden on the command line.

I had to do the following two things for the build to pass:

1. Disable `-Werror` for the `javafx.graphics` module until [JDK-8334137](https://bugs.openjdk.org/browse/JDK-8334137) is fixed. The warning generated when using `sun.misc.Unsafe` in JDK 22 or later cannot be suppressed, so we must not use `-Werror` when compiling the graphics module.
2. Define module-specific `extraLintOptions` in the swing module to disable the "options" lint warning, which is enabled by default. That warning is generated because we (necessarily) use `--source 21` instead of `--release 21` for the Swing module, and don't set the location of the system modules.

I tested this, both locally and via GHA, in the following two branches, each of which introduces an error, the first tests that a removal warning will fail the build and the second tests that a missing explicit constructor will fail the build:

* [test-Werror-removal](https://github.com/kevinrushforth/jfx/tree/test-Werror-removal) : [GHA workflow](https://github.com/kevinrushforth/jfx/actions/runs/9664971290)
* [test-Werror-implicit-ctor](https://github.com/kevinrushforth/jfx/tree/test-Werror-implicit-ctor) : [GHA workflow](https://github.com/kevinrushforth/jfx/actions/runs/9664924813)

---------
### Progress

### Issue
 * [JDK-8334161](https://bugs.openjdk.org/browse/JDK-8334161): Enable -Werror for javac tasks to fail on any warnings (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1475/head:pull/1475` \
`$ git checkout pull/1475`

`$ git checkout pull/1475` \
`$ git pull https://git.openjdk.org/jfx.git pull/1475/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1475`

`$ git pr show -t 1475`

</details>

<a href="https://git.openjdk.org/jfx/pull/1475.diff">https://git.openjdk.org/jfx/pull/1475.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1475#issuecomment-2189245281)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -576,6 +576,15 @@ def defaultTestLintOptions = ""
 defineProperty("TEST_LINT", defaultTestLintOptions)
 ext.IS_TEST_LINT = TEST_LINT != ""
 
+defineProperty("JAVAC_WERROR", "true")
+ext.IS_JAVAC_WERROR = Boolean.parseBoolean(JAVAC_WERROR)
+
+defineProperty("TOOL_JAVAC_WERROR", "false")
+ext.IS_TOOL_JAVAC_WERROR = Boolean.parseBoolean(TOOL_JAVAC_WERROR)
+
+defineProperty("TEST_JAVAC_WERROR", "false")
+ext.IS_TEST_JAVAC_WERROR = Boolean.parseBoolean(TEST_JAVAC_WERROR)
+
 // Doclint options (all enabled by default)
 defineProperty("DOC_LINT", "all")
 ext.IS_DOC_LINT = DOC_LINT != ""
@@ -2258,6 +2267,10 @@ project(":graphics") {
     project.ext.moduleRuntime = true
     project.ext.moduleName = "javafx.graphics"
 
+    // FIXME: Remove this setting when JDK-8334137 is fixed
+    // Disable javac -Werror until we stop using sun.misc.Unsafe
+    project.ext.disableJavacWerror = true
+
     getConfigurations().create("antlr");
 
     sourceSets {
@@ -2789,6 +2802,11 @@ project(":swing") {
     // excluded by "--release NN". This will fall back to using
     // "-source NN -target NN" for this module.
     project.ext.skipJavaCompilerOptionRelease = true
+
+    // We also need to disable the implicitly-enabled lint warning that
+    // results from the above
+    project.ext.extraLintOptions =
+        "-options"
 
     tasks.all {
         if (!COMPILE_SWING) it.enabled = false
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -2258,6 +2267,10 @@ project(":graphics") {
     project.ext.moduleRuntime = true
     project.ext.moduleName = "javafx.graphics"
 
+    // FIXME: Remove this setting when JDK-8334137 is fixed
```
andy-goryachev-oracle, line 2270: perhaps refer to https://bugs.openjdk.org/browse/JDK-8334161 as well here?

andy-goryachev-oracle (approved):
introduced a javadoc problem and the build failed:

```
error: warnings found and -Werror specified
1 error
1 warning

> Task :javadoc FAILED
```

one minor comment.

build.gradle
kevinrushforth: Probably not in this case. We generally don't refer to the bug ID that added a line of code or a comment.

I will add a comment to [JDK-8334137](https://bugs.openjdk.org/browse/JDK-8334137) with a note to remove this block as part of fixing it.

arapte:
A few observations on usage:

1. `gradle -PLINT=all` command would fail, or if we enable any other lint warning which if occurs would fail the build.
2. `gradle -PTEST_JAVAC_WERROR=true :controls:test` does not fail even if a warning is present, as by default lint warnings are disabled for test. We need to enable any specific warning that we are looking for; an example as
`gradle -PTEST_LINT=options,unchecked -PTEST_JAVAC_WERROR=true :controls:test`
3. Currently, `gradle -PTEST_LINT=all,-options -PTEST_JAVAC_WERROR=true :controls:test` fails. (with any test task)

arapte (approved):

kevinrushforth:
Thanks for the reviews and the additional observations.

/integrate

PR #1476, opened 2024-06-19
8088923: IOOBE when adding duplicate categories to the BarChart

drmarmac:
This PR provides the test case given in the JBS issue, and a simple fix for the index calculation when inserting data after previous data with duplicate categories.

Also, I've added a comment to `BarChart`s javadoc, clarifying the behavior that was apparently assumed (but broken) previously.

The index lookup is skipped for performance reasons if there are no duplicates, corresponding to the previous implementation.
Further optimizations would be possible, but probably are not really helpful without more extensive changes. The previous code already loops over all categories to check if they are present, typically nested in a loop adding many data items, thus already scaling quadratically when adding lots of mostly unique data points.

---------
### Progress

### Issue
 * [JDK-8088923](https://bugs.openjdk.org/browse/JDK-8088923): IOOBE when adding duplicate categories to the BarChart (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1476/head:pull/1476` \
`$ git checkout pull/1476`

`$ git checkout pull/1476` \
`$ git pull https://git.openjdk.org/jfx.git pull/1476/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1476`

`$ git pr show -t 1476`

</details>

<a href="https://git.openjdk.org/jfx/pull/1476.diff">https://git.openjdk.org/jfx/pull/1476.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1476#issuecomment-2178328333)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
@@ -60,7 +60,7 @@ import javafx.css.StyleableProperty;
 /**
  * A chart that plots bars indicating data values for a category. The bars can be vertical or horizontal depending on
  * which axis is a category axis.
- *
+ * <p>
  * Adding data with multiple occurences of a category to a series shows the last occurence.
  *
  * @param <X> the category axis value type
--- a/modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
@@ -221,6 +221,7 @@ public class BarChart<X,Y> extends XYChart<X,Y> {
             } else {
                 // There may be data items with duplicate categories. Find category insertion index on the axis
                 // by looking at the concatenation of the data of all series, skipping duplicate categories.
+                // The category insertion index is found when the new data's index is reached within its series.
                 categoryIndex = 0;
                 var uniqueCategories = new HashSet<String>();
                 for (var entry : seriesCategoryMap.entrySet()) {
--- a/modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
@@ -211,11 +211,32 @@ public class BarChart<X,Y> extends XYChart<X,Y> {
         }
         // check if category is already present
         if (!categoryAxis.getCategories().contains(category)) {
-            // find category index in case data contains duplicate categories
-            int i = series.getData().size() != categoryAxis.getCategories().size() ? series.getItemIndex(item) :
-                    itemIndex;
+            int seriesCount = getDataSize();
+            int categoryCount = categoryAxis.getCategories().size();
+
+            int categoryIndex;
+            if (seriesCount == 1 && itemIndex == categoryCount) {
+                // shortcut if there is only one series and data contains no duplicates
+                categoryIndex = categoryCount;
+            } else {
+                // There may be data items with duplicate categories. Find category insertion index on the axis
+                // by looking at the concatenation of the data of all series, skipping duplicate categories.
+                categoryIndex = 0;
+                var uniqueCategories = new HashSet<String>();
+                for (Map<String, Data<X,Y>> catMap : seriesCategoryMap.values()) {
+                    for (String cat : catMap.keySet()) {
+                        if (cat == category) {
+                            break;
+                        }
+                        if (uniqueCategories.add(cat)) {
+                            categoryIndex++;
+                        }
+                    }
+                }
+            }
+
             // note: cat axis categories can be updated only when autoranging is true.
-            categoryAxis.getCategories().add(i, category);
+            categoryAxis.getCategories().add(categoryIndex, category);
         } else if (categoryMap.containsKey(category)){
             // RT-21162 : replacing the previous data, first remove the node from scenegraph.
             Data<X,Y> data = categoryMap.get(category);
--- a/modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
@@ -61,6 +61,8 @@ import javafx.css.StyleableProperty;
  * A chart that plots bars indicating data values for a category. The bars can be vertical or horizontal depending on
  * which axis is a category axis.
  *
+ * Adding data with multiple occurences of a category to a series shows the last occurence.
+ *
  * @param <X> the category axis value type
  * @param <Y> the data value type
  * @since JavaFX 2.0
@@ -209,8 +211,11 @@ public class BarChart<X,Y> extends XYChart<X,Y> {
         }
         // check if category is already present
         if (!categoryAxis.getCategories().contains(category)) {
+            // find category index in case data contains duplicate categories
+            int i = series.getData().size() != categoryAxis.getCategories().size() ? series.getItemIndex(item) :
+                    itemIndex;
             // note: cat axis categories can be updated only when autoranging is true.
-            categoryAxis.getCategories().add(itemIndex, category);
+            categoryAxis.getCategories().add(i, category);
         } else if (categoryMap.containsKey(category)){
             // RT-21162 : replacing the previous data, first remove the node from scenegraph.
             Data<X,Y> data = categoryMap.get(category);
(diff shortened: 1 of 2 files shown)


modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
```diff
@@ -61,6 +61,8 @@
  * A chart that plots bars indicating data values for a category. The bars can be vertical or horizontal depending on
  * which axis is a category axis.
  *
+ * Adding data with multiple occurences of a category to a series shows the last occurence.
```
andy-goryachev-oracle, line 64: you probably want to add <p> on line 63

this PR does not require CSR as the change in javadoc is mere clarification rather than a substantial API change.

modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
```diff
@@ -209,8 +211,11 @@ public BarChart(@NamedArg("xAxis") Axis<X> xAxis, @NamedArg("yAxis") Axis<Y> yAx
         }
         // check if category is already present
         if (!categoryAxis.getCategories().contains(category)) {
+            // find category index in case data contains duplicate categories
+            int i = series.getData().size() != categoryAxis.getCategories().size() ? series.getItemIndex(item) :
```
andy-goryachev-oracle, line 215: this is clever: little impact on performance in the case of non-duplicate categories (the most common case).

andy-goryachev-oracle:
@drmarmac can you assign the ticket to yourself please?

drmarmac:
ok, done!

modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
andy-goryachev-oracle: However, I don't see the issue if I just add a duplicate data point after the chart is rendered (the code jumps to line 218 in the master branch).

This makes me think that perhaps the fix should include fixing the logic of adding category to the right hashtables at the right moment.  What do you think?

andy-goryachev-oracle: I've added a popup menu on the BarChart page in the Monkey Tester to test the case ...

drmarmac: You'll need to add an additional point after adding the one with duplicate categories to see the exception in the console. Using the "Add Point" button in the monkey tester after using the context menu entry should reproduce the issue.
Regarding category hashtables, do you have any specific issues in mind, or see any regressions due to this PR? Otherwise I'd suggest to create separate issues / failing tests and do follow-up PRs.

drmarmac: The second added Point specifically needs to be in another, later category for this issue to occur - I created such a point in the monkey tester with the "Add Point" button to the right, not by repeatedly using the new context menu. Maybe try multiple times (depending on how the points random values are implemented there?)

Regarding categories: My understanding is that `series.getData()` is supposed to include all data, including duplicates. On the other hand, `categoryAxis.getCategories()` only contains distinct categories. So the new insertion index of the category of a (not necessarily duplicate) newly added point needs to be shifted only if there are any previous duplicates with lower indices. Thus, I think everything is accounted for correctly unless I'm missing something.

drmarmac:
I'm not able to reproduce this with either master or this PR, can you check if that was the correct order? Note that `chart.getData().getFirst().setData(list);` in `addDuplicateSeries` in the monkey tester actually replaces the data of the previous series, not the newly added one.

andy-goryachev-oracle:
My apologies!  The steps to reproduce (swap the two menus):

- run the latest Monkey Tester
- select BarChart
- right click on chart, select "Add Series with Duplicate Category" item twice
- right click, select "Add Duplicate Category"

andy-goryachev-oracle:
Also (this might be a separate issue), repeated invocation of "Add Series with Duplicate Category" menu causes XAxis legend to display ever increasing number of series:

Screenshot 2024-06-20 at 11 59 46

drmarmac:
Thanks, I can reproduce this now, both in master and this PR, so it may even be another unrelated issue. I can have a look if it actually touches this PR or not.

The legends being added in your screenshot are correct, that's actually what the monkey tester method is currently doing (adding empty series, because the code just modifies the first series). However, the tick marks on the axis are in incorrect places (which I'm pretty sure is the next unrelated bug...)

andy-goryachev-oracle:
> adding empty series

ah, right!



> the tick marks on the axis are in incorrect

this might be JDK-8314754

drmarmac:
Turns out this problem is even worse. If there are multiple series, and some data has duplicate categories within one or across multiple series, the insertion index on the category axis could be more or less anywhere.

I've enhanced the fix by choosing an ordering of the categories on the axis according to the concatenation of the data of each series, skipping duplicates. I've added some more unit tests, and checked the manual monkey tester issue discussed above as well.

Note this category ordering on the axis is not necessarily maintained, e.g. removing previously duplicate data items might cause some rather arbitrary ordering again, so it's probably not safe to optimize this much more without much more control over the methods being called.
Performance is still O(n) per item added, however n now is the sum of all categories of all series.

andy-goryachev-oracle (approved):
Looks good to me.
I'd like a second reviewer to take a look at this.

drmarmac:
The last implementation could have caused a regression if anyone relied on the previous ordering when there are no duplicates and an insertion is made within a series.
I've updated the algorithm and added unit tests to ensure this should now be consistent with the previous behavior, and also works when there are duplicates. @andy-goryachev-oracle @mstr2 Please re-review.

modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
```diff
@@ -223,14 +223,18 @@ public BarChart(@NamedArg("xAxis") Axis<X> xAxis, @NamedArg("yAxis") Axis<Y> yAx
                 // by looking at the concatenation of the data of all series, skipping duplicate categories.
                 categoryIndex = 0;
                 var uniqueCategories = new HashSet<String>();
-                for (Map<String, Data<X,Y>> catMap : seriesCategoryMap.values()) {
+                for (var entry : seriesCategoryMap.entrySet()) {
+                    Series s = entry.getKey();
```
andy-goryachev-oracle, line 227: would it make sense to add a comment explaining the reasoning for this more complex code?

andy-goryachev-oracle:
Found another problem.  It looks like it is present in the master branch, so this is a separate issue.  To reproduce with the Monkey Tester:

- select bar chart page
- right click, select "Add Series with Duplicate Category" twice
- click "Add Series" button on the right, under XY Chart
- resize the chart using the split pane on the right
The series added via context menu are not getting resized

This may or may not be related to https://bugs.openjdk.org/browse/JDK-8314754

Screenshot 2024-06-21 at 15 59 15

modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
drmarmac: I've extended the code comment.

drmarmac:
I can reproduce this, looks unrelated to the issue this PR is trying to fix. The category axis entries look reasonable, while the calculated plot coordinates are off. Not sure if this is the same as [8314754](https://bugs.openjdk.org/browse/JDK-8314754) either, probabaly should get its own ticket.

mstr2 (approved):

modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
andy-goryachev-oracle: thank you

andy-goryachev-oracle (approved):
lgtm.

There is additional performance hit which I think is ok because javafx charts' design is not suitable for large amount of data points anyway.  A typical use of the chart is to render relatively few data points where the impact is minor.

andy-goryachev-oracle:
> I can reproduce this, looks unrelated to the issue this PR is trying to fix.

Created https://bugs.openjdk.org/browse/JDK-8334873

andy-goryachev-oracle:
thank you for fixing these two bugs!

drmarmac:
I found quite a few more, opened a new PR for one of those...

PR #1486, opened 2024-06-24
8334874: Horizontal scroll events from touch pads should scroll the TabPane tabs

jperedadnr:
This PR considers the horizontal scroll events that can be generated on a trackpad, on an horizontally sided `TabPane` control (top or bottom), adding to the existing vertical scroll events from mouse wheel and trackpad, to scroll its tabs.

Therefore, scrolling a `TabPane` will behave in the same way that `ScrollBar` does regarding those same events and control orientation (vertical/horizontal).

---------
### Progress

### Issue
 * [JDK-8334874](https://bugs.openjdk.org/browse/JDK-8334874): Horizontal scroll events from touch pads should scroll the TabPane tabs (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1486/head:pull/1486` \
`$ git checkout pull/1486`

`$ git checkout pull/1486` \
`$ git pull https://git.openjdk.org/jfx.git pull/1486/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1486`

`$ git pr show -t 1486`

</details>

<a href="https://git.openjdk.org/jfx/pull/1486.diff">https://git.openjdk.org/jfx/pull/1486.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1486#issuecomment-2187133415)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -934,17 +934,23 @@ public class TabPaneSkin extends SkinBase<TabPane> {
             // Scrolling the mouse wheel downwards results in the tabs scrolling left (i.e. exposing the right-most tabs)
             // Scrolling the mouse wheel upwards results in the tabs scrolling right (i.e. exposing th left-most tabs)
             addEventHandler(ScrollEvent.SCROLL, (ScrollEvent e) -> {
+                double dx = e.getDeltaX();
+                double dy = e.getDeltaY();
+
                 Side side = getSkinnable().getSide();
                 side = side == null ? Side.TOP : side;
                 switch (side) {
                     default:
                     case TOP:
                     case BOTTOM:
-                        setScrollOffset(scrollOffset + e.getDeltaY());
+                        // Consider vertical scroll events (dy > dx) from mouse wheel and trackpad,
+                        // and horizontal scroll events from a trackpad (dx > dy)
+                        dx = Math.abs(dy) > Math.abs(dx) ? dy : dx;
+                        setScrollOffset(scrollOffset + dx);
                         break;
                     case LEFT:
                     case RIGHT:
-                        setScrollOffset(scrollOffset - e.getDeltaY());
+                        setScrollOffset(scrollOffset - dy);
                         break;
                 }
 
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/TabPaneTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/TabPaneTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -38,6 +38,7 @@ import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import javafx.scene.control.SelectionModel;
+import javafx.scene.input.ScrollEvent;
 import test.com.sun.javafx.scene.control.infrastructure.StageLoader;
 import javafx.application.Platform;
 import javafx.beans.property.BooleanProperty;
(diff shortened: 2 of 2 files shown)


andy-goryachev-oracle:
Considering that this is an improvement and a change in behavior, I would like to suggest a couple of changes:

1. For Side.LEFT/RIGHT, reverse the direction.  What we have now is inconsistent with the movement of e.g. ListView (can be seen if using track pad with the list view and TabPane in the Monkey Tester):

Screenshot 2024-06-24 at 12 32 18

2. Since now the TabPane movement correctly reflects the orientation of track pad gesture, we need to reverse the movement when TabPane effective nodeOrientation is RTL.

3. We don't have the behavior documented currently (some initial docs can be found in /doc-files/behavior).  Maybe it's a good opportunity at least start a file for TabPane?

And, a question - does this change require a CSR, since we are changing the behavior?

kevinrushforth:
> And, a question - does this change require a CSR, since we are changing the behavior?

Good question. Thinking out loud, if what this PR does is to change the existing behavior of top/bottom TabPane scrolling (e.g., horizontal scrollbar) to align with what ScrollPane already does, then... maybe not? (a CSR wouldn't be a bad idea, though)

I'd be happy to hear others thoughts on this.

It does need two reviewers.

/reviewers 2

kevinrushforth:
> 1. For Side.LEFT/RIGHT, reverse the direction. What we have now is inconsistent with the movement of e.g. ListView (can be seen if using track pad with the list view and TabPane in the Monkey Tester):

I don't recommend doing that as part of this PR. That would be a noticeable behavior change, and warrants more discussion.

> 2. Since now the TabPane movement correctly reflects the orientation of track pad gesture, we need to reverse the movement when TabPane effective nodeOrientation is RTL.

Do ScrollPane or ListView work this way already? If they do, then it seems reasonable to do the same here. Otherwise, I'm not sure we should do this for TabPane in isolation.

andy-goryachev-oracle:
> I don't recommend doing that as part of this PR. That would be a noticeable behavior change, and warrants more discussion

we **are** changing the behavior already by changing the direction of the gesture, so I thought we should be able to tackle this issue within the same PR.

The other point is the discussion - these comments are already on the mailing list, what alternative discussion is required?

And what are the acceptance criteria for that discussion?

andy-goryachev-oracle:
> Do ScrollPane or ListView work this way already?

Good point - both have the same issue (I am sure other controls as well, TableView, others did not check).  Definitely deserves a separate ticket, possibly even an umbrella task (RTL navigation).

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
```diff
@@ -934,17 +934,23 @@ public TabHeaderArea() {
...
+
                 Side side = getSkinnable().getSide();
                 side = side == null ? Side.TOP : side;
                 switch (side) {
                     default:
                     case TOP:
                     case BOTTOM:
-                        setScrollOffset(scrollOffset + e.getDeltaY());
+                        // Consider vertical scroll events (dy > dx) from mouse wheel and trackpad,
+                        // and horizontal scroll events from a trackpad (dx > dy)
```
andy-goryachev-oracle, line 947: Maybe we should explain why - to preserve the existing behavior with the mouse (scrolling up/down)?

The trackpad events can have both `dx` and `dy` of various magnitude, both non-zero, so the logic seems to me a bit contrived - like filtering some events, although at the end it still feels ok.

What do you think?

jperedadnr: (sorry I missed to reply)

I believe the comment already states what are we taking into account, but this is a more detailed explanation:

- a mouse wheel only has vertical scroll events (`dx == 0`, so `abs(dy) > abs(dx)`), then we just take `dy` to produce the horizontal scroll offset.
- a trackpad has both scroll events, and we take whichever is bigger. Typically the user will do a horizontal scrolling if tabs are horizontally laid out, and therefore abs(dy) <<< abs(dx), so we take `dx` to produce the horizontal scroll offset, ignoring the smaller `dy` (that wouldn't have any effect in any case).

I'm happy to reword the comment if needed.

andy-goryachev-oracle: Thank you!

A larger question is that ideally we would need a mechanism to differentiate trackpad from mouse events, as they sometimes require different handling.  We can't fully rely on the fact that either dx or dy is exactly 0.0, since theoretically we could get those events from the trackpad.

What do you think?

modules/javafx.controls/src/test/java/test/javafx/scene/control/TabPaneTest.java
```diff
@@ -1241,4 +1242,105 @@ public void testSMLeakOnSwitchSkinAndSM() {
...
+        if (side.equals(Side.TOP) || side.equals(Side.BOTTOM)) {
+            double delta = Math.abs(deltaY) > Math.abs(deltaX) ? deltaY : deltaX;
+            assertEquals(firstTabBounds.getMinX() + delta, newFirstTabBounds.getMinX(), 0);
+            assertEquals(firstTabBounds.getMinY(), newFirstTabBounds.getMinY(), 0);
+        } else {
+            assertEquals(firstTabBounds.getMinX(), newFirstTabBounds.getMinX(), 0);
+            assertEquals(firstTabBounds.getMinY() - deltaY, newFirstTabBounds.getMinY(), 0);
+        }
+    }
+
```
andy-goryachev-oracle, line 1345: please remove unnecessary newline

andy-goryachev-oracle (approved):
I am ok with fixing the direction of vertical scrolling in a separate ticket, as well as RTL reversal of the horizontal scrolling, also as a separate ticket.

Left a couple of minor comments, otherwise looks good.  I'll re-approve if you decide to make changes.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TabPaneTest.java
jperedadnr: done

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
jperedadnr: In this case, the bigger value from the scrolling gesture prevails, ignoring the other small one (it doesn't matter if it is exactly 0 or not).
 
For  the`TabPane` case, that is good enough, since we only want to scroll in one direction, and it really doesn't matter if the source is a mouse wheel or a trackpad.

kevinrushforth (approved):
LGTM

I ran the program attached to the bug report and it now recognizes horizontal scroll as expected. I also verified that the newly added test fails without your fix and passes with your fix.

andy-goryachev-oracle (approved):

PR #1489, opened 2024-06-25
8314754: Minor ticks are not getting updated both the axes in LineChart

drmarmac:
This PR ensures the `tickMarksUpdated()` method (overriden with minor tick mark layout code in subclasses) is called during `Chart` `Axis` layout when needed.
Previously, it was only called when tick mark length or range was changed, which is not happening during axis animations, causing minor ticks to become outdated when animations are enabled.

I've added a test that fails with the previous code, specifically checking the tick mark spacing.
Alternatively the test could be implemented by accessing ValueAxis.minorTickMarkValues, which would need to be made at least package-private however (with a shim). This implementation checks the tick mark `Path` shape instead, which is already publicly accessible (but could be considered an implementation detail).

---------
### Progress

### Issue
 * [JDK-8314754](https://bugs.openjdk.org/browse/JDK-8314754): Minor ticks are not getting updated both the axes in LineChart (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1489/head:pull/1489` \
`$ git checkout pull/1489`

`$ git checkout pull/1489` \
`$ git pull https://git.openjdk.org/jfx.git pull/1489/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1489`

`$ git pr show -t 1489`

</details>

<a href="https://git.openjdk.org/jfx/pull/1489.diff">https://git.openjdk.org/jfx/pull/1489.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1489#issuecomment-2189242931)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/chart/Axis.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/chart/Axis.java
@@ -698,8 +698,6 @@ public abstract class Axis<T> extends Region {
                 }
             }
 
-            // call tick marks updated to inform subclasses that we have updated tick marks
-            tickMarksUpdated();
             // mark all done
             oldLength = length;
             rangeValid = true;
@@ -750,6 +748,9 @@ public abstract class Axis<T> extends Region {
                 }
             }
             updateTickMarks(side, length);
+
+            // call tick marks updated to inform subclasses that we have updated tick marks
+            tickMarksUpdated();
         }
     }
 
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/chart/XYNumberLineChartsTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/chart/XYNumberLineChartsTest.java
@@ -27,7 +27,6 @@ package test.javafx.scene.chart;
 
 import javafx.scene.shape.MoveTo;
 import javafx.scene.shape.Path;
-import org.junit.jupiter.api.Assertions;
 import test.com.sun.javafx.scene.control.infrastructure.ControlTestUtils;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
@@ -170,8 +169,8 @@ public class XYNumberLineChartsTest extends XYNumberChartsTestBase {
         double minorTickSpacing = minorTickYValues.get(1) - minorTickYValues.get(0);
 
         double delta = 0.001;
-        Assertions.assertEquals(5, yAxis.getMinorTickCount());
-        Assertions.assertEquals(5, majorTickSpacing / minorTickSpacing, delta);
+        assertEquals(5, yAxis.getMinorTickCount());
+        assertEquals(5, majorTickSpacing / minorTickSpacing, delta);
     }
 
     @Override
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/chart/XYNumberLineChartsTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/chart/XYNumberLineChartsTest.java
@@ -25,11 +25,17 @@
 
 package test.javafx.scene.chart;
 
+import javafx.scene.shape.MoveTo;
+import javafx.scene.shape.Path;
+import org.junit.jupiter.api.Assertions;
 import test.com.sun.javafx.scene.control.infrastructure.ControlTestUtils;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+
 import javafx.scene.chart.AreaChart;
 import javafx.scene.chart.Axis;
 import javafx.scene.chart.Chart;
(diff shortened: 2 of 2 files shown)


modules/javafx.controls/src/main/java/javafx/scene/chart/Axis.java
```diff
@@ -750,6 +748,9 @@ protected void tickMarksUpdated(){}
                 }
             }
             updateTickMarks(side, length);
+
+            // call tick marks updated to inform subclasses that we have updated tick marks
+            tickMarksUpdated();
```
andy-goryachev-oracle, line 753: is this the only place where this call needs to be added?

for example, in BarChart the x axis is not updated after "Add series with duplicate category" - is this a separate issue?


Screenshot 2024-06-25 at 11 58 52

modules/javafx.controls/src/test/java/test/javafx/scene/chart/XYNumberLineChartsTest.java
```diff
@@ -137,6 +143,38 @@ public void testSeriesClearWithoutSymbolsAnimated_8150264() {
...
+                .filter(path -> path instanceof MoveTo)
+                .map(moveTo -> ((MoveTo) moveTo).getY())
+                .toList();
+
+        List<Double> minorTickYValues = yAxis.getChildrenUnmodifiable().stream()
+                .filter(obj -> obj instanceof Path && obj.getStyleClass().contains("axis-minor-tick-mark"))
+                .flatMap(obj -> ((Path) obj).getElements().stream())
+                .filter(path -> path instanceof MoveTo)
+                .map(moveTo -> ((MoveTo) moveTo).getY())
+                .toList();
```
andy-goryachev-oracle, line 168: very clever - I think this is the right approach to obtain the necessary information via public APIs rather than shims.

just need to be aware that only the nodes present in the scene graph can be found this way (I think, please correct me if I am wrong)

andy-goryachev-oracle:
testing looks good.  also tried AreaChart, BubbleChart, ScatterChart, StackedAreaChart.

modules/javafx.controls/src/test/java/test/javafx/scene/chart/XYNumberLineChartsTest.java
```diff
@@ -137,6 +143,38 @@ public void testSeriesClearWithoutSymbolsAnimated_8150264() {
...
+    public void testMinorTicksMatchMajorTicksAfterAnimation() {
+        startAppWithSeries();
+        chart.setAnimated(true);
+        // increase ranges, starting axis animation
+        chart.getData().getFirst().getData().add(new XYChart.Data<>(100, 100));
+        pulse();
+        toolkit.setAnimationTime(1000);
+        NumberAxis yAxis = (NumberAxis)chart.getYAxis();
+
+        List<Double> majorTickYValues = yAxis.getChildrenUnmodifiable().stream()
```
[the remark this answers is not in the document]
drmarmac: this is after the animation is finished, I added a clarifying comment to the test code.

modules/javafx.controls/src/main/java/javafx/scene/chart/Axis.java
drmarmac: yes, the BarChart issue (which I actually want to get fixed) will need additional fixes in other classes. I could add those to this PR, but probably it's best to handle them in another ticket+PR to keep this PR affecting all `Chart`s minimal?

andy-goryachev-oracle: a separate PR would be best.  I am getting lost - do we have a ticket for that issue already or do we need to create one?

drmarmac: I think we need to create that as it turned out it's not the same as the minor tick problem fixed by this PR.

andy-goryachev-oracle: but wait: this may not be a different issue.  I see a slightly different behavior relative to the master branch:

- select bubble chart page in the monkey tester
- add series with duplicate category

now I see the minor tick marks bunched up in the middle of Y Axis:

Screenshot 2024-06-25 at 14 49 42

modules/javafx.controls/src/test/java/test/javafx/scene/chart/XYNumberLineChartsTest.java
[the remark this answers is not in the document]
andy-goryachev-oracle: thank you

andy-goryachev-oracle (approved):
Simple enough fix, I think 1 reviewer is sufficient, but feel free.

Also tested with AreaChart, BubbleChart, ScatterChart, StackedAreaChart.

modules/javafx.controls/src/main/java/javafx/scene/chart/Axis.java
drmarmac: ok, good, I did not manage to reproduce the minor tick problem on the Y axis in the bar chart unless I switch back to master.

andy-goryachev-oracle: thank you for finding and fixing these issues!

Maran23 (commented):
Left one comment, looks good to me otherwise as well!

modules/javafx.controls/src/test/java/test/javafx/scene/chart/XYNumberLineChartsTest.java
```diff
@@ -25,11 +25,15 @@
 
 package test.javafx.scene.chart;
 
+import javafx.scene.shape.MoveTo;
+import javafx.scene.shape.Path;
+import org.junit.jupiter.api.Assertions;
```
Maran23, line 30: Note: This mixes JUnit 4 with JUnit 5 API. AFAIK, this is nothing we want, so it is better to use the already imported `assertEquals` here.

drmarmac: right, I've changed this to JUnit 4 and removed the JUnit 5 import.

Maran23 (approved):

andy-goryachev-oracle (approved):

PR #1490, opened 2024-06-26
8326712: Robot tests fail on XWayland

azvegint:
Most of the headful test failures on XWayland are due to screen capture is not working.

Wayland, unlike X11, does not allow arbitrary applications to capture the screen contents directly.
Instead, screen capture functionality is managed by the compositor, which can enforce stricter controls and permissions, requiring explicit user approval for screen capture operations.

This issue is already resolved in OpenJDK ([base issue](https://bugs.openjdk.org/browse/JDK-8280982), there are  subsequent fixes) by using the [ScreenCast XDG portal](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.ScreenCast.html).

The XDG ScreenCast portal is utilized for capturing screen data as it provides a secure and standardized method for applications to access screen content.
Additionally, it allows for the reuse of user permissions without requiring repeated confirmations, streamlining the user experience and enhancing convenience.

<hr>

So this changeset is a copy of the OpenJDK fixes with the addition of the JavaFX customization. 
For ease of review, you can skip the changes in the first two commits:
- First commit is a direct copy of files from OpenJDK
- Second commit removes the `gtk-` prefix before the `gtk_...` and `g_...` function calls (in AWT all gtk functions are dynamically loaded, we don't need that in JavaFX).

properties added:

- `javafx.robot.screenshotMethod`, accepts `gtk`(existing gtk method) and `dbusScreencast`(added by this changeset, used by default for Wayland)
- `javafx.robot.screenshotDebug` prints debug info if it is set to `true`

<hr>

What are the remaining issues?

1. https://bugs.openjdk.org/browse/JDK-8335468

After applying this fix, system tests will pass except for the `SwingNodeJDialogTest` test.

This interop test calls `java.awt.Robot#getPixelColor` which internally `gtk->g_main_context_iteration(NULL, TRUE);` causes a blocking of javafx gtk loop, so the test hangs.
So a change is required on OpenJDK side to fix this issue.

2. https://bugs.openjdk.org/browse/JDK-8335470

Even after solving the `#1`, the `SwingNodeJDialogTest.testNodeRemovalBeforeShow` case is still failing.

3. https://bugs.openjdk.org/browse/JDK-8335469

Internally the ScreenCast session keeps open for [2s](https://github.com/openjdk/jdk/blob/d457609f700bbb1fed233f1a04501c995852e5ac/src/java.desktop/unix/classes/sun/awt/screencast/ScreencastHelper.java#L62).
This is to reduce overhead in case of frequent consecutive screen captures.

There is a crash when an AWT ScreenCast session overlaps with the FX ScreenCast session.  E.g.  `java.awt.Robot#getPixelColor()` and `javafx.scene.robot.Robot#getPixelColor()` are called within 1 second.

---------
### Progress

### Issue
 * [JDK-8326712](https://bugs.openjdk.org/browse/JDK-8326712): Robot tests fail on XWayland (**Bug** - P3)

### Reviewers
 * [Joeri Sykora](https://openjdk.org/census#sykora) (@tiainen - Author) 🔄 Re-review required (review applies to [3aa56a2c](https://git.openjdk.org/jfx/pull/1490/files/3aa56a2c0d065ed818f0517f4d1ebd8be8ccc554))

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1490/head:pull/1490` \
`$ git checkout pull/1490`

`$ git checkout pull/1490` \
`$ git pull https://git.openjdk.org/jfx.git pull/1490/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1490`

`$ git pr show -t 1490`

</details>

<a href="https://git.openjdk.org/jfx/pull/1490.diff">https://git.openjdk.org/jfx/pull/1490.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1490#issuecomment-2196353894)


tests/system/src/test/java/test/robot/javafx/embed/swing/SwingNodeJDialogTest.java
```diff
@@ -34,7 +34,7 @@
 
 public class SwingNodeJDialogTest extends SwingNodeBase {
 
-    @Test
+    @Test(timeout = 15000)
```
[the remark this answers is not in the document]
azvegint: Yes.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/gtk/screencast/ScreencastHelper.java
```diff
@@ -0,0 +1,217 @@
...
+
+
+    private ScreencastHelper() {
+    }
+
+    static {
+        SCREENCAST_DEBUG =
+                AccessController.doPrivileged((PrivilegedAction<Boolean>) () -> {
+                    final String str =
+                            System.getProperty("fx.robot.screenshotDebug", "false");
```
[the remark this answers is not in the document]
azvegint: Done.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/gtk/screencast/TokenStorage.java
```diff
@@ -0,0 +1,461 @@
...
+ * The restore token allows the ScreenCast session to be restored
+ * with previously granted screen access permissions.
+ */
+@SuppressWarnings("removal")
+final class TokenStorage {
+
+    private TokenStorage() {}
+
+    private static final String REL_NAME =
+            ".fx/robot/screencast-tokens.properties";
```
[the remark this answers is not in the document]
azvegint: Updated to check two locations:

1. `.java/.robot/screencast-tokens.properties` - the default, should be used by both OpenJDK and JavaFx
2. `.awt/robot/screencast-tokens.properties` - this is currently the only location used by OpenJDK, but it should be changed to the first location after [JDK-8335267](https://bugs.openjdk.org/browse/JDK-8335267) is resolved.

If there is a writable file at location `#2`, we won't try to read or write to location `#1`.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/gtk/GtkRobot.java
```diff
@@ -32,8 +33,30 @@
 import com.sun.glass.ui.GlassRobot;
 import com.sun.glass.ui.Screen;
 
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
+@SuppressWarnings("removal")
```
kevinrushforth, line 39: Maybe this can go on the individual elements that use the security manager? To do this, you might need to create a local "tmp" variable and then assign that to `screenshotMethod`.
(either way, don't worry about the files you copied from AWT)

modules/javafx.graphics/src/main/native-glass/gtk/fp_pipewire.h
```diff
@@ -0,0 +1,91 @@
...
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+#ifdef HEADLESS
```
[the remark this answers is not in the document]
kevinrushforth: I don't have a strong opinion. This file is unchanged relative to the original in AWT. The other three already have some changes, so I'd probably lean toward removing it.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/gtk/screencast/TokenStorage.java
```diff
@@ -109,26 +111,33 @@ private static Path setupPath() {
             return null;
         }
 
-        Path path = Path.of(userHome, REL_NAME);
+        Path secondaryPath = Path.of(userHome, REL_NAME_SECONDARY);
+
+        // use secondary path only if it already exists and writable
+        Path path = Files.isWritable(secondaryPath)
```
kevinrushforth, line 117: If both exist and are writable, then wouldn't you want to default to the primary path? If so, then I think the test should be something like `Files.isWritable(secondaryPath) && !Files.isWritable(primaryPath)`.

kevinrushforth (commented):
I did a cursory review, just looking at the changes since the first two commits, and it looks fine. Preliminary testing looks good.

I think this is ready to take out of Draft.

kevinrushforth:
We'll want extra pairs of eyes on this one (so at least two "R"eviewers). I'd also like Gluon to verify that it builds on their CI system.

Reviewers: @kevinrushforth @prrace @tiainen 

/reviewers 2 reviewers

kevinrushforth:
@azvegint When you do take it out of Draft, please write up something in the Description about the changes as a help to those who will be reviewing it.

Also, can you merge in the latest master? In particular, I want to see a GHA run with the `javac -Werror` fix.

modules/javafx.graphics/src/main/native-glass/gtk/fp_pipewire.h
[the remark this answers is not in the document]
johanvos: With the headless platform that I started in the sandbox project, there is the concept of headless, but rather than compiling native code with `#ifdef HEADLESS` the current approach is to avoid those native compilations at all.
However, I don't exclude that some parts of a file might make sense for a headless glass platform, where other parts should not be considered. In that case, the `ifdef HEADLESS` and in general the compile option `-DHEADLESS` would make sense.
While this is not needed at this moment (and not even sure it will ever be needed), I am slightly in favor of keeping it here.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/gtk/screencast/ScreencastHelper.java
```diff
@@ -0,0 +1,233 @@
...
+
+        checkReturnValue(retVal);
+    }
+
+    private static boolean checkReturnValue(int retVal) {
+        if (retVal == DENIED) {
+            // user explicitly denied the capture, no more tries.
+            throw new SecurityException(
+                    "Screen Capture in the selected area was not allowed"
+            );
```
azvegint, line 204: @kevinrushforth 
I think this SecurityException should be removed, as it is not mentioned in the javadoc for the `getPixelColor` or `getScreenCapture`.  And we also do not want to throw another runtime exception.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/gtk/GtkRobot.java
azvegint: Done for all cases.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/gtk/screencast/TokenStorage.java
azvegint: Sure, it sounds reasonable, updated.

modules/javafx.graphics/src/main/native-glass/gtk/fp_pipewire.h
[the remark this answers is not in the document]
azvegint: I already did the removal earlier, but I can bring it back once we come to an agreement.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/gtk/screencast/ScreencastHelper.java
johanvos: If the SecurityException is not thrown, how does the caller of `getRGBPixels` knows that the result will not be correct because of the missing permission? (as opposed to being incorrect due to a failure)

azvegint: > If the SecurityException is not thrown, how does the caller of getRGBPixels knows that the result will not be correct because of the missing permission? (as opposed to being incorrect due to a failure)

He won't know it. An indirect result of failure can be a black image.
The [current documentation](https://github.com/openjdk/jfx/blob/master/modules/javafx.graphics/src/main/java/javafx/scene/robot/Robot.java#L240) documentation specifies only IllegalStateException and NullPointerException.

modules/javafx.graphics/src/main/native-glass/gtk/fp_pipewire.h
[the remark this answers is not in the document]
kevinrushforth: Johan makes a good point, so go ahead and add it back.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/gtk/screencast/ScreencastHelper.java
kevinrushforth: Since `SecurityException` is deprecated, we should not throw that.

In a similar situation on macOS (user has not enabled screen capture), we return a blank (all black) image, so I think that's what we would do here.

modules/javafx.graphics/src/main/native-glass/gtk/fp_pipewire.h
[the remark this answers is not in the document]
kevinrushforth: Worth noting is that this `HEADLESS` check is just an assertion to verify that the build logic properly excludes the file from the build when building headless. It otherwise does nothing.

tiainen (approved):
The build still works as expected and documented on our CI infrastructure.

tsayao:
Just of out curiosity, please don't mind it too much:

Why the piperwire include files are included? Since the infraestucture should be present for it to work (pipewire is quite new and not present on older linux distributions) a runtime check is needed anyways.

What's the advantage of calling dbus functions directly and not use libportal?

azvegint:
> Just of out curiosity, please don't mind it too much:
> 
> Why the piperwire include files are included? Since the infraestucture should be present for it to work (pipewire is quite new and not present on older linux distributions) a runtime check is needed anyways.
> 
> What's the advantage of calling dbus functions directly and not use libportal?

This is mostly a copy and paste of the OpenJDK code, which should be able to be built on fairly old Linux distributions (consider OpenJDK backports to earlier versions), and we didn't want to add any new external dependencies there.
So with this changeset, we are just bringing tested code with minimal changes to JFX.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/gtk/screencast/ScreencastHelper.java
azvegint: Updated to return blank image in case of permission denial.

modules/javafx.graphics/src/main/native-glass/gtk/fp_pipewire.h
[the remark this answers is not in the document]
azvegint: The `HEADLESS` check is back.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/gtk/screencast/ScreencastHelper.java
prrace: > Since `SecurityException` is deprecated, we should not throw that.
> 
> In a similar situation on macOS (user has not enabled screen capture), we return a blank (all black) image, so I think that's what we would do here.

No, SecurityException is NOT deprecated and there are no plans to deprecate it.

https://download.java.net/java/early_access/jdk23/docs/api/java.base/java/lang/SecurityException.html#%3Cinit%3E()

This kind of platform reason for a "SecurityException" is precisely why it isn't deprecated even though the Java SecurityManager is deprecated.

kevinrushforth: Thanks for the correction about "SecurityException" not being deprecated.

I still think that for this PR, we should not throw an exception, since we A) don't specify it to do so, and B) don't do that on macOS in a nearly identical situation.

We could consider a future RFE to change the spec (and maybe the behavior).

kevinrushforth:
The code changes all look good. I've done a fair bit of testing already, and will finish up my testing tomorrow.

tests/system/src/test/java/test/robot/javafx/embed/swing/SwingNodeBase.java
```diff
@@ -199,6 +204,12 @@ public void testAbove(boolean above) {
             robot.mouseMove(clickLoc, clickLoc);
             robot.mousePress(InputEvent.BUTTON1_DOWN_MASK);
             robot.mouseRelease(InputEvent.BUTTON1_DOWN_MASK);
+
+            // Emulating mouse clicks on XWayland with XTEST does not currently affect the Wayland compositor,
+            // so trying to click on the window title or window body will not bring the window to the front.
```
azvegint, line 209: This is the [same issue](https://bugs.openjdk.org/browse/JDK-8280990) we have in OpenJDK, when the emulated mouse click doesn't put one window in front of another.
After this fix and applying a fix for [the hang issue](https://bugs.openjdk.org/browse/JDK-8335468), all the `SwingNodeJDialogTest` tests are passing.

kevinrushforth (approved):
Testing is complete. No new issues. I only see the `SwingNodeJDialogTest` failure.

I did leave one minor naming comment and will reapprove if you make the suggested change.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/gtk/screencast/TokenStorage.java
```diff
@@ -60,18 +60,20 @@
  * The restore token allows the ScreenCast session to be restored
  * with previously granted screen access permissions.
  */
-@SuppressWarnings("removal")
 final class TokenStorage {
 
     private TokenStorage() {}
 
     private static final String REL_NAME =
+            ".java/.robot/screencast-tokens.properties";
```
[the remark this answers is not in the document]
azvegint: Sure, updated.

kevinrushforth (approved):

prrace (approved):

modules/javafx.graphics/src/main/java/com/sun/javafx/geom/Dimension.java
```diff
@@ -0,0 +1,50 @@
...
+public class Dimension {
+    public int width;
+    public int height;
+
+    public Dimension(int width, int height) {
+        this.width = width;
+        this.height = height;
+    }
+
+    public boolean equals(Object obj) {
```
andy-goryachev-oracle, line 39: [JDK-8335633](https://bugs.openjdk.org/browse/JDK-8335633)
Missing @Override in Dimension

PR #1491, opened 2024-06-26
8335218: Eclipse Config: Remove Gradle Integration

andy-goryachev-oracle:
This might be controversial. I am proposing to remove the Gradle integration in the Eclipse config files.

Problem
=======
Eclipse Gradle integration (Buildship) cannot import the OpenJFX build.gradle cleanly. Every time the project is imported into a new workspace (or re-opened after being closed) it executes Gradle, creates and modifies a number of Eclipse .project and .classpath files, all of which need to be reverted for Eclipse workspace to become usable again.

Solution
======
Remove Gradle nature from the Eclipse project files. This change only affects Eclipse config files and does not impact build.gradle or other IDEs.

Advantages
=========
1. The multiple nested projects in the repo will get imported cleanly on the first attempt, will not require additional steps to clear the Buildship changes.
2. completely removes the dependency on the Eclipse Buildship and its idiosyncrasies.

NOTES:
- even though the reverse was done for IntelliJ, but its gradle import still does not import tests cleanly, see [JDK-8223373](https://bugs.openjdk.org/browse/JDK-8223373)
- this improvement contradicts [JDK-8223374](https://bugs.openjdk.org/browse/JDK-8223374) as without Eclipse files in the repo, it will be impossible to use Eclipse in a meaningful way without the fully functional Buildship support, and that is a big IF.
- once integrated, Eclipse users would only need to re-import the main project with 'search for nested projects' enabled

---------
### Progress

### Issue
 * [JDK-8335218](https://bugs.openjdk.org/browse/JDK-8335218): Eclipse Config: Remove Gradle Integration (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1491/head:pull/1491` \
`$ git checkout pull/1491`

`$ git checkout pull/1491` \
`$ git pull https://git.openjdk.org/jfx.git pull/1491/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1491`

`$ git pr show -t 1491`

</details>

<a href="https://git.openjdk.org/jfx/pull/1491.diff">https://git.openjdk.org/jfx/pull/1491.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1491#issuecomment-2192685852)

### DIFF
--- a/.project
+++ b/.project
@@ -4,10 +4,5 @@
 	<projects>
 	</projects>
 	<buildSpec>
-		<buildCommand>
-			<name>org.eclipse.buildship.core.gradleprojectbuilder</name>
-			<arguments>
-			</arguments>
-		</buildCommand>
 	</buildSpec>
 </projectDescription>
--- a/.project
+++ b/.project
@@ -10,7 +10,4 @@
 			</arguments>
 		</buildCommand>
 	</buildSpec>
-	<natures>
-		<nature>org.eclipse.buildship.core.gradleprojectnature</nature>
-	</natures>
 </projectDescription>
--- a/apps/samples/3DViewer/.project
+++ b/apps/samples/3DViewer/.project
@@ -10,14 +10,8 @@
 			<arguments>
 			</arguments>
 		</buildCommand>
-		<buildCommand>
-			<name>org.eclipse.buildship.core.gradleprojectbuilder</name>
-			<arguments>
-			</arguments>
-		</buildCommand>
 	</buildSpec>
 	<natures>
 		<nature>org.eclipse.jdt.core.javanature</nature>
-		<nature>org.eclipse.buildship.core.gradleprojectnature</nature>
 	</natures>
 </projectDescription>
--- a/apps/samples/Ensemble8/.project
+++ b/apps/samples/Ensemble8/.project
@@ -10,14 +10,8 @@
 			<arguments>
 			</arguments>
 		</buildCommand>
-		<buildCommand>
-			<name>org.eclipse.buildship.core.gradleprojectbuilder</name>
-			<arguments>
-			</arguments>
-		</buildCommand>
 	</buildSpec>
 	<natures>
 		<nature>org.eclipse.jdt.core.javanature</nature>
-		<nature>org.eclipse.buildship.core.gradleprojectnature</nature>
 	</natures>
 </projectDescription>
--- a/apps/samples/Modena/.project
+++ b/apps/samples/Modena/.project
@@ -10,14 +10,8 @@
 			<arguments>
 			</arguments>
 		</buildCommand>
-		<buildCommand>
-			<name>org.eclipse.buildship.core.gradleprojectbuilder</name>
-			<arguments>
-			</arguments>
-		</buildCommand>
 	</buildSpec>
 	<natures>
 		<nature>org.eclipse.jdt.core.javanature</nature>
-		<nature>org.eclipse.buildship.core.gradleprojectnature</nature>
 	</natures>
 </projectDescription>


kevinrushforth:
Reviewers: @nlisker 

This will need discussion among and review by the Eclipse users on the list.

/reviewers 2

lukostyra:
+1 for this change from me.

I use VS Code for development, on the marketplace most popular and recommended Java support extension is the Red Hat extension which is essentially repackaged Eclipse. It was always frustrating to try and open JFX only to find out that something doesn't load correctly because the plugin overwrote .project and .classpath files.

I checked this PR and with default extension settings it loads properly. All regular features also work fine, so the plugin definitely works correctly after this PR. I only noticed that there are still modified `.project` files in `apps` directory - I wonder if it's just VS Code plugin deal, but it doesn't seem to interrupt the build process for the plugin.

andy-goryachev-oracle:
@lukostyra thank you for testing!  there should be no changes made to the .project or any other files.  could you reset your workspace and maybe try again?

if the changes do appear, what are they?

nlisker:
Continuing the [mailing list](https://mail.openjdk.org/pipermail/openjfx-dev/2024-June/047467.html) discussion with more technical details.

Gradle tries to configure Eclipse based on its build file. This includes things like source folders, compiler settings, dependencies and others. In the cases that I have tried in my projects, this works well. One problem that still remains with this configuration is that it's not aware of the locations of the modules. This causes missing module errors when launching, which require `-add-modules` to remedy. The [eclipse plugin](https://docs.gradle.org/current/userguide/eclipse_plugin.html) solves this by resolving the paths locally for your gradle cache.

This tends to be the preferred way to do things from my experience since you don't need to push eclipse-specific files to the repo. The gradle files need to be there anyway, so it makes sense to configure eclipse from them and you maintain uniformity between the different development environments. Where this fails is with OpenJFX projects. The monolithic and old gradle file uses conventions from old gradle versions. This causes faulty generation of the eclipse files with and without the eclipse plugin. For example, all the source folders (`src/main/java`, `src/test/java`...) are prepended with the project/module name (`base/src/main/java`) because of manual source set configurations that shouldn't exist, at least not the way they are now.

The problem presented in the issue has 2 solutions:
1. Opt out: Include the Gralde nature by default, do all the initial building, then revert the Eclipse files. If you want to disable the Gradle integration, you can do so as well be removing the nature within the IDE.
2. Opt in: Don't include the Gradle nature by default and do all the initial building. If you want the Gradle integration, you can add it within the IDE.
Both of these are one-time actions.

I'm fine with either since I find the one-time action to be cheap.

By the way, with this change you will also have to update the [wiki page](https://wiki.openjdk.org/display/OpenJFX/Using+an+IDE#UsinganIDE-UsingEclipse) to a state where the Java import is the standard and there's an opt-in for the Gradle integration.

@hjohn Do you have any remarks?

hjohn:
> Do you have any remarks?

I personally don't launch build tools from the IDE (neither Gradle, nor Maven) but only use such tools to import projects; after the import I expect the IDE to be able to build and run tests without intervention of any build tool, and will use the build tooling (usually on the command line) as a last resort if I can't get it to work that way (which is sometimes the case for a complicated project like JavaFX, especially when I need to run a headless test).  Importing and working this way works for almost any project, but never has worked for me with JavaFX (which was a big barrier for me when I first tried contributing to FX -- I even tried switching IDE's, but it wasn't much better there either).

What annoys me the most with the FX project is the inability to run certain tests directly from the IDE by right clicking and selecting `Run as... JUnit test`.  I work around this by running the test once, then changing its configuration and adding a standard incantation to its command line (`-Djavafx.toolkit`, `-Djava.library.path`, some `--add-modules`) -- I can live with that, although it is less than ideal (right clicking and running a test would be nice to have working out of the box for JavaFX).

I doubt however this change will make this any better or worse :)

So, I'm fine with either solution here.  As I don't use Gradle integration, I'd lean slightly to just removing it, or alternatively, fixing the problems in the gradle build so Buildship recognizes it properly.  Since I suspect the latter is not an easy task, removing the Gradle nature seems like a good alternative.

andy-goryachev-oracle:
Thank you @lukostyra  for checking!

> All of `.project`s have the same bit added (seems like something internal for the plugin):

This change looks innocuous enough - at least it does not break the build.  Why VSCode does it is a mystery, but I suppose that's a bug in VSCode plugin.

andy-goryachev-oracle:
> What annoys me the most with the FX project is the inability to run certain tests directly from the IDE by right clicking and selecting `Run as... JUnit test`.

Yep, same issue here, and same workaround.

I wonder if it is possible to create a separate project(s) for tests that will have all the necessary dependencies added (as a separate issue).

For example, headless tests require addition of these vm args:

```
-Djava.library.path="../../../build/sdk/lib"
-Djavafx.toolkit="test.com.sun.javafx.pgstub.StubToolkit"
-ea
```

while system tests are slightly different:

```
--add-modules=javafx.base,javafx.graphics,javafx.controls,javafx.swing
--add-opens javafx.controls/test.com.sun.javafx.scene.control.test=javafx.base
--add-exports javafx.base/com.sun.javafx=ALL-UNNAMED
-Djava.library.path="../../../../build/sdk/lib"
```

andy-goryachev-oracle:
thank you @nlisker for your thoughts!

> The [eclipse plugin](https://docs.gradle.org/current/userguide/eclipse_plugin.html) solves this by resolving the paths locally for your gradle cache.

so this is a non-standard setup, meaning it is not described in the https://wiki.openjdk.org/display/OpenJFX/Using+an+IDE#UsinganIDE-UsingEclipse wiki page.

Still, I would like to propose we proceed with this change even if it is a temporary one (until  the Buildship plug-in gets mature enough to be able to import the gradle configuration cleanly) because it prevents the said Buildship from messing up the project files and breaking the build every time the project(s) are imported or reopened.

nlisker:
> I work around this by running the test once, then changing its configuration and adding a standard incantation to its command line (`-Djavafx.toolkit`, `-Djava.library.path`, some `--add-modules`) 

The `--add-modules` could be taken care of by the eclipse plugin. The other 2 are javafx specific.

> fixing the problems in the gradle build so Buildship recognizes it properly. Since I suspect the latter is not an easy task, removing the Gradle nature seems like a good alternative.

Yes, https://github.com/openjdk/jfx/pull/1232 should be part of the effort to simplify the gradle build file. It will take several iterations on that file to modernize OpenJFX's build.

> so this is a non-standard setup, meaning it is not described in the https://wiki.openjdk.org/display/OpenJFX/Using+an+IDE#UsinganIDE-UsingEclipse wiki page.

Yes, it's not described in the wiki because it can't read the gradle file properly to begin with.

> Still, I would like to propose we proceed with this change even if it is a temporary one (until the Buildship plug-in gets mature enough to be able to import the gradle configuration cleanly) because it prevents the said Buildship from messing up the project files and breaking the build every time the project(s) are imported or reopened.

I'm fine with this. Just noting that we're not waiting on Buildship to catch up, we need our `build.gradle` to catch up :)

nlisker (approved):

hjohn (approved):

andy-goryachev-oracle:
thank you for reviewing!  will integrate after the fork.

PR #1492, opened 2024-06-27
8198830: BarChart: auto-range of CategoryAxis not working on dynamically setting data

drmarmac:
This PR fixes the placement of `BarChart` bars and category tick marks, particularly when adding data / multiple series and enabling animations.
It covers all cases mentioned in the JBS ticket and adds a unit test.

The settable `barGap` and `categoryGap` now should also behave as expected.

There's a change regarding consistent placement of bars when there is more than one series:
Now, for example, if there are two series S1 and S2, S1 bars will always be on the left and S2 bars will always be on the right side of the tick mark.
This means that if some data category (=x-value) is only present in S2, but not S1, the bar will still be drawn on the right. The previous behavior, if the marks weren't off completely, would have put it on the left which I'd say looks inconsistent.
There's a test for this situation as well.

Note this does not fix [JDK-8334873](https://bugs.openjdk.org/browse/JDK-8334873) where bars get stuck while having different widths. There seem to be additional issues which seem not directly related to the changed code and are probably out of scope for this PR (unless we see some regressions).

---------
### Progress

### Issue
 * [JDK-8198830](https://bugs.openjdk.org/browse/JDK-8198830): BarChart: auto-range of CategoryAxis not working on dynamically setting data (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1492/head:pull/1492` \
`$ git checkout pull/1492`

`$ git checkout pull/1492` \
`$ git pull https://git.openjdk.org/jfx.git pull/1492/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1492`

`$ git pr show -t 1492`

</details>

<a href="https://git.openjdk.org/jfx/pull/1492.diff">https://git.openjdk.org/jfx/pull/1492.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1492#issuecomment-2195612462)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
@@ -423,9 +423,9 @@ public class BarChart<X,Y> extends XYChart<X,Y> {
                         bar.resizeRelocate( bottom, categoryPos + barOffset + (barWidth + getBarGap()) * index,
                                             top-bottom, barWidth);
                     }
-
-                    index++;
                 }
+
+                index++;
             }
             catIndex++;
         }
--- a/modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
@@ -384,7 +384,7 @@ public class BarChart<X,Y> extends XYChart<X,Y> {
     @Override protected void layoutPlotChildren() {
         double catSpace = categoryAxis.getCategorySpacing();
         // calculate bar spacing
-        final double availableBarSpace = catSpace - (getCategoryGap() + getBarGap());
+        final double availableBarSpace = catSpace - getCategoryGap() + getBarGap();
         double barWidth = (availableBarSpace / getSeriesSize()) - getBarGap();
         final double barOffset = -((catSpace - getCategoryGap()) / 2);
         final double zeroPos = (valueAxis.getLowerBound() > 0) ?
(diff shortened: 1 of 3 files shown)


modules/javafx.controls/src/main/java/javafx/scene/chart/CategoryAxis.java
```diff
@@ -68,8 +68,17 @@ public final class CategoryAxis extends Axis<String> {
     // -------------- PRIVATE FIELDS -------------------------------------------
     private List<String> allDataCategories = new ArrayList<>();
     private boolean changeIsLocal = false;
-    /** This is the gap between one category and the next along this axis */
-    private final DoubleProperty firstCategoryPos = new SimpleDoubleProperty(this, "firstCategoryPos", 0);
+
+    /** This is the position of the first category along this axis */
+    private final DoubleProperty firstCategoryPos =
+            new SimpleDoubleProperty(this, "firstCategoryPos", 0) {
```
andy-goryachev-oracle, line 74: question: shouldn't line 74 be indented 4 spaces instead 8?

also, we probably should update copyright year in all the touched files, like so:
```
 * Copyright (c) 2010, 2024, Oracle and/or its affiliates. All rights reserved.
```

(I'll do a full review tomorrow)

andy-goryachev-oracle (approved):
Thank you @drmarmac for fixing all this bug (and adding unit tests!)

Verified that the bug is fixed, including the two scenarios mentioned in the comments, with animation on and off.

The fix is pretty straightforward, I think one reviewer is sufficient.

modules/javafx.controls/src/main/java/javafx/scene/chart/BarChart.java
```diff
@@ -423,9 +423,9 @@ public BarChart(@NamedArg("xAxis") Axis<X> xAxis, @NamedArg("yAxis") Axis<Y> yAx
                         bar.resizeRelocate( bottom, categoryPos + barOffset + (barWidth + getBarGap()) * index,
                                             top-bottom, barWidth);
                     }
-
-                    index++;
                 }
+
+                index++;
```
andy-goryachev-oracle, line 428: this is correct.  not sure how the existing tests did not find this problem.

PR #1493, opened 2024-06-28
8334593: Adding, removing and then adding a JFXPanel again leads to NullPointerException

prsadhuk:
Adding, then removing, and then adding a JFXPanel to the same component in the Swing scene graph leads to a NullPointerException in GlassScene because the sceneState is null.
Removing JFXPanel calls JFXPanel.removeNotify which calls Window.hide which calls SceneHelper.disposePeer -> Scene.disposePeer -> EmbeddedScene.dispose -> GlassScene.dispose which sets "sceneState" to null...
so when GlassScene.updateSceneState is called, it results in NPE.
Fix is to check if `host` (which is usually javafx.embed.swing.JFXPanel$HostContainer for active JFXPanel) has been reset/deleted which is done when `EmbeddedScene.dispose` is called during removeNotify and abstain from updating the scene state

---------
### Progress

### Issue
 * [JDK-8334593](https://bugs.openjdk.org/browse/JDK-8334593): Adding, removing and then adding a JFXPanel again leads to NullPointerException (**Bug** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1493/head:pull/1493` \
`$ git checkout pull/1493`

`$ git checkout pull/1493` \
`$ git pull https://git.openjdk.org/jfx.git pull/1493/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1493`

`$ git pr show -t 1493`

</details>

<a href="https://git.openjdk.org/jfx/pull/1493.diff">https://git.openjdk.org/jfx/pull/1493.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1493#issuecomment-2196442641)

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/EmbeddedScene.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/EmbeddedScene.java
@@ -157,7 +157,9 @@ final class EmbeddedScene extends GlassScene implements EmbeddedSceneInterface {
         entireSceneNeedsRepaint();
         Platform.runLater(() -> {
             QuantumToolkit.runWithRenderLock(() -> {
-                updateSceneState();
+                if (host != null) {
+                    updateSceneState();
+                }
                 return null;
             });
         });
--- a/tests/system/src/test/java/test/javafx/embed/swing/JFXPanelNPETest.java
+++ b/tests/system/src/test/java/test/javafx/embed/swing/JFXPanelNPETest.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright (c) 2024, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package test.javafx.embed.swing;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
+import javax.swing.JFrame;
+import javax.swing.SwingUtilities;
+
+import javafx.application.Application;
+import javafx.application.Platform;
+import javafx.embed.swing.JFXPanel;
+import javafx.scene.Scene;
+import javafx.scene.control.Button;
+import javafx.stage.Stage;
+
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import test.util.Util;
+
+public class JFXPanelNPETest implements Thread.UncaughtExceptionHandler {
+    // Used to launch the application before running any test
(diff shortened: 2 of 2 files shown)


kevinrushforth:
Can you add an automated test to cover this fix?

Worth noting, this won't fully solve the problem that the reporter of this bug filed. The null check will prevent the NPE, but the scene will still not be visible. To allow removing and then re-adding the `JFXPanel`, we would need to not call `Window.hide` when the `JFXPanel` is removed. We should file a follow-on Enhancement to consider doing this, but that will need more discussion. The main point that would need to be solved is to figure out when to call `Window.hide` if not when the `JFXPanel` is  removed.

kevinrushforth:
@prsadhuk One more question: Do you know what changed in JavaFX 21 to trigger the NPE?

andy-goryachev-oracle:
Since this problem straddles two different toolkits, I would suggest to specify a minimum of 2 reviewers.

prsadhuk:
> To allow removing and then re-adding the `JFXPanel`, we would need to not call `Window.hide` when the `JFXPanel` is removed.

I am not sure if this is a problem as just as `JFXPanel.removeNotify` does `Window.hide` but at the same time when we add JFXPanel to frame, it calls `JFXPanel.addNotify` which does `Window.show` so tit gets hidden when JFXPanel is removed and shown when JFXPanel is added

prsadhuk:
> @prsadhuk One more question: Do you know what changed in JavaFX 21 to trigger the NPE?

It is not reproducible in jfx21 as per my testing...The problematic code to call updateSceneState was added for [JDK-8274932](https://bugs.openjdk.org/browse/JDK-8274932) so it has regressed from jfx22

prsadhuk:
Check is not without precedence in this file..There are instance of  `assert host != null;` and `host != null` check like below
https://github.com/openjdk/jfx/blob/6a586b662592be3eb81670f0c5ce48061c2fc07c/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/EmbeddedScene.java#L185

prsadhuk:
Automated test added

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/EmbeddedScene.java
```diff
@@ -157,7 +157,9 @@ public void setPixelScaleFactors(float scalex, float scaley) {
         entireSceneNeedsRepaint();
         Platform.runLater(() -> {
             QuantumToolkit.runWithRenderLock(() -> {
-                updateSceneState();
+                if (host != null) {
```
andy-goryachev-oracle, line 160: I think we need a different solution here, as mentioned in line 58

```
// TODO: synchronize access to embedder from ET and RT
```

andy-goryachev-oracle:
Kevin is right, this fix does not solve the issue mentioned in the ticket.  Once the fxPanel is added back, its content is not visible.  Does not matter whether removing/adding happens at startup or the button event handler.

(attaching a slightly modified test case to the ticket, notice lines 30 and 44.

Also, I think swing requires validate() and repaint() called after modifying the component's children.

tests/system/src/test/java/test/javafx/embed/swing/JFXPanelNPETest.java
```diff
@@ -0,0 +1,120 @@
...
+
+        Platform.runLater(() -> {
+            Scene scene = new Scene(new Button("Testbutton"));
+            fxPanel.setScene(scene);
+            Thread.currentThread().setUncaughtExceptionHandler(this);
+            SwingUtilities.invokeLater(() -> {
+                // fxPanel removed from frame
+                frame.remove(fxPanel);
+                // fxPanel added to frame again
+                frame.add(fxPanel); // <-- leads to NullPointerException
```
Maran23, line 106: You can use JUnit5 to write this test (class) and use `assertDoesNotThrow(() -> frame.add(fxPanel));`, which makes this a bit better to understand

prsadhuk:
> Kevin is right, this fix does not solve the issue mentioned in the ticket. Once the fxPanel is added back, its content is not visible. Does not matter whether removing/adding happens at startup or the button event handler.
> 
> (attaching a slightly modified test case to the ticket, notice lines 30 and 44.
> 
> Also, I think swing requires validate() and repaint() called after modifying the component's children.

OK..I guess I probably misunderstood the expectation...it says
```
EXPECTED -
The JFXPanel is visible to the user of the application and no Exceptions are thrown.
ACTUAL -
The JFXPanel is visible but the following Exception is thrown
```
which I deciphered as the JFXPanel window being visible and I guess in original testcase execution, the "TestButton" was still visible, only NPE was thrown..

ANyway, I guess it was mentioned that "We should file a follow-on Enhancement to consider doing this" and regarding synchro mentioned at line58, I guess we already have `QuantumToolkit.runWithRenderLock` in the problematic code

andy-goryachev-oracle:
> QuantumToolkit.runWithRenderLock

locking a class field may not be sufficient.  or rather, cannot be sufficient when multiple threads are accessing the field, as there are no guarantees as to the order of invocations.

Instead, we might need to pass the reference as a variable to the lambda which executes in a different thread.  This way no matter what happens to the field, the code that executes in a different context will always have the right object.

And, it might be better not to share the field between the threads and toolkits.  Always mutate a particular field in the context of one toolkit - like fieldAWT / fieldFX.  

Just a suggestion.

PR #1506, opened 2024-07-15
8335630: Crash if Platform::exit called with fullScreen Stage on macOS 14

kevinrushforth:
This PR solves two related bugs that are caused by events being delivered while the FX runtime is in the process of shutting down. These bugs are related enough that I think they need to be addressed at the same time. While debugging and developing the test, I saw one or the other or both in various test runs. The system test included with this program verifies that there is no crash and no exception, so will catch either failure mode.

This can happen when there is a full-screen window showing during shutdown on macOS, since full screen enter / exit uses a nested event loop that affects the order of operation. This could theoretically happen in other cases, thus the fix does not involve checking whether we are in full-screen or whether there is a nested event loop.

The fix does the following:

1. `GlassWindow+Overrides.m`: Check that the JVM is still running before calling the Java notification method in `windowDidResignKey` and `windowShouldClose`. This matches what is already done for other similar methods in this class. This is the fix for JDK-8335630.
2. `Window.java` and `View.java`: Check whether the Application instance is null in the event notification callbacks, and skip handling the event if it is. This is the fix for JDK-8299738.

Note that the fix for 2 is in platform-independent code. This should be fine, since we would want to skip the event handling callback on any other platform if it were done during shutdown, in order to avoid the ISE.

I have included a system test that passes on all platforms. On macOS, the test will fail without the fix and pass with the fix.

Details of the fix are below:

`Platform::exit` is called by the application to shutdown the JavaFX toolkit. The implementation of `Platform::exit` calls `Toolkit::exit` on the JavaFX Application thread, which in turn calls glass `Application::terminate` to shut it down. `Application::terminate` will close all the native windows, and call `finishTerminating` to terminate the native event loop and detach the thread from the JVM. This is asynchronous, at least on macOS, and will schedule the termination by posting a native event to the event loop.

on macOS 13, and sometimes on macOS 14, a Window or View event handler can be called between `Toolkit::exit`/`Application::terminate` (which sets the Toolkit's cached thread to null and the glass Application instance to null) and the JVM shutdown. This will make a JNI upcall to the appropriate Window or View `handleXxxxx` method, which will deliver it to the registered event handler. Because the Toolkit thread has been set to null, the "are we on the JavaFX Application thread" check will fail and it will throw an ISE. Since it can't usefully do anything anyway, we just want to skip handling it.

On macOS 14, `windowDidResignKey` can be called after the JVM is shutdown, causing an assertion print statement and then a crash when it tries to call a JNI method using a NULL `env` pointer. We need to use the `GET_MAIN_JENV_NOWARN` flavor of the macro to get the JNI `env`, and then check `env` for non-null.

NOTE: I intend to backport this to the `jfx23` stabilization branch for JavaFX 23 after it is integrated into mainline.

/issue add 8299738
/reviewers 2

---------
### Progress

### Issues
 * [JDK-8335630](https://bugs.openjdk.org/browse/JDK-8335630): Crash if Platform::exit called with fullScreen Stage on macOS 14 (**Bug** - P3)
 * [JDK-8299738](https://bugs.openjdk.org/browse/JDK-8299738): ISE if Platform::exit called with fullScreen Stage on macOS 13 (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1506/head:pull/1506` \
`$ git checkout pull/1506`

`$ git checkout pull/1506` \
`$ git pull https://git.openjdk.org/jfx.git pull/1506/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1506`

`$ git pr show -t 1506`

</details>

<a href="https://git.openjdk.org/jfx/pull/1506.diff">https://git.openjdk.org/jfx/pull/1506.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1506#issuecomment-2229251819)


modules/javafx.graphics/src/main/java/com/sun/glass/ui/View.java
```diff
@@ -528,17 +528,26 @@ public void setEventHandler(EventHandler eventHandler) {
         this.eventHandler = eventHandler;
     }
 
+    private boolean shouldHandleEvent() {
+        // Don't send any more events if the application has shutdown
+        if (Application.GetApplication() == null) {
```
andy-goryachev-oracle, line 533: Should we declare `Application.application` field as volatile?

I see interleaved multi-threaded access pattern:

```
thread=JavaFX-Launcher
CREATE thread=JavaFX-Launcher
thread=JavaFX-Launcher
thread=Thread-2
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX-Launcher
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX-Launcher
thread=JavaFX Application Thread
thread=PulseTimer-CVDisplayLink thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=QuantumRenderer-0
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=PulseTimer-CVDisplayLink thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX-Launcher
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX-Launcher
thread=JavaFX-Launcher
thread=JavaFX-Launcher
thread=QuantumRenderer-0
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
thread=JavaFX Application Thread
FINISH thread=JavaFX Application Thread
```

It might be ok though, since both creation and setting to null is done by the FX App thread.

andy-goryachev-oracle (approved):
The fix produces no exception and no crash on macOS 14.5 running on M1.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/View.java
mstr2: Other threads may still not see the current value of the `Application.application` field. Is it safe for other threads to use the `Application` instance after the field has been set to `null`?

kevinrushforth: This is unrelated to the current bug, since all accesses are happening on the FX app thread, so I'd like to file a follow-up to look into this.

To answer the question, there might be a potential problem with other threads not seeing that the `Application.application` field has been set to null.

Other threads do not access the glass Application directly. Most methods requires the app to be on the application thread, and will correctly detect and fail whether they see the old or new value of `Application.application`, so those operations will be fine.

One scenario that should be examined is `Platform.runLater`, which could call into glass and see the old value of application. There are related fields in `Tooklit` (e.g., the `fxUserThread` variable) that ought to be examined at the same time.

andy-goryachev-oracle: > since all accesses are happening on the FX app thread

Not quite right: the following threads access the field, apart from JavaFX Application Thread:

```
thread=JavaFX-Launcher
thread=Thread-2
thread=PulseTimer-CVDisplayLink thread
thread=QuantumRenderer-0
```

The output above logs thread names that access `Application.application` field, running the reproducer attached to both tickets.

kevinrushforth: To clarify, I'm talking about the accesses that have anything to do with this bug, and specifically, the methods that I modified. Those are all on the FX application thread. Unrelated to this bug, there are plenty of calls to Platform.runLater on other threads, which will indeed access the `Application.application` field from that other thread.

kevinrushforth: I filed [JDK-8336476](https://bugs.openjdk.org/browse/JDK-8336476) as a follow-up bug to address the race condition.

mstr2 (approved):
The fix looks good for the intended single-threaded usage.

PR #1519, opened 2024-07-28
8301761: The sorting of the SortedList can become invalid

LoayGhreeb:
Fix an issue in `SortedList` where the sorting became incorrect when adding new items that are equal to existing items according to the comparator. The `SortedList` should consider the insertion index of these items to maintain the correct order.

---------
### Progress

### Issue
 * [JDK-8301761](https://bugs.openjdk.org/browse/JDK-8301761): The sorting of the SortedList can become invalid (**Bug** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1519/head:pull/1519` \
`$ git checkout pull/1519`

`$ git checkout pull/1519` \
`$ git pull https://git.openjdk.org/jfx.git pull/1519/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1519`

`$ git pr show -t 1519`

</details>

<a href="https://git.openjdk.org/jfx/pull/1519.diff">https://git.openjdk.org/jfx/pull/1519.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1519#issuecomment-2276589532)


hjohn:
The PR title "The sorting of the SortedList can become invalid" seems to imply that the sorting can become incorrect, as in the incorrect order.

This is not the case. The sorting is still correct. Claiming that the sorting is invalid because a newly added item was not placed in a specific location relative to other equal items is a bit of misrepresentation.

So, if this PR wants to move forward, I think we first need to decide if we want to guarantee a specific behavior when inserting new but equal elements.  If we do, then it should be documented, and also maintained when the list gets resorted (ie. resorts must be stable).

LoayGhreeb:
> This is not the case. The sorting is still correct. Claiming that the sorting is invalid because a newly added item was not placed in a specific location relative to other equal items is a bit of misrepresentation.
> 
> So, if this PR wants to move forward, I think we first need to decide if we want to guarantee a specific behavior when inserting new but equal elements.

The issue isn’t just with adding new elements. It also affects the `TableView` that populated with items from a `SortedList`, The `SortedList`'s comparator is bound to the `TableView`'s comparator.
When the table is sorted by multiple columns, removing a column from the sort order should reorder the items. However, currently, removing any column from the sort order doesn't change the items order until all sorting columns are removed. At that point, the items returned to the default insertion order.

For example:
image

1. The initial order is [0, 1, 2, 3] (index column).
2. Sort by `Col1`, the order will be: [1, 0, 2, 3].
3. Add a second sort (Shift + Click) on `Col2`, the order will be: [1, 2, 3, 0].
4. Shift + Click again on `Col2` to sort in descending order: [1, 0, 3, 2].
5. Shift + Click again on `Col2` to remove the sort from `Col2`. The order will stay the same [1, 0, 3, 2], but the expected order is [1, 0, 2, 3] as in step 2.

<details>
<summary> Code </summary>

```java
import javafx.beans.property.ReadOnlyObjectWrapper;
import javafx.collections.transformation.SortedList;
import javafx.scene.Scene;
import javafx.stage.Stage;

public class Main extends Application {

    @Override
        primaryStage.setScene(new Scene(createTableView()));
        primaryStage.show();
    }

    private TableView<Data> createTableView() {
        TableView<Data> tableView = new TableView<>();

        TableColumn<Data, Integer> indexColumn = new TableColumn<>("#");
        indexColumn.setCellValueFactory(cell -> new ReadOnlyObjectWrapper<>(cell.getValue().index()));
        indexColumn.setSortable(false);

        TableColumn<Data, Integer> column1 = new TableColumn<>("Col1");
        column1.setCellValueFactory(cell -> new ReadOnlyObjectWrapper<>(cell.getValue().val1()));

        TableColumn<Data, Integer> column2 = new TableColumn<>("Col2");
        column2.setCellValueFactory(cell -> new ReadOnlyObjectWrapper<>(cell.getValue().val2()));

        tableView.getColumns().addAll(indexColumn, column1, column2);

        ObservableList<Data> data = FXCollections.observableArrayList(
                new Data(0, 5, 4),
                new Data(1, 1, 7),
                new Data(2, 5, 2),
                new Data(3, 5, 3)
        );

        SortedList<Data> sortedData = new SortedList<>(data);

        sortedData.comparatorProperty().bind(tableView.comparatorProperty());
        tableView.setItems(sortedData);
        return tableView;
    }

    public record Data(int index, int val1, int val2) {
    }
}
```

</details>

hjohn:
> Shift + Click again on Col2 to remove the sort from Col2. The order will stay the same [1, 0, 3, 2], but the expected order is [1, 0, 2, 3] as in step 2.

Why is that the expected order?  It is "a" possible, but valid order.  When only sorting on `Col1` then in my opinion [1, 0, 2, 3], [1, 0, 3, 2], [1, 2, 0, 3], etc.. are all valid sort orders.

If we want to distinguish between equal elements to give them a fixed sub-ordering (and the inverse one I suppose when the sort is descending) then what should the distinguishing factor be?  Its position in the underlying unsorted list?  The time it was added to the underlying list?  They need not be the same.  Using the position in the underlying list could be quite random if the source of the list doesn't guarantee a fixed position (from an unsorted database query for example).

What if the underlying list is another sorted list or a filtered list, can we even rely on the indices those provide?

nlisker:
@hjohn I was confused as well, but the [`TableView` docs](https://openjfx.io/javadoc/22/javafx.controls/javafx/scene/control/TableView.html#sorting-heading) say:
> Starting with JavaFX 8.0 (and the introduction of [SortedList](https://openjfx.io/javadoc/22/javafx.base/javafx/collections/transformation/SortedList.html)), it is now possible to have the collection return to the unsorted state when there are no columns as part of the TableView [sort order](https://openjfx.io/javadoc/22/javafx.controls/javafx/scene/control/TableView.html#getSortOrder()). To do this, you must create a SortedList instance, and bind its [comparator](https://openjfx.io/javadoc/22/javafx.base/javafx/collections/transformation/SortedList.html#comparatorProperty()) property to the TableView [comparator](https://openjfx.io/javadoc/22/javafx.controls/javafx/scene/control/TableView.html#comparatorProperty()) property

So I think that the confusion stems from the fuzzy guarantee that removing the column-based sorting will return the table to *the* unsorted state before that sorting (and not to *a* unsorted state). That might mean that `TableView` will have to maintain the initial sorting somewhere. Or... it's a bad documentation :)

I still don't see how this is a bug in `SortedList`. It could be a bug in `TableView`.

koppor:
TL;DR: This change is covered by an additional test. All other tests remain intact. Thus, this change does not break anything; it merely introduces a new feature.

> > Shift + Click again on Col2 to remove the sort from Col2. The order will stay the same [1, 0, 3, 2], but the expected order is [1, 0, 2, 3] as in step 2.
> 
> Why is that the expected order?

In the role of a user, I persieve removing an additional sort order as undo operation. 

Applying that thought to the example: "undo" is the reverse operation of a "do" operation. Meaning: First, Col2 should be considered as additional sorting. Then, it should not be considered as additional sorting. 

Applying that thought to a longer table: I as user have a dataset with n columns. If the dataset is in [1NF](https://en.wikipedia.org/wiki/First_normal_form) only, there might be many columns having the same values. If I as user begin to sort starting from column 1, then column 2, then column 3. Then I see: Oh, wait, no, I want to skip that column and sort by column 4 instead of 3. Then I undo the sort by column 3 and continue with 4. -- If the sort is not reverted to the previous state, but some other state, I need to start from the beginning. For two columnsd, this "annoyance" might be OK, but what for 10 columns? If I made a sorting mistake at column 9, I really need to redo all 8 columns before?

> If we want to distinguish between equal elements to give them a fixed sub-ordering (and the inverse one I suppose when the sort is descending) then what should the distinguishing factor be? Its position in the underlying unsorted list? The time it was added to the underlying list? They need not be the same. Using the position in the underlying list could be quite random if the source of the list doesn't guarantee a fixed position (from an unsorted database query for example).

I like the position in the list. At the time of "materialization" of the list, the position of an element is known - indepenent of it was retrieved by a SQL query or other sources.

Since this change "only" narrows the order, does not contradict other expectations, and really helps our desktop application to be more usable, I am strongly in favour to use the index as second criterion. -- An indication for no-change-for-others is that there was only an additional test case needed - and all existing test cases were unmodified.

> What if the underlying list is another sorted list or a filtered list, can we even rely on the indices those provide?

It would be even natural if the index change, the result of the SortedList also changes. I think, this is OK.

PR #1520, opened 2024-07-29
8337400: [Linux] Initial window position is not centered on Ubuntu 24.04 / Xorg

tsayao:
On Ubuntu 24.04 with Xorg windows are not shown centered.

This is due a configure event being received with position 0,0 which is wrong.

The fix ignores the configure events until the window is mapped.

---------
### Progress

### Issue
 * [JDK-8337400](https://bugs.openjdk.org/browse/JDK-8337400): [Linux] Initial window position is not centered on Ubuntu 24.04 / Xorg (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1520/head:pull/1520` \
`$ git checkout pull/1520`

`$ git checkout pull/1520` \
`$ git pull https://git.openjdk.org/jfx.git pull/1520/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1520`

`$ git pr show -t 1520`

</details>

<a href="https://git.openjdk.org/jfx/pull/1520.diff">https://git.openjdk.org/jfx/pull/1520.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1520#issuecomment-2256563228)

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/gtk/GlassApplication.cpp
+++ b/modules/javafx.graphics/src/main/native-glass/gtk/GlassApplication.cpp
@@ -545,6 +545,7 @@ static void process_events(GdkEvent* event, gpointer data)
                     process_dnd_target(ctx, &event->dnd);
                     break;
                 case GDK_MAP:
+                    ctx->process_map();
                     // fall-through
                 case GDK_UNMAP:
                 case GDK_CLIENT_EVENT:
--- a/modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
+++ b/modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
@@ -731,7 +731,8 @@ WindowContextTop::WindowContextTop(jobject _jwindow, WindowContext* _owner, long
             geometry(),
             resizable(),
             on_top(false),
-            is_fullscreen(false) {
+            is_fullscreen(false),
+            map_received(false) {
     jwindow = mainEnv->NewGlobalRef(_jwindow);
     gdk_windowManagerFunctions = wmf;
 
@@ -959,6 +960,10 @@ void WindowContextTop::process_state(GdkEventWindowState* event) {
     WindowContextBase::process_state(event);
 }
 
+void WindowContextTop::process_map() {
+    map_received = true;
+}
+
 void WindowContextTop::process_realize() {
     gdk_window = gtk_widget_get_window(gtk_widget);
     if (frame_type == TITLED) {
@@ -975,6 +980,10 @@ void WindowContextTop::process_realize() {
 }
 
 void WindowContextTop::process_configure(GdkEventConfigure* event) {
+    if (!map_received) {
+        return;
+    }
+
     int ww = event->width + geometry.extents.left + geometry.extents.right;
     int wh = event->height + geometry.extents.top + geometry.extents.bottom;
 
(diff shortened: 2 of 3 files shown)


kevinrushforth:
I've noticed a few Ubuntu 24.04 failures both on Xorg and on Wayland. I'll test them with this patch.

/reviewers 2

lukostyra (approved):
LGTM, fixes the problem on Ubuntu 24.04. Noticed no regressions otherwise.

I did see some new test failures on 24.04, but these don't go away after this patch. Must be some different reason for them.

kevinrushforth (commented):
This causes a regression on Ubuntu 22.04. The following tests fail with this fix (they all pass with current master), when using X.org or Wayland:

```
SizeToSceneTest > testInitialSizeOnFullscreenThenSizeToScene() FAILED
    org.opentest4j.AssertionFailedError: 360.0 >= 1870.0 ==> expected: <true> but was: <false>

SizeToSceneTest > testInitialSizeOnMaximizedThenSizeToScene() FAILED
    org.opentest4j.AssertionFailedError: 360.0 >= 1870.0 ==> expected: <true> but was: <false>

SizeToSceneTest > testInitialSizeOnSizeToSceneThenMaximized() FAILED
    org.opentest4j.AssertionFailedError: 360.0 >= 1870.0 ==> expected: <true> but was: <false>
```

On Ubuntu 24.04 there are also some new failures in `SizeToSceneTest ` with this PR, but several of the test fail anyway on 24.04

tsayao:
The problem is a configure event is telling that the window is at 0,0 (that's wrong). I will investigate further way this event is being sent and find a better solution.

tsayao:
> This causes a regression on Ubuntu 22.04. The following tests fail with this fix (they all pass with current master), when using X.org or Wayland:
> 
> ```
> SizeToSceneTest > testInitialSizeOnFullscreenThenSizeToScene() FAILED
>     org.opentest4j.AssertionFailedError: 360.0 >= 1870.0 ==> expected: <true> but was: <false>
> 
> SizeToSceneTest > testInitialSizeOnMaximizedThenSizeToScene() FAILED
>     org.opentest4j.AssertionFailedError: 360.0 >= 1870.0 ==> expected: <true> but was: <false>
> 
> SizeToSceneTest > testInitialSizeOnSizeToSceneThenMaximized() FAILED
>     org.opentest4j.AssertionFailedError: 360.0 >= 1870.0 ==> expected: <true> but was: <false>
> ```
> 
> On Ubuntu 24.04 there are also some new failures in `SizeToSceneTest ` with this PR, but several of the test fail anyway on 24.04

It always failing for me, with this PR or without.

Maran23:
> It always failing for me, with this PR or without.

Since I wrote them, maybe we can fix those tests for you as well and make them more stable across Linux.
I think generally those tests are good to have as they can catch serious issues when it comes to stage showing and sizing. But they should be as stable as they can be.

tsayao:
The problem is that on Ubuntu 24.04 with Xorg (aparrently does not affect XWayland - but not sure if it's true on all scenarios) there's an extra configure event that causes problems. For example, if the window is set to begin fullscreen it will go back to floating (because of this event) and the size test will fail.

It's something non-obvious that is causing this, problably.

tsayao:
This is probably a Mutter bug as it is reproducible on Mutter/Xorg with a Gtk C program and is not reproducible on i3.

Gtk is receiving a wrong configure event that causes this. It also causes the failing `SizeToSceneTest` and some other bugs such as fullscreen or maximized windows going back to floating.

PR #1521, opened 2024-07-29
8320232: Cells duplicated when table collapsed and expanded

Maran23:
This PR fixes a bug where a `TableView` inside a `TitledPane` may 'duplicate' the cells. This actually has nothing to do with the `TitledPane`, but it triggered the bug easily due to its nature to change the size of his content when collapsing.

The expansion change of a `TitledPane` triggered an event where the underlying `VirtualFlow` was adding cells to the pile (for reuse) and later cleaning them all up without resetting the index to -1. 
This led to a bug where two cells had the same index and therefore received an edit event, although just one should (and is visible, the other cell has no parent -> orphan node).

The fix is to always reset the index to -1. This was already done before, just not everywhere, for all cells.
So before we clear the pile (or cells), we always reset the index to -1.
Added a bunch of tests, some pass before and after.

---------
### Progress

### Issue
 * [JDK-8320232](https://bugs.openjdk.org/browse/JDK-8320232): Cells duplicated when table collapsed and expanded (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1521/head:pull/1521` \
`$ git checkout pull/1521`

`$ git checkout pull/1521` \
`$ git pull https://git.openjdk.org/jfx.git pull/1521/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1521`

`$ git pr show -t 1521`

</details>

<a href="https://git.openjdk.org/jfx/pull/1521.diff">https://git.openjdk.org/jfx/pull/1521.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1521#issuecomment-2256668585)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
@@ -805,6 +805,9 @@ public class VirtualFlow<T extends IndexedCell> extends Region {
         if (vertical == null) {
             vertical = new BooleanPropertyBase(true) {
                 @Override protected void invalidated() {
+                    resetIndex(cells);
+                    resetIndex(pile);
+
                     pile.clear();
                     sheetChildren.clear();
                     cells.clear();
@@ -1072,9 +1075,10 @@ public class VirtualFlow<T extends IndexedCell> extends Region {
             lastHeight = -1;
             releaseCell(accumCell);
             sheet.getChildren().clear();
-            for (int i = 0, max = cells.size(); i < max; i++) {
-                cells.get(i).updateIndex(-1);
-            }
+
+            resetIndex(cells);
+            resetIndex(pile);
+
             cells.clear();
             pile.clear();
             releaseAllPrivateCells();
@@ -1082,9 +1086,9 @@ public class VirtualFlow<T extends IndexedCell> extends Region {
             lastWidth = -1;
             lastHeight = -1;
             releaseCell(accumCell);
-            for (int i = 0, max = cells.size(); i < max; i++) {
-                cells.get(i).updateIndex(-1);
-            }
+
+            resetIndex(cells);
+
             addAllToPile();
             releaseAllPrivateCells();
         } else if (needsReconfigureCells) {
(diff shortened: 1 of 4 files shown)


andy-goryachev-oracle:
Does this change affect `TreeView`?

Maran23:
> Does this change affect `TreeView`?

All components that use the `VirtualFlow`,but it might be more noticeable for some than others. In fact, I first noticed this behavior  in `Dialog`s. I will add a reproducer for that one ticket as well soon, so it contains more information for people that may dig the backlog or git history in the future.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -805,6 +805,9 @@ public final BooleanProperty verticalProperty() {
         if (vertical == null) {
             vertical = new BooleanPropertyBase(true) {
                 @Override protected void invalidated() {
+                    resetIndex(cells);
```
andy-goryachev-oracle, line 808: minor: [cells|pile]clear() can be combined with resetIndex(), something like

```
    private void clear(ArrayLinkedList<T> cells) {
        for (T cell : cells) {
            cell.updateIndex(-1);
        }
        cells.clear();
    }
```

as these operations seem to be called in pairs

andy-goryachev-oracle (commented):
This is a definitely the right fix.

The TitledPane reproducer fails without and works correctly with the change.  I don't see the duplicate startEdit() with the Dialog reproducer, please clarify the steps to reproduce and what is expected/observed.

Left a couple of minor comments, will re-approve if you choose to make changes.

andy-goryachev-oracle (approved):
As with everything involving VirtualFlow, I'd like a second reviewer.  Maybe @johanvos ?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
Maran23: Unfortunately not in the `needsRebuildCells` if condition, where we only reset the index and add all cells to the pile

andy-goryachev-oracle: you are right.

andy-goryachev-oracle:
I see the issue with the dialog reproducer on Windows 11.  macOS does not exhibit the issue with the dialog one, but does have an issue with the titled pane reproducer.

This change fixes the bug on macOS and Windows 11.

kevinrushforth (approved):
Looks good. I confirm that 3 of the newly-added test methods fail without the fix and pass with the fix.

I agree with Andy about the unrelated change. I'll reapprove if you fix it.

Maran23:
/integrate

I agree the minor change is unrelated, but since it is that small for a test I had written in the past, I decided to keep it.

PR #1532, opened 2024-08-08
8333919: dragViewOffsetX/dragViewOffsetY are ignored for the dragView image

lukostyra:
When fixing [JDK-8233955](https://bugs.openjdk.org/browse/JDK-8233955) offset calculation was left out, which made Dragboard offset API not work on macOS.

This change fixes this mistake. Now drag image should be properly offset.

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassViewDelegate.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassViewDelegate.m
@@ -1021,7 +1021,7 @@ static jint getSwipeDirFromEvent(NSEvent *theEvent)
 
         if (image != nil)
         {
-            DNDLOG("created an image with dim %dx%d", [image size].width, [image size].height);
+            DNDLOG("created an image with dim %fx%f", [image size].width, [image size].height);
             // check the drag image size and scale it down as needed using Safari behavior (sizes) as reference
             rect.size.width = [image size].width;
             rect.size.height = [image size].height;
@@ -1066,6 +1066,9 @@ static jint getSwipeDirFromEvent(NSEvent *theEvent)
                 if (offset.y > imageHalfY || offset.y < -imageHalfY) {
                     offset.y = imageHalfY * (offset.y > 0 ? 1 : -1);
                 }
+
+                dragPoint.x += offset.x;
+                dragPoint.y += offset.y;
             }
         }
         else


lukostyra:
I have one more question to this, as I don't have the best way to test this. I'm not sure how offsets should be interpreted and if they are interpreted the right way.

At this point:
* For X offset: positive values shift image RIGHT (mouse cursor is closer to the LEFT side of drag image), negative shift LEFT (mouse cursor is closer to the RIGHT side of drag image)
* For Y offset: positive values shift image DOWN (mouse cursor is closer to the TOP of the drag image), negative values shift UP (mouse cursor is closer to the BOTTOM of the drag image)

Let me know if this behavior is desired and if it matches pre-[JDK-8233955](https://bugs.openjdk.org/browse/JDK-8233955) behavior.

lukostyra:
I did not realize that Windows/macOS behavior was matched in this regard, so I'll use Windows as reference for further testing. Thanks!

I agree offsets should be flipped (with `-=` rather than `+=`) so that is working well. I also removed the centering of the drag image, so now both shift directions and offset origin match Windows.

However, from my initial testing, it seems like macOS prefers the mouse cursor to be always within the drag image bounds. As an example, `DndTest.java` defined the Y offset equal to -30. While the drag image is initially offset correctly, macOS immediately "snaps" it back so that the mouse cursor stays within the drag image bounds. This seems to be a platform-specific behavior though (Windows keeps the offset as we define it). I'll do some more experiments and see if that can be remedied somehow, but for now I doubt it.

kevinrushforth:
Reviewers: @jperedadnr @kevinrushforth 

/reviewers 2

modules/javafx.graphics/src/main/native-glass/mac/GlassViewDelegate.m
```diff
@@ -1046,12 +1046,9 @@ - (void)startDrag:(NSDragOperation)operation withItems:(NSArray<NSDraggingItem*>
 
         if (image != nil)
         {
-            // select the center of the image as the drag origin
```
jperedadnr, line 1049: The old comment about RT-17629 (https://bugs.openjdk.org/browse/JDK-8091965) does not apply anymore: after this PR, the offset, which comes from the Java layer, does adjust the drag image origin.

So I believe we can remove the whole comment as well (and possibly close RT-17629).

modules/javafx.graphics/src/main/native-glass/mac/GlassViewDelegate.m
```diff
@@ -1066,6 +1063,9 @@ - (void)startDrag:(NSDragOperation)operation withItems:(NSArray<NSDraggingItem*>
                 if (offset.y > imageHalfY || offset.y < -imageHalfY) {
                     offset.y = imageHalfY * (offset.y > 0 ? 1 : -1);
                 }
+
+                dragPoint.x -= offset.x;
+                dragPoint.y -= offset.y;
```
jperedadnr, line 1068: This works fine now, _if_ the offset values are smaller than the imageHalf X,Y values, which doesn't make sense anymore:

- If you drag from {0,0} (top-left), it works from that position
- But if you drag from {imageW, imageH} (bottom-right), the offset is clamped to the image centre, which it is not expected.

So now we need to change the above lines to clamp the offset between 0,0 and imageW,H (though Windows doesn't clamp the offset at all).

lukostyra: Out of curiosity I removed offset clamping (macOS snaps the drag image regardless) and technically it still works, but when you cancel the drag the image snaps back to the offset position. With high enough (or negative) values drag image can animate back even outside the application bounds, which makes little sense and looks weird. I'll clamp it between 0,0 and imageW,H as you mentioned.

I also didn't find a way to disable above drag image "snapping". However, if it makes sense to clamp it on macOS, maybe it would also make sense to introduce similar limits to offsets on other platforms (probably as a separate issue)?

jperedadnr: Right, there is a built-in native animation from the offset point to the mouse cursor (in and out). 

When the drag event is cancelled, the image returns to the offset location with this animation. We still want that, so I wouldn't remove the animation (if we could), but of course, with the original position of the source, no far away from it, so it makes sense to clamp between the image bounds. 

And doing the same (clamping) for Windows would be a separate issue, indeed.

jperedadnr: By the way, this is how you can disable the cancel animation:

```
diff --git a/modules/javafx.graphics/src/main/native-glass/mac/GlassDraggingSource.m b/modules/javafx.graphics/src/main/native-glass/mac/GlassDraggingSource.m
index 50a79cabf6..373537f24e 100644
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassDraggingSource.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassDraggingSource.m
@@ -40,6 +40,7 @@
 
 - (NSDragOperation)draggingSession:(NSDraggingSession *)session sourceOperationMaskForDraggingContext:(NSDraggingContext)context
 {
+     session.animatesToStartingPositionsOnCancelOrFail = NO;
     // The Command key masks out every operation other than NSDragOperationGeneric. We want
     // internal Move events to get through this filtering so we copy the Move bit into the
     // Generic bit and treat Generic as a synonym for Move.
```

I'm still in favour of this native animation, as opposed to the Windows case that doesn't have it. This can be a separated issue as well (unifying the animation in all platforms).

modules/javafx.graphics/src/main/native-glass/mac/GlassViewDelegate.m
```diff
@@ -1046,26 +1046,26 @@ - (void)startDrag:(NSDragOperation)operation withItems:(NSArray<NSDraggingItem*>
 
         if (image != nil)
         {
-            // select the center of the image as the drag origin
```
jperedadnr, line 1049: There are two consecutive `if (image != nil)` blocks, so you can remove the 4 lines above this one.

jperedadnr (commented):
Looking good, I still have two minor comments

modules/javafx.graphics/src/main/native-glass/mac/GlassViewDelegate.m
andy-goryachev-oracle: I wanted to chime in, thought it might be unrelated:

In the context of a docking framework, it is impossible to disable the cancel animation.  We should have a way to disable it.  The use case is when the user drags and drops a tab/pane outside of the dockable windows.  In this case a new window should appear at the drop location, without the drag image animated back to the start position.

Another thing that is impossible is to control the drag image: sometimes one needs a scaled icon, sometimes one wants the full content.

These two issues basically prevent implementing a docking framework via DnD, one needs to use other solution, for example
https://github.com/andy-goryachev/FxDock

jperedadnr: I'm not sure if I follow, but about the cancel animation, do you mean that is not possible to disable it on macOS? Have you tested `session.animatesToStartingPositionsOnCancelOrFail`? 
As for the image, `DragBoard::setDragView` allows you setting any image that you need for the drag gesture, or am I missing something?

andy-goryachev-oracle: it's been a while - java8 time frame - back then I was unable to find a way to disable animation on cancel.

as for image - I was setting large image and fx tried to scale it to be smaller, I was unable to force it to show the unscaled full size image.

also, why does the drag image look fuzzy on mac retina?

lukostyra:
Also I noticed the PR title does not match the JBS issue title - missed the macOS tag while copying. Will update it to match.

andy-goryachev-oracle:
FYI, I've added Drag and Drop page to the Monkey Tester for testing various options:

andy-goryachev-oracle:
Question: how should we handle the negative offsets?

The way javadoc is currently phrased, "offsetX x position of the cursor **over** the image" implies that the offset is clamped to be inside of the (scaled) image.  If so, the behavior appears to be correct even in the presence of infinities and NaNs (the latter produces 0 offset).

andy-goryachev-oracle:
Another issue - out of scope for this PR - is the DRAG_DONE event missing any coordinates, regardless of whether it was dropped over the accepting target or non-accepting target or no target.  I think this was another reason I could not implement the docking framework via DnD.

```
{event=DRAG_DONE, screenX=0.0, screenY=0.0, sceneX=0.0, sceneY=0.0}
```

modules/javafx.graphics/src/main/native-glass/mac/GlassViewDelegate.m
```diff
@@ -1042,30 +1042,27 @@ - (void)startDrag:(NSDragOperation)operation withItems:(NSArray<NSDraggingItem*>
...
                 NSPoint offset = NSPointFromString(offsetString);
                 //Adjust offset to the image size
-                float imageHalfX = [image size].width/2.0f;
-                float imageHalfY = [image size].height/2.0f;
+                float imageW = [image size].width;
+                float imageH = [image size].height;
 
-                if (offset.x > imageHalfX || offset.x < -imageHalfX) {
-                    offset.x = imageHalfX * (offset.x > 0 ? 1 : -1);
+                if (offset.x < 0.0f) {
```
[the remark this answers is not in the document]
kevinrushforth: That seems out of scope for this PR. We typically don't do any special processing to deal with NaN unless there is a problem. If needed, maybe it could be done as part of the follow-up bug [JDK-8338204](https://bugs.openjdk.org/browse/JDK-8338204).

kevinrushforth (approved):
Looks good to me. Unifying the behavior on Windows and Linux can be done in the follow-up bug [JDK-8338204](https://bugs.openjdk.org/browse/JDK-8338204).

If needed, additional follow-up issues can be filed.

modules/javafx.graphics/src/main/native-glass/mac/GlassViewDelegate.m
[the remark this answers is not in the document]
andy-goryachev-oracle: Sounds good, especially since it seems to treat NaNs as 0's.

andy-goryachev-oracle (approved):

jperedadnr (approved):
Looks good

PR #1545, opened 2024-08-26
8339178: [macos] Swing InterOp Platform.exit() crash

kevinrushforth:
A Swing / FX interop app will crash if an application creates a new AWT / Swing window after calling Platform.exit. The root cause of the crash is that AWT caches the JNI env pointer for the AppKit thread, and assumes that it is valid for the life of the application. In the case where JavaFX is the owner of the NSApplication, we detach the AppKit thread from the JVM, after which AWT's env pointer is no longer valid. AWT will therefore crash the next time it does a JNI upcall.

This PR fixes the crash by leaving the macOS AppKit thread attached to the JVM after the JavaFX main event loop terminates. This requires attaching the AppKit thread to the JVM as a daemon thread when JavaFX is the NSApplication owner, matching what AWT does when it is the owner. In order to prevent a JavaFX application from exiting prematurely, create a non-daemon "KeepAlive" thread that can be terminated when the FX toolkit exits. This also solves a somewhat-related problem where the JavaFX toolkit will exit prematurely if the AWT toolkit is started first, and all AWT windows are disposed.

This fix is in addition to the AWT fix: [JDK-8190329](https://bugs.openjdk.org/browse/JDK-8190329) / openjdk/jdk#20688. Either the AWT fix or the JavaFX fix is sufficient to avoid this specific problem, but there is value in fixing it in both places, so I cloned the AWT bug to create a JavaFX bug that we can use for this PR.

Summary of the changes:

* Attach the AppKit thread to the JVM as a daemon
* Do not detach the thread when the FX main event loop terminates
* Create and start a KeepAlive thread in MacApplication, when the FX toolkit starts
* Terminate the KeepAlive thread when the FX toolkit finishes

Testing:

I created automated systems tests from the manual test programs in the bugs as well as a test to ensure that we don't regress and exit prematurely in the pure JavaFX case (which would happen without the KeepAlive thread). Two of the new tests fail on macOS without the fix and pass with the fix. All three pass on all platforms with the fix.

---------
### Progress

### Issues
 * [JDK-8339178](https://bugs.openjdk.org/browse/JDK-8339178): [macos] Swing InterOp Platform.exit() crash (**Bug** - P3)
 * [JDK-8339183](https://bugs.openjdk.org/browse/JDK-8339183): [macos] Premature exit in Swing interop when last JFrame is disposed (**Bug** - P4)

### Reviewers
 * [Prasanta Sadhukhan](https://openjdk.org/census#psadhukhan) (@prsadhuk - **Reviewer**)
 * [Alexander Zvegintsev](https://openjdk.org/census#azvegint) (@azvegint - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1545/head:pull/1545` \
`$ git checkout pull/1545`

`$ git checkout pull/1545` \
`$ git pull https://git.openjdk.org/jfx.git pull/1545/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1545`

`$ git pr show -t 1545`

</details>

<a href="https://git.openjdk.org/jfx/pull/1545.diff">https://git.openjdk.org/jfx/pull/1545.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1545#issuecomment-2316224131)

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacApplication.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacApplication.java
@@ -63,6 +63,28 @@ final class MacApplication extends Application implements InvokeLaterDispatcher.
     private boolean isTaskbarApplication = false;
     private final InvokeLaterDispatcher invokeLaterDispatcher;
 
+    private static final CountDownLatch keepAliveLatch = new CountDownLatch(1);
+
+    private static void startKeepAliveThread() {
+        // Start a non-daemon thread to keep the JavaFX toolkit alive
+        // This is necessary on macOS because the JavaFX Application thread
+        // is attached to the macOS Appkit Thread as a daemon thread
+        Thread thr = new Thread(() -> {
+            try {
+                keepAliveLatch.await();
+            } catch (InterruptedException ex) {
+                throw new RuntimeException("Unexpected exception: ", ex);
+            }
+        });
+        thr.setName("JavaFX-KeepAlive");
+        thr.setDaemon(false);
+        thr.start();
+    }
+
+    private static void finishKeepAliveThread() {
+        keepAliveLatch.countDown();
+    }
+
     MacApplication() {
         // Embedded in SWT, with shared event thread
         @SuppressWarnings("removal")
@@ -104,6 +126,11 @@ final class MacApplication extends Application implements InvokeLaterDispatcher.
             });
         isTaskbarApplication = tmp;
 
+        // Since both FX and AWT attaches to the AppKit thread as a daemon
+        // We need to create a non-Daemon KeepAlive thread so the FX toolkit
+        // doesn't exit prematurely
+        startKeepAliveThread();
+
         ClassLoader classLoader = MacApplication.class.getClassLoader();
         _runLoop(classLoader, wrappedRunnable, isTaskbarApplication);
     }
@@ -135,6 +162,7 @@ final class MacApplication extends Application implements InvokeLaterDispatcher.
     @Override
     protected void finishTerminating() {
         _finishTerminating();
+        finishKeepAliveThread();
 
         super.finishTerminating();
     }
--- a/modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacApplication.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacApplication.java
@@ -65,10 +65,19 @@ final class MacApplication extends Application implements InvokeLaterDispatcher.
 
     private static final CountDownLatch keepAliveLatch = new CountDownLatch(1);
 
+    /**
+     * Starts a non-daemon KeepAlive thread to ensure that the
+     * JavaFX toolkit keeps running until the toolkit exits. On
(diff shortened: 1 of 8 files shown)


modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacApplication.java
```diff
@@ -63,6 +63,28 @@ final class MacApplication extends Application implements InvokeLaterDispatcher.
...
+        Thread thr = new Thread(() -> {
+            try {
+                keepAliveLatch.await();
+            } catch (InterruptedException ex) {
+                throw new RuntimeException("Unexpected exception: ", ex);
+            }
+        });
+        thr.setName("JavaFX-KeepAlive");
+        thr.setDaemon(false);
+        thr.start();
```
andy-goryachev-oracle, line 81: is it possible to reuse some other thread for this purpose?
this may not be an issue when virtual threads take hold, but right now threads are precious native resource.

kevinrushforth: No. The whole reason we are doing this is that there are no other non-daemon threads. I updated the docs to make this more clear. Anyway, threads aren't _that_ scarce as long as you don't create hundreds of them.

Loosely related: This reminds me that `javafx.application.Application.launch` does create an unnecessary thread (on all platforms). There is a long-standing issue to get rid of it:

[JDK-8090323](https://bugs.openjdk.org/browse/JDK-8090323): Consider eliminating the JavaFX-Launcher Thread

It isn't high priority, but would be good cleanup.

andy-goryachev-oracle: thank you for clarification!
 
never underestimate the happiness with which the app and library developers create new threads...

kevinrushforth:
@prsadhuk Can you be one of the reviewers?

/reviewers 2

prsadhuk (approved):
Tested with and without AWT fix https://github.com/openjdk/jdk/pull/20688 and it looks good without crash..

tests/system/src/test/java/test/launchertest/TestAppPlatformExitAWT.java
```diff
@@ -0,0 +1,77 @@
...
+ * This is launched by MainLauncherTest.
+ */
+public class TestAppPlatformExitAWT extends Application {
+
+    private void createSwing() {
+        JDialog d = new JDialog();
+        Platform.runLater(()-> {
+            Platform.exit();
+        });
+        d.setVisible(true);
```
kevinrushforth, line 50: By code inspection, I realized that the `JDialog` might be shown before `Platform::exit` has been called, meaning that the test could pass even if the bug is present. I'll push a fix to move the setVisible inside the runLater lambda, using a `SwingUtilities::invokeLater`. This will ensure that it provokes the bug.

I ran the updated test and confirm that the test still fails without the fix and passes with the fix.

prsadhuk (approved):

azvegint (approved):

PR #1546, opened 2024-08-27
8339068: [Linux] NPE: Cannot read field "firstFont" because "<local4>" is null

jperedadnr:
This PR adds a couple of null checks to `LogicalFont` and `FTFactory`, that make use of `FontConfigManager::getFontConfigFont`, which under certain cases, can return null.

I've tested this PR on both Linux (Ubuntu 22.04.4) and Android, just using `-Dprism.useFontConfig=false`.

On Ubuntu, I've manually added some font files to the JDK path `$JAVA_HOME/lib/fonts` (which didn't exist), and the test passes now, though this message is still printed out:
```
Error: JavaFX detected no fonts! Please refer to release notes for proper font configuration
```
which is confusing to me, because fonts where found in the JDK path after all, and even in the case that there were no fonts found, "the release notes" is an ambiguous reference for the user. 

Also, instead of adding fonts to the JDK, I tested adding a `logicalfonts.properties` file with `-Dprism.fontdir` and without `fontConfig`, but this case was already working before the patch for this PR, and still works after it.

Note that if there are no fonts in the JDK path, and `prism.fontdir` is not provided, without `fontConfig`, after this PR, there will still be an exception:

```
Caused by: java.lang.NullPointerException: Cannot invoke "com.sun.javafx.font.FontResource.getDefaultAAMode()" because the return value of "com.sun.javafx.font.LogicalFont.getSlot0Resource()" is null
        at javafx.graphics@24-internal/com.sun.javafx.font.LogicalFont.getDefaultAAMode(LogicalFont.java:456)
        at javafx.graphics@24-internal/com.sun.javafx.font.LogicalFont.getStrike(LogicalFont.java:461)
        at javafx.graphics@24-internal/com.sun.javafx.font.PrismFont.getStrike(PrismFont.java:79)
        at javafx.graphics@24-internal/com.sun.javafx.text.PrismTextLayout.setContent(PrismTextLayout.java:139)
        at javafx.graphics@24-internal/javafx.scene.text.Text.getTextLayout(Text.java:303)
        at javafx.graphics@24-internal/javafx.scene.text.Text.needsFullTextLayout(Text.java:258)
        at javafx.graphics@24-internal/javafx.scene.text.Text$6.invalidated(Text.java:576)
 ...
```

because `LogicalFont::getSlot0Resource` will return null. Adding null checks after this point will require many changes (`LogicalFont::getSlotResource(0)` is used in many places). Shouldn't we fail gracefully after `LogicalFont::getSlot0Resource` if `slot0FontResource` is finally null?

---------
### Progress

### Issue
 * [JDK-8339068](https://bugs.openjdk.org/browse/JDK-8339068): [Linux] NPE: Cannot read field "firstFont" because "&lt;local4&gt;" is null (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1546/head:pull/1546` \
`$ git checkout pull/1546`

`$ git checkout pull/1546` \
`$ git pull https://git.openjdk.org/jfx.git pull/1546/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1546`

`$ git pr show -t 1546`

</details>

<a href="https://git.openjdk.org/jfx/pull/1546.diff">https://git.openjdk.org/jfx/pull/1546.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1546#issuecomment-2312945622)

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/font/freetype/FTFactory.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/font/freetype/FTFactory.java
@@ -133,11 +133,13 @@ public class FTFactory extends PrismFontFactory {
         boolean isItalic = primaryResource.isItalic();
         FontConfigManager.FcCompFont font =
             FontConfigManager.getFontConfigFont("sans", isBold, isItalic);
-        ArrayList<String> linkedFontFiles = FontConfigManager.getFileNames(font, false);
-        ArrayList<String> linkedFontNames = FontConfigManager.getFontNames(font, false);
         FontFallbackInfo info = new FontFallbackInfo();
-        for (int i=0; i<linkedFontNames.size(); i++)  {
-            info.add(linkedFontNames.get(i), linkedFontFiles.get(i), null);
+        if (font != null) {
+            ArrayList<String> linkedFontFiles = FontConfigManager.getFileNames(font, false);
+            ArrayList<String> linkedFontNames = FontConfigManager.getFontNames(font, false);
+            for (int i=0; i<linkedFontNames.size(); i++)  {
+                info.add(linkedFontNames.get(i), linkedFontFiles.get(i), null);
+            }
         }
         return info;
     }
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/font/freetype/FTFactory.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/font/freetype/FTFactory.java
@@ -27,6 +27,7 @@ package com.sun.javafx.font.freetype;
 
 import java.util.ArrayList;
 
+import com.sun.javafx.PlatformUtil;
 import com.sun.javafx.font.FontConfigManager;
 import com.sun.javafx.font.FontFallbackInfo;
 import com.sun.javafx.font.FontResource;
@@ -40,6 +41,7 @@ import com.sun.javafx.text.TextRun;
 public class FTFactory extends PrismFontFactory {
 
     static boolean LCD_SUPPORT;
+    static boolean isLinux = PlatformUtil.isLinux();
 
     public static PrismFontFactory getFactory() {
         PrismFontFactory factory = null;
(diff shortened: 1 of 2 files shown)


kevinrushforth:
@prrace Can you take a look at this?

hjohn:
The changes seem to deal with the case where Font Config is unavailable on Linux.  I think those changes are fine.  However, if it turns out there isn't any font at all (via an alternative source) I think it would be best to just fail.

modules/javafx.graphics/src/main/java/com/sun/javafx/font/freetype/FTFactory.java
```diff
@@ -133,11 +133,13 @@ public FontFallbackInfo getFallbacks(FontResource primaryResource) {
...
-        ArrayList<String> linkedFontNames = FontConfigManager.getFontNames(font, false);
         FontFallbackInfo info = new FontFallbackInfo();
-        for (int i=0; i<linkedFontNames.size(); i++)  {
-            info.add(linkedFontNames.get(i), linkedFontFiles.get(i), null);
+        if (font != null) {
+            ArrayList<String> linkedFontFiles = FontConfigManager.getFileNames(font, false);
+            ArrayList<String> linkedFontNames = FontConfigManager.getFontNames(font, false);
+            for (int i=0; i<linkedFontNames.size(); i++)  {
+                info.add(linkedFontNames.get(i), linkedFontFiles.get(i), null);
+            }
```
hjohn, line 142: Downstream code seems to be able to deal with empty `FontFallbackInfo`, so I guess this is also okay.

prrace:
I don't understand why this change is needed.

Perhaps I'm misunderstanding, but if usefontconfig is false, that means "I have supplied embedded fonts instead".
In which case the API calls where the null checks are added should not be returning null.

In other words, setting this to false, and not supplying fonts is a misconfiguration.
Arguably we should disable this property and always look for fontconfig.
But one would only know how to misconfigure it by reading the sources to discover this property.

jperedadnr:
On Linux, if you use `-Dprism.useFontConfig=false`, but then don't set valid embedded fonts, you get the reported NPE, instead of a different failure, with a more explicit message.
 
But since this property is undocumented, and you indeed need to read the sources to find out about it, you might also discover that setting a fontDir with some fonts and a property file with some properties to use those fonts, will solve the issue and there won't be a NPE anymore.

So I agree that setting this to false, and not supplying fonts is a misconfiguration, but should probably be handled better, with a more explicit message or fast failure.

On Android, however, the situation is different: `fontpath_linux.c`(with the native implementation of `FontConfigManager::getFontConfig`) is not used, but `FTFactory::getFallbacks` calls it nonetheless. 

Using `-Dprism.useFontConfig=false` prevents the runtime issue (UnsatisfiedLinkError), but then since `FontConfigManager.getFontConfigFont` returns null (note there is an `AndroidFontFinder` class that should manage fonts on Android), then the NPE can't be avoided, unless with add this null check. 

I agree that a null check hides the underlying issue, so prabably we should use `isLinux` instead. `PrismFontFactory` has this gate before calling `FontConfigManager`.

jperedadnr:
So far, the documentation in https://wiki.openjdk.org/display/OpenJFX/Font+Setup is still somehow valid in this situation (no fontConfig and providing embedded fonts).

So I can revert the change in `LogicalFont`, and modify the message in `FontConfigManager::initFontConfigLogFonts`:
```
System.err.println("Error: JavaFX detected no fonts! " +
                "Please refer to release notes for proper font configuration");
```         
to something more meaningful, like?
```
System.err.println("Error: JavaFX detected no fonts!\n " +
                "For proper font configuration, please check https://wiki.openjdk.org/display/OpenJFX/Font+Setup");
```            

The NPE will be still present though.

Thoughts?

jperedadnr:
@prrace could you have another look at this PR, after my change and comments?

prrace:
Sorry, I think I was out the week you updated it and then it disappeared off my radar.
Seems like a reasonable compromise

prrace (approved):

PR #1554, opened 2024-09-02
8339335: set number of parallel jobs when building webkit

johanvos:
When building webkit, the perl script in `modules/javafx.web/src/main/native/Tools/Scripts/build-webkit` will check for a `-j` argument in `$makeArgs`. If that doesn't exist, it will default to `numberOfCPUs()`.
This PR passes the `NUM_COMPILE_THREADS` from `build.gradle` to the `build-webkit` script. When not setting `NUM_COMPILE_THREADS` explicitly,it will default to the number of CPU's as well, hence it will have the same behavior as without this PR.
When setting `NUM_COMPILE_THREADS` explicitly, the provided number will be used by cmake as well now.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3722,6 +3722,7 @@ project(":web") {
                 exec {
                     workingDir("$webkitOutputDir")
                     def cmakeArgs = "-DENABLE_TOOLS=1"
+                    def makeArgs = "-j $NUM_COMPILE_THREADS"
                     if (IS_STATIC_BUILD) {
                         cmakeArgs = " $cmakeArgs -DSTATIC_BUILD=1 -DUSE_THIN_ARCHIVES=OFF";
                     }
@@ -3792,6 +3793,7 @@ project(":web") {
                     cmakeArgs += " -DJAVAFX_RELEASE_VERSION=${jfxReleaseMajorVersion}"
                     commandLine("perl", "$projectDir/src/main/native/Tools/Scripts/build-webkit",
                         "--java", "--icu-unicode", targetCpuBitDepthSwitch,
+                        "--makeArgs=${makeArgs}",
                         "--no-experimental-features", "--cmakeargs=${cmakeArgs}")
                 }
             }


kevinrushforth:
I'll run this through or CI as well as do some local builds. The one concern I have is that `NUM_COMPILE_THREADS` is set to 1 on macOS, meaning that this will increase the build time on macOS.

/reviewers 2

johanvos:
It is easy to create (yet) another property for this, but that will make the already complicated build.gradle even more complicated, so if possible I'd like to avoid that.

I read the cryptic comment in build.gradle
`// On Mac we limit it to 1 by default due to problems running gcc in parallel`
What is the problem that is being referred to here? And why would that problem occur in the other native compile jobs but not in the (resource intensive) webkit compile jobs?

johanvos:
I see 3 options:
1. the current approach in this PR. Drawback: huge increase in build time on mac
2. a separate option for the number of parallel jobs for webkit. Drawbacks: not elegant because
* we don't know why there are no issues compiling webkit with parallel jobs while there are issues compiling JavaFX native code with parallel jobs.
* adding another property increases complexity
3. Finding and fixing the cause for the parallel compilation issues on mac (JDK-8089417). Drawback: finding the cause may take a very long time -- if it is even still possible to reproduce
4. Removing the 1-thread restriction for mac. Drawback: unless we do (3), we don't know if this is not going to cause issues.

I'm ok with either approach. I consider option 3 the most elegant one, and option 2 the most pragmatic one (but also the least elegant).

kevinrushforth:
As you say, option 3 is the cleanest, however I believe it will very quickly morph into option 4. I am reasonably certain that [JDK-8089417](https://bugs.openjdk.org/browse/JDK-8089417) can be closed as "Cannot reproduce" because:

* The font T2K code that was hitting the failure is no longer part of our repo; T2K was a closed-source font rasterizer that we removed in JDK 10 via [JDK-8187147](https://bugs.openjdk.org/browse/JDK-8187147).
* Xcode switched from gcc to clang by default several years ago
 
I will locally remove the workaround, and then run the following in a loop a few hundred times to check whether we see any issues:

```
$ gradle cleanNative sdk
```

I did it a few times on my MacBook M1 and the only impact I see of removing the workaround is that a rebuild of the SDK without media or WebKit (so just rebuilding the graphics / glass native code) goes from 27 seconds to 12 seconds. :)

kevinrushforth:
Presuming that my testing shows no problems, I propose the following:

1. I will close [JDK-8089417](https://bugs.openjdk.org/browse/JDK-8089417) as "Cannot reproduce"
2. I will file a new bug to remove the workaround that currently disabled parallel compilation on mac for the graphics module (I think that's cleaner from a bookkeeping perspective than repurposing JDK-8089417)
3. You can add the removal of the logic that forced `NUM_COMPILE_THREADS` to 1 on macOS to this PR, and `/issue add` the new JBS issue

Thoughts?

johanvos:
> Presuming that my testing shows no problems, I propose the following:
> 
> 1. I will close [JDK-8089417](https://bugs.openjdk.org/browse/JDK-8089417) as "Cannot reproduce"
> 2. I will file a new bug to remove the workaround that currently disabled parallel compilation on mac for the graphics module (I think that's cleaner from a bookkeeping perspective than repurposing JDK-8089417)
> 3. You can add the removal of the logic that forced `NUM_COMPILE_THREADS` to 1 on macOS to this PR, and `/issue add` the new JBS issue
> 
> Thoughts?

Sounds like a plan. I'll add that to the PR once you confirm testing went as expected.

kevinrushforth:
Btw, I ran another CI build with the `NUM_COMPILE_THREADS=1` workaround for Mac removed, and the build times are good again. In fact, the x64 build runs even faster than before -- possibly because WebKit build defaults to a lower number than num_processors, and it happens to be 12 on that x64 build system.

Here are the updated numbers:

OS / Arch|Current|This PR|PR + don't limit num threads on macOS
---------|-------|-------|-------
macOS / aarch64|13m:41s|56m:36s|13m:43s
macOS / x64|28m:30s|2h:24m:52s|23m:01s

I should also add that the Linux and Windows build times are largely unaffected by this PR (which is good). They do run slightly faster with your patch than current, but the percentage gain is small.

kevinrushforth:
> Sounds like a plan. I'll add that to the PR once you confirm testing went as expected.

So far so good. I'll let it run a b it longer, but already I'm over 120 iterations with no failures.

johanvos:
> Btw, I ran another CI build with the `NUM_COMPILE_THREADS=1` workaround for Mac removed, and the build times are good again. In fact, the x64 build runs even faster than before -- possibly because WebKit build defaults to a lower number than num_processors, and it happens to be 12 on that x64 build system.
> 
> Here are the updated numbers:
> 
> OS / Arch	Current	This PR	PR + don't limit num threads on macOS
> macOS / aarch64	13m:41s	56m:36s	13m:43s
> macOS / x64	28m:30s	2h:24m:52s	23m:01s
> I should also add that the Linux and Windows build times are largely unaffected by this PR (which is good). They do run slightly faster with your patch than current, but the percentage gain is small.

Interesting.
The default NUM_COMPILE_THREADS comes from what Gradle gives us, using `${Runtime.runtime.availableProcessors()}`
The default threads that webkit uses `numberOfCPUs()` which is on mac implemented as `sysctl -n hw.ncpu`
On my macOS/aarch64, both implementations give "10", but this can indeed give different values.

kevinrushforth:
It's probably coincidental then.

Btw, my local test of `gradle cleanNative sdk` in a loop has hit 350 iterations with no errors. I'm running one last test on our CI build nodes (both x64 and aarch64) with 100 iterations. Once that completes, I'll close lose [JDK-8089417](https://bugs.openjdk.org/browse/JDK-8089417) and file the new bug.

kevinrushforth:
I'm done testing and have closed [JDK-8089417](https://bugs.openjdk.org/browse/JDK-8089417) as "Cannot reproduce". I filed a new bug:

[JDK-8339505](https://bugs.openjdk.org/browse/JDK-8339505): Enable parallel compilation of native code on macOS

Go ahead and remove the force-NUM_COMPILE_THREADS -to-1 workaround and add that new issue to this PR.

kevinrushforth:
The latest build changes look good. I'll do a final CI build and report results...probably tomorrow.

andy-goryachev-oracle:
I am using this command to build:

```
(rm -rf ./build ; gradle -PCOMPILE_MEDIA=true -PCOMPILE_WEBKIT=true all 2>&1) | tee ~/`date +"build-%Y-%m%d-%H%M%S"`.log
```

on macOS 14.6.1 / M1 Pro

looking at the Activity Monitor, I see 1 thread per clang compiler. Is this expected, or am I supposed to see more concurrent threads?

Screenshot 2024-09-03 at 13 25 55

kevinrushforth:
This setting (nor likely any setting we can, or would want to, control) doesn't affect the parallelism of an individual clang process, but rather controls the maximum number of concurrent clang processes that will be launched.

The next effect of this patch on macOS should be that we will now do parallel compilation for native code in the graphics module by default. There should be no change, by default, in the parallelism of WebKit compilation (there would have been before the most recent commit that Johan did).

What this patch will do is to apply the same `NUM_COMPILE_THREADS` flag to native WebKit as well as native graphics compilation. Prior to this fix, setting `NUM_COMPILE_THREADS` to, say 2, would have been used for the graphics build and ignored for the WebKit build.

andy-goryachev-oracle:
```
-PNUM_COMPILE_THREADS=1: BUILD SUCCESSFUL in 1h 9m 44s
-PNUM_COMPILE_THREADS=10: BUILD SUCCESSFUL in 17m 4s
-PNUM_COMPILE_THREADS=32: BUILD SUCCESSFUL in 19m 49s
```

andy-goryachev-oracle (approved):

kevinrushforth (approved):
All good.

PR #1556, opened 2024-09-05
8339603: Seal the class hierarchy of Node, Camera, LightBase, Shape, Shape3D

mstr2:
None of these classes can be extended by user code, and any attempt to do so will fail at runtime with an exception. For this reason, we can seal the class hierarchy and remove the run-time checks to turn this into a compile-time error instead.

In some cases, `Node` and `Shape` are extended by JavaFX classes in other modules, preventing those derived classes from being permitted subclasses. A non-exported `AbstractNode` and `AbstractShape` class is provided just for these scenarios. Note that introducing a new superclass is a source- and binary-compatible change (see [JLS ch. 13](https://docs.oracle.com/javase/specs/jls/se22/html/jls-13.html)).

I'm not sure if this change requires a CSR, as it doesn't change the specification in any meaningful way. There can be no valid JavaFX program that is affected by this change.

/reviewers 2

---------
### Progress
- [x] Change requires CSR request [JDK-8341659](https://bugs.openjdk.org/browse/JDK-8341659) to be approved

### Issues
 * [JDK-8339603](https://bugs.openjdk.org/browse/JDK-8339603): Seal the class hierarchy of Node, Camera, LightBase, Shape, Shape3D (**Enhancement** - P4)
 * [JDK-8341659](https://bugs.openjdk.org/browse/JDK-8341659): Seal the class hierarchy of Node, Camera, LightBase, Shape, Shape3D (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1556/head:pull/1556` \
`$ git checkout pull/1556`

`$ git checkout pull/1556` \
`$ git pull https://git.openjdk.org/jfx.git pull/1556/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1556`

`$ git pr show -t 1556`

</details>

<a href="https://git.openjdk.org/jfx/pull/1556.diff">https://git.openjdk.org/jfx/pull/1556.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1556#issuecomment-2331350730)
</details>


andy-goryachev-oracle:
What is the benefit of this change?

Also, this change is breaking.  One can currently create a Node:
```
public class CanCreateNode extends Application {
    @Override
    public void init() {
        Node n1 = new Node() {
        };
        class Yo extends Node {
        }
        Yo n2 = new Yo();
    }
```

mstr2:
> What is the benefit of this change?

It enforces a system invariant at compile-time instead of run-time.

> Also, this change is breaking. One can currently create a Node:
> 
> ```
> public class CanCreateNode extends Application {
>     @Override
>     public void init() {
>         Node n1 = new Node() {
>         };
>         class Yo extends Node {
>         }
>         Yo n2 = new Yo();
>     }
> ```

You can't do anything meaningful with this node. As soon as you add it to a scene graph, you'll get a runtime exception. The fact that you can't extend `Node` is a system invariant, finding pathological situations in which you won't get an exception is just a loophole.

kevinrushforth:
If this PR goes forward, it will need a CSR, since it does impact source code compatibility. As you point out, you can't usefully extend Node today, since you will get a runtime exception as soon as you try to create an instance of a subclassed Node, so what this does is move a runtime error to compile time. That's generally a good thing, but..is it worth doing? We can continue to discuss.

/csr

mstr2:
> If this PR goes forward, it will need a CSR, since it does impact source code compatibility. As you point out, you can't usefully extend Node today, since you will get a runtime exception as soon as you try to create an instance of a subclassed Node, so what this does is move a runtime error to compile time. That's generally a good thing, but..is it worth doing? We can continue to discuss.
> 
> /csr

In addition to moving a runtime error to compile time, it also documents intent for future JavaFX maintainers. Given that the change is not invasive and has pretty much no downsides, I think it is worth it.

andy-goryachev-oracle:
> The fact that you can't extend `Node` is a system invariant

But I **can**, and the code provided above runs with no exception - in either `init()` or `start()`, making this change a breaking one.

Whether such code is meaningful is another story.

I would vote against this change.

andy-goryachev-oracle:
can you point me to where it says one can't extend Node or that it is verboten?

nlisker:
I thought about doing this some years ago for `Shape3D`, `Material` and `LightBase` when I was adding light types. `Shape3D` state:

> An application should not extend the Shape3D class directly. Doing so may lead to an UnsupportedOperationException being thrown.

However, I didn't see enough value in this. There is some small benefit for `LightBase` because it allows an exhaustive switch on its subtypes in `NGShape3D` (IIRC). Also, for `Material` and `LightBase`, even before compile time - at code writing time - the user will find it impossible to subtype these. `Shape3D` can be made to allow inheritance as its subtypes don't require native-level handling. I assume that this is why it has a comment about subtyping, and there is more benefit in sealing it.

The downside is minor too: custom shaders will allow user implementations, but that's not coming any time soon.

Since removing `sealed`/`final` from a class can be done in the future without much hassle, I'm not against sealing the above classes (and removing the comments about not extending them). It just doesn't overcome the initial inertia for me. I will review the changes for these classes if this goes forward.

As for `Node` and `Shape`, I didn't look at subclassing them and never thought about it, so I have no opinion there.

mstr2:
> can you point me to where it says one can't extend Node or that it is verboten?

https://github.com/openjdk/jfx/blob/4647367ce4590440fbb85675055225c869502314/modules/javafx.graphics/src/main/java/com/sun/javafx/scene/NodeHelper.java#L81

nlisker:
The documentation for `Node` has the line:

> An application should not extend the Node class directly. Doing so may lead to an UnsupportedOperationException being thrown.

Just above [StringID](https://openjfx.io/javadoc/22/javafx.graphics/javafx/scene/Node.html#StringID).

andy-goryachev-oracle:
a comment in some internal helper class is not, in my opinion, a normative reference.

I am not against this change in principle - it might be a welcome enhancement, if it were solving a real world problem.  I am against this change because it has a risk of existing application failing (compatibility risk).

App developers are a creative bunch, they often do things that the platform developers did not expect.

andy-goryachev-oracle:
"should" and "may" are not the same as "must" and "will".

but anyway, my objection is due to possible compatibility risk.

nlisker:
I forgot to mention `Camera` in my reply. Same as the 3D classes: the user can't do it if they try, so it will fail before compilation. I don't mind sealing it and will review changes for this as well. I also wonder if there are even other camera types in the literature and if the class can be built for subtyping at all.

andy-goryachev-oracle:
I wonder if it is possible to have some kind of interim solution that will warn the app developers that they are doing something they are not supposed to?

- Printing a warning won't work: it will be ignored.
- Shutting down the app (and allowing it to run with a system property) is probably even worse

What do you think?

mstr2:
> "should" and "may" are not the same as "must" and "will".

I think this is just suboptimal language, and not a material distinction. You _will_ get a runtime error if you use it in any meaningful way. Anyway, if a developer _really_ wants to create an instance of `Node`, they can use the non-exported `AbstractNode` at their own peril.

kevinrushforth:
> > "should" and "may" are not the same as "must" and "will".
> 
> I think this is just suboptimal language, and not a material distinction. You _will_ get a runtime error if you use it in any meaningful way. Anyway, if a developer _really_ wants to create an instance of `Node`, they can use the non-exported `AbstractNode` at their own peril.

Indeed it is a case of suboptimal language. The intention was to make it clear that this is _not_ supported. Period. And it isn't in any meaningful way. If there are corner cases where we miss throwing an exception, that is a bug.

There might be other reasons to not make this change -- I haven't formed an opinion yet -- but concern over compatibility isn't one of them.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -412,7 +416,9 @@
  * @since JavaFX 2.0
  */
 @IDProperty("id")
-public abstract class Node implements EventTarget, Styleable {
+public abstract sealed class Node
+        implements EventTarget, Styleable
+        permits AbstractNode, Camera, LightBase, Parent, SubScene, Canvas, ImageView, Shape, Shape3D {
```
andy-goryachev-oracle, line 421: We may need to use stronger wording in javadoc in regards to extending Nodes.  should -> must, etc.  (or, rather, 'cannot' and possibly explain why?)

mstr2: I've removed the language that you can't extend those classes, as it is now redundant because it is enforced by the compiler. We shouldn't document obvious facts without providing any added value (for example, adding context and reasoning _why_ it's the way it is).

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/CameraHelper.java
```diff
@@ -57,7 +57,7 @@ public static void initHelper(Camera camera) {
 
     @Override
     protected NGNode createPeerImpl(Node node) {
-        throw new UnsupportedOperationException("Applications should not extend the Camera class directly.");
+        throw new AssertionError();
```
andy-goryachev-oracle, line 60: is this the right error type?
why change UnsupportedOperationException?

nlisker:
If this goes forward, please seal `Material` as well.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/CameraHelper.java
mstr2: When `Camera` is sealed, this method cannot be called by user code. `UnsupportedOperationException` would be okay if it could reasonably be called, but this is now an invariant that would signal a bug in JavaFX.

mstr2:
> If this goes forward, please seal `Material` as well.

If we seal `Material`, we should probably also seal `Paint`. What do you think?

nlisker:
> > If this goes forward, please seal `Material` as well.
> 
> If we seal `Material`, we should probably also seal `Paint`. What do you think?

I don't see why they are related, but if `Paint` can't be subclassed then I don't see a reason not to seal it. I can envision possibilities to create new paints, like other types of gradients. There's nothing inherently stopping that except for access to methods like platform paint.

mstr2:
> I don't see why they are related, but if `Paint` can't be subclassed then I don't see a reason not to seal it. I can envision possibilities to create new paints, like other types of gradients. There's nothing inherently stopping that except for access to methods like platform paint.

They are in the same package, and they are related insofar as an argument for sealing `Material` would also apply for `Paint`. I'll do that in a separate PR.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/CameraHelper.java
andy-goryachev-oracle: maybe an explanatory comment should be added?

kevinrushforth: We typically use `InternalError` for the case where there should be no possible reason for a particular case, but also use `AssertionError` in some cases. Either is OK with me. I agree that a comment to the affect of "Should not ever get here" wouldn't be a bad idea.

mstr2: I've added a comment to `CameraHelper` and `LightBaseHelper`.

andy-goryachev-oracle (approved):
thank you for adding the comment!

kevinrushforth (approved):
This looks good. I left one question about whether (or not) to throw `InternalError` on a null (this was a case of a runtime check that should no longer be possible).

I also left a question in the CSR about the two public classes (MediaView and SwingNode) that extend `AbstractNode`. They show up in the javadocs, even though they are inaccessible. I think this is OK, but wanted Joe's take on it.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/NodeHelper.java
```diff
@@ -65,23 +63,7 @@ protected NodeHelper() {
...
-            } else {
-                nodeType = "Node";
-            }
-
-            throw new UnsupportedOperationException(
-                    "Applications should not extend the "
-                    + nodeType + " class directly.");
-        }
-        return helper;
+        return nodeAccessor.getHelper(node);
```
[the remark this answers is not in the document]
mstr2: I think that would be unnecessarily verbose, and it doesn't seem like it would offer a benefit: as of now, this can't happen. Let's assume that the class hierarchy is changed in the future, such that `null` might be returned. This wouldn't result in a compile-time error. The exception would only be thrown when someone tries to acquire the `NodeHelper`, but in this case, we'd most likely get a NPE anyway.

andy-goryachev-oracle (approved):
looks good.

PR #1562, opened 2024-09-11
8339517: Add --enable-native-access flag when running tests

kevinrushforth:
When running JavaFX tests and apps with JDK 24, we now get illegal access warnings of the following form:

```
WARNING: java.lang.System::load has been called by com.sun.glass.utils.NativeLibLoader in module javafx.graphics (file:build/shims/javafx.graphics/)
```

With the current jfx master, I get 301 such warnings on my macOS system (see the `test.log` file attached to the JBS bug report).

In order to avoid those warnings, and to avoid the eventual errors in the future, we need to pass one of the following two options when running test programs:

* "java --enable-native-access=javafx.graphics,javafx.media,javafx.web" -- for unit tests in the web and systemTests subprojects, as well as in the generated `build/run.args` file
* "java --enable-native-access=javafx.graphics" -- for all other unit tests

With this patch, we are down to 5 such warnings on macOS (on Linux there are few more, but the reasons are the same). The remaining warnings are due to the following two issues:

1. We are making a native JNI call from javafx.swing, which we should not be doing. This is a JavaFX bug. The native library in question is loaded by the javafx.graphics modules, so the solution will likely be to move the native method call to somewhere in javafx.graphics and call that new method from javafx.swing.
2. We have two unit tests that use a custom `ModuleLayer` to load the JavaFX modules at runtime, after the test application is launched. In this case the "--enable-native-access" is ineffective. I suspect this is a JDK bug.

I will file a pair of follow-on bugs for these.

---------
### Progress

### Issue
 * [JDK-8339517](https://bugs.openjdk.org/browse/JDK-8339517): Add --enable-native-access flag when running tests (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1562/head:pull/1562` \
`$ git checkout pull/1562`

`$ git checkout pull/1562` \
`$ git pull https://git.openjdk.org/jfx.git pull/1562/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1562`

`$ git pr show -t 1562`

</details>

<a href="https://git.openjdk.org/jfx/pull/1562.diff">https://git.openjdk.org/jfx/pull/1562.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1562#issuecomment-2343893457)


kevinrushforth:
Reviewer: @arapte

kevinrushforth:
Update on problem 2:

> We have two unit tests that use a custom `ModuleLayer` to load the JavaFX modules at runtime, after the test application is launched. In this case the "--enable-native-access" is ineffective. I suspect this is a JDK bug.

This turned out to be a test bug. When using a custom `ModuleLayer` to load classes, a test (or application) needs to call `ModuleLayer.Controller::enableNativeAccess`. So the follow-up bug will be to do this in those tests.

(the reason it will be a follow-up bug is that `ModuleLayer.Controller::enableNativeAccess` was preview in JDK 21 and final in JDK 22 along with Panama, so unless we want to do this via reflection, we need to wait until we bump the minimum JDK for JavaFX).

kevinrushforth:
> This turned out to be a test bug. When using a custom `ModuleLayer` to load classes, a test (or application) needs to call `ModuleLayer.Controller::enableNativeAccess`. So the follow-up bug will be to do this in those tests.
> 
> (the reason it will be a follow-up bug is that `ModuleLayer.Controller::enableNativeAccess` was preview in JDK 21 and final in JDK 22 along with Panama, so unless we want to do this via reflection, we need to wait until we bump the minimum JDK for JavaFX).

Since I need a follow-up bug for this either way, I decided to fix this using reflection for now, and file a follow-up bug to replace the reflection calls with direct calls to `ModuleLayer.Controller::enableNativeAccess` once we bump the minimum to JDK 22.

kevinrushforth:
I filed the following two follow-up bugs:

[JDK-8340005](https://bugs.openjdk.org/browse/JDK-8340005): Eliminate native access calls from javafx.swing
[JDK-8340004](https://bugs.openjdk.org/browse/JDK-8340004): [TestBug] Call ModuleLayer.Controller::enableNativeAccess directly rather than via reflection

arapte (approved):
Verified that warnings disappear with this change.

PR #1567, opened 2024-09-15
8339515: [TestBug] Convert web tests to JUnit 5

jaybhaskar:
Successfully converted web tests from JUnit 4 to JUnit 5, ensuring all tests are fully compliant with the JUnit 5 framework.

---------
### Progress

### Issue
 * [JDK-8339515](https://bugs.openjdk.org/browse/JDK-8339515): [TestBug] Convert web tests to JUnit 5 (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1567/head:pull/1567` \
`$ git checkout pull/1567`

`$ git checkout pull/1567` \
`$ git pull https://git.openjdk.org/jfx.git pull/1567/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1567`

`$ git pr show -t 1567`

</details>

<a href="https://git.openjdk.org/jfx/pull/1567.diff">https://git.openjdk.org/jfx/pull/1567.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1567#issuecomment-2351573490)


modules/javafx.web/src/test/java/test/com/sun/webkit/network/CookieTest.java
```diff
@@ -443,33 +443,25 @@ public CookieModel(String name, String value) {
...
-                expected.creationTime.baseTime(), actual.getLastAccessTime());
-        assertEquals("Unexpected persistent flag" + s,
-                expected.persistent, actual.getPersistent());
-        assertEquals("Unexpected hostOnly flag" + s,
-                false, actual.getHostOnly());
-        assertEquals("Unexpected secureOnly flag" + s,
-                expected.secureOnly, actual.getSecureOnly());
-        assertEquals("Unexpected httpOnly flag" + s,
-                expected.httpOnly, actual.getHttpOnly());
+        assertEquals(expected.name, actual.getName(),"Unexpected name" + s);
```
Maran23, line 454: minor: missing space after the comma -> `, `

modules/javafx.web/src/test/java/test/com/sun/webkit/network/CookieTest.java
```diff
@@ -443,33 +443,25 @@ public CookieModel(String name, String value) {
...
+        assertEquals(expected.value, actual.getValue(),"Unexpected value" + s);
```
Maran23, line 455: minor: missing space after the comma -> `, `

modules/javafx.web/src/test/java/test/javafx/scene/web/FileReaderTest.java
```diff
@@ -226,10 +238,14 @@ private void testLatch(CountDownLatch latch) {
...
-                                       expectedArrayBuffer.length, obj.getMember("length"));
+                String filePath = fileList[0];
+                int expectedLength = expectedArrayBuffer.length;
+                int actualLength = (Integer) obj.getMember("length");
+                String message = String.format("File at %s: Expected length %d but got %d",
+                        filePath, expectedLength, actualLength);
+                assertEquals(expectedLength, actualLength, message);
                 for (int i = 0; i < expectedArrayBuffer.length; i++) {
-                    assertEquals("Unexpected file content received", expectedArrayBuffer[i], ((Number)(obj.getSlot(i))).byteValue());
+                    assertEquals(expectedArrayBuffer[i], ((Number)(obj.getSlot(i))).byteValue(),"Unexpected file content received");
```
Maran23, line 248: minor: missing space after the comma -> `, `

modules/javafx.web/src/test/java/test/javafx/scene/web/LoadTest.java
```diff
@@ -114,56 +119,59 @@ private void testLoadEmpty(String url) {
...
         submit(() -> {
             Document doc = web.getDocument();
-            assertNotNull("Document should not be null", doc);
+            assertNotNull(doc, "Document should not be null");
 
             Element html = doc.getDocumentElement();
-            assertNotNull("There should be an HTML element", html);
-            assertEquals("HTML element should have tag HTML", "HTML", html.getTagName());
+            assertNotNull(html, "There should be an HTML element");
+            assertEquals("HTML", html.getTagName(),"HTML element should have tag HTML");
```
Maran23, line 132: minor: missing space after the comma -> `, `

modules/javafx.web/src/test/java/test/javafx/scene/web/LoadTest.java
```diff
@@ -180,31 +188,35 @@ private void testLoadEmpty(String url) {
...
+    public void testLoadTitleChanged() {
         final String FILE = "src/test/resources/test/html/ipsum.html";
         final CountDownLatch latch = new CountDownLatch(1);
 
         submit(() -> {
             WebEngine webEngine = new WebEngine();
             webEngine.titleProperty().addListener((observable, oldValue, newValue) -> {
-                assertTrue("loadContent in SUCCEEDED State", webEngine.getLoadWorker().getState() == SUCCEEDED);
-                assertEquals("Title", "Lorem Ipsum", webEngine.getTitle());
+                assertTrue(webEngine.getLoadWorker().getState() == SUCCEEDED,"loadContent in SUCCEEDED State");
```
Maran23, line 218: minor: missing space after the comma -> `, `

modules/javafx.web/src/test/java/test/javafx/scene/web/SubresourceIntegrityTest.java
```diff
@@ -90,26 +87,26 @@ public void setup() throws Exception {
...
+                        "<body>%s</body>\n" +
+                        "</html>", scriptUrl, hashValue, NOT_LOADED);
         out.write(html.getBytes());
         out.close();
     }
 
-    @Test
+    @ParameterizedTest
+    @MethodSource("dataProvider")
     public void testScriptTagWithCorrectHashValue() {
```
Maran23, line 99: Major: This methods need the parameters from the `MethodSource` AFAIK, that is:
`testScriptTagWithCorrectHashValue(String hashValue, String expected) { .. }`

The member variables should be removed then:
``` 
private final String hashValue = "";
private final String expected = "";
```

Maran23 (changes requested):
There is one major issue with the parameterized test as far as I can see. Some minor style guide things, looks good otherwise!

Maran23:
Also one note: The copyright year was not updated, but this could also be done by the script afterwards I guess.

modules/javafx.web/src/test/java/test/javafx/scene/web/SubresourceIntegrityTest.java
jaybhaskar: In this case, keeping the values as instance variables is more reliable. Since all the tests pass when hashValue and expected are instance variables and  some of them fail when I  adding parameter to testScriptTagWithCorrectHashValue , 

it's likely that their assignment happens at the correct time relative to the @BeforeEach setup. Using parameters may interfere with the test lifecycle, since hasValue is being used in 
    @BeforeEach
    public void setup() ....{}

 it's safer and more straightforward to use instance variables.This would also make it easier to manage the state across @BeforeEach

Maran23: I think you need to do the same as done by @andy-goryachev-oracle in his JUnit5 replacement PR:
```
    public void setup(String hashValue, String expected) throws Exception {
        htmlFile = new File("subresource-integrity-test.html");
        ...
    }

    @ParameterizedTest
    @MethodSource("dataProvider")
    public void testScriptTagWithCorrectHashValue(String hashValue, String expected) {
        setup(hashValue, expected);
        ...
    }
```

modules/javafx.web/src/test/java/test/com/sun/webkit/SharedBufferTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2024, Oracle and/or its affiliates. All rights reserved.
```
kevinrushforth, line 2: You removed the starting copyright year, which is not OK. I recommend to revert all of the copyright changes. Then, if you prefer, you can either update it correctly to leave the starting year, or better still, leave it alone and let the script that @arapte runs periodically take care of it.

kevinrushforth (commented):
You need to revert the copyright year change, since you've mistakenly removed the start year.

modules/javafx.web/src/test/java/test/javafx/scene/web/SubresourceIntegrityTest.java
andy-goryachev-oracle: @Maran23 is right.  In the absence of support for parameterized class-level tests the easiest solution (I think) is to parameterize every `@Test`, which means:
- add parameters to each method, replace `@Test` with `@ParameterizedTest` and `@MethodSource`
- add parameters to setup method (`@Before`, remove the annotation)
- call the setup method from each test method
- remove the fields that were set in the constructor and the constructor itself

Things might be a bit more complicated in the case of test class hierarchies, so additional work and checks will be needed.

modules/javafx.web/src/test/java/test/com/sun/webkit/SharedBufferTest.java
jaybhaskar: reverted copyright date change

andy-goryachev-oracle:
looks like there is a merge conflict label added

modules/javafx.web/src/test/java/test/javafx/scene/web/SubresourceIntegrityTest.java
jaybhaskar: When hashValue and expected are instance variables, they are initialized once per test case due to how JUnit creates a new instance of the test class for each @ParameterizedTest. This ensures that the state is clean between tests and no variable is shared between different test cases. So I think it is ok.

modules/javafx.web/src/test/java/test/javafx/scene/web/FormControlsTest.java
```diff
@@ -36,46 +36,38 @@
...
 import javafx.scene.web.WebEngineShim;
 
-import org.junit.After;
-import org.junit.Test;
-import org.junit.runner.RunWith;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.runners.Parameterized.Parameters;
 import org.junit.runners.Parameterized;
```
arapte, line 43: These two imports should be removed.

modules/javafx.web/src/test/java/test/javafx/scene/web/SubresourceIntegrityTest.java
```diff
@@ -29,59 +29,56 @@
...
 import java.util.Collection;
-import org.junit.After;
-import org.junit.Before;
-import org.junit.Test;
-import org.junit.runner.RunWith;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.runners.Parameterized.Parameters;
```
arapte, line 36: This import should be removed.

arapte (changes requested):
Suggesting a few minor changes.

modules/javafx.web/src/test/java/test/javafx/scene/web/MiscellaneousTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2022, Oracle and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
jaybhaskar: already did like Copyright (c) 2011, 2022,

modules/javafx.web/src/test/java/test/javafx/scene/web/HistoryTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2022, Oracle and/or its affiliates. All rights reserved.
```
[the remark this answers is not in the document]
jaybhaskar: already did like Copyright (c) 2011, 2022,

modules/javafx.web/src/test/java/test/javafx/scene/web/SubresourceIntegrityTest.java
jaybhaskar: resoled

modules/javafx.web/src/test/java/test/javafx/scene/web/FormControlsTest.java
jaybhaskar: resolved

modules/javafx.web/src/test/java/test/javafx/scene/web/HistoryTest.java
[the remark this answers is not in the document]
kevinrushforth: No, what happened is that the _way_ you reverted your copyright change caused a change that you had earlier merged in from master to be lost. So you have effectively undone a change that was done by another commit that went into master. Please re-revert the change as Ambarish pointed out.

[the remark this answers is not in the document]
jaybhaskar: got it, Thanks

kevinrushforth:
Note to reviewers: The GHA build compiles the web tests, but doesn't run them. Someone will need to do an offline test run with this patch.

modules/javafx.web/src/test/java/test/javafx/scene/web/SubresourceIntegrityTest.java
arapte: I also think the test needs to be modified as suggested by @Maran23 and @andy-goryachev-oracle .
With the changes in this PR the member variables are never assigned with value.
I verified the values of the two member variables by printing them before line 102, They are always empty.
Apply below change, after applying this PR changes.
```
diff --git a/modules/javafx.web/src/test/java/test/javafx/scene/web/SubresourceIntegrityTest.java b/modules/javafx.web/src/test/java/test/javafx/scene/web/SubresourceIntegrityTest.java
index e65ff7d32f..0d2d670777 100644
--- a/modules/javafx.web/src/test/java/test/javafx/scene/web/SubresourceIntegrityTest.java
+++ b/modules/javafx.web/src/test/java/test/javafx/scene/web/SubresourceIntegrityTest.java
@@ -99,6 +99,8 @@ public final class SubresourceIntegrityTest extends TestBase {
         load(htmlFile);
         final String bodyText = (String) executeScript("document.body.innerText");
         assertNotNull("document.body.innerText must be non null for " + hashValue, bodyText);
+        System.err.println("hashValue = " + hashValue);
+        System.err.println("expected = " + expected);
         assertEquals(hashValue, expected, bodyText);
     }
```

Screenshot 2024-09-17 at 4 51 46 PM

arapte (changes requested):
Providing a few comments, majority are related to correcting indentation correction.
1. Only space changes: Some indentation changes only add or remove spaces. I would suggest to avoid these changes. Or add only if the existing formatting is very bad. But would recommend to avoid.
2. Some indentations are bad, which needs correction for example having `);` on a separate line.

=> I have not pointed out all the indentation corrections. Please do look at other similar changes too.

Another comment is regarding **Parameterized** test: `modules/javafx.web/src/test/java/test/javafx/scene/web/SubresourceIntegrityTest.java`
Looks like the test is not converted correctly. The parameters are never assigned to the member variables. Hence never tested. Please take a look at the specific comment on the test. (please check [my reply to previous discussion)](https://github.com/openjdk/jfx/pull/1567#discussion_r1763063040)

modules/javafx.web/src/test/java/test/javafx/scene/web/SubresourceIntegrityTest.java
kevinrushforth: @jaybhaskar Yes, there is an issue as the above comment from @arapte shows. Please don't mark this as "resolved" until it is, and @arapte is satisfied.

jaybhaskar: Thanks , I have modified and push according

jaybhaskar: I found issue , and done changes according to review, Thanks all

arapte (approved):
LGTM

Maran23 (approved):
looks good now

modules/javafx.web/src/test/java/test/javafx/scene/web/MiscellaneousTest.java
```diff
@@ -150,7 +152,9 @@ public void start() {
            submit(obj::start);
            // Try accessing the resultObject created in JavaFX Application Thread
            // from someother thread. It should throw an exception.
-           obj.resultObject.toString();
+           assertThrows(IllegalStateException.class, () -> {
```
andy-goryachev-oracle, line 155: on second thought, I think this and the previous changes, even though not technically equivalent, are perfectly fine, in the spirit of the test.

andy-goryachev-oracle (commented):
- query for old imports produces 0 hits (good)
- left some minor comments, one (assertArrayEquals) might need a code change

PR #1574, opened 2024-09-19
8340405: JavaFX shutdown hook can hang preventing app from exiting

kevinrushforth:
This PR fixes a hang on exit which can happen if QuantumToolkit.dispose hangs when called from the QuantumToolkit shutdown hook. A shutdown hook should never run indefinitely, so the fix is to call dispose from a background thread and wait for up to 5 seconds for it to finish normally, and then throw an Error so that the application will be able to exit anyway.

The QuantumToolkit registers a "Glass/Prism Shutdown Hook" that calls the QuantumToolkit.dispose method to gracefully shutdown the toolkit and renderer. If dispose hangs or deadlocks, this will prevent the application from exiting, requiring the process to be killed. This has been observed on macOS, due to [JDK-8238505](https://bugs.openjdk.org/browse/JDK-8238505), which is is an intermittent deadlock that occurs during shutdown. We ought to eventually fix that intermittent deadlock, but even in the presence of a deadlock or hang, we don't want the shutdown hook to hang.

This can happen at any time, but is especially a problem during an automated test run, which will then hang until the Jenkins job is either killed or its global timer (currently set to 12 hours) kills the job. Even in that case, processes can be left lying around which can affect subsequent runs.

I discovered this while testing [JDK-8328629](https://bugs.openjdk.org/browse/JDK-8328629), which adds a default Unit 5 timeout, and without fixing the deadlock, the timeout will be unable to abort the hanging test.

### Testing instructions

I tested this in connection with PR #1575 / [JDK-8328629](https://bugs.openjdk.org/browse/JDK-8328629) using my [test-timeout-and-shutdown-hang](https://github.com/kevinrushforth/jfx/tree/test-timeout-and-shutdown-hang) branch. That branch has the fixes from both this PR and PR #1575 as well as tests that will hang indefinitely without the fixes and fail due to the expected timeout with the fixes.

---------
### Progress

### Issue
 * [JDK-8340405](https://bugs.openjdk.org/browse/JDK-8340405): JavaFX shutdown hook can hang preventing app from exiting (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1574/head:pull/1574` \
`$ git checkout pull/1574`

`$ git checkout pull/1574` \
`$ git pull https://git.openjdk.org/jfx.git pull/1574/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1574`

`$ git pr show -t 1574`

</details>

<a href="https://git.openjdk.org/jfx/pull/1574.diff">https://git.openjdk.org/jfx/pull/1574.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1574#issuecomment-2361892582)

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
@@ -259,7 +259,25 @@ public final class QuantumToolkit extends Toolkit {
          */
         shutdownHook = new Thread("Glass/Prism Shutdown Hook") {
             @Override public void run() {
-                dispose();
+                // Run dispose in a background thread and wait for up to
+                // 5 seconds for it to finish. If it doesn't, then throw an
+                // error, so that if dispose hangs or deadlocks, it won't
+                // prevent the JVM from exiting.
+                var disposeLatch = new CountDownLatch(1);
+                var thr = new Thread(() -> {
+                    dispose();
+                    disposeLatch.countDown();
+                });
+                thr.setDaemon(true);
+                thr.start();
+
+                try {
+                    if (!disposeLatch.await(5, TimeUnit.SECONDS)) {
+                        throw new InternalError("dispose timed out");
+                    }
+                } catch (InterruptedException ex) {
+                    throw new InternalError(ex);
+                }
             }
         };
         @SuppressWarnings("removal")


kevinrushforth:
Reviewer: @arapte and either @lukostyra or @andy-goryachev-oracle 

This should be a safe fix, but I would like a second reviewer.

/reviewers 2

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
```diff
@@ -259,7 +259,25 @@ boolean get() {
          */
         shutdownHook = new Thread("Glass/Prism Shutdown Hook") {
             @Override public void run() {
-                dispose();
+                // Run dispose in a background thread and wait for up to
+                // 5 seconds for it to finish. If it doesn't, then throw an
```
andy-goryachev-oracle, line 263: is 5 seconds a good timeout value?

could it be shorter?
is it possible for the normal shutdown on a real system take more than 5 seconds?

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
```diff
@@ -259,7 +259,25 @@ boolean get() {
...
+                var thr = new Thread(() -> {
+                    dispose();
+                    disposeLatch.countDown();
+                });
+                thr.setDaemon(true);
+                thr.start();
+
+                try {
+                    if (!disposeLatch.await(5, TimeUnit.SECONDS)) {
+                        throw new InternalError("dispose timed out");
```
andy-goryachev-oracle, line 276: could this happen on a real system?  if so, should the message be more clear and user-friendly?

also, if it can be encountered in the field, it might be better to declare a constant and add a javadoc to it explaining what it is.

what do you think?

arapte: I think javadoc and a property might be an overkill at this moment. The root problem is the deadlock and is very intermittent, and it hinders with the jenkins jobs. The change seems good means to avoid such situations where process keep running for hours. And if in case the frequency of issue increases or we could find a definite way of reproducing the issue, then we should try to fix the root cause.
Approving, I shall re-approve if updated.

arapte (approved):

johanvos:
I'll review this too.

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
kevinrushforth: I think 5 seconds is a reasonable timeout value. I wouldn't want it to be any shorter nor too much longer. Normally, this is something that would take a few milliseconds.

FWIW, we have a similar 5 second timeout in the (macOS-specific) initialization code that waits for app activation.

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
kevinrushforth: I also think that's overkill to document this or make it programmable. This timed wait is to handle the (very rare) case of a hang/deadlock that will otherwise prevent the application from exiting at all. Remember that this code only gets run when the application calls `System.exit(int)`.

I could provide a slightly more informative message. Maybe something like:

```
Timeout shutting down the JavaFX toolkit
```

johanvos:
This fix looks good. While we should avoid that the QuantumToolkit.dispose hangs, it is indeed good practice to take into account the (rare) possibility that it hangs. In this case, there is not much a developer or user can do, so the message returned with the InternalError doesn't matter to me.
Regardless of this fix, it would still be good to have a fix for [JDK-8238505](https://bugs.openjdk.org/browse/JDK-8238505) (although that is likely less urgent now that the CI blocking is tackled) -- and there might be other issues that cause the dispose thread to not finish.

johanvos (approved):

andy-goryachev-oracle (approved):
thanks for clarifications!
if this message is unlikely to be seen, the code is good as is.

kevinrushforth:
Thanks for the reviews. I'll integrate it this afternoon (this is a product fix and it hasn't been "rfr" for 24 hours yet) if there are no other comments.

kevinrushforth:
> Regardless of this fix, it would still be good to have a fix for [JDK-8238505](https://bugs.openjdk.org/browse/JDK-8238505) (although that is likely less urgent now that the CI blocking is tackled) -- and there might be other issues that cause the dispose thread to not finish.

Agreed on both points. While [JDK-8238505](https://bugs.openjdk.org/browse/JDK-8238505) is less urgent after this fix, it's still a bug. I instrumented the timer and dispose code as part of evaluating and then filing this shutdown hook bug, and I'll add some comments by way of a partial evaluation of that bug.

PR #1569, opened 2024-09-16
8339510: [TestBug] Convert system tests to JUnit 5

andy-goryachev-oracle:
Converting system tests to junit5.

Please see migration notes:
https://github.com/andy-goryachev-oracle/Test/blob/main/doc/Tests/JUnit5Migration.md

### Notes:

I see shutdown timeout on linux, this will be addressed by [JDK-8340403](https://bugs.openjdk.org/browse/JDK-8340403)
```
QPathTest > executionError FAILED
    org.opentest4j.AssertionFailedError: Exceeded timeout limit of 10000 msec
        at app//org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:39)
        at app//org.junit.jupiter.api.Assertions.fail(Assertions.java:134)
        at app//test.util.Util.runAndWait(Util.java:156)
        at app//test.util.Util.runAndWait(Util.java:127)
        at app//test.util.Util.shutdown(Util.java:365)
        at app//test.com.sun.marlin.QPathTest.teardownOnce(QPathTest.java:146)
```

This test might fail intermittently (?) on linux only:
```
SetSceneScalingTest > testShowAndSetScene() FAILED
    org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
        at app//org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:55)
        at app//org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:40)
        at app//org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:35)
        at app//org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:179)
        at app//test.robot.javafx.stage.SetSceneScalingTest$TestShowSetSceneApp.test(SetSceneScalingTest.java:129)
        at app//test.robot.javafx.stage.SetSceneScalingTest$TestApp.runTest(SetSceneScalingTest.java:89)
        at app//test.robot.javafx.stage.SetSceneScalingTest.testShowAndSetScene(SetSceneScalingTest.java:193)
```

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issue
[...]

PR #1575, opened 2024-09-19
8328629: JUnit test without a timeout value can hang indefinitely

kevinrushforth:
This PR adds a default timeout of 120 seconds for JUnit 5 tests that do not have an explicit `@Timeout` on either the methods or the class, and a default timeout of 20 seconds for lifecycle methods (e.g., `@BeforeEach`, `@BeforeAll`, `@AfterEach`, `@AfterAll`).

JUnit 5 effects its timeout by interrupting the test thread. We have several tests that catch and ignore InterruptedException for sleep or await, so I fixed these tests to throw an error if InterruptedException is caught (they may or may not be a problem, but if any of them can happen in a loop, it would block the timeout from killing the test).

While testing this, I discovered that by itself, this JUnit 5 timeout is ineffective if the deadlock described in [JDK-8238505](https://bugs.openjdk.org/browse/JDK-8238505)  occurs. The reason for this is that the deadlock almost always occurs when called from a shutdown hook (called by `System.exit`), and that prevents the process from exiting even when interrupted by the test runner after the timeout. A separate PR is out for that bug fix.

### Testing instructions

I tested this in connection with PR #1574 / [JDK-8340405](https://bugs.openjdk.org/browse/JDK-8340405) using my [test-timeout-and-shutdown-hang](https://github.com/kevinrushforth/jfx/tree/test-timeout-and-shutdown-hang) branch. That branch has the fixes from both this PR and PR #1574 as well as tests that will hang indefinitely without the fixes and fail due to the expected timeout with the fixes.

The default timeout values for test and lifecycle methods can be modified using a gradle property:

* `JUNIT_TEST_TIMEOUT` : default = 120s
* `JUNIT_LIFECYCLE_TIMEOUT` : default = 20s

For example, to set a timeout of 60 seconds for tests and 15 seconds for lifecycle methods:

```
gradle -PJUNIT_TEST_TIMEOUT=60s -PJUNIT_LIFECYCLE_TIMEOUT=15s test
```

---------
### Progress

### Issue
 * [JDK-8328629](https://bugs.openjdk.org/browse/JDK-8328629): JUnit test without a timeout value can hang indefinitely (**Bug** - P4)

### Reviewers
 * [Ambarish Rapte](https://openjdk.org/census#arapte) (@arapte - **Reviewer**) Review applies to [e16ef7ec](https://git.openjdk.org/jfx/pull/1575/files/e16ef7ece7cb95ab0baf2fd37349396e4986fcbf)
 * [Marius Hanl](https://openjdk.org/census#mhanl) (@Maran23 - Committer) Review applies to [8a4078fe](https://git.openjdk.org/jfx/pull/1575/files/8a4078fed1941bdd7052c1271079c4bfed378fba)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1575/head:pull/1575` \
`$ git checkout pull/1575`

`$ git checkout pull/1575` \
`$ git pull https://git.openjdk.org/jfx.git pull/1575/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1575`

`$ git pr show -t 1575`

</details>

<a href="https://git.openjdk.org/jfx/pull/1575.diff">https://git.openjdk.org/jfx/pull/1575.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1575#issuecomment-2361894421)


modules/javafx.controls/src/test/java/test/javafx/scene/control/AlertTest.java
```diff
@@ -92,7 +92,8 @@ private void showAndHideDialog(Dialog<?> dialog, boolean normalClose) {
                         Platform.runLater(() -> getStage(dialog).close());
                     }
                 } catch (InterruptedException e) {
-                    e.printStackTrace();
+                    // fail(e);
+                    throw new AssertionError(e);
```
kevinrushforth, line 96: I left the commented out "fail(e)" as a possibly better alternative once these tests are updated to JUnit 5, which is under review at PR #1569. Depending on the timing of integrating that PR and this one, I will either merge in master (I've already done a test merge and there are no merge conflicts) and update these calls to use "fail" or file a follow-on bug.

kevinrushforth:
Reviewer: @arapte or @andy-goryachev-oracle

tests/system/src/test/java/test/robot/javafx/scene/control/behavior/BehaviorRobotTestBase.java
```diff
@@ -320,14 +320,17 @@ protected Runnable shift(KeyCode k) {
...
 
     /**
-     * Convenience alias for Thread.sleep() that does not throw an exception.
+     * Convenience alias for Thread.sleep() that rethrows InterruptedException
+     * as an AssertionError.
      * @param ms the timout in milliseconds
      */
     protected void sleep(int ms) {
         try {
+            // KCR: this should be sleep(ms) not a hard-coded 1
```
andy-goryachev-oracle, line 329: good thing it's never called, but it is my fault.
this method was later replaced with `waitForIdle`

I can make this change as part of #1569

andy-goryachev-oracle (approved):
looks good.

I think it's better to wait until #1569 is integrated.

arapte (approved):
LGTM

tests/system/src/test/java/test/robot/javafx/scene/control/behavior/BehaviorRobotTestBase.java
kevinrushforth: If it isn't being used, then I like your idea of deleting it. Especially since this seems redundant given `Util.sleep(ms)`. I'll revert my changes and let you delete it as part of #1569.

kevinrushforth:
> I think it's better to wait until #1569 is integrated.

Either should be fine, since there is no merge conflict, but I can wait to integrate this until after you integrate that.

tests/system/src/test/java/test/robot/javafx/scene/control/behavior/BehaviorRobotTestBase.java
kevinrushforth: On closer inspection, I don't think this should be removed as part of PR #1569, since that file is already using JUnit 5 and is not touched by your PR.

Either I can remove the method with this PR or we just leave it for a follow-up. Removing it with this PR seems easiest, since I am already modifying that method as part of this PR (along with other methods that catch and ignore InterruptedException from sleep).

andy-goryachev-oracle: I agree with removing the method in this PR.  Thanks!

kevinrushforth: Done. Thanks.

andy-goryachev-oracle (approved):
looks good, thanks for removing the vestigial method.

this PR will likely need more changes once #1569 gets integrated (I would recommend to integrate this one after, simply because the other has more files)

Maran23 (approved):
Makes sense and is a good imrovement!

andy-goryachev-oracle:
#1569 has been integrated, we can make the final changes

kevinrushforth:
> #1569 has been integrated, we can make the final changes

Done.

tests/system/src/test/java/test/javafx/scene/text/ArabicWrappingTest.java
```diff
@@ -108,6 +109,7 @@ public void testWrapping() {
            try {
                Thread.sleep(2000);
            } catch (Exception e) {
+                fail(e);
```
andy-goryachev-oracle, line 112: something wrong with indent?

andy-goryachev-oracle (approved):
one minor formatting issue, otherwise looks good.

tests/system/src/test/java/test/javafx/scene/text/ArabicWrappingTest.java
kevinrushforth: Good catch. I'll fix.

andy-goryachev-oracle (approved):

PR #1393, opened 2024-03-06
8313138: Scrollbar Keyboard enhancement

andy-goryachev-oracle:
Adding alt-ctrl-LEFT/RIGHT/UP/DOWN (option-command-LEFT/RIGHT/UP/DOWN) key bindings to

- ListView
- TreeView
- TableView
- TreeTableView

to support keyboard-only horizontal and vertical scrolling.  The main reason for the change is to improve accessibility.

**NOTE: For controls in right-to-left orientation, the direction of horizontal scrolling is reversed.**

As far as I can tell, these key combinations do not interfere with editing.

The proposed solution can be further optimized by adding a public method to the VirtualFlow class, something like

```
public void horizontalUnitScroll(boolean right);
public void verticalUnitScroll(boolean down);
```

Q: Does this change require a CSR to explain the change in the controls' behavior?  We don't yet have the key bindings documented in /doc-files/behavior

Note:
Jenkins headful test passed on all mac configurations, failed on all linux configurations (master branch failed also, so it is test issue), while windows configuration is not yet available.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Change requires CSR request [JDK-8329336](https://bugs.openjdk.org/browse/JDK-8329336) to be approved
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issues
 * [JDK-8313138](https://bugs.openjdk.org/browse/JDK-8313138): Scrollbar Keyboard enhancement (**Enhancement** - P3)
 * [JDK-8329336](https://bugs.openjdk.org/browse/JDK-8329336): Scrollbar Keyboard enhancement (**CSR**)

### Reviewers
 * [Alexander Zuev](https://openjdk.org/census#kizune) (@azuev-java - Author) 🔄 Re-review required (review applies to [5bae5e7a](https://git.openjdk.org/jfx/pull/1393/files/5bae5e7a1725006fd8a9cb32993c67ce828b893b))
 * [Ambarish Rapte](https://openjdk.org/census#arapte) (@arapte - **Reviewer**)
 * [Kevin Rushforth](https://openjdk.org/census#kcr) (@kevinrushforth - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
[...]

PR #54, opened 2019-11-27
8212034: Potential memory leaks in jpegLoader.c in error case

arapte:
Memory allocated in initDecompressor() and decompressIndirect() is not freed in error case.
In error case,
1. Allocated memory should be freed.
2. Appropriate de-initialization jpeg library calls should be added.

Verified that,
1. All unit and systems tests pass on three platforms, and
2. Memory consumption with and without fix is similar by comparing memory before and after showing 10 jpeg images for 100 times.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed

### Issue
 * [JDK-8212034](https://bugs.openjdk.java.net/browse/JDK-8212034): Potential memory leaks in jpegLoader.c in error case

### Reviewers
 * Johan Vos ([jvos](@johanvos) - **Reviewer**)
 * Kevin Rushforth ([kcr](@kevinrushforth) - **Reviewer**)

### Download
`$ git fetch https://git.openjdk.java.net/jfx pull/54/head:pull/54`
`$ git checkout pull/54`

PR #1582, opened 2024-09-26
JDK-8340852 ScrollPane should not consume navigation keys when it doesn't have direct focus

hjohn:
This change modifies `ScrollPaneBehavior` to only consume keys that are targetted at it.  As `KeyEvent`s are in almost all cases only intended for the targetted node (as visually that's where the user expects the keyboard input to go, as per normal UI rules) consuming key events that bubble up is simply incorrect.  When the `ScrollPane` is focused directly (it has the focused border) then it is fine for it to respond to all kinds of keys.

In FX controls normally there is no need to check if a `Control` is focused (although they probably should **all** do this) as they do not have children that could have received the Key Events involved, and Key Events are always sent to the focused Node.  When `ScrollPane` was developed this was not taken into account, leading to it consuming keys not intended for it.

This fixes several unexpected problems for custom control builders.  A custom control normally benefits from standard navigation out of the box (TAB/shift+TAB) and directional keys.  However, this breaks down as soon as this custom control is positioned within a `ScrollPane`, which is very surprising behavior and not at all expected.  This makes it harder than needed for custom control developers to get the standard navigation for the directional keys, as they would have to specifically capture those keys before they reach the `ScrollPane` and trigger the correct navigation action themselves (for which as of this writing there is no public API).

The same goes for all the other keys captured by `ScrollPane` when it does not have focus, although not as critical as the UP/DOWN/LEFT/RIGHT keys.

### DIFF
--- a/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/ScrollPaneBehavior.java
+++ b/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/ScrollPaneBehavior.java
@@ -28,6 +28,9 @@ package com.sun.javafx.scene.control.behavior;
 import javafx.scene.control.ScrollBar;
 import javafx.scene.control.ScrollPane;
 import com.sun.javafx.scene.control.inputmap.InputMap;
+import com.sun.javafx.scene.control.inputmap.KeyBinding;
+
+import javafx.scene.input.KeyEvent;
 import javafx.scene.input.MouseEvent;
 import javafx.scene.control.skin.ScrollPaneSkin;
 
@@ -71,18 +74,18 @@ public class ScrollPaneBehavior extends BehaviorBase<ScrollPane> {
 
         // scrollpane-specific mappings for key and mouse input
         addDefaultMapping(inputMap,
-            new InputMap.KeyMapping(LEFT, e -> rtl(scrollPane, this::horizontalUnitIncrement, this::horizontalUnitDecrement)),
-            new InputMap.KeyMapping(RIGHT, e -> rtl(scrollPane, this::horizontalUnitDecrement, this::horizontalUnitIncrement)),
+            new InputMap.KeyMapping(new KeyBinding(LEFT), e -> rtl(scrollPane, this::horizontalUnitIncrement, this::horizontalUnitDecrement), this::isNotFocused),
+            new InputMap.KeyMapping(new KeyBinding(RIGHT), e -> rtl(scrollPane, this::horizontalUnitDecrement, this::horizontalUnitIncrement), this::isNotFocused),
 
-            new InputMap.KeyMapping(UP, e -> verticalUnitDecrement()),
-            new InputMap.KeyMapping(DOWN, e -> verticalUnitIncrement()),
+            new InputMap.KeyMapping(new KeyBinding(UP), e -> verticalUnitDecrement(), this::isNotFocused),
+            new InputMap.KeyMapping(new KeyBinding(DOWN), e -> verticalUnitIncrement(), this::isNotFocused),
 
-            new InputMap.KeyMapping(PAGE_UP, e -> verticalPageDecrement()),
-            new InputMap.KeyMapping(PAGE_DOWN, e -> verticalPageIncrement()),
-            new InputMap.KeyMapping(SPACE, e -> verticalPageIncrement()),
+            new InputMap.KeyMapping(new KeyBinding(PAGE_UP), e -> verticalPageDecrement(), this::isNotFocused),
+            new InputMap.KeyMapping(new KeyBinding(PAGE_DOWN), e -> verticalPageIncrement(), this::isNotFocused),
+            new InputMap.KeyMapping(new KeyBinding(SPACE), e -> verticalPageIncrement(), this::isNotFocused),
 
-            new InputMap.KeyMapping(HOME, e -> verticalHome()),
-            new InputMap.KeyMapping(END, e -> verticalEnd()),
+            new InputMap.KeyMapping(new KeyBinding(HOME), e -> verticalHome(), this::isNotFocused),
+            new InputMap.KeyMapping(new KeyBinding(END), e -> verticalEnd(), this::isNotFocused),
 
             new InputMap.MouseMapping(MouseEvent.MOUSE_PRESSED, this::mousePressed)
         );
@@ -143,6 +146,9 @@ public class ScrollPaneBehavior extends BehaviorBase<ScrollPane> {
         return Optional.ofNullable(((ScrollPaneSkin)getNode().getSkin()).getHorizontalScrollBar());
     }
 
+    private boolean isNotFocused(KeyEvent keyEvent) {
+        return !getNode().isFocused();
+    }
 
     /***************************************************************************
      *                                                                         *


kevinrushforth:
When this is ready for review, it will need plenty of testing. Since it is a behavioral change (albeit to fix a bug), we might consider whether it needs a CSR and/or a release note.

/reviewers 2

andy-goryachev-oracle:
Would it be possible to provide an SCCE that clearly illustrates the problem? This can be a base for a unit test in the PR.

Another question - do you think it might make sense to wait until we get the traversal policy API #1555 and instead install a policy by default which is identical to the current behavior, but at the same time it can be disabled or changed to a different policy?

hjohn:
I'm sorry, what is an SCCE?

As for the problem, I think I explained this several times now, but I'll attempt it again:

When the user uses the keyboard to control an application, there generally is a border or some other indication indicating the control that will be accepting those keyboard commands.  In FX, KeyEvents therefore are always send to the control that has the current focus, as that's where the user expects them to go.  For any controls that don't have children, you can assume that a KeyEvent that is received is intended for you -- your control is a leaf in the hierarchy, and so if you received the KeyEvent, you must have had the focus.

ScrollPane is different.  It is also a Control, but is rarely the target of the focus and even more rarely a leaf in a scene.  As such, KeyEvents that were not consumed by the actual target Control can bubble up to it.  If a Control wishes to use standard navigation provided by Scene, it must let KeyEvents bubble up.  If there is no ScrollPane in between anywhere, this works perfectly.  For example, take five Buttons, put them in a Scene in an HBox, remove the standard navigation behavior from them, and observe that even without these keys being part of the InputMap that the directional navigation works for these controls.

Now, do this same experiment but put a ScrollPane in between (ie. Scene -> ScrollPane -> HBox -> 5 buttons).  Observe that now directional navigation no longer works, despite one of the Buttons having the focus.  This is because the ScrollPane, even though it is not focused (no :focused pseudo state, no focused border) is responding to bubbled up KeyEvents intended for one of the Buttons.

Custom Control authors outside FX are facing this exact problem -- their Control works flawlessly with standard navigation out of the box when put in standard containers like HBox/VBox etc.  However, as soon as a ScrollPane is part of the hierarchy, the directional navigation will no longer function.  Again, this is because ScrollPane is consuming KeyEvents not targetted at it.  There is no good work-around for this problem, as custom control authors don't have access to input maps nor the traversal engine.  Their only option is to recreate the same navigation offered by FX by responding themselves to directional keys which is cumbersome.

But why are we working around this problem at all?  ScrollPane has no business consuming those keys when it is not focused itself.  Controls in FX are basically installing navigation keys in their input maps **because** they otherwise stop functioning correctly when a ScrollPane is in the hierarchy.  Navigation is not the business of individual controls, unless their very nature requires a different form of navigation (like ToggleButtons).

So in short: create a custom control, put two of them in a HBox and observe that you get directional navigation for free.  Now put a ScrollPane in that mix, observe that directional navigation no longer works.

> Another question - do you think it might make sense to wait until we get the traversal policy API https://github.com/openjdk/jfx/pull/1555 and instead install a policy by default which is identical to the current behavior, but at the same time it can be disabled or changed to a different policy?

I really don't think the current ScrollPane behavior has any place in FX, so I don't see a need to offer both the old and the new behavior.  I really feel we're going overboard here to avoid potentially breaking something when it is clearly not correct behavior from a UI point of view.  You can't make any changes at all anywhere to anything if the criteria is to not cause regressions in undocumented behavior.

As for installing custom policies, I'm still not convinced that's the right way to go.  I haven't seen any convincing use cases that aren't incredibly niche or are primarily there to work around other more fundamental problems (like this PR that attempts to address one of them).  The use cases I did see could be solved by offering navigation methods on `Scene`, and wouldn't be solved by custom policies (ie. if I receive an ActionEvent, and in response to it I want to navigate to the logical next control, how would policies help here?)

My take on all of this: if the navigation offered by default in Scene is insufficient for your control, then it is easy enough to look at the scene graph and pick your own next/previous control in response to a KeyEvent.  Can FX help here?   Sure, offer a few methods that given a Node find a logical/directional next/previous Node, or `null` if not found.  There is no need to wrap this all up in a policy:

In `Node`:

      Optional<Node> findFocusable(direction);

Or if you want to get really fancy offer a few standard traversal policies that can be accessed statically:

    public interface TraversalPolicy {
        Optional<Node> findFocusable(Node from);
    }

    public enum DirectionalNavigation implements TraversalPolicy {
        UP,
        DOWN,
        LEFT,
        RIGHT;

        @Override
        public Optional<Node> findFocusable(Node from) {
            Node next = null; // Calculate next

            return Optional.of(next);
        }
    }

    public enum LogicalNavigation implements TraversalPolicy {
        NEXT,
        PREVIOUS;

        @Override
        public Optional<Node> findFocusable(Node from) {
            Node next = null; // Calculate next

            return Optional.of(next);
        }
    }

    public enum InsaneCustomNavigation implements TraversalPolicy {
        RANDOM,
        DIAGONAL_THREE_DOWN;

        @Override
        public Optional<Node> findFocusable(Node from) {
            Node next = null; // Calculate next

            return Optional.of(next);
        }
    }

You can them super easily use these like this:

        Button button = new Button();

        button.setOnAction(e -> LogicalNavigation.NEXT.findFocusable(button).ifPresent(Node::requestFocus));

You wouldn't even need the interface `TraversalPolicy` unless you want to make them configurable for the above code.  With the interface you could do this:

        button.setOnAction(e -> configuredPolicy.findFocusable(button).ifPresent(Node::requestFocus));

andy-goryachev-oracle:
SCCE = Self-contained code example, or "complete code sample" in https://wiki.openjdk.org/display/OpenJFX/Submitting+a+Bug+Report

I am not entirely convinced that the proposed solution addresses compatibility concerns, unless this can be demonstrated.  I would imagine the idea behind SP handling these keys might have been a desire to be able to scroll the content in the view without changing the currently focused node within possibly?

Alternatively, should we use a different set of key bindings instead of simple arrow keys, like for instance alt-ctrl-LEFT/RIGHT/UP/DOWN (option-command-LEFT/RIGHT/UP/DOWN) key bindings, similarly to #1393 ?

hjohn:
> I do not see much behavioral difference between controls inside a regular VBox and VBox inside a ScrollPane
> 
> https://github.com/andy-goryachev-oracle/Test/blob/main/src/goryachev/research/FocusPolicyResearch.java
> 
> Am I doing things wrong?

Yeah, as I explained, FX provided controls, like `ToggleButton` will capture navigation keys **before** they bubble up to `ScrollPane`.  See here in `ToggleButtonBehavior`:

            new KeyMapping(RIGHT, e -> traverse(e, "ToggleNext-Right")),
            new KeyMapping(LEFT, e -> traverse(e, "TogglePrevious-Left")),
            new KeyMapping(DOWN, e -> traverse(e, "ToggleNext-Down")),
            new KeyMapping(UP, e -> traverse(e, "TogglePrevious-Up"))
            
So, with FX provided controls, this issue is being worked around.  The controls themselves handle navigation, and so `ScrollPane` won't interfere.  `ToggleButton` however is not a good example, as it has special needs for its navigation (which I hope are documented) and so is correct to override how navigation works.

Using a regular `Button` instead for your tests would be better, however this also won't work as FX provided controls are all providing their own navigation to avoid letting events bubble up to `ScrollPane`.

To make a real comparison, you need to either:

1. Remove the navigation behavior from one of the FX controls, ie. in `ButtonBehavior`, remove the line:

        // add focus traversal mappings
        addDefaultMapping(buttonInputMap, FocusTraversalInputMap.getFocusTraversalMappings());

2. Set a different skin on `Button` (which will also kill its standard behavior).

I'll provide an example.

hjohn:
> SCCE = Self-contained code example, or "complete code sample" in https://wiki.openjdk.org/display/OpenJFX/Submitting+a+Bug+Report

Ah, thank you, I hadn't seen this before.

> I am not entirely convinced that the proposed solution addresses compatibility concerns, unless this can be demonstrated. I would imagine the idea behind SP handling these keys might have been a desire to be able to scroll the content in the view without changing the currently focused node within possibly?

That would be counter to how all scrollable views in almost all UI's work.  Generally, a ScrollPane only scrolls when it has the focus itself.  As an example, in Eclipse, use the Package Explorer and ensure it has a horizonal scroll bar.  Now use the mouse to scroll to the right.  Now press the left arrow key to go "up the tree" and close nodes.  You'll notice that when you are at a root item, and pressing left again will not "bubble" up to the ScrollPane and it starts scrolling left.

At the risk of repeating myself, this is not a matter of what the original programmers may have thought (it is highly likely an oversight, considering how bubbling works and how you don't need to worry about with this in "normal" controls), but what people expect from UI's.  The focused component deals with key presses, and nothing else does.  The only exceptions to these are well understood:

- Menu Hotkeys (usually with a specific kind of menu modifier key as it would otherwise be very confusing)
- Underscored keys to change focus (also only with a modifier, and made visual to the user)
- Navigation
 
> Alternatively, should we use a different set of key bindings instead of simple arrow keys, like for instance alt-ctrl-LEFT/RIGHT/UP/DOWN (option-command-LEFT/RIGHT/UP/DOWN) key bindings, similarly to #1393 ?

That's certainly possible, if that's considered normal for such a control on a given platform.  AFAIK, FX is still doing its best to emulate how controls work on the platforms it supports (as evidenced by specific mac/linux/windows bindings in many behaviors).

On Windows however, I'm not aware of any such bindings.  I've also looked into the Mac platform, and I could not find any combinations that would make the container scroll while another control within it has focus (like an edit box for example).

hjohn:
Here is a minimal working example that demonstrates the issue when people create a custom control.

Note that the standard controls and the custom control all have directional navigation working, except in the case where the custom control is embedded inside a `ScrollPane`.  This creates a big burden on custom control developers as to make their control work when used in a scroll pane, they now need to implement their own navigation...

```java
package app;

import javafx.scene.Node;
import javafx.scene.Scene;
import javafx.scene.control.ScrollPane;
import javafx.scene.control.Skin;
import javafx.scene.layout.GridPane;
import javafx.scene.layout.StackPane;
import javafx.stage.Stage;

public class App extends Application {

    @Override

        GridPane gp = new GridPane(10, 10);

        gp.add(new VBox(new Label("Standard Buttons in normal container"), new HBox(5, new Button("A"), new Button("B"), new Button("C"))), 0, 0);
        gp.add(new VBox(new Label("Standard Buttons in ScrollPane"), new ScrollPane(new HBox(5, new Button("A"), new Button("B"), new Button("C")))), 1, 0);
        gp.add(new VBox(new Label("Custom Buttons in normal container"), new HBox(5, new CustomButton("A"), new CustomButton("B"), new CustomButton("C"))), 0, 1);
        gp.add(new VBox(new Label("Custom Buttons in normal container"), new ScrollPane(new HBox(5, new CustomButton("A"), new CustomButton("B"), new CustomButton("C")))), 1, 1);

        Scene scene = new Scene(gp);

        primaryStage.setScene(scene);

        primaryStage.show();
    }

    static class CustomButton extends Button {

        CustomButton(String title) {
            super(title);

            setSkin(new CustomButtonSkin(this));
        }
    }

    static class CustomButtonSkin implements Skin<Button> {
        private final Button control;
        private final StackPane container;

        public CustomButtonSkin(Button button) {
            this.control = button;
            this.container = new StackPane();
            this.container.getChildren().add(new Label(button.getText()));
        }

        @Override
        public Button getSkinnable() {
            return control;
        }

        @Override
        public Node getNode() {
            return container;
        }

        @Override
        public void dispose() {
            container.getChildren().clear();
        }
    }
}
```

andy-goryachev-oracle:
> Here is a minimal working example

Very helpful, thank you.

I would say the use of arrow keys to traverse focus in this case _looks to me_ to be application-specific.  In other words, this may or may not be the desired behavior.

In the context of a simple dialog which only has buttons - yes, it `is` the desired behavior.  In the case of a more complex UI where text input components are present - probably not, but it may depend on the application requirements.

In your example, the navigation with TAB / shift+TAB is there, and it works correctly in all four panes.  So it's not like the functionality is severely impeded.

What I am saying is that what we currently have is probably ok and JDK-8340852 may not be needed.

andy-goryachev-oracle:
I would like to get feedback from a wider audience, if possible.

A cursory search in JBS showed no complaints about ScrollPane consuming the events.

There is a bunch of issues related to focus traversal and/or arrow keys though:

[JDK-8087583](https://bugs.openjdk.org/browse/JDK-8087583)
https://bugs.openjdk.org/browse/JDK-8284878
[JDK-8091294](https://bugs.openjdk.org/browse/JDK-8091294)
[JDK-8091825](https://bugs.openjdk.org/browse/JDK-8091825)

andy-goryachev-oracle:
Also: swing apparently does not use arrow keys in a similar situation:

https://github.com/andy-goryachev-oracle/Test/blob/main/src/goryachev/research/JScrollPane_Buttons.java

hjohn:
> > Here is a minimal working example
> 
> Very helpful, thank you.
> 
> I would say the use of arrow keys to traverse focus in this case _looks to me_ to be application-specific. In other words, this may or may not be the desired behavior.
> 
> In the context of a simple dialog which only has buttons - yes, it `is` the desired behavior. In the case of a more complex UI where text input components are present - probably not, but it may depend on the application requirements.

I am sorry, you're saying that you think it is okay that setting a Skin suddenly makes your control malfunction, but only if it has a `ScrollPane` somewhere in its hierarchy?  Where is the logic exactly?  Should it not be either:

- All navigation stops working for custom controls, regardless of placement
- Navigation keeps working for custom controls, regardless of placement

... this in between stance is baffling me, and seems to be just defending the status quo in favor of a work-around solution that is not addressing core problems.

The fact that I used buttons has no bearing on this.  Other controls will have the exact same problems, if they let directional keys bubble up.  For text input controls this is usually not the case as they have other purposes for those keys, but that may change with read only status, or if they want to say allow scrolling from one text component to the next.

> In your example, the navigation with TAB / shift+TAB is there, and it works correctly in all four panes. So it's not like the functionality is severely impeded.

Yes, I never claimed tab didn't work. I only claimed that FX is duplicating navigation everywhere in its own controls, just so users wouldn't have to deal with ScrollPane affecting navigation keys.

It sounds like you would be okay with having FX controls behave in the same way, then at least it would be consistent. So let's remove directional navigation from all FX controls, and only have that bubble up to Scene when there is no ScrollPane in between.  Then custom controls and FX controls would be on the same footing.

Or, alternatively, let's remove default navigation from Scene and make navigation not a global concern (as it should really be) but the business of each and every control built..
 
> What I am saying is that what we currently have is probably ok and JDK-8340852 may not be needed.

I'm unable to follow your reasoning.  If you could explain in more detail why you think the current system should:

a) Have navigation decentralized in every control
b) Should lose only half of the standard navigation keys (not all, or none) when a Skin is assigned
c) Custom skinned controls have different navigation when placed in ScrollPane while FX controls don't

kevinrushforth:
I tend to think John is right and that the current behavior of ScrollPane is a bug. In general, dropping a control into a ScrollPane shouldn't change aspects of the behavior that aren't directly related to the functioning of the ScrollPane. The fact that JavaFX's controls have to install their own keyboard handler to get what would otherwise be the default behavior underscores this.

I'll admit that I haven't looked at this in enough detail to be sure, but I don't see any good reason for a non-focused ScrollPane to consume keyboard events that bubble up to it.

beldenfox:
I agree that the current ScrollPane implementation is wrong. Traversal is an accessibility issue so it’s imperative that we don’t block the traversal keys. We can’t expect custom controls to work around the ScrollPane's faulty behavior.

I don’t agree with the idea that a ScrollPane should only process key events when it has the focus largely because I can’t think of an instance where a ScrollPane receives focus. Focus either lies on the node that owns the ScrollPane (like a TableView) or one of the nodes inside the ScrollPane but not on the ScrollPane itself.

In any case there are instances in any UI toolkit where key event handling is split between a control and non-focused containers further up the tree (for example, I just verified that on macOS the NSScrollView implements PgUp and PgDn but never has focus). I haven’t dug into JavaFX controls in depth but I would be surprised if there weren’t similar instances. It’s a pretty common occurrence.

If we remove the existing PgUp, PgDn, and Home behavior folks will notice and accessibility will take a hit. I’m not entirely sure anyone will notice if we remove the arrow key behavior since the other Controls work so hard to hide them from the ScrollPane to begin with. I do agree that these bindings need to be removed OR completely re-implemented so they don’t block traversal (e.g. they could try to traverse and scroll if traversal isn't possible).

TL;DR Don’t try to filter events based on whether the ScrollPane has focus. Remove the ScrollPane bindings that block the traversal keys. Leave the rest.

andy-goryachev-oracle:
> I can’t think of an instance where a ScrollPane receives focus.

when there is a non-interactive content inside (such as image or a label).
they you *do* want to handle arrow keys to scroll.

andy-goryachev-oracle:
> Focus either lies on the node that owns the ScrollPane (like a TableView) or one of the nodes inside the ScrollPane but not on the ScrollPane itself.

this seems to be incorrect: unless you set `mouseTransparent`, ScrollPane can be focused with a mouse click, even when `focusTraversable` is false.

beldenfox:
> > Focus either lies on the node that owns the ScrollPane (like a TableView) or one of the nodes inside the ScrollPane but not on the ScrollPane itself.
> 
> this seems to be incorrect: unless you set `mouseTransparent`, ScrollPane can be focused with a mouse click, even when `focusTraversable` is false.

I didn't say that a ScrollPane couldn't take focus, just that it normally doesn't. And if focus lies with a node inside the ScrollPane we can't expect a user to move focus to the ScrollPane to get PgUp and PgDn working.

> > I can’t think of an instance where a ScrollPane receives focus.
> 
> when there is a non-interactive content inside (such as image or a label). they you _do_ want to handle arrow keys to scroll.

Unless that would make the ScrollPane a traversal dead-end.

It's a conundrum. There are cases where the arrows should scroll particularly for top-level content (I can use them to scroll through a web page in Safari). But from what I can tell this is not the default behavior in the macOS toolkit probably because it interferes with traversal.

Whatever behavior we want the way it's implemented in JavaFX at the moment is wrong.

BTW, I'm not optimistic that we can find some combination of modifiers + arrow keys to use. The behavior of Shift is already defined and many of the other modifier + arrow combinations are used by one OS or another.

andy-goryachev-oracle:
> Whatever behavior we want the way it's implemented in JavaFX at the moment is wrong.

I changed my mind and I agree with you and John that this change makes sense.

> BTW, I'm not optimistic that we can find some combination of modifiers + arrow keys to use. The behavior of Shift is already defined and many of the other modifier + arrow combinations are used by one OS or another.

There was a recent change made to a number of scrollable views (ListView, TreeView, TableView, TreeTableView) to add alt-ctrl-LEFT/RIGHT/UP/DOWN (option-command-LEFT/RIGHT/UP/DOWN) key bindings just for that.

Perhaps we could apply the same idea to the ScrollPane.

andy-goryachev-oracle:
just in case, my variant:
https://github.com/andy-goryachev-oracle/Test/blob/main/src/goryachev/research/ScrollPane_ConsumeNavEvents2.java

hjohn:
> I agree that the current ScrollPane implementation is wrong. Traversal is an accessibility issue so it’s imperative that we don’t block the traversal keys. We can’t expect custom controls to work around the ScrollPane's faulty behavior.
> 
> I don’t agree with the idea that a ScrollPane should only process key events when it has the focus largely because I can’t think of an instance where a ScrollPane receives focus. Focus either lies on the node that owns the ScrollPane (like a TableView) or one of the nodes inside the ScrollPane but not on the ScrollPane itself.

ScrollPane however has no idea how to behave correctly to process those keys.  ScrollPanes are so universal that processing keys like Home/End, PgUp/PgDown cannot be given a proper meaning without knowing what they contain.

Examples:
- A ScrollPane contains a street map, currently centered on a location that's being tracked. 
  - Press PgUp -- what do we do?  Moving the vertical bar "one 'page' up" would leave you without seeing the location you'd want to see.  Zooming out may make more sense.  Doing nothing may also make more sense, yet, this key **cannot** be disabled...
  - Press Home or End... where do we go?  Home probably should center on the location being tracked.  End has no purpose whatsoever.  Scrolling to the top makes zero sense.  Should this key scroll to just the top, or top left is another question you can ask... what if the map is infinitely large? Where do you scroll to then?

- A ScrollPane contains a bunch of focusable control (like a variable amount of input fields and buttons).  Let's say we've scrolled down a bit and a Button is focused.  Now press pgup/pgdown/home or end.  End result: button remains focused, but is now no longer visible in the pane.

- A ScrollPane contains a rich text editable area, like Java code.
  - Press PgUp/Down: ScrollPane has no notion of lines or units, nor does it know how large the font is that is being used.  It scrolls but ends up displaying the top line only half instead of showing only discrete lines.
 
- A ScrollPane contains a 3D rotatable representation of a building. What should the pgup/pgdown/home/end keys do?

- A ScrollPane is nested within another ScrollPane.  What should pressing Home do?  First scroll the inner scroll pane to the top, then on a 2nd press the outer ScrollPane?  Scroll both immediately to their Home position?  Neither happens currently, and it seems unlikely that a ScrollPane abstraction could possibly know what the correct thing to do is here without knowing anything about its contents.

- A ScrollPane contains **actual** pages (zoomed out, like in say Word).
  -  PgUp/PgDown will not be aware of this and scroll up/down a self determined amount.  This will **not** end nicely at a page boundary... you could "configure" this on the ScrollPane, but it would be near impossible to get right as different units are used, nor could it handle something like larger and smaller pages (or portrait/landscape pages).

What the above examples show primarily is that what a `ScrollPane` should do is highly dependent on its content.  Applications aiming to be accessible should therefore define their own handling for accessible keys.  Standard View controls that incorporate a scrollable area should handle such keys themselves (and they do), here's an excerpt from `ListViewBehavior`:

            new KeyMapping(new KeyBinding(HOME), e -> selectFirstRow(), isInEditableComboBox),
            new KeyMapping(new KeyBinding(END), e -> selectLastRow(), isInEditableComboBox),
            new KeyMapping(new KeyBinding(HOME).shift(), e -> selectAllToFirstRow(), isInComboBox),
            new KeyMapping(new KeyBinding(END).shift(), e -> selectAllToLastRow(), isInComboBox),
            new KeyMapping(new KeyBinding(PAGE_UP).shift(), e -> selectAllPageUp()),
            new KeyMapping(new KeyBinding(PAGE_DOWN).shift(), e -> selectAllPageDown()),

I'm afraid that the current key handling 'behavior' of `ScrollPane` has been added with a very narrow point of view that it would only be used to display something like a list, that has sensible home and end locations and units suitable to paging.  Even when used with something that may look like a list, `ScrollPane` has no notion of "rows" or "pages" and so will just scroll some fixed (although configurable) amount.

> In any case there are instances in any UI toolkit where key event handling is split between a control and non-focused containers further up the tree (for example, I just verified that on macOS the NSScrollView implements PgUp and PgDn but never has focus). I haven’t dug into JavaFX controls in depth but I would be surprised if there weren’t similar instances. It’s a pretty common occurrence.

I'd be interested in how you verified this.  AFAIK, contained controls will handle paging (like NSTextView).

> If we remove the existing PgUp, PgDn, and Home behavior folks will notice and accessibility will take a hit. I’m not entirely sure anyone will notice if we remove the arrow key behavior since the other Controls work so hard to hide them from the ScrollPane to begin with. I do agree that these bindings need to be removed OR completely re-implemented so they don’t block traversal (e.g. they could try to traverse and scroll if traversal isn't possible).

Trying to traverse, then scroll if you can't would just be annoying.  You'd press Up to go the previous control (perhaps expecting it to "wrap around" or stop at the first control if holding the key), but there isn't any and instead the system does something else which you didn't ask for (scrolls up a bit).  Assigning such double meanings to keys without a visual clue what will happen will only confuse users.  This is why we have the focus indicator, and that's why a ScrollPane should not be acting upon keys by default unless it visually has the focus.

> TL;DR Don’t try to filter events based on whether the ScrollPane has focus. Remove the ScrollPane bindings that block the traversal keys. Leave the rest.

That could be a compromise for now.

However, the larger point still stands: As shown in the examples, I doubt that a ScrollPane, being a very low level component, could ever sensibly implement the keys that it currently does without knowing what it is displaying.  Again, the View controls in FX don't rely on ScrollPane for Home/End/PgUp/PgDown either.  Nobody should rely on these to be honest.  If your situation requires these keys to do something, then you should add these yourself (where it makes sense, this could be at the ScrollPane level, or on any focusable controls within the pane).

I wish I could provide a "one size fits all" key handling for `ScrollPane`, but I don't think that's realistic.  At least with this PR, we're making it possible to provide your own implementation that may make more sense, as you can now actually capture those keys when they arrive at a `ScrollPane` without the default handler consuming them first (as with this PR it will only consume them when it has focus).

Currently, your only other option to "fix" this is to create a ScrollPane subclass and completely reimplement its Skin and, as you can't copy its behavior since those are private, you then won't get the default event handler that steals your keys.

---

Here's an example illustrating one of the problems with a ScrollPane's default scroll behavior.  A window opens with 20 buttons in a ScrollPane.  Make the Window small enough.  Then Scroll halfway.  Select a Button.  Now press keys like Home/End and see how the focused control disappears from view.  More sensible actions would be: pgup/pgdown focus a button 5 down or 5 up.  Home/End: focus first, last button.

```java
package app;

import javafx.scene.Scene;
import javafx.scene.control.ScrollPane;
import javafx.stage.Stage;

public class App2 extends Application {

    @Override
        VBox vbox = new VBox(new Label("Standard Buttons in ScrollPane"), new ScrollPane(
            new VBox(5, createButtons())
        ));

        Scene scene = new Scene(vbox);

        primaryStage.setScene(scene);

        primaryStage.show();
    }

    Button[] createButtons() {
        Button[] buttons = new Button[20];

        for(int i = 0; i < 20; i++) {
            buttons[i] = new Button("Button " + (i + 1));
        }

        return buttons;
    }
}
```

hjohn:
> > > Focus either lies on the node that owns the ScrollPane (like a TableView) or one of the nodes inside the ScrollPane but not on the ScrollPane itself.
> > 
> > 
> > this seems to be incorrect: unless you set `mouseTransparent`, ScrollPane can be focused with a mouse click, even when `focusTraversable` is false.
> 
> I didn't say that a ScrollPane couldn't take focus, just that it normally doesn't. And if focus lies with a node inside the ScrollPane we can't expect a user to move focus to the ScrollPane to get PgUp and PgDn working.
> 
> > > I can’t think of an instance where a ScrollPane receives focus.
> > 
> > 
> > when there is a non-interactive content inside (such as image or a label). they you _do_ want to handle arrow keys to scroll.
> 
> Unless that would make the ScrollPane a traversal dead-end.
> 
> It's a conundrum. There are cases where the arrows should scroll particularly for top-level content (I can use them to scroll through a web page in Safari). But from what I can tell this is not the default behavior in the macOS toolkit probably because it interferes with traversal.

It makes sense to not make something that doesn't apply in all situations the default behavior.   A browser is mostly a read-only affair, and consuming directional keys for scrolling makes a lot of sense.  They don't allow directional navigation for focusable controls, except in very specific cases (like an embedded toolbar) so they are not dealing with this conundrum at all.

The toolkit browsers are built on however do allow directional navigation (their toolbars and other native controls support it), but it was a **choice** to not to do this on the web page itself (the nested application).  It was also a choice there to not be able to escape the nested application easily (when tabbing through things on the web page, you wouldn't want to accidentally tab to the back button or address bar, that would break the immersion of the nested application).  Instead browsers provide special short cuts to "escape" the web page (F6 to focus the address bar for example).

FX is similar to most other toolkits.  It provides directional navigation as that makes sense for productivity applications.  It however sets up its scroll pane's default behavior to act like a web browser would.  It is taking away the choice here, as you can't disable this easily without making a new skin for `ScrollPane`.  It would have been better to not set such an opiniated default that can't be overridden easily.  In the opposite situation, users can easily install handlers to make the scroll pane act the way they want (and deal with the consequences of their actions by consuming certain keys before they bubble up).  So to sum up there are two options:

- ScrollPane aggressively consumes navigational and other keys **[current]**
  - To prevent this, must create a new skin for ScrollPane to get rid of its default behavior; possibly this can be improved with a behavior or input map proposal, but the main point stands: why have such an opiniated default that doesn't apply in all situations?
  - Or alternatively: have all embedded controls handle such keys before they reach the scroll pane (what FX does with the controls it provides in order to get the expected result of navigation before scrolling).

- ScrollPane by default remains impassive when not focused **[proposed]**
  - To get the existing behavior, just install an event handler on ScrollPane (a choice the user makes that will block directional navigation in favor of scrolling)
  - If you want to allow directional navigation first, then the handler should be installed on Scene.  Any left over keys can then be converted into actions for a specific scroll pane (probably a top level one, browser style)
  - If you want to have exact control over the scrolls a scroll pane should do, then consider handling those at the embedded control that is aware of how its content is best represented (like ListView, TableView) -- this is how pgup/pgdown/home/end should be handled -- a scroll pane is not smart enough to make such keys behave sensibly in all situations

> BTW, I'm not optimistic that we can find some combination of modifiers + arrow keys to use. The behavior of Shift is already defined and many of the other modifier + arrow combinations are used by one OS or another.

Yeah, I think we definitely should not do this as part of this problem.  A scroll pane is too generic; it makes much more sense to leave this up to the specific implementation instead (ie. TableView, ListView).  There is little sense to provide something out of the box that for any realistic control probably needs to be overridden to do what the user expects.

hjohn:
> I played a bit more with your example. Your CustomButtonSkin is weird as it does not place the original button in the scene graph, not sure if that's important.

I think there is a misunderstanding.  The `CustomButton` **is** part of the scene graph (it's added by the user).  It is after all the container in which the Skin is placing the controls that give this custom control its appearance.  The easiest way to give it an appearance is to add a Text or Label, which I did.  Adding the Node provided to the Skin to the scene graph again (instead of things like shapes, texts, labels, etc) would create a cycle (and FX wil throw an immediate exception :-) )

The structure is:

     CustomButton (situated in a container somewhere in the Scene, added by the user to the scene graph)
        |
        \--> Label (displaying the button text, added by the Skin)

The default skins work similar.

> I've added a number of custom components to the example
> 
> ```
>     static class L extends Label {
>         public L(String s) {
>             super(s);
>             setFocusTraversable(true);
>             setMouseTransparent(false);
>         }
>     }
> ```
> 
> together with monitoring the scene's focusOwnerProperty (don't want to mess with CSS).
> 
> These "L" components do traverse with the arrow keys and your fix with and without the ScrollPane. So perhaps I am wrong and we do want this change.

Yeah, without the fix these would also need to install navigational keys in their input maps just in case someone made them focus traversable.  It's a good point that I didn't think of to justify this change, thanks for finding it.

> A couple of notes: the existing example, at least when considering its top two panels with the regular buttons, exhibits reduced accessibility when the scroll pane shows a scroll bar. One can still move focus and the scroll bar does scroll the focused component to view, but one cannot induce it to scroll (for example, if one adds a long label after the last button).

I also discovered the directional navigation seems to bug and skip controls if you reduce the spacing in the HBox's to zero, but that's a separate problem for sure.

andy-goryachev-oracle:
> or perhaps indeed something with traversal policies would need to be created by the user to solve this on a case by case basis. It will be hard to always do the expected thing with a control as generic as ScrollPane.

I played with your examples and this PR a bit more.  I am now convinced you are right and we should fix the ScrollPane as you proposed (with a minor suggestion).

On the traversal subject, the keyboard navigation in FX is sometimes weird, which further confirms the need for a public traversal APIs (#1555).  For example, arrow down unexpectedly navigates away from the scroll pane content into the side panel (with or without this PR):

Screenshot 2024-10-03 at 09 30 13

hjohn:
> > or perhaps indeed something with traversal policies would need to be created by the user to solve this on a case by case basis. It will be hard to always do the expected thing with a control as generic as ScrollPane.
> 
> I played with your examples and this PR a bit more. I am now convinced you are right and we should fix the ScrollPane as you proposed (with a minor suggestion).
> 
> On the traversal subject, the keyboard navigation in FX is sometimes weird, which further confirms the need for a public traversal APIs (#1555). For example, arrow down unexpectedly navigates away from the scroll pane content into the side panel (with or without this PR):

The directional navigation does a best fit approach here, it looks for the closest target in a 90 degree arc from the origin to find the nearest matching target -- I think it just clips the HBar policy dropdown, which is nearest, so it goes there.  Otherwise it would have gone most likely to the HMax drop down.  It is not aware of things like rows or layout :)

image

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/ScrollPaneBehavior.java
```diff
@@ -71,18 +74,18 @@ public ScrollPaneBehavior(ScrollPane scrollPane) {
...
-            new InputMap.KeyMapping(PAGE_DOWN, e -> verticalPageIncrement()),
-            new InputMap.KeyMapping(SPACE, e -> verticalPageIncrement()),
+            new InputMap.KeyMapping(new KeyBinding(PAGE_UP), e -> verticalPageDecrement(), this::isNotFocused),
+            new InputMap.KeyMapping(new KeyBinding(PAGE_DOWN), e -> verticalPageIncrement(), this::isNotFocused),
+            new InputMap.KeyMapping(new KeyBinding(SPACE), e -> verticalPageIncrement(), this::isNotFocused),
 
-            new InputMap.KeyMapping(HOME, e -> verticalHome()),
-            new InputMap.KeyMapping(END, e -> verticalEnd()),
+            new InputMap.KeyMapping(new KeyBinding(HOME), e -> verticalHome(), this::isNotFocused),
+            new InputMap.KeyMapping(new KeyBinding(END), e -> verticalEnd(), this::isNotFocused),
```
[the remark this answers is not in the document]
hjohn: It's an interesting suggestion, but it is not needed.  `javac` will already deduplicate these.

You can even verify that this is the case.  Use `javap` to decompile the class file with `javap -c <classname>`.  In there, `invokedynamic` is used to represent the lambda's.  It looks like this for example:

     120: invokedynamic #54,  0             // InvokeDynamic #1:test:(Lcom/sun/javafx/scene/control/behavior/ScrollPaneBehavior;)Ljava/util/function/Predicate;

Later on, you'll see another:

     152: invokedynamic #54,  0             // InvokeDynamic #1:test:(Lcom/sun/javafx/scene/control/behavior/ScrollPaneBehavior;)Ljava/util/function/Predicate;

What you can see here is that the same constant (# 54) is used to reference the method. So, there's no need to help the compiler here.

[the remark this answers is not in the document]
andy-goryachev-oracle: hmmm... when I set up a breakpoint in Eclipse in KeyMapping:785 (after it hits the ScrollPaneBehavior<init>) the value for `interceptor` changes

```
interceptor= 0x0000007001340d50  (id=208)	
interceptor= 0x00000070013411c0  (id=212)	
interceptor= 0x0000007001341630  (id=216)	
```

I think it's still creates a different lambda object.  (I recall testing this assumption with some unit test a while back).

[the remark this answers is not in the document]
andy-goryachev-oracle: yep, it's a different object.  the constant you are referring to is just a name of the method, I think.  here:

```
public class AppTestLauncher {
    public static void main(String[] args) throws Throwable {
        Runnable a = AppTestLauncher::func;
        Runnable b = AppTestLauncher::func;
        System.out.println("a=" + a + " b=" + b + " == " + (a == b));
    }
    
    static void func() { }
```

outputs

`a=goryachev.apps.AppTestLauncher$$Lambda/0x000000f801008c78@16c0663d b=goryachev.apps.AppTestLauncher$$Lambda/0x000000f801008e90@23223dd8 == false
`

andy-goryachev-oracle (approved):
Thank you John for explanations, patience, and the fix!

andy-goryachev-oracle:
I don't think this change needs a CSR, but should be definitely mentioned in the release notes.

kevinrushforth:
> Thank you John for explanations, patience, and the fix!

+1 to this!

> I don't think this change needs a CSR, but should be definitely mentioned in the release notes.

I had pretty much come to the same conclusion. I added the `release-note=yes` label in JBS, and will confirm whether or not it needs a CSR when I review the PR. One other thing is that the JBS issue should be changed from Enhancement to Bug, so I did that.

beldenfox:
Again, it’s really, really rare for a ScrollPane to have focus. Making these bindings contingent on focus is the same as removing them altogether.

These are behavioral changes and developers will expect a rationale for each change. The removal of the arrow key bindings is an easy sell; the bindings interfere with traversal. And as we've discovered most toolkits don't implement these keys by default.

Two rationales have been presented for removing PgUp, PgDn, Home, and End. One is that it’s inappropriate for a node to process a key event when it doesn’t have direct focus. This is not a generally accepted rule in any UI toolkit I’ve ever worked with (just think of how Shift+Right to extend the selection is implemented in most table views). The other is that since the default behavior isn’t always correct we should remove it. But that’s true for just about any default behavior we provide. These are not strong arguments for removing existing behaviors.

If nothing else we should make removing PgUp, PgDn, Home, End, and Space a separate issue.

andy-goryachev-oracle:
> Again, it’s really, really rare for a ScrollPane to have focus.

incorrect.  if you have a non-interactive scroll pane content and click on it, the scroll pane gets focus and handles the keys correctly.  and even if there is interactive content, clicking on any pane inside causes the scroll pane to receive focus.

you can try with the monkey tester, as it allows for changing various properties at run time (focusTraversable, mouseTransparent, etc):

Screenshot 2024-10-04 at 14 37 20

hjohn:
> Again, it’s really, really rare for a ScrollPane to have focus. Making these bindings contingent on focus is the same as removing them altogether.
> 
> These are behavioral changes and developers will expect a rationale for each change. The removal of the arrow key bindings is an easy sell; the bindings interfere with traversal. And as we've discovered most toolkits don't implement these keys by default.
> 
> Two rationales have been presented for removing PgUp, PgDn, Home, and End. One is that it’s inappropriate for a node to process a key event when it doesn’t have direct focus. This is not a generally accepted rule in any UI toolkit I’ve ever worked with (just think of how Shift+Right to extend the selection is implemented in most table views). 

That will not change, those keys are processed by the focused control.  Using Shift + Navigation key to extend selection will keep working as normal and the pane will scroll if necessary, but not because the scroll pane is handling that key (it would have no clue what to do on shift + right) but because the control involved will try to keep the cursor visible.

> The other is that since the default behavior isn’t always correct we should remove it. But that’s true for just about any default behavior we provide. These are not strong arguments for removing existing behaviors.

It's more than that, most FX controls will not let these keys bubble up at all.  

- None of the `*View` controls let any of these keys bubble up, because the default scroll behavior is not suitable (they want to scroll by rows, or real pages, and so all those keys are handled directly).
- Editable fields will consume all keys (yes, a TextField consumes pgup/pgdown even though it has no use for them).  This is probably desirable as it would be very annoying to have a wrong key press scroll your pane.
- Other FX controls, like Buttons, or controls that have been made focus traversable (like Labels) do let these keys bubble up, but it is probably very questionable to let scroll pane act on them for the simple reason that the focused control can now be scrolled out of view and become invisible... now we're in a state where the user does not know what is focused, and has no idea what their next key press will do (assuming they have a short memory or left the application like this for a while).

So for FX controls that have scrolling behavior baked in, nothing changes.  For controls that do not, either they all consume those keys or the scroll behavior you get by default is highly undesirable as it leaves the user guessing what has focus.

Now, **custom** controls are a different matter.  These may have relied on this behavior, and I think that's why this change should be mentioned explicitly in the release notes.  They no longer get "scrolling" for free.  However, I think even for these controls the default behavior is undesirable. We're again talking about focusable controls, and letting scroll pane have its way will mean those controls can scroll out of view.  If it is a single view filling control (like a `*View`) it is likely these keys have been overridden by the custom control to provide better than default behavior.

However, to say no custom controls will be unimpacted by this change is a stretch.  There may indeed be custom controls that liked and desired the standard scroll pane behavior.  These will need to be modified to capture these keys themselves and provide appropriate scrolling.  This is easy to do however.

> If nothing else we should make removing PgUp, PgDn, Home, End, and Space a separate issue.

We can do that, if the above hasn't convinced you.  We'd be rehashing all these points likely again though:
- Default behavior is likely wrong (pane is unaware of what is displayed, and will make odd jumps)
- Default behavior can scroll view away from focused control, leaving the user guessing what their key presses will do next once focused control becomes invisible
- Default behavior is overridden in all FX controls that offer scrolling
- Default behavior is likely overridden in custom controls where scrolling makes sense to get better scroll positioning
- You still get the default behavior when displaying something that can't get focus (ie. a map, or an image, or a big piece of text) if of course you allowed the scroll pane to get focus in the first place

I also checked some controls outside FX.  In ControlsFX at least they seem to rely on internal skins for most of their scrollable controls, and require an `add-opens` to make use of the library with FX.  This means they are likely not impacted by this change either.

beldenfox:
> > Two rationales have been presented for removing PgUp, PgDn, Home, and End. One is that it’s inappropriate for a node to process a key event when it doesn’t have direct focus. This is not a generally accepted rule in any UI toolkit I’ve ever worked with (just think of how Shift+Right to extend the selection is implemented in most table views).
> 
> That will not change, those keys are processed by the focused control. Using Shift + Navigation key to extend selection will keep working as normal and the pane will scroll if necessary, but not because the scroll pane is handling that key (it would have no clue what to do on shift + right) but because the control involved will try to keep the cursor visible.

I said "think of how Shift+Right to extend the selection is implemented in most table views". I never claimed it was handled by the scroll pane or even that it should be. It is typically implemented by allowing an event/action/selector/whatever to bubble up to the enclosing table view to be acted on even if the focus lies on one of its descendants. I was just countering this (new) concept that a control shouldn't act on a keystroke unless it has direct focus. That's just not a generally accepted rule.

> Now, **custom** controls are a different matter. These may have relied on this behavior, and I think that's why this change should be mentioned explicitly in the release notes. They no longer get "scrolling" for free. However, I think even for these controls the default behavior is undesirable. We're again talking about focusable controls, and letting scroll pane have its way will mean those controls can scroll out of view. If it is a single view filling control (like a `*View`) it is likely these keys have been overridden by the custom control to provide better than default behavior.

If I understand you correctly you're once again citing a new rule, namely that PgUp and PgDn shouldn't allow the focused node to scroll out of view.

It is also not a generally accepted rule that if the default behavior isn't correct in all cases it should be removed.

I'm going to tap out of this conversation. I don't agree with these changes but don't think they will inflict much harm. I believe that if there's pushback from the developer community you will not get far using the novel principles you're citing (like the one outlined in the bug title). But perhaps I'm out of step on this and it's time for others to chime in.

hjohn:
> I said "think of how Shift+Right to extend the selection is implemented in most table views". I never claimed it was handled by the scroll pane or even that it should be. It is typically implemented by allowing an event/action/selector/whatever to bubble up to the enclosing table view to be acted on even if the focus lies on one of its descendants. 

I'm sorry, I misunderstood what you meant here.  This kind of bubbling up is not the problem.  Such events bubble up from the sub-structure of a control (say a table cell) to the table.  The table is still the thing that has focus, which also manages a cursor (the active cell, or the more common cursor in an edit box).  There is no confusion here for users what the focused control is and where input is going, and so it is absolutely fine to let this bubble up to a `Node` that is from a user perspective  "focused" and part of the same control.

This is not the case with `ScrollPane`. Here it clearly bubbles "out" of the focused area to a different control.  If there was a focused box around the entire control (including the scroll bars, like a TableView, TextArea or ListView have) then it is clear to the user where input is going.

> I was just countering this (new) concept that a control shouldn't act on a keystroke unless it has direct focus. That's just not a generally accepted rule.

Perhaps I worded this poorly.  What I always meant to say is a key press should go to the control that has visible focus.  If that control is composed of (to the user irrelevant) other components, then it is fine to let events bubble up within the control and its substructure, as long as the event is not acted upon "outside" the user visible focused area.

Acting on a key press bubbled up from a focused `Button` to a `ScrollPane` would clearly violate this.  The scroll pane does not have visible focus, the button does.

Acting on a key press bubbled up from a table cell doesn't violate this; the `TableView` has the focus (and shows this with a visual cue), the cell is just its cursor and is part of the table view's sub-structure.  Here you can see that the `TableView` has the focused border, and so is free to act on key presses that bubble up to that point:

image

> If I understand you correctly you're once again citing a new rule, namely that PgUp and PgDn shouldn't allow the focused node to scroll out of view.

I'll admit that this just seems more like common sense than a rule... how else are you supposed to provide a UI that the user can intuitively grasp?  You were talking about accessibility earlier, wouldn't an accessible UI also require that the control that has input focus is kept visible?

> It is also not a generally accepted rule that if the default behavior isn't correct in all cases it should be removed.

It would definitely be hard to make that a general rule.  It will have to be judged for each case if it does more harm than good.

> I'm going to tap out of this conversation. I don't agree with these changes but don't think they will inflict much harm. I believe that if there's pushback from the developer community you will not get far using the novel principles you're citing (like the one outlined in the bug title). But perhaps I'm out of step on this and it's time for others to chime in.

I understand, thanks for your feedback, it is appreciated and you brought several good points.  Apologies if I sometimes word things poorly or come across a bit too invested in issues; I really do believe this is worth changing the current behavior for.

kevinrushforth:
@hjohn I now think this does warrant a CSR, if for no other reason than to capture the outcome of this discussion (once it is finished), and any potentially visible behavioral change that a custom control writer might see. Further, it might be helpful to document this in the `ScrollPane` class docs.

/csr

hjohn:
@kevinrushforth I've added a paragraph in `ScrollPane`'s javadoc:

> ScrollPane by default only acts on key presses when it has the focus ({@link #isFocused()} returns {@code true}) and won't respond to unconsumed key events that bubble up from a focused child control. The key presses it acts on are platform specific, but by default should include keys for panning and paging the view.

I don't think we should document each key press there (that could be something to document for the specific behavior type if they ever become public).  To compare, `Button` mentions how it works when it is a default or cancel button, but doesn't mention what keys it responds to for general operation.

hjohn:
@kevinrushforth Should the CSR include more than the change in the ScrollPane documentation, like reasoning for the change, and perhaps a code example of how to get the old behavior back (for custom control authors that may have been relying on this behavior)?

modules/javafx.controls/src/main/java/javafx/scene/control/ScrollPane.java
```diff
@@ -73,8 +73,12 @@
  * node in a Group.
  * <p>
  * ScrollPane sets focusTraversable to false.
- * </p>
- *
+ * <p>
+ * ScrollPane by default only acts on key presses when it has the focus
+ * ({@link #isFocused()} returns {@code true}) and won't respond to unconsumed
+ * key events that bubble up from a focused child control. The key presses it
+ * acts on are platform specific, but by default should include keys for panning
```
andy-goryachev-oracle, line 80: which keys are platform specific?  I could not find any - the behavior registers no platform-specific bindings.

hjohn: Perhaps it should read "may be platform specific" -- I just don't want to pin FX down on a specific implementation, as FX is aligning with the platforms it runs on, which it doesn't control.  So if one of the platforms at some point starts using different keys, then I'd expect FX to align with that.

kevinrushforth:
> @kevinrushforth Should the CSR include more than the change in the ScrollPane documentation, like reasoning for the change, and perhaps a code example of how to get the old behavior back (for custom control authors that may have been relying on this behavior)?

That might be a good idea. A short version of this would be useful as part of the release note.

modules/javafx.controls/src/main/java/javafx/scene/control/ScrollPane.java
andy-goryachev-oracle: right.  it just feels to me like the sentence "The key presses..." is out of place as it does not specify something that SP has to do.  remove?

hjohn:
For users to get back to old behavior, some calculations need to be done as `ScrollPane` doesn't offer any helpful programmatic methods to scroll the pane (which is perhaps something that could be addressed in a separate issue, I filed https://bugs.openjdk.org/browse/JDK-8342654).

The most useful helper is one that scroll by a fraction, shown below:

    static void scrollByFraction(ScrollPane scrollPane, double x, double y) {
        Node content = scrollPane.getContent();

        if (content == null) {
            return;
        }

        Bounds viewportBounds = scrollPane.getViewportBounds();
        Bounds layoutBounds = content.getLayoutBounds();

        if (x != 0) {
            double visibleFraction = viewportBounds.getWidth() / layoutBounds.getWidth();
            double range = scrollPane.getHmax() - scrollPane.getHmin();
            double scrollFactor = range * visibleFraction / (1 - visibleFraction);

            scrollPane.setHvalue(scrollPane.getHvalue() + x * scrollFactor);
        }

        if (y != 0) {
            double visibleFraction = viewportBounds.getHeight() / layoutBounds.getHeight();
            double range = scrollPane.getVmax() - scrollPane.getVmin();
            double scrollFactor = range * visibleFraction / (1 - visibleFraction);

            scrollPane.setVvalue(scrollPane.getVvalue() + y * scrollFactor);
        }
    }

Then to get something similar to the old behavior back is to install an event handler:

        scrollPane.addEventHandler(KeyEvent.KEY_PRESSED, e -> {
            double x = 0;
            double y = 0;

            switch(e.getCode()) {
                case KeyCode.LEFT -> x = -0.1;
                case KeyCode.RIGHT -> x = 0.1;
                case KeyCode.UP -> y = -0.1;
                case KeyCode.DOWN -> y = 0.1;
                case KeyCode.PAGE_UP -> y = -0.9;
                case KeyCode.PAGE_DOWN -> y = 0.9;
                case KeyCode.SPACE -> y = 0.9;
                case KeyCode.HOME -> x = y = Double.NEGATIVE_INFINITY;
                case KeyCode.END -> x = y = Double.POSITIVE_INFINITY;
                default -> {}
            }

            if (x != 0 || y != 0) {
                scrollByFraction(scrollPane, x, y);
                e.consume();
            }
        });

Note that this doesn't exactly replicate the old behavior, but is probably a much more useful implementation than what was provided as standard before.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/ScrollPaneBehavior.java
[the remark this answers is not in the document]
hjohn: I've updated this, thanks for making me aware of it.

modules/javafx.controls/src/main/java/javafx/scene/control/ScrollPane.java
hjohn: I've removed the entire sentence, and also the "by default" part.

modules/javafx.controls/src/main/java/javafx/scene/control/ScrollPane.java
```diff
@@ -73,8 +73,10 @@
  * node in a Group.
  * <p>
  * ScrollPane sets focusTraversable to false.
- * </p>
- *
+ * <p>
+ * ScrollPane only acts on key presses when it has the focus ({@link #isFocused()}
+ * returns {@code true}) and won't respond to unconsumed key events that bubble
```
andy-goryachev-oracle, line 78: what about consumed events?  does it respond to those?

i.e. you may want to remove the word 'unconsumed'

kevinrushforth (approved):
The latest version looks good. I tested it (using the test program attached to the bug) and it works as expected.

I have also Reviewed the CSR, and it is ready to Finalize.

kevinrushforth:
@hjohn In case you missed the notification, this is ready to integrate.

hjohn:
Yeah, I was planning to write the release note first, but I guess we don't need to wait for that.

kevinrushforth:
> Yeah, I was planning to write the release note first, but I guess we don't need to wait for that.

Right. Which reminds me that I now have a backlog of 4 release notes to write.

PR #1587, opened 2024-10-01
8341372: BackgroundPosition, BorderImage, BorderStroke, CornerRadii should be final

mstr2:
Backgrounds and borders are comprised of immutable and final types. The following types are marked with the `final` modifier:

* Background
* BackgroundFill
* BackgroundImage
* BackgroundSize
* Border
* BorderWidths

The following types are not marked with the `final` modifier:

* BackgroundPosition
* BorderImage
* BorderStroke
* CornerRadii

This is probably just an oversight, as there is no value in allowing a subsection of these types to be extended.

/reviewers 2
/csr

---------
### Progress
- [x] Change requires CSR request [JDK-8341434](https://bugs.openjdk.org/browse/JDK-8341434) to be approved

### Issues
 * [JDK-8341372](https://bugs.openjdk.org/browse/JDK-8341372): BackgroundPosition, BorderImage, BorderStroke, CornerRadii should be final (**Enhancement** - P4)
 * [JDK-8341434](https://bugs.openjdk.org/browse/JDK-8341434): BackgroundPosition, BorderImage, BorderStroke, CornerRadii should be final (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1587/head:pull/1587` \
`$ git checkout pull/1587`

`$ git checkout pull/1587` \
`$ git pull https://git.openjdk.org/jfx.git pull/1587/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1587`

`$ git pr show -t 1587`

</details>

<a href="https://git.openjdk.org/jfx/pull/1587.diff">https://git.openjdk.org/jfx/pull/1587.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1587#issuecomment-2387139929)

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/layout/Background.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/layout/Background.java
@@ -144,7 +144,7 @@ public final class Background implements Interpolatable<Background> {
      * @interpolationType <a href="../../animation/Interpolatable.html#pairwise">pairwise</a>
      */
     public final List<BackgroundFill> getFills() { return fills; }
-    final List<BackgroundFill> fills;
+    private final List<BackgroundFill> fills;
 
     /**
      * The list of BackgroundImages which together define the image portion
@@ -155,7 +155,7 @@ public final class Background implements Interpolatable<Background> {
      * @interpolationType <a href="../../animation/Interpolatable.html#pairwise">pairwise</a>
      */
     public final List<BackgroundImage> getImages() { return images; }
-    final List<BackgroundImage> images;
+    private final List<BackgroundImage> images;
 
     /**
      * The outsets of this Background. This represents the largest
@@ -169,7 +169,7 @@ public final class Background implements Interpolatable<Background> {
      * @interpolationType the intermediate value is derived from {@link #getFills() fills}
      */
     public final Insets getOutsets() { return outsets; }
-    final Insets outsets;
+    private final Insets outsets;
 
     /**
      * Gets whether the background is empty. It is empty if there are no fills or images.
@@ -185,18 +185,17 @@ public final class Background implements Interpolatable<Background> {
     private final boolean hasOpaqueFill;
 
     /**
-     * Package-private immutable fields referring to the opaque insets
-     * of this Background.
+     * The opaque insets of this Background.
      */
     private final double opaqueFillTop, opaqueFillRight, opaqueFillBottom, opaqueFillLeft;
-    final boolean hasPercentageBasedOpaqueFills;
+    private final boolean hasPercentageBasedOpaqueFills;
 
     /**
      * True if there are any fills that are in some way based on the size of the region.
      * For example, if a CornerRadii on the fill is percentage based in either or both
      * dimensions.
      */
-    final boolean hasPercentageBasedFills;
+    private final boolean hasPercentageBasedFills;
 
     /**
      * The cached hash code computation for the Background. One very big
@@ -314,7 +313,7 @@ public final class Background implements Interpolatable<Background> {
             // The common case is to NOT have percent based radii
             final boolean b = fill.getRadii().hasPercentBasedRadii;
             hasPercentFillRadii |= b;
-            if (fill.fill.isOpaque()) {
+            if (fill.getFill().isOpaque()) {
                 opaqueFill = true;
                 if (b) {
                     hasPercentOpaqueInsets = true;
@@ -452,7 +451,7 @@ public final class Background implements Interpolatable<Background> {
                     final double fillBottom = fillInsets.getBottom();
                     final double fillLeft = fillInsets.getLeft();
 
-                    if (fill.fill.isOpaque()) {
+                    if (fill.getFill().isOpaque()) {
                         // Some possible configurations:
                         //     (a) rect1 is completely contained by rect2
                         //     (b) rect2 is completely contained by rect1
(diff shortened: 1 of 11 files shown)


andy-goryachev-oracle:
There might be a very narrow use case (like debugging) when developer might need to extend to supply a custom `toString()` or some other method.

From a high level design standpoint, I do agree with making these classes final.

mstr2:
> There might be a very narrow use case (like debugging) when developer might need to extend to supply a custom `toString()` or some other method.
> 
> From a high level design standpoint, I do agree with making these classes final.

Maybe, but that's not a sufficient reason to allow subclassing, as the same could be said for all cases of encapsuluation and inheritance control. Alternatives are:
1) Use the debugger to mark objects
2) Use a tool to change method definitions at runtime
3) Download OpenJFX and build it locally with the required changes

kevinrushforth:
> > There might be a very narrow use case (like debugging) when developer might need to extend to supply a custom `toString()` or some other method.
> > From a high level design standpoint, I do agree with making these classes final.
> 
> Maybe, but that's not a sufficient reason to allow subclassing, as the same could be said for all cases of encapsuluation and inheritance control.

I agree that this use case should not be a consideration.

As noted, this is an incompatible change, but we have a precedent for doing this for similar classes when we made the Color class final in JDK 8 with [JDK-8118119](https://bugs.openjdk.org/browse/JDK-8118119).

modules/javafx.graphics/src/main/java/javafx/scene/layout/BackgroundPosition.java
```diff
@@ -52,7 +52,7 @@
  * the Region's right edge.
  * @since JavaFX 8.0
  */
-public class BackgroundPosition implements Interpolatable<BackgroundPosition> {
+public final class BackgroundPosition implements Interpolatable<BackgroundPosition> {
```
andy-goryachev-oracle, line 55: should we also declare the fields `final` as well?
horizontalSide, etc.

modules/javafx.graphics/src/main/java/javafx/scene/layout/BorderImage.java
```diff
@@ -50,7 +50,7 @@
  * When applied to a Region with a defined shape, a BorderImage is ignored.
  * @since JavaFX 8.0
  */
-public class BorderImage implements Interpolatable<BorderImage> {
+public final class BorderImage implements Interpolatable<BorderImage> {
```
andy-goryachev-oracle, line 53: same question about final fields.

except for:
- innerEdge/outerEdge (used by border) and
- image use in Region:799 (could be accessed via getImage())

modules/javafx.graphics/src/main/java/javafx/scene/layout/BorderStroke.java
```diff
@@ -48,7 +48,7 @@
  *
  * @since JavaFX 8.0
  */
-public class BorderStroke implements Interpolatable<BorderStroke> {
+public final class BorderStroke implements Interpolatable<BorderStroke> {
```
andy-goryachev-oracle, line 51: same suggestion with final fields, except innerEdge/outerEdge

modules/javafx.graphics/src/main/java/javafx/scene/layout/BackgroundPosition.java
mstr2: All fields of `BackgroundPosition` are already final. Same for `BorderImage` and `BorderStroke`.

andy-goryachev-oracle: sorry, I meant `private`

mstr2: Yes, that's a sensible change.

modules/javafx.graphics/src/main/java/javafx/scene/layout/Background.java
```diff
@@ -314,7 +313,7 @@ private Background(List<BackgroundFill> fills, List<BackgroundImage> images, int
             // The common case is to NOT have percent based radii
             final boolean b = fill.getRadii().hasPercentBasedRadii;
             hasPercentFillRadii |= b;
-            if (fill.fill.isOpaque()) {
+            if (fill.getFill().isOpaque()) {
```
andy-goryachev-oracle, line 316: it was probably ok to leave some fields as package protected: I don't actually know if the compiler replaces the getFill() function call with direct access to the field.

on the other hand, a public API is always a better choice.

andy-goryachev-oracle (approved):

modules/javafx.graphics/src/main/java/javafx/scene/layout/Background.java
hjohn: Once the code is sufficiently often called, it will be optimized and subsequently inline the call. It then becomes equivalent to a direct field access.

hjohn:
> > There might be a very narrow use case (like debugging) when developer might need to extend to supply a custom `toString()` or some other method.
> > From a high level design standpoint, I do agree with making these classes final.
> 
> Maybe, but that's not a sufficient reason to allow subclassing, as the same could be said for all cases of encapsuluation and inheritance control. Alternatives are:
> 
> 1. Use the debugger to mark objects
> 2. Use a tool to change method definitions at runtime
> 3. Download OpenJFX and build it locally with the required changes

4. Copy/paste class without changing its package into your code and modify it (only works when not running with modules) -- it's so much faster to test changes this way than building JFX.

hjohn (approved):
Changes look good; those classes should definitely be final.

nlisker:
I've previously looked at these classes when records came out and they would be good candidates for being records, but it breaks binary compatibility. They would also be good value classes when that comes out. In any case, they need to be final and immutable.

kevinrushforth (approved):
Looks good. I'll review the CSR shortly.

PR #1590, opened 2024-10-03
8333374: Cannot invoke "com.sun.prism.RTTexture.contentsUseful()" because "this.txt" is null 

lukostyra:
Contrary to issue description, the problem was not because we referenced a texture after it was freed, but we referenced an RTT that we just tried to allocate and failed to do so because of lack of space in Prism's vram pool. In case of attached `WebViewAnimationTest.java` test app, the problem exists because WebView tries to allocate a new RTT for newly opened WebView scene and the Prism pool does not have 16MB needed while previous WebView is still being displayed. Only after the old WebView is disposed of, then the RTT allocation can proceed and WebView can continue as normal.

I looked around and did not find other solution than adding the null-checks to silence the NPE being thrown. NPE was originally in `RTImage.java`, however after adding a null check there triggered another NPE in `WCGraphicsPrismContext.java`, so I also added another null check there. Upper layers of web module seem to handle those just fine, the NPEs disappeared after that and the test still works properly once the pool gets enough room to display.

All of our tests run fine with this change (I do not expect any major regressions, as this happens only with a very limited memory pool scenario - it was extremely hard and time consuming to trigger this bug on the test app with default 512MB pool limit). For the test app, it might take a couple frames until the old WebView is disposed of and there is enough room for new WebView's RTT in the pool, after that the scene renders properly.

---------
### Progress

### Issue
 * [JDK-8333374](https://bugs.openjdk.org/browse/JDK-8333374): Cannot invoke "com.sun.prism.RTTexture.contentsUseful()" because "this.txt" is null (**Bug** - P4)

### Reviewers
 * [Hima Bindu Meda](https://openjdk.org/census#hmeda) (@HimaBinduMeda - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1590/head:pull/1590` \
`$ git checkout pull/1590`

`$ git checkout pull/1590` \
`$ git pull https://git.openjdk.org/jfx.git pull/1590/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1590`

`$ git pr show -t 1590`

</details>

<a href="https://git.openjdk.org/jfx/pull/1590.diff">https://git.openjdk.org/jfx/pull/1590.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1590#issuecomment-2390967879)

### DIFF
--- a/modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/RTImage.java
+++ b/modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/RTImage.java
@@ -115,6 +115,10 @@ final class RTImage extends PrismImage implements ResourceFactoryListener {
                     (int) Math.ceil(width * pixelScale),
                     (int) Math.ceil(height * pixelScale),
                     Texture.WrapMode.CLAMP_NOT_NEEDED);
+            if (txt == null) {
+                log.fine("RTImage::getTexture : return null because rt texture not allocated");
+                return null;
+            }
             txt.contentsUseful();
             txt.makePermanent();
             if (registeredWithFactory == null || registeredWithFactory.get() != f) {
--- a/modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/WCGraphicsPrismContext.java
+++ b/modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/WCGraphicsPrismContext.java
@@ -146,6 +146,11 @@ class WCGraphicsPrismContext extends WCGraphicsContext {
                     ? l.getGraphics()
                     : baseGraphics;
 
+            if (cachedGraphics == null) {
+                log.fine("getGraphics failed - couldn't acquire cachedGraphics");
+                return null;
+            }
+
             ResourceFactory rf = cachedGraphics.getResourceFactory();
             if (!rf.isDisposed()) {
                 state.apply(cachedGraphics);


kevinrushforth:
Reviewers: @kevinrushforth @HimaBinduMeda 

/reviewers 2

kevinrushforth:
At first glance this  PR looks good, but I'll look more closely and do some testing.

Is it feasible to add a test for this? (automated if possible, manual if not)

lukostyra:
As for the manual test that was added to the issue, I am not sure if we can easily add it to the repo, especially considering that it uses some HTML/JS code to generate the "memory intensive" scenarios.

I might try triggering the problem from a unit test, but I'm not sure if I'll succeed. I'll come back to this in a short bit.

modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/RTImage.java
```diff
@@ -115,6 +115,10 @@ private RTTexture getTexture() {
                     (int) Math.ceil(width * pixelScale),
                     (int) Math.ceil(height * pixelScale),
                     Texture.WrapMode.CLAMP_NOT_NEEDED);
+            if (txt == null) {
+                log.fine("RTImage::getTexture : return null because rt texture not allocated");
+                return null;
```
mstr2, line 120: If `RTImage.getTexture()` can return `null`, then L147 should probably also check for `null` before calling `readPixels()`.

lukostyra:
I played around with testing this issue a bit and I couldn't find a good automatic solution. The main problem here is that the exception we're looking for is caught by the RenderJob thread, so we can't really easily capture it via system tests. On the other hand, since it depends on Prism's vram pool, it is quite difficult to make it happen via unit tests. I also had issues creating a stressful enough scenario for the Web engine that would trigger the exception, and the file attached to the issue - `ngrams.html` - uses an external JavaScript library and a lot of additional code that I am not sure of when it comes to its source.

In the process I also found another exception, this one thrown inside Prism:
```
    java.lang.NullPointerException: Cannot invoke "com.sun.prism.d3d.D3DRTTexture.contentsUseful()" because "rtt" is null
        at javafx.graphics@24-internal/com.sun.prism.d3d.D3DResourceFactory.createPresentable(D3DResourceFactory.java:383)
        at javafx.graphics@24-internal/com.sun.javafx.tk.quantum.PresentingPainter.run(PresentingPainter.java:81)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
        at java.base/java.util.concurrent.FutureTask.runAndReset(FutureTask.java:358)
        at javafx.graphics@24-internal/com.sun.javafx.tk.RenderJob.run(RenderJob.java:58)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
        at javafx.graphics@24-internal/com.sun.javafx.tk.quantum.QuantumRenderer$PipelineRunnable.run(QuantumRenderer.java:126)
        at java.base/java.lang.Thread.run(Thread.java:1570)
```
Another side effect of not properly null-checking the data we allocate, this time in PresentingPainter. However, since it is a separate module, I think I would solve this separately (and, again, it is very very hard to hit it in "normal" conditions - it only appeared when vram pool was limited to 40M).

kevinrushforth (approved):
The fix looks good to me. I also tested it, and it works as expected. I added one reply to an earlier question.

kevinrushforth:
@jperedadnr Since you reported the bug, do you want to check that this does what you expect?

lukostyra:
> We should file a follow-up bug to check the thread-safety of the draw method in RTImage in the case of printing.

Filed [JDK-8341852](https://bugs.openjdk.org/browse/JDK-8341852) to track the threading issue

jperedadnr (approved):
Tested on macOS, it works fine after applying this fix.

HimaBinduMeda (approved):
Fix looks good. No NPE is seen after applying the fix, when testing the attached WebViewAnimationTest on Mac

PR #1592, opened 2024-10-05
8341514: Add reducedMotion and reducedTransparency preferences

mstr2:
This PR adds the `Platform.Preferences.reducedMotion` and `Platform.Preferences.reducedTransparency` accessibility preferences:
```java
interface Preferences {
    /**
     * Specifies whether applications should minimize the amount of non-essential animations,
     * reducing discomfort for users who experience motion sickness or vertigo.
     * <p>
     * If the platform does not report this preference, this property defaults to {@code false}.
     *
     * @return the {@code reducedMotion} property
     * @defaultValue {@code false}
     * @since 24
     */
    ReadOnlyBooleanProperty reducedMotionProperty();

    /**
     * Specifies whether applications should minimize the amount of transparent or translucent
     * layer effects, which can help to increase contrast and readability for some users.
     * <p>
     * If the platform does not report this preference, this property defaults to {@code false}.
     *
     * @return the {@code reducedTransparency} property
     * @defaultValue {@code false}
     * @since 24
     */
    ReadOnlyBooleanProperty reducedTransparencyProperty();

    ...
}
```

On Windows, these preferences correspond to "Transparency effects" and "Animation effects" in the system settings:

On macOS, these preferences correspond to "Reduce motion" and "Reduce transparency" in the system settings:

On Linux, the `reducedMotion` preference corresponds to `gtk-enable-animations`. On my Ubuntu 24.04 system, GTK3 does not pick up the "Settings / Accessibility / Reduce Animations" system setting, and therefore GTK always reports `gtk-enable-animations == true`. Changing the preference requires manually editing `etc/gtk-3.0/settings.ini`. The `reducedTransparency` preference has no corresponding GTK setting.

/reviewers 2
/csr

---------
### Progress
- [x] Change requires CSR request [JDK-8341667](https://bugs.openjdk.org/browse/JDK-8341667) to be approved

### Issues
 * [JDK-8341514](https://bugs.openjdk.org/browse/JDK-8341514): Add reducedMotion and reducedTransparency preferences (**Enhancement** - P4)
 * [JDK-8341667](https://bugs.openjdk.org/browse/JDK-8341667): Add reducedMotion and reducedTransparency preferences (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1592/head:pull/1592` \
`$ git checkout pull/1592`

`$ git checkout pull/1592` \
`$ git pull https://git.openjdk.org/jfx.git pull/1592/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1592`

`$ git pr show -t 1592`

</details>

<a href="https://git.openjdk.org/jfx/pull/1592.diff">https://git.openjdk.org/jfx/pull/1592.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1592#issuecomment-2394910311)

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2023, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -30,6 +30,8 @@ import com.sun.javafx.util.Utils;
 import javafx.application.ColorScheme;
 import javafx.beans.InvalidationListener;
 import javafx.beans.property.Property;
+import javafx.beans.property.ReadOnlyBooleanProperty;
+import javafx.beans.property.ReadOnlyBooleanWrapper;
 import javafx.beans.property.ReadOnlyObjectProperty;
 import javafx.beans.property.ReadOnlyObjectPropertyBase;
 import javafx.beans.property.ReadOnlyObjectWrapper;
@@ -43,15 +45,50 @@ import java.util.Objects;
  */
 final class PreferenceProperties {
 
-    private final ColorProperty backgroundColor = new ColorProperty("backgroundColor", Color.WHITE);
-    private final ColorProperty foregroundColor = new ColorProperty("foregroundColor", Color.BLACK);
-    private final ColorProperty accentColor = new ColorProperty("accentColor", Color.rgb(21, 126, 251));
-    private final List<ColorProperty> allColors = List.of(backgroundColor, foregroundColor, accentColor);
+    private final DeferredProperty<Color> backgroundColor = new DeferredProperty<>("backgroundColor", Color.WHITE);
+    private final DeferredProperty<Color> foregroundColor = new DeferredProperty<>("foregroundColor", Color.BLACK);
+    private final DeferredProperty<Color> accentColor = new DeferredProperty<>("accentColor", Color.rgb(21, 126, 251));
+    private final List<DeferredProperty<Color>> allColors = List.of(backgroundColor, foregroundColor, accentColor);
     private final ColorSchemeProperty colorScheme = new ColorSchemeProperty();
+    private final DeferredProperty<Boolean> reducedMotion = new DeferredProperty<>("reducedMotion", false);
+    private final DeferredProperty<Boolean> reducedTransparency = new DeferredProperty<>("reducedTransparency", false);
+    private final List<DeferredProperty<Boolean>> allFlags = List.of(reducedMotion, reducedTransparency);
+    private final ReadOnlyBooleanWrapper reducedMotionFlag;
+    private final ReadOnlyBooleanWrapper reducedTransparencyFlag;
     private final Object bean;
 
     PreferenceProperties(Object bean) {
         this.bean = bean;
+
+        reducedMotionFlag = new ReadOnlyBooleanWrapper(bean, reducedMotion.getName());
+        reducedMotionFlag.bind(reducedMotion);
+
+        reducedTransparencyFlag = new ReadOnlyBooleanWrapper(bean, reducedTransparency.getName());
+        reducedTransparencyFlag.bind(reducedTransparency);
+    }
+
+    public ReadOnlyBooleanProperty reducedMotionProperty() {
+        return reducedMotionFlag.getReadOnlyProperty();
+    }
+
+    public boolean isReducedMotion() {
+        return reducedMotion.get();
+    }
+
+    public void setReducedMotion(boolean value) {
+        reducedMotion.setValueOverride(value);
+    }
+
+    public ReadOnlyBooleanProperty reducedTransparencyProperty() {
+        return reducedTransparencyFlag.getReadOnlyProperty();
+    }
+
+    public boolean isReducedTransparency() {
+        return reducedTransparency.get();
+    }
+
+    public void setReducedTransparency(boolean value) {
+        reducedTransparency.setValueOverride(value);
     }
 
     public ReadOnlyObjectProperty<ColorScheme> colorSchemeProperty() {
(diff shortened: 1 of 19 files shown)


modules/javafx.graphics/src/main/java/javafx/application/Platform.java
```diff
@@ -580,6 +585,34 @@ public static Preferences getPreferences() {
...
+         * Specifies whether applications should minimize the amount of non-essential animations,
+         * reducing discomfort for users who experience motion sickness or vertigo.
+         * <p>
+         * If the platform does not report this preference, this property defaults to {@code false}.
+         *
+         * @return the {@code reducedMotion} property
+         * @defaultValue {@code false}
+         * @since 24
+         */
+        ReadOnlyBooleanProperty reducedMotionProperty();
```
kevinrushforth, line 598: Adding a new method  without a default implementation to an interface is an incompatible change. Since we do not expect that an application would ever implement this interface, this seems OK to me.

We should either add an `@implSpec` to the class docs indicating that applications should not implement  `Platform.Preferences` or make it a sealed interface.

kevinrushforth (commented):
This seems like a reasonable enhancement. Since this adds new interface methods, it is (technically) an incompatible change. I recommend either documenting that applications should not implement the interface, or else make it a sealed class so that they can't.

modules/javafx.graphics/src/main/java/javafx/application/Platform.java
mstr2: Good point, I've sealed the interface.

kevinrushforth (commented):
The API changes look good. I'll review the impl changes later; create the CSR whenever you are ready.

andy-goryachev-oracle:
I've added the Platform Preferences Monitor to the Monkey Tester (Tools -> ...)

Screenshot 2024-10-08 at 14 55 41


One thing I could not do is to enumerate the properties getter via reflection or BeanInfo/PropertyDescriptor because the class is not accessible `com.sun.javafx.application.preferences.PlatformPreferences`

This presents a problem for any application that is expected to work with different JFX versions.

modules/javafx.graphics/src/main/java/javafx/application/Platform.java
```diff
@@ -572,13 +576,43 @@ public static Preferences getPreferences() {
...
      *         <tr><td>{@code GTK.success_color}</td><td>{@link Color}</td></tr>
+     *         <tr><td>{@code GTK.enable_animations}</td><td>{@link Boolean}</td></tr>
      *         <tr></tr>
      *     </tbody>
      * </table>
      *
      * @since 22
      */
-    public interface Preferences extends ObservableMap<String, Object> {
+    public sealed interface Preferences extends ObservableMap<String, Object>
```
andy-goryachev-oracle, line 586: is there any way we can convert this to a regular class here, so it can be inspected?

what's the point of declaring an interface just to allow for one possible implementation?

kevinrushforth:
> One thing I could not do is to enumerate the properties getter via reflection or BeanInfo/PropertyDescriptor because the class is not accessible com.sun.javafx.application.preferences.PlatformPreferences

Did try to use `pref.getClass()` for your reflection? That won't work. You should be able to look up the methods of `Platform.Preferences.class`.

modules/javafx.graphics/src/main/java/javafx/application/Platform.java
kevinrushforth: No, that would be a mistake.

kevinrushforth:
To add a little more detail on what is a common mistake when doing reflection on a returned type from any API:

Fails:

```
    var pref = Platform.getPreferences();
    var methods = pref.getClass().getMethods();
```

Works:

```
    var methods = Platform.Preferences.class.getMethods();
```

modules/javafx.graphics/src/main/java/javafx/application/Platform.java
andy-goryachev-oracle: please ignore.  Using reflection and/or BeanInfo on Platform.Properties.class worked.

andy-goryachev-oracle:
@mstr2 I see the map values change, but both new properties showing wrong initial values and not being updated.

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
```diff
@@ -102,14 +139,23 @@ public void setAccentColor(Color color) {
...
+        outerLoop:
         for (Map.Entry<String, ChangedValue> entry : changedPreferences.entrySet()) {
-            String key = platformKeyMappings.get(entry.getKey());
-            if (key != null) {
-                for (ColorProperty colorProperty : allColors) {
-                    if (colorProperty.getName().equals(key)) {
-                        updateColorProperty(colorProperty, entry.getValue().newValue());
-                        break;
+            PreferenceMapping<?> mapping = platformKeyMappings.get(entry.getKey());
+            if (mapping != null) {
```
andy-goryachev-oracle, line 147: this code seems rather inefficient:

why are we doing linear search?
also why are we searching through colors when, for example, the changed property is a boolean?

is it possible to do a simple hash map lookup instead?

andy-goryachev-oracle:
just to be clear, the code I used to verify properties in the monkey tester is both via UI and `addListener`:

https://github.com/andy-goryachev-oracle/MonkeyTest/blob/b78a34ddf028c8e987249ed8f4ace6fca6469efc/src/com/oracle/tools/fx/monkey/sheets/PropertiesMonitor.java#L318

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
mstr2: Good idea, this has also removed quite a bit of code.

mstr2:
> @mstr2 I see the map values change, but both new properties showing wrong initial values and not being updated (on macOS):

This should also be fixed now.

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
```diff
@@ -226,11 +192,13 @@ public T get() {
...
 
         /**
-         * Only called by {@link #updateDeferredProperty}, this method doesn't fire a change notification.
+         * Only called from {@link PreferenceProperties#update}, this method doesn't fire a change notification.
          * Change notifications are fired after the new values of all deferred properties have been set.
          */
-        public void setValue(T value) {
-            this.platformValue = value;
+        @SuppressWarnings("unchecked")
+        public void setPlatformValue(Object value) {
```
andy-goryachev-oracle, line 199: should this method be declared package protected?

mstr2: It doesn't really matter, since it's a nested class. The outer class has access to all members, regardless of visibility. Sometimes I use access modifiers in nested classes to informally document intent (which methods are supposed to be called by the outer class, and which are not). For consistency, I've added the `public` modifier to `fireValueChangeIfNecessary`.

andy-goryachev-oracle: you are probably right: it does not matter.

at first I thought that since the instances of `DeferredProperty` are public, one can use reflection to get to its public methods, but as we learned earlier, reflection will likely fail due to this class being unaccessible.

andy-goryachev-oracle (approved):
looks good, tested on macOS 14.6.1 M1 and win 11.
did not test it on linux

kevinrushforth:
API looks good. I reviewed the CSR. The impl also looks good. I just need to finish my testing.

kevinrushforth (approved):
Tested on Linux. All good.

PR #1595, opened 2024-10-07
8341090: Remove support for security manager from JavaFX

kevinrushforth:
This PR removes support for running JavaFX applications with the Java Security Manager.

The initial work was done in 4 separate commits as follows:

* Fail fast at startup if the Security Manager is enabled
* Remove `-Djava.security.manager` and all security policy files; delete `SandboxAppTest`
* Delete remaining tests that depend on the Security Manager; as part of this, I renamed`CustomSecurityManagerTest` to `StageRobotTest` and removed all of the tests that depend on the security manager, since the test methods that don't depend on SM are useful functional tests of a Stage's FullScreen and AlwaysOnTop mode.
* API spec change to remove mention of security manager (including mention of throwing SecurityException)

This PR is both necessary and sufficient to remove support for running with the Java Security Manager enabled. I have tested this with JDK 23, which still has SM support (although deprecated for removal), and with a local JDK built from the [jdk-sandbox:jep486](https://github.com/openjdk/jdk-sandbox/tree/jep486) branch, which has the SM disabled. With both JDKs, I can build and run this PR branch; all tests pass. Attempting to enable the security manager with JDK 23 will now fail in `<clinit>` of either `LauncherImpl` (called by `Application.launch`) or `PlatformImpl` (called by `Platform.startup`) with an informative error message.

This PR should be integrated prior to the integration of JEP 486, so that we can continue to run JavaFX tests on JDK 24 after the SM is disabled. Otherwise they will fail to run with a fatal error launching the JVM (we currently pass `-Djava.security.manager=allow` when running our tests).

#### Follow-on work

I will file additional JBS issues to track follow-up work to cleanup the remaining calls to the security manager methods that are deprecated for removal. The follow-up work can be done independently of the timing of the integration of JEP 486 into JDK 24, but all of them will be targeted for JavaFX 24.

Follow-on issues for JavaFX 24:

* Remove all calls to `doPrivileged` (there are 350 such calls in 168 files)
* Remove all other calls to `AccessController` (48 calls in 33 files)
* Remove all remaining calls to `System::getSecurityManager` (45 calls in 27 files)
* Remove calls to deprecated SM methods from `PlatformUtil` and `MethodUtil` in `javafx.base` (these are lightly modified copies of those classes in `java.base` so I will check how they are being handled)
* Remove all uses (either throw or catch) of the deprecated `AccessControlException` : : 4 occurrences in 4 files
* Remove all occurrences where we throw a `SecurityException`; check all place where we catch it (and consider whether it is needed)
* Check and remove any remaining usages of other SM methods that are deprecated for removal by JEP 411
    * SecurityManager 2 files (in addition to the ones that call `getSecurityManager`)
    * AccessControlContext: 169 occurrences in 65 files (many are associated with doPriviliged calls, so should be rechecked once doPriv is removed)
* Utility method for fail fast SM check (if we keep the check, else remove it)

NOTE: I had earlier thought that the "fail fast" could be temporary until all follow-up work is done, but I now think we should leave it in for JavaFX 24, possibly adding it to two other classes -- `ReflectUtil` and `MethodUtil` -- depending on the follow-up work needed for those classes. If we do end up leaving it, I'll create a utility method in `javafx.base` that will use reflection to check whether the security manager is enabled, so we don't leave in a call to a method (`System::getSecurityManager`) that is deprecated for removal.

We might also file additional follow-up issues, not necessarily targeted for JavaFX 24:

* Consider deprecating for removal the following security-related APIs in JavaFX
    * `FXPermission` (this PR changes the spec to make it clear that it is no longer used to control access to resources)
    * `WebErrorEvent.USER_DATA_DIRECTORY_SECURITY_ERROR` (if there is no other use than to report a SecurityException that can no longer happen after the security manager is removed)

---------
### Progress
- [x] Change requires CSR request [JDK-8341858](https://bugs.openjdk.org/browse/JDK-8341858) to be approved

### Issues
 * [JDK-8341090](https://bugs.openjdk.org/browse/JDK-8341090): Remove support for security manager from JavaFX (**Enhancement** - P2)
 * [JDK-8341858](https://bugs.openjdk.org/browse/JDK-8341858): Remove support for security manager from JavaFX (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1595/head:pull/1595` \
`$ git checkout pull/1595`

`$ git checkout pull/1595` \
`$ git pull https://git.openjdk.org/jfx.git pull/1595/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1595`

`$ git pr show -t 1595`

</details>

<a href="https://git.openjdk.org/jfx/pull/1595.diff">https://git.openjdk.org/jfx/pull/1595.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1595#issuecomment-2402570653)


tests/system/src/test/java/test/robot/helloworld/StageRobotTest.java
```diff
@@ -0,0 +1,182 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
+import com.sun.javafx.PlatformUtil;
+import test.robot.testharness.VisualTestBase;
+
+/**
+ * Basic visual tests using glass Robot to sample pixels.
+ */
+@Timeout(value=15000, unit=TimeUnit.MILLISECONDS)
+public class StageRobotTest extends VisualTestBase {
```
andy-goryachev-oracle, line 53: this file seem unrelated.  merge issue?

kevinrushforth: It's related. See this comment in the PR Description:

> ...as part of this, I renamed`CustomSecurityManagerTest` to `StageRobotTest` and removed all of the tests that depend on the security manager, since the test methods that don't depend on SM are useful functional tests of a Stage's FullScreen and AlwaysOnTop mode.

Because I ended up deleting so much of the file, git's heuristic for detecting renames doesn't think that it is.

If you run this diff manually on the PR branch, you'll see it:

```
git diff --find-renames=40 master... -- tests/system/src/test/java/test/robot/helloworld
```

andy-goryachev-oracle: sorry, missed that.  please ignore the comment.

arapte (approved):
A full build/test run looks good. Javadoc has no mention of security manager anymore.

johanvos:
@kevinrushforth looks good at first sight. I'm very happy to see the SandboxAppTest being removed!
When you say "all tests pass", what set of tests are you referring to? We have a number of properties that determine which tests are actually executed, and I believe in general it would be good if we could standardize on what set of tests we consider crucial (e.g. not the ones marked as unstable).
I'd like to run all tests as well on all supported platforms, but your explanation is very clear and makes sense, hence I don't expect issues.

kevinrushforth:
> When you say "all tests pass", what set of tests are you referring to? We have a number of properties that determine which tests are actually executed, and I believe in general it would be good if we could standardize on what set of tests we consider crucial (e.g. not the ones marked as unstable).

Agreed. By "all tests", I meant a full headful test run except unstable tests, meaning that I ran `gradle test` with: `-PFULL_TEST=true -PUSE_ROBOT=true`.

Our typical headful test run is:

```
gradle --info sdk shims
gradle --info --continue -PTEST_ONLY=true -PFULL_TEST=true -PUSE_ROBOT=true test
```

I usually run it like that so that I get a fail fast on the compilation (i.e., without the `--continue`) and so that repeat runs, or runs of specific tests don't need to recompile the sdk and shims, but if all you want is a single full test run, you don't need to split (in which case, omit the `TEST_ONLY` flag).

> I'd like to run all tests as well on all supported platforms, but your explanation is very clear and makes sense, hence I don't expect issues.

That would be great. Let me know what you find.

I did a full test run using JDK 23 (our default boot JDK) on the following platforms:

Windows 11
Mac aarch64 (both macOS 13 and macOS 14)
Mac x64 (both macOS 13 and macOS 14)
Ubuntu 22.04

I also did the same full test run setting `JDK_HOME` to a test build of the [jdk-sandbox:jep486](https://github.com/openjdk/jdk-sandbox/tree/jep486) branch on the same platforms, excluding Mac x64 (I didn't have a JDK jep486 build handy on our test machine, and since there is nothing arch-specific about the SM removal, that seemed OK).

kevinrushforth:
@johanvos Did you want a chance to review this before I integrate it?

johanvos (approved):

PR #1596, opened 2024-10-08
8341670: [Text,TextFlow] Public API for Text Layout Info

andy-goryachev-oracle:
Please refer to

https://github.com/andy-goryachev-oracle/Test/blob/main/doc/Text/LayoutInfo.md

The RichTextArea control ([JDK-8301121](https://bugs.openjdk.org/browse/JDK-8301121)), or any custom control that needs non-trivial navigation within complex or wrapped text needs a public API to get information about text layout.

This change fixes the missing functionality by adding a new public method to the `Text` and `TextFlow` classes.\:

```
    /**
     * Obtains the snapshot of the current text layout information.
     * @return the layout information
     * @since 25
     */
    public final LayoutInfo getLayoutInfo()
```

The `LayoutInfo` provides a view into the text layout within `Text`/`TextFlow` nodes such as:

- caret information
- text lines: offsets and bounds
- overall layout bounds
- text selection geometry
- strike-through geometry
- underline geometry

## WARNINGS

Presently, information obtained via certain Text/TextField methods is incorrect with non-zero padding and borders, see [JDK-8341438](https://bugs.openjdk.org/browse/JDK-8341438).

This PR provides correct answers in the new API, leaving the behavior of the existing methods unchanged (there is a compatibility risk associated with trying to fix [JDK-8341438](https://bugs.openjdk.org/browse/JDK-8341438) ).

Also, the RTL support is out of scope for this PR, due to multiple pre-existing conditions, see https://bugs.openjdk.org/browse/JDK-8343557

## Testing

The new APIs can be visually tested using the Monkey Tester on this branch:
https://github.com/andy-goryachev-oracle/MonkeyTest/tree/text.layout.api

in the Text and TextFlow pages:
Screenshot 2024-11-04 at 11 38 21

Two very basic headful tests have been added.

## See Also

https://github.com/FXMisc/RichTextFX/pull/1246

/reviewers 2
/csr

---------
### Progress
- [x] Change requires CSR request [JDK-8344648](https://bugs.openjdk.org/browse/JDK-8344648) to be approved

### Issues
 * [JDK-8341670](https://bugs.openjdk.org/browse/JDK-8341670): [Text,TextFlow] Public API for Text Layout Info (**Enhancement** - P4)
 * [JDK-8344648](https://bugs.openjdk.org/browse/JDK-8344648): [Text,TextFlow] Public API for Text Layout Info (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1596/head:pull/1596` \
`$ git checkout pull/1596`

`$ git checkout pull/1596` \
`$ git pull https://git.openjdk.org/jfx.git pull/1596/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1596`

`$ git pr show -t 1596`

</details>

<a href="https://git.openjdk.org/jfx/pull/1596.diff">https://git.openjdk.org/jfx/pull/1596.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1596#issuecomment-2400532550)
</details>


modules/javafx.graphics/src/main/java/javafx/scene/text/LayoutInfo.java
```diff
@@ -0,0 +1,95 @@
...
+package javafx.scene.text;
+
+import javafx.geometry.Rectangle2D;
+
+/**
+ * Represents an immutable snapshot of certain aspects of the text layout.
+ *
+ * @since 24
+ */
+public interface LayoutInfo {
```
prrace, line 34: Why an interface ? It is easier to evolve a class when you want to add more info.
Also it should be sealed unless you can show why it is important that
applications can create instances.

It is being exposed for the benefit of RT but the class doc itself doesn't say when you might want to get one and why.
It might also benefit from a succinct description of what "the text layout" means and encompasses.

andy-goryachev-oracle: good points, thanks!  please let me know if the update addressed your concerns.

a couple of notes:

- converted to an abstract class.  I feel weird referring to an internal implementation at the API level (`public sealed abstract class LayoutInfo permits com.sun.javafx.text.PrismLayoutInfo`, yikes!)
- why we get it: we don't explain why we need HitInfo or caret shapes or range shapes, do we?

modules/javafx.graphics/src/main/java/javafx/scene/text/LayoutInfo.java
```diff
@@ -0,0 +1,106 @@
...
+ * questions.
+ */
+package javafx.scene.text;
+
+import javafx.geometry.Rectangle2D;
+
+/**
+ * Represents an immutable snapshot of certain aspects of the text layout
+ * in a {@code Text} or {@code TextFlow} node,
+ * such as break up of the text into lines and the their bounds.
```
prrace, line 32: "the their"

modules/javafx.graphics/src/main/java/javafx/scene/text/LayoutInfo.java
```diff
@@ -0,0 +1,106 @@
...
+ * <p>
+ * The snapshot is valid until the layout changes due to any change that
+ * triggers that, such as resizing of the container or modification of properties.
```
prrace, line 35: "until the layout changes due to any change that triggers that"

it's not wrong, but writing change twice reads a bit off.
How about

The layout snapshot is no longer valid after actions such as resizing of the container, or modification of certain properties.
For example updating the text or the font would invalidate the layout snapshot, but a change of color would not.


I still don't see how applications can KNOW this has happened.
There's nothing on the (immutable) layout which says its invalidated.
There's no property to monitor.
So people will in effect never be able to do anything but use it and toss it and get a new one every time.
Perhaps you can make it 'cheap' to do this.
Since it is immutable, the Text or TextFlow can cache it.
If it is invalidated, the Text or TextFlow should know and can flush its cached version.
So the SAME immutable instance can be returned over and over again until that happens.

andy-goryachev-oracle: thank you!

I am struggling with how to explain that it should neither be cached, nor compared to another snapshot.

Maybe an example would help.  Use case is an editable rich text component based on `TextFlow`.  The user presses `HOME` and the caret needs to go to the beginning of the line.  We get the `LayoutInfo`, determine which line the caret is currently at and determine the start offset of that line to move the caret to.

Another example: in `Text`, to determine whether the mouse is clicked inside of the text or beyond the last character (see [JDK-8091012](https://bugs.openjdk.org/browse/JDK-8091012) ).

We are not expecting the application to monitor the text layout, so no notifications or observable properties or equals() are needed.

(I've added some words to that effect in `Text/TextFlow.getLayoutInfo()` )

prrace: Clients will usually obtain a LayoutInfo in order to respond to some user action or an update of the target node.
LayoutInfo should be considered an ephemeral snapshot, to be used and immediately discarded.
Clients should always obtain a new one when they need to respond again as likely the trigger to respond again means the target node has changed.

prrace: Although I don't see the harm in being more efficient.

andy-goryachev-oracle: I hear a suggestion to create a think wrapper/proxy instead of the snapshot, right?

This might, in fact, be a better idea.  The API stays the same.
Let me think.

Thanks!

andy-goryachev-oracle:
@jugen do you think this PR addresses the issue https://bugs.openjdk.org/browse/JDK-8091012 ?

andy-goryachev-oracle:
One additional thought/question:

There is a need to extract a few more data points from the text layout in addition to the text lines, specifically:
- bounds of the (selection) range, as rectangles
- strikethrough range shape (as `PathElement[]` and/or rectangles)
- underline range shape (as `PathElement[]` and/or rectangles)

We currently have `Text/TextFlow.underlineShape(int,int)`, `.rangeShape(int,int)` which return `PathElement[]` in an undocumented way.

There is no corresponding `.strikeThroughShape()` method, or a way to retrieve the geometric bounds as lines/rectangles.

Granted, one can attempt to analyze the `PathElement[]` and convert it to geometry, but it seems we should avoid asking the clients to do that.

So the question are:
- should we add these data points to LayoutInfo
- if so, in what form?  i.e. as individual methods or as an array of Line/Rectangle objects,
- should we add the `.strikeThroughShape()` to Text/TextFlow, or simply adding it here should be sufficient?

Jugen:
@andy-goryachev-oracle First off thanks for all your work on the richtext control you're working on.

> @Jugen do you think this PR addresses the issue https://bugs.openjdk.org/browse/JDK-8091012 ?

Unfortunately I don't think this helps for the hit problem Tomas was having, see `hit(x,y)` in RichTextFX's [TextFlowExt](https://github.com/FXMisc/RichTextFX/blob/9405b4f03a4a7865741a8216671f4bbbd3fe27bc/richtextfx/src/main/java/org/fxmisc/richtext/TextFlowExt.java#L142) for the relevant code.

However I was able to use **all** the methods in the `LayoutInfo` interface of this PR in RichTextFX to replace the custom layout info code, which is great.

Jugen:
Wrt to rangeShape & underlineShape: RichTextFX quite successfully uses the PathElement[] returned to it's advantage and so wouldn't gain from Rectangles.

`.strikeThroughShape()` should be added to Text/TextFlow for consistency sake.

andy-goryachev-oracle:
Thank you you all @prrace @Jugen @kevinrushforth for the feedback!
I am going to take this PR back to draft to make the suggested changes.

modules/javafx.graphics/src/main/java/javafx/scene/text/CaretInfo.java
```diff
@@ -0,0 +1,58 @@
...
+package javafx.scene.text;
+
+import com.sun.javafx.text.PrismCaretInfo;
+
+/**
+ * Provides the information associated with the caret.
+ *
+ * @since 24
+ */
+public sealed abstract class CaretInfo permits PrismCaretInfo {
```
kevinrushforth, line 34: Given the API you define, is there a reason not to make this an interface?

modules/javafx.graphics/src/main/java/javafx/scene/text/LayoutInfo.java
```diff
@@ -0,0 +1,121 @@
...
+ * with purpose of querying the details of the layout such as break up of the text into lines,
+ * as well as geometry of other shapes derived from the layout (selection, underline, etc.).
+ * <p>
+ * The information obtained via this object may change after the next layout cycle, which may come as a result
+ * of actions such as resizing of the container, or modification of certain properties.
+ * For example updating the text or the font might change the layout, but a change of color would not.
+ *
+ * @since 24
+ */
+public sealed abstract class LayoutInfo permits com.sun.javafx.text.PrismLayoutInfo {
```
kevinrushforth, line 41: Is there a reason to not make this an interface?

modules/javafx.graphics/src/main/java/javafx/scene/text/LayoutInfo.java
```diff
@@ -0,0 +1,121 @@
...
+     * @return the number of text lines
+     */
+    public abstract int getTextLineCount();
+
+    /**
+     * Returns the list of text lines in the layout.
+     *
+     * @return the list of {@code TextLineInfo} objects
+     */
+    public abstract List<TextLineInfo> getTextLines();
```
kevinrushforth, line 72: Is the list immutable? If so (which I think it should be), can you specify that?

modules/javafx.graphics/src/main/java/javafx/scene/text/LayoutInfo.java
```diff
@@ -0,0 +1,121 @@
...
+     * Returns the {@code TextLineInfo} object which contains information about
+     * the text line at index {@code index}.
+     *
+     * @param index the line index
+     * @return the {@code TextLineInfo} object
+     */
+    public abstract TextLineInfo getTextLine(int index);
+
+    /**
+     * Returns the geometry of the text selection, as an array of {@code Rectangle2D} objects,
```
kevinrushforth, line 84: "as an array ..." --> "as a list ...".

Is the list immutable?

kevinrushforth (commented):
I took a quick initial pass of the API. It looks like what I would expect. I left a couple questions inline. Also, you might want to document whether the newly added methods can return null (maybe in the LayoutInfo class docs, if none of them can return null).

modules/javafx.graphics/src/main/java/javafx/scene/text/CaretInfo.java
andy-goryachev-oracle: Originally it was an interface.  Then @prrace said "Why an interface ? It is easier to evolve a class when you want to add more info."

Since this thing is sealed, it probably does not matter, and access to class methods is _two_ nanoseconds faster (not that's important).

The only good thing about interface is the ease of mocking in testing, but we don't do that.

modules/javafx.graphics/src/main/java/javafx/scene/text/LayoutInfo.java
andy-goryachev-oracle: A new instance gets created for each caller, we don't care what they do to it.  Maybe we should say that?

From the API purity perspective, it could be, but an immutable list would incur a copy overhead, do we really need that?

It also looks like there is no easy way of creating an immutable wrapper similar to JDK's List12, so we'd need to invent one.

(Note: the original design called for an array, but List was recommended because of stream/etc)

kevinrushforth: You can use `Collections.unmodifiableList(List)` to wrap the list without copying.

modules/javafx.graphics/src/main/java/javafx/scene/text/CaretInfo.java
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * Constructor for subclasses to call.
+     */
+    protected CaretInfo() {
+    }
+
+    /**
+     * Returns the number of lines representing the caret.
```
kevinrushforth, line 42: Since this is in the text class, the term "lines" is ambiguous. You mean geometric lines, not text lines, right?

modules/javafx.graphics/src/main/java/javafx/scene/text/CaretInfo.java
```diff
@@ -0,0 +1,58 @@
...
+     *
+     * @return the number of parts representing the caret
+     */
+    public abstract int getLineCount();
+
+    /**
+     * Returns the geometry of the line at the specified index.
+     * <p>
+     * The geometry is encoded in an array of [x, ymin, ymax] values which
+     * represent a line drawn from (x, ymin) to (x, ymax).
```
kevinrushforth, line 52: This seems overly restrictive, unless I don't understand what this does. How would this work for caret that is "caret" shaped -- `^` -- or shaped line an "I" with serifs? Presuming that specifying it with a series of geometric lines is sufficient, should it be defined as `[xmin, xmax, ymin, ymax]` instead?

kevinrushforth (commented):
Finished the first pass through the API, with a few more questions.

modules/javafx.graphics/src/main/java/javafx/scene/text/LayoutInfo.java
```diff
@@ -0,0 +1,121 @@
...
+    protected LayoutInfo() {
+    }
+
+    /**
+     * Returns the logical bounds of the layout:
+     * <ul>
+     * <li>{@code minX} is always zero
+     * <li>{@code minY} is the ascent of the first line (negative)
+     * <li>{@code width} the width of the widest line
+     * <li>{@code height} the sum of all lines height
```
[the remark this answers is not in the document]
andy-goryachev-oracle: There is an existing related issues
[JDK-8317120](https://bugs.openjdk.org/browse/JDK-8317120) RFE: TextFlow.rangeShape() ignores lineSpacing
[JDK-8317122](https://bugs.openjdk.org/browse/JDK-8317122) RFE: TextFlow.preferredHeight ignores lineSpacing

so perhaps we need to add a `boolean` argument to qualify whether the return value should include the last line's `lineSpacing` or not.

modules/javafx.graphics/src/main/java/javafx/scene/text/CaretInfo.java
```diff
@@ -0,0 +1,58 @@
...
+     *
+     * @param index the line index
+     * @return the array of [x, ymin, ymax] values
+     */
+    public abstract double[] getLineAt(int index);
```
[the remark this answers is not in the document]
andy-goryachev-oracle: well, we could have returned Line2D but we have no such public class!
We have javafx.geometry.Rectangle2D but no corresponding Line2D.

I could use Rectangle2D here, which might also take care of the vertical text support which `will have been added` in the year 252525.

modules/javafx.graphics/src/main/java/javafx/scene/text/CaretInfo.java
andy-goryachev-oracle: We are talking about "geometry", assuming two things:

- the caret "geometry" represents a boundary between two characters and therefore is a line (or two lines in the case of a split caret)
- the text is horizontal

the latter poses a problem - on one hand, FX currently provides no support whatsoever for the vertical text, but that might change in the (distant) future.  So perhaps Rectangle2D should work with the width of 0.

andy-goryachev-oracle:
@prrace please take a look

andy-goryachev-oracle:
keep-alive

andy-goryachev-oracle:
keep-alive

johanvos:
@jperedadnr Can you review this?

modules/javafx.graphics/src/main/java/com/sun/javafx/text/PrismLayoutInfo.java
```diff
@@ -0,0 +1,160 @@
...
+    private final TextLayout layout;
+
+    public PrismLayoutInfo(TextLayout layout) {
+        this.layout = layout;
+    }
+
+    @Override
+    public Rectangle2D getBounds(boolean includeLineSpacing) {
+        BaseBounds b = layout.getBounds();
+        Insets m = insets();
```
jperedadnr, line 57: See my comments below about adding the insets to the layout information. Maybe we could also add another boolean `includeInsets` for this?

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/text/TextLayout.java
```diff
@@ -79,6 +77,12 @@ public interface TextLayout {
 
     public static final int DEFAULT_TAB_SIZE = 8;
 
+    /** Callback to be called for each rectangular shape */
+    @FunctionalInterface
+    public static interface GeometryCallback {
+        public void addRectangle(float left, float top, float right, float bottom);
```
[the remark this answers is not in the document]
andy-goryachev-oracle: the use of float in this interface is correct, but the code in TextUtils::getRange is bad, and i should feel bad.  good catch!

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/text/TextLayout.java
```diff
@@ -233,8 +237,36 @@ public String toString() {
...
+     * Queries the range geometry of the range of text within the text layout for one of the three possible types:
+     * <ul>
+     * <li>{@link #TYPE_STRIKETHROUGH} - strike-through shape
+     * <li>{@link #TYPE_TEXT} - text selection shape
+     * <li>{@link #TYPE_UNDERLINE} - underline shape
+     * </ul>
+     *
+     * @param start the start offset
+     * @param end the end offset
+     * @param the type of the geometry
```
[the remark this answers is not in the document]
andy-goryachev-oracle: eagle eye!

modules/javafx.graphics/src/main/java/com/sun/javafx/text/PrismLayoutInfo.java
andy-goryachev-oracle: Good question!  I don't think this is needed: the call returns the layout bounds (in the owning node coordinate system).

The insets only change the position of the layout within the owning node, unlike lineSpacing which actually changes the shape by adding or removing the last line spacing.

I think this method is ok as is.

modules/javafx.graphics/src/main/java/javafx/scene/text/LayoutInfo.java
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * Returns the geometry of the text selection, as an immutable list of {@code Rectangle2D} objects,
+     * for the given start and end offsets.
+     *
+     * @param start the start offset
+     * @param end the end offset
+     * @param includeLineSpacing determines whether the result includes the line spacing
+     * @return the immutable list of {@code Rectangle2D} objects
+     */
+    public abstract List<Rectangle2D> selectionShape(int start, int end, boolean includeLineSpacing);
```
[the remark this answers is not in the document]
andy-goryachev-oracle: Good point.  The word 'shape' might be confusing, although it is still correct in human terms (it is a shape, after all).  I think it's ok, but we could use the word "geometry" for these methods `selectionGeometry` / `strikeThroughGeometry` / `underlineGeometry`, what do other people think?

As for using `javafx.scene.shape.Rectangle` - it is too heavy of an object, with properties and everything, it's not what we ask from this method.

modules/javafx.graphics/src/main/java/javafx/scene/text/TextLineInfo.java
```diff
@@ -0,0 +1,57 @@
...
+ * justification.
+ * <li>
+ * {@code height} - the height of the line.
+ * The height of the line is sum of the max ascent, max descent, and
+ * max line gap of all the fonts in the line.
+ * </ul>
+ *
+ * @since 25
+ */
+public record TextLineInfo(int start, int end, Rectangle2D bounds) {
```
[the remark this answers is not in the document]
andy-goryachev-oracle: it was intentional decision to use the light weight `Rectangle2D` class instead of a `Shape` (which extends `Node`).

andy-goryachev-oracle:
thank you for thoughtful comments, @jperedadnr !

andy-goryachev-oracle:
please review

modules/javafx.graphics/src/main/java/com/sun/javafx/text/PrismCaretInfo.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public int getSegmentCount() {
+        return parts.length;
+    }
+
+    @Override
+    public Rectangle2D getSegmentAt(int index) {
+        return parts[index];
```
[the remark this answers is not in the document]
mstr2: While true, I think we should throw `IndexOutOfBoundsException` rather than `ArrayIndexOutOfBoundsException` (use `Objects.checkIndex()` for that). This should also be specified with a `@throws` javadoc tag in `CaretInfo`.

modules/javafx.graphics/src/main/java/com/sun/javafx/text/PrismCaretInfo.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        sb.append("PrismCaretInfo{parts=[");
```
mstr2, line 54: I think you shouldn't leak implementation details via the `toString()` method (the name of the implementing class, as well as `parts=`, which is not a term that appears in the API). Maybe just drop `PrismCaretInfo{parts=` entirely.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/text/TextLayout.java
```diff
@@ -233,8 +237,36 @@ public String toString() {
...
+     * <li>{@code [x, y, h]} - corresponds to a single line from (x, y) to (x, y + h)
+     * <li>{@code [x, y, x2, h]} - corresponds to a split caret drawn as two lines, the first line
+     * drawn from (x, y) to (x, y + h/2), the second line drawn from (x2, y + h/2) to (x2, y + h).
+     * </ul>
+     *
+     * @param offset the character offset
+     * @param leading whether the caret is biased on the leading edge of the character
+     * @return the caret geometry
+     */
+    public float[] getCaretGeometry(int offset, boolean leading);
```
mstr2, line 256: Does this method throw any exceptions? If so, please specify.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/text/TextLayout.java
```diff
@@ -233,8 +237,36 @@ public String toString() {
...
+     * <li>{@link #TYPE_TEXT} - text selection shape
+     * <li>{@link #TYPE_UNDERLINE} - underline shape
+     * </ul>
+     *
+     * @param start the start offset
+     * @param end the end offset
+     * @param type the type of the geometry
+     * @param client the callback to invoke for each rectangular shape
+     */
+    public void getRange(int start, int end, int type, GeometryCallback client);
```
mstr2, line 271: Does this method throw any exceptions?

modules/javafx.graphics/src/main/java/javafx/scene/text/LayoutInfo.java
```diff
@@ -0,0 +1,121 @@
...
+ * with purpose of querying the details of the layout such as break up of the text into lines,
+ * as well as geometry of other shapes derived from the layout (selection, underline, etc.).
+ * <p>
+ * The information obtained via this object may change after the next layout cycle, which may come as a result
+ * of actions such as resizing of the container, or modification of certain properties.
+ * For example updating the text or the font might change the layout, but a change of color would not.
+ *
+ * @since 25
+ */
+public abstract class LayoutInfo {
```
mstr2, line 41: Any exceptions thrown by methods of this class?

modules/javafx.graphics/src/main/java/com/sun/javafx/text/PrismCaretInfo.java
andy-goryachev-oracle: makes sense.

On a related note, we don't seem to have a consistent standard for toString() output (probably no one does!), so maybe it's worth polling the devs about it.

Standardized log output might be useful when one has to go through masses of old/transient logs, or when scraping the logs for analytics.  A JSON output might be useful since it offers a cheap structured form, although it suffers from two issues: verbosity and difficulty to encode binary data and object types easily.

Any ideas?

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/text/TextLayout.java
andy-goryachev-oracle: This method does not throw any exceptions, so none are specified.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/text/TextLayout.java
andy-goryachev-oracle: no.

modules/javafx.graphics/src/main/java/javafx/scene/text/CaretInfo.java
```diff
@@ -56,6 +56,8 @@ protected CaretInfo() {
      *
      * @param index the line index
      * @return the bounds of the caret segment
+     * @throws IndexOutOfBoundsException if the index is outside of the range if the index is out of range
```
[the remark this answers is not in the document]
andy-goryachev-oracle: daylight saving time strikes!
thanks for noticing.

modules/javafx.graphics/src/main/java/javafx/scene/text/LayoutInfo.java
mstr2: Can this class be sealed?

modules/javafx.graphics/src/main/java/javafx/scene/text/TextLineInfo.java
```diff
@@ -0,0 +1,57 @@
...
+import javafx.geometry.Rectangle2D;
+
+/**
+ * Provides the information about a text line in a text layout.
+ *
+ * @param start the start offset for the line
+ * @param end the end offset for the line (index of the last character + 1)
+ * @param bounds the bounds of the text line, in local coordinates:
+ * <ul>
+ * <li>
```
mstr2, line 36: What do you think about indenting the `<ul>` (so that it's clearer that it belongs to the `@param` tag), and also indent the `<li>` elements to make them stand out better:

```
@param bounds the bounds of the text line, in local coordinates:
       <ul>
           <li>{@code minX} - the x origin of the line (relative to the layout).
               The x origin is defined by TextAlignment of the text layout, always zero
               for left-aligned text.
           <li>{@code minY} - the ascent of the line (negative).
               The ascent of the line is the max ascent of all fonts in the line.
           ...

```

I find this formatting to be more readable in source code form.

modules/javafx.graphics/src/main/java/javafx/scene/text/CaretInfo.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Returns the geometry of the segment at the specified index.
+     *
+     * @param index the line index
+     * @return the bounds of the caret segment
+     * @throws IndexOutOfBoundsException if the index is out of range
+     *     {@code (index < 0 || index > getSegmentCount())}
+     */
+    public abstract Rectangle2D getSegmentAt(int index);
```
mstr2, line 62: The naming style in `CaretInfo`, `LayoutInfo`, and `TextLineInfo` is a bit inconsistent in that it uses both `getFoo()` as well as `foo()` (both with and without method arguments). I think you should decide on one way or the other.

andy-goryachev-oracle:
> Can this class be sealed?

Theoretically, yes, but what problem would that solve?

I can understand when sealing would make the platform more secure, but in this case I see no value whatsoever.  Could you give me an example where sealing the `LayoutInfo` class might help?

modules/javafx.graphics/src/main/java/javafx/scene/text/CaretInfo.java
andy-goryachev-oracle: `TextLineInfo` is a record, I don't want to add duplicate getXXX() aliases there.

The other (in `LayoutInfo`) are modeled after `Text`/`TextFlow` (e.g. `underlineShape()`), I have no problem with any of these.  I don't think we have strict rules regarding naming, and these are not java beans to utilize some sort of naming pattern in accessors, but I could be missing something.

Could you be more specific?

mstr2:
> > Can this class be sealed?
> 
> Theoretically, yes, but what problem would that solve?
> 
> I can understand when sealing would make the platform more secure, but in this case I see no value whatsoever. Could you give me an example where sealing the `LayoutInfo` class might help?

I think a class should always be final or sealed by default, no further justification required (it's one of the defaults the Java language arguably got wrong). Justification should be given for a class to be extensible, as that will invariably add to the incompatibility budget required for future changes. Designing a class for extensibility is also usually very different from designing a sealed class, and there are many examples in JavaFX where careless extensibility leads to problems down the line.
 
> edit: I can see the value of sealing in some security context, but I really dislike how this exposes the internals via `permits`.

How so? You can't do anything with non-exported permitted subclasses, so it's not leaking anything in that sense. Javadoc also doesn't show you non-exported permitted classes.

modules/javafx.graphics/src/main/java/javafx/scene/text/CaretInfo.java
mstr2: I understand the argument for `TextLineInfo`, I guess I'm looking more at `LayoutInfo` here. My question is whether it is really the best choice to sacrifice local consistency within a single class just to have it look more similar to equally inconsistently named methods in another class? Anyway, I just wanted to point this out, that's not a hill for me to die on.

andy-goryachev-oracle:
> I think a class should always be final or sealed by default

contr-argument: testing.  having extendable classes makes it easier to mock, although in the case of a platform your argument carries greater weight.

I can make it `sealed`.

modules/javafx.graphics/src/main/java/javafx/scene/text/TextLineInfo.java
```diff
@@ -0,0 +1,57 @@
...
+import javafx.geometry.Rectangle2D;
+
+/**
+ * Provides the information about a text line in a text layout.
+ *
+ * @param start the start offset for the line
+ * @param end the end offset for the line (index of the last character + 1)
+ * @param bounds the bounds of the text line, in local coordinates:
+ * <ul>
+ *   <li>
```
mstr2, line 36: We usually indent with 4 spaces.

mstr2 (approved):
Looks good. I've left one minor comment.

modules/javafx.graphics/src/main/java/javafx/scene/text/TextLineInfo.java
andy-goryachev-oracle: in this economy, spaces are expensive ;-)

kevinrushforth (commented):
I did a first pass review of the public API -- I didn't look at the implementation yet.

modules/javafx.graphics/src/main/java/javafx/scene/text/CaretInfo.java
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public abstract int getSegmentCount();
+
+    /**
+     * Returns the geometry of the segment at the specified index.
+     *
+     * @param index the line index
+     * @return the bounds of the caret segment
+     * @throws IndexOutOfBoundsException if the index is out of range
+     *     {@code (index < 0 || index > getSegmentCount())}
```
kevinrushforth, line 60: Shouldn't this be `>= getSegmentCount()`.

modules/javafx.graphics/src/main/java/javafx/scene/text/LayoutInfo.java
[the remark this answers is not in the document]
kevinrushforth: I also would like to hear thoughts about compatibility concerns around fixing [JDK-8341438](https://bugs.openjdk.org/browse/JDK-8341438).

If we are not going to fix it in the same release (JavaFX 25) as this Text LayoutInfo enhancement, I think we need a NOTE in the javadoc to clarify this difference.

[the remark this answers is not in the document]
andy-goryachev-oracle: I like the idea of `getX*Geometry()`, to further distance the new API from the buggy ones in `Text`/`TextFlow`.

modules/javafx.graphics/src/main/java/javafx/scene/text/TextLineInfo.java
andy-goryachev-oracle: what's a "tidy" and is it a part of our build/validation?

mstr2: The closing `</li>` tag is not required in HTML, and Javadoc doesn't seem to object. Is this a requirement of another tool?

kevinrushforth: OK. I (mistakenly) thought that this was required. To answer Andy's question, [tidy](https://www.html-tidy.org/) is a toolto validate HTML5 for various formatting problems which can cause problems with some browsers. The JDK team used to periodically run it on their javadoc, but I don't think it is run continuously. Jon Gibbons filed [JDK-8184220](https://bugs.openjdk.org/browse/JDK-8184220) several years ago to "tidy" up our docs, which we did. This did not flag any missing `</li>` tags.

So: no change needed.

modules/javafx.graphics/src/main/java/javafx/scene/text/CaretInfo.java
andy-goryachev-oracle: The widths are 0 with the PrismTextLayout.  I thought it'll be easier to package this information into a rectangle than simply return a bunch of values in an array and then try to describe the layout.  It also better in terms of extensibility (e.g. if we ever support vertical layouts).

modules/javafx.graphics/src/main/java/javafx/scene/text/Text.java
```diff
@@ -859,7 +862,8 @@ public final PathElement[] getSelectionShape() {
     }
 
     /**
-     * The shape of the selection in local coordinates.
+     * The shape of the selection in coordinates
+     * relative to the font base line.
```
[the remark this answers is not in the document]
andy-goryachev-oracle: reverted.

modules/javafx.graphics/src/main/java/javafx/scene/text/Text.java
```diff
@@ -2070,4 +2090,33 @@ public Object queryAccessibleAttribute(AccessibleAttribute attribute, Object...
...
         }
     }
+
+    /**
+     * Returns the object which provides a view into the text layout for this node, which allows for querying
+     * the details of the layout.
+     * <p>
+     * While there is no general guarantee that successive invocations of this method return the same instance,
+     * it is safe to either cache this object or call this method each time, since the information obtained from
+     * this lightweight object remains valid until the next layout cycle.
```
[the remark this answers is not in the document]
andy-goryachev-oracle: good comment.  how does this sound:
```
Returns the object which provides a snapshot of the text layout geometry for this node.
```

modules/javafx.graphics/src/main/java/javafx/scene/text/LayoutInfo.java
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * Returns the logical bounds of the layout.
+     * Depending on {@code includeLineSpacing}, the return value may include the line spacing after the
+     * last line of text.
+     *
+     * @param includeLineSpacing determines whether the line spacing after last text line should be included
+     * @return the layout bounds
+     */
+    public abstract Rectangle2D getBounds(boolean includeLineSpacing);
```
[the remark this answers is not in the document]
andy-goryachev-oracle: No, this is strictly a 2D API, everything is expressed in lightweight Rectangle2D's.

modules/javafx.graphics/src/main/java/javafx/scene/text/CaretInfo.java
andy-goryachev-oracle: (hangs head in shame)

modules/javafx.graphics/src/main/java/javafx/scene/text/LayoutInfo.java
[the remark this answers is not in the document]
andy-goryachev-oracle: Good point!  I think we should do the following in https://bugs.openjdk.org/browse/JDK-8341438 (after this PR gets integrated):

- deprecate the existing (buggy) methods, but keep them for compatibility
- add new methods, named `getX*Shape()`, with additional arguments where needed
- update javadoc for the existing methods, explaining the problem (possibly)

This will fix the issue and avoid breaking existing applications which applied workarounds.

modules/javafx.graphics/src/main/java/javafx/scene/text/TextFlow.java
```diff
@@ -244,6 +248,18 @@ public final PathElement[] underlineShape(int start, int end) {
...
 
+    /**
+     * Returns the shape for the strike-through in local coordinates.
+     *
+     * @param start the beginning character index for the range
+     * @param end the end character index (non-inclusive) for the range
+     * @return an array of {@code PathElement} which can be used to create a {@code Shape}
+     * @since 25
+     */
+    public final PathElement[] strikeThroughShape(int start, int end) {
```
andy-goryachev-oracle, line 259: Perhaps this method should be renamed to `getStrikeThroughShape()` to be consistent with the proposed evolution of the buggy methods in https://bugs.openjdk.org/browse/JDK-8341438

andy-goryachev-oracle:
I'll update the CSR once the javadoc settles.

kevinrushforth (commented):
The changes to the API look good. I left a few wording suggestions. Once those are resolved, update the CSR.

I see that this PR lists [JDK-8341671](https://bugs.openjdk.org/browse/JDK-8341671) and [JDK-8341672](https://bugs.openjdk.org/browse/JDK-8341672) as additional issues solved. They are included as part of the text LayoutInfo, so it would be better to close them as duplicates and remove them from this PR (using `/issue remove`).

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/text/TextLayout.java
mstr2: Instead of returning `float[]` here, this would be a good candidate to return a discriminated union instead, something like:
```java
sealed interface CaretGeometry {
    record Single(float x, float y, float height) implements CaretGeometry {}
    record Split(float x1, float x2, float y, float height) implements CaretGeometry {}
}
```

This has two advantages:
1. It's more expressive, as you don't need to look up what the array values represent.
2. Instead of disciminating between the two cases by inspecting the array length as you do in `PrismLayoutInfo::caretInfoAt`:

   ```java
   Rectangle2D[] parts;
   if (c.length == 3) {
       ..
       parts = ...
   } else {
       ...
       parts = ...
   }
   ```
   ...you can use an exhaustive switch expression instead:
   ```
   Rectangle2D[] parts = switch (c) {
       case Single s -> new Rectangle2D[] {
           new Rectangle2D(s.x() + dx, s.y() + dy, 0, s.height())
       };
   
       case Split s -> new Rectangle2D[] {
           new Rectangle2D(...),
           new Rectangle2D(...)
       };
   };
   ```

   This is easier to read, and you'll get the compiler checks for free.

modules/javafx.graphics/src/main/java/javafx/scene/text/Text.java
```diff
@@ -1108,6 +1112,18 @@ public final PathElement[] underlineShape(int start, int end) {
...
 
+    /**
+     * Returns the shape for the strike-through in local coordinates.
+     *
+     * @param start the beginning character index for the range
+     * @param end the end character index (non-inclusive) for the range
+     * @return an array of {@code PathElement} which can be used to create a {@code Shape}
+     * @since 25
+     */
+    public final PathElement[] getStrikeThroughShape(int start, int end) {
```
kevinrushforth, line 1123: I just noticed that this is named `getStrikeThroughShape` rather than `strikeThroughShape`. The latter would match the similar `rangeShape`, `caretShape`, and `underlineShape` methods. We had discussed that you might propose to fix [JDK-8341438](https://bugs.openjdk.org/browse/JDK-8341438) by deprecating `rangeShape` and `underlineShape` and adding new `getRangeShape` and `getUnderlineShape` methods (although that might be problematic given the `caretShape` property, which necessarily already has a no-arg `getCaretShape()` method).

Did you name this method `getStrikeThroughShape` with the intention that it would have the fixed semantics of applying the insets from its inception? If so, then I recommend documenting that fact in this method.

kevinrushforth (commented):
Updated API looks good. I left a comment on the CSR.

I left a comment on one other naming / semantic issue that I just noticed.

modules/javafx.graphics/src/main/java/javafx/scene/text/Text.java
andy-goryachev-oracle: Yes, you are right - and I plan to address this in https://bugs.openjdk.org/browse/JDK-8341438 (as it will be a separate change).

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/text/TextLayout.java
mstr2: The point is that it's less complicated to have self-describing, strongly typed, and compiler-checked code. Your code is more complicated because it's brittle, untyped, and less robust against refactoring.

Just to make sure that we're on the same page here: "so it was made as low level as possible" sounds like you deliberately chose to forgo strong typing... to achieve _what_?

mstr2: Discriminated unions are an amazing tool in the type system if you want to return a number of differently shaped things, which is exactly what you're doing here. You make it sound like giving names to things is somehow a downside, when it's the exact opposite (classes/records are not a finite resource, it's what we use to name things).

I do consider readibility and maintainability to be important, yes. A float array is none of that.

modules/javafx.graphics/src/main/java/javafx/scene/text/Text.java
andy-goryachev-oracle: We discussed this with @kevinrushforth - I am going to remove the `getStrikeThroughShape()` method from Text/TextFlow in this PR, and add it under either a separate ticket, or as a part of https://bugs.openjdk.org/browse/JDK-8341438

andy-goryachev-oracle: Created https://bugs.openjdk.org/browse/JDK-8357594 for `Text::strikeThroughShape` and `TextFlow::getStrikeThroughShape`.

modules/javafx.graphics/src/main/java/com/sun/javafx/text/PrismLayoutInfo.java
```diff
@@ -0,0 +1,138 @@
...
+    @Override
+    public int getTextLineCount() {
+        return layout.getLines().length;
+    }
+
+    @Override
+    public List<TextLineInfo> getTextLines(boolean includeLineSpacing) {
+        TextLine[] lines = layout.getLines();
+        Insets m = insets();
+        double dx = m.getLeft(); // TODO rtl?
```
mstr2, line 74: This comment looks like it's a question to the reader. Does the code not work for RTL layouts?

andy-goryachev-oracle: JavaFX does not support RTL fully - there is a whole series of issues waiting to be fixed.
See https://bugs.openjdk.org/browse/JDK-8343557

modules/javafx.graphics/src/main/java/javafx/scene/text/TextLineInfo.java
```diff
@@ -0,0 +1,57 @@
...
+ *     {@code minX} - the x origin of the line (relative to the layout).
+ *     The x origin is defined by TextAlignment of the text layout, always zero
+ *     for left-aligned text.
+ *   <li>
+ *     {@code minY} - the ascent of the line (negative).
+ *     The ascent of the line is the max ascent of all fonts in the line.
```
kevinrushforth, line 42: Shouldn't that be "glyphs" not "fonts"?

modules/javafx.graphics/src/main/java/javafx/scene/text/TextLineInfo.java
```diff
@@ -0,0 +1,57 @@
...
+ *   <li>
+ *     {@code width} - the width of the line.
+ *     The width for the line is sum of all the run widths in the line, it is not
+ *     affect by the wrapping width but it will include any changes caused by
+ *     justification.
+ *   <li>
+ *     {@code height} - the height of the line.
+ *     The height of the line is sum of the max ascent, max descent, and
+ *     max line gap of all the fonts in the line.
```
kevinrushforth, line 51: Shouldn't that be "glyphs" not "fonts"?

modules/javafx.graphics/src/main/java/javafx/scene/text/TextLineInfo.java
```diff
@@ -0,0 +1,57 @@
...
+ *   <li>
+ *     {@code minX} - the x origin of the line (relative to the layout).
+ *     The x origin is defined by TextAlignment of the text layout, always zero
+ *     for left-aligned text.
+ *   <li>
+ *     {@code minY} - the ascent of the line (negative).
+ *     The ascent of the line is the max ascent of all fonts in the line.
+ *   <li>
+ *     {@code width} - the width of the line.
+ *     The width for the line is sum of all the run widths in the line, it is not
```
kevinrushforth: minor: "for the line" --> "of the line"

modules/javafx.graphics/src/main/java/javafx/scene/text/CaretInfo.java
```diff
@@ -0,0 +1,63 @@
...
+     */
+    protected CaretInfo() {
+    }
+
+    /**
+     * Returns the number of segments representing the caret.
+     *
+     * @return the number of segments representing the caret
+     */
+    public abstract int getSegmentCount();
```
kevinrushforth, line 52: Do you think it's worth adding a `List<Rectangle2D>getSegments()` method in addition to (or instead of) `getSegmentCount` and `getSegmentAt`?

kevinrushforth (commented):
The API looks good given the removal of the Text/TextFlow getStrikeTrhoughShape methods. I added one question about whether you want to consider adding a getting to CaretInfo to return the list of rectangles in a single call, but I'll leave that up to you.

I pointed out what I think are a couple doc typos in TextLine info.

I think this is close to being ready. I'll review the implementation and do some testing.

kevinrushforth:
@prrace Can you also review the API changes?

modules/javafx.graphics/src/main/java/javafx/scene/text/TextLineInfo.java
andy-goryachev-oracle: This definition is copied from `com.sun.javafx.scene.text.TextLine::getBounds` L58.
I'll let Phil confirm that, but since it's a max ascent, it might be a function of the font rather than the particular glyphs.

modules/javafx.graphics/src/main/java/javafx/scene/text/CaretInfo.java
andy-goryachev-oracle: Probably unnecessary, because the caret has one segment most of the time.  When bidi is involved, the caret might have 2 segments.  It will never have 3 or more segments.

modules/javafx.graphics/src/main/java/javafx/scene/text/TextLineInfo.java
kevinrushforth: I asked Phil offline and he confirmed that this should be "font" not "glyph", which makes sense, since what you want to know is how tall the line should be to accommodate any glyph that you might display with that font.

modules/javafx.graphics/src/main/java/javafx/scene/text/TextLineInfo.java
kevinrushforth: Nope. It's correct as is.

kevinrushforth:
I ran the tests included with this fix on my macOS 14 system and see 3 failures:

```
$ gradle --continue --info -PTEST_ONLY=true -PFULL_TEST=true -PUSE_ROBOT=true :systemTests:test --tests TextFlow_TextLayout_Test --tests Text_TextLayout_Test

Text_TextLayout_Test > testSelection() FAILED
    org.opentest4j.AssertionFailedError: expected: <0.0> but was: <-23.203125>
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:81)
        at app//org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1014)
        at app//test.robot.javafx.scene.Text_TextLayout_Test.testSelection(Text_TextLayout_Test.java:205)

Text_TextLayout_Test > testTextLines() FAILED
    org.opentest4j.AssertionFailedError: expected: <0.0> but was: <-23.203125>
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:81)
        at app//org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1014)
        at app//test.robot.javafx.scene.Text_TextLayout_Test.testTextLines(Text_TextLayout_Test.java:155)

Text_TextLayout_Test > testCaretInfo() FAILED
        at app//test.robot.javafx.scene.Text_TextLayout_Test.testCaretInfo(Text_TextLayout_Test.java:87)
```

andy-goryachev-oracle:
good catch!  the failures not specific to macOS 14.
modified the headful tests to use top text origin (all the other choices are handled by headless tests).

launched an internal headful run: job 378

edit: apart from one unrelated failure (QPathTest timeout on macOS) all headful tests pass.

tests/system/src/test/java/test/robot/javafx/scene/TextFlow_TextLayout_Test.java
```diff
@@ -0,0 +1,382 @@
...
+    @Test
+    public void testCaretInfo() {
+        setText("__________\n", "______\n", "_\n");
+        waitForIdle();
+        LayoutInfo la = textFlow.getLayoutInfo();
+
+        CaretInfo ci = la.caretInfoAt(0, true);
+
+        // caret is one line
+        assertEquals(ci.getSegmentCount(), 1);
```
kevinrushforth: Expected and actual are reversed, here and a few others below (most are OK).

tests/system/src/test/java/test/robot/javafx/scene/TextFlow_TextLayout_Test.java
```diff
@@ -0,0 +1,382 @@
...
+    public void testTextLines() {
+        setText("__\n", "____\n", "______");
+        waitForIdle();
+        LayoutInfo la = textFlow.getLayoutInfo();
+
+        // spacing = 0
+
+        assertEquals(la.getTextLineCount(), 3);
+        List<TextLineInfo> ls = la.getTextLines(false);
+        assertNotNull(ls);
```
kevinrushforth, line 136: Minor: maybe check the following, too?

```
        assertEquals(3, ls.size());
```

tests/system/src/test/java/test/robot/javafx/scene/Text_TextLayout_Test.java
```diff
@@ -0,0 +1,366 @@
...
+    @Test
+    public void testCaretInfo() {
+        setText("__________\n______\n_\n");
+        waitForIdle();
+        LayoutInfo la = text.getLayoutInfo();
+
+        CaretInfo ci = la.caretInfoAt(0, true);
+
+        // caret is one line
+        assertEquals(ci.getSegmentCount(), 1);
```
kevinrushforth: Expected and actual are reversed, here and a few others below (most are OK).

kevinrushforth (commented):
The fix looks good. The tests look good, although I noted a few "assertEquals" in the test with expected and actual parameters reserved (meaning the error message would be confusing if they ever failed).

tests/system/src/test/java/test/robot/javafx/scene/TextFlow_TextLayout_Test.java
andy-goryachev-oracle: in the other test as well...  good catch!

andy-goryachev-oracle:
Thank you all for reviewing this PR!

PR #1598, opened 2024-10-11
8329098: Support "@1x" image naming convention as fallback

lukostyra:
This PR adds a title-mentioned fallback to `ImageStorage.java` and a test for that specific scenario.

In `ImageStorage.loadAll()` we still prefer the Image path that was provided by the user, but if it is missing we will now try to add a `@1x` suffix to Image name and give it one last try.

Added test includes a binary file `checkers@1x.png`. This was added to differentiate it from the collection of `checker.png` files and their higher scale variants, but it is a direct copy of already existing `checker.png` test resource, just with a different name and suffix. Test should fail without a change in `ImageStorage.java` and pass with it.

---------
### Progress

### Warnings
&nbsp;⚠️ Patch contains a binary file (modules/javafx.graphics/src/test/resources/test/com/sun/javafx/iio/lightblue.png)
&nbsp;⚠️ Patch contains a binary file (modules/javafx.graphics/src/test/resources/test/com/sun/javafx/iio/lightblue@1x.png)

### Issue
 * [JDK-8329098](https://bugs.openjdk.org/browse/JDK-8329098): Support "@1x" image naming convention as fallback (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1598/head:pull/1598` \
`$ git checkout pull/1598`

`$ git checkout pull/1598` \
`$ git pull https://git.openjdk.org/jfx.git pull/1598/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1598`

`$ git pr show -t 1598`

</details>

<a href="https://git.openjdk.org/jfx/pull/1598.diff">https://git.openjdk.org/jfx/pull/1598.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1598#issuecomment-2407507827)


kevinrushforth (approved):
I left one minor comment on the fallback and one minor comment on the test. I don't think either of them are that important, so I'll approve it as is. I'll reapprove if you decide to make any changes.

modules/javafx.graphics/src/main/java/com/sun/javafx/iio/ImageStorage.java
```diff
@@ -380,7 +380,16 @@ public ImageFrame[] loadAll(String input, ImageLoadListener listener,
...
+                        try {
+                            theStream = ImageTools.createInputStream(input);
+                        } catch (IOException ignored) {
+                        }
+                    }
+
+                    if (theStream == null) {
+                        // last fallback, try to see if the file exists with @1x suffix
+                        String scaled1xName = ImageTools.getScaledImageName(input, 1);
+                        theStream = ImageTools.createInputStream(scaled1xName);
```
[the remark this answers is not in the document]
lukostyra: Capturing the earlier message and rethrowing it uses the original stack trace:
```
java.io.FileNotFoundException: C:\Users\Lukasz\dev\code\jfx\rt\modules\javafx.graphics\build\resources\test\test\com\sun\javafx\iio\checkers.png (The system cannot find the file specified)
    at java.base/java.io.FileInputStream.open0(Native Method)
    at java.base/java.io.FileInputStream.open(FileInputStream.java:213)
    ...
    at javafx.graphics@24-internal/com.sun.javafx.iio.ImageStorage.loadAll(ImageStorage.java:385)
```

I figured wrapping it with an additional message mentioning original path is a better solution, that way the stack trace is not misleading.

modules/javafx.graphics/src/main/java/com/sun/javafx/iio/ImageStorage.java
```diff
@@ -380,7 +380,20 @@ public ImageFrame[] loadAll(String input, ImageLoadListener listener,
...
+                        try {
+                            theStream = ImageTools.createInputStream(input);
+                        } catch (IOException ignored) {
+                        }
+                    }
+
+                    if (theStream == null) {
+                        try {
+                            // last fallback, try to see if the file exists with @1x suffix
+                            String scaled1xName = ImageTools.getScaledImageName(input, 1);
```
hjohn, line 392: I think this function does not properly handle cases where the input may already have a scale specified.

If I want to load `testimg@1x.png` then this function will return `testimg@1x@1x.png`

I'm not sure what is normal here, but I'd think that if a scale was already specified that we shouldn't do any fallback.

modules/javafx.graphics/src/main/java/com/sun/javafx/iio/ImageStorage.java
```diff
@@ -380,7 +380,20 @@ public ImageFrame[] loadAll(String input, ImageLoadListener listener,
                     }
 
                     if (theStream == null) {
-                        theStream = ImageTools.createInputStream(input);
+                        try {
+                            theStream = ImageTools.createInputStream(input);
+                        } catch (IOException ignored) {
```
hjohn, line 385: It looks like this abstracts away errors like `FileNotFoundException`, and it will do a fallback in all cases, even if it is not a 404 or `FileNotFoundException` -- errors like network problems, permission problems, etc, will also trigger a fallback (likely resulting in the same error).

Maybe I'm being too picky :)

hjohn (commented):
@mstr2 as you're busy in the `ImageStorage` class as well and know how it works, perhaps you could take a peek at this PR as well.

modules/javafx.graphics/src/main/java/com/sun/javafx/iio/ImageStorage.java
```diff
@@ -380,7 +380,20 @@ public ImageFrame[] loadAll(String input, ImageLoadListener listener,
...
+                        }
+                    }
+
+                    if (theStream == null) {
+                        try {
+                            // last fallback, try to see if the file exists with @1x suffix
+                            String scaled1xName = ImageTools.getScaledImageName(input, 1);
+                            theStream = ImageTools.createInputStream(scaled1xName);
+                        } catch (IOException e) {
+                            throw new IOException(e.getMessage() + " (original resource path: " + input + ")");
```
mstr2, line 395: I don't like this exception message, since it will now say something like
> yourImage@1x.jpeg was not found (original resource path: yourImage.jpeg)

That's confusing, because in all likelihood, I didn't specify that file name in my application code. What is the significance of mentioning `@1x.jpeg` in the exception message? We don't mention any of the other scale factors.

I'd rather just call `ImageTools.createInputStream(input)` again, and let the resulting exception bubble up.

We should _only_ mention the scale factor in the exception message if the scale factor was explicitly specified by the application developer.

modules/javafx.graphics/src/main/java/com/sun/javafx/iio/ImageStorage.java
lukostyra: This technically translates to the for-loop higher up that tries to fetch other scaled versions of an image - if you want to explicitly load `foo@1x.png` it will look for `foo@1x@2x.png` etc.

Should we just assume that with `@Nx` provided in the name we will only load that one specific scale?

mstr2: I would say so, yes. If the user specifies any `@Nx`, then we only look for this particular version.

kevinrushforth: Probably, but this is out of scope for this PR / JBS bug. Please file a follow-up bug.

lukostyra: Filed [JDK-8342530](https://bugs.openjdk.org/browse/JDK-8342530)

modules/javafx.graphics/src/main/java/com/sun/javafx/iio/ImageStorage.java
lukostyra: That is technically solved by re-calling `createInputStream(input)` like @mstr2 suggested to get a correct exception with valid message, so I will go with that approach.

mstr2: This is a bit minor, but what do you think about storing the previous exception in L385, and then re-throwing it? This is most likely exceedingly rare, but it might be possible that the reason why the previous call to `createInputStream` failed was only transient, we're not 100% sure that we can reproduce the exception by attempting it again.

lukostyra: This would technically create an ambiguity in the call stack (I checked and it points at in this case L385 instead of where the throw actually happened) but if it's not a problem I can change it.

mstr2: I would think that this is acceptable, since that's the exception message that we expect to see (if the `@1x` version also can't be found).

hjohn: I think the call stack being different is fine; call stacks are there for developers to debug code, and the users that are seeing these will not be debugging that particular code so the callstack being slightly off at the deepest level is fine.  Altenatively, you could still consider adding the `@1x` exception as a supressed exception to the first one (with Exception#addSuppressed`) but I wouldn't bother for this case.

lukostyra: All done, with fallback exception added as suppressed

mstr2 (approved):

hjohn (approved):

kevinrushforth (approved):

PR #1601, opened 2024-10-16
8342462: TextAreaSkin: remove USE_MULTIPLE_NODES

andy-goryachev-oracle:
Removed "not yet fully implemented" USE_MULTIPLE_NODES and related code.

I would like to remove the early unfinished idea of using multiple Text nodes in TextAreaSkin to clean up the code, to make it easier to do fixes for [JDK-8342233](https://bugs.openjdk.org/browse/JDK-8342233) and [JDK-8296266](https://bugs.openjdk.org/browse/JDK-8296266).

Also some minor cleanup.

## Summary of Changes

- removed USE_MULTIPLE_NODES and code paths that correspond to its `true` value
- permanently adding one Text node to `paragraphNodes` Group, keeping the latter for compatibility purposes
- removed any code that scans `paragraphNodes` children
- using `getTextNode()` in place of `((Text)paragraphNodes.getChildren().get(0))`

---------
### Progress

### Issue
 * [JDK-8342462](https://bugs.openjdk.org/browse/JDK-8342462): TextAreaSkin: remove USE_MULTIPLE_NODES (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1601/head:pull/1601` \
`$ git checkout pull/1601`

`$ git checkout pull/1601` \
`$ git pull https://git.openjdk.org/jfx.git pull/1601/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1601`

`$ git pr show -t 1601`

</details>

<a href="https://git.openjdk.org/jfx/pull/1601.diff">https://git.openjdk.org/jfx/pull/1601.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1601#issuecomment-2419784305)

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextAreaSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextAreaSkin.java
@@ -817,13 +817,10 @@ public class TextAreaSkin extends TextInputControlSkin<TextArea> {
         Text paragraphNode = getTextNode();
 
         if (y < contentView.snappedTopInset()) {
-            // Select the character at x in the first row
             return 0;
         } else if (y > contentView.snappedTopInset() + contentView.getHeight()) {
-            // Select the character at x in the last row
             return (textArea.getLength() - paragraphNode.getText().length());
         } else {
-            // Select the character at x in the row at y
             Bounds bounds = paragraphNode.getBoundsInLocal();
             double paragraphViewY = paragraphNode.getLayoutY() + bounds.getMinY();
             if (y >= paragraphViewY
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextAreaSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextAreaSkin.java
@@ -89,9 +89,6 @@ public class TextAreaSkin extends TextInputControlSkin<TextArea> {
 
     final private TextArea textArea;
 
-    // *** NOTE: Multiple node mode is not yet fully implemented *** //
-    private static final boolean USE_MULTIPLE_NODES = false;
-
     private final TextAreaBehavior behavior;
 
     private double computedMinWidth = Double.NEGATIVE_INFINITY;
@@ -161,12 +158,10 @@ public class TextAreaSkin extends TextInputControlSkin<TextArea> {
     public TextAreaSkin(final TextArea control) {
         super(control);
 
+        this.textArea = control;
         // install default input map for the text area control
         this.behavior = new TextAreaBehavior(control);
         this.behavior.setTextAreaSkin(this);
-//        control.setInputMap(behavior.getInputMap());
-
-        this.textArea = control;
 
         caretPosition = new IntegerBinding() {
             { bind(control.caretPositionProperty()); }
@@ -189,8 +184,6 @@ public class TextAreaSkin extends TextInputControlSkin<TextArea> {
                 updateTextNodeCaretPos(control.getCaretPosition());
             }
         });
-
-//        setManaged(false);
 
         // Initialize content
         scrollPane = new ScrollPane();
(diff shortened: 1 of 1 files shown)


kevinrushforth:
Either @arapte or I will review this. I think a single reviewer will be enough, given that the functionality hasn't changed.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextAreaSkin.java
```diff
@@ -745,52 +724,24 @@ private void paragraphEnd(boolean goPastInitialNewline, boolean select) {
...
-            int pEnd = pStart + p.textProperty().getValueSafe().length();
-            if (pEnd >= start) {
-                return p.underlineShape(start - pStart, end - pStart);
-            }
-            pStart = pEnd + 1;
-        }
-        return null;
+    @Override
+    protected PathElement[] getUnderlineShape(int start, int end) {
+        return getTextNode().underlineShape(start, end);
```
kevinrushforth, line 729: The former code would return `null` if `start` > `textNode.textProperty().getValueSafe().length()`. I think this is OK, given that the implementation of `Text::underlineShape` checks -- what do you think? This applies to `getRangeShape`, too.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextAreaSkin.java
```diff
@@ -847,76 +798,43 @@ public Bounds getCaretBounds() {
...
-                        break;
-                    }
-
-                    paragraphOffset += paragraphNode.getText().length() + 1;
-                }
+        Text paragraphNode = getTextNode();
+
+        if (y < contentView.snappedTopInset()) {
+            // Select the character at x in the first row
+            return 0;
```
kevinrushforth, line 821: I see that the implementation of the now-removed `getNextInsertionPoint` method also returned `0` unconditionally, so this is equivalent, but... isn't the comment wrong, since `x` isn't used in this case?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextAreaSkin.java
```diff
@@ -847,76 +798,43 @@ public Bounds getCaretBounds() {
...
+        } else if (y > contentView.snappedTopInset() + contentView.getHeight()) {
+            // Select the character at x in the last row
+            return (textArea.getLength() - paragraphNode.getText().length());
```
kevinrushforth, line 824: Same question about the comment being wrong.

kevinrushforth (approved):
This looks equivalent to the current code, and I couldn't see anything wrong in my testing. I left a couple questions inline.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextAreaSkin.java
andy-goryachev-oracle: I think it's ok.  The javadoc for `TextInputControlSkin::underlineShape()` does not specify a null return value.

In fact, the javadoc for `Text::rangeShape()` and `::underlineShape()` say

`@return an array of {@code PathElement} which can be used to create a {@code Shape}`

implying that it should never be `null` as the return value is typically used in `Shape::getElements().setAll()` or `.addAll()` which will throw an NPE if the thing is null.  In fact, `Text::rangeShape()` (used for underline and range shape) always returns a non-null value.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextAreaSkin.java
andy-goryachev-oracle: comment is indeed wrong, will fix.

kevinrushforth (approved):

PR #1604, opened 2024-10-18
8091673: Public focus traversal API for use in custom controls (with the traversal policy hidden)

andy-goryachev-oracle:
Public focus traversal API for use in custom controls

https://github.com/andy-goryachev-oracle/Test/blob/main/doc/FocusTraversal/FocusTraversal-v3.md

This work is loosely based on the patch
https://cr.openjdk.org/~jgiles/8061673/

And is a scaled down version (with the public traversal policy API removed) of
#1555 

/csr
/reviewers 2

---------
### Progress
- [ ] Change requires CSR request [JDK-8342906](https://bugs.openjdk.org/browse/JDK-8342906) to be approved

### Integration blocker
&nbsp;⚠️ Title mismatch between PR and JBS for issue [JDK-8091673](https://bugs.openjdk.org/browse/JDK-8091673)

### Issues
 * [JDK-8091673](https://bugs.openjdk.org/browse/JDK-8091673): Public focus traversal API for use in custom controls (**Enhancement** - P4) ⚠️ Title mismatch between PR and JBS.
 * [JDK-8342906](https://bugs.openjdk.org/browse/JDK-8342906): Public focus traversal API for use in custom controls (**CSR**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1604/head:pull/1604` \
`$ git checkout pull/1604`

`$ git checkout pull/1604` \
`$ git pull https://git.openjdk.org/jfx.git pull/1604/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1604`

`$ git pr show -t 1604`

</details>

<a href="https://git.openjdk.org/jfx/pull/1604.diff">https://git.openjdk.org/jfx/pull/1604.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1604#issuecomment-2430436718)
</details>


modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -10459,4 +10438,38 @@ void releaseAccessible() {
...
 
+    /**
+     * Requests focus traversal from this {@code Node} in the specified direction.
+     * A successful traversal results in the newly focused {@code Node} visibly indicating its focused state.
+     *
+     * @param direction the direction of focus traversal
+     * @return true if traversal was successful
+     * @since 24
+     */
+    public final boolean requestFocusTraversal(TraversalDirection direction) {
```
andy-goryachev-oracle, line 10449: possible alternative names:

moveFocus
requestFocusTraversal
tryToFocus
tryMoveFocus

kevinrushforth: Of these, I like `requestFocusTraversal` best. Another alternative: `requestMoveFocus`.

kevinrushforth (commented):
API changes look good, although I recommend removing what looks like commented out API, complete with an `@since` that is guaranteed to be wrong.

andy-goryachev-oracle:
cleaned up and keeping `requestFocusTraversal` (though I do like the "try" prefix)

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
hjohn: Plain `requestFocus` can also work as it can be overloaded with the direction parameter.

nlisker: I like the overloaded `requestFocus` idea as well.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -10459,4 +10438,38 @@ void releaseAccessible() {
...
+        TraversalDirectionInternal d;
+        switch (direction) {
```
nlisker, line 10451: Using an expression switch will be less verbose:
```java
TraversalDirectionInternal d = switch (direction) {
    case DOWN -> TraversalDirectionInternal.DOWN;
    case LEFT -> TraversalDirectionInternal.LEFT;
    ...
    default -> null;
};
return d == null ? false : TraversalUtils.traverse(this, d, true);
```

modules/javafx.graphics/src/main/java/javafx/scene/TraversalDirection.java
```diff
@@ -22,20 +22,24 @@
...
  */
-
-package com.sun.javafx.scene.traversal;
+package javafx.scene;
 
 /**
- * Specifies the traversal method.
+ * Specifies the direction of focus traversal.
+ *
+ * @since 24
```
nlisker, line 30: I would add `@see` for the method that uses this enum, otherwise it's rather disconnected from anything.

modules/javafx.graphics/src/main/java/javafx/scene/TraversalDirection.java
```diff
@@ -22,20 +22,24 @@
...
-     * Traversal was initiated programmatically or by clicking.
-     */
-    DEFAULT,
-
-    /**
-     * Traversal was initiated by pressing a key on the keyboard.
-     */
-    KEY
+public enum TraversalDirection {
+    /** Moves focus downward. */
```
nlisker, line 33: The direction itself doesn't move the focus. I would use a different phrasing, maybe something like "Indicates a focus change to the node below the currently focused node".

Same with the others.

nlisker (commented):
Took a quick look, mostly of the API.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -10459,4 +10438,38 @@ void releaseAccessible() {
         }
     }
 
+    /**
+     * Requests focus traversal from this {@code Node} in the specified direction.
+     * A successful traversal results in the newly focused {@code Node} visibly indicating its focused state.
```
mstr2, line 10443: What's the rationale behind always setting the `focusVisible` bit when doing programmatic focus traversal? Visible focus is always cleared when calling the existing `requestFocus()` method, since it is specified to _only_ be set when focus is acquired by keyboard navigation.

Maybe we should have the next node visibly indicate focus if (and only if) the current node visibly indicates focus. This would be consistent with the [Selectors 4 spec](https://www.w3.org/TR/selectors-4/#the-focus-visible-pseudo):
> If the previously-focused element indicated focus, and a script causes focus to move elsewhere, the newly focused element should indicate focus.

> 
> Conversely, if the previously-focused element did not indicate focus, and a script causes focus to move elsewhere, the newly focused element should also not indicate focus.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
andy-goryachev-oracle: I really dislike overloading in this case - the semantics of these two methods is totally different: one moves the focus to the node, one moves the focus somewhere else.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/ListenerHelper.java
```diff
@@ -444,4 +448,60 @@ public void handle(T ev) {
...
+     * @param control the control
+     * @param callback the callback
+     */
+    public void addSceneFocusOwnerListener(Control control, Consumer<Node> callback) {
+        items.add(new ChLi<Scene>() {
+            private ChangeListener<Node> focusListener = (s, p, n) -> {
+                if (isParent(control, n)) {
+                    callback.accept(n);
+                }
+            };
```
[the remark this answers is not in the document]
andy-goryachev-oracle: the two places that use this mechanism do not use `Subscription`.

modules/javafx.graphics/src/main/java/javafx/scene/TraversalDirection.java
andy-goryachev-oracle: good point!

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
andy-goryachev-oracle: The rationale is that this PR does not change the way the current focus subsystem works, only provides the public API.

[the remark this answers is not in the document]
andy-goryachev-oracle: thanks!

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -10459,4 +10438,38 @@ void releaseAccessible() {
         }
     }
 
+    /**
+     * Requests focus traversal from this {@code Node} in the specified direction.
```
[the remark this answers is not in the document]
andy-goryachev-oracle: thanks!

modules/javafx.graphics/src/main/java/javafx/scene/TraversalDirection.java
```diff
@@ -28,18 +28,19 @@
...
  */
 public enum TraversalDirection {
-    /** Moves focus downward. */
+    /** Indicates a focus change to the node below the currently focused node. */
     DOWN,
-    /** Moves focus left. */
+    /** Indicates a focus change to the node to the left of the currently focused node. */
     LEFT,
-    /** Moves focus to the next focusable Node, possibly traversing into the children of the current parent. */
+    /** Indicates a focus change to the next focusable Node, possibly traversing into the children of the current parent. */
```
nlisker, line 38: Minor: "Node" -> "node" like the other places it's used.

modules/javafx.graphics/src/main/java/javafx/scene/TraversalDirection.java
```diff
@@ -28,18 +28,19 @@
...
     NEXT,
-    /** Moves focus to the previous focusable Node. */
+    /** Indicates a focus change to the previous focusable node. */
```
nlisker, line 40: `NEXT` talks about traversing children, does `PREVIOUS` need to mention it too? Could it traverse out of the children of the current parent, or is it not symmetric?

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -10439,37 +10439,24 @@ void releaseAccessible() {
...
-            d = TraversalDirectionInternal.RIGHT;
-            break;
-        case UP:
-            d = TraversalDirectionInternal.UP;
-            break;
-        default:
-            return false;
-        }
-        return TraversalUtils.traverse(this, d, true);
+        TraversalDirectionInternal d = switch(direction) {
```
nlisker, line 10451: Minor: space after `switch`.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -10439,37 +10439,24 @@ void releaseAccessible() {
...
+        case DOWN -> TraversalDirectionInternal.DOWN;
+        case LEFT -> TraversalDirectionInternal.LEFT;
+        case NEXT -> TraversalDirectionInternal.NEXT;
+        case PREVIOUS -> TraversalDirectionInternal.PREVIOUS;
+        case RIGHT -> TraversalDirectionInternal.RIGHT;
+        case UP -> TraversalDirectionInternal.UP;
+        default -> null;
```
nlisker, line 10458: Do we indent labels inside a switch? I find it more readable since `switch` opens a new context. This is also what I remember seeing in OpenJFX. Don't know if it's enforced.

nlisker (commented):
The changes look good, left minor comments. I only reviewed the API.

modules/javafx.graphics/src/main/java/javafx/scene/TraversalDirection.java
andy-goryachev-oracle: I think the whole "children" thing can be dropped, since there is no NEXT_IN_LINE in the public API anymore.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
mstr2: So you're proposing to change the specification of `focusVisible`. That needs to be discussed first.

mstr2: Because `focusVisible` is specified to _only_ be set as a result of keyboard navigation, not as a result of programmatic navigation.

`Node.focusVisible` specifies:
```
    /**
     * Indicates whether this {@code Node} should visibly indicate focus.
     * This flag is set when the node acquires input focus via keyboard navigation,
     * and it is cleared when the node loses focus or when {@link #requestFocus()}
     * is called.
     *
     * @defaultValue false
     * @since 19
     */
```

The [JBS issue](https://bugs.openjdk.org/browse/JDK-8268225) says:
> For JavaFX, "Node.focusVisible" would be set in addition to "Node.focused" if a scene graph node gained focus as a consequence of keyboard input. In all other cases, "Node.focusVisible" would not be set on a node.

andy-goryachev-oracle: To clarify, this PR changes nothing in the way FX behaves in relation to focus.  Nothing.  I only provided a public API for use in custom controls. 

So whatever discrepancy between FX and w3.org spec you mention (not sure why, this isn't an FX spec) applies equally to the existing implementation (if not, please let me know - you found a bug).

However, perhaps, you pointing out that the docs may be improved - and you are doing so in a very difficult manner.  Or maybe you are pointing out the fact that we should bring back the traverse(Node, TraversalDirection, boolean focusVisible) call?

mstr2: I just quoted you the JavaFX specification, not W3C. It very clearly says that the `focusVisible` flag is only set when a node acquires focus as a result of keyboard navigation. You are proposing to extend the API to change focus programmatically, and in the process also set `focusVisible`, thus violating that specification.

andy-goryachev-oracle: Maybe we should say that `Node.requestFocusTraversal()` should only be called by custom components/skins as a response to key press?  I can't see a way to enforce that particular rule.

Or perhaps `Node.focusVisible` doc should be updated to include the reference to `Node.requestFocusTraversal()` ?

What's your opinion?

mstr2: Let's consider a few scenarios:

1. A node has acquired focus via keyboard navigation. Its `focusVisible` bit is set. When I press "tab", I expect the next node to also have visible focus. This is the status quo.
2. Same as 1, but instead of pressing "tab", I press "enter", which has a custom handler that changes focus with the new `requestFocusTraversal` API. I think is is quite natural to assume that the visible bit will also carry over to the new node.
3. A node acquires focus with mouse navigation. Its `focusVisible` bit is not set. When pressing "tab", I expect the next node to have the visible bit.
4. Same as 3, but instead of pressing "tab", I press "enter". The custom handler changes focus programmatically, and the new node does not have the visible bit.

It seems we have the same problem that we originally had with `requestFocus()`, namely that we can't distinguish whether the API was called in response to a key press, or in response to any other event (mouse, programmatically, etc.).

Back when `focusVisible` was added, we decided to always clear the visible bit, because this seemed more correct than to always set it. Like you say, there's no way to tell why someone called `requestFocusTraversal`, and thus we can never clearly know whether to set or to clear the visible bit.

Since we can't know, maybe the right course of action is to allow the caller to specify whether to clear or set the visible bit, because the caller should know that. The API could then be something like: `requestFocusTraveral(TraversalDirection, boolean visible)`.

andy-goryachev-oracle: I think the javadoc for `Node.focusVisible` (and its sibling `Node.focusWithin`) is rather insufficient.  The JBS ticket is not a normative document, perhaps this should be clarified in a follow-up ticket (by moving some of the verbiage from JDK-8268225 ?)

I am not against adding a boolean for `visible` (or `focusVisible` ?) argument.

@mstr2 what would be a good description for this parameter?

mstr2: What do you think of the following:
```
* @param visible Specifies whether the {@link #focusVisibleProperty() focusVisible} flag will be
*                set on the node that receives focus. Callers must specify {@code true} if this
*                method is called as a result of keyboard navigation, or if the current node
*                visibly indicates focus; in all other cases, callers must specify {@code false}.
```

andy-goryachev-oracle: I like that, thank you!

andy-goryachev-oracle: also updated the JEP:
https://github.com/andy-goryachev-oracle/Test/blob/main/doc/FocusTraversal/FocusTraversal-v3.md

andy-goryachev-oracle: Somewhat tangential question: should we start placing JEPs in `/doc-files/jeps` ?

The use of private repos is ok at the initial stages of discussion, but at some point we are risking the bit rot: private repos can be removed but the mailing list posts remain...

kevinrushforth: > I like that, it clearly explains what to do. thank you!

I do, too. When I read the discussion up to the point Michael proposed this, my thought was that the only way to solve it is with a parameter like this.

> Somewhat tangential question: should we start placing JEPs in /doc-files/jeps ?

There was a thread a while ago about this. Let's restart that thread rather than discussing it in any particular PR.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
kevinrushforth: > I don't indent case labels inside of switch.

Our typical code style is to indent. I don't have a strong preference, but I switched (pun intended) may years ago from not indenting to always indenting because: A) the IDEs (at least most of them) indent by default; B) most of the code in the JDK and JavaFX does.

> I don't think we enforce one way or the other.

No, we don't. Feel free to leave it as you have it.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
hjohn: If this flag is suddenly so important, then why doesn't `requestFocus` have it?

I feel that we've found an internal inconsistency, and that the solution is to then drop this on the user to ensure correctness (the wording "callers must specify" definitely alludes to this).  It's like asking users to pass along some important flag, and if they get it wrong the system breaks.

I think the request focus call should take the most sensible value for focusVisible on programmatic changes, and that FX should not be using this API itself for mouse focus actions.  Now we're just exposing focusVisible in a round-about way (as a boolean flag of some other call) to make it user controllable -- then why not just make it writable?  If I don't like the calculated default (or perhaps always `true`), then I can call `Node#setFocusVisible(false)` after the request focus call.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
hjohn: This is a bit silly.  You have an opening brace, you should be indenting as you would in every other case when an opening brace appears and you break off the line.  So unless there is a **really** good reason to suddenly not do so that has to do with readability, I think this is an really odd stand to take.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
andy-goryachev-oracle: Good question, @hjohn .  Why indeed, @mstr2  ?

mstr2: > If this flag is suddenly so important, then why doesn't `requestFocus` have it?

That's easy: because we didn't deem it necessary. Back when we added the flag, only internal code would be able to set it, and that has worked out fine. Focus traversal using keyboard navigation didn't use `requestFocus()`.

Now we have a public API, and we face the same question again. There is a clear advantage of setting the focusVisible flag as a parameter of `requestFocusTraversal`: we ensure consistency between `focused` and `focusVisible`, such that a node cannot be focusVisible if not focused. Having focusVisible be a writable property would allow applications to have a node visibly indicate focus, or even multiple independent nodes indicating visible focus simultaneously, while yet another node is actually focused. That's not good.

mstr2: How about this for `focusVisible`:
```
    /**
     * Indicates whether this {@code Node} should visibly indicate focus.
     * <p>
     * This flag is set when the node is {@link #focusedProperty() focused} via keyboard navigation,
     * or when the input focus was programmatically moved from another node to this node with
     * {@link #requestFocusTraversal} while the other node visibly indicated focus.
     * <p>
     * The {@code focusVisible} flag can be used by applications that do not want a permanent focus
     * indicator, and opt for a focus indicator that is only visible when it is most helpful to the
     * user. This is not the case when the user clicks on a control (because they would know which
     * control they clicked on), but only when a keyboard-initiated focus traversal moves the input
     * focus to another control.
     */
```

As `requestFocusTraversal` is concerned, I think we can simply this a bit. We don't need users to check whether `focusVisible` is set on the current node; the implementation can do this internally. Here's a revised parameter doc:
```
* @param keyNavigation {@code true} if this method is called as a result of keyboard navigation;
*                      {@code false} otherwise.
```

The implementation would then set `focusVisible` to true if either the `keyNavigation` argument is true, or if the current node's `focusVisible` flag is true.

kevinrushforth: This is better than the previous, since it only asks the caller of this method, typically the Skin of a custom control, to provide the one missing piece of information -- whether or not the method is in response to keyboard navigation -- and not ask the caller to compute the final answer. That's good. The only question I have is: can we do even better? Is there enough information for the implementation of `requestFocusTraversal` to know whether it was called from  keyboard event? If not, then this updated proposal seems reasonable.

hjohn: I think it could be possible, but I'm not sure if it would be pretty.  Scene is the one dispatching these events, and since FX has a single threaded model, Scene could have a flag that indicates it is currently dispatching a key event.  If `requestFocusTraversal` is called, it can detect if this was in response to a key event higher up the call stack by checking the flag on Scene.

I do think there are other mechanisms in FX that pass along information from shallower callstacks to deeper callstacks in some form or another.  `ExpressionHelper` for example is aware of nested calls and takes action accordingly, and I'm almost certain there are more in places were FX is dealing with peers and syncing the scene graph.

andy-goryachev-oracle: What happens when a different type of event is created by a `KeyEvent` handler?  How does the downstream code differentiate between the secondary events coming from a key event handler vs. from one originated normally?

> Is there enough information for the implementation of requestFocusTraversal to know whether it was called from keyboard event?

The (original) intent of this API is to enable the custom components trigger the focus traversal as a response to key events only, the problem is that we can't prevent the application code from misusing the API and calling it for other reason.  (There is no other reason for the application code or custom component to call this method with the `visible=false`, other than misuse)

I would rather specify that this method must be called as a response to KeyEvent, and drop the argument.

kevinrushforth: That seems reasonable, given that it is the expected use case. I wouldn't say "_must_ be called", but rather that it "is _expected_ to be called" in response to a KeyEvent, and then document the behavior accordingly -- namely, the node that receives focus will be `focusVisible`.

@mstr2 What do you think?

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
kevinrushforth: That isn't an example of "everything on one line". That's effectively the "switch case NNNN:, on a line by itself", which is defensible, although still not preferred.

Anyway, what you currently have is what I called option 1, and it doesn't seem very readable. I'll still leave it up to you, but please consider the comments made in this thread.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
hjohn: > What happens when a different type of event is created by a `KeyEvent` handler? How does the downstream code differentiate between the secondary events coming from a key event handler vs. from one originated normally?

This wouldn't matter IMHO, if you fire another event, that just deepens the call stack (the secondary event is fired immediately, not after the key event finishes); the original key event will still be the one that initiated it -- ie. user pressed a key, and focus changed as a result -- no matter how that exactly happened it is then FX's decision to turn on focus visible highlighting.  The whole focus visible highlighting is a feature provided by FX to give it a more native look and feel, but it may be platform dependent which is why I think the user should probably not have a say in the matter. 

I already suggested to make the property writable, but if the user can't be trusted to use it correctly, then they can't be trusted to set such a boolean correctly either.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
nlisker: I'm not sure why this is a counter argument. It's not used in the JDK or JavaFX. I still think option 2 is better, but if we don't enforce a style then it's your choice.

andy-goryachev-oracle: I think this discussion is not productive.  Can we focus on essentials please?

kleopatra: readability __is__ essential ;) Same as others, wondering why you insist on not going to max it?

andy-goryachev-oracle: I agree with readability.  At the same time, every single suggestion is perfectly readable, in my opinion.

The problem is that all this is subjective.  I find the K&R style unreadable, especially when indents are all over the place.  I don't like spaces after if or switch, or inconsistent newlines at the beginning of the class.  And I especially dislike the trailing whitespace rule in jcheck.  But I keep quiet (except this once).

"This is not a productive area of discussion"

andy-goryachev-oracle:
A general question: what would be an example of when `focused != focusVisible` (and `!= focusWithin`) ?

Shouldn't the `focused` be a sort of "primary" property, and all other derived from it?

What is the purpose of `focusVisible` exactly?

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
mstr2: Is there really no conceivable scenario in which this method could legitimately be called _not_ in response to a key event, or is it simply that we're not thinking of such a scenario right now? I'd be bothered by an API that was created with a too-narrow focus in mind, and as a result doesn't allow me to do what I need.

andy-goryachev-oracle: Could you think of a scenario when we want a component focused but not indicate its focused state?

mstr2:
> A general question: what would be an example of when `focused != focusVisible` (and `!= focusWithin`) ?
> 
> Shouldn't the `focused` be a sort of "primary" property, and all other derived from it?
> 
> What is the purpose of `focusVisible` exactly?

`focusVisible` is almost always `false`, because most of the time, users of a GUI will interact with a mouse (or touch), not with the keyboard. Most desktop user interfaces _only_ show a focus ring around interactive controls if you tab to the element, and _not_ if you click it. This is what this property allows JavaFX applications to replicate (in fact, the built-in themes of JavaFX are an outlier in desktop UI design in that they have permanent focus rings).

Naturally, a control can only visibly indicate focus if it has the input focus. This is why `focusVisible=true` can only occur when `focused=true` and `focusWithin=true`.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
mstr2: Yes, that's actually the most common case (focused, but not focusVisible) in contemporary user interfaces, see my post below.

mstr2:
Just to complete the picture, I also want to explain why I am strongly against making `focusVisible` writable.

The (much older) `focused` property is an example of bad API design: it is a read-only property that is controlled by the focus subsystem, but in a stroke of complete hackery, also allows subclasses to change its value arbitrarily with the protected `setFocused` method. From this it necessarily follows that the documentation of `focused` is lying:
```
    /**
     * Indicates whether this {@code Node} currently has the input focus.
     * To have the input focus, a node must be the {@code Scene}'s focus
     * owner, and the scene must be in a {@code Stage} that is visible
     * and active. See {@link #requestFocus()} for more information.
     *
     * @see #requestFocus()
     * @defaultValue false
     */
```
Since `focused` can be set (and will be set) arbitrarily by subclasses, the presence of this flag does _not_ necessarily indicate that the node has the input focus, it does _not_ necessarily indicate that it is the scene's focus owner, and it does _not_ necessarily indicate that it is in an active scene.

You might think that this is only a small loophole in the implementation, but sadly it is not: Several controls, most notably TableView, abuse the `focused` property for an entirely different purpose. They use it to mark the cell that is currently _selected_, and then hack the `focused` property value to reflect the _selection state_. This selection persists even when the cell doesn't have the input focus (and thus, the `focused` property lies about the input focus).

Not only is this unexpected, it regularly leads users to think that the JavaFX focus system is defective: https://bugs.openjdk.org/browse/JDK-8317426

Repeating the same mistake with `focusVisible` would be crazy.

andy-goryachev-oracle:
>  users of a GUI will interact with a mouse (or touch), not with the keyboard.

It's one of the falsehoods programmers should be aware of.

But regardless of how they interact with the UI, the focus indicator should always indicate the focus owner, no?  There might be levels of indication, such as grayed focus border when the component does not have input focus, vs blue focus border when it does, but there should never be a situation when the component **has input focus and no indication**, right?

So perhaps the focus subsystem should have been designed to have these three properties:

- focus-owner
- focus-highlighted
- parent-of-focus-owner

Applying this logic to the issue at hand, we probably should drop the boolean argument and say it "is expected to be called in response to a KeyEvent", that is, do not give the user the choice they don't need (or want).

What do you think?

mstr2:
> It's one of the falsehoods programmers should be aware of.
> 
> But regardless of how they interact with the UI, the focus indicator should always indicate the focus owner, no? There might be levels of indication, such as grayed focus border when the component does not have input focus, vs blue focus border when it does, but there should never be a situation when the component **has input focus and no indication**, right?

No, you're clearly in the wrong here. Just go to a random native macOS or Windows 11 application and show me the focus indicators. You won't see them in almost all cases. You're trying to re-litigate common concensus across operating systems and [on the web](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible), and in the process ignore all precedent and the thinking that lead to the state of the art.


> Applying this logic to the issue at hand, we probably should drop the boolean argument and say it "is expected to be called in response to a KeyEvent", that is, do not give the user the choice they don't need (or want).
> 
> What do you think?

I don't see how your argument against `focusVisible` has anything to do with the question at hand.

andy-goryachev-oracle:
OK, I am just trying to understand.

Getting back to the API, are you ok with dropping the boolean `visible` argument and documenting the expectation of the key event?

hjohn:
> Just to complete the picture, I also want to explain why I am strongly against making `focusVisible` writable.

I don't disagree, I'm just a bit on the fence if making it writable by proxy is a good idea then when it is a concept that should be controlled by FX's implementation.

> The (much older) `focused` property is an example of bad API design: it is a read-only property that is controlled by the focus subsystem, but in a stroke of complete hackery, also allows subclasses to change its value arbitrarily with the protected `setFocused` method. From this it necessarily follows that the documentation of `focused` is lying:
> 
> ```
>     /**
>      * Indicates whether this {@code Node} currently has the input focus.
>      * To have the input focus, a node must be the {@code Scene}'s focus
>      * owner, and the scene must be in a {@code Stage} that is visible
>      * and active. See {@link #requestFocus()} for more information.
>      *
>      * @see #requestFocus()
>      * @defaultValue false
>      */
> ```
> 
> Since `focused` can be set (and will be set) arbitrarily by subclasses, the presence of this flag does _not_ necessarily indicate that the node has the input focus, it does _not_ necessarily indicate that it is the scene's focus owner, and it does _not_ necessarily indicate that it is in an active scene.

Yeah, it's not a very useful property in its current state; I see it as more of an indication that something has the focused look (like disabled, pressed and armed for example) and not necessarily that such a Node actually has the focus.  Better to use focusOwner for that.

> You might think that this is only a small loophole in the implementation, but sadly it is not: Several controls, most notably TableView, abuse the `focused` property for an entirely different purpose. They use it to mark the cell that is currently _selected_, and then hack the `focused` property value to reflect the _selection state_. This selection persists even when the cell doesn't have the input focus (and thus, the `focused` property lies about the input focus).

I think it is sometimes desirable to have a cell look like it has focus, as well as the TableView itself, but I could be wrong, or it should have been implemented differently.  Perhaps the concepts are more like the TableView has the focus, and its cursor is on a specific cell (similar to a simple text box, where the box has the focus, and there's a cursor for navigation/editing -- we don't say that position 43 has the focus either when the cursor is located there...).

As for `focusVisible`, I think a good default would definitely be to set it to true when the programmatic methods are called.  Yes, you could call these methods in response to a mouse click, but I think that would be really silly (ie. the user clicks at location X,Y, and you programmatically change focus to something that wasn't clicked).  Regardless, indicating **where** the focus went, even if it was a mouse click, is probably a good idea.

If we can't think of a good use case now to set it to `false` after programmatic navigation, then I would strongly suggest to wait for such a use case to appear (on the bug tracker), evaluating that, and then potentially adding an overload or some other solution (like the one I suggested that checks if a key event was in the call stack).

Also something that just occurred to me, but perhaps I'm late to the party; with the current API being on `Node` you could call `requestFocusTraversal` on basically any `Node`, even one that is currently not focused.  I'm not sure if this is intentional or that the API should perhaps only work from the current focused node.

ps.  the diff for this PR still looks huge here?

andy-goryachev-oracle:
> ps. the diff for this PR still looks huge here?

Mainly because of TraversalEngine -> TraversalPolicy refactoring and skins.

And yes, one can call requestFocusTraversal on any Node, focusable or not.  There was some language to that effect somewhere, but it should be explicitly described here - you are right pointing this out.

nlisker:
> Also something that just occurred to me, but perhaps I'm late to the party; with the current API being on `Node` you could call `requestFocusTraversal` on basically any `Node`, even one that is currently not focused. I'm not sure if this is intentional or that the API should perhaps only work from the current focused node.

I saw this already in the previous PR and thought it's intentional. Even with this method being outside of `Node`, it takes a `Node` argument, which is arbitrary too. If you want to limit it, you'll have to throw or no-op on an unfocused node. I don't think that's necessary.

andy-goryachev-oracle:
It was in the removed FocusTraversal:

```
 * The methods provided in this class allow for transferring focus away from the
 * specific {@code Node} (which serves as a reference and does not have to be focused or
 * focusable), within the owning {@link Scene} or {@link SubScene}.
```

hjohn:
> > Also something that just occurred to me, but perhaps I'm late to the party; with the current API being on `Node` you could call `requestFocusTraversal` on basically any `Node`, even one that is currently not focused. I'm not sure if this is intentional or that the API should perhaps only work from the current focused node.
> 
> I saw this already in the previous PR and thought it's intentional. Even with this method being outside of `Node`, it takes a `Node` argument, which is arbitrary too. If you want to limit it, you'll have to throw or no-op on an unfocused node. I don't think that's necessary.

Well, the API could be on `Scene`, in which case you just call `Scene#requestFocusTraversal(Direction)`, as `Scene` knows what the focused node is.

nlisker:
> > > Also something that just occurred to me, but perhaps I'm late to the party; with the current API being on `Node` you could call `requestFocusTraversal` on basically any `Node`, even one that is currently not focused. I'm not sure if this is intentional or that the API should perhaps only work from the current focused node.
> > 
> > 
> > I saw this already in the previous PR and thought it's intentional. Even with this method being outside of `Node`, it takes a `Node` argument, which is arbitrary too. If you want to limit it, you'll have to throw or no-op on an unfocused node. I don't think that's necessary.
> 
> Well, the API could be on `Scene`, in which case you just call `Scene#requestFocusTraversal(Direction)`, as `Scene` knows what the focused node is.

In that case, you are right. I don't have a strong opinion as I am not an advanced user of focus. The current behavior is good enough for me.

By the way, if you have multiple scenes in the stage, and each one has its own focused node, which node receives keyboard input?

hjohn:
> In that case, you are right. I don't have a strong opinion as I am not an advanced user of focus. The current behavior is good enough for me.
> 
> By the way, if you have multiple scenes in the stage, and each one has its own focused node, which node receives keyboard input?

Keyboard input goes to the active window, and so it would go to the scene attached to that window (and a window can only have one scene, not counting subscenes).  The node receiving the input then would be focus owner of the scene that belongs to the current active window.

mstr2:
> Getting back to the API, are you ok with dropping the boolean `visible` argument and documenting the expectation of the key event?

Yes, I think that's okay.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -10440,29 +10440,40 @@ void releaseAccessible() {
...
+     * The {@code Node} serves as a reference point and does not have to be focused or focusable.
      * A successful traversal results in a new {@code Node} being focused.
      * <p>
-     * The {@code visible} parameter specifies whether the {@link #focusVisibleProperty() focusVisible}
-     * flag will be set on the node that receives focus.
-     * Callers must specify {@code true} if this method is called as a result of keyboard navigation,
-     * or if the current node visibly indicates focus;
-     * in all other cases, callers must specify {@code false}.
+     * This method is expected to be called in response to a {@code KeyEvent}, since the {@code Node}
+     * receiving focus will have {@link #focusVisibleProperty() focusVisible} property set.
```
mstr2, line 10447: I think there should be the definite article "the" in front of `{@link #focusVisibleProperty() focusVisible}`.

andy-goryachev-oracle: you are right.  fixed.

andy-goryachev-oracle:
Closing in favor of #1628.

PR #1607, opened 2024-10-21
8342703: CSS transition is not started when initial value was not specified

mstr2:
When the initial value of a styleable property is not specified in a stylesheet, no transition is started:

    .button {
        transition: -fx-opacity 1s;
    }
    
    .button:hover {
        -fx-opacity: 0.5;
    }

The expected behavior is that a transition is started in this case, since the default value of `-fx-opacity` is 1.

The reason for this bug is that `StyleableProperty` implementations do not start a CSS transition when the value is applied for the first time. The intention behind this is that a node that is added to the scene graph should not start transitions. CSS transitions should only be started _after_ the node has been shown for the first time.

The logic to detect this situation is currently as follows:
```java
    // If this.origin == null, we're setting the value for the first time.
    // No transition should be started in this case.
    TransitionDefinition transition = this.origin != null && getBean() instanceof Node node ?
        NodeHelper.findTransitionDefinition(node, getCssMetaData()) : null;
```

However, this does not work. When no initial style is specified in the stylesheet, `this.origin` will not be set, and thus no transition will be started even after the node has been shown. The new logic works like this:

A `Node.initialCssState` flag is added. Initially, this is `true`. Manually calling `applyCss` or similar methods will not clear this flag, as we consider all manual CSS processing to be part of the "initial CSS state". Only at the end of `Scene.doCSSPass` will this flag be cleared on all nodes that have expressed their interest. This mechanism ensures that a node will be eligible for CSS transitions only after the following conditions have been satisfied:
1. The node was added to a scene graph
2. CSS processing was completed for a scene pulse

/reviewers 2

---------
### Progress

### Issue
 * [JDK-8342703](https://bugs.openjdk.org/browse/JDK-8342703): CSS transition is not started when initial value was not specified (**Bug** - P4)

### Reviewers
 * [Markus Mack](https://openjdk.org/census#mmack) (@drmarmac - Author)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1607/head:pull/1607` \
`$ git checkout pull/1607`

`$ git checkout pull/1607` \
`$ git pull https://git.openjdk.org/jfx.git pull/1607/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1607`

`$ git pr show -t 1607`

</details>

<a href="https://git.openjdk.org/jfx/pull/1607.diff">https://git.openjdk.org/jfx/pull/1607.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1607#issuecomment-2427037403)
</details>


modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -10066,6 +10079,25 @@ private void doProcessCSS() {
         }
     }
 
+    /**
+     * A node is considered to be in its initial CSS state if it wasn't shown in a scene graph before.
+     * This flag is cleared after CSS processing was completed in a Scene pulse event. Note that manual
+     * calls to {@link #applyCss()} or similar methods will not clear this flag, since we consider all
+     * CSS processing before the Scene pulse to be part of the node's initial state.
+     */
+    private boolean initialCssState = true;
```
andy-goryachev-oracle, line 10088: is it possible to find a solution that does not involve adding a field to Node?

mstr2: I don't think so, because we need to track the "initial CSS state" flag for each individual node. It can change during runtime when the node is added to or removed from the scene graph.

I've looked at adding it to `CssFlags` or `DirtyBits`, but both have different purposes. The addition of this boolean field doesn't change the object size for most HotSpot implementations, as boolean fields are usually packed together. Currently, `Node` has 9 boolean fields (plus 1 byte field), with an extra 2 bytes of unused space.

modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
```diff
@@ -589,6 +589,29 @@ private void doCSSPass() {
...
+            clearInitialCssStateNodes.clear();
+        }
+    }
+
+    /**
+     * A list of nodes that have expressed their interest to be notified when the next CSS pass
+     * is completed. Nodes will use this event to determine whether they are in their initial
+     * CSS state (see {@link Node#initialCssState}.
+     */
+    private final Set<Node> clearInitialCssStateNodes = Collections.newSetFromMap(new IdentityHashMap<>());
```
Maran23, line 607: why not use a normal `HashSet`? What advantage has this approach?

mstr2: `HashSet` uses `Object.equals`, which can be overridden by user code, and this would break the logic. It's the _instance_ that wants to be notified.

Maran23: Good point, I never ever did that for `Node`s (and I don't know why I would need to), but you are right, it is indeed possible and therefore a possible scenario.

hjohn: I think this is overly cautious.  Overriding equals on a class that did not implement equals in a hierarchy you don't control  is not a very reasonable scenario.  You will not be able to call `super.equals` with reasonable results if the hierarchy did not implement it in the first place.  This is because there may be private data that you can't access for your equality comparison (and `Node` has lots of that).  We also use `WeakHashMap`s in several areas already with subtypes of `Node`s as keys (I found `Region`, `Parent` and `TreeView` being used as keys), so this kind of override should probably be documented as being unsupported on `Node`.

mstr2: Makes sense. I think we should make `Node.equals()` final in this case.

drmarmac (approved):
I did some testing, adding transitions to css properties that have no explicit initial values in css. The issue is clearly reproducible with the current master, and fixed with this PR.

One thing I noticed is that transitions of "-fx-background-insets" jump without a smooth animation. Probably this is due to a difference in the number of list elements (-fx-background-insets is an array), which does not seem to have animation support (yet), so it's not an issue with this PR.

Code changes look good as well, I see the addition to `Node` has already been discussed.

modules/javafx.graphics/src/main/java/javafx/css/StyleableBooleanProperty.java
```diff
@@ -70,7 +70,8 @@ public StyleableBooleanProperty(boolean initialValue) {
     /** {@inheritDoc} */
     @Override
     public void applyStyle(StyleOrigin origin, Boolean v) {
-        TransitionDefinition transition = this.origin != null && getBean() instanceof Node node ?
+        // If the value is applied for the first time, we don't start a transition.
+        TransitionDefinition transition = getBean() instanceof Node node && !NodeHelper.isInitialCssState(node) ?
```
andy-goryachev-oracle, line 74: minor: to simplify the code in many places, you could change `isInitialState` to accept an `Object` argument instead of `Node` and do an `instanceof` there.

mstr2: I don't think this would work, since we need the `Node`-typed `node` variable as an argument for `NodeHelper.findTransitionDefinition`, so we can't remove the `instanceof` test here.

andy-goryachev-oracle: NodeHelper:

```
    public static boolean isInitialCssState(Object node) {
        if(node instanceof Node n) {
            return nodeAccessor.isInitialCssState(n);
        }
        return false;
    }
```

andy-goryachev-oracle: or better yet, move the instanceof to NodeHelper.findTransitionDefinition()

mstr2: I'll see whether I can refactor these methods a bit, but not as part of this PR.

PR #1608, opened 2024-10-23
8342453: Remove calls to doPrivileged in javafx.graphics/com.sun.javafx.tk

kevinrushforth:
This PR removes all doPrivileged calls from `com.sun.javafx.tk**` in the `javafx.graphics` module. 

Here is a quick overview of what I did for this fix:

1. Changed all simple cases of `doPrivileged((PrivilegedAction<T>) () -> LAMBDA)` to `LAMBDA`, removing the `@SuppressWarnings("removal")` if possible. In case of an unused or unneeded local variable, I removed the local variable.
2. Remove unused `AccessControlContext` variables, meaning those whose only use was to be passed into a doPrivileged call that is now gone.
3. Removed all `@SuppressWarnings("removal")` annotations that were unneeded after the removed doPrivileged calls. In some cases there are annotations on a method or class. Those can only be removed if the doPrivileged calls were the only use of deprecated SM API in that method or class.
4. Remove unused imports.

Finally, here are a few "best practices" that I tried to follow, and would ask others to follow when doing their piece of this. The idea is to reduce the cognitive load on the reviewers. It might take you a couple extra minutes, but will save time during the review:

* When removing the doPrivileged calls, please do the minimum amount of reformatting necessary to properly indent the body of the doPrivileged after it is removed. For example, don't wholesale reformat a method (or worse, an entire class) just because a couple doPrivileged calls are removed.
* Please try to not reorder the import statements when removing unused imports.

#### Notes to reviewers

As a helpful hint for reviewers, I recommend reviewing this using the "Hide whitespace" option.

An initial, limited review of this was done in my personal fork at kevinrushforth/jfx#4 if other reviwers are interested.

---------
### Progress

### Issue
 * [JDK-8342453](https://bugs.openjdk.org/browse/JDK-8342453): Remove calls to doPrivileged in javafx.graphics/com.sun.javafx.tk (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1608/head:pull/1608` \
`$ git checkout pull/1608`

`$ git checkout pull/1608` \
`$ git pull https://git.openjdk.org/jfx.git pull/1608/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1608`

`$ git pr show -t 1608`

</details>

<a href="https://git.openjdk.org/jfx/pull/1608.diff">https://git.openjdk.org/jfx/pull/1608.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1608#issuecomment-2441524385)
</details>


kevinrushforth:

@jayathirthrao @lukostyra you will also want to look at this if you didn't get a chance to look at kevinrushforth/jfx#4

/reviewers 2

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
```diff
@@ -136,70 +134,58 @@
...
-    @SuppressWarnings("removal")
-    private static boolean debug =
-            AccessController.doPrivileged((PrivilegedAction<Boolean>) () -> Boolean.getBoolean("quantum.debug"));
+    public static final boolean pulseDebug = Boolean.getBoolean("quantum.pulse");
 
-    @SuppressWarnings("removal")
-    private static Integer pulseHZ =
-            AccessController.doPrivileged((PrivilegedAction<Integer>) () -> Integer.getInteger("javafx.animation.pulse"));
+    private static final boolean multithreaded = initMultithreaded();
+    private static boolean initMultithreaded() {
```
mstr2, line 142: Have you considered a pattern such as this?

```java
    private static final boolean multithreaded = ((Supplier<Boolean>)() -> {
        // If it is not specified, or it is true, then it should
        // be true. Otherwise it should be false.
        String value = System.getProperty("quantum.multithreaded");
        if (value == null) return true;
        final boolean result = Boolean.parseBoolean(value);
        if (verbose) {
            System.out.println(result ? "Multi-Threading Enabled" : "Multi-Threading Disabled");
        }
        return result;
    }).get();
```

This would prevent a proliferation of many new init methods.

kevinrushforth: That's an interesting suggestion: trading a new method for a lambda (the prior code used a lambda to pass to the doPrivileged, so it's also less of a delta). I'll take a look at it and see if it looks better to me.

kevinrushforth: I decided to make this change. Thanks for the suggestion.

andy-goryachev-oracle: This looks useful.
I wonder if it would make sense to create a utility method in some common Util class?

```
public static boolean initBoolean(BooleanSupplier c) { return c.getAsBoolean(); }
```

mstr2 (approved):

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
kevinrushforth: We could, but I don't see enough value in doing it. The only thing you would save is the explicit cast to `(Supplier<Boolean>)` and the `.get()`.

Rather than that, it might be useful  (as a follow-up, not as part of the various doPriv removal PRs), to have a centralized utility method that also did the reading and parsing of the property. I'm fairly certain we already have some localized methods of this sort somewhere.

In any case, I don't want to make a refactoring change that relies on a new utility method as part of this effort

andy-goryachev-oracle: ok

andy-goryachev-oracle (approved):

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassViewEventHandler.java
```diff
@@ -60,22 +60,17 @@
...
 
-import java.security.AccessController;
-import java.security.PrivilegedAction;
+import java.util.function.Supplier;
 
 class GlassViewEventHandler extends View.EventHandler {
 
     static boolean zoomGestureEnabled;
     static boolean rotateGestureEnabled;
     static boolean scrollGestureEnabled;
```
Glavo, line 69:
```java
    private static final boolean zoomGestureEnabled = Boolean.valueOf(System.getProperty("com.sun.javafx.gestures.zoom", "false"));
    private static final boolean rotateGestureEnabled = Boolean.valueOf(System.getProperty("com.sun.javafx.gestures.rotate", "false"));
    private static final boolean scrollGestureEnabled = Boolean.valueOf(System.getProperty("com.sun.javafx.gestures.scroll", "false"));
```

kevinrushforth: Yes, I could have done that, but chose to minimize the diffs in this case. Now that it is integrated, it's probably not worth revisiting.

PR #1618, opened 2024-10-31
8343336: Add persistentScrollBars preference

mstr2:
This PR adds another accessibility preference that is available on all supported desktop platforms:

1. **Windows**: Settings -> Accessibility -> Visual Effects -> Always show scroll bars
2. **macOS**: System Settings -> Appearance -> Show scroll bars
3. **Ubuntu**: Settings -> Accessibility -> Seeing -> Always show scroll bars

When the OS setting is changed, all platforms report updates that can be observed in JavaFX.
The manual test application `test/manual/events/PlatformPreferencesChangedTest` can be used to test the feature.

/reviewers 2
/csr

---------
### Progress
- [x] Change requires CSR request [JDK-8343451](https://bugs.openjdk.org/browse/JDK-8343451) to be approved

### Issues
 * [JDK-8343336](https://bugs.openjdk.org/browse/JDK-8343336): Add persistentScrollBars preference (**Enhancement** - P4)
 * [JDK-8343451](https://bugs.openjdk.org/browse/JDK-8343451): Add persistentScrollBars preference (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1618/head:pull/1618` \
`$ git checkout pull/1618`

`$ git checkout pull/1618` \
`$ git pull https://git.openjdk.org/jfx.git pull/1618/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1618`

`$ git pr show -t 1618`

</details>

<a href="https://git.openjdk.org/jfx/pull/1618.diff">https://git.openjdk.org/jfx/pull/1618.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1618#issuecomment-2448969624)
</details>


andy-goryachev-oracle:
This looks like a useful data point.

I am curious as to what are the criteria for including or excluding a particular preference?  What is our long term goal - should we support all the preferences available on a particular platform, or the intersection, or what?  May be it's something to discuss first on the mailing list.

mstr2:
My thinking is that a particular preference should be useful and almost universally supported on the desktop platforms to make it into JavaFX API. The number of preferences we might consider are pretty limited, so I don't expect `Platform.Preferences` to explode in size and scope.

I think this is a minor addition that doesn't change anything substantial about `Platform.Preferences`, and as such doesn't need a major discussion since the arguments are basically the same since the introduction of platform preferences. However, if you or others are interested in having a discussion, we can of course do that.

mstr2:
Changed the naming from `reducedScrollBars` to `persistentScrollBars`, as this matches the logic of the OS settings more closely.

kevinrushforth:
> My thinking is that a particular preference should be useful and almost universally supported on the desktop platforms to make it into JavaFX API.

This seems like the right criteria to me.

> The number of preferences we might consider are pretty limited, so I don't expect `Platform.Preferences` to explode in size and scope.

Good.

> I think this is a minor addition that doesn't change anything substantial about `Platform.Preferences`, and as such doesn't need a major discussion since the arguments are basically the same since the introduction of platform preferences. However, if you or others are interested in having a discussion, we can of course do that.

With this explanation, I agree.

kevinrushforth:
Reviewers: @andy-goryachev-oracle @lukostyra 

(I'll look over the public API and the CSR)

modules/javafx.graphics/src/main/java/javafx/application/Platform.java
```diff
@@ -586,6 +586,20 @@ public static Preferences getPreferences() {
...
+         * Specifies whether applications should always show scroll bars. If not set, an application may
+         * choose to hide scroll bars that are not actively used, or make them smaller or less noticeable.
+         * <p>
+         * If the platform does not report this preference, this property defaults to {@code false}.
+         *
+         * @return the {@code persistentScrollBars} property
+         * @defaultValue {@code false}
+         * @since 24
+         */
+        ReadOnlyBooleanProperty persistentScrollBarsProperty();
```
kevinrushforth, line 599: Do you also plan to add new platform-specific preferences or do you think this is unnecessary?

mstr2: I added the respective platform-specific keys to the `Preferences` documentation.

andy-goryachev-oracle (commented):
the code changes look good, and both the property and the key reflect the platform preference on macOS M1 14.7.

I'll test windows tomorrow, and if someone can test it on linux, that would be fantastic!

andy-goryachev-oracle (approved):
looks good on win11, both property and windows key get updated in the monkey tester.

kevinrushforth (approved):
Looks good. In addition to reviewing the code and the docs, I tested it on all platforms. On Linux I tested it on Ubuntu 24.04  and on a _very_ old 16.04. There is a minor bug on 16.04 (a warning printed), but it could be done as a follow-up if you prefer.

I reviewed the CSR, so it is ready to finalize.

modules/javafx.graphics/src/main/native-glass/gtk/PlatformSupport.cpp
```diff
@@ -121,6 +121,10 @@ jobject PlatformSupport::collectPreferences() const {
         gboolean enableAnimations = true;
         g_object_get(settings, "gtk-enable-animations", &enableAnimations, NULL);
         putBoolean(env, prefs, "GTK.enable_animations", enableAnimations);
+
+        gboolean overlayScrolling = true;
+        g_object_get(settings, "gtk-overlay-scrolling", &overlayScrolling, NULL);
```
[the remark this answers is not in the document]
mstr2: I've added a runtime check for the `gtk-overlay-scrolling` property.

modules/javafx.graphics/src/main/native-glass/gtk/PlatformSupport.cpp
```diff
@@ -121,6 +121,12 @@ jobject PlatformSupport::collectPreferences() const {
         gboolean enableAnimations = true;
         g_object_get(settings, "gtk-enable-animations", &enableAnimations, NULL);
         putBoolean(env, prefs, "GTK.enable_animations", enableAnimations);
+
+        if (g_object_class_find_property(G_OBJECT_GET_CLASS(settings), "gtk-overlay-scrolling")) {
```
kevinrushforth, line 125: Looks good, presuming that we want the property to be absent on systems that don't support this setting.

kevinrushforth (approved):
The warning is gone on my Ubuntu 16.04 system. I note that the `GTK.overlay_scrolling` preference is not reported on that system (which seems fine) and that the `persistentScrollBarsProperty` is always `false`. Everything still works as expected on systems with a newer GTK.

modules/javafx.graphics/src/main/native-glass/gtk/PlatformSupport.cpp
mstr2: Yes, a property that isn't available shouldn't be included in the preferences map. We've documented that in `Platform.Preferences`:
```java
     * The preferences that are reported by the platform may be dependent on the operating system version
     * and its current configuration, so applications should not assume that a particular preference is
     * always available.
```

kevinrushforth: Yes, so this is fine.

andy-goryachev-oracle (approved):
ship it!

PR #1627, opened 2024-11-05
8342992: Security manager check should not use deprecated methods

kevinrushforth:
As specified in the security manager removal CSR, [JDK-8341858](https://bugs.openjdk.org/browse/JDK-8341858), JavaFX now checks at startup whether the security manager is enabled and fails fast with an `UnsupportedOperationException` if it is. The check is currently done in the `<clinit>` methods of `PlatformImpl` and `LauncherImpl` by calling the deprecated `System::getSecurityManager` method.

This PR creates a new `SecurityUtility::checkSecurityManager` utility method in `javafx.base` that uses reflection to avoid calling API that is deprecated for removal. I also added a call to `checkSecurityManager` in `ReflectUtil` and `MethodUtil` in `javafx.base` for non-graphical applications that only use `javafx.base`.

---------
### Progress

### Issue
 * [JDK-8342992](https://bugs.openjdk.org/browse/JDK-8342992): Security manager check should not use deprecated methods (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1627/head:pull/1627` \
`$ git checkout pull/1627`

`$ git checkout pull/1627` \
`$ git pull https://git.openjdk.org/jfx.git pull/1627/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1627`

`$ git pr show -t 1627`

</details>

<a href="https://git.openjdk.org/jfx/pull/1627.diff">https://git.openjdk.org/jfx/pull/1627.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1627#issuecomment-2458028151)
</details>

### DIFF
--- a/modules/javafx.base/src/main/java/com/sun/javafx/SecurityUtil.java
+++ b/modules/javafx.base/src/main/java/com/sun/javafx/SecurityUtil.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright (c) 2024, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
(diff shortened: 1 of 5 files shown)


kevinrushforth:
Reviewer: @arapte

modules/javafx.base/src/main/java/com/sun/javafx/SecurityUtil.java
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Check for the presence of a security manager (from an older JDK) and
+     * throw UnsupportedOperationException if enabled. Use reflection to avoid
+     * a dependency on an API that is deprecated for removal. This method does
+     * nothing if the security manager is not enabled or if
+     * System::getSecurityManager cannot be invoked.
+     *
+     * @throws UnsupportedOperationException if the security manager is enabled
+     */
+    public static void checkSecurityManager() {
```
Glavo, line 48: Can we cache the results of the check?

kevinrushforth: I thought about doing that. In fact, I even coded it up, but since it only called at most 4 times (once for each of the 4 call sites) at startup, I wasn't sure it was worth it. I don't mind either way.

@arapte do you have an opinion?

arapte: Given that usage of this method is very limited and it would not change in future. It seems Ok not to cache the result.

andy-goryachev-oracle (approved):

PR #1628, opened 2024-11-06
8091673: Public focus traversal API for use in custom controls

andy-goryachev-oracle:
Public focus traversal API for use in custom controls.

https://github.com/andy-goryachev-oracle/Test/blob/main/doc/FocusTraversal/FocusTraversal-v3.md

This is a lightweight change that only adds the public API for focus traversal, containing neither the public API for the traversal policy (#1555) nor with the changes for the traversal policy hidden (#1604).

/csr
/reviewers 2

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -8531,6 +8531,45 @@ public abstract class Node implements EventTarget, Styleable {
         return getScene().traverse(this, dir, method);
     }
 
+    /**
+     * Tries to move the focus from this {@code Node} in the specified direction.
+     * The {@code Node} serves as a reference point and does not have to be focused or focusable.
+     * A successful traversal results in a new {@code Node} being focused.
+     * <p>
+     * This method is expected to be called in response to a {@code KeyEvent}, since the {@code Node}
+     * receiving focus will have the {@link #focusVisibleProperty() focusVisible} property set.
+     *
+     * @param direction the direction of focus traversal
+     * @return {@code true} if traversal was successful
+     * @since 24
+     */
+    public final boolean requestFocusTraversal(TraversalDirection direction) {
+        Direction d;
+        switch (direction) {
+        case DOWN:
+            d = Direction.DOWN;
+            break;
+        case LEFT:
+            d = Direction.LEFT;
+            break;
+        case NEXT:
+            d = Direction.NEXT;
+            break;
+        case PREVIOUS:
+            d = Direction.PREVIOUS;
+            break;
+        case RIGHT:
+            d = Direction.RIGHT;
+            break;
+        case UP:
+            d = Direction.UP;
+            break;
+        default:
+            return false;
+        }
+        return traverse(d, TraversalMethod.KEY);
+    }
+
     //--------------------------
     //  Private Implementation
     //--------------------------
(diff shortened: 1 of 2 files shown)


modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -8531,6 +8531,45 @@ final boolean traverse(Direction dir, TraversalMethod method) {
         return getScene().traverse(this, dir, method);
     }
 
+    /**
+     * Tries to move the focus from this {@code Node} in the specified direction.
+     * The {@code Node} serves as a reference point and does not have to be focused or focusable.
+     * A successful traversal results in a new {@code Node} being focused.
+     * <p>
+     * This method is expected to be called in response to a {@code KeyEvent}, since the {@code Node}
```
kevinrushforth, line 8539: Suggestion: `, since the Node` --> `; therefore, the Node`

kevinrushforth (commented):
I like this approach of providing the minimal API to do programmatic focus traversal without any additional (and unneeded) implementation changes.

The API looks good with a couple wording suggestions on the docs.

Can you provide some unit tests?

kevinrushforth:
One comment on the JEP:

The [Risks and Assumptions](https://github.com/andy-goryachev-oracle/Test/blob/main/doc/FocusTraversal/FocusTraversal-v3.md#risks-and-assumptions) section no longer matches the current proposal, since you are not proposing to "make the existing internal focus traversal methods public" nor is there any refactoring to introduce risk.

andy-goryachev-oracle:
@nlisker or @mstr2 would you like to be the second reviewer?

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -8531,6 +8531,45 @@ final boolean traverse(Direction dir, TraversalMethod method) {
...
+     * receiving focus will have the {@link #focusVisibleProperty() focusVisible} property set.
+     *
+     * @param direction the direction of focus traversal
+     * @return {@code true} if traversal was successful
+     * @since 24
+     */
+    public final boolean requestFocusTraversal(TraversalDirection direction) {
+        Direction d;
+        switch (direction) {
+        case DOWN:
```
mstr2, line 8549: Have you considered moving this code to the `Direction` enum, for example like this:
```java
enum Direction {
    ...
    public static Direction of(TraversalDirection direction) {
        return switch (direction) {
            case DOWN -> DOWN;
            case LEFT -> LEFT;
            case NEXT -> NEXT;
            case PREVIOUS -> PREVIOUS;
            case RIGHT -> RIGHT;
            case UP -> UP;
        }
    }
}
```

This way we won't clutter up the already very long `Node` file with simple conversions, and the co-location of the conversion with the actual enum makes it easier to maintain the code as well.

mstr2:
I think the tests should also assert the `focusVisible` flag, as it is part of the specification of the new method.

andy-goryachev-oracle:
> tests should also assert the `focusVisible` flag

good point, thanks!

mstr2 (approved):

andy-goryachev-oracle:
@hjohn do you want to take a look as well?

kevinrushforth (approved):
Looks good.

hjohn (approved):
Looks good, a lot easier to review this way, thanks!

PR #1632, opened 2024-11-09
8343956: Focus delegation API

mstr2:
Implementation of [focus delegation](https://gist.github.com/mstr2/44d94f0bd5b5c030e26a47103063aa29).

---------
### Progress
- [ ] Change requires a CSR request matching fixVersion jfx26 to be approved (needs to be created)

### Issue
 * [JDK-8343956](https://bugs.openjdk.org/browse/JDK-8343956): Focus delegation API (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1632/head:pull/1632` \
`$ git checkout pull/1632`

`$ git checkout pull/1632` \
`$ git pull https://git.openjdk.org/jfx.git pull/1632/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1632`

`$ git pr show -t 1632`

</details>

<a href="https://git.openjdk.org/jfx/pull/1632.diff">https://git.openjdk.org/jfx/pull/1632.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1632#issuecomment-3043474437)
</details>


beldenfox:
PopupWindows have to solve a similar problem; multiple nodes can receive key events and they all act as if they have focus. This gets (mildly) complicated when dealing with input methods. There may be multiple focused nodes but there can only be one caret so you have to have a mechanism to determine which node will respond to input method requests. You also have to determine when to enable and disable IM events at the OS level. Spinners have never dealt with input methods but ComboBoxes do.

I've submitted PR #1634 to get input methods working in popups. If you want input methods to work with focus delegation you would need to do similar bookkeeping.

I noticed that PopupWindows use EventRedirectors for forwarding messages. I didn't dig into the EventRedirector implementation but it is an existing class that seems to be tackling the same problem as this PR.

hjohn:
This looks really good.  I'm wondering if this could be simplified further.  Specifically, I think the `hoistFocus` flag and manual management of the focus delegate may not be needed.

It seems to me that a Control could share some similarities with a Scene, in that Control has properties that track a focus owner (similar to focus delegate). In effect, a Control is a focus root similar to scene.  When a Scene receives focus, it determines the best Node to "delegate" focus to; similarly, when a Control receives focus, it determines which skin control should be focused.  The normal focus rules should do the right thing here and for example select the TextField of a Spinner as the delegate automatically (some children may need to be marked as not focusable to guide the auto selection, but this is an already existing standard mechanism).

When determining where to send events, if the target is a focus root, it queries its focus owner (or focus delegate) and extends the event to that target.  If that target is also a focus root, the process repeats.

The request focus function should operate differently as well.  It should look for the closest focus root (a Control or Scene) and call the appropriate request focus function on the root it finds.  If that root is Scene, everything works as usual.  If it is another focus root like a Control, Control can determine the best way to focus one of its child nodes (likely you can just apply a normal search for an eligible focusable control for this).

Perhaps the focus root functionality can be captured in an interface that both Scene and Control implement.  I think it would need to specify a `requestFocus` method and `focusOwnerProperty`.  This interface would then replace the `focusScope` flag.

mstr2:
@hjohn I think what's missing in your model is the option to have an independently focusable node inside of a control. For example, think of a Button placed within a Button (via its `graphic` property). Clicking the nested button should not transfer the focus to its parent button. Whether a focus request should be hoisted is a choice that depends on the specifics of the control, and we probably need an API for that.

beldenfox:
@hjohn Could you provide an outline of the algorithm that a control would use to automatically determine the focus delegate? I really can't envision what that would look like. For that matter I'm not sure why we would rely on an algorithm when in the few existing cases where focus delegation is needed the control knows exactly which node to delegate to.

hjohn:
> @hjohn Could you provide an outline of the algorithm that a control would use to automatically determine the focus delegate? I really can't envision what that would look like. For that matter I'm not sure why we would rely on an algorithm when in the few existing cases where focus delegation is needed the control knows exactly which node to delegate to.

Well, `Scene` determines the initial focus node by doing a depth first search on its children and focusing the first child that is focus traversable and enabled.  I don't see why this wouldn't work for controls exactly the same way.  It's almost always the correct default, and in cases when it isn't, it could still be overridden manually.  Take a ComboBox for example.  It consists of a field and a button (or vice versa if you have a different skin).  In either case however it would focus the field as the button in the combo is not focus traversable.

Skins therefore can control the focus delegate by making appropriate use of focus traversable and enabled, just like you can control what Scene would pick for its initial focus.

hjohn:
> @hjohn I think what's missing in your model is the option to have an independently focusable node inside of a control. For example, think of a Button placed within a Button (via its `graphic` property). Clicking the nested button should not transfer the focus to its parent button. Whether a focus request should be hoisted is a choice that depends on the specifics of the control, and we probably need an API for that.

Okay, agreed, so it would be good to have Nodes control whether or not the focus traverses up.  

I could imagine the flag could also in the future perhaps allow the Root Node or Scene to determine whether the Window should receive focus (perhaps special cases like popups or transparent overlays would allow interaction without hoisting focus to the Window?)

Then what about the focus delegate?  This is not a full FX property in this API, but seems very similar to `focusOwner` in Scene, in that it determines to which `Node` (keyboard) events are directed.  One could say that, since Window/Scene are also focusable, Scene is **delegating** focus to a node.  Would it be reasonable to have `focusDelegate` be a read-only property like in `Scene`, and perhaps rename it to focus owner (or if that's too drastic, ensure that `Scene` at some later stage --might-- implement the same interface and then duplicating its focus owner property as focus delegate)?

This can be achieved later, just interested in hearing your thoughts:
- A getter can later become a property
- A getter can later become part of an interface
- Scene can later implement an interface, even if it means duplicating the functionality of a property (if not named the same already)

I see some overlap, and having Nodes/Scene/Window implement a common interface (like `EventTarget`) is not unheard of.

mstr2:
> > @hjohn I think what's missing in your model is the option to have an independently focusable node inside of a control. For example, think of a Button placed within a Button (via its `graphic` property). Clicking the nested button should not transfer the focus to its parent button. Whether a focus request should be hoisted is a choice that depends on the specifics of the control, and we probably need an API for that.
> 
> Okay, agreed, so it would be good to have Nodes control whether or not the focus traverses up.
> 
> I could imagine the flag could also in the future perhaps allow the Root Node or Scene to determine whether the Window should receive focus (perhaps special cases like popups or transparent overlays would allow interaction without hoisting focus to the Window?)

Do you mean window focus in a JavaFX sense, or in a window manager sense? For example, in Windows you can make a window non-activatable, which means it won't steal the focus from other windows when interacted with (for example, an on-screen keyboard window).


> Then what about the focus delegate? This is not a full FX property in this API, but seems very similar to `focusOwner` in Scene, in that it determines to which `Node` (keyboard) events are directed. One could say that, since Window/Scene are also focusable, Scene is **delegating** events/focus to a node. Would it be reasonable to have `focusDelegate` be a read-only property like in `Scene`, and perhaps rename it to focus owner (or if that's too drastic, ensure that `Scene` at some later stage --might-- implement the same interface and then duplicating its focus owner property as focus delegate)?

It works quite well to say that a `Scene` delegates focus to a `Node`. It doesn't seem to work quite as well (linguistically) to say that a `TextField` is the focus owner of a `ComboBox` (the `ComboBox` is also focused, why would the `TextField` be its focus owner?). But linguistics aside, the question is whether `focusOwner`/`focusDelegate` should be a read-only property.

That's a very good question. It would allow observers to know when the focus delegate changes. But can the focus delegate change? It can probably go from non-null to null, and vice versa. Let's consider a hypothetical control that is like a combo box, but it contains two separate text fields. It seems reasonable that the focus delegate can probably change between the two text fields.

But the proposed API doesn't seem to easily support this scenario. When I click on the second text field, it hoists the focus request up to the combo box, which in turn delegates its focus to... the first or the second text field? How would it choose? Is this a scenario that we need to solve?

beldenfox:
> It works quite well to say that a `Scene` delegates focus to a `Node`. It doesn't seem to work quite as well (linguistically) to say that a `TextField` is the focus owner of a `ComboBox` (the `ComboBox` is also focused, why would the `TextField` be its focus owner?).

For what it's worth I am not a fan of using the term focus owner to refer to a focus delegate. The focusOwner of a Scene is the target of key events. The focus delegate of a ComboBox is the *second* target of key events. Key events are sent to the ComboBox first and the delegate only receives the events that it doesn't consume. To me that's an important distinction.

But really I just don't want to end up having a conversation about the focusOwner's focusOwner.

> But the proposed API doesn't seem to easily support this scenario. When I click on the second text field, it hoists the focus request up to the combo box, which in turn delegates its focus to... the first or the second text field? How would it choose? Is this a scenario that we need to solve?

Presumably we would need a way of telling the focus scope node which sub-node hoisted the focus so it could select the correct delegate. But it's difficult to imagine a control that's trying to pass itself off as a monolithic entity having two internal TextFields. Wouldn't that require enabling keyboard traversal inside a monolithic control? What does that mean? I don't think this is a scenario we need to solve.

hjohn:
> > It works quite well to say that a `Scene` delegates focus to a `Node`. It doesn't seem to work quite as well (linguistically) to say that a `TextField` is the focus owner of a `ComboBox` (the `ComboBox` is also focused, why would the `TextField` be its focus owner?).
> 
> For what it's worth I am not a fan of using the term focus owner to refer to a focus delegate. The focusOwner of a Scene is the target of key events. The focus delegate of a ComboBox is the _second_ target of key events. Key events are sent to the ComboBox first and the delegate only receives the events that it doesn't consume. To me that's an important distinction.

Fair enough, just be aware that Scene also consumes some key events before sending them on, so the analogy holds quite well (menu shortcuts take precedence and are not dispatched, while navigation keys are only processed by Scene when they are unconsumed after bubbling back up).

I just too often see solutions that form a hierarchy, but then name give a part of the hierarchy a custom name, with custom API, while they're really the same thing.  If focusOwner is not a good name (which is quite possible as it was chosen primarily for the case of Scene delegating to Nodes) then keeping focus delegate is fine -- Scene could always be extended to duplicate this property under that name if it makes sense to do so.

> But really I just don't want to end up having a conversation about the focusOwner's focusOwner.

That may happen anyway for more complicated controls, or perhaps with two-level navigation :)
 
> > But the proposed API doesn't seem to easily support this scenario. When I click on the second text field, it hoists the focus request up to the combo box, which in turn delegates its focus to... the first or the second text field? How would it choose? Is this a scenario that we need to solve?
> 
> Presumably we would need a way of telling the focus scope node which sub-node hoisted the focus so it could select the correct delegate. But it's difficult to imagine a control that's trying to pass itself off as a monolithic entity having two internal TextFields. Wouldn't that require enabling keyboard traversal inside a monolithic control? What does that mean? I don't think this is a scenario we need to solve.

Perhaps not, I'm not sure how something like ColorPicker works, but I'm pretty sure it allows keyboard navigation when its drop down is open (same for an open ComboBox).  The delegate however will likely always be the same component though; I think it could be save to say that the delegate is either a fixed value or `null`.

> > > @hjohn I think what's missing in your model is the option to have an independently focusable node inside of a control. For example, think of a Button placed within a Button (via its `graphic` property). Clicking the nested button should not transfer the focus to its parent button. Whether a focus request should be hoisted is a choice that depends on the specifics of the control, and we probably need an API for that.
> > 
> > 
> > Okay, agreed, so it would be good to have Nodes control whether or not the focus traverses up.
> > I could imagine the flag could also in the future perhaps allow the Root Node or Scene to determine whether the Window should receive focus (perhaps special cases like popups or transparent overlays would allow interaction without hoisting focus to the Window?)
> 
> Do you mean window focus in a JavaFX sense, or in a window manager sense? For example, in Windows you can make a window non-activatable, which means it won't steal the focus from other windows when interacted with (for example, an on-screen keyboard window).

Perhaps the analogy does break down here.  I was thinking, if you could interact with a control on an inactive window, the control could decide if interacting with it should focus the window or not with the hoist focus flag.  However, that would presumably also mean the Scene's focus owner shouldn't be changed (or be `null`).

> > Then what about the focus delegate? This is not a full FX property in this API, but seems very similar to `focusOwner` in Scene, in that it determines to which `Node` (keyboard) events are directed. One could say that, since Window/Scene are also focusable, Scene is **delegating** events/focus to a node. Would it be reasonable to have `focusDelegate` be a read-only property like in `Scene`, and perhaps rename it to focus owner (or if that's too drastic, ensure that `Scene` at some later stage --might-- implement the same interface and then duplicating its focus owner property as focus delegate)?
> 
> It works quite well to say that a `Scene` delegates focus to a `Node`. It doesn't seem to work quite as well (linguistically) to say that a `TextField` is the focus owner of a `ComboBox` (the `ComboBox` is also focused, why would the `TextField` be its focus owner?). But linguistics aside, the question is whether `focusOwner`/`focusDelegate` should be a read-only property.

I agree that focusOwner probably only works for well for Scene, although it could have been named focusDelegate and work quite well if there had been sufficient foresight :)

> That's a very good question. It would allow observers to know when the focus delegate changes. But can the focus delegate change? It can probably go from non-null to null, and vice versa. Let's consider a hypothetical control that is like a combo box, but it contains two separate text fields. It seems reasonable that the focus delegate can probably change between the two text fields.
> 
> But the proposed API doesn't seem to easily support this scenario. When I click on the second text field, it hoists the focus request up to the combo box, which in turn delegates its focus to... the first or the second text field? How would it choose? Is this a scenario that we need to solve?

Why doesn't it just set the delegate to whatever hoisted focus in the first place?  A not entirely hypothetical `DateEntryControl` with 3 separate fields internally (with borders stripped) could have a day-month-year area (such controls were quite common before fancy date pickers).  Each of the subfields could set hoist focus.  Clicking on a specific subfield would put the cursor there, but focus the `DateEntryControl` as a whole.

hjohn:
> > It works quite well to say that a `Scene` delegates focus to a `Node`. It doesn't seem to work quite as well (linguistically) to say that a `TextField` is the focus owner of a `ComboBox` (the `ComboBox` is also focused, why would the `TextField` be its focus owner?).
> 
> For what it's worth I am not a fan of using the term focus owner to refer to a focus delegate. The focusOwner of a Scene is the target of key events. The focus delegate of a ComboBox is the _second_ target of key events. Key events are sent to the ComboBox first and the delegate only receives the events that it doesn't consume. To me that's an important distinction.

I don't think the distinction quite holds.  Scene does not delegate all key events.  Menu shortcuts for example are consumed and never dispatched, and I think the same goes for mnemonics.  Navigation keys are dispatched, and only acted upon by Scene when bubbled back up.

However, I now think that "focus delegate" is indeed the better name for this.
 
> Presumably we would need a way of telling the focus scope node which sub-node hoisted the focus so it could select the correct delegate. But it's difficult to imagine a control that's trying to pass itself off as a monolithic entity having two internal TextFields. Wouldn't that require enabling keyboard traversal inside a monolithic control? What does that mean? I don't think this is a scenario we need to solve.

See my example in my reply to Michael, I think it may be worth having a few thoughts about.

beldenfox:
> I don't think the distinction quite holds. Scene does not delegate all key events. Menu shortcuts for example are consumed and never dispatched, and I think the same goes for mnemonics. Navigation keys are dispatched, and only acted upon by Scene when bubbled back up.

What you describe is a common pattern in other UI toolkits but is not how JavaFX works. All key events are immediately fired at the Scene's focus owner and all processing happens within the resulting dispatch chain. The Scene provides a dispatcher that processes mnemonics during the capturing phase (early) and menu accelerators and navigation in the bubbling phase (late). The accelerators also cover the default dialog buttons.

hjohn:
> > I don't think the distinction quite holds. Scene does not delegate all key events. Menu shortcuts for example are consumed and never dispatched, and I think the same goes for mnemonics. Navigation keys are dispatched, and only acted upon by Scene when bubbled back up.
> 
> What you describe is a common pattern in other UI toolkits but is not how JavaFX works. All key events are immediately fired at the Scene's focus owner and all processing happens within the resulting dispatch chain. The Scene provides a dispatcher that processes mnemonics during the capturing phase (early) and menu accelerators and navigation in the bubbling phase (late). The accelerators also cover the default dialog buttons.

But isn't that a distinction without a difference? Technically, you're right, all keys go to the focus owner.  In practice though, some never make it out the door.

Looks like I recalled wrong how menu keys are processed, surprisingly they can be blocked if you want.  I should have checked more closely :)

mstr2:
> Why doesn't it just set the delegate to whatever hoisted focus in the first place? A not entirely hypothetical `DateEntryControl` with 3 separate fields internally (with borders stripped) could have a day-month-year area (such controls were quite common before fancy date pickers). Each of the subfields could set hoist focus. Clicking on a specific subfield would put the cursor there, but focus the `DateEntryControl` as a whole.

I've added a parameter to `getFocusDelegate(Node hoistingNode)`, which indicates the node that hoisted the last focus request to the node on which `getFocusDelegate` is called. The hoisting node is not necessarily a focus delegate, it can be any descendant with the `hoistFocus` flag set.

When `getFocusDelegate(Node hoistingNode)` is called, the control can use the additional information to decide which of its children will be the focus delegate.

hjohn:
Keep open IMHO.

mstr2:
Dispatching an event in the presence of focus delegation is slightly different, because it requires cooperation of the `Node` that is the event target. Previously, we would dispatch an event like this:
```java
EventUtil.fireEvent(EventTarget, Event);
```

With focus delegation, we need to know what the final delegate will be, and we dispatch the event like this:
```java
EventUtil.fireEvent(EventTarget firstTarget, EventTarget finalDelegate, Event event);
```

Of course, this is internal code that can only be called within the `Node` class (as only `Node` knows its final delegate). But we need a way for users to correctly dispatch an event without knowing `Node` internals. [JDK-8303209](https://bugs.openjdk.org/browse/JDK-8303209) comes in handy here, so I've pulled that proposal into this feature.

Old code like `Event.fireEvent(EventTarget, Event)` is simply redefined to call `EventTarget.dispatchEvent(Event)`, ensuring that the overridden implementation `Node.dispatchEvent(Event)` is called. This ensures that events are always fired correctly.

Note that it is not enough to manually construct an event dispatch chain with `EventTarget.buildEventDispatchChain(EventDispatchChain)`, and call `EventDispatchChain.dispatchEvent(Event)` on it. This skips the very important step of setting the correct delegation target. I've added documentation to the existing method, clarifying that it should not be used by application developers. There is no reason to use it in any case, since we now have the `EventTarget.dispatchEvent(Event)` method instead.

mstr2:
This PR is now ready for review. We should finish the discussion of whether we want to have this feature in JavaFX. If we decide to go ahead, it is not required to migrate all controls to use focus delegation at the same time. This PR includes the `Spinner` control as an example, and more controls can be migrated step by step.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -8501,6 +8560,127 @@ public String getName() {
...
+     * node and treats this node as having no focus delegate.
+     * <p>
+     * Focus delegation is often combined with {@link #isFocusScope() focus scoping}.
+     *
+     * @param hoistingNode the descendant of this {@code Node} that hoisted the focus request
+     *                     (not necessarily the focus delegate), or {@code null}
+     * @return the focus delegate, which is a descendant of this {@code Node}
+     * @since 26
+     */
+    Node getFocusDelegate(Node hoistingNode) {
```
beldenfox, line 8814: I tried to tweak the InputMethodStateManager to deal with delegates and discovered there’s no easy way to get a node’s delegate. Shouldn’t there be a public way of retrieving the current delegate?

I don’t think getFocusDelegate should take a parameter. If I was creating a node with multiple focus delegates I would probably want to track the current one in some simple way. But getFocusDelegate doesn’t allow me to do that, I always have to respect the hoisting node parameter being passed in and dig through my descendant nodes to figure out which delegate is being requested. And it implies the only way to change the delegate is to requestFocus on the new delegate and take it through the hoisting process. That seems overly restrictive. I can see why a focus scope node would need to know which node hoisted focus but only at the time the hoist happens.

We’re spending a lot of time thinking about controls that may have more than one delegate. I really doubt anyone will ever create one. If they do it raises a bunch of questions related to accessibility and keyboard traversal that we’re not even discussing.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/DelegatingEventDispatcher.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    public Event dispatchEvent(Event event, EventDispatchChain tail) {
+        boolean retarget = false;
+
+        // Focus delegation is the only scenario in which the event target may be the parent node.
+        // Since we are in the capturing phase, we need to retarget the event to the focus delegate.
+        if (event.getTarget() == parent) {
+            retarget = true;
+            event = event.copyFor(event.getSource(), delegate);
+        }
```
hjohn: I think this improves the code by avoiding a reassignment:
```suggestion
        boolean retarget = event.getTarget() == parent;

        // Focus delegation is the only scenario in which the event target may be the parent node.
        // Since we are in the capturing phase, we need to retarget the event to the focus delegate.
        if (retarget) {
            event = event.copyFor(event.getSource(), delegate);
        }
```

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
hjohn: Does `resolveFocusDelegate` do what you need?  Perhaps that method should be public (and maybe renamed) and the version taking the parameter can be kept internal.  I think it would be unwise to assume there can only be one delegate as this mechanism is there primarily for skins, which are public API and would be used to construct advanced controls that would benefit from multiple delegates.  While FX itself is unlikely to have need of this as its focus is (and must be) on basic, composable controls, there is an infinite amount of useful composed controls that would benefit from having multiple focus delegates.

Examples that would work well with multiple delegates:

- The Date/Time entry control I mentioned with separated TextFields
- Similar to the above but for credit card entry, ip addresses, or any other field with fixed separators; of course this can be done with a single TextField as well, but that's much harder (and you get less context for each field for accessibility -- ie. for a date field it can specifically mention you are in the "month" part instead of just asking for a full date where it may need to explain the order of the individual parts).
- Spreadsheet cells can have controls embedded while the cell retains focus
- Tag editor (where you type something, which is autocompleted from a list of options, and when selected becomes a removable/editable tag -- the address fields of an email client are such an example)

At a minimum IMHO we should ensure this design doesn't prevent the possibility of multiple focus delegates, but preferably it should support it immediately.  I'm less well versed in the accessibility points you raised, could you elaborate?  For example, for keyboard traversal, I don't see why this wouldn't work with multiple delegates (as its similar to having a Skin that has multiple focusable controls)?

beldenfox: > Does `resolveFocusDelegate` do what you need?

Currently we turn on input method processing at the platform level if and only if there's any control in the focus chain that is set up to receive input method events and respond to input method requests. That could be any focusOwner in any focused Scene or any one of the focusOwners' delegates. So I need to see the entire delegate chain.

If we don't expect the focus delegate to change while a focus scope node has focus then all I need is getFocusDelegate with no parameters. If we expect the focus delegate to change dynamically it should become a property.

> For example, for keyboard traversal, I don't see why this wouldn't work with multiple delegates (as its similar to having a Skin that has multiple focusable controls)?

The existing traversal machinery is designed to update a Scene's focusOwner not which delegate is active within the current focusOwner.

In the Date/Time control you give as an example how would the user move focus between the various delegates? What would be the most intuitive model for the user? Personally I prefer the one used in macOS Calendar where Tab is used to move between the month, day, and year (or hour and minute). But that creates a conflict with using Tab to move focus into and out of the Date/Time control as a whole.

In any case the question is whether a monolithic control with multiple delegates is expected to roll its own internal traversal mechanism or if we need to extend the existing mechanism to include delegates. Having each control roll its own internal traversal would hide it from the accessibility frameworks. And if we expect the existing traversal keys (like Tab) to work internally we should extend the existing traversal machinery to work with delegates.

hjohn: > > Does `resolveFocusDelegate` do what you need?
> 
> Currently we turn on input method processing at the platform level if and only if there's any control in the focus chain that is set up to receive input method events and respond to input method requests. That could be any focusOwner in any focused Scene or any one of the focusOwners' delegates. So I need to see the entire delegate chain.

Alright, so a method to get the current delegate would really help there, and I think that information should be available easily enough.

> If we don't expect the focus delegate to change while a focus scope node has focus then all I need is getFocusDelegate with no parameters. If we expect the focus delegate to change dynamically it should become a property.

I think the variant with a parameter has a different intent, and so I think we should not attempt to combine the two.

> > For example, for keyboard traversal, I don't see why this wouldn't work with multiple delegates (as its similar to having a Skin that has multiple focusable controls)?
> 
> The existing traversal machinery is designed to update a Scene's focusOwner not which delegate is active within the current focusOwner.
> 
> In the Date/Time control you give as an example how would the user move focus between the various delegates? What would be the most intuitive model for the user? Personally I prefer the one used in macOS Calendar where Tab is used to move between the month, day, and year (or hour and minute). But that creates a conflict with using Tab to move focus into and out of the Date/Time control as a whole.

Tab would definitely be the most intuitive model, and that's also what I would expect (and how this has worked for such controls since ages past).  So what I think is missing here is that in order to do correct navigation, Scene must be able to query the delegate of its current focus owner.  I think this could be done transparently:

- This PR already directs events to the current delegate, including events that are interpreted as navigation events
- Events targeted at a delegate already bubble up in the usual way, eventually reaching Scene
- Scene, when it wants to act on such an event for navigation, can ask its current focus owner to find out the actual control that received the event (the delegate).  This could be a parameterless `getFocusDelegate` or some such or a property.
- Scene then initiates the navigation logic from the delegate node (let's say it is in a field of a composed DateEntryControl -- the normal logic would find a next/previous field, or if there is none, a next unrelated node).  Navigation in FX works like this everywhere, where the child order determines navigation order, and I think we should keep that.
- A new Node now has been found that should receive focus.  If this node was a node external to the DateEntryControl, it won't have hoist focus set; proceed as normal as make it the focus owner.
- If the new Node does have hoist focus set, then find out its parent, and inform it that the delegate changed; this parent (whether it is the same or different) becomes or already is the new focus owner -- the focus owner may not have changed at all if the navigation only changed the active delegate -- which is probably a good thing

For input methods then I think the missing piece is to find out the delegate chain, but as Scene would need to know this for correct navigation (and so we don't have to build a navigation override system on a per Node basis) I think it could use the same mechanism.

> In any case the question is whether a monolithic control with multiple delegates is expected to roll its own internal traversal mechanism or if we need to extend the existing mechanism to include delegates. Having each control roll its own internal traversal would hide it from the accessibility frameworks. And if we expect the existing traversal keys (like Tab) to work internally we should extend the existing traversal machinery to work with delegates.

I would really not want to see custom traversal mechanisms, because on top of what you said I think it would be unintuitive as well.  I think relying on child order is more than sufficient.  Any custom mechanism is I think more likely to be unintuitive for the user as there is IMHO only one way to navigate intuitively through a UI -- tab/shift tab to focus the next **logical** component (defined by child order and UI hierarchy), and arrow keys go to the visually best next component -- this is what Scene does already, and allowing controls to deviate here further is just not something that is a requirement ever.  Controls already have more than sufficient leeway to define the next "logical" node, even to the point that it would be unintuitive for users...

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -8501,6 +8560,127 @@ public String getName() {
...
+     *
+     * @since 26
+     */
+    private BooleanProperty hoistFocus;
+
+    public final boolean isHoistFocus() {
+        return hoistFocus == null ? false : hoistFocus.get();
+    }
+
+    public final void setHoistFocus(boolean value) {
```
beldenfox, line 8738: In this PR setting hoistFocus on a control only updates the flag in the control itself. What if the control’s skin introduces sub-nodes? I would assume that setting hoistFocus on a control would automatically set that flag on the control’s entire skin. Is that correct? If so what are the mechanics of making that happen?

It seems hoistFocus is serving two roles. One is an optimization so that requestFocus knows whether it should walk the scene graph looking for a focus-scope node. For that use the flag should be propagated through descendants in the graph. But it’s also being proposed as a hoisting barrier e.g. when it’s *not* set it prevents focus from hoisting upward even if there's a focus-scope node higher up in the graph. But a barrier applies just to a single node and should not be propagated. So I’m having a difficult time figuring out when and how this flag should be propagated through the tree. But I’m no expert on skins so maybe I’m assuming this is more complicated than it actually is.

(P.S. I like this PR, it's a huge improvement over what we've got. I'm just trying to understand how it all works.)

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
beldenfox: Yeah, I'm sure traversal is solvable, I just wanted to spend some cycles figuring it out. For example, would you set focusTraversable on the date/time control AND it's delegates? My guess is just the delegates but this is the sort of thing one would want to prototype.

To be clear I'm not trying to solve traversal in this round. I just want to make sure we have mapped out the path forward so the API will accommodate traversal eventually.

hjohn: When I have a few moments, I'll see if I can create the date entry control, and see how it works with this solution.  I can then override Scene code to see if navigation is a quick fix or a huge issue.

mstr2: > Currently we turn on input method processing at the platform level if and only if there's any control in the focus chain that is set up to receive input method events and respond to input method requests. That could be any focusOwner in any focused Scene or any one of the focusOwners' delegates. So I need to see the entire delegate chain.
> 
> If we don't expect the focus delegate to change while a focus scope node has focus then all I need is getFocusDelegate with no parameters. If we expect the focus delegate to change dynamically it should become a property.

The focus delegate can change while a focus scope has focus, but only due to a focus request (i.e. navigation or calling `Node.requestFocus()`). I'm reasonably certain that we don't need or want the delegate to be a property. As of now, the delegate chain is simply re-evaluated in response to a focus request.

However, there's currently no easy way for `Scene` to know when the delegate chain has changed.

> The existing traversal machinery is designed to update a Scene's focusOwner not which delegate is active within the current focusOwner.

Focus traversal is not aware of what a delegate is, nor should it be. This means that focus traversal _will_ update the delegate chain if you traverse into a node that hoists focus.

I think it's important to point out that the primary use case of focus delegation is the creation of black-box controls, but focus delegation is not the same level of abstraction as black-box controls. It's better to think of focus delegation as a basic building block, and black-box controls as a higher-level abstraction.

At the `Node` level (where controls are not a thing yet), it is more approriate to picture focus delegation as a kind of openly visible multi-level focus. In this model, focus traversal doesn't differentiate between nodes based on how they came into being (e.g. as part of a skin), it simply applies the traversal algorithm on the scene graph as it is, being completely unaware of the logical grouping that is established by controls.

> In the Date/Time control you give as an example how would the user move focus between the various delegates? What would be the most intuitive model for the user? Personally I prefer the one used in macOS Calendar where Tab is used to move between the month, day, and year (or hour and minute). But that creates a conflict with using Tab to move focus into and out of the Date/Time control as a whole.

I don't know what it would mean to move focus into and out of a control as a whole. You can only meaningfully move focus into a composite control if it has at least one sub-node that can be focused. If there is no such node, then you trivially can't use focus delegation. If, on the other hand, you have a focusable sub-node, then "moving into" the composite control just means focusing the sub-node. And "moving out" of the composite control means focusing the next logically focusable node in the scene graph. I don't see any conflict here.

> In any case the question is whether a monolithic control with multiple delegates is expected to roll its own internal traversal mechanism or if we need to extend the existing mechanism to include delegates. Having each control roll its own internal traversal would hide it from the accessibility frameworks. And if we expect the existing traversal keys (like Tab) to work internally we should extend the existing traversal machinery to work with delegates.

I haven't extensively tested this yet with custom composite controls, but I think the existing traversal logic will just work as it is.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
mstr2: > In this PR setting hoistFocus on a control only updates the flag in the control itself. What if the control’s skin introduces sub-nodes? I would assume that setting hoistFocus on a control would automatically set that flag on the control’s entire skin. Is that correct? If so what are the mechanics of making that happen?

The `hoistFocus` flag only applies to the node on which it is set, and not to its descendants. Consider a control that uses a sub-control in its skin. If the sub-control also delegates focus to one of its own descendants, then those descendants individually get to decide whether to hoist focus to the sub-control; and if so, the sub-control will further hoist focus to the outer control.

Automatically setting this flag precludes the option to have sub-nodes that _don't_ hoist focus. Consider a Button: it has an optional sub-node (`graphic`) that is independently focusable and doesn't hoist focus. A custom button could conceivably have both: a delegate (maybe a text field), as well as an independently focusable sub-node.

> It seems hoistFocus is serving two roles. One is an optimization so that requestFocus knows whether it should walk the scene graph looking for a focus-scope node. For that use the flag should be propagated through descendants in the graph. But it’s also being proposed as a hoisting barrier e.g. when it’s _not_ set it prevents focus from hoisting upward even if there's a focus-scope node higher up in the graph. But a barrier applies just to a single node and should not be propagated. So I’m having a difficult time figuring out when and how this flag should be propagated through the tree. But I’m no expert on skins so maybe I’m assuming this is more complicated than it actually is.

`hoistFocus` is not an optimization, it's semantically relevant in the way I described.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
beldenfox: > Focus traversal is not aware of what a delegate is, nor should it be. This means that focus traversal _will_ update the delegate chain if you traverse into a node that hoists focus.

If a key event invokes the traversal machinery it will start at the focusOwner, not one of the focusOwner's delegates. There's no way for the existing traversal machinery to tab between the sub-fields of a date/time control that is the focus owner.

I've seen two methods for traversing a date/time control. In one method you use Tab to traverse to the control, through the month/day/year fields, and then beyond the control. That doesn't work well in grid-based layouts where the user expects three presses of Tab to reliably move them three columns to the right. In those designs it's not uncommon for traversal between month/day/year to move to other keys.

> At the `Node` level (where controls are not a thing yet), it is more approriate to picture focus delegation as a kind of openly visible multi-level focus.

Traversal needs to pick one node as a starting point.

The Scene needs to pick one node for processing input method requests (the IM window needs to track the caret and there can only be one).

I'm sure accessibility also has issues with multi-level focus but don't understand the platform API's well enough to go into any detail on that. I can see how JavaFX currently sets this up for ComboBoxes and am assuming it's correct. If it is I'm not sure the existing accessibility attributes are sufficient to convey focus beyond two levels. (We really need an accessibility guru to weigh in on this).

I don't think any of these issues are necessarily deep, just that they need to be hashed out.

> I'm reasonably certain that we don't need or want the delegate to be a property.

Currently you're tracking the hoisting node as an internal reference and then using it to constantly re-compute the delegate whenever you need it (going through a loop to validate the hoisting node each time). It seems you could use the hoisting node to determine the delegate while processing requestFocus and just track the delegate directly. Do you expect the delegate associated with a particular hoisting node to change over time? If not, why not just compute it once?

In any case the input method state manager needs to know if the delegate chain changes. That's still a problem that needs to be solved.

mstr2:
> _Mailing list message from [Andy Goryachev](mailto:andy.goryachev@oracle.com) on [openjfx-dev](mailto:openjfx-dev@mail.openjdk.org):_
> 
> I think I narrowed down my objection to this proposal. My main problem with it is that it tries to add properties to the wrong entity (Node).
> 
> In a typical application, the focus traversal requirements usually come in the context of a higher level aggregate such as a form, or a panel, or a window, and not the individual nodes. An example from Swing and AWT is the FocusTraversalPolicy class. This is no accident - the focus traversal might be dynamically dependent on other aspects of the UI state, for example the kind of information already present in a form.
> 
> Trying to implement it at the Node level would be extremely difficult, and one will inevitably resolve to creating a kind of controller that determines the traversal order at the higher level.
> 
> A better approach, in my opinion, would be to revive the FocusTraversalPolicy API [0], [1] (after some modernization). The most important feature that would be added is enabling ability to plug in a custom traversal policy that can address any application-level requirements with relative ease.

I don't think that these two proposals are at odds, in fact they are completely orthogonal. Focus delegation has nothing to do with focus traversal. We're not trying to answer the question what happens when (or how) we move the input focus to another control.

Focus delegation is an implementation of multi-level focus for composite controls. It spans different levels of abstraction, something that JavaFX already attempts to do right now, but fails due to a lack of supporting infrastructure. I don't see how a customizable focus traversal API addresses _any_ of the two major improvements that focus delegation brings to the table:

1. A consistent focus delegation chain: focusing a descendant control keeps the outer control focused, which is what would be expected in a black-box control model.
2. Events are not duplicated, they are logically consistent for all observers.

hjohn:
> _Mailing list message from [Andy Goryachev](mailto:andy.goryachev@oracle.com) on [openjfx-dev](mailto:openjfx-dev@mail.openjdk.org):_
> 
> I think I narrowed down my objection to this proposal. My main problem with it is that it tries to add properties to the wrong entity (Node).
> 
> In a typical application, the focus traversal requirements usually come in the context of a higher level aggregate such as a form, or a panel, or a window, and not the individual nodes. An example from Swing and AWT is the FocusTraversalPolicy class. This is no accident - the focus traversal might be dynamically dependent on other aspects of the UI state, for example the kind of information already present in a form.

The reason Swing has a FocusTraversalPolicy is because its layout system is decoupled from focus traversal. Layout managers in Swing determine the positions of components but do not expose any information about a good traversal order of those components. As a result, the container has no built-in understanding of how focus should move between components based on layout.

So Swing needed a FocusTraversalPolicy because layout managers (although flexible) do not provide any inherent information on focus traversal order. In contrast, JavaFX integrate layout and focus order more tightly through hierarchical scene graphs, node ordering, and orientation (for some controls).  In other words, there is more than enough ways to influence focus traversal order:

- Making logical groups which comes naturally with the various containers
- Adding nodes to containers in the desired focus order
- Leveraging disabled, visible, and focusTraversable properties for even more control

For instance, if a GridPane should traverse top-to-bottom rather than left-to-right, it’s simply a matter of adding the nodes in top-down order. JavaFX respects the order in which nodes are added, regardless of their visual grid positions.

So, I don't think grabbing this idea from Swing and adding it to FX can ever be the right solution as it was introduced there to solve a completely different problem (not using order of children + layout managers being decoupled and unable to confer focus traversal information).  If there are currently problems with FX's focus traversal order that can't easily be solved by using the existing tools (especially node order), that are common problems for users, then please share these, and let's see how these concerns can be addressed by enhancing what is already there instead of overriding it.  Perhaps a `Node.setViewOrder` type mechanism is needed, but for focus -- but I struggle to think why this couldn't be solved by simply moving the child in the correct position...

hjohn:
> _Mailing list message from [Andy Goryachev](mailto:andy.goryachev@oracle.com) on [openjfx-dev](mailto:openjfx-dev@mail.openjdk.org):_
> 
> My Swing is a bit rusty now, but I think it managed to handle geometry-based navigation just fine. But it is not the reason I refer to Swing's FocusTraversalPolicy.
> 
> The application requirements might call for traversal that is governed not only by geometry, but also by the state of the UI in general, even dynamic, based on the information already entered by the user (as in "skip to item 25 if you answered "no" here).

This can easily be handled by disabling the controls that are not relevant when "no" was answered.  This can be done dynamically and something that most UI's have been doing already since UI's were invented (I remember doing this in the early 90's).  This is not a good enough reason to offer an alternative (and not complimentary) focus traversal mechanism.  

If for some reason those controls must remain "enabled", you can instead of disabling them remove their focus traversable flag -- note that you're actively confusing the user then though.
 
> The value of the Parent.traversalPolicy is not only that it helps with composite controls and skins, but also that it enables the application developers to implement whatever crazy requirements they might have.

FX shouldn't cater to crazy requirements; for inclusion in FX, this must be a **common** requirement, while offering sufficient power for "crazy" requirements.  FX makes the common cases easy (I rarely find fault with FX's default focus traversal), and there's more than sufficient controls available to support more exotic requirements.

> Also, the FocusTraversalPolicy (whatever form it eventually might take) is not a foreign concept transplanted from an older system. JavaFX, in fact, already has these policies internally implemented, offering a small, but important benefit of not altering the behavior of the existing code.

Yes, and almost universally when I encounter its usage in the behavior/skins, they actually show a lack of understanding of how FX works, or are no longer relevant as FX evolved.  I'm very sure we can remove the entire internal system (think of how much code that would safe, and how similar the user model then becomes for their own skins) and not lose any functionality by just using what FX offers with regards to focus traversal.  One of the biggest stumbling blocks for this has been removed already (see #1582 ).

As you didn't share any use cases that are even remotely common, it is still hard for me to imagine why we need an alternative navigation system when JavaFX's system out-of-the-box was by default 10x better than what Swing offered before it introduced the focus traversal policy.

hjohn:
> There is more: I think the kind of issues we see stem from the sub-optimal event dispatching mechanism in JavaFX, the weirdest part of which is creating multiple event instances for the same event. 

You've mentioned this a number of times.

Events are copied because FX uses the **source** field to indicate where the receiving event handled was attached.  As it bubbles down, the source field must be updated if you want it to contain the correct value.  This is actually a performance optimization for cases where a single listener can be attached to say all table cells, and use the `source` field to distinguish which cell was involved.

It's worth noting that this is actually contrary to how the source field is documented in Event's superclass `EventObject`.
 
A better solution for this would have been to use a `ThreadLocal` to hold the source (or the soon to be finalized `ScopedValue`).  It may be possible to back-implement this still without introducing any incompatibilities.  It may safe a few copies (although there won't be any performance wins as even MouseMove events are simply too infrequent to matter).  Events would still not be fully immutable though as they still have the consumed flag, and custom implementations could allow for tracking additional information (like "ifNotConsumed" actions).

> The key events should be targeting the keyboard focus owner, in case of the Spinner and similar controls, the TextField inside of it. "Control" and "Skin" are abstractions that have nothing to do with that, and the Spinner itself might conceptually be a part of some other custom control.

I disagree.  The Control is intended to function as single encapsulated unit.  Its substructure is pure happenstance, and any control (no matter how complex) could choose to not use any sub-controls at all but completely implement them anew.  The use of sub-controls is for implementation convenience and reuse, but should not be exposed to users. They reason about a `ComboBox` as a single unit.  How such a control handles events is up to it (it could for example choose not to forward events at all, but instead directly alter its sub-controls).  The forwarding of events here is again for convenience and re-use, not because we want users to see or directly interact with those events, they are part of the black box that a Control embodies.

What Spinner and related controls currently do is taking a poorly constructed short-cut to do as little work as possible to make the Spinner work (poorly duplicating events, and re-firing them at the top level).  They didn't have to do it this way at all.

mstr2:
> There is more: I think the kind of issues we see stem from the sub-optimal event dispatching mechanism in JavaFX, the weirdest part of which is creating multiple event instances for the same event. The key events should be targeting the keyboard focus owner, in case of the Spinner and similar controls, the TextField inside of it. "Control" and "Skin" are abstractions that have nothing to do with that, and the Spinner itself might conceptually be a part of some other custom control.
> 
> To answer the last two points:
> 
> 1. the leaf TextField should be the focused Node, not its parent Control, or its more complex custom control.

But that would be a bad experience for application developers. In most cases, application developers interact with a scene graph of controls. The skins are just an implementation detail that is mostly invisible, and easily changeable with a different theme.

Developers should be able to expect that when they `requestFocus()` on a control, the control will be focused. This is what JavaFX currently tries to do (but it fails to do it right). Your suggestion would be a major breaking change, it would mean that a skin arbitrarily gets to decide which node will _actually_ be focused.

This leaks an implementation detail into the user model with huge consequences: simply changing a skin becomes brittle and potentially impossible, as user code can no longer reason about controls independently from their skins. User code would no longer be able to know which node will be focused, or which node will be the target of events.

> 2. to avoid duplicate events, don't duplicate events.

Yes, this is the logical solution for your suggestion. But then observers will not be able to know that a key event is targeted at a spinner. Let's assume that a scene listener is installed that intercepts key events going to spinners. Using your suggestion, the listener would have to look whether the key event is targeted at a TextField (which is an implementation detail of a skin, and could change with another theme), and then figure out whether the parent of that TextField is a Spinner (which is another implementation detail).

hjohn:
> John,
> 
> Thank you for continuing this discussion. This particular response might need its own thread and a topic (FocusTraversalPolicy).
> 
> I want to emphasize that JavaFX should provide support for "crazy" as you put it requirements. In my experience, the world of enterprise software is full of "crazy" requirements that have solid business justification.

Andy, fine.  However, this is a quote from you, emphasis mine:

> The value of the Parent.traversalPolicy is not only that it helps with composite controls and skins, but also that it enables the application developers to implement **whatever crazy requirements** they might have.

Maran23:
After I did a deep check on the focus/focus traversable implementation and usage, I think something like this is 100% needed. 

I just wonder if we can achieve this a bit easier/simple.

Especially `ScrollBar` and Table related `Controls` had interesting implementations to 'trap' the focus.

An easy improvement would be that every `Node` can have a proxy where the focus should be requested (instead). `ScrollBar` needs it set for whatever `Control` it should scroll in. The Table Column Header needs that, the `VirtualFlow` as well, rows and cells as well. But I'm not even 100% sure about the idea. We probably need to check every case and find the usecase. I did not checked the `FakeFocusTextField` and related yet.

The Idea that a `Control` is the top level node that receives focus and everything else is a black box sounds interesting. But there are controls that do not request focus at all, like the `Accordion`. There, only the `TitledPane` will.

Or a complete different approach: We retrofit the `focusWithin` in those controls. As far as I could see, the goal was always to have the focus rect when pressing an inner `Control`. Well, this property is exactly there for this. And the focus request hacks can be removed. 

Needs more research.

andy-goryachev-oracle:
I also thought about this issue, and what @Maran23 just said resonates with me.

The `Control` is not special.  All it brings is `Skin` and a tooltip, otherwise it's a `Region`.  What's more, it might be a composite thing: a Control can be constructed out of other Controls, which can be composite as well.  So I think it's a bit of a mistake to elevate it to a special place (with regards to focus, in this case).

The `Scene` keeps track of a focused **Node**, not a focused `Control`.  In the case of a composite Control, it might be some inner Node which may not be a `Control`.  And that is how it should be.  

Event dispatching deals with the focused `Node`, and if a more complex control hierarchy wants to change the way the focus is indicated, it should handle/filter the events correspondingly (focusWithin).  

The idea of a black-box Control is just a way of saying that Control (regardless of whether it's a simple or a composite one) is just a convenient façade from the application perspective.  It allows the application to deal with the complexity without knowing the internals.

Does it mean we can bind the outer control's `focused` property to the `focused` property of one of its inner components?  I don't know, needs to be investigated.

Maran23:
So I wasted nearly my whole weekend trying to see how we could get rid of `setFocused(..)` (deprecate) and fix everything that needs it. I came across nearly all corner cases we have now, and checked all Controls and their traversable behavior.

Those are all my notes I wrote while doing so, straight up copying them here. Maybe they are useful.

Requirements:
- MANUALLY synchronize Focus from ComboBox/DatePicker/Spinner also into FakeFocusTextField. But rather than being a focus owner, it should just synchronize the focus so that it can react correctly (caret etc.) and also look correct (focused, CSS).
- RenderThemeImpl: Should synchronize focus from DOM state into the component without requesting real focus. Again, this is the same usecase as above. We want to manually synchronize focus so that the Controls will react correctly (caret etc.) and also look correct (focused, CSS)
- Cell: This one would be completely avoidable, but we cannot break it now. This is only done to cancel (or soon, stop) edit and for CSS. Both could easily be implemented on their own (Like TabPane)
- Some Controls need to know if something else was interacted with (focus changed) so they commit or cancel their value
- Complex Control like ScrollPane or Table want to always show the focused rectangle even though sub controls in there are focused, like the ScollBar or Row/Cell/Cell TextField. This can be achieved by making those not focus traversable and using focusWithin. As focusWithin was not available when the API was made, those Controls are usually requesting their focus back into the Complex Control when received or rely on hacks. And they do it wrong, e.g. VirtualFlow sometimes receives focus when scrolling
- There are Controls like Accordion or ScrollPane that never want to have focus, only their TitledPanes/Content should have it (It still can be enforced by setting focus traversable back to true, which is a bit weird)

Disadvantages:
- Due to the nature of how ComboBox/DatePicker/Spinner are built up, they need various hacks: FakeFocusTextField, redirecting events and overwriting the traversal engine.
- Committing or Cancelling based on the focusedProperty with focusTraversable set to false is not possible and not even clear if it should. You may want to do something like writing something in a TextArea and click on Buttons in a ToolBar to make text bold, which should not shift focus away and not commit
- Making some Controls focus traversable like Accordion will suddenly focus Accordion, which is not really visible (no indicator as well), makes no sense and even worse, you are getting focus trapped (tested today)
- As Cells inside a Table are focused with hacky `setFocused,` they will contribute to focusWithin, which will make the the table focusWithin true even though something completely different outside the Table is now focusOwner (focused)

So what do we need then:
- Some Controls need a manual focus sync (from A -> B), but will never become focus owner and should not contribute to focusWithin and focusVisible. They also should not be traversable to and should receive the events correctly.
  - In this case, the Control should be a blackbox, and everything under it should be not accessible with focus and focus traversable, while still receiving events if needed.
  - This seems like a strong candidate for focus delegation, but it would be good to keep it simple. Even keeping FakeFocusTextField with the focus sync and redirecting the key events in a smart way (not like now), while everything else is fixed (no custom traversal engine) seems like a good compromise
- Some Controls are rather a whitebox, never want focus and never be focus-traversable. Everything it contains should be accessible and focus traversable. 
  - Basically focus delegation reversed somewhat
  - This is a bit weird, because they have set focus traversable to false, but we can set them focus traversable and they behave completey weird and it makes no sense, even trapping focus
    - Maybe it makes more sense that those SHOULD be focus traversable, but that only means that their children can be. Making e.g. the ScrollPane not focus traversable means the content can not be access with TAB, even if the content is focus traversable.
- Complex Controls like the Table are in between black and whitebox. Most of the children should also never be focused, never be focus traversable, while some may should. Example: Column Header for sorting/filtering or cells
  - Focus Rect can be solved easily with focusWithin
  - Cells may should receive focus and be focus owner, finalizing the weird implementation we have now and making it complete. FocusWithin will make sure the everything still looks fine

Conclusion:
- We need a way to control (for a `Control`) which children are accessible and which not, without hacks, traversal engine hacks, key redirection hacks, messing up focusWithin and still be focusable (manually), even if there is a little bit of manual work required for manual focus and key events.
- A `Control` should specify if it itself can receive focus / be focus traversable (so the Control itself can get focus) or just redirect it completely (focus traversable is true, but it will not be focused itself but the children)
  - Accessible children can be traversed to and receive focus and key events as normal
  - Non Accessible childrens can not be traversed to and do not receive focus, unless it is manually synced. They will receive key events manually - we need to specify that as well
  - Non accessible children can request focus (there is no way for us to forbid that probably, but it may break expectations when further pressing TAB or SHIFT+TAB. 
    - We may can implement a warning for this case, although I would like to not do a search in the Scene Tree' to MAY find a mismatch and make everything more complex and slow
    - In general, I would like to avoid doing complex 'search the Scene Tree' operations for that
 
The biggest motivation is to have a straight baseline which solves all the cases, not just one. I think all those cases do share a somewhat common problem.
I think it should be okay to use other Controls inside the Skin. 
I also think it should be possible for us JavaFX Users to write a Control and Skin that can also facilitate whatever we come up with. 
This would make writing own Controls with Skins much easier. Whatever we learn should also influence the behaviors.

beldenfox:
@maran23 I’ve never paid much attention to the Node focused property since it doesn’t play any role in the delivery of key events. It would be really helpful if you could give a brief summary of what it does and, more importantly, how it’s used.

Based on your notes it looks like text input controls consult this property to determine when to show their carets. In the context of this PR this means that the “focused” property needs to be set on the focus owner’s delegate (the TextField) rather than the focus owner (the Spinner or ComboBox) as the spec currently reads.

This is just another part of the system that needs to be shifted from the focus owner to the delegate. We also need to use the delegate as the target for Key and InputMethod events and ensure it’s consulted for input method requests.

(I do wonder if it would make more sense for Spinner and ComboBox to set the focus owner to the TextField instead. This seems to be how the system was designed to function. The focused property would work as expected, the pseudo-classes would work as expected, events would automatically be targeted at the right control, input method requests would work. Even the accessibility API seems to be designed for this; the FOCUS_ITEM accessibility property makes it easy to retrieve the Spinner or ComboBox from the TextField.)

It also looks like container controls like TableView are setting the “focused” property on their cells. Why? To add the “focus” pseudo-class to the cell so it displays a focus ring? Whatever TableView is doing with “focused” what’s your proposed replacement?

Maran23:
@beldenfox,
@mstr2 and my idea is to deprecate `setFocused(..)`.
Because it does not work well with `focusWithin` (as can be seen in the table right now). It is hacky, the focus is not cleared automatically (You basically need to do everything manually), which makes it confusing. Especially as we have `requestFocus`.

It probably exists to workaround the problems I mentioned above. So those need a proper fix. In an old ticket somewhere, I did read something about 'sub focus'. So back then, something like that was planned it looks like.

> This is just another part of the system that needs to be shifted from the focus owner to the delegate. We also need to use the delegate as the target for Key and InputMethod events and ensure it’s consulted for input method requests.

Yes. I don't know how (a delegate system is one idea), but this needs a proper solution for that reasons.

> (I do wonder if it would make more sense for Spinner and ComboBox to set the focus owner to the TextField instead. This seems to be how the system was designed to function. The focused property would work as expected, the pseudo-classes would work as expected, events would automatically be targeted at the right control, input method requests would work. Even the accessibility API seems to be designed for this; the FOCUS_ITEM accessibility property makes it easy to retrieve the Spinner or ComboBox from the TextField.)

The problem with that is that Controls like the `ComboBox` have a listener bound to the `focusedProperty` (and probably a lot of user code out there as well). So as soon as the `TextField` with be the focus owner, the Control thinks the user clicked somewhere else and commit their value.

> It also looks like container controls like TableView are setting the “focused” property on their cells. Why? To add the “focus” pseudo-class to the cell so it displays a focus ring? Whatever TableView is doing with “focused” what’s your proposed replacement?

My idea is to set the pseudoclass manually (for backward compatibility with CSS) and just decide whether to commit the value or not. No need to use the `focusedProperty` at all.

With that fixed, we can utilize the `focusWithin` CSS style to always show the focus ring even if something inside the table is focus owner, like a `TextField` or even a Cell, if a developer makes it focus traversable (default is false right now). This enables us to remove all workarounds that try to keep/shift the focus just for the sake of the focus ring (like the `ScrollBar` ot table headers, that always try to shift the focus away just for the focus ring).

beldenfox:
@maran23 These days I try to avoid using the term “focus”. In JavaFX that can mean many different things. A node might have “focus” if:

1) it’s the focus owner of a scene;

2) it’s the focus owner of a scene and the scene’s stage is the active window;

3) it matches the “focus” pseudo-class and displays a focus ring;

4) it’s the target of key events.

The Node focused property used to track definition 2. Now it’s a fuzzy combination of 2 and 3.

When you say you want to implement “commit on focus lost” does that include when the stage loses OS focus (definition 2)? That’s the behavior you would get using the Node focused property. If committing could cause an alert to pop up you might be creating a usability problem; when the user tries to check their e-mail the JavaFX app would keep popping an alert.

The Controls side of JavaFX is using a composition approach that the Graphics side of JavaFX wasn’t designed for. A ComboBox wants the focus owner to be the logical control (the ComboBox) and the key event target to be the internal TextField. The Scene doesn't currently make that distinction.

(It’s interesting to compare JavaFX with the Win UI toolkit which was designed for composition. Win UI makes a distinction between the logical control and the key event target and its Event object can track both elements.)

This PR separates the focus owner and the event target by introducing the delegate. The downside is that a delegate can in turn have a delegate which complicates things. It might make more sense for Scene to track the key event target as a property separate from the focus owner. Based on your comments it looks like the "focused" Node property should remain with the focus owner but the "focus" pseudo-class should follow the key event target.

Maran23:
> (It’s interesting to compare JavaFX with the Win UI toolkit which was designed for composition. Win UI makes a distinction between the logical control and the key event target and its Event object can track both elements.)

This sounds interesting indeed. IMO it makes sense to compare and check other UI frameworks and how they dealt with this situation. That makes sense, I just wonder how we could implement something like that in retrospect without changing and breaking too much.

PR #1635, opened 2024-11-12
8344067: TableCell indices may not match the TableRow index

Maran23:
This PR fixes a bug where the `TableCell` indices can be outdated (not synchronized) with the `TableRow` index.

What normally happens is:
- Index is changed: Cell update is requested when the row is empty (otherwise noop)
- Item is changed: Cell update is requested, which will now update the indices of the underlying `TableCells`

Under some circumstances, when a `TableRow` is reused (e.g. index 60 -> 1) the item can be the same 
-> `oldIndex != newIndex && oldItem == newItem`
This can happen when the items of a `TableView` are changed, but some items are the same in both item lists (Think about a filter, where we filter down 60 to 2 items).

-> In this scenario, the cell update is not triggered, so the underlying `TableCell` indices will not be updated ever (e.g. they still have index 60 set, but the row has 1 now). 

The fix is to always update the underlying `TableCell` indices when the `TableRow` index changed. 
While usually the item is different when the index changed, this is not always the case (there is no guarantee that the item changed, as we can see in the example, where the cell is reused).

---

Also made sure that the issues linked in the code and ticket do not regress:
- https://bugs.openjdk.org/browse/JDK-8095357
- https://bugs.openjdk.org/browse/JDK-8115269

---------
### Progress

### Issue
 * [JDK-8344067](https://bugs.openjdk.org/browse/JDK-8344067): TableCell indices may not match the TableRow index (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1635/head:pull/1635` \
`$ git checkout pull/1635`

`$ git checkout pull/1635` \
`$ git pull https://git.openjdk.org/jfx.git pull/1635/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1635`

`$ git pr show -t 1635`

</details>

<a href="https://git.openjdk.org/jfx/pull/1635.diff">https://git.openjdk.org/jfx/pull/1635.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1635#issuecomment-2471600990)
</details>

### DIFF
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewRowTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewRowTest.java
@@ -219,16 +219,16 @@ public class TableViewRowTest {
 
         row.updateIndex(0);
 
-        List<TableCell<String, String>> cells = row.getChildrenUnmodifiable().stream().
-                filter(TableCell.class::isInstance).map(e -> (TableCell<String, String>) e).toList();
+        List<TableCell<String, String>> cells = row.getChildrenUnmodifiable().stream()
+                .filter(TableCell.class::isInstance).map(e -> (TableCell<String, String>) e).toList();
         for (TableCell<String, String> cell : cells) {
             assertEquals(0, cell.getIndex());
         }
 
         row.updateIndex(1);
 
-        cells = row.getChildrenUnmodifiable().stream().
-                filter(TableCell.class::isInstance).map(e -> (TableCell<String, String>) e).toList();
+        cells = row.getChildrenUnmodifiable().stream()
+                .filter(TableCell.class::isInstance).map(e -> (TableCell<String, String>) e).toList();
         for (TableCell<String, String> cell : cells) {
             assertEquals(1, cell.getIndex());
         }
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewRowTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewRowTest.java
@@ -27,6 +27,8 @@ package test.javafx.scene.control;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import javafx.scene.control.SelectionMode;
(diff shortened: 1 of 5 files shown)


andy-goryachev-oracle:
I'll review, thanks!

andy-goryachev-oracle:
The bug is present in TableView and TreeTableView, but not TreeView - is this correct?

modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewRowTest.java
```diff
@@ -205,4 +207,30 @@ protected boolean isItemChanged(String oldItem, String newItem) {
...
+        TableView<String> table = ControlUtils.createTableView();
+
+        TableRow<String> row = new TableRow<>();
+        row.updateTableView(table);
+
+        stageLoader = new StageLoader(row);
+
+        row.updateIndex(0);
+
+        List<TableCell<String, String>> cells = row.getChildrenUnmodifiable().stream().
```
andy-goryachev-oracle, line 222: minor suggestion: maybe reformat to keep one statement per line

```
row.getChildrenUnmodifiable().stream().
   filter(TableCell.class::isInstance).
   map(e -> (TableCell<String, String>) e).
   toList();
```

andy-goryachev-oracle: (here and elsewhere)

modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewTest.java
```diff
@@ -6329,4 +6329,54 @@ public void startEdit() {
 
         assertEquals(1, startEditCounter.get());
     }
+
+    @Test
+    void testReSetItemsWithSameItemShouldUpdateCellIndices() {
+        table.setFixedCellSize(24);
```
andy-goryachev-oracle, line 6335: fixed size is not essential to the test, right?  I've tested with the SCCE in the ticket, it has no effect.

it's probably ok to leave it here though.

andy-goryachev-oracle (approved):
tested with SCCE found in the ticket on macOS M1 14.7, also some limited testing with the monkey tester, no issues found.

looks good; left some very minor formatting suggestions - will re-approve should you decide to fix.

thanks for writing the tests!

1 reviewer should be enough.

Maran23:
> The bug is present in TableView and TreeTableView, but not TreeView - is this correct?

Yes since `ListView` and `TreeView` do not need to do any synchronization when their index or item changed. But I did not test them.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewTest.java
Maran23: Ah yes, I used it initially to better predict the `ScrollBar`, but is indeed not necessary.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewRowTest.java
Maran23: I usually keep it this way as it is still readable and less space. But changed the formatting so that the statement with the dot `.filter` is in the new line. This is as far as I know what most people do.

andy-goryachev-oracle: there is generally added benefit in case of an exception the line number would point to the offending code, rather than to the whole line.

it's probably not important here as we don't deal with deciphering ancient logs as much.

andy-goryachev-oracle (approved):

PR #1637, opened 2024-11-14
8343196: Add build property to identify experimental builds of JavaFX

arapte:
Introduce a new build property `jfx.experimental.release.suffix` to be used for the early access builds of an under development feature in JavaFX.
This property would be set to a value ONLY in a branch specific to an experiment in [jfx-sandbox](https://github.com/openjdk/jfx-sandbox) repo.
For example:
For the metal branch in jfx-sandbox repo, it would be set to `-metal`.  The javafx.version for the early access build generated from that branch would be `24-metal`.

And it would always be empty for master branch in both main jfx repo and in jfx-sandbox repo.
This change has no effect on a regular developer build from master branch.

---------
### Progress

### Issue
 * [JDK-8343196](https://bugs.openjdk.org/browse/JDK-8343196): Add build property to identify experimental builds of JavaFX (**Enhancement** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1637/head:pull/1637` \
`$ git checkout pull/1637`

`$ git checkout pull/1637` \
`$ git pull https://git.openjdk.org/jfx.git pull/1637/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1637`

`$ git pr show -t 1637`

</details>

<a href="https://git.openjdk.org/jfx/pull/1637.diff">https://git.openjdk.org/jfx/pull/1637.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1637#issuecomment-2476182250)
</details>


arapte:
@kevinrushforth  Please take a look.

kevinrushforth:
Reviewers: @kevinrushforth  @johanvos 

/reviewers 2

kevinrushforth:
@arapte Can you merge in the latest master? The macOS GHA failure is due to your being out of date.

The fix looks good to me. I'll do some testing before approving.

arapte:
> @arapte Can you merge in the latest master?

Thanks @kevinrushforth , The checks pass after merging in the latest master.

build.gradle
```diff
@@ -695,7 +695,15 @@ if (HUDSON_JOB_NAME == "not_hudson") {
...
 } else {
-    relSuffix = IS_MILESTONE_FCS ? "" : jfxReleaseSuffix
+    if (IS_MILESTONE_FCS) {
+        if (jfxExperimentalReleaseSuffix != "") {
+            fail("An experimental release branch should not be used for FCS build")
+        }
+        relSuffix = ""
+    } else {
+        relSuffix = jfxExperimentalReleaseSuffix != "" ?
+                        jfxExperimentalReleaseSuffix : jfxReleaseSuffix;
```
kevinrushforth, line 705: One thought I had: we might also want to set `relOpt = "-${buildTimestamp}"` here like we do for internal builds so we always see a time stamp for experimental builds. What do you think?

build.properties
```diff
@@ -38,6 +38,12 @@
 # will cause the suffix to become blank.
 jfx.release.suffix=-ea
 
+# The experimental release suffix is set only in jfx-sandbox repository
+# for a speficic branch. When this property is set it overrides
```
kevinrushforth, line 42: Do you want to say something about it needing to start with a `-`? Or do you think the example of `-ea` above will be clear enough?

kevinrushforth (approved):
Looks good. I left a couple comments and will reapprove if you decide to make the changes.

johanvos:
I might be confused, but wasn't the idea to use a new property (e.g. jfx.feature.variant) instead of overloading the existing jfx.release.suffix?

kevinrushforth:
> I might be confused, but wasn't the idea to use a new property (e.g. jfx.feature.variant) instead of overloading the existing jfx.release.suffix?

This PR does add a new property in `build.properties`,`jfx.experimental.release.suffix` and does not require modiffying `jfx.release.suffix`. For example, in the "metal" branch of the jfx-sandbox, we would define this as:

```
jfx.experimental.release.suffix=-metal
```

build.gradle
johanvos: I agree with that. It is often useful to have timestamp info in the builds during development.
I know I'm a broken record on this topic, but if possible, it would be great if (parts of) this logic can be moved into a separate file in order to make the build.gradle smaller and more readable. I'm not insisting on this for this PR, but I think it is worth thinking about this whenever changes to build.gradle are made.

johanvos:
> > I might be confused, but wasn't the idea to use a new property (e.g. jfx.feature.variant) instead of overloading the existing jfx.release.suffix?
> 
> This PR does add a new property in `build.properties`,`jfx.experimental.release.suffix`, and does not require modifying `jfx.release.suffix`. For example, in the "metal" branch of the jfx-sandbox, we would define this as:
> 
> ```
> jfx.experimental.release.suffix=-metal
> ```

My apologies, I misread the proposed code. 
It is a bit confusing to have "jfx.release.suffix" and "jfx.experimental.release.suffix" where the former refers to an artifact property and the latter refers to a major feature/variant/... like metal. But then, consistent naming is always difficult so I'm ok with the proposed suggestion.

build.gradle
nlisker: > I know I'm a broken record on this topic, but if possible, it would be great if (parts of) this logic can be moved into a separate file in order to make the build.gradle smaller and more readable.

Also at the risk of sounding like a broken record, https://github.com/openjdk/jfx/pull/1232 has been waiting to start this process for a while now and it's stuck waiting for explanations on how publishing works, something only Oracle and Gluon do. If someone can help Jendrik Johannes there then the ball will get rolling. I tried to split the big build file into separate files for each module, but it's entangled with the code in that PR. Completing that PR might also remove the need for the JavaFX plugin.

kevinrushforth:
> It is a bit confusing to have "jfx.release.suffix" and "jfx.experimental.release.suffix" where the former refers to an artifact property and the latter refers to a major feature/variant/... like metal. 

I think you have a good point. Maybe we should drop the '.suffix' in the name of this property (in which case, it wouldn't have a `-` so the logic in `build.gradle` would add the `-` when needed). It would probably be easier to document it as well.

What do you think?

kevinrushforth:
If we do this, then here are some possible names:

* jfx.experimental.release
* jfx.experimental.release.name
* jfx.variant.release
* jfx.variant.release.name
* jfx.variant
* jfx.feature.variant

Thoughts?

johanvos:
> > It is a bit confusing to have "jfx.release.suffix" and "jfx.experimental.release.suffix" where the former refers to an artifact property and the latter refers to a major feature/variant/... like metal.
> 
> I think you have a good point. Maybe we should drop the '.suffix' in the name of this property (in which case, it wouldn't have a `-` so the logic in `build.gradle` would add the `-` when needed). It would probably be easier to document it as well.
> 
> What do you think?

That makes sense (removing the word ".suffix")

johanvos:
> If we do this, then here are some possible names:
> 
> * jfx.experimental.release
> * jfx.experimental.release.name
> * jfx.variant.release
> * jfx.variant.release.name
> * jfx.variant
> * jfx.feature.variant
> 
> Thoughts?

I don't have objections against any of those.
If I have to vote, my vote would go to "jfx.feature.name" (which is not in the list ;) ) because I associate variant rather with a single option from a wider set (e.g. prism variants are sw, es2,,...)
But I am more than happy with any name on this list.

build.gradle
johanvos: > > I know I'm a broken record on this topic, but if possible, it would be great if (parts of) this logic can be moved into a separate file in order to make the build.gradle smaller and more readable.
> 
> Also at the risk of sounding like a broken record, #1232 has been waiting to start this process for a while now and it's stuck waiting for explanations on how publishing works, something only Oracle and Gluon do. If someone can help Jendrik Johannes there then the ball will get rolling. I tried to split the big build file into separate files for each module, but it's entangled with the code in that PR. Completing that PR might also remove the need for the JavaFX plugin.

Fair remark. I just commented on the PR.

nlisker:
Variants are also something [Gradle](https://docs.gradle.org/current/userguide/variant_model.html#sec:abm-configuration-attributes) uses to select one of several components, so I suggest to avoid this overload. JavaFX should use Gradle variants for selection of operating systems.

I would have went with `jfx.experimental.name` :) because the sandbox has experimental features, as opposed to other features that are directly integrated into the main stream.

build.gradle
arapte: Added `relOpt = "-${buildTimestamp}" `

build.properties
arapte: Added a comment to describe the use of -, with example of -metal and -d3d12.

arapte:
Why we chose the name `jfx.experimental.release.suffix`.
The JavaFX version string contains the value of `jfx.release.suffix` i.e  `-ea` as a suffix. The new variable would be doing same, so it is also a suffix.
`24-ea` would be now `24-metal`

and use of `experimental` is justified by @nlisker as:
> I would have went with jfx.experimental.name :) because the sandbox has experimental features, as opposed to other features that are directly integrated into the main stream.

`jfx.release.suffix` **:** JavaFX relase version suffix for early access releases i.e `-ea`
`jfx.experimental.release.suffix` **:** JavaFX relase version suffix for experimental releases.

With this understanding, I am happy with the current name :) but shall update the PR we finalize any other name.

kevinrushforth:
I'd rather have the property be something without `suffix` for the reasons Johan and I mentioned.

Even though it's a little longer, I like a combination of what Johan and Nir suggested:

`jfx.experimental.feature.name` -- The name of the experimental feature

No one will ever specify this on the command line or have to type it, so having it be more explicit is good, even though the name is a little long.

You can say something like:

Set this property to the name of the experimental feature for your branch. Do not include a dash in the name. For example, if you have a sandbox branch named `myfeature` you might set `jfx.experimental.feature.name=myfeature`.

arapte:
> `jfx.experimental.feature.name` -- The name of the experimental feature

Thanks @kevinrushforth , Updated the property name as suggested.

build.gradle
```diff
@@ -696,13 +696,14 @@ if (HUDSON_JOB_NAME == "not_hudson") {
...
+        if (jfxExperimentalFeatureName != "") {
+            fail("An experimental feature branch should not be used for FCS build")
         }
         relSuffix = ""
     } else {
-        relSuffix = jfxExperimentalReleaseSuffix != "" ?
-                        jfxExperimentalReleaseSuffix : jfxReleaseSuffix;
+        relSuffix = jfxExperimentalFeatureName != "" ?
+                        "-${jfxExperimentalFeatureName}" : jfxReleaseSuffix;
+        relOpt = "-${buildTimestamp}"
```
kevinrushforth, line 706: This will unconditionally add a timestamp to _all_ EA builds of JavaFX, not just experimental builds. Changing the version string for EA builds seems out of scope for this PR.

```suggestion
        if (jfxExperimentalFeatureName != "") {
            relSuffix = "-${jfxExperimentalFeatureName}"
            relOpt = "-${buildTimestamp}"
        } else {
            relSuffix = jfxReleaseSuffix
        }
```

kevinrushforth (commented):
Looks good with one suggested change.

build.gradle
arapte: Thanks @kevinrushforth , Included the change.

kevinrushforth:
Looks good now. @johanvos this should be ready for your review.

johanvos (approved):

PR #1640, opened 2024-11-16
8341687: Memory leak in TableView after interacting with TableMenuButton

Maran23:
There are multiple issues in the `TableMenu` logic that result in a memory leak.

The following problems are now fixed with this PR:
- The listener, that is registered to the `col.visibleProperty()` was not weak nor was it ever unregistered
   - The fix is to do pretty much the same that was already done with the `col.textProperty()` listener
- The `col.visibleProperty()` and `col.textProperty()` where added again and again and again to the column when the columns changed (which happens when toggling the visibility).
  - The fix is to add the listeners once - when the `MenuItem` is created. This way, when the `TableMenu` is rebuild and the cached `MenuItem` is used, the whole listener logic is not run again for the column

---------
### Progress

### Issue
 * [JDK-8341687](https://bugs.openjdk.org/browse/JDK-8341687): Memory leak in TableView after interacting with TableMenuButton (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1640/head:pull/1640` \
`$ git checkout pull/1640`

`$ git checkout pull/1640` \
`$ git pull https://git.openjdk.org/jfx.git pull/1640/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1640`

`$ git pr show -t 1640`

</details>

<a href="https://git.openjdk.org/jfx/pull/1640.diff">https://git.openjdk.org/jfx/pull/1640.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1640#issuecomment-2480251159)
</details>

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -158,6 +158,14 @@ public class TableHeaderRow extends StackPane {
         }
     };
 
+    private InvalidationListener visibleColumnListener = observable -> {
+        TableColumnBase<?, ?> column = (TableColumnBase<?, ?>) ((BooleanProperty) observable).getBean();
+        CheckMenuItem menuItem = columnMenuItems.get(column);
+        if (menuItem != null) {
+            menuItem.setSelected(column.isVisible());
+        }
+    };
+
     private final ChangeListener<Boolean> cornerPaddingListener = (obs, ov, nv) -> updateCornerPadding();
 
     private final WeakInvalidationListener weakTableWidthListener =
@@ -175,10 +183,11 @@ public class TableHeaderRow extends StackPane {
     private final WeakInvalidationListener weakColumnTextListener =
             new WeakInvalidationListener(columnTextListener);
 
+    private final WeakInvalidationListener weakVisibleColumnListener =
+            new WeakInvalidationListener(visibleColumnListener);
+
     private final WeakChangeListener<Boolean> weakCornerPaddingListener =
             new WeakChangeListener<>(cornerPaddingListener);
-
-
 
     /* *************************************************************************
      *                                                                         *
@@ -580,9 +589,7 @@ public class TableHeaderRow extends StackPane {
         CheckMenuItem item = columnMenuItems.remove(col);
         if (item != null) {
             col.textProperty().removeListener(weakColumnTextListener);
-            item.selectedProperty().unbindBidirectional(col.visibleProperty());
-
-            columnPopupMenu.getItems().remove(item);
+            col.visibleProperty().removeListener(weakVisibleColumnListener);
         }
 
         if (! col.getColumns().isEmpty()) {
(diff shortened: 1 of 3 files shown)


modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
```diff
@@ -627,25 +634,29 @@ private void createMenuItem(TableColumnBase<?,?> col) {
         if (item == null) {
             item = new CheckMenuItem();
             columnMenuItems.put(col, item);
+
+            item.setSelected(col.isVisible());
+
+            final CheckMenuItem _item = item;
+            // fake bidrectional binding (a real one was used here but resulted in JBS-8136468)
```
andy-goryachev-oracle, line 641: that's JDK-8136468

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TableViewTableHeaderRowTest.java
```diff
@@ -225,6 +228,70 @@ void testOverriddenColumnPopupMenu() {
         assertNull(columnPopupMenu);
     }
 
+    /**
+     * Tests that re-setting the same columns does not cause memory leaks.
+     * See also: <a href="https://bugs.openjdk.org/browse/JDK-8341687">JDK-8341687</a>.
```
andy-goryachev-oracle, line 233: technically, this test does not test for memory leak - it seems to test the fact that the listener gets disconnected.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TableViewTableHeaderRowTest.java
```diff
@@ -225,6 +228,70 @@ void testOverriddenColumnPopupMenu() {
...
+
+        // The count should be the same still.
+        for (TableColumn<String, ?> column : tableView.getColumns()) {
+            assertEquals(3, getVisibilityListenerCount(column));
+        }
+    }
+
+    /**
+     * Tests that toggling the column visibility does not cause memory leaks.
+     * See also: <a href="https://bugs.openjdk.org/browse/JDK-8341687">JDK-8341687</a>.
```
andy-goryachev-oracle, line 261: same comment - does not test for memory leak.  the memory leak test might involve `JMemoryBuddy` - see for example `SystemMenuBarTest`.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableViewTableHeaderRowTest.java
```diff
@@ -226,6 +229,70 @@ void testOverriddenColumnPopupMenu() {
         assertNull(columnPopupMenu);
     }
 
+    /**
+     * Tests that re-setting the same columns does not cause memory leaks.
+     * See also: <a href="https://bugs.openjdk.org/browse/JDK-8341687">JDK-8341687</a>.
```
andy-goryachev-oracle, line 234: same, does not test for leaks.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TableViewTableHeaderRowTest.java
Maran23: I tried, but it isn't that easy - since the memory is not leaked because we keep some references we should not keep - rather we are adding listener again and again and again. So I did not managed to make a test with `JMemoryBuddy`, but can prove the point with the amount of listener instead.

andy-goryachev-oracle: I think this is an acceptable explanation as it tests the root cause of the leak.

Do you think changing the skin would allow us to create a test for the memory leak specifically.  Maybe some variation of `SkinMemoryLeakTest` ?

Maran23: Found a way to test it with `JMemoryBuddy`. Kept the listener tests as well, since they also prove the point.

andy-goryachev-oracle (approved):
thank you for adding the memory leak test!
the tests fail in the master, and pass with the fixes in this PR.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableHeaderRow.java
Maran23: maybe we we should do a small cleanup ticket where we replace all `JBS-` and `RT-` references with the `JDK-` one? A quick check led to around 120 occurrences.

andy-goryachev-oracle: and maybe fix the RT- references as well, though it might be a bit harder because a  lookup is required.

Maran23: That is correct. I still think it's worth it, especially since not everyone knows the 'trick' of pasting the `RT-` entry into the bug tracker to get the corresponding `JDK-` entry (in fact, I only learned that this year 😄)

andy-goryachev-oracle: I'd review that!  The only question is how to do RT- efficiently, so it can be easily reviewed (without going through manually copying and pasting every id)...

kevinrushforth: It would be pretty easy to do it with a JIRA REST query: Given an input list of `RT-nnnn` bug IDs, produce the corresponding output list of `JDK-mmmmmmm` bug IDs. I'll volunteer to do this part of it.

Someone could then easily turn that into a script using `sed` to do the replacement.

kevinrushforth: I filed [JDK-8344899](https://bugs.openjdk.org/browse/JDK-8344899) to track the cleanup and attached a `grep.log` file of all occurrences of `"RT-[0-9]+"` and a `rt2jdk.txt` file with the mapping from `RT-nnnn` to `JDK-mmmmmmm`.

PR #1644, opened 2024-11-22
8185887: TableRowSkinBase fails to correctly virtualize cells in horizontal direction

Maran23:
This PR fixes the horizontal virtualization performed in the `TableRowSkinBase`, which significantly improves performance with many columns (and thus many cells). Scrolling up and down as well as scrolling left and right feels much more snappy.

In order to do that, there are multiple things needed:
- the `isColumnPartiallyOrFullyVisible` needs to be fixed to take the `VirtualFlow` into account, not the `TableView`. Since rows are inside the `VirtualFlow`, we need to use the width from there
  - The wrong implementation right now leads to [JDK-8276326](https://bugs.openjdk.org/browse/JDK-8276326), where if you scroll to the right with many columns, cells on the left start to be empty (since they are removed when they shouldn't be)
  - It also does not help performance when scrolled to the left, as the right cells are not removed (only when scrolling to the right, cells on the left are removed)
- To improve performance, `isColumnPartiallyOrFullyVisible` was refactored to take in everything that is needed directly, without reiterating through all columns
- As before, the `TableRow` adds or removes cells that are visible or not. 
Note that this is only done when a fixed cell size is set. 
The reason for that is that we always know the row height. If not set, we need all cells so we can iterate over them to get the max height. I'm not sure if this can be improved, but this is not the goal of this PR and can be checked later
- The other issue mentioned in [JDK-8276326](https://bugs.openjdk.org/browse/JDK-8276326) happens only when a fixed cell size is set (empty rows). This is related and also fixed:
  - Cells start to be empty when scolling around a lot. This is because cells that are in the pile (ready to be reused) will not receive any layout requests (`requestLayout`) while all cells in the viewport will receive them. As soon as they are reused, they are visually outdated and not updated, leading to empty cells
  Fix is to request layout to those cells as well, and as soon as they are reused, they will layout themself
  - This has revealed another error: Cells that are not used anymore (added to the pile and invisible) are STILL inside the `VirtualFlow` sheet (as children). This will cause them to actually do the layout when requested, and especially when the height of the `TableView` changed drastically (e.g. from 50 visible cells to just 10), we have 40 cells laying around, receiving the layout request I added to fix [JDK-8276326](https://bugs.openjdk.org/browse/JDK-8276326), as mentioned above
  
  The fix is to remove those cells from the viewport when not needed anymore.
  NOTE: The `VirtualFlow` does a lot of 'clean up everything and decide which cells are visible', so I chose a performant fix (instead of removing all cells and readding them again)
  NOTE2: This makes the logic to make cells invisible and visible again obsolete. This was there so those unused cells laying around in the `VirtualFlow` does not receive any Mouse or KeyEvents. This can be cleaned up in a follow up PR

Considerations:
-
- I decided to not do more logic like batching all cells that should be removed (`removeAll`), as the scenario that mostly happens is:
  - You start scrolling to the right, we add and remove cells one by one
  - Therefore, we mostly do not remove or add multiple cells at once anyway
- Other bigger refactoring for performance can be done in a follow up (if we find things, I do not want to blow up this PR)

Testing:
-
- I added many tests to cover all possible scenarios, including:
  - Reordering a column
  - Decrease/Increase column width
  - Adding and removing columns
  - Scrolling around and checking, that no empty cells are around ([JDK-8276326](https://bugs.openjdk.org/browse/JDK-8276326))
- I did lots of testing on my own and also on JavaFX applications. I found no regression
- Some tests need to be adjusted since the fix also improves `TreeTableRow` code, which before added `LabeledText` sometimes (from `LabeledSkinBase`), although it was not needed, leading to code that need to count that particular node in, in some tests
- Remove ` VirtualFlowTestUtils.BLOCK_STAGE_LOADER_DISPOSE` which was hacky and replaced it with the normal `StageLoader` mechanism (as we use pretty much everywhere else)

/issue add JDK-8276326

---------
### Progress

### Issues
 * [JDK-8185887](https://bugs.openjdk.org/browse/JDK-8185887): TableRowSkinBase fails to correctly virtualize cells in horizontal direction (**Bug** - P3)
 * [JDK-8276326](https://bugs.openjdk.org/browse/JDK-8276326): Empty Cells in TableView supposedly after using setFixedCellSize() (**Bug** - P4)
 * [JDK-8252566](https://bugs.openjdk.org/browse/JDK-8252566): TreeTableView: broken row layout for fixedCellSize (**Bug** - P3)
 * [JDK-8346824](https://bugs.openjdk.org/browse/JDK-8346824): Collapsing tree view causes rendering issues (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1644/head:pull/1644` \
`$ git checkout pull/1644`

`$ git checkout pull/1644` \
`$ git pull https://git.openjdk.org/jfx.git pull/1644/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1644`

`$ git pr show -t 1644`

</details>

<a href="https://git.openjdk.org/jfx/pull/1644.diff">https://git.openjdk.org/jfx/pull/1644.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1644#issuecomment-2494177779)
</details>


kevinrushforth:
We will need to check that this doesn't impact accessibility, since there are some "interesting" cases where cells are requested and a layout pass is done even when they aren't visible.

@arapte As I recall, you were looking into VirtualFlow in connection with an a11y at one point. Can you test this?

kevinrushforth:
@johanvos You may want to take a look since this touches VirtualFlow.

Maran23:
/issue add JDK-8252566

same issue as https://bugs.openjdk.org/browse/JDK-8276326, but for the `TreeTableView`

Maran23:
> We will need to check that this doesn't impact accessibility, since there are some "interesting" cases where cells are requested and a layout pass is done even when they aren't visible.

Good point! Checking `getPrivateCell`, I wonder why we do not use the `accumCell` here, we rather add a cell into the sheet (although it is not visible) and return it. With the changes made in this PR, this cell will be cleaned up later. So worth checking!

arapte:
> @arapte As I recall, you were looking into VirtualFlow in connection with an a11y at one point. Can you test this?

I tested the PR changes with a few a11y scenarios, and did not observe any issues.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
```diff
@@ -229,30 +213,19 @@ public final DoubleProperty indentProperty() {
...
+        Node disclosureNode = getDisclosureNode();
+        if (disclosureNode != null && disclosureNode.getParent() == null) {
+            disclosureNodeDirty = true;
         }
 
-        Node disclosureNode = getDisclosureNode();
-        if (disclosureNode != null && disclosureNode.getScene() == null) {
+        if (disclosureNodeDirty) {
             updateDisclosureNodeAndGraphic();
+            disclosureNodeDirty = false;
```
andy-goryachev-oracle, line 228: shouldn't `disclosureNodeDirty = false;` be moved to `updateDisclosureNodeAndGraphic();` ?
here and in L257, L216

andy-goryachev-oracle (commented):
Testing on macOS 14.7.1 with the latest Monkey Tester (where I've added 200 columns case for Tree/TableView, 
https://github.com/andy-goryachev-oracle/MonkeyTest )
looks good.  

It feels like the horizontal scrolling has improved a bit, the vertical exhibits the same slight hiccups as with the master branch.

kevinrushforth:
> I've got an exception on the TreeTableView page in the monkey tester after clicking a cell and then pressing control+command+RIGHT_ARROW with the VoiceOver on:

Is this reproducible? Does it happen without this fix? If this is a regression then it needs to be fixed.

andy-goryachev-oracle:
Created [JDK-8235989](https://bugs.openjdk.org/browse/JDK-8345536) for the a11y exception issue.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
```diff
@@ -160,6 +137,13 @@ private void updateCachedFixedSize() {
             if (t != null) {
                 fixedCellSize = t.getFixedCellSize();
                 fixedCellSizeEnabled = fixedCellSize > 0.0;
+
+                if (fixedCellSizeEnabled) {
```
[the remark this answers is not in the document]
Maran23: Not clash, but there will be merge conflicts that are easy to solve

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
Maran23: Could be an idea. I will check. I just tried to not make to many changes, but this might be an easy one.

Maran23:
> It feels like the horizontal scrolling has improved a bit, the vertical exhibits the same slight hiccups as with the master branch.

Yes, especially if you try with many columns (> 100), scrolling up and down is much faster (especially when scrolled completely to the left). When you scroll completely to the right, it is just a little bit faster than master.

I have also some ideas how to improve vertial scrolling with many items, but for another day. :)

Maran23:
> I tested the PR changes with a few a11y scenarios, and did not observe any issues.

Thank you for testing!
Out of interesting, is it 'enough' to test with the `Narrator` application on Windows 11?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
Maran23: Checked the code, that is a safe improvement, so will do.

Maran23:
> By the way, a Tree/TableView with 500 columns is barely usable - locks up the UI giving me a macOS spinning beach ball quite often. Not that it's a reasonable use case. 200 is ok. May be the 500 case can be used for profiling?

With fixed cell size, it seems okay to me. How is your table configured?

> Speaking of the vertical scrolling - I could not use the idea implemented by the VirtualFlow in the RichTextArea as it won't work in the case of very large models.

There is a lot of stuff that can improved in the `VirtualFlow`. Most of the modern table components all use fixed cells or have some kind of dynamic height calculation logic. They only rely on calculating the size themself when it really needs to (e.g. the developer turning it on).

You can for example check AG-Grid, which is a pretty nicely implemented table that is very fast and has a nice API design:
https://www.ag-grid.com/javascript-data-grid/row-height/
They also do not recommend setting the auto height, and also explain the side effects (cells and scroll bar jumping).

We only have the fixed cell size implemented, but it is not the default (and too late to change). This helps a lot since the math is very easy then. 
For something like the `RichTextArea`, this might be more complex, so I can see why the `VirtualFlow` does not fit. It is more complex as a Table, where you most likely know the height in advance.

johanvos:
I checked the changes in VirtualFlow and those are ok.
In general, we need a better approach to test and quantify performance, but realistically that should not be done as part of this PR. There are some tests in ListViewTest and TableViewTest that are based on the number of invocations of updateItem() (which is in most cases a good indication of performance), but they are mainly ad-hoc.

In absence of those tests, I believe this PR is a good step in the good direction.

andy-goryachev-oracle (approved):
looks good, thanks!

scrolling performance is not an issue: the horizontal one has improved, and the vertical one is on par with the master.

andy-goryachev-oracle:
> I believe this PR is a good step in the good direction.

@johanvos would you approve this PR then?

johanvos:
> > I believe this PR is a good step in the good direction.
> 
> @johanvos would you approve this PR then?

I only looked into the changes in VirtualFlow, so I can't say anything about the other changes -- and approving a PR means the whole PR has been reviewed.
I can do a review later, but tbh, the TableRowSkin is a beast I never managed to tame :)

andy-goryachev-oracle:
Thank you for the partial review nevertheless!  Every little bit helps...

johanvos:
@jperedadnr can you review?

henrykdz:
Hey Guys, i am just new here, pls anyone help with my observation: 
Bug Report: JavaFX TreeView - Incorrect Cell Index Update on Expand/Collapse
Issue:
When expanding or collapsing a TreeItem in a TreeView, an incorrect cell index is updated. Specifically, a cell index mirrored vertically around the center of the visible area is updated instead of the correct cell. This indicates a flaw in the cell index update logic or the visible cell list update mechanism.

This issue is particularly noticeable when cells are styled using CSS, as the incorrect update results in a visual glitch or flicker. Even without CSS, the problem manifests as a subtle graphical irregularity due to the rapid update process.

Steps to Reproduce:

Create a TreeView with enough items to enable scrolling.
Scroll to make a specific range of indices visible (e.g., indices 4 to 13).
Apply distinct CSS styles to cells to make updates visually apparent.
Expand or collapse a TreeItem within the visible range (e.g., index 5).
Observe that the mirrored index (e.g., index 12) is updated instead of the correct index.
Observed Behavior:
When a TreeItem at index i is expanded/collapsed, the index updated is:

Mirrored Index = 2 ⋅ Center − 𝑖

Examples:
Expanding/collapsing index 5 incorrectly updates index 12.
Expanding/collapsing index 10 incorrectly updates index 3.
Expected Behavior:
Only the cell corresponding to the expanded/collapsed TreeItem should be updated.

Impact:
Visual inconsistencies and graphical glitches in the TreeView.
Subtle flickering or graphical irregularities during expand/collapse actions, especially noticeable with styled cells.
Suggested Cause:
A miscalculation or incorrect implementation in the logic for updating cell indices or the visible cell list during expand/collapse operations.

Environment:
Observed in JavaFX versions 16 through 23.
Present across different platforms and configurations.

andy-goryachev-oracle:
@henrykdz : 
thank you for the bug report!  I can help you to file a bug.  Two questions:
1. is the issue you describe still present when tested with the changes in this PR?  Or is it a separate issue?
2. can you create a simple SCCE (self-contained short example), see for instance https://github.com/andy-goryachev-oracle/Test/blob/main/src/goryachev/bugs/TreeView_RenderingFailure_8346824_8347357.java

henrykdz:
https://github.com/user-attachments/assets/6a07c5f9-7ac3-4148-b1db-07957af06e7a



> @henrykdz : thank you for the bug report! I can help you to file a bug. Two questions:
> 
> 1. is the issue you describe still present when tested with the changes in this PR?  Or is it a separate issue?
> 2. can you create a simple SCCE (self-contained short example), see for instance https://github.com/andy-goryachev-oracle/Test/blob/main/src/goryachev/bugs/TreeView_RenderingFailure_8346824_8347357.java


https://github.com/user-attachments/assets/b0d65fa0-0556-4687-a80b-1168866324f7

henrykdz:
I just updated to JavaFX 23 using Azure Zulu with FX, hope it helps anyhow!

henrykdz:
Did you recognized?

andy-goryachev-oracle:
@henrykdz : 
thank you for providing SCCE.

I've created https://bugs.openjdk.org/browse/JDK-8347464 to track the new issue.  The new issue is not related to this PR, so sorry @Maran23 for adding unrelated comments.

In the future, it's probably better to file the bug via normal channel described in https://bugs.java.com/bugdatabase/faq or by sending a message to [openjfx-dev@openjdk.org](mailto:openjfx-dev@openjdk.org)

andy-goryachev-oracle:
@henrykdz :
could you please attach the video clip to https://bugs.openjdk.org/browse/JDK-8347464 ?

modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewVirtualizationTest.java
```diff
@@ -0,0 +1,549 @@
...
+        }
+    }
+
+    @Test
+    void testHorizontalVirtualizationAddFirstColumn() {
+        TableColumn<String, String> tableColumn = new TableColumn<>("NEW");
+        tableColumn.setPrefWidth(50);
+        tableView.getColumns().addFirst(tableColumn);
+
+        // Needs a double pulse so that the viewport breadth is correctly calculated.
```
johanvos, line 177: This worries me a bit (the double pulse happens on a few places in the new tests). I would expect the platform to deal with is, so when a test requires 2 explicit pulse requests, it sounds like an issue.
Why can't this be done with a single request? (and then waiting until the pulse and pending runnables on the FX Thread have finished)?

Maran23: Unfortunately I don't think there is any other way. The `VirtualFlow` needs two pulses (in real life applications) as the first time, the layout is not yet correct for some cases (e.g. for `No ScrollBar` -> `ScrollBar`). I even used some watchpoints to confirm this behavior.

You can see the same thing in the `VirtualFlowTest.setUp` method, which initializes the flow and does two `pulse` calls after. So I think this is a premature problem, which probably can be fixed, but more refactorings/optimzations are needed (some of which I want to file a PR when I have more time). 
Maybe at some point we can completely eliminate this problem, but I don't think I can do that here in this PR (unless you have an idea, which is very much welcome!)

jperedadnr: Maybe the name `pulse` in the test you point out is misleading, since it actually calls `flow.layout()`, and two layout passes is not the same as two pulses.

jperedadnr (commented):
I tested the PR, and it improves the scrolling performance in some cases, without noticeable side effects. 
I left some comments.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewVirtualizationTest.java
Maran23: That is right, but the `pulse` call will call `layout`, which will call `layoutChildren` (when needed). I could also just call `layout` directly, but want to keep it the same way as the actual Application would behave.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
```diff
@@ -229,30 +213,19 @@ public final DoubleProperty indentProperty() {
...
-        }
+        disclosureNodeDirty = false;
     }
 
     /** {@inheritDoc} */
     @Override protected void layoutChildren(double x, double y, double w, double h) {
-        if (disclosureNodeDirty) {
-            updateDisclosureNodeAndGraphic();
-            disclosureNodeDirty = false;
+        Node disclosureNode = getDisclosureNode();
```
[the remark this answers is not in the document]
Maran23: This is more likely an 'accidential' fix. I changed this code because the code in the super class changed (e.g. `updateCells` and `updateChildren`). So while I see your point, if possible I would not try to cherrypick this out (if even possible) and keep it in here. 
Good point, I will check if I can write a test for it.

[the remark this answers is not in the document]
Maran23: EDIT: The issue is about the `TreeView` specifically, so this code is not related to the issue.
The fix is in the `VirtualFlow`, which I did for the tables, which also fixes the issue for the `TreeView`.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewVirtualizationTest.java
andy-goryachev-oracle: Relying on two pulses may cause flicker (I had a similar issue with the RichTextArea).

The solution is to re-compute the layout in layoutChildren() when needed (i.e. when scroll bar status changes).

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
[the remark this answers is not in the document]
Maran23: EDIT2: I wrote a test, that fails before and succeeds after for that specific problem!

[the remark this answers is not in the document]
andy-goryachev-oracle: I second @jperedadnr concern about combining separate issues into one PR.  Small focused PRs will get tested, reviewed, and integrated almost always faster than the big complicated ones.

andy-goryachev-oracle:
Can attest to the fact that two pulses are indeed needed when testing with the monkey tester (unconstrained resize policy, columns with "all constraints", 10 rows; resize to the point where both scrollbars are visible but very close to where the SB are not needed, then increase the height slowly).

The good news is that I could not get it to flicker (at least on macOS, with 1:1 scale).  I see SB disappears, then another ("two pulses").  Visually, it's the same behavior as with the master branch.

The not-so-good news is that there is a configuration when the HSB remains, but once you click on it it disappears.  On the master branch, the HSB remains even though it's not needed:

Screenshot 2025-01-16 at 12 50 25

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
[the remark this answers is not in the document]
Maran23: Completely agree, but in this case, the fix for [JDK-8276326](https://bugs.openjdk.org/browse/JDK-8276326) is also the same for [JDK-8346824](https://bugs.openjdk.org/browse/JDK-8346824) - that is the `VirtualFlow` changes.
It is not really possible to separate those issues. 
What I could try is to cherrypick the changes from the `VirtualFlow`, leaving out the virtualization optimizations. Although this was not the original point here.

The reason those are together is that they are very closely related. Fixing the virtualization without the `VirtualFlow` will still break applications and tests because you can somewhat randomly get empty cells.

Fixing the `VirtualFlow` first might work, but if you scroll enough you will get empty cells because of broken virtualization.
So this PR got bigger in order to have a working virtualization and also scrolling behavior with no empty cells.
If it really helps, I can split them up, as mentioned above.

Maran23:
> The not-so-good news is that there is a configuration when the HSB remains, but once you click on it it disappears (might be hard to reproduce). In both this PR and with the master branch, the HSB remains even though it's not needed:

When I debugged the double pulse, I found out that the viewport breadth changed a little bit between the first and the second call. Which might also change the HSB. 
My guess is that this issue and the double pulse are related, I already thought about this when I found this out first when doing this PR. So yes, worth investigating at one point.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeTableRowSkin.java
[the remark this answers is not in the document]
andy-goryachev-oracle: probably not worth it at this point, I think.  We just need to be extra careful during the review to consider all the scenarios involved.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TableViewVirtualizationTest.java
Maran23: Good to know, we should document your solution in more detail (and with the concrete fix -> code snippet from the RTA) together with the HSB bug as a new ticket probably? Then we may already have a potential solution with code, and of course a ticket that tracks the bug, if we do not have already (I think I saw a similar bug already, but not 100% sure).

andy-goryachev-oracle: VFlow LL1529-1555

```
        // scroll bars
        boolean vsbVisible = useContentHeight ?
            false :
            (topCellIndex() > 0) ?
                true :
                (arrangementHeight + contentPaddingTop + contentPaddingBottom) > viewPortHeight;

        if (vsbVisible != vscroll.isVisible()) {
            vscroll.setVisible(vsbVisible);
            // do another layout pass with the scrollbar updated
            layoutCells();
            return;
        }
        if (vsbVisible) {
            width -= vsbWidth;
        }

        boolean hsbVisible = (wrap || useContentWidth) ?
            false :
            (unwrappedWidth + contentPaddingLeft + contentPaddingRight) > viewPortWidth;

        if (hscroll.isVisible() != hsbVisible) {
            hscroll.setVisible(hsbVisible);
            // do another layout pass with the scrollbar updated
            layoutCells();
            return;
        }
```


https://github.com/andy-goryachev-oracle/jfx/blob/d07d408fa79e25a02b0d5e2f9aeb3990a7136077/modules/jfx.incubator.richtext/src/main/java/com/sun/jfx/incubator/scene/control/richtext/VFlow.java#L1529

Maran23: Thanks. That makes sense IMO. Can probably also be applied to the `VirtualFlow`.
Just need to be careful that we never can run in a stack overflow ofc. :)

andy-goryachev-oracle (approved):
The changes and testing LGTM.

henrykdz:
> @henrykdz : could you please attach the video clip to https://bugs.openjdk.org/browse/JDK-8347464 ?

I've deleted it from here, didn't knowing it would cause this, sry. i guess i cannot connect to the old url again?

mstr2 (approved):
I tested this patch, which improves performance with many columns. The code changes look good.

Regarding the discussion around requiring two layout passes: there's an old PR that attempted to fundamentally change that (iterative layout), which I think might solve a lot of these kinds of problems.

PR #1616, opened 2024-10-29
8309381: Support JavaFX incubator modules

kevinrushforth:
This PR add the necessary support for [JavaFX Incubator Modules](https://github.com/kevinrushforth/jfx/blob/jfx.incubator/INCUBATOR-MODULES.md). It includes the following:

1. Changes to the build scripts `build.gradle` and `settings.gradle` to document where to add your incubator module. Also added `jlink` flags to not resolve incubator modules by default and warn when resolving (same as is done for JDK incubator modules).
2. A utility class to produce warnings when first using an incubator module.
3. Changes to `javafx.base/module-info.java` to document where to add the needed qualified exports to your module to access the utility class in 2.

See PR #1617 for a sample module that is built on top of this PR's source branch.

This is almost ready to review. I will take it out of Draft in the next day or so.

/reviewers 2 reviewers
/csr

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change requires CSR request [JDK-8344644](https://bugs.openjdk.org/browse/JDK-8344644) to be approved

### Issues
 * [JDK-8309381](https://bugs.openjdk.org/browse/JDK-8309381): Support JavaFX incubator modules (**Enhancement** - P3)
 * [JDK-8344644](https://bugs.openjdk.org/browse/JDK-8344644): Support JavaFX incubator modules (**CSR**)

### Reviewers
 * [Andy Goryachev](https://openjdk.org/census#angorya) (@andy-goryachev-oracle - **Reviewer**)
 * [Ambarish Rapte](https://openjdk.org/census#arapte) (@arapte - **Reviewer**) Review applies to [1d262fb3](https://git.openjdk.org/jfx/pull/1616/files/1d262fb3eb14f153f927652fb31c75cbc819249d)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1616/head:pull/1616` \
`$ git checkout pull/1616`

Update a local copy of the PR: \
`$ git checkout pull/1616` \
`$ git pull https://git.openjdk.org/jfx.git pull/1616/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 1616`

View PR using the GUI difftool: \
[...]

PR #1646, opened 2024-11-22
Draft: 8309381: Support JavaFX incubator modules (v2)

kevinrushforth:
This is a Draft PR of a slightly modified set of build changes to be able to add incubator modules without modifying `build.gradle` in several places. See [this comment in PR #1616](https://github.com/openjdk/jfx/pull/1616#issuecomment-2494645757) for more information.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1646/head:pull/1646` \
`$ git checkout pull/1646`

`$ git checkout pull/1646` \
`$ git pull https://git.openjdk.org/jfx.git pull/1646/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1646`

`$ git pr show -t 1646`

</details>

<a href="https://git.openjdk.org/jfx/pull/1646.diff">https://git.openjdk.org/jfx/pull/1646.diff</a>

</details>


settings.gradle
```diff
@@ -23,7 +23,16 @@
  * questions.
  */
 
-include "base", "graphics", "controls", "swing", "swt", "fxml", "jsobject", "web", "media", "systemTests"
+include("base")
```
andy-goryachev-oracle, line 26: what's the difference between these two notations?

README-incubator.md
```diff
@@ -0,0 +1,166 @@
...
+To add an incubator module, do the following:
+
+1. Add an entry for your module in `settings-incubator.gradle` and `build-incubator.gradle` in the designated place
+2. Create your module under `modules/jfx.incubator.YOURMODULENAME`, including your source code and test code as is done for other modules
+3. Add the needed build logic in `modules/jfx.incubator.YOURMODULENAME/project.gradle`
+
+Here is an example patch:
+
+```
+diff --git a/incubator-build.gradle b/incubator-build.gradle
```
andy-goryachev-oracle, line 18: the diffs are rather hard to read.

maybe instead we should provide a clean example, with the module name token in UPPERCASE that will be easy to replace.

the idea is to give the author of an incubator module a template that they can easily use, rather than forcing them to decipher the diffs.

example:

```
ext.incubatorProjectNames = [
     'incubator.INCUBATORMODULE'
]
```



what do you think?

build.gradle
```diff
@@ -4039,7 +4060,18 @@ project(":systemTests") {
...
-    def dependentProjects = [ 'base', 'graphics', 'controls', 'media', 'jsobject', 'web', 'swing', 'fxml' ]
+    def dependentProjects = [
+        'base',
+        'graphics',
+        'controls',
+        'media',
+        'jsobject',
+        'web',
+        'swing',
+        'fxml'
```
andy-goryachev-oracle, line 4071: +1 for placing each item on its own line

andy-goryachev-oracle (commented):
I have mixed feelings:

On plus side, this change separates incubators from the main code.

On the other side, it might make it more difficult to make changes to build.gradle because

1. the separation hides the dependency
2. it makes it more difficult to grep since now the logic is scattered between multiple files

I wonder if it might be better to split declarative part of the build.gradle from the code/logic part.  Then any subsequent changes would go into declarative part (projects, flags, dependencies, etc), while logic remains stable and out of view.
That would be a much larger project though.

Overall, I think I would rather not separate the incubators' build files.

PR #1653, opened 2024-11-29
8345261: Refactor the Dimension2D classes

nlisker:
A small refactoring of the Dimension classes.

* `com.sun.javafx.geom.Dimension` was removed and its uses were replaced by `com.sun.javafx.geom.Dimension2D`.
* `com.sun.javafx.geom.Dimension2D` became a record.
* `javafx.geometry.Dimension2D`: fields became `final`.

I'm not sure we need the implementation class at all considering we are free to use the public one.

---------
### Progress

### Issue
 * [JDK-8345261](https://bugs.openjdk.org/browse/JDK-8345261): Refactor the Dimension2D classes (**Enhancement** - P4)

### Reviewers
 * [John Hendrikx](https://openjdk.org/census#jhendrikx) (@hjohn - **Reviewer**) 🔄 Re-review required (review applies to [e1ce604a](https://git.openjdk.org/jfx/pull/1653/files/e1ce604ae88c002de0a3bd39963c3a692c36510a))

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1653/head:pull/1653` \
`$ git checkout pull/1653`

`$ git checkout pull/1653` \
`$ git pull https://git.openjdk.org/jfx.git pull/1653/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1653`

`$ git pr show -t 1653`

</details>

<a href="https://git.openjdk.org/jfx/pull/1653.diff">https://git.openjdk.org/jfx/pull/1653.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1653#issuecomment-2508171658)
</details>


hjohn (approved):
Looks good

nlisker:
Thanks John. Because it's the weekend I'll wait until Monday evening with this.

kevinrushforth:
@azvegint Can you take a quick look at this since you added the Dimension class as part of your initial Wayland Robot fix?

azvegint:
These 3 classes are all different from each other. They use different types of variables to store data (`double`, `float`, `int`).

We have `com.sun.javafx.geom.Rectangle` which uses `int`, and the `com.sun.javafx.geom.Dimension` was specifically introduced as its pair which also uses `int`.
(just as we have a pair of `javafx.geometry.Rectangle2D` and `javafx.geometry.Dimension2D.java`).

Switching from `int` to `float` may have some side effects on stream processing due to the imprecise comparison.

So I would prefer to keep the `com.sun.javafx.geom.Dimension`.

nlisker:
Thanks @azvegint. If the `int`-based dimensions are coupled with `com.sun.javafx.geom.Rectangle` (and from the code, that's the only place they are used at), would you object to making `Dimension` an inner class of `Rectangle`? The `float`-based dimensions are used in other 2D shapes, so are more general purpose.

nlisker:
I also looked at the uses of `com.sun.javafx.geom.Rectangle` and in almost all cases I've looked at, its values are widened to floats later on. I'm not going to change anything there, but there is an indication that it's not the right tool for the job.

azvegint:
> would you object to making Dimension an inner class of Rectangle?

I am fine with it.

modules/javafx.graphics/src/main/java/com/sun/javafx/geom/Dimension2D.java
```diff
@@ -26,18 +26,6 @@
 package com.sun.javafx.geom;
 
 /**
- * The <code>Dimension2D</code> class is to encapsulate a width
- * and a height dimension.
- * <p>
+ * A 2D dimension object that contains a width and a height.
```
andy-goryachev-oracle, line 29: maybe mention that this class uses `float`s?

modules/javafx.graphics/src/main/java/com/sun/glass/ui/gtk/screencast/TokenStorage.java
```diff
@@ -365,13 +366,10 @@ static Set<TokenItem> getTokens(List<Rectangle> affectedScreenBounds) {
...
         // screens may have been moved while the token is still valid
-        List<Dimension> dimensions =
+        List<Dimension2D> dimensions =
                 affectedScreenBounds
                 .stream()
-                .map(rectangle -> new Dimension(
-                        rectangle.width,
-                        rectangle.height
-                ))
+                .map(rectangle -> new Dimension2D(rectangle.width, rectangle.height))
```
andy-goryachev-oracle, line 372: the largest positive integer that can be stored in a float exactly is ~16M.  we are fine here.

modules/javafx.graphics/src/main/java/com/sun/javafx/geom/Dimension2D.java
```diff
@@ -26,18 +26,6 @@
...
-    public float height;
-
-    public Dimension2D() { }
-
-    public Dimension2D(float w, float h) {
-        width = w;
-        height = h;
-    }
-}
+public record Dimension2D(float width, float height) {}
```
andy-goryachev-oracle, line 31: I wish this class was named differently to signify it's based on float...

nlisker:
You reviewed the PR before the requested changes were made :)
I'm going to move the `int` one into an inner class anyway.

nlisker:
@azvegint Please confirm that the change is acceptable.

modules/javafx.graphics/src/main/java/com/sun/javafx/geom/Dimension2D.java
nlisker: I changed the names of the non-public classes to indicate the types they use, it's just easier to see at a glance.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/gtk/screencast/TokenStorage.java
nlisker: Yes, I don't think we need all of them, but I was asked to keep the types.

modules/javafx.graphics/src/main/java/com/sun/javafx/geom/Dimension2Df.java
```diff
@@ -26,18 +26,9 @@
...
-    public float height;
-
-    public Dimension2D() { }
-
-    public Dimension2D(float w, float h) {
-        width = w;
-        height = h;
-    }
-}
+public record Dimension2Df(float width, float height) {}
```
andy-goryachev-oracle, line 34: is this class really needed?

Looks like it's only referenced from `Arc2D::setArc()` and `RectanglularShape::setFrame()`, as far as I can tell these two methods are never called.

Should all three be removed?

nlisker: I'm not familiar with these methods, but I did wonder why we need extra implementations in the initial comment.

andy-goryachev-oracle: looks like dead code left over from the original implementation.

I removed the class along with the two methods and jfx builds and runs (i.e. shows arcs and rectangles) just fine.  Eclipse IDE also has no issues.  I think we can remove these.

nlisker: I don't mind removing them in this PR if others agree.

mstr2:
> Thanks @azvegint. If the `int`-based dimensions are coupled with `com.sun.javafx.geom.Rectangle` (and from the code, that's the only place they are used at), would you object to making `Dimension` an inner class of `Rectangle`? The `float`-based dimensions are used in other 2D shapes, so are more general purpose.

In my opinion, inner classes are justified when the implementations are coupled, or when one thing only makes sense in the context of another thing. That doesn't seem to be the case here: `Dimension` is neither used in `Rectangle`, nor are their implementations coupled.

What this suggests to me is that either `Dimension` should remain its own top-level class, or if it is not useful there, be moved to where it _is_ used instead.

nlisker:
> > Thanks @azvegint. If the `int`-based dimensions are coupled with `com.sun.javafx.geom.Rectangle` (and from the code, that's the only place they are used at), would you object to making `Dimension` an inner class of `Rectangle`? The `float`-based dimensions are used in other 2D shapes, so are more general purpose.
> 
> In my opinion, inner classes are justified when the implementations are coupled, or when one thing only makes sense in the context of another thing. That doesn't seem to be the case here: `Dimension` is neither used in `Rectangle`, nor are their implementations coupled.
> 
> What this suggests to me is that either `Dimension` should remain its own top-level class, or if it is not useful there, be moved to where it _is_ used instead.

The int-based Dimension is only used *by* `Rectangle` even if in another class. I can couple their implementations by adding a `Rectangle#asDimension2D`, and this will be the only instantiation of this class.

hjohn:
> The int-based Dimension is only used by Rectangle even if in another class

I've taken another look.  Rectangle is not using this class at all.  I think it would be much better to give the GTK screencast logic its own little int/int record (not necessarily called anything like dimension).  Just define it package private or as an inner type of `TokenItem`.  This code really has no relation at all to the rest of FX (it is even GTK specific).  

Alternatively (which may be even better), avoid the creation of a dimension and just call `hasAllScreensOfSameSize(dimensions)` with the actual rectangles and have `TokenItem` figure it out using that information.

As for the `float` version of `Dimension2D`, I think we don't need this.  Just use the `double` one; everything in FX is doubles, and to be honest, the conversion we sometimes do to floats is a precision loss risk.

So, if it is up to me:
- Don't create any new dimension types; FX uses doubles, down casting to floats is not without risk.  Int variants are not needed in core FX code.
- Have TokenItem accept Rectangles in its `hasAllScreensOfSameSize` function.  I think you can change the logic to avoid creating a dimension class at all, or you can make a tiny `record ScreenSize(int w, int h)` defined as an inner class or even method class to do what it is currently doing.

PR #915, opened 2022-10-13
8260528: Clean glass-gtk sizing and positioning code

tsayao:
This cleans size and positioning code, reducing special cases, code complexity and size.

Changes:

- cached extents: 28, 1, 1, 1 are old defaults - modern gnome uses different sizes. It does not assume any size because it varies - it does cache because it's unlikely to vary on the same system - but if it does occur, it will only waste a resize event.
- window geometry, min/max size are centralized in `update_window_constraints`;
- Frame extents (the window decoration size used for "total window size"):
    - frame extents are received in `process_property_notify`;
    - removed quirks in java code;
    - When received, call `set_bounds` again to adjust the size (to account decorations later received);
- Removed `activate_window` because it's the same as focusing the window. `gtk_window_present` will deiconify and focus it.
- `ensure_window_size` was a quirk - removed;
- `requested_bounds` removed - not used anymore;
- `window_configure` incorporated in `set_bounds` with `gtk_window_move` and `gtk_window_resize`;
- `process_net_wm_property` is a work-around for Unity only (added a check if Unity - but it can probably be removed at some point)
- `restack` split in `to_front()` and `to_back()` to conform to managed code;

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issue
 * [JDK-8260528](https://bugs.openjdk.org/browse/JDK-8260528): Clean glass-gtk sizing and positioning code

### Reviewers
 * [Johan Vos](https://openjdk.org/census#jvos) (@johanvos - **Reviewer**)
 * [Kevin Rushforth](https://openjdk.org/census#kcr) (@kevinrushforth - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/915/head:pull/915` \
`$ git checkout pull/915`

Update a local copy of the PR: \
`$ git checkout pull/915` \
`$ git pull https://git.openjdk.org/jfx.git pull/915/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 915`

View PR using the GUI difftool: \
`$ git pr show -t 915`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
[...]

PR #1654, opened 2024-11-30
8344372: Setting width for TRANSPARENT Stage -> gtk_window_resize: assertion 'height > 0'

tsayao:
The bug happened when setting only width or height (not both) on a Stage that was oriented by the view size.

The fix just uses the previous value for width/height if not set.

The test class is designed to accept more test cases in the "mixed sizing" scenarios.

---------
### Progress

### Issue
 * [JDK-8344372](https://bugs.openjdk.org/browse/JDK-8344372): Setting width for TRANSPARENT Stage -&gt; gtk_window_resize: assertion 'height &gt; 0' (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1654/head:pull/1654` \
`$ git checkout pull/1654`

`$ git checkout pull/1654` \
`$ git pull https://git.openjdk.org/jfx.git pull/1654/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1654`

`$ git pr show -t 1654`

</details>

<a href="https://git.openjdk.org/jfx/pull/1654.diff">https://git.openjdk.org/jfx/pull/1654.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1654#issuecomment-2508987959)
</details>


kevinrushforth:

/reviewers 2

kevinrushforth (commented):
The fix works as expected.

The test will catch the failure, but fails for me almost always on one of my machines and very infrequently on another. There is a race condition in the test where the `onShown` event can be delivered before the initial stage size has been realized. I recommend sleeping for 500 msec after getting the `onShown` event and before changing the width of the Stage.

tests/system/src/test/java/test/javafx/stage/StageMixedSizeTest.java
```diff
@@ -96,7 +98,9 @@ private void createStage(BiConsumer<Stage, StackPane> beforeShown,
             }
 
             mainStage.setScene(new Scene(sp));
+
             mainStage.setOnShown(e -> {
+                Util.sleep(500);
```
kevinrushforth, line 103: We don't want to sleep on the FX application thread. I think you'll need to split this into separate `Runnable`s for separate calls to `Util::runAndWait` and have the sleeps between them.

tests/system/src/test/java/test/javafx/stage/StageMixedSizeTest.java
```diff
@@ -105,9 +109,9 @@ private void createStage(BiConsumer<Stage, StackPane> beforeShown,
             });
 
             mainStage.show();
+            Util.sleep(500);
         });
 
         Util.waitForLatch(showLatch, 5, "Stage failed to setup and show");
-        Util.sleep(500);
```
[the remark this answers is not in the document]
tsayao: Restored

tests/system/src/test/java/test/javafx/stage/StageMixedSizeTest.java
tsayao: I did it with a Timeline, is it ok?

kevinrushforth: That should be OK. I'll test it tomorrow.

kevinrushforth:
> On my end the test does pass on Ubuntu 22.04 with Xorg and with the fix. However, without the fix, the tests still succeed despite the GTK assertions showing up:

When I tested the earlier version yesterday, the test failed for me on Ubuntu 22.04 without the fix. I haven't tested the updated version, but if it no longer catches the error by failing without the fix, then something is wrong with the updated test.

kevinrushforth:
I see the same behavior as @lukostyra -- the current version of the test no longer catches the error. I took a quick look and didn't spot anything. The timeline looks correct; the sleep now looks redundant since you wait for 500ms in the timeline and then again after the latch.await, but that wouldn't cause the test to miss catching the error. Odd.

tsayao:
The problem is that JavaFx has the wrong size, so I moved it to a Robot test to do color checking at the position.

tsayao:
Updated the explanation of the tests above.

"Pre-submit tests" are failing. Is it "my fault"?

kevinrushforth:
> "Pre-submit tests" are failing. Is it "my fault"?

No, it looks like a transient glitch of some sort. I was going to suggest that you rerun them, and I see that you did.

tsayao:
I updated the tests to use Robot, using color validation. This ensures that the tests can verify whether the window has correctly shrunk as expected.

kevinrushforth:
I'll re-test when I get a chance.

lukostyra (approved):
Test works well and now it indeed fails when the fix is not there. LGTM

tests/system/src/test/java/test/robot/javafx/stage/StageMixedSizeTest.java
```diff
@@ -0,0 +1,110 @@
...
+            Scene scene = new Scene(pane);
+            testStage.setScene(scene);
+            testStage.setX(0);
+            testStage.setY(0);
+            testStage.setOnShown(e -> onShown.run());
+            testStage.show();
+        });
+    }
+
+
```
kevinrushforth, line 102: Minor: extra blank line can be removed

kevinrushforth (approved):
The updated test looks good. It is stable on both my systems and it reliably catches the error.

I left a couple minor suggestions and will reapprove if you decide to make changes.

tests/system/src/test/java/test/robot/javafx/stage/StageMixedSizeTest.java
tsayao: Done.

tests/system/src/test/java/test/robot/javafx/stage/StageMixedSizeTest.java
```diff
@@ -0,0 +1,110 @@
...
+        CountDownLatch latch = new CountDownLatch(1);
+        final int finalWidth = 200;
+
+        setupContentSizeTestStage(300, 300,
+                () -> doTimeLine(Map.of(500L, () -> testStage.setWidth(finalWidth),
+                                        1000L, latch::countDown)));
+
+        assertTrue(latch.await(TIMEOUT, TimeUnit.MILLISECONDS), "Timeout waiting for test stage to be shown");
+
+        runAndWait(() -> assertColorDoesNotEqual(BACKGROUND_COLOR, getColor(290, 100), TOLERANCE));
```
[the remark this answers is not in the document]
tsayao: Done.

tests/system/src/test/java/test/robot/javafx/stage/StageMixedSizeTest.java
```diff
@@ -0,0 +1,110 @@
...
+    private static final Color BACKGROUND_COLOR = Color.YELLOW;
+    private static final double TOLERANCE = 0.07;
+    private Stage testStage;
+
+    @Test
+    public void testSetWidthOnlyAfterShownOnContentSizeWindow() throws Exception {
+        CountDownLatch latch = new CountDownLatch(1);
+        final int finalWidth = 200;
+
+        setupContentSizeTestStage(300, 300,
```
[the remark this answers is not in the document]
tsayao: Done.

kevinrushforth (approved):
Looks good.

Since this is a very minor change, I am lowering the re-review requirement to 1.

/reviewers 1

lukostyra (approved):

kevinrushforth:
@tsayao This is ready to integrate when you are ready.

yushijinhun:
Hi, I reproduced this bug on OpenJFX 17.0.10, but not on 17.0.9. Therefore, I think it might be introduced in https://github.com/openjdk/jfx/pull/915. Could you please consider backporting this fix to JavaFX 17?

PR #1655, opened 2024-12-02
8345348: CSS media feature queries

mstr2:
Implementation of [CSS media queries](https://gist.github.com/mstr2/cbb93bff03e073ec0c32aac317b22de7).

---------
### Progress
- [x] Change requires CSR request [JDK-8358332](https://bugs.openjdk.org/browse/JDK-8358332) to be approved

### Issues
 * [JDK-8345348](https://bugs.openjdk.org/browse/JDK-8345348): CSS media feature queries (**Enhancement** - P4)
 * [JDK-8358332](https://bugs.openjdk.org/browse/JDK-8358332): CSS media feature queries (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1655/head:pull/1655` \
`$ git checkout pull/1655`

`$ git checkout pull/1655` \
`$ git pull https://git.openjdk.org/jfx.git pull/1655/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1655`

`$ git pr show -t 1655`

</details>

<a href="https://git.openjdk.org/jfx/pull/1655.diff">https://git.openjdk.org/jfx/pull/1655.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1655#issuecomment-2513198119)
</details>


modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
```diff
@@ -1175,6 +1187,155 @@ public String getName() {
...
+     *         <tr><th>For</th><td><code>@media</code></td></tr>
+     *         <tr><th>Value</th><td><code>light</code> | <code>dark</code></td></tr>
+     *         <tr><th>Boolean Context</th><td>no</td></tr>
+     *     </tbody>
+     * </table>
+     *
+     * @defaultValue {@link Platform.Preferences#getColorScheme()}
+     * @since 24
+     */
+    private final ObjectProperty<ColorScheme> colorScheme = new MediaProperty<>(
```
andy-goryachev-oracle, line 1212: is it possible for two different scenes to have different color schemes or any other properties derived from the platform preferences?

would it make more sense to use the platform preferences properties directly?

mstr2:
Let's try this again.

/open

hjohn:
I think you should consider moving `MediaRule`, `MediaQuery` and `MediaQueryContext` to `com.sun.javafx.css` (or similar internal package).  The javafx.css module IMHO should not be further extended in its current form, as IMHO it needs a rethink first.  From what I gather it was only made public to support writing binary CSS files, which was done primarily to be able to load huge stylesheets like caspian/modena a bit faster; it's public interface was never really thought through, and there are no known users (aside from Scene Builder), and AFAIK no use cases for the API (`StyleSheet` can't be directly used by the CSS engine, and doing so would probably cripple it as the `StyleSheet` class seems to be observable in all its aspects -- I have doubts any of that is currently in use, but perhaps Scene Builder uses it).

I verified on a local build that this is easy to do (you need to make some constructors public, and the writeBinary/readBinary functions after they've been moved to the internal package).

andy-goryachev-oracle:
In the .md document, you make a reference to w3.org media queries, so one question comes right away: do you eventually plan to implement the whole spec as defined there, or we are really talking about a very small subset of features?

For example, do you plan to support media query operations like `and`, `not`, `only`, etc.?  Also, the w3.org spec might be way too complicated even to be referenced - things there will never apply to JavaFX (`tty`, `projection`, `speech`, etc.)

Alternatively, wouldn't it make _more_ sense to add variables to the JavaFX CSS instead?  Or maybe in conjunction with a simpler media-query-like syntax, something like @if(scene.isDark)?

My concern is that the media query features (whatever subset of it we decide to implement) might make the code way too complex without providing substantial benefit, and perhaps we can achieve the same goal (of dynamically activating parts of the stylesheet) using simpler solutions like variables?

What do you think?

mstr2:
> In the .md document, you make a reference to w3.org media queries, so one question comes right away: do you eventually plan to implement the whole spec as defined there, or we are really talking about a very small subset of features?

We'd probably never implement the whole spec, but only the parts that make sense for JavaFX. Starting with user-preference media features is an obvious choice, because we already have a matching set of platform preferences (the astute reader might notice just how closely the existing platform preferences, the proposed media queries, and the W3C standard align -- almost as if by design :wink:).

> For example, do you plan to support media query operations like `and`, `not`, `only`, etc.? Also, the w3.org spec might be way too complicated even to be referenced - things there will never apply to JavaFX (`tty`, `projection`, `speech`, etc.)

The `only` keyword makes no sense here, it's a legacy keyword for the web. Other than that, this PR already implements the full syntax for media feature queries, including `and`, `or`, and `not` (except for the _range form_, which we only need to support if we decide to also support viewport characteristics features like `width`).

I reference the W3C specification because like with all recently added CSS features, this is an implementation of the standard. The JavaFX CSS reference documents the extent to which the standard is implemented.

> Alternatively, wouldn't it make _more_ sense to add variables to the JavaFX CSS instead? Or maybe in conjunction with a simpler media-query-like syntax, something like @if(scene.isDark)?

CSS variables can't be implemented without a [major rewrite of the CSS system](https://gist.github.com/mstr2/f416996caf48e11193f0b6a5883a3926). We can do this, but it is not really related to the feature proposed here. Even with CSS variables, you need a way to toggle their value based on external configuration, and then you're right back at the start of the problem.

I am very much against inventing _any_ non-standard CSS syntax whatsoever. Every time JavaFX did that, the result was not good. Let's face the reality here: the JavaFX CSS implementation is severely undercooked, and that's the very best thing I can say about it. I don't want to make it any worse.

> My concern is that the media query features (whatever subset of it we decide to implement) might make the code way too complex without providing substantial benefit, and perhaps we can achieve the same goal (of dynamically activating parts of the stylesheet) using simpler solutions like variables?

You seem to be arguing against media queries themselves, not against their implementation in JavaFX. But they are a thing in the technology that 99,9% of developers know as CSS, it's a brute fact of nature. I think we should acknowledge that, and not try to do our own little thing again.

mstr2:
> Also, would it be possible to add Goals and Non-Goals to the .md, make it more JEP-like?

I added the two sections.

modules/javafx.graphics/src/test/java/test/javafx/scene/Scene_preferences_Test.java
```diff
@@ -0,0 +1,108 @@
...
+                    prefs -> prefs.setPersistentScrollBars(true)),
+            new TestRun("prefers-reduced-motion",
+                        prefs -> prefs.setReducedMotion(false),
+                        prefs -> prefs.setReducedMotion(true)),
+            new TestRun("prefers-reduced-transparency",
+                    prefs -> prefs.setReducedTransparency(false),
+                    prefs -> prefs.setReducedTransparency(true)),
+            new TestRun("prefers-reduced-data",
+                    prefs -> prefs.setReducedData(false),
+                    prefs -> prefs.setReducedData(true))
```
andy-goryachev-oracle, line 86: it's probably an IDE artifact, but indentation is all over the place.

andy-goryachev-oracle (commented):
Partial review: the test cases are pretty comprehensive, thanks!

andy-goryachev-oracle:
> You seem to be arguing against media queries themselves

Probably.  I am afraid the CSS subsystem will grow into a monster (I mean, it already is).

The follow up question is what you guys think about updating `modena.css` to support the dark mode?  I understand the platform preferences and these media queries are prerequisites, what are other things that we'll need?

mstr2:
> > You seem to be arguing against media queries themselves
> 
> Probably. I am afraid the CSS subsystem will grow into a monster (I mean, it already is).

It is, and it’s held together by duct tape in many places. That’s why I think we should consider the rewrite at some point in the future to put it on solid footing (and it’s the only way to get really useful things like variables).

Media queries are pretty orthogonal to that, as most of the implementation will be the same with a potential rewrite.

> The follow up question is what you guys think about updating `modena.css` to support the dark mode? I understand the platform preferences and these media queries are prerequisites, what are other things that we'll need?

What’s missing to properly support dark mode is really only the external configuration of stylesheets, so I think with media queries in place we have everything we need for a dark version of Modena.

hjohn:
> > You seem to be arguing against media queries themselves
> 
> Probably. I am afraid the CSS subsystem will grow into a monster (I mean, it already is).
> 
> The follow up question is what you guys think about updating `modena.css` to support the dark mode? I understand the platform preferences and these media queries are prerequisites, what are other things that we'll need?

With some minimal style changes (and a hack to change the window border to dark mode on Windows) I get this out of modena.css:

image

mstr2:
> With some minimal style changes (and a hack to change the window border to dark mode on Windows) I get this out of modena.css:

Nice! Once we have a per-scene color scheme, I plan to have window borders adjust to the color scheme of the scene.

By the way, can I interest you in a review of this PR?

hjohn:
> By the way, can I interest you in a review of this PR?

Yes, I already partially checked it out, and was intending to do a closer review soon.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/parser/TokenStream.java
```diff
@@ -0,0 +1,114 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package com.sun.javafx.css.parser;
+
+import java.util.List;
+import java.util.function.Predicate;
+
+public final class TokenStream {
```
hjohn, line 31: I think the naming of the methods in this class leaves something to be desired.

- `Token consume()` **OK**
- `Token consume(Predicate)` -> `consumeIf` or `consumeIfMatches`
    Makes it clearer, as `consume` seems to imply something is always consumed (ie. it still skips one token if the predicate didn't match returning `null`).
- `Token peek()` **OK**
- `reconsume` -> `unconsume`, `undoConsume`, `undo`, `previous`, `resetToPrevious`, `decrementIndex`
    Nothing is consumed which reconsume seems to imply, instead it moves the index back one place so the next call to `consume` may indeed reconsume a token; reconsume as-is would do nothing (and if it returned anything it would be same the as `current`.
- `boolean peekMany(Predicate...)` -> `matches`
    It doesn't work like `peek`.  `peekMany` would imply it returns a `List<Token>` 
- `reset(int)` -> `setIndex`
    This seems to be similar to what say `InputStream` provides, but `InputStream` hides the `index` parameter so you can't set it to some arbitrary value (like skipping ahead).  If you want to mimic this pattern, I'd suggest removing the parameter and providing a `mark` method.

modules/javafx.graphics/src/main/java/javafx/css/Rule.java
```diff
@@ -48,6 +50,17 @@
...
+            @Override
+            public MediaRule getMediaRule(Rule rule) {
+                return rule.mediaRule;
+            }
+        });
+    }
+
+    private final MediaRule mediaRule;
+
     private List<Selector> selectors = null;
```
hjohn, line 64: I dislike the accessor pattern somewhat; I wonder, would it be possible to:

- Make `Rule` sealed: `public sealed abstract class Rule permits InternalRule` (*) similar to how this was done for `Selector`: `public abstract sealed class Selector permits SimpleSelector, CompoundSelector`
- `InternalRule` would not be public API, and can therefore introduce a public method to get `MediaRule`
- `Rule` does not have a public constructor, so IMHO there's no risk in there ever being an instance of `Rule` directly (similar reasoning was used to when the `SimpleSelector` / `CompountSelector` refactoring was done).

This way getting access to the `getMediaRule` method is just a simple `instanceof` check away:

     MediaRule rule = rule instanceof InternalRule ir ? ir.getMediaRule() : null;

(*) `InternalRule` is just a place holder name

hjohn (commented):
A partial review, I didn't look at the tests.

I noticed some non-intuitive equals/hashCode overrides, but found no evidence they're needed to make the non-test code work; if they're only there to make testing more convenient, I really think the overrides should be removed.  In general, IMHO, any equals/hashCode override that doesn't fully test equality (apart from `transient` or derived fields) should be avoided as it will be surprising if such classes are ever used with `equals` or any hash based collection classes.

If there's need for partial equality, then externalize this.

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
```diff
@@ -230,19 +231,19 @@ public T get() {
...
 
-        public void setValueOverride(T value) {
+        public synchronized void setValueOverride(T value) {
             this.overrideValue = value;
             updateEffectiveValue();
             fireValueChangedEvent();
         }
 
-        public void fireValueChangedIfNecessary() {
+        public synchronized void fireValueChangedIfNecessary() {
```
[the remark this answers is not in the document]
mstr2: Platform preference change events always happen on the FX thread. The reason for `synchronized` is as follows:

A `Scene` can be created on any thread as per spec. When the scene is created, the `ScenePreferences` class and its associated properties are instantiated, which in turn subscribe internally to the `Platform.Preferences` properties. The synchronization is to protect the addition and removal of listeners.

This is not ideal, though. As soon as we're subscribed to `Platform.Preferences`, the `Scene.Preferences` can receive change notifications on the FX thread, which interferes with the mandated capability to be created on any thread.

I think we have a few options here:
1. Don't subscribe scene preferences to platform preferences until the scene is shown. The downside of this is that we don't know the actual values of the preferences up until that point.
2. With limited synchronization, fetch the current preference values from the platform when the scene is created, but don't subscribe to change notifications until the scene is shown.
3. Specify that scene preference changes can happen on the FX thread, even when the scene is created in a background thread.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/expression/FunctionExpression.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(featureName, featureValue, value);
+    }
+
+    @Override
+    public String toString() {
+        return "(" + (featureValue != null ? featureName + ": " + featureValue : featureName) + ")";
+    }
```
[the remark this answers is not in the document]
mstr2: The expression is entirely defined by `featureName`, `featureValue`, and `value`. Since the associated function is usually a method reference, it won't equal other method references (even though the referenced method is the same). There's no point in equating different method references of the same method because this just means that no two expressions are ever equal.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/parser/TokenStream.java
mstr2: I renamed `consume(Predicate)` to `consumeIf(Predicate)`, and `peekMany(Predicate...)` to `matches(Predicate...)`.

`index()` and `reset()` don't need to be public, so I've removed that (also I don't want to mimic the problematic mark/reset pattern of `InputStream`).

`reconsume` is [CSS lingo](https://www.w3.org/TR/css-syntax-3/#tokenizer-definitions), I'm keeping that as-is.

modules/javafx.graphics/src/main/java/javafx/css/Rule.java
mstr2: I can see where you're coming from, and this really makes sense when you're having a public interface and an internal implementation. But that's not the case here, the implementation is out there in `javafx.css`. I don't mind the accessor pattern here (it's simple enough, and we can change it at any time), and would rather revisit the entire public CSS API at some point and either remove it or make it useful.

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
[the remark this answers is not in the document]
mstr2: I've decided to go with option 2.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/parser/TokenStream.java
mstr2: I don't want to die on the `reconsume` hill here, but I think there is some value in being able to read through the W3C specification of the CSS algorithms, and find similar terminology in our implementation. Makes it a bit easier to correlate specification and implementation.

andy-goryachev-oracle: I think there is value in clarity, especially if we are not implementing 100% of the "spec".  I agree with John this name is suboptimal.

andy-goryachev-oracle:
Would it be possible to expand the examples in the .md file?  For example, an example that uses operator(s)?  

Another question: is there any special prioritization given to media queries relative to other rules?

mstr2:
> Another question: is there any special prioritization given to media queries relative to other rules?

No, a media query just either matches, or it doesn't. If it matches, the rule enclosed by the media query is treated like any other rule; if it doesn't match, the rule doesn't apply.

andy-goryachev-oracle:
> if it doesn't match, the rule doesn't apply.

and if does apply, it overrides any previous rules (but can be overridden by any subsequent rules), right?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaQuerySerializer.java
```diff
@@ -0,0 +1,117 @@
...
+            };
+        }
+
+        static final QueryType[] VALUES = values();
+    }
+
+    public static void writeBinary(MediaQuery mediaQuery,
+                                   DataOutputStream os,
+                                   StyleConverter.StringStore stringStore) throws IOException {
+        os.writeByte(QueryType.of(mediaQuery).ordinal());
```
andy-goryachev-oracle, line 72: potential problem:
someone may insert/reorder QueryType enums and break backward compatibility.
at a minimum, a warning comment should be added to QueryType, or perhaps some more reliable code in read/writeBinary?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaQuerySerializer.java
```diff
@@ -0,0 +1,117 @@
...
+
+            case DisjunctionExpression expr -> {
+                writeBinary(expr.left(), os, stringStore);
+                writeBinary(expr.right(), os, stringStore);
+            }
+        }
+    }
+
+    public static MediaQuery readBinary(DataInputStream is, String[] strings) throws IOException {
+        return switch (QueryType.VALUES[is.readByte()]) {
```
andy-goryachev-oracle, line 104: same issue with backward compatibility - what if an old core tries to read the newer format and AIOOBE is thrown instead of IOException?  Should we validate first and throw IOE instead?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaQuerySerializer.java
andy-goryachev-oracle: or perhaps use hard-coded values instead 'c', 'f', 'C', 'd', 'n' ?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaRule.java
```diff
@@ -0,0 +1,130 @@
...
+public final class MediaRule {
+
+    private final List<MediaQuery> queries;
+    private final MediaRule parent;
+    private final int hash;
+
+    public MediaRule(List<MediaQuery> queries, MediaRule parent) {
+        this.queries = List.copyOf(queries);
+        this.parent = parent;
+        this.hash = queries.hashCode();
```
andy-goryachev-oracle, line 48: I would rather see `this.queries.hasCode()` here

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaRule.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /**
+     * Returns the list of media queries.
```
andy-goryachev-oracle, line 52: should we say "immutable"?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaRule.java
```diff
@@ -0,0 +1,130 @@
...
+        int size = stream.readInt();
+        var queries = new MediaQuery[size];
+
+        for (int i = 0; i < size; i++) {
+            queries[i] = MediaQuerySerializer.readBinary(stream, strings);
+        }
+
+        boolean hasParent = stream.readBoolean();
+
+        return new MediaRule(List.of(queries), hasParent ? readBinary(stream, strings) : null);
```
andy-goryachev-oracle, line 118: the exceptions in java have improved, but there is 3 statements in this line.
anyone who ever debugged code after the fact using two-week old logs would say: please keep one statement per line, if possible.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaRule.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        return obj instanceof MediaRule rule && rule.getQueries().equals(queries);
```
andy-goryachev-oracle, line 123: should we add `if(obj == this)` as we usually do?
also, `parent` is missing in the check

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaRule.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return hash;
+    }
```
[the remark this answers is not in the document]
andy-goryachev-oracle: As far as I can tell, MediaRule is not exposed publicly, neither it's put into a hashtable, so we might as well remove `hash`, `hashCode()` and `equals()`

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/expression/ConjunctionExpression.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public boolean evaluate(MediaQueryContext context) {
+        return left.evaluate(context) && right.evaluate(context);
+    }
+
+    @Override
+    public String toString() {
+        return "(" + left.toString() + " and " + right.toString() + ")";
```
andy-goryachev-oracle, line 67: `return "(" + left + " and " + right + ")";`

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/expression/DisjunctionExpression.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public boolean evaluate(MediaQueryContext context) {
+        return left.evaluate(context) || right.evaluate(context);
+    }
+
+    @Override
+    public String toString() {
+        return "(" + left.toString() + " or " + right.toString() + ")";
```
andy-goryachev-oracle, line 67: `return "(" + left + " or " + right + ")";`

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/expression/FunctionExpression.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Override
+    public boolean evaluate(MediaQueryContext context) {
+        return Objects.equals(function.apply(context), value);
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        return obj instanceof FunctionExpression<?> expr
```
andy-goryachev-oracle, line 58: `if(obj == this)` ?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/expression/FunctionExpression.java
```diff
@@ -0,0 +1,73 @@
...
+            && expr.featureName.equals(featureName)
+            && Objects.equals(expr.featureValue, featureValue)
+            && Objects.equals(expr.value, value);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(featureName, featureValue, value);
```
andy-goryachev-oracle, line 66: very minor: maybe add `FunctionExpression.class` to the mix?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/parser/CssLexer.java
```diff
@@ -23,57 +23,52 @@
...
-    final static int WS = 40;
-    final static int NL = 41;
-    final static int FONT_FACE = 42;
-    final static int URL = 43;
-    final static int IMPORT = 44;
-    final static int SECONDS = 45;
-    final static int MS = 46;
-    final static int AT_KEYWORD = 47;
+public final class CssLexer {
+    public final static int STRING = 10;
```
andy-goryachev-oracle, line 34: minor: `public static final` maybe?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/parser/TokenStream.java
```diff
@@ -0,0 +1,106 @@
...
+        Token nextToken = consume();
+        if (nextToken != null && predicate.test(nextToken)) {
+            return nextToken;
+        }
+
+        reconsume();
+        return null;
+    }
+
+    public void reconsume() {
```
andy-goryachev-oracle, line 73: maybe a javadoc would suffice if you like the name so much

modules/javafx.graphics/src/main/java/com/sun/javafx/css/parser/TokenStream.java
```diff
@@ -0,0 +1,106 @@
...
+            for (Predicate<Token> predicate : predicates) {
+                if (consumeIf(predicate) == null) {
+                    return false;
+                }
+            }
+
+            return true;
+        } finally {
+            currentIndex = index;
+            currentItem = index >= 0 ? source.get(index) : null;
```
andy-goryachev-oracle, line 103: general question: will this `finally` block work as expected if an exception happened somewhere in the `try` block?

modules/javafx.graphics/src/main/java/javafx/css/CssParser.java
```diff
@@ -4282,14 +4322,94 @@ private void parse(Stylesheet stylesheet, CssLexer lexer) {
...
+                if (!consumeRBrace(lexer)) {
+                    return;
+                }
+
+                if (mediaRule != null) {
+                    mediaRule = mediaRule.getParent();
+                }
+
+                currentToken = nextToken(lexer);
+                expectedRBraces--;
```
andy-goryachev-oracle, line 4339: what happens in case of unbalanced braces and `expectedRBraces` < 0 ?

modules/javafx.graphics/src/main/java/javafx/css/Rule.java
```diff
@@ -361,6 +379,15 @@ final void writeBinary(DataOutputStream os, StyleConverter.StringStore stringSto
     static Rule readBinary(int bssVersion, DataInputStream is, String[] strings)
             throws IOException
     {
+        MediaRule mediaRule = null;
+
+        if (bssVersion >= 7) {
```
andy-goryachev-oracle, line 384: please add a comment pointing to Stylesheet.BINARY_CSS_VERSION.

Ideally, there would be constants declared somewhere with explanation instead of the magic number '7'

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -63,8 +64,9 @@ public class Stylesheet {
      * binary stream changes. This number does not correlate with JavaFX versions.
      * Version 5: persist @font-face
      * Version 6: converter classes moved to public package
+     * Version 7: media queries
```
andy-goryachev-oracle, line 67: this, in my opinion, deserves a constant (it does not have to be public).

modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
```diff
@@ -6506,4 +6521,207 @@ Accessible getAccessible() {
...
+        ObjectProperty<Boolean> persistentScrollBarsProperty();
+
+        /**
+         * Gets the value of the {@code persistentScrollBars} property.
+         *
+         * @return the value of the {@code persistentScrollBars} property
+         * @see #persistentScrollBarsProperty()
+         * @see #setPersistentScrollBars(Boolean)
+         */
+        boolean isPersistentScrollBars();
```
andy-goryachev-oracle, line 6593: do these accessors need individual javadocs?

andy-goryachev-oracle (commented):
first batch of comments.  more to follow.

modules/javafx.graphics/src/test/java/test/com/sun/javafx/css/media/MediaQuerySerializerTest.java
```diff
@@ -0,0 +1,128 @@
...
+import org.junit.jupiter.api.Test;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class MediaQuerySerializerTest {
```
andy-goryachev-oracle, line 46: just a thought: this could have been a parameterized test (this version is ok though).

modules/javafx.graphics/src/test/java/test/com/sun/javafx/css/parser/CssLexerTest.java
```diff
@@ -45,7 +44,7 @@ public class CssLexerTest {
     public CssLexerTest() {
     }
 
-    private void checkTokens(List<TokenShim> resultTokens, TokenShim... expectedTokens) {
+    private void checkTokens(List<Token> resultTokens, Token... expectedTokens) {
```
andy-goryachev-oracle, line 47: should there be new tests in `CssLexerTest` for the `@media` queries?

modules/javafx.graphics/src/test/java/test/javafx/css/CssParser_mediaQuery_Test.java
```diff
@@ -0,0 +1,310 @@
...
+                    new FunctionExpression<>("prefers-reduced-motion", "reduce", _ -> false, true)),
+                new FunctionExpression<>("prefers-persistent-scrollbars", "persistent", _ -> false, true)
+            ),
+            mediaRule.getQueries().getFirst()
+        );
+    }
+
+    @Test
+    void disjunctionAndConjunctionCanNotBeCombined() {
+        Stylesheet stylesheet = new CssParser().parse("""
```
andy-goryachev-oracle, line 185: could we also test for "or not" and "and not" ?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaQueryParser.java
```diff
@@ -0,0 +1,285 @@
...
+        if (tokens.matches(LPAREN, IDENT, RPAREN) || tokens.matches(LPAREN, IDENT, COLON)) {
+            return parseMediaFeature(tokens);
+        }
+
+        // ( <media-condition> )
+        if (tokens.consumeIf(LPAREN) != null) {
+            MediaQuery expression = parseMediaCondition(tokens);
+
+            if (tokens.consumeIf(RPAREN) == null) {
+                errorHandler.accept(tokens.consume(), "Expected RPAREN");
```
andy-goryachev-oracle, line 216: RPAREN -> right parenthesis '('?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaQueryParser.java
```diff
@@ -0,0 +1,285 @@
...
+                return null;
+            }
+
+            return expression;
+        }
+
+        errorHandler.accept(tokens.consume(), "Expected LPAREN");
```
andy-goryachev-oracle, line 223: LPAREN -> left parenthesis ')'?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaQueryParser.java
```diff
@@ -0,0 +1,285 @@
...
+     *     <mf-plain> = <ident> : <any>
+     *     <mf-boolean> = <ident>
+     * }</pre>
+     *
+     * @param tokens the token stream
+     * @return the expression
+     */
+    private MediaQuery parseMediaFeature(TokenStream tokens) {
+        if (tokens.consumeIf(LPAREN) == null) {
+            errorHandler.accept(tokens.consume(), "Expected LPAREN");
```
andy-goryachev-oracle, line 241: LPAREN?

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
```diff
@@ -824,6 +826,86 @@ <h3><a id="introatrules">&#64; Rules</a></h3>
...
+    <h4>@media</h4>
+    <p>A media query is a method of testing certain aspects of the <a href="../../../javafx/scene/Scene.html">Scene</a>.
+        Media queries are independent of the contents of the scene graph, its styling, or any other internal aspect;
+        they're only dependent on "external" configuration of the <code>Scene</code>.
+    <p>Several media queries can be combined into a comma-separated <strong>media query list</strong>.
+        A media query list evaluates to <code>true</code> if <em>any</em> of the media queries is <code>true</code>,
+        and evaluates to <code>false</code> only if <em>all</em> the media queries are <code>false</code>.
+        An empty media query list evaluates to <code>true</code>.
+        <figure style="margin: 0">
+            <img src="media-query.svg" width="210" alt="Media Query List">
```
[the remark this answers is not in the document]
mstr2: Yes.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaQueryParser.java
```diff
@@ -0,0 +1,285 @@
...
+                        }
+
+                        // Invalid expressions always evaluate to false.
+                        expressions.add(new ConstantExpression(false));
+                    }
+                }
+
+                default -> {
+                    errorHandler.accept(stream.current(), "Unexpected token");
+                    return expressions;
```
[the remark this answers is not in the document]
mstr2: The return statement? Yes, we want token processing to stop here if we encounter an unexpected token, as there's no recovery strategy at this point.

modules/javafx.graphics/src/test/java/test/javafx/css/CssParser_mediaQuery_Test.java
```diff
@@ -0,0 +1,310 @@
...
+            }
+            """);
+
+        var mediaRule = RuleHelper.getMediaRule(stylesheet.getRules().getFirst());
+        assertEquals(1, mediaRule.getQueries().size());
+        assertEquals(new ConstantExpression(false), mediaRule.getQueries().getFirst());
+    }
+
+    @Test
+    void parserRecoversWhenMediaQueryIsMalformed() {
```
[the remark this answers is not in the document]
mstr2: I've added a test for an invalid operator combination, as well as unbalanced parentheses.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaQuerySerializer.java
mstr2: I've added an explicit `serializedId` value.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaQuerySerializer.java
mstr2: That would be forward compatibility, which is not supported. JavaFX will not even try to load a binary file with a higher version than it can understand, and error out immediately (see Stylesheet.java:L317).

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaRule.java
mstr2: I've removed it instead.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaRule.java
mstr2: This is now removed.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/expression/FunctionExpression.java
mstr2: Why?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/parser/CssLexer.java
mstr2: I want to keep the diffs minimal here, because I didn't change anything except making the constants public.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/parser/TokenStream.java
mstr2: I've added javadocs to all methods. By the way, I don't really get the opposition to the name. The arguments presented so far do not convince me to drop CSS lingo.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/parser/TokenStream.java
mstr2: That's its purpose. It will reset the stream back to the state it was in before the `matches(...)` method was called.

modules/javafx.graphics/src/main/java/javafx/css/CssParser.java
mstr2: The behavior is the same as with a normal unbalanced curly brace. I've added a test for it.

modules/javafx.graphics/src/main/java/javafx/css/Rule.java
mstr2: I've added comments to that effect.

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
mstr2: I'd rather do this in a refactor PR with the rest of the magic numbers.

modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
mstr2: Yes, because the javadoc tool doesn't recognize a method that returns `boolean` as a getter for `ObjectProperty<Boolean>` (it expects `BooleanProperty`). This is a null-coalescing property: you can set it to `null`, but it will always evaluate to `true` or `false`, so it doesn't make sense to return `Boolean` from its getter.

modules/javafx.graphics/src/test/java/test/com/sun/javafx/css/media/MediaQuerySerializerTest.java
mstr2: Yes, with the downside that parameterized tests are usually a bit harder to debug in isolation.

modules/javafx.graphics/src/test/java/test/com/sun/javafx/css/parser/CssLexerTest.java
mstr2: No, media queries are not a thing at the lexer stage. The media query grammar only comes in at a later stage.

modules/javafx.graphics/src/test/java/test/javafx/css/CssParser_mediaQuery_Test.java
mstr2: I've added a test.

mstr2:
> Should this work?

No. I've added a test for it.

> Also, are people supposed to know what LPAREN is? Can we provide more user friendly error messages?

The error messages of `CssParser` are what they are, I guess. For example, it will also emit messages like "Expected LBRACE" and "Expected RBRACE". If we change that, then probably with a cleanup ticket.

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/media-query.svg
```diff
@@ -0,0 +1,94 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
```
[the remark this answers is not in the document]
mstr2: I don't really see a major difference on your screenshot, both versions look blurry to me. When I open `cssref.html` in Chrome, it looks quite sharp.

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
```diff
@@ -824,6 +826,86 @@ <h3><a id="introatrules">&#64; Rules</a></h3>
...
+        <li>Any media feature can be negated by placing the <code>not</code> operator before it.
+        <li>Two or more media features can be chained together, such that the query is only true if <em>all</em> the
+            media features are true, by placing the <code>and</code> operator between them.
+        <li>Two or more media features can be chained together, such that the query is true if <em>any</em> of the
+            media features are true, by placing the <code>or</code> operator between them.
+        <li>Expressions can be grouped by wrapping them in parentheses.
+        <li>It is invalid to mix different boolean operators at the same "level" of a media query. For example,
+            <code>(prefers-color-scheme: dark) and (prefers-reduced-motion) or (prefers-reduced-transparency)</code>
+            is illegal, as it is unclear what was meant. In this case, parentheses must be used to group expressions.
+    </ul>
```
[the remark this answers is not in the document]
mstr2: I don't want to overload cssref with too many examples, as this is not supposed to be a general-purpose CSS documentation. The media query doc is already quite comprehensive as it is. For detailed information, there's always the W3C specification, and hundreds of third-party sources.

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
```diff
@@ -173,15 +170,19 @@ public void setAccentColor(Color color) {
...
+                }
             }
-        }
 
-        for (DeferredProperty<?> property : deferredProperties.values()) {
-            property.fireValueChangedIfNecessary();
+            colorScheme.update();
+
+            for (DeferredProperty<?> property : deferredProperties.values()) {
+                property.fireValueChangedIfNecessary();
```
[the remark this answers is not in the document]
mstr2: I've moved the `fireValueChangedIfNecessary` out of the synchronized block. We need the synchronized block as a memory barrier to see the latest value of the `effectiveValue` field, as we need to be able to read it from any thread.

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/media-query.svg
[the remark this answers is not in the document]
hjohn: It looks sharp to me as well.  The w3.org one looks better when I view it on the actual website (not your snapshot above).  

Mac renders things slightly differently and leans heavily on anti-aliasing to represent visuals as accurately as possible; this works well on high-DPI displays but tends to look blurry on lower-DPI screens. Windows, on the other hand, limits anti-aliasing more to achieve a crisper, sharper result; this is more effective on lower-DPI displays and doesn't really make a difference for high-DPI screens.

andy-goryachev-oracle (commented):
Thank you for making changes and adding tests!  And sorry for all the nitpicking.

Looks good, with only two sticky points remaining:

- I would like to see an example of correct syntax for the operator (and/or how to write `and not` case)
- still think synchronization should not be used, instead use `volatile` and check for fx thread in the mutators.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/expression/FunctionExpression.java
andy-goryachev-oracle: > Why?

made me think.  the reason is _probably_ so that `class B extends A` would produce a different `hashCode`, but then `A.hashCode` should include not the class instance, but the `getClass()`:

```java
class A {
  public int hashCode() {
    return Objects.hash(getClass(), ...);
  }
}

class B extends A {
  // B.hashCode() produces a value different from A.hashCode()
}
```

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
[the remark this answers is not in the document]
mstr2: `volatile` is not enough, because the different values are correlated (for example `foreground`, `background`, and `colorScheme`). All values need to be updated in a single transaction, as otherwise we could end up reading a transient combination of values.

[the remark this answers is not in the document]
andy-goryachev-oracle: I see.  The I suggest we enforce single-threaded nature of it: everything must happen in the FX Application Thread.  I don't think it's a good idea to start synchronizing things in FX now all of a sudden.

Any code that wants to access FX from a background thread must do so using a combination of `runLater` and a `CountDownLatch`.

What do you think?

modules/javafx.graphics/src/test/java/test/com/sun/javafx/css/parser/CssLexerTest.java
mstr2: `CssLexer` is responsible for converting the stylesheet text into a token stream. In this stream, `@media` appears as an `AT_KEYWORD` token, but no further syntactic processing is done (i.e. the parser doesn't care at all if the sequence of tokens makes any sense). This only happens in `CssParser`, where the actual value of the token is inspected and subsequent tokens are interpreted according to a specific grammar (in this case, the media query grammar). That's why all of the parsing tests are in `CssParser_mediaQuery_Test`.

modules/javafx.graphics/src/test/java/test/javafx/css/CssParser_mediaQuery_Test.java
```diff
@@ -291,6 +291,38 @@ void invalidFeatureValueEvaluatesToFalse() {
...
+                new FunctionExpression<>("prefers-reduced-motion", "reduce", _ -> null, true),
+                new ConstantExpression(false) // the rest of the query is malformed and evaluates to false
+            ),
+            mediaRule.getQueries());
+    }
+
+    @Test
+    void invalidCombinationOfConjunctionAndNegationEvaluatesToFalse() {
+        Stylesheet stylesheet = new CssParser().parse("""
+            @media (prefers-reduced-motion: reduce) and not (prefers-color-scheme: dark) {
```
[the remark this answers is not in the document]
mstr2: That would be `(prefers-A) and (not (prefers-B))`. A bit convoluted, but that's what the grammar says...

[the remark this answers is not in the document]
andy-goryachev-oracle: please add this example to `cssref.html`

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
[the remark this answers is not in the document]
mstr2: That would be okay, except that `Scene` is allowed to be created on a background thread (it's in the class documentation), so we can't completely enforce it to be single-threaded. Once a scene is created, a user can also request its preferences. This is when we need some way to get the current platform preferences on a background thread, as otherwise users would potentially see the scene preferences suddenly change once the scene is shown.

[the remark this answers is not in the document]
andy-goryachev-oracle: That's why I am against this background thread nonsense.  Create - yes (even though I am against this).  Access and modify properties, especially those that interact with other parts of the FX - no.

I think we should enforce the FX application thread access.

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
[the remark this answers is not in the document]
mstr2: I've added more examples to cssref.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/parser/CssLexer.java
mstr2: Changed to canonical modifier order.

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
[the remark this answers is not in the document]
mstr2: This would be a significant change, as currently `Scene` is explicitly allowed to be created _and modified_ on a background thread. This was probably always a very bad idea (because it makes the implementation so much more difficult), but how do you think we should change that?

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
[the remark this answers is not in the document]
andy-goryachev-oracle: Thank you!

I might suggest to drop `<ul>` since these bullets are somewhat of a distraction.

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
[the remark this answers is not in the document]
andy-goryachev-oracle: how about: 

- the implementation must not fire any events before the Scene is shown
- the preferences getters should disallow calls from a background thread

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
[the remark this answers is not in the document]
mstr2: Without it, it just looks like a wall of text to me.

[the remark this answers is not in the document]
andy-goryachev-oracle: ok

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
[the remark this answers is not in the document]
andy-goryachev-oracle: PlatformPreferences.update() seems to be called only from the fx thread, if I read the code right.  What happens when the Scene is created in a bg thread and a listener added to its preferences, and an event is fired?  Will the statement `#1` be still valid?  (I don't see any code that delays the dispatch until after the Scene is shown).

`#2` - all this is good, but we seem to hand the gun to user and say: here, shoot yourself in the foot.  And they do.  I think it (creating things in a background thread) was a mistake in the first place, but there seems to be zero desire to change it, so who am I to say otherwise?

andy-goryachev-oracle (approved):
Thank you for expanding CSS media query examples!  Verified that they work as expected:

```css
@media (prefers-color-scheme: dark) and (not (prefers-persistent-scrollbars)) {
  .root {
    -fx-accent: yellow;
  }
}
```

I am still convinced javaFX should have been a single-threaded framework with no exceptions, but the current code seems ok as long as it's not abused.

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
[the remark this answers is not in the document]
mstr2: `NullCoalescingPropertyBase` (used in `ScenePreferences`) is initially disconnected. It will set its initial value to the value is reads from `PlatformPreferences`, but it will not subscribe to change notifications. The reason for this is to prevent a scene that is created on a background thread from immediately receiving change notifications on the FX thread, which could not only potentially corrupt the property state (because they're not concurrently thread-safe), but also mess up the application state.

When the scene is shown, `NullCoalescingPropertyBase` will subscribe to change notifications from `PlatformPreferences`, because now we're safely on the FX thread. The code for that is in ScenePreferences:L142.

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
```diff
@@ -268,51 +272,60 @@ private void updateEffectiveValue() {
...
+
+            fireValueChangeIfNecessary();
         }
 
-        @Override
-        public ColorScheme get() {
-            synchronized (mutex) {
-                return super.get();
+        // This method must only be called when synchronized on 'mutex'.
+        public void updateEffectiveValue() {
```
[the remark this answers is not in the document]
mstr2: No, the instance of `PreferenceProperties` only stored as a private field in `PlatformPreferences`. I'm usually using public methods in nested classes to indicate that the method is supposed to be called from outside of the nested class. That's not strictly necessary, but I think it serves to differentiate them from "purely" private methods.

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
```diff
@@ -235,23 +238,24 @@ public T get() {
...
                 this.overrideValue = value;
                 updateEffectiveValue();
-                fireValueChangedIfNecessary();
             }
+
+            fireValueChangedIfNecessary();
         }
 
-        // This method must only be called when synchronized on 'mutex'.
+        // This method must only be called on the FX application thread.
```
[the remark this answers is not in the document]
mstr2: The `Toolkit.getToolkit().checkFxUserThread()` check is in the calling method.

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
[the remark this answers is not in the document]
kevinrushforth: > I think it (creating things in a background thread) was a mistake in the first place, but there seems to be zero desire to change it

Your objection is noted, but we are not going to relitigate this. JavaFX explicitly allows creation and modification of a scene graph on a thread other than the JavaFX application thread up to the point before that scene is attached to a window.

This means that we do need to be careful in our implementation, especially avoiding anything that might run on the application thread prior to the scene becoming "live". Sometimes that means deferring the adding of listeners as was recently done for charts. It sounds like that's what is being done here by deferring the subscription to platform preferences until the scene is shown.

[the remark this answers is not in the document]
hjohn: @kevinrushforth ... what about manipulating a Scene that was at some point attached to a Window?  I mean, I've built a solution specifically for just-in-time registration **and** unregistration of listeners.  So IMHO when a Scene is detached or the window is no longer showing (basically any time it doesn't have a peer) then listeners need to detach as well (and re-attach when that situation changes again).

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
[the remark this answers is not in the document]
andy-goryachev-oracle: let me try again:

`Platform.getPreferences()` returns an instance of `PlatformPreferences`
`PlatformPreferences.colorSchemeProperty()` returns a `ColorSchemeProperty` with a public `updateEffectiveValue()`

the app can use reflection to invoke a public method, though admittedly in a convoluted way.

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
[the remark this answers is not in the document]
andy-goryachev-oracle: since it's a public method accessible to the application via the same method, I think you either need to make it private or add the check.

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PreferenceProperties.java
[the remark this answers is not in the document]
kevinrushforth: No, an application would not be able to do that. A getter whose implementation happens to return a subtype in a non-exported package does not allow the caller of that getter to access any methods in the non-exported class (even public ones). If it did, then you could never return a type that implements an interface, since all interface methods are public.

@mstr2 is right. There is no problem here that I can see.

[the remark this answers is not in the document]
kevinrushforth: Having said that, if the methods in question are not needed outside the package, consider making them package scope.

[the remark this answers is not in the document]
andy-goryachev-oracle: yes it can:
```
--add-opens javafx.graphics/com.sun.javafx.application.preferences=andy_test
```

```java
        var platformPreferences = Platform.getPreferences();
        var colorSchemeProperty = platformPreferences.colorSchemeProperty();
        Method m = colorSchemeProperty.getClass().getMethod("updateEffectiveValue");
        m.setAccessible(true);
        m.invoke(colorSchemeProperty);
```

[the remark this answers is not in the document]
kevinrushforth: > --add-opens javafx.graphics/com.sun.javafx.application.preferences=andy_test

You can access any number of internal classes or methods with "--add-exports" or "--add-opens", so that is not a valid argument.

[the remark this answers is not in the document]
andy-goryachev-oracle: okay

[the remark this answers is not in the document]
andy-goryachev-oracle: yes, I agree - thanks for the discussion!

(and `setAccessible()` is not needed with `--add-opens`)

I withdraw my comments - making these methods `private` is not needed, `public` or package-scope is fine.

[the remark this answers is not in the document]
mstr2: > So at the risk of repeating myself, there is no concern that making this method public exposes it to applications. However, making it package-scope if it doesn't need to be public, is a best practice if it doesn't need to be accessed outside the package.

These methods can have any visibility and it doesn't change their semantics at all, because the methods are declared in a private nested class (so they are not visible outside of the enclosing class). As a reminder, the enclosing class has access even to private members of its nested classes. The reason I chose to make the methods public is merely to indicate "this is supposed to be public API for the enclosing class".

I'm not inclined to change that, as all of the arguments presented so far do not apply. The methods are not visible in other top-level classes (even in the same package), and making them any less visible doesn't change that.

modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
```diff
@@ -6568,7 +6568,7 @@ public sealed interface Preferences permits ScenePreferences {
          * <table class="striped">
          *     <caption>Media Feature</caption>
          *     <tbody>
-         *         <tr><th>Name</th><td><code>prefers-persistent-scrollbars</code></td></tr>
+         *         <tr><th>Name</th><td><code>-fx-prefers-persistent-scrollbars</code></td></tr>
```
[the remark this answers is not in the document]
mstr2: The other queries are standardized and universally agreed upon, whereas this one isn't. It's there because we have a corresponding platform preference, so I think a vendor prefix is in order.

hjohn (approved):

modules/javafx.graphics/src/main/java/javafx/application/Platform.java
```diff
@@ -612,10 +612,9 @@ public sealed interface Preferences extends ObservableMap<String, Object>
         /**
          * Specifies whether applications should always show scroll bars. If not set, an application may
          * choose to hide scroll bars that are not actively used, or make them smaller or less noticeable.
-         * <p>
-         * If the platform does not report this preference, this property defaults to {@code false}.
          *
          * @return the {@code persistentScrollBars} property
+         * @mediaFeature <a href="../scene/doc-files/cssref.html#mediafeatures">{@code -fx-prefers-persistent-scrollbars}</a>
```
kevinrushforth: I note that this tag seems to be ineffective (perhaps because this is a property and gets special javadoc treatment).

kevinrushforth (commented):
The API changes look good. The cssref docs and the class docs for the new Scene.Preferences class look good. The one problem I see is that the `@mediaFeature` javadoc tag doesn't render anything, at least not when I run `gradle javadoc` using the default JDK 24.

You might need to revert this change and link to the cssref manually.

I note that the CSR spec was done before this change, so depending on how you resolve this, the spec should be regenerated.

modules/javafx.graphics/src/main/java/javafx/application/Platform.java
mstr2: That's correct, the javadoc tool copies javadocs from the property field to its accessor methods. But it doesn't do a throrough job: it only copies standard tags and a few custom tags (`@defaultValue` for example). I think that's either a bug or a missing feature of the javadoc tool. I can't find a single example where it would be useful to _not_ copy a javadoc tag from the property field to its accessor methods.

So it seems like we have several options:
1. Keep the custom `@mediaFeature` tag in JavaFX, and file an enhancement with the javadoc tool to copy _all_ tags to property accessors. Then the media feature tags would start to show up in generated docs once the javadoc enhancement is in.
2. File an enhancement, but only start using the `@mediaFeature` tag once javadoc is fixed.
3. Don't use a custom javadoc tag at all.

I would prefer option 1, because it is the most structured form of documentation, and doesn't require us to come back later. In general, custom javadoc tags are a very useful feature to present structured information, and I think we should prefer it to prose.

mstr2: I notice that I've described the behavior of the javadoc tool for property fields, but this is a property method. However, my conclusion is the same: not rendering custom tags is a bug in the tool.

kevinrushforth: Very likely this is a javadoc limitation. I don't know how likely it will be to get it fixed, but it would be no earlier than in JDK 26, which means that it would be (at least) JavaFX 27 before it started working. That makes option 1 less attractive.

I would probably recommend option 2. Otherwise, we won't see any reference to media feature queries in the Preferences doc any time soon.

kevinrushforth: Whichever option you decide, update the CSR to match. Then I'll Review the CSR and you can finalize it.

mstr2: I've updated the CSR to match.

andy-goryachev-oracle (approved):
#mediafeatures works from `Platform.Preferences` and `Scene.Preferences`

hjohn (approved):
This looks good to me now.

kevinrushforth (approved):
I think this is ready to go in. I've reviewed the API, and will rely on Andy's and John's review of the implementation.

PR #1662, opened 2024-12-10
8345937: Update JUnit to 5.11.3

Maran23:
The JUnit Platform got some updates over the last years, so it is a good idea to update it to the newest version.

That is from `5.8.1` to `5.11.3`.
Affected dependencies:
- `junit.jupiter`
- `junit.platform`
- `opentest4j`

Release notes: https://junit.org/junit5/docs/5.11.3/release-notes/

Note:
I did a handful of CI runs and got a test failure once in `ServiceLifecycleTest`. Not sure yet if related to the update or not.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issue
 * [JDK-8345937](https://bugs.openjdk.org/browse/JDK-8345937): Update JUnit to 5.11.3 (**Enhancement** - P4)

### Reviewers
 * [Michael Strauß](https://openjdk.org/census#mstrauss) (@mstr2 - **Reviewer**)
 * [Kevin Rushforth](https://openjdk.org/census#kcr) (@kevinrushforth - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1662/head:pull/1662` \
`$ git checkout pull/1662`

Update a local copy of the PR: \
`$ git checkout pull/1662` \
`$ git pull https://git.openjdk.org/jfx.git pull/1662/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 1662`

View PR using the GUI difftool: \
`$ git pr show -t 1662`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jfx/pull/1662.diff">https://git.openjdk.org/jfx/pull/1662.diff</a>

</details>
<details><summary>Using Webrev</summary>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1662#issuecomment-2532507120)
</details>

PR #1664, opened 2024-12-13
8346222: SwingNodePlatformExitCrashTest fails with JUnit 5.11.3

kevinrushforth:
This PR fixes a latent test bug in test bug in `SwingNodePlatformExitCrashTest` by setting a flag to skip shutdown rather than relying on a quirk of JUnit 5.8.1.

I discovered this while testing PR #1662 which updates JUnit to 5.11.3.

```
$ gradle sdk shims
$ gradle -PTEST_ONLY=true -PFULL_TEST=true -PUSE_ROBOT=true :systemTests:test --tests SwingNodePlatformExitCrashTest
...
SwingNodePlatformExitCrashTest > executionError FAILED
    org.opentest4j.AssertionFailedError: Exceeded timeout limit of 10000 msec
        at app//test.util.Util.runAndWait(Util.java:168)
        at app//test.util.Util.runAndWait(Util.java:139)
        at app//test.util.Util.shutdown(Util.java:316)
        at app//test.robot.javafx.embed.swing.SwingNodeBase.teardownOnce(SwingNodeBase.java:81)
        ...
```

The abstract `SwingNodeBase` base class calls `Util::shutdown` from its static `tearDownOnce` method, annotated with `@AfterAll`. The `SwingNodePlatformExitCrashTest` shuts down the FX toolkit as part of the test so must not call shutdown a second time. The current test logic attempts to "override" the static method in the base class with an @AfterAll annotation on a static method of the same name in the subclass. This accidentally works in JUnit 5.8 by hiding the method in the parent class, but no longer does in JUnit 5.11.

The right fix is for the subclass to set a flag such that the superclass will skip the call to `Util::shutdown`.

### DIFF
--- a/tests/system/src/test/java/test/robot/javafx/embed/swing/SwingNodeBase.java
+++ b/tests/system/src/test/java/test/robot/javafx/embed/swing/SwingNodeBase.java
@@ -53,6 +53,7 @@ public class SwingNodeBase {
     public static final int LONG_WAIT_TIME = 2500;
 
     protected static Robot robot;
+    protected static boolean doShutdown = true;
 
     // Used to launch the application before running any test
     private static final CountDownLatch launchLatch = new CountDownLatch(1);
@@ -78,7 +79,9 @@ public class SwingNodeBase {
 
     @AfterAll
     public static void teardownOnce() {
-        Util.shutdown();
+        if (doShutdown) {
+            Util.shutdown();
+        }
     }
 
     public static void runWaitSleep(Runnable run) {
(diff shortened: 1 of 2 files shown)


kevinrushforth:
@prsadhuk Please review

@Maran23 Can you also take a look at this?

kevinrushforth:
As an FYI, the behavior change that exposed this latent bug in our test was a deliberate change in JUnit 5.11.

https://github.com/junit-team/junit5/issues/3553

Indeed it was the case that before 5.11, a static method annotated with `@BeforeAll` or `@AfterAll` in a subclass would hide the same named method in the superclass, and starting in 5.11, it no longer does.

Maran23:
Thanks for investigating this in https://github.com/openjdk/jfx/pull/1662. I have read all comments and can confirm your investigation!
Funny enough, I found a similar problem when updating another project to the latest JUnit version. 
So it seems like a good idea to update JUnit so that we can no longer write tests that violate this behavior, only to find out later.

Maran23 (approved):

prsadhuk (approved):
Though not failing in windows, but I understand why it can fail.. looks good.

PR #1665, opened 2024-12-13
8346227: Seal Paint and Material

mstr2:
The `Paint` and `Material` classes can't be extended by user code, because their implementations require special support in internal JavaFX code. The classes should be sealed.

/csr
/reviewers 2

---------
### Progress
- [x] Change requires CSR request [JDK-8346461](https://bugs.openjdk.org/browse/JDK-8346461) to be approved

### Issues
 * [JDK-8346227](https://bugs.openjdk.org/browse/JDK-8346227): Seal Paint and Material (**Enhancement** - P4)
 * [JDK-8346461](https://bugs.openjdk.org/browse/JDK-8346461): Seal Paint and Material (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1665/head:pull/1665` \
`$ git checkout pull/1665`

`$ git checkout pull/1665` \
`$ git pull https://git.openjdk.org/jfx.git pull/1665/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1665`

`$ git pr show -t 1665`

</details>

<a href="https://git.openjdk.org/jfx/pull/1665.diff">https://git.openjdk.org/jfx/pull/1665.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1665#issuecomment-2542551066)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/Toolkit.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/Toolkit.java
@@ -594,23 +594,12 @@ public abstract class Toolkit {
     }
 
     public Object getPaint(Paint paint) {
-        if (paint instanceof Color) {
-            return createColorPaint((Color) paint);
-        }
-
-        if (paint instanceof LinearGradient) {
-            return getPaint((LinearGradient) paint);
-        }
-
-        if (paint instanceof RadialGradient) {
-            return getPaint((RadialGradient) paint);
-        }
-
-        if (paint instanceof ImagePattern) {
-            return createImagePatternPaint((ImagePattern) paint);
-        }
-
-        return null;
+        return switch (paint) {
+            case Color color -> createColorPaint(color);
+            case LinearGradient gradient -> getPaint(gradient);
+            case RadialGradient gradient -> getPaint(gradient);
+            case ImagePattern pattern -> createImagePatternPaint(pattern);
+        };
     }
 
     protected static final double clampStopOffset(double offset) {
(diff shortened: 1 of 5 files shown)


nlisker (approved):

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/Toolkit.java
```diff
@@ -594,23 +594,12 @@ private Object getPaint(RadialGradient paint) {
...
-        if (paint instanceof RadialGradient) {
-            return getPaint((RadialGradient) paint);
-        }
-
-        if (paint instanceof ImagePattern) {
-            return createImagePatternPaint((ImagePattern) paint);
-        }
-
-        return null;
+        return switch (paint) {
```
andy-goryachev-oracle, line 597: standard warning: not equivalent change (in the case of null `paint`).
it's ok in this case, since it looks like this method is never called with a null argument thanks to guards like this replicated across the code base:

```
peer.setFillPaint(getFill() == null ? null : tk.getPaint(getFill()));
```

would it have been easier to allow null argument and return null from `getPaint()` instead?

mstr2: `null` is not a valid argument for this method, and has never been. `acc_getPlatformPaint()` previously asserted that the return value is not `null`.

andy-goryachev-oracle: this is not what I am asking though: in more than one place we have _callers_ like this
```
peer.setFillPaint(getFill() == null ? null : tk.getPaint(getFill()));
```

it could be changed to 
```
peer.setFillPaint(tk.getPaint(getFill()));
```

mstr2: Yes, that would make the callsite a bit easier. However, from an API perspective, I think it's not the best approach to handle nulls in the toolkit layer. A system that channels nulls thorugh various layers tends to become harder to reason about over time.

andy-goryachev-oracle: I disagree: it makes it more resilient by doing what the caller has to do anyway.
However, the answer is acceptable.

andy-goryachev-oracle (approved):

PR #1668, opened 2024-12-23
8346281: [Windows] RenderScale doesn't update to HiDPI changes

jperedadnr:
This PR adds the missing native implementation for Windows `GlassWindow::HandleDPIEvent`, to notify the (Java) window when there is a DPI change event, which can happen when the user changes the resolution of the screen (via Settings -> System -> Display -> scale), while the JavaFX application is running.

When such `WM_DPICHANGED` event happens, `GlassWindow::HandleDPIEvent` notifies the (Java) window, which changes its platform scale via `Window::notifyScaleChanged`, and `GlassScreen::HandleDisplayChange();` is needed too, to update the platform scale of the screen where the window is at as well.

There are no tests added to this PR, since these would require manual intervention to change the display. In any case, the test case added to the [issue](https://bugs.openjdk.org/browse/JDK-8346281) runs fine now when the app runs on a given screen and the user changes its resolution. 

There is a follow-up issue after this PR, that might require a more complex fix, for the case where the user changes the resolution of a different screen that is placed before the one that has the application (as location coordinates of the latter depend on the former), because there is no `WM_DPICHANGED` event in this case.

---------
### Progress

### Issue
 * [JDK-8346281](https://bugs.openjdk.org/browse/JDK-8346281): [Windows] RenderScale doesn't update to HiDPI changes (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1668/head:pull/1668` \
`$ git checkout pull/1668`

`$ git checkout pull/1668` \
`$ git pull https://git.openjdk.org/jfx.git pull/1668/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1668`

`$ git pr show -t 1668`

</details>

<a href="https://git.openjdk.org/jfx/pull/1668.diff">https://git.openjdk.org/jfx/pull/1668.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1668#issuecomment-2559441958)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/win/GlassWindow.cpp
+++ b/modules/javafx.graphics/src/main/native-glass/win/GlassWindow.cpp
@@ -209,6 +209,10 @@ LRESULT CALLBACK GlassWindow::CBTFilter(int nCode, WPARAM wParam, LPARAM lParam)
     return ::CallNextHookEx(GlassWindow::sm_hCBTFilter, nCode, wParam, lParam);
 }
 
+#ifndef USER_DEFAULT_SCREEN_DPI
+#define USER_DEFAULT_SCREEN_DPI 96
+#endif
+
 #ifndef WM_DPICHANGED
 #define WM_DPICHANGED       0x02E0
 #endif
@@ -359,6 +363,10 @@ LRESULT GlassWindow::WindowProc(UINT msg, WPARAM wParam, LPARAM lParam)
                     break;
             }
             HandleViewSizeEvent(GetHWND(), msg, wParam, lParam);
+            break;
+        case WM_DPICHANGED:
+            HandleDPIEvent(wParam, lParam);
+            GlassScreen::HandleDisplayChange();
             break;
         case WM_MOVING:
             m_winChangingReason = WasMoved;
@@ -739,6 +747,15 @@ void GlassWindow::HandleSizeEvent(int type, RECT *pRect)
 
     env->CallVoidMethod(m_grefThis, midNotifyResize,
                         type, pRect->right-pRect->left, pRect->bottom-pRect->top);
+    CheckAndClearException(env);
+}
+
+void GlassWindow::HandleDPIEvent(WPARAM wParam, LPARAM lParam)
+{
+    JNIEnv* env = GetEnv();
+    float scale = (float) LOWORD(wParam) / (float) USER_DEFAULT_SCREEN_DPI;
+
+    env->CallVoidMethod(m_grefThis, midNotifyScaleChanged, scale, scale, scale, scale);
     CheckAndClearException(env);
 }
 


modules/javafx.graphics/src/main/native-glass/win/GlassWindow.cpp
```diff
@@ -209,6 +209,10 @@ LRESULT CALLBACK GlassWindow::CBTFilter(int nCode, WPARAM wParam, LPARAM lParam)
     return ::CallNextHookEx(GlassWindow::sm_hCBTFilter, nCode, wParam, lParam);
 }
 
+#ifndef USER_DEFAULT_SCREEN_DPI
+#define USER_DEFAULT_SCREEN_DPI 96
```
andy-goryachev-oracle, line 213: shouldn't this value come from some windows header?

andy-goryachev-oracle:
please add the test case to https://bugs.openjdk.org/browse/JDK-8346281

andy-goryachev-oracle:
Does not seem to work.  Here is my scenario:

1. win11 with two monitors, the built-in laptop screen at scale 100%, the external monitor at 225% (located "above" the build-in one).
2. the monkey tester opens up on the external monitor (since it remembers its position between launches)
3. when the user clicks on the menu, the menu pops up in the wrong location (way outside of the main window in fact)
4. change resolution to 200%
5. click on the menu, the menu also appears in the wrong location
6. it's only after I drag the window to the main monitor the menus start appearing at the right location.  drag it back to the external monitor, the menu appears at the correct location again.
7. change external scale back to 225%, the menu opens up at the right location.

jperedadnr:
I've added the test case to the JBS issue (sorry about that, I thought I did when I created it).

About the test case scenario:
- you have a JavaFX application already running in a given monitor
- without changing resolution, the label shows the current output scale, and the menu popup shows up in the correct position (if it doesn't, this is a different issue than this one).
- Then, after showing the menu at least once, you change the screen resolution (the screen that has the app, that is)

With this PR you should see: 
- the label showing the correct output scale once you change the resolution (this doesn't happen without the PR),
- and the menu popup showing up at the correct position (this doesn't happen without the PR).

andy-goryachev-oracle:
hmm, not sure where the disconnect is.
- before I launch the monkey tester, there is no fx app running anywhere
- the monkey tester appears on the external monitor, it's status bar shows scale 2.25
- the menu appears in the wrong location

let me try attaching the screenshot:

jperedadnr:
Do you have the same issue (unexpected offset for the popup) as well without this PR?

andy-goryachev-oracle:
I think the difference between your scenario and mine is that the monkey tester appears on the external monitor instead of the main one.

If I close the monkey tester while it is on the primary laptop screen and restart it, the menu shows up at the right location from the start.

Maybe you can change your test app to open one window on each screen?

andy-goryachev-oracle:
> Do you have the same issue (unexpected offset for the popup) as well without this PR?

The behavior is different.  With the MT showing up on the external monitor, the initial location of the menus is correct.  Once I change the scale, the menus pop up in two wrong location *above and to the left* (the change in scale was 2.25 -> 1.5)

andy-goryachev-oracle:
attached an updated SCCE to the ticket.  the initial scale value seems to be incorrect in this PR.

kevinrushforth:
I can review / test this tomorrow. There are various cases to consider even without changing the scale, and we need to confirm that there are no regressions.

jperedadnr:
Thanks for the link @kevinrushforth , it helps a lot clarifying a few things. 

I didn't find issues with `WM_DPICHANGED` (yet), but I didn't test those edge cases that are mentioned.

I see several different issues related to DPI changes (with and without my patch), and I can reproduce the issue that @andy-goryachev-oracle  mentioned, so definitely this PR needs more work, and probably some follow-ups.

PR #1679, opened 2025-01-15
8347753: VetoableListDecorator doesn't accept its own sublists for bulk operations

mstr2:
Passing a `VetoableListDecorator.subList()` to any of its bulk operations (`addAll`, `setAll`, `removeAll`, `retainAll`) throws `ConcurrentModificationException`. The reason is that the `VetoableListDecorator.modCount` field is incremented before the underlying list's bulk operation is invoked, which causes a mismatch when the sublist is interrogated by the bulk operation.

However, simply updating the `modCount` field _after_ the underlying list was modified also doesn't work, as in this case listeners can't see the correct value for `modCount` in their callback. The fix is to make a defensive copy of the sublist before invoking the underlying list's bulk operation.

/reviewers 2

---------
### Progress

### Issue
 * [JDK-8347753](https://bugs.openjdk.org/browse/JDK-8347753): VetoableListDecorator doesn't accept its own sublists for bulk operations (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1679/head:pull/1679` \
`$ git checkout pull/1679`

`$ git checkout pull/1679` \
`$ git pull https://git.openjdk.org/jfx.git pull/1679/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1679`

`$ git pr show -t 1679`

</details>

<a href="https://git.openjdk.org/jfx/pull/1679.diff">https://git.openjdk.org/jfx/pull/1679.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1679#issuecomment-2591391894)
</details>


modules/javafx.base/src/main/java/com/sun/javafx/collections/VetoableListDecorator.java
```diff
@@ -387,14 +387,37 @@ public int hashCode() {
         return list.hashCode();
     }
 
-    private class VetoableSubListDecorator implements List<E> {
+    /**
+     * Returns the specified collection as an unmodifiable list that can safely be used in all bulk
```
andy-goryachev-oracle, line 391: Do you think it might be easier to create a defensive copy **always**?

In other words, can we guarantee that it is impossible for the user to create a convoluted code involving maybe two `VetoableListDecorators` where the second one loops back the changes to the first one, however ridiculous that might sound?

mstr2: The way I see it, the situation that erroneously triggers `ConcurrentModificationException` only happens when `VetoableListDecorator` accesses its own sublist:
```java
    try {
        modCount++;
        boolean ret = list.addAll(index, c); // --> c is its own sublist
        ...
```

Since `modCount` is modified first, and the sublist refers back to the same modified `modCount`, the exception occurs. It can't occur when we are dealing with another list (or a sublist of another list), since in this case there is no self-referential conflict.

The way `ArrayList` circumvents this problem is by incrementing `modCount` only after the operation is done, not before it has started; it doesn't create a defensive copy.

andy-goryachev-oracle: Thank you for clarification!  I'll try to come up with a test (and I think even if I succeed, it does not mean that your solution is not good, since the caller can always create a defensive copy themselves).

BTW, what were you doing to trigger this bug?

mstr2: > BTW, what were you doing to trigger this bug?

This happened when I was writing some tests for the other `VetoableListDecorator` PRs...

modules/javafx.base/src/test/java/test/javafx/collections/VetoableObservableListTest.java
```diff
@@ -205,12 +205,24 @@ public void testAddAll() {
         assertSingleCall(new String[] {"a", "b"}, new int[] {4,4});
     }
 
+    @Test
+    public void testAddAll_subList() {
+        list.addAll(list.subList(0, 2));
+        assertSingleCall(new String[] {"foo", "bar"}, new int[] {4, 4});
+    }
```
andy-goryachev-oracle, line 212: suggestion: also check that the list contains the newly added elements?
(here and in added tests that involve subList?)

andy-goryachev-oracle (commented):
The scenario I was referring to is impossible, I think.

One possible alternative is to create the defensive copy each time, this will save one extra pointer every time an iterator or a sublist gets created (these objects might be long lived).  The code in this PR creates a copy in many (most?) cases anyway, and in my opinion, the memory is more precious resource that CPU cycles (i.e. using extra memory costs many more CPU cycles in garbage collection etc.), so please consider that.

+ some minor suggestions

mstr2:
> One possible alternative is to create the defensive copy each time, this will save one extra pointer every time an iterator or a sublist gets created (these objects might be long lived). The code in this PR creates a copy in many (most?) cases anyway, and in my opinion, the memory is more precious resource that CPU cycles (i.e. using extra memory costs many more CPU cycles in garbage collection etc.), so please consider that.

I don't quite understand what you mean. Can you elaborate?

modules/javafx.base/src/test/java/test/javafx/collections/VetoableObservableListTest.java
mstr2: I've added checks for the list content in all modified tests.

andy-goryachev-oracle:
> Can you elaborate?
> what does "save one extra pointer" mean?

What I mean is simply get rid of the extra pointer like `VetoableListIteratorDecorator.parent` and the related logic, and simply create a copy of List/Collection each time.

mstr2:
> > Can you elaborate?
> > what does "save one extra pointer" mean?
> 
> What I mean is simply get rid of the extra pointer like `VetoableListIteratorDecorator.parent` and the related logic, and simply create a copy of List/Collection each time.

Okay, I see. We can't do that, because a sublist is specified to be a live view onto the source list. After all, `list.subList(0, 3).clear()` is valid code, and the result is that three elements are removed from the source list.

andy-goryachev-oracle:
> We can't do that, because a sublist is specified to be a live view onto the source list.

good point, thanks for pointing this out.

andy-goryachev-oracle (approved):
lgtm

kevinrushforth (approved):
Code changes look good. I confirm that the new tests fail without the fix and pass with the fix.

PR #1685, opened 2025-01-28
8348744: Application window not always activated on macOS 15

kevinrushforth:
The fix for [JDK-8319066](https://bugs.openjdk.org/browse/JDK-8319066) added a needed call to activate a JavaFX window on macOS 14 or later via `[NSApp activate]`. In macOS 15, this doesn't always activate the window; even on macOS 14 there are certain cases where it might not. In all other places where we need to activate a window we call `[NSApp activateIgnoringOtherApps:YES]`. This is also what AWT uses in all places.

The fix is to replace the call to `activate` with a call to `activateIgnoringOtherApps`. I ran a full set of headful tests on macOS 15, 14, and 13 (although the code is not executed on  macOS 13), and everything looks good on my end.

Worth noting is that `NSApp::activateIgnoringOtherApps` is deprecated as of MacOS SDK 14, which means that we might have problems in the future when we update the Xcode toolchain (or if they eventually degrade it so that the `ignoringOtherApps` part of this call stops doing anything), but that will be a problem we need to address anyway, affecting other places in  JavaFX and AWT.

I intend to backported this fix to `jfx24` after it is integrated into mainline (for 25).

/reviewers 2

---------
### Progress

### Issue
 * [JDK-8348744](https://bugs.openjdk.org/browse/JDK-8348744): Application window not always activated on macOS 15 (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1685/head:pull/1685` \
`$ git checkout pull/1685`

`$ git checkout pull/1685` \
`$ git pull https://git.openjdk.org/jfx.git pull/1685/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1685`

`$ git pr show -t 1685`

</details>

<a href="https://git.openjdk.org/jfx/pull/1685.diff">https://git.openjdk.org/jfx/pull/1685.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1685#issuecomment-2618893565)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
@@ -283,11 +283,8 @@ jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved)
         // but it doesn't get activated, so this is needed:
         LOG("-> need to active application");
         dispatch_async(dispatch_get_main_queue(), ^{
-            [NSApp performSelector: @selector(activate)];
+            [NSApp activateIgnoringOtherApps:YES];
         });
-        // TODO: performSelector is used only to avoid a compiler
-        // warning with the 13.3 SDK. After updating to SDK 14
-        // this can be converted to a standard call.
     }
 }
 


kevinrushforth:
Reviewers: @andy-goryachev-oracle @jperedadnr

modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
```diff
@@ -283,11 +283,8 @@ - (void)applicationDidFinishLaunching:(NSNotification *)aNotification
         // but it doesn't get activated, so this is needed:
         LOG("-> need to active application");
         dispatch_async(dispatch_get_main_queue(), ^{
-            [NSApp performSelector: @selector(activate)];
+            [NSApp activateIgnoringOtherApps:YES];
```
andy-goryachev-oracle, line 286: should we add a comment about possible removal of this API in newer platforms?
(I would suggest also include the JBS reference so the context can be obtained by the person who will be fixing this code in the future).

andy-goryachev-oracle (approved):
works as expected (test app stage gets focused even when the terminal window running the test script loses focus)  on macOS 15.1.1 M1, running the following script:

```
#! /bin/bash

set -x

sleep 5
gradle --continue --info -PBUILD_TOOLS_DOWNLOAD_SCRIPT=../build-tools.gradle -PTEST_ONLY=true -PFULL_TEST=true -PUSE_ROBOT=true :systemTests:test --tests CSSRoundingTest
```

modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
kevinrushforth: I don't want to do either of these in this PR. My comment in the PR Description about possible API removal is independent of this bug and something I just happened to notice -- plus it's speculative on my part. I will file a follow-on place-holder bug for this, but there isn't anything actionable at this point. Also, I don't like the practice of putting bug IDs in for fixed bugs except in unusual cases (this isn't one of them).

andy-goryachev-oracle: Yes, one can typically rely on git history to make connection to the original PR/JBS.  The problem is that a merge might make this information not (easily) accessible.
And yes, the comments are subject to the usual decay, when they lose the relevance (or stop being applicable), but at least one can easily find the original issue and related discussion.

Just my two yen.

jperedadnr, line 286: I see that the CanvasTest mentioned in the JBS issue passes now with this fix, and fails with `[NSApp activate];`

I'm not sure why `[NSApp activateIgnoringOtherApps:YES];` works but `[NSApp activate];` doesn't, given that the former is deprecated, as you already mentioned, and the latter is what Apple precisely [recommends](https://developer.apple.com/documentation/appkit/nsapplication/activate(ignoringotherapps:)?language=objc) to use instead.
In any case, probably something for https://bugs.openjdk.org/browse/JDK-8348878.

jperedadnr (approved):

modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
andy-goryachev-oracle: @jperedadnr can you add this comment to https://bugs.openjdk.org/browse/JDK-8348878 please?

kevinrushforth: > I'm not sure why [NSApp activateIgnoringOtherApps:YES]; works but [NSApp activate]; doesn't, given that the former is deprecated, as you already mentioned, and the latter is what Apple precisely [recommends](https://developer.apple.com/documentation/appkit/nsapplication/activate(ignoringotherapps:)?language=objc) to use instead.

The default activation mode is to only activate only when no other app is active, after first deactivating the launching application. This is true for the new `activate` method as well as the existing `activateWithOptions` method. `[NSApp activate]` is equivalent to `[NSApp activateIgnoringOtherApps:NO]`.

The `activateIgnoringOtherApps` method says this:

> By default, activation deactivates the calling app (assuming it was active), and then the new app is activated only if there’s no currently active application. This prevents the new app from stealing focus from the user, if the app is slow to activate and the user has switched to a different app in the interim. However, if you specify NSApplicationActivateIgnoringOtherApps, the application is activated regardless of the currently active app, potentially stealing focus from the user.

Let's continue looking into this in the context of [JDK-8348878](https://bugs.openjdk.org/browse/JDK-8348878).

PR #1690, opened 2025-01-29
8348423: [TestBug] stress test Nodes initialization from a background thread

andy-goryachev-oracle:
Created a test that validates various Nodes can be initialized in a background thread.

---------
### Progress

### Issue
 * [JDK-8348423](https://bugs.openjdk.org/browse/JDK-8348423): [TestBug] stress test Nodes initialization from a background thread (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1690/head:pull/1690` \
`$ git checkout pull/1690`

`$ git checkout pull/1690` \
`$ git pull https://git.openjdk.org/jfx.git pull/1690/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1690`

`$ git pr show -t 1690`

</details>

<a href="https://git.openjdk.org/jfx/pull/1690.diff">https://git.openjdk.org/jfx/pull/1690.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1690#issuecomment-2625950431)
</details>

### DIFF
--- a/tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationBackgroundThreadTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationBackgroundThreadTest.java
@@ -1,0 +1,252 @@
+/*
+ * Copyright (c) 2025, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package test.robot.javafx.scene;
+
+import static org.junit.jupiter.api.Assertions.fail;
+import java.time.LocalDate;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+import javafx.application.Platform;
+import javafx.geometry.Point2D;
+import javafx.geometry.Pos;
+import javafx.scene.Node;
+import javafx.scene.control.Accordion;
+import javafx.scene.control.DatePicker;
+import javafx.scene.control.Label;
+import javafx.scene.control.PasswordField;
+import javafx.scene.control.TextArea;
+import javafx.scene.control.TextField;
+import javafx.scene.control.TextInputControl;
+import javafx.scene.control.TitledPane;
+import javafx.scene.control.Tooltip;
+import javafx.scene.control.skin.AccordionSkin;
+import javafx.scene.control.skin.DatePickerSkin;
+import javafx.scene.control.skin.LabelSkin;
+import javafx.scene.control.skin.TextAreaSkin;
+import javafx.scene.control.skin.TextFieldSkin;
+import javafx.scene.layout.BorderPane;
+import javafx.util.Duration;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import test.robot.testharness.RobotTestBase;
+
+/**
+ * Tests Node initialization from a background thread, per
+ *
+ * https://openjfx.io/javadoc/23/javafx.graphics/javafx/scene/Node.html
+ *
+ * "Node objects may be constructed and modified on any thread as long they are not yet attached to a Scene in a Window
+ * that is showing. An application must attach nodes to such a Scene or modify them on the JavaFX Application Thread."
+ */
+public class NodeInitializationBackgroundThreadTest extends RobotTestBase {
+    private static final int THREAD_COUNT = 100;
+    private static final int DURATION = 2000;
+    private static final AtomicLong seq = new AtomicLong();
+    private static final AtomicBoolean failed = new AtomicBoolean();
+
+    @Test
+    public void accordion() {
+        test(() -> {
+            Accordion c = new Accordion();
+            c.setSkin(new AccordionSkin(c));
+            c.getPanes().add(new TitledPane("Accordion", new BorderPane()));
+            return c;
+        }, (c) -> {
+            c.prefHeight(-1);
+            c.prefWidth(-1);
+        });
+    }
+
+    @Disabled("JDK-8349004") // FIX
+    @Test
+    public void datePicker() {
+        test(() -> {
+            DatePicker c = new DatePicker();
+            c.setSkin(new DatePickerSkin(c));
+            return c;
+        }, (c) -> {
+            c.show(); // fails here
+            c.setValue(LocalDate.now());
+            c.prefHeight(-1);
+            c.setValue(LocalDate.EPOCH);
+            c.prefWidth(-1);
+        });
+    }
+
+    @Test
+    public void passwordField() {
+        test(() -> {
+            PasswordField c = new PasswordField();
+            c.setSkin(new TextFieldSkin(c));
+            return c;
+        }, (c) -> {
+            // could not get it to fail
+            access(c);
+            c.setAlignment(Pos.CENTER);
+            c.getCharacters();
+        });
+    }
+
+    @Disabled("JDK-8347392") // FIX
+    @Test
+    public void textArea() {
+        test(() -> {
+            TextArea c = new TextArea();
+            c.setSkin(new TextAreaSkin(c));
+            return c;
+        }, (c) -> {
+            access(c);
+        });
+    }
+
+    @Test
+    public void textField() {
+        test(() -> {
+            TextField c = new TextField();
+            c.setSkin(new TextFieldSkin(c));
+            return c;
+        }, (c) -> {
+            // could not get it to fail
+            access(c);
+            c.setAlignment(Pos.CENTER);
+            c.getCharacters();
+        });
+    }
+
+    @Disabled("JDK-8348100") // FIX
+    @Test
+    public void tooltip() {
+        test(() -> {
+            Tooltip t = new Tooltip("this is a tooltip");
+            t.setShowDelay(Duration.ZERO);
+            t.setHideDelay(Duration.ZERO);
+            Label c = new Label("testing tooltip");
+            c.setSkin(new LabelSkin(c));
+            c.setTooltip(t);
+            return c;
+        }, (c) -> {
+            Tooltip t = c.getTooltip();
+            t.isShowing();
+            t.setGraphic(new Label("yo!"));
+            if (Platform.isFxApplicationThread()) {
+                Point2D p = c.localToScreen(c.getWidth() / 2.0, c.getHeight() / 2.0);
+                robot.mouseMove(p);
+            }
+        });
+    }
+
+    private void access(TextInputControl c) {
+        c.setPrefWidth(20);
+        c.setPromptText("yo");
+        c.setText(nextString());
+        c.prefHeight(-1);
+        c.getControlCssMetaData();
+    }
+
+    private <T extends Node> void test(Supplier<T> generator, Consumer<T> operation) {
+        T visibleNode = generator.get();
+        String title = visibleNode.getClass().getSimpleName();
+        
+        setContent(visibleNode);
+        setTitle(title);
+
+        AtomicBoolean running = new AtomicBoolean(true);
+        CountDownLatch counter = new CountDownLatch(THREAD_COUNT);
+
+        try {
+            for (int i = 0; i < THREAD_COUNT; i++) {
+                new Thread(title) {
+                    @Override
+                    public void run() {
+                        try {
+                            T n = generator.get();
+                            int count = 0;
+                            while (running.get()) {
+                                operation.accept(n);
+    
+                                count++;
+                                if ((count % 100) == 0) {
+                                    inFx(() -> {
+                                        operation.accept(visibleNode);
+                                    });
+                                }
+                            }
+                        } finally {
+                            counter.countDown();
+                        }
+                    }
+                }.start();
+            }
+
+            sleep(DURATION);
+        } finally {
+            running.set(false);
+        }
+
+        // let them finish
+        try {
+            counter.await(500, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            fail(e);
+        }
+    }
+
+    private static String nextString() {
+        long ix = seq.incrementAndGet();
+        if ((ix % 10) == 0) {
+            return null;
+        }
+        return "_a" + ix + "\nyo!";
+    }
+
+    @BeforeAll
+    public static void beforeAll() {
+        // this might be made a part of the base class (RobotTestBase)
+        Thread.setDefaultUncaughtExceptionHandler((t, e) -> {
+            e.printStackTrace();
+            failed.set(true);
+            // we could also accumulate stack trace(s) and send them to fail() in afterEach()
+        });
+    }
+
+    @BeforeEach
+    public void beforeEach() {
+        failed.set(false);
+    }
+
+    @AfterEach
+    public void afterEach() {
+        if (failed.get()) {
+            fail();
+        }
+    }
+}
--- a/tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationBackgroundThreadTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationBackgroundThreadTest.java
@@ -144,10 +144,11 @@ import test.robot.testharness.RobotTestBase;
  * - WebView
  *
  * NOTE: I suspect this test might be a bit unstable and/or platform-dependent, due to its multi-threaded nature.
+ *
+ * TODO add remaining Nodes to the test.
  */
 public class NodeInitializationBackgroundThreadTest extends RobotTestBase {
-    private static final int THREAD_COUNT = 100;
-    private static final int DURATION = 2000;
+    private static final int DURATION = 5000;
     private static final AtomicLong seq = new AtomicLong();
     private static final AtomicBoolean failed = new AtomicBoolean();
 
@@ -161,7 +162,7 @@ public class NodeInitializationBackgroundThreadTest extends RobotTestBase {
         }, this::accessControl);
     }
 
-    @Disabled("JDK-8349091")
+    @Disabled("JDK-8349091") // FIX
     @Test
     public void areaChart() {
         test(() -> {
@@ -174,7 +175,7 @@ public class NodeInitializationBackgroundThreadTest extends RobotTestBase {
         });
     }
 
-    @Disabled("JDK-8349091")
+    @Disabled("JDK-8349091") // FIX
     @Test
     public void barChart() {
         test(() -> {
@@ -187,7 +188,7 @@ public class NodeInitializationBackgroundThreadTest extends RobotTestBase {
         });
     }
 
-    @Disabled("JDK-8349091")
+    @Disabled("JDK-8349091") // FIX
     @Test
     public void bubbleChart() {
         test(() -> {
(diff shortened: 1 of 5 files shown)


tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationBackgroundThreadTest.java
```diff
@@ -0,0 +1,906 @@
...
+
+    @Disabled("JDK-8349004") // FIX
+    @Test
+    public void datePicker() {
+        test(() -> {
+            DatePicker c = new DatePicker();
+            c.setSkin(new DatePickerSkin(c));
+            return c;
+        }, (c) -> {
+            c.show(); // fails here
```
kevinrushforth, line 305: As mentioned in the JBS issue, I think we should consider filing an issue to update the spec of `ComboBoxPopupControl::show` to require calling it on the JavaFX app thread (throwing an exception if called off thread). Either way, we should ensure that the skins themselves never call show unless the control is showing.

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationBackgroundThreadTest.java
```diff
@@ -0,0 +1,906 @@
...
+
+    @Test
+    public void colorPicker() {
+        test(() -> {
+            ColorPicker c = new ColorPicker();
+            c.setSkin(new ColorPickerSkin(c));
+            c.setValue(Color.GREEN);
+            return c;
+        }, (c) -> {
+            c.show(); // does not fail here, unlike DatePicker?
```
kevinrushforth, line 272: Interesting. I still think we should disallow it as I mentioned in the `DatePicker` case.

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationBackgroundThreadTest.java
```diff
@@ -0,0 +1,906 @@
...
+    public void menuButton() {
+        test(() -> {
+            MenuButton c = new MenuButton();
+            c.setSkin(new MenuButtonSkin(c));
+            return c;
+        }, (c) -> {
+            c.getItems().setAll(new MenuItem("MenuButton"));
+            c.setPopupSide(Side.RIGHT);
+            accessControl(c);
+            c.show();
```
kevinrushforth, line 377: This is another case of a `show` method that we should consider changing to require access only on the FX app thread.

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationBackgroundThreadTest.java
```diff
@@ -0,0 +1,906 @@
...
+        // let all threads finish
+        try {
+            counter.await(500, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            fail(e);
+        }
+    }
+
+    private static boolean nextBoolean() {
+        return new Random().nextBoolean();
```
kevinrushforth, line 779: It is a best practice when using Random to share a single instance and not allocate a new one each time. It is more performant and more random.

kevinrushforth (commented):
I like the overall structure of this test. I left several mostly "fit and finish" type comments.

I will test it as part of finishing the review. What testing have you done? I recommend the following CI headful test runs:

1. A run using this branch "as is" to verify no failures on our headful systems
2. A run using a branch with all of the `@Disabled` tests enabled to see how well the test catches the bugs.

andy-goryachev-oracle:
> What testing have you done?

All disabled tests fail on my mac, I did not run the CI job (no need, I know they fail).
Additionally, I ran the CI job with the tests disabled - this test runs fine (there were some unrelated failures on some platform)

Curiously, I am having trouble reproducing the tooltip test today - it definitely failed yesterday.  I am not sure what the difference is, so I am investigating this as a part of https://bugs.openjdk.org/browse/JDK-8348100

tests/system/src/test/java/test/robot/testharness/RobotTestBase.java
```diff
@@ -0,0 +1,190 @@
...
+
+    @BeforeAll
+    public static void initFX() throws Exception {
+        startupLatch = new CountDownLatch(1);
+        Util.launch(startupLatch, App.class);
+    }
+
+    @AfterAll
+    public static void teardownOnce() {
+        Util.shutdown();
```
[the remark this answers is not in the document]
andy-goryachev-oracle: No, this method closes all the open stages.

kevinrushforth:
> > What testing have you done?
> 
> All disabled tests fail on my mac, I did not run the CI job (no need, I know they fail). Additionally, I ran the CI job with the tests disabled - this test runs fine (there were some unrelated failures on some platform)

OK, good. I'll test on Windows.

> Curiously, I am having trouble reproducing the tooltip test today - it definitely failed yesterday. I am not sure what the difference is, so I am investigating this as a part of https://bugs.openjdk.org/browse/JDK-8348100

It's a Friday and your system is tired after a long week? Seriously, that's one of the challenges of this sort of MT testing. For many race conditions, it's not easy to verify.

tests/system/src/test/java/test/robot/testharness/RobotTestBase.java
```diff
@@ -0,0 +1,190 @@
...
+        Util.runAndWait(() -> {
+            r.run();
+        });
+    }
+
+    /**
+     * Thread.sleep() alias that fails the test if an {@code InterruptedException} gets thrown.
+     * @param milliseconds the number of milliseconds to sleep
+     */
+    protected void sleep(int milliseconds) {
```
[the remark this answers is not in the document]
andy-goryachev-oracle: it does not matter, no test should run 23 days

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationBackgroundThreadTest.java
```diff
@@ -0,0 +1,906 @@
...
+                            T n = generator.get();
+                            int count = 0;
+                            while (running.get()) {
+                                operation.accept(n);
+
+                                count++;
+                                if ((count % 100) == 0) {
+                                    inFx(() -> {
+                                        operation.accept(visibleNode);
+                                    });
```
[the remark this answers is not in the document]
andy-goryachev-oracle: yes, jiggling the visible node during the test adds stress to the system, which is the intended behavior.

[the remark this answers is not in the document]
kevinrushforth: Yes, I know. I was pointing out the fact that the frequency of the jitter will increase as the number of background threads increases. Which I think is fine.

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationBackgroundThreadTest.java
andy-goryachev-oracle: yes, we'll deal with later.
I think there might be a solution which allows show() of an unconnected popup, and if that we'll use the ticket to add a javadoc comment and modify the test.

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationBackgroundThreadTest.java
andy-goryachev-oracle: while true, the shared instance will add synchronization and this is something I would like to avoid.  The degree of randomness is not that important in this case.

the only thing that we lose is reproducibility, but in the case of a headful test it's probably impossible anyway, due to many other factors.

In any case, I think the main purpose of this test is a sort of sanity check of the requirement, not an exhaustive test of all the properties and all the methods that the application can call from a background thread - at least that what I thought at the time of writing.  We can always add more stress to it.

andy-goryachev-oracle:
> It's a Friday and your system is tired after a long week?

It is also raining today...

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationBackgroundThreadTest.java
```diff
@@ -727,8 +728,16 @@ private void accessRegion(Region c) {
     }
 
     private <T extends Node> void test(Supplier<T> generator, Consumer<T> operation) {
-        T visibleNode = generator.get();
-        String title = visibleNode.getClass().getSimpleName();
+        AtomicReference<T> ref = new AtomicReference();
+        runAndWait(() -> {
```
kevinrushforth, line 732: What was the rationale for this change? This means we are no longer testing the ability to construct a node off thread.

Related to this, I think a separate set of tests where we construct instances of the node in a loop would be useful, in addition to the set of tests that mutates the node in a loop.

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationBackgroundThreadTest.java
kevinrushforth: Yes, whatever we end up doing will be done as a follow-up.

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationBackgroundThreadTest.java
kevinrushforth: OK, that sounds reasonable. Maybe one other thing to consider would be `ThreadLocalRandom`, but it's probably OK ei ther way.

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationBackgroundThreadTest.java
kevinrushforth: I think this change is why the Tooltip test is no longer failing. The failure mode reported in [JDK-8348100](https://bugs.openjdk.org/browse/JDK-8348100) happens when  constructing Tooltip on a background thread.

andy-goryachev-oracle: The very first (visible) node should be created in the FX application thread (something I missed earlier).  All other instances are created in the background thread - notice the second call to `generator.get()` inside the background thread lambda.

andy-goryachev-oracle: > Related to this, I think a separate set of tests where we construct instances of the node in a loop would be useful, in addition to the set of tests that mutates the node in a loop.

this is a good idea!  thank you.

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationStressTest.java
```diff
@@ -0,0 +1,947 @@
...
+            Tooltip t = new Tooltip("this is a tooltip");
+            t.setShowDelay(Duration.ZERO);
+            t.setHideDelay(Duration.ZERO);
+            Label c = new Label("testing tooltip");
+            c.setSkin(new LabelSkin(c));
+            c.setTooltip(t);
+            c.setId("Tooltip");
+            return c;
+        }, (c) -> {
+            Tooltip t = c.getTooltip();
```
kevinrushforth, line 671: For the background thread, construct a new object.

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationStressTest.java
```diff
@@ -0,0 +1,947 @@
...
+            t.isShowing();
+            t.setGraphic(new Label("yo!"));
+            if (Platform.isFxApplicationThread()) {
+                Point2D p = c.localToScreen(c.getWidth() / 2.0, c.getHeight() / 2.0);
```
kevinrushforth, line 675: In order to reliably reproduce it with the manual test, I had to move the mouse off and on the tooltip (so the tooltip is shown / hidden repeatedly). You can do that here by alternately moving the mouse to the center (as you've done here) and outside the window.

kevinrushforth (commented):
I left a couple comments. Regarding the tooltip issue, I was able to make it fail reliably by doing two things:

1. When not on the FX app thread, create a new tooltip rather than using the one created in the generator
2. When on the FX app thread, alternate between moving the mouse onto / off of the control

Like this:

```
        AtomicBoolean overTooltip = new AtomicBoolean(false);
        test(() -> {
            Tooltip t = new Tooltip("this is a tooltip");
            t.setShowDelay(Duration.ZERO);
            t.setHideDelay(Duration.ZERO);
            Label c = new Label("testing tooltip");
            c.setSkin(new LabelSkin(c));
            c.setTooltip(t);
            c.setId("Tooltip");
            return c;
        }, (c) -> {
            Tooltip t;
            if (Platform.isFxApplicationThread()) {
                t = c.getTooltip();
            } else {
                t = new Tooltip("this is a tooltip");
            }
            t.isShowing();
            t.setGraphic(new Label("yo!"));
            if (Platform.isFxApplicationThread()) {
                boolean isOverTooltip = overTooltip.get();
                Point2D p = isOverTooltip ?
                        c.localToScreen(c.getWidth() + 5.0, c.getHeight() + 5.0) :
                        c.localToScreen(c.getWidth() / 2.0, c.getHeight() / 2.0);
                robot.mouseMove(p);
                overTooltip.set(!isOverTooltip);
            }
        });
```

Feel free to use / adapt / whatever this suggestion.

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationBackgroundThreadTest.java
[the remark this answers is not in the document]
andy-goryachev-oracle: I like it.

Also, I think we may want to add a bit of randomness to the process - in the "jiggler" thread, as well as to tests themselves.

kevinrushforth:
> Re: tooltip issue: I was able to modify the test to reproduce the issue reliably (thanks for earlier suggestion to create tooltips in a tight loop).

Good to know. On Windows, I could only get it to fail by also moving the mouse on / off the tooltip I commented on above.

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationBackgroundThreadTest.java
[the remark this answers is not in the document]
kevinrushforth: A bit of randomness in the "jiggler" thread seems good. Not sure about adding more randomness in the tests -- maybe, but having them free running (rather than sleeping) is probably more stressful.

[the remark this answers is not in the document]
andy-goryachev-oracle: in some cases the animation and transitions are getting involved, which might uncover additional issues.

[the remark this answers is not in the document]
kevinrushforth: That makes sense. Worth noting it that the animation and transitions only run on the app thread (although they can be started on any thread).

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationStressTest.java
andy-goryachev-oracle: I am going to address the tooltip issue in a different PR so as not to prolong the review process...

andy-goryachev-oracle:
Created https://bugs.openjdk.org/browse/JDK-8349257 for any improvements to the existing tests and also to add the remaining Node classes (shapes etc.)

andy-goryachev-oracle:
It looks like this test might depend on specifics of JIT: the very first run after re-compilation ("clean" in Eclipse) it fails (with all the tests enabled), but any number of subsequent runs pass just fine (macOS 15.1.1 M1).

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationStressTest.java
```diff
@@ -0,0 +1,979 @@
...
+
+    @Test
+    public void scrollPane() {
+        assumeFalse(SKIP_TEST);
+        test(() -> {
+            ScrollPane c = new ScrollPane(new Label("ScrollPane"));
+            c.setSkin(new ScrollPaneSkin(c));
+            return c;
+        }, (c) -> {
+            c.setPannable(nextBoolean());
```
kevinrushforth, line 486: Do you think it's worth adding some scrolling here? That might be a good follow-on test enhancement.

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationStressTest.java
```diff
@@ -0,0 +1,979 @@
...
+    public void spinner() {
+        assumeFalse(SKIP_TEST);
+        test(() -> {
+            Spinner c = new Spinner();
+            c.setSkin(new SpinnerSkin(c));
+            c.setValueFactory(new SpinnerValueFactory.IntegerSpinnerValueFactory(0, 100, 50, 1));
+            return c;
+        }, (c) -> {
+            c.setEditable(nextBoolean());
+            accessControl(c);
```
kevinrushforth, line 501: Is it worth adding code to increment / decrement the spinner?

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationStressTest.java
```diff
@@ -0,0 +1,979 @@
...
+        });
+    }
+
+    @Test
+    public void treeView() {
+        assumeFalse(SKIP_TEST);
+        Supplier<TreeItem<String>> gen = () -> {
+            TreeItem<String> root = new TreeItem<>(null);
+            return root;
+        };
```
kevinrushforth, line 712: `gen` is an unused local var.

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationStressTest.java
```diff
@@ -0,0 +1,979 @@
...
+        if (title == null) {
+            title = visibleNode.getClass().getSimpleName();
+        }
+
+        setTitle(title);
+        setContent(visibleNode);
+
+        int threadCount = 1 + Runtime.getRuntime().availableProcessors() * 2;
+        AtomicBoolean running = new AtomicBoolean(true);
+        int additionalThreads = 2; // jiggler + tight loop
```
kevinrushforth, line 775: I wonder if it is worth having more than 1 "tight loop" thread? Maybe dedicate 1/2 of the threads to a tight loop (`generator` in the loop) and 1/2 to access (`generator` once and `operation` in the loop) This is something that you could consider for a future improvement,

kevinrushforth (approved):
Overall, this looks good. I think the framework is in a good enough place to integrate it now; we can always make improvements in the future if needed.

Here are a couple general observations based on my testing (mostly on Windows):

1. Once a test fails, subsequent tests are likely to be impacted. I've seen several instances of a timeouts in later tests after a failure (when I enable a test associated with a known bug). In at least one case, the visible stage froze and became non-responsive. This isn't surprising to me, given that all tests are run in the same process. If any of the exceptions hit the FX application thread it will likely cause this behavior.
2. Some of the disabled tests only fail infrequently -- meaning that either the payload for the test or the 5 second duration might not be enough to trigger the bug reliably (although race conditions are notoriously hard to test "reliably"). We can look at these individual cases when reviewing the bug fix that will enable the corresponding test.

I left a couple in-line questions and comments. Many would be questions for follow-up work (if at all). I'll re-approve if you decide to make any changes now.

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationStressTest.java
```diff
@@ -143,32 +138,45 @@
...
  * TODO add remaining Nodes to the test.
  */
-public class NodeInitializationBackgroundThreadTest extends RobotTestBase {
+public class NodeInitializationStressTest extends RobotTestBase {
     private static final int DURATION = 5000;
     private static final AtomicLong seq = new AtomicLong();
     private static final AtomicBoolean failed = new AtomicBoolean();
+    // for debugging purposes: setting this to true will skip working tests
+    // TODO remove once all the tests pass
+    private static final boolean SKIP_TEST = false;
```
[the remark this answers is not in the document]
andy-goryachev-oracle: correct.

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationStressTest.java
```diff
@@ -0,0 +1,979 @@
...
+        int threadCount = 1 + Runtime.getRuntime().availableProcessors() * 2;
+        AtomicBoolean running = new AtomicBoolean(true);
+        int additionalThreads = 2; // jiggler + tight loop
+        CountDownLatch counter = new CountDownLatch(threadCount + additionalThreads);
+
+        try {
+            // construct nodes in a tight loop
+            new Thread(() -> {
+                try {
+                    while (running.get()) {
```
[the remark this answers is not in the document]
andy-goryachev-oracle: Probably not worth doing it - the test is relatively quick (even after we add all other  Nodes).

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationStressTest.java
andy-goryachev-oracle: follow-up

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationStressTest.java
andy-goryachev-oracle: follow-up

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationStressTest.java
andy-goryachev-oracle: this is an interesting idea, esp. for intermittent tests.  I'll keep this as is for now, but will use the idea if we encounter more stubborn tests.

kevinrushforth (approved):

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationStressTest.java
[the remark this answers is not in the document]
kevinrushforth: Also, this would only bail out of a single test method, so very likely not worth it.

PR #1667, opened 2024-12-20
8342565: [TestBug] StubTextLayout

andy-goryachev-oracle:
Changed the StubTextLayout to use product PrismTextLayout with much simplified glyph layout (via stubbed fonts).  The new layout assumes all the glyphs are squares of font size, while the bold type face produces wider glyphs (by 1 pixel).  The default font size has changed from 10 to 12 to make it closer to win/linux.

This brings the test environment closer to the product configuration and expands the capabilities of our headless testing pipeline, which will be useful for upcoming behavior tests.

Existing tests have been adjusted/reworked mainly due to default font size change.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issue
 * [JDK-8342565](https://bugs.openjdk.org/browse/JDK-8342565): [TestBug] StubTextLayout (**Enhancement** - P4)

### Reviewers
 * [Marius Hanl](https://openjdk.org/census#mhanl) (@Maran23 - Committer)
 * [Kevin Rushforth](https://openjdk.org/census#kcr) (@kevinrushforth - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1667/head:pull/1667` \
`$ git checkout pull/1667`

Update a local copy of the PR: \
`$ git checkout pull/1667` \
`$ git pull https://git.openjdk.org/jfx.git pull/1667/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 1667`

View PR using the GUI difftool: \
`$ git pr show -t 1667`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jfx/pull/1667.diff">https://git.openjdk.org/jfx/pull/1667.diff</a>

</details>
<details><summary>Using Webrev</summary>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1667#issuecomment-2616512250)
</details>

PR #1691, opened 2025-01-30
8347392: Thread-unsafe implementation of c.s.j.scene.control.skin.Utils

andy-goryachev-oracle:
Thread-safe and re-entrant implementation of Utils.

The new code still uses the static instances of Text and TextLayout for performance reasons, but adds a thread-safe mechanism to keep track of whether any of the instances is in use and when that happens, supplies a new instance instead.  This solves both thread safety and re-entrancy.

/reviewers 2

---------
### Progress

### Issue
 * [JDK-8347392](https://bugs.openjdk.org/browse/JDK-8347392): Thread-unsafe implementation of c.s.j.scene.control.skin.Utils (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1691/head:pull/1691` \
`$ git checkout pull/1691`

`$ git checkout pull/1691` \
`$ git pull https://git.openjdk.org/jfx.git pull/1691/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1691`

`$ git pr show -t 1691`

</details>

<a href="https://git.openjdk.org/jfx/pull/1691.diff">https://git.openjdk.org/jfx/pull/1691.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1691#issuecomment-2638313473)
</details>

### DIFF
--- a/.jcheck/conf
+++ b/.jcheck/conf
@@ -24,7 +24,7 @@
 [general]
 project=openjfx
 jbs=jdk
-version=jfx24
+version=jfx25
 
 [repository]
 tags=(jdk-){0,1}([1-9]([0-9]*)(\.(0|[1-9][0-9]*)){0,3})(\+(([0-9]+))|(-ga))|[1-9]((\.\d{1,3}){0,2})-((b\d{2,3})|(ga))|[1-9]u(\d{1,3})-((b\d{2,3})|(ga))
--- a/apps/samples/RichTextAreaDemo/build.xml
+++ b/apps/samples/RichTextAreaDemo/build.xml
@@ -77,28 +77,28 @@ unless built as a part of openjfx repository where the default value is sufficie
 
 	<target name="run-codearea-demo">
 		<exec executable="java">
-			<arg line="--module-path '${javafx.home}/lib' --add-modules 'javafx.base,javafx.graphics,javafx.controls,jfx.incubator.input,jfx.incubator.richtext' -classpath 'dist/${TARGET}.jar' com.oracle.demo.richtext.codearea.CodeAreaDemoApp" />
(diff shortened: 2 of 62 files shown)


andy-goryachev-oracle:
@kevinrushforth I'd suggest to take a look at this first, as it blocks some other fixes.  thanks!

kevinrushforth (approved):
The fix looks good. My testing is good as well. The now-enabled tests fail pretty reliably for me on Windows without your fix and all pass with your fix.

I left a few minor questions / comments, but will approve as is.

mstr2:
What's the rationale for using a borrow pattern here, instead of simply synchronizing access to the object in question? I don't think we use such a pattern anywhere else. Do you expect heavy contention?

andy-goryachev-oracle:
> What's the rationale for using a borrow pattern here, instead of simply synchronizing access to the object in question?

A very good question!

This pattern seemed to be the most applicable here: the solution needs to be both thread-safe and reentrant, with the expectation that in most cases the simple scenario happens where the static instance is used.

We actually use that pattern using synchronization, see `GlyphLayout::getInstance()` for example.  I think the use of AtomicBoolean is far better, so I took the same approach in #1667 .

Is there a specific scenario that would not work or cause issues?

kevinrushforth (approved):

mstr2 (approved):

andy-goryachev-oracle:
Hmmm, to further make a point: options 2 and 3 may not be re-entrant, so they probably won't work.

We haven't encounter this re-entrancy issue, but I suppose it might be possible to have a text component embedded in another text component (e.g. a TextArea embedded inside a RichTextArea for some reason).  Even if the access happens within the FX application thread, the inner layout might clobber the outer one.

Thank you for a good discussion though!

kevinrushforth:
> Hmmm, to further make a point: options 2 and 3 may not be re-entrant, so they probably won't work.

Your solution is basically what I meant by option 2. The way you did it, always using the return value of a method that will return a "safe" instance to use, is reentrant. You are right that a simple approach to option 3 would not be reentrant.

I like the solution you chose.

PR #1692, opened 2025-01-31
8333275: ComboBox: adding an item from editor changes editor text

mstr2:
When the current editor value of a `ComboBox` is added to its items list, the editor value may be changed to an entirely unrelated item in the list. However, our expectation would be that the editor reflects the newly added item instead.

Under normal circumstances, adding items to the list in front of the current item causes `selectedIndex` to be shifted by the number of items added, keeping the editor value in sync with the currently selected item.

Let's illustrate this with an example:
`items = [foo, bar, baz]`, `selectedIndex = 0`, `selectedItem = foo`

The user changes the editor value to `qux` and inserts the value at list position 0, causing the current index to be shifted to the right (in `ComboBoxSelectionModel::itemsContentObserver`):
`items = [qux, foo, bar, baz]`, `selectedIndex = 1`, `selectedItem = foo`

The index is shifted a second time in `ListViewBitSetSelectionModel::updateSelection`:
`items = [qux, foo, bar, baz]`, `selectedIndex = 2`, `selectedItem = bar`

Now `bar` is selected and shown in the editor, but we would expect `qux` instead. The fix is to detect that we're inserting the current editor value in `ComboBoxSelectionModel::itemsContentObserver`, and select the correct item directly. We also need to account for this situation in `ListViewBitSetSelectionModel`, where we check whether the current value is already contained in the added values.

/reviewers 2

---------
### Progress

### Issue
 * [JDK-8333275](https://bugs.openjdk.org/browse/JDK-8333275): ComboBox: adding an item from editor changes editor text (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1692/head:pull/1692` \
`$ git checkout pull/1692`

`$ git checkout pull/1692` \
`$ git pull https://git.openjdk.org/jfx.git pull/1692/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1692`

`$ git pr show -t 1692`

</details>

<a href="https://git.openjdk.org/jfx/pull/1692.diff">https://git.openjdk.org/jfx/pull/1692.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1692#issuecomment-2626180203)
</details>

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/ComboBox.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/ComboBox.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/ComboBox.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/ComboBox.java
@@ -598,20 +598,35 @@ public class ComboBox<T> extends ComboBoxBase<T> {
                     }
                 }
 
+                int index = -1;
                 int shift = 0;
+                T editorValue;
+
+                try {
+                    editorValue = comboBox.getConverter().fromString(comboBox.getEditor().getText());
+                } catch (RuntimeException ex) {
+                    editorValue = null;
+                }
+
                 while (c.next()) {
                     comboBox.wasSetAllCalled = comboBox.previousItemCount == c.getRemovedSize();
 
                     if (c.wasReplaced()) {
-                        // no-op
+                        index = -1;
+                    } else if (c.wasAdded() && editorValue != null && c.getAddedSubList().contains(editorValue)) {
+                        index = c.getFrom() + c.getAddedSubList().indexOf(editorValue);
                     } else if (c.wasAdded() || c.wasRemoved()) {
+                        index = -1;
+
                         if (c.getFrom() <= getSelectedIndex() && getSelectedIndex()!= -1) {
                             shift += c.wasAdded() ? c.getAddedSize() : -c.getRemovedSize();
                         }
                     }
                 }
 
-                if (shift != 0) {
+                if (index >= 0) {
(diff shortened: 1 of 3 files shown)


modules/javafx.controls/src/main/java/javafx/scene/control/ListView.java
```diff
@@ -1306,6 +1306,14 @@ public ListViewBitSetSelectionModel(final ListView<T> listView) {
                 while (c.next()) {
                     final T selectedItem = getSelectedItem();
                     final int selectedIndex = getSelectedIndex();
+                    int addedItemOffset = -1;
+
+                    if (c.wasAdded()
+                            && selectedItem != null
+                            && (addedItemOffset = c.getAddedSubList().indexOf(selectedItem)) >= 0
+                            && selectedIndex == c.getFrom() + addedItemOffset) {
+                        doSelectionUpdate = false;
```
andy-goryachev-oracle, line 1315: which scenario does this code path handle outside of tests?

I could never hit a breakpoint in line 1315, except in ensureEditorValueDoesNotChangeWhenCurrentEditorValueIsAddedToItemsList().

modules/javafx.controls/src/main/java/javafx/scene/control/ComboBox.java
```diff
@@ -598,20 +598,35 @@ public ComboBoxSelectionModel(final ComboBox<T> cb) {
                     }
                 }
 
+                int index = -1;
```
andy-goryachev-oracle, line 601: you may want to change the copyright year to 2025 in all the modified files

modules/javafx.controls/src/main/java/javafx/scene/control/ListView.java
mstr2: It's only when `ComboBox.getEditor().setText(...)` is called, which doesn't seem to immediately update the current value. This shouldn't really happen in most applications.

andy-goryachev-oracle:
I'd like a second pair of eyes on this, due to change in ListView (though the change looks reasonable and the test seems to be doing the right thing).

Also, I only tested it with the reproducer on macOS 15.1.1 M1, can we get this tested on Windows and/or Linux?

/reviewers 2

andy-goryachev-oracle (approved):

andy-goryachev-oracle:
(just noticed you already specified 2 reviewers, my "contribution" was unnecessary)

kevinrushforth (approved):
Looks good.

PR #1694, opened 2025-02-04
8335587: TextInputControl: Binding prompt text that contains linebreak causes exception

Ziad-Mid:
When binding the promptTextProperty of a TextInputControl (TextField or TextArea) to a text that contains linebreaks/newlines ("\n") the "bind" call causes a RuntimeException to be thrown, the solution to it is to unbind before calling the set(txt) method to set the new value for the property.
Also added tests to test this new fix

---------
### Progress

### Issue
 * [JDK-8335587](https://bugs.openjdk.org/browse/JDK-8335587): TextInputControl: Binding prompt text that contains linebreak causes exception (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1694/head:pull/1694` \
`$ git checkout pull/1694`

`$ git checkout pull/1694` \
`$ git pull https://git.openjdk.org/jfx.git pull/1694/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1694`

`$ git pr show -t 1694`

</details>

<a href="https://git.openjdk.org/jfx/pull/1694.diff">https://git.openjdk.org/jfx/pull/1694.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1694#issuecomment-2634183503)
</details>

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/TextInputControl.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/TextInputControl.java
@@ -320,6 +320,9 @@ public abstract class TextInputControl extends Control {
             String txt = get();
             if (txt != null && txt.contains("\n")) {
                 txt = txt.replace("\n", "");
+                if(isBound()){
+                    unbind();
+                }
                 set(txt);
             }
         }
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/TextInputControl.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/TextInputControl.java
@@ -320,7 +320,7 @@ public abstract class TextInputControl extends Control {
             String txt = get();
             if (txt != null && txt.contains("\n")) {
                 txt = txt.replace("\n", "");
-                if(isBound()){
+                if (isBound()) {
                     unbind();
                 }
                 set(txt);
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/TextInputControlTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/TextInputControlTest.java
@@ -25,11 +25,6 @@
 
 package test.javafx.scene.control;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertNull;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assertions.fail;
 import java.util.Collection;
 import java.util.List;
 import javafx.beans.property.BooleanProperty;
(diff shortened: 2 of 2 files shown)


mstr2:
In my opinion, both the original implementation, as well as the proposed fix, are defective.

The implementation seems to try to enforce the (undocumented) invariant that the prompt text cannot contain line breaks. If that is indeed a sensible invariant, then the property should reject any attempt to set an invalid value by throwing an exception. Silently changing the property value is unexpected for developers.

However, it doesn't seem to me that this is a sensible invariant. `TextFieldSkin` and `TextAreaSkin` render the prompt text with a `Text` node, which doesn't have a single-line restriction. Simply not doing anything to the prompt text would be a good choice. Another good choice would be to clean up the prompt text where it is used (i.e. in `TextFieldSkin` and `TextAreaSkin`). We already do this in other places, for example in `LabeledSkinBase::getCleanText`.

modules/javafx.controls/src/main/java/javafx/scene/control/TextInputControl.java
```diff
@@ -320,6 +320,9 @@ public String getName() {
             String txt = get();
             if (txt != null && txt.contains("\n")) {
                 txt = txt.replace("\n", "");
+                if(isBound()){
```
andy-goryachev-oracle, line 323: minor: try to match the existing formatting:
`if (isBound()) {`

andy-goryachev-oracle:
@mstr2 you bring a good point: the behavior (removing newlines) is undocumented, and it should be.

I propose to do this in a separate issue, more specifically in https://bugs.openjdk.org/browse/JDK-8335547 which intends to remove the filtering for `TextArea` since that controls is capable of displaying multi-line prompts (and `TextField`/`PasswordField` are not)

mstr2:
I guess I don't see the point of this particular ticket. This PR doesn't contain any useful change, because it is entirely subsumed in JDK-8335547. Why add code here and bother reviewers (even slightly), when the actual fix is to just remove the code? Even as an interim fix, it's mostly a waste of time for an issue that has existed for 13 years.

andy-goryachev-oracle:
> the point of this particular ticket

[NEW DRIVER. PLEASE BE PATIENT]

andy-goryachev-oracle (approved):
Looks good in the MonkeyTester.

Thank you for providing unit tests and especially covering a complete set of scenarios (from the applied logic standpoint): null, with, and without newlines.

The fix is localized and well defined, so one reviewer is probably enough, but let's wait the prescribed time period (24h) to give other people a chance to take a look, especially @mstr2 :-)

Maran23 (approved):
I agree with @mstr2, but I think fixing the exception at least is good IMO.
There are multiple locations in JFX where things are unbound to set a value. I think this should never happen, as it is unexpected for the developer.

But that is a bigger problem to solve at one point.

PR #1695, opened 2025-02-05
8337960: Improve performance of mfwrapper by reusing GStreamer media buffers for decoded video

sashamatveev:
- Added new class `CMFGSTBuffer` which can allocate memory internally or provide GStreamer allocated memory to Media Foundation.
- Added `GstBufferPool` to limit allocation of output buffers used for rendering (memory will not be allocated for each buffer, but instead will be reused from pool). Limits are 3 min buffers and 6 max buffers. During testing 3 buffers was enough.
- Changed `CoInitializeEx` to `COINIT_MULTITHREADED` as per Media Foundation requirements.
- Added error handling for `ProcessOutput` in case of https://bugs.openjdk.org/browse/JDK-8329227. With error handling `MediaPlayer` fails nicely instead of silent hang.


tiainen (approved):
All builds went fine.

modules/javafx.media/src/main/native/gstreamer/plugins/mfwrapper/mfgstbuffer.h
```diff
@@ -0,0 +1,85 @@
...
+    // GStreamer interface
+    HRESULT GetGstBuffer(GstBuffer **ppBuffer);
+    HRESULT SetCallbackData(sCallbackData *pCallbackData);
+    HRESULT SetGetGstBufferCallback(void (*function)(GstBuffer **ppBuffer,
+            long lSize, sCallbackData *pCallbackData));
+
+private:
+    HRESULT AllocateOrGetBuffer(BYTE **ppbBuffer);
+
+private:
```
kevinrushforth, line 69: Minor: isn't this redundant?

modules/javafx.media/src/main/native/gstreamer/plugins/mfwrapper/mfgstbuffer.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+    return S_OK;
+}
+
+HRESULT CMFGSTBuffer::AllocateOrGetBuffer(BYTE **ppbBuffer)
+{
+    if (ppbBuffer == NULL)
+        return E_INVALIDARG;
+
+    // If we have GStreamer get buffer cllback set, then call it to get
```
kevinrushforth, line 215: Minor typo: `cllback` --> `callback`

modules/javafx.media/src/main/native/gstreamer/plugins/mfwrapper/mfwrapper.cpp
```diff
@@ -844,6 +886,118 @@ static HRESULT mfwrapper_init_colorconvert(GstMFWrapper *decoder,
...
+        return;
+
+    // Pool might fail in case of flushing, but MF still might want buffer.
+    // It is better to give buffer to MF just in case, then fail
+    // CMFGSTBuffer::Lock().
+    (*ppBuffer) = gst_buffer_new_allocate(NULL, lSize, NULL);
+}
+
+// Gets max length of configured media buffer we using for final rendering from
+// decoder ot color convert.
```
kevinrushforth, line 911: Minor: "ot" --> "of" ?

modules/javafx.media/src/main/native/gstreamer/plugins/mfwrapper/mfgstbuffer.cpp
```diff
@@ -0,0 +1,262 @@
...
+            m_pGstBuffer = NULL;
+            return E_FAIL;
+        }
+
+        m_bUnmapGstBuffer = TRUE;
+
+        (*ppbBuffer) = m_GstMapInfo.data;
+    }
+    // Lock can be called multiple times, so if we have GStreamer buffer
+    // allocated just return it.
```
kevinrushforth, line 242: Is it actually possible that Lock can be called multiple times? If it is, then I see that you don't keep track of the number of calls to Lock, so is it correct to assume that the caller does not match Lock / Unlock calls, but rather calls Unlock only once regardless of the number of calls to Lock?

kevinrushforth (approved):
The code looks reasonable. I left a few minor questions / comments. I'll reapprove if you make any changes.

All my testing looks good (I only tested on Windows, since this is in Windows-only code).

modules/javafx.media/src/main/native/gstreamer/plugins/mfwrapper/mfgstbuffer.h
sashamatveev: Yes, fixed.

modules/javafx.media/src/main/native/gstreamer/plugins/mfwrapper/mfgstbuffer.cpp
sashamatveev: Fixed.

modules/javafx.media/src/main/native/gstreamer/plugins/mfwrapper/mfwrapper.cpp
sashamatveev: Should be "or". Fixed.

modules/javafx.media/src/main/native/gstreamer/plugins/mfwrapper/mfgstbuffer.cpp
sashamatveev: Yes, based on documentation Lock() can be called multiple times to acquire pointer and it should be valid until last Unlock() is called. The caller is responsible to match Lock / Unlock calls. So, current implementation is not correct. I will fixed it. Also, based on documentation Lock() can be called from separate threads, so it should be thread safe. I will fix it as well.

sashamatveev:
8337960: Improve performance of mfwrapper by reusing GStreamer media buffers for decoded video [v2]
- Fixed typos.
- Lock / Unlock is now thread safe and last Unlock will unmap buffer.
- AllocateOrGetBuffer() is updated to remap unmapped buffer and split into two logical sections.

kevinrushforth (approved):
This looks good to me now with the updated lock/unlock now handling multiple matched lock/unlock.

arapte (approved):
LGTM...

PR #1696, opened 2025-02-06
8348100: Tooltips cannot be instantiated on background thread

andy-goryachev-oracle:
Navigating up the parent hierarchy during CSS processing in a background thread encounters a null parent, causing a NPE.

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -200,7 +200,9 @@ final class CssStyleHelper {
         // found when matching selectors
         Styleable parent = styleable.getStyleableParent();
         for(int n=1; n<depth; n++) {
-
+            if (parent == null) {
+                return;
+            }
             // TODO: this means that a style like .menu-item:hover won't work. Need to separate CssStyleHelper tree from scene-graph tree
             if (parent instanceof Node == false) {
                 parent=parent.getStyleableParent();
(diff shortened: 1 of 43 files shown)


modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
```diff
@@ -200,7 +200,9 @@ private static void updateParentTriggerStates(Styleable styleable, int depth, Ps
         // found when matching selectors
         Styleable parent = styleable.getStyleableParent();
         for(int n=1; n<depth; n++) {
-
+            if (parent == null) {
+                return;
+            }
```
kevinrushforth, line 205: While this does prevent the exception, I think this is fixing a symptom rather than the root cause. The only reason that parent can ever be null here and in the other place you added the null check is that `Tooltip::getStyleableParent` is returning a different answer between the time it is called to calculate `depth` and the time it is used here in a loop. This is happening because `Tooltip::getStyleableParent` uses static state that is only valid when on the FX application thread.

kevinrushforth (commented):
I think the root cause of the problem is in `Tooltip::getStyleableParent`, which does not return a stable result when called from a background thread. The implementation does this:

```
        if (BEHAVIOR.hoveredNode == null) {
            return super.getStyleableParent();
        }
        return BEHAVIOR.hoveredNode;
```

Where `BEHAVIOR` is a static field of Tooltip.

This might be a better fix:

```
    @Override public Styleable getStyleableParent() {
        if (!Platform.isFxApplicationThread() || BEHAVIOR.hoveredNode == null) {
            return super.getStyleableParent();
        }
        return BEHAVIOR.hoveredNode;
    }
```

and then you don't even need the null checks you added to  `CssStyleHelper`.

andy-goryachev-oracle:
> `BEHAVIOR` is a static field of Tooltip.

You are right, this is the root cause and a better fix!  (What was I thinking?  The ticket points to it directly in the description).

Fun fact: one can add junit5 annotation `@RepeatedTest(value = 5)` to a test to make it repeat.

kevinrushforth (approved):
Looks good now. I confirm that the new test fails reliably without the fix and passes with the fix.

modules/javafx.controls/src/main/java/javafx/scene/control/Tooltip.java
```diff
@@ -25,50 +25,53 @@
...
 import javafx.scene.image.ImageView;
 import javafx.scene.input.MouseEvent;
 import javafx.scene.text.Font;
 import javafx.scene.text.TextAlignment;
 import javafx.stage.Window;
 import javafx.util.Duration;
+import com.sun.javafx.beans.IDProperty;
+import com.sun.javafx.css.StyleManager;
+import com.sun.javafx.scene.NodeHelper;
+import com.sun.javafx.stage.PopupWindowHelper;
```
mstr2, line 74: That's a lot of import noise for such a small patch...

mstr2 (approved):

PR #1697, opened 2025-02-06
8349091: Charts: exception initializing in a background thread

andy-goryachev-oracle:
Root Cause:
(Multiple) properties are getting bound to the global `Platform.accessibilityActive` property. Binding (and I say, accessing) of properties is not thread-safe.

I also changed the design a bit.  Originally, every symbol in a chart had its `focusTraversableProperty` bound to `Platform.accessibilityActive` in order to enable the accessibility navigation across the chart data points.  This is rather inefficient, as the property has to manage (thousands?) of listeners.

Instead, a single boolean property is added to each chart, with a listener added to it which iterates over data symbols to toggle the `focusTraversableProperty` directly.

The exact moment when the new property gets bound is also important, and has to happen in the FX application thread.

With this change, it is safe to create and populate charts with data in a background thread.

---

## NOTES

1. It looks like the `Platform.accessibilityActive` property never transitions back to false  after it transitioned to true.  Some say it is because there is no mechanism in the platform to get notified which cannot possibly be true.
2. The javadoc for `Platform.accessibilityActiveProperty()` method stipulates that "_This method may be called from any thread_" which is patently not true as the current implementation is simply not thread-safe.

## Note to the Reviewers

To avoid merge conflicts, the preferred order of integrations:

#1697 
#1713 
#1717

/reviewers 2

---------
### Progress

### Issue
 * [JDK-8349091](https://bugs.openjdk.org/browse/JDK-8349091): Charts: exception initializing in a background thread (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1697/head:pull/1697` \
`$ git checkout pull/1697`

`$ git checkout pull/1697` \
`$ git pull https://git.openjdk.org/jfx.git pull/1697/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1697`

`$ git pr show -t 1697`

</details>

<a href="https://git.openjdk.org/jfx/pull/1697.diff">https://git.openjdk.org/jfx/pull/1697.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1697#issuecomment-2640854663)
</details>


kevinrushforth:
Reviewers: @kevinrushforth @azuev-java

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationStressTest.java
```diff
@@ -182,26 +181,28 @@ public void areaChart() {
             c.getData().setAll(createNumberSeries());
             return c;
         }, (c) -> {
-            c.getData().setAll(createNumberSeries());
             accessChart(c);
+            c.getData().setAll(createNumberSeries());
+            c.getData().setAll(createNumberSeries());
```
azuev-java, line 186: Just curious - why do we do setAll twice in all the tests?

azuev-java:
Ok, changes look good and they seems to work on Mac OS. I am curious why do we make labels in charts focus traversable when a11y is switched on? May be something to do with accessibility on Windows? I haven't tested this patch on Windows yet. On Mac OS user can move accessibility cursor to non-focusable elements such as text labels and images so we should not have a problem navigating to the chart labels. Needs to be investigated more.

azuev-java (approved):

andy-goryachev-oracle:
> I am curious why do we make labels in charts focus traversable when a11y is switched on?

I think the reason is to announce the axis values for each data point (that's what the VoiceOver announces on macOS for me) - you can readily try it in the MonkeyTester.

andy-goryachev-oracle:
Sorry, forgot to enable the pie chart test.
On a positive note, I've tested the last commit in the cumulative branch
https://github.com/andy-goryachev-oracle/jfx/tree/ag.thread.safe.init
which includes all the fixes, the heap size lowered from 1g to 128m, and the test duration increased 3x to 15 seconds.

andy-goryachev-oracle:
Hi @kevinrushforth  and @azuev-java , could you please re-approve this PR to unblock two more?

kevinrushforth (commented):
This is heading in the right direction, but has some bugs that will need to be fixed.

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationStressTest.java
andy-goryachev-oracle: added a class-level comment.

modules/javafx.controls/src/main/java/javafx/scene/chart/Chart.java
```diff
@@ -511,6 +509,56 @@ public StyleableProperty<Boolean> getStyleableProperty(Chart node) {
...
+                accessibilityActive.addListener((src, prev, on) -> {
+                    handleAccessibilityActive(on);
+                });
+            }
+            return accessibilityActive.get();
+        } else {
+            // chart and its data are allowed to be constructed in a background thread
+            ChangeListener<Scene> li = new ChangeListener<>() {
+                @Override
+                public void changed(ObservableValue<? extends Scene> src, Scene prev, Scene scene) {
```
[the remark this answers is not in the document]
andy-goryachev-oracle: thanks for noticing!  the old code was indeed flawed.

kevinrushforth (approved):
What you have now works in all cases I've tried. I left a couple suggestions and will reapprove if you decide to make changes.

modules/javafx.controls/src/main/java/javafx/scene/chart/Chart.java
```diff
@@ -104,6 +102,9 @@ public abstract class Chart extends Region {
     /** Animator for animating stuff on the chart */
     private final ChartLayoutAnimator animator = new ChartLayoutAnimator(chartContent);
 
+    // SimpleBooleanProperty or ObjectBinding
+    private volatile Object accessibilityActive;
```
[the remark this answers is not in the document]
andy-goryachev-oracle: It's a design decision - I won't want to waste an extra pointer.
The cpu cycles are much cheaper nowadays than bytes.
Extra bytes cost much more in cpu cycles (gc) and electricity.

modules/javafx.controls/src/main/java/javafx/scene/chart/Chart.java
```diff
@@ -533,32 +535,47 @@ void updateSymbolFocusable(boolean on) {
...
-                        if (isAccessibilityActive()) {
-                            handleAccessibilityActive(true);
+            if (accessibilityActive == null) {
+                // set up a listener which, once the chart gets connected to a window (always in the context of
+                // the fx application thread), calls isAccessibilityActive() again,
+                // which in turn installs the listener on the Platform.accessibilityActiveProperty.
+                ObservableValue<Window> winProp = sceneProperty().flatMap(Scene::windowProperty);
+                accessibilityActive = winProp; // keep the reference so it won't get gc
+
+                // lambda cannot be used in place of a ChangeListener in removeListener()
```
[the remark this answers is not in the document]
andy-goryachev-oracle: I don't know how to use Subscription in this case.  
This does not work:

```
                ObservableValue<Window> winProp = sceneProperty().flatMap(Scene::windowProperty);
                accessibilityActive = winProp; // keep the reference so it won't get gc
                Subscription sub = winProp.subscribe((win) -> {
                    if (win != null) {
                        if (accessibilityActive == winProp) {
                            accessibilityActive = null;
                        }
                        if (isAccessibilityActive()) {
                            handleAccessibilityActive(true);
                        }
                        //winProp.removeListener(this);
                        sub.unsubscribe(); <-- COMPILE ERROR
                    }
                });
```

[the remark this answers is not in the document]
andy-goryachev-oracle: @hjohn could you help here please?  How could we use Subscription in a situation when it has to be unsubscribed from within the lambda?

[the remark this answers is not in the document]
hjohn: I haven't been tracking these fixes for allowing initialization in background threads, but it seems to me that basically anything should be allowed as long as you're not part of a visible scene graph -- and I think there's also no expectation that all functionality of a control "works" as long as it is not yet part of such a graph (ie. the listeners are only needed once it is part of a scene graph).

If you make listeners conditional on being part of a scene graph, then I think you can handle these with a single code path.  Such a condition would be:

    ObservableValue<Boolean> partOfSceneGraph = node.sceneProperty()
      .flatMap(Scene::windowProperty)
      .flatMap(Window::showingProperty)
      .orElse(false);
      
You can then safely install any listeners directly, regardless if they're on a background thread or not:

    someProperty().when(partOfSceneGraph).subscribe( ... );

Or:

    someProperty().when(partOfSceneGraph).addListener( ... );

On a background thread, `partOfSceneGraph` will be `false`, and no listener gets installed (yet).  As soon as it becomes `true` all listeners with that same condition become active.  It becomes `true` automatically when the node has a scene belonging to a window that is visible.  Vice versa, if it ever becomes `false` again, (which it may when the Node is removed or replaced) all listeners using this condition are removed again.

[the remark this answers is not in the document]
andy-goryachev-oracle: Thank you!

This is not what I asked though: my question was about using `Subscription` in one-off case, and more specifically, about whether it is even possible to unsubscribe from within the lambda.

What you are proposing is a slightly more involved change: for example, registering and unregistering listeners each time might introduce other issues (especially since we have no way to prioritize listeners/handlers).  It also complicates the management of skin (as skins can be added/removed), something I would rather avoid.

modules/javafx.controls/src/main/java/javafx/scene/chart/AreaChart.java
```diff
@@ -541,7 +543,7 @@ private Node createSymbol(Series<X,Y> series, int seriesIndex, final Data<X,Y> i
             symbol = new StackPane();
             symbol.setAccessibleRole(AccessibleRole.TEXT);
             symbol.setAccessibleRoleDescription("Point");
-            symbol.focusTraversableProperty().bind(Platform.accessibilityActiveProperty());
+            symbol.setFocusTraversable(isAccessibilityActive());
```
hjohn, line 546: So, if I understand correctly, we can't directly do this `bind`:

    symbol.focusTraversableProperty().bind(Platform.accessibilityActiveProperty())
    
Because on a background thread there may not yet be an initialized `Platform`?  Or perhaps, `Platform` would need to initialize and we don't want to do that yet on a background thread?

modules/javafx.controls/src/main/java/javafx/scene/chart/Chart.java
```diff
@@ -511,6 +511,71 @@ public StyleableProperty<Boolean> getStyleableProperty(Chart node) {
...
+                return p.get();
+            } else {
+                boolean aa = Platform.accessibilityActiveProperty().get();
+                SimpleBooleanProperty active = new SimpleBooleanProperty(aa);
+                accessibilityActive = active;
+                active.bind(Platform.accessibilityActiveProperty());
+                active.addListener((src, prev, on) -> {
+                    handleAccessibilityActive(on);
+                });
+                return active.get();
```
hjohn, line 550: I'm confused as to what is happening here.  A boolean property `active` is created, which is bound to `Platform.accessibilityActiveProperty` (this creates a listener on `accessibilityActiveProperty`). We add then another listener on `active`.  

What confuses me is what this is supposed to achieve, and also why we're initialising the property, but then bind it anyway... `active = new SimpleBooleanProperty(aa)` and later `active.bind` means that the variable `aa` is completely unnecessary as the `bind` will do this `get` for you.

modules/javafx.controls/src/main/java/javafx/scene/chart/AreaChart.java
hjohn: Was this solution considered:

    ObservableValue<Boolean> partOfSceneGraph = this.sceneProperty()
      .flatMap(Scene::windowProperty)
      .flatMap(Window::showingProperty)
      .orElse(false);
      
    symbol.focusTraversableProperty()
        .bind(Platform.accessibilityActiveProperty().when(partOfSceneGraph));
        
What the above code does is create a binding on the `when` statement (a listener is added on the `when` result).  However, the `when` property will not add a listener on `Platform.accessibilityActiveProperty()` until `partOfSceneGraph` is `true`.  As soon as it does though, a listener is installed on  `Platform.accessibilityActiveProperty()` and its latest value is set to `focusTraversableProperty` via the bind.

hjohn: The above does require though that `Platform.accessibilityActiveProperty()` is properly synchronized.  I think it may be a good idea to fix that first.  Perhaps all platform provided properties (if there are more) should ensure they're only initialized once.

modules/javafx.controls/src/main/java/javafx/scene/chart/Chart.java
andy-goryachev-oracle: aa is unnecessary, you are right, bind() will set the value.

the property is created as a way to signal any subsequent calls that no more work is needed.

andy-goryachev-oracle:
Modified to use Subscription (thanks, @hjohn !)

Check out the code sample

https://github.com/andy-goryachev-oracle/Test/blob/main/src/goryachev/tests/Chart_ThreadInitTest.java 

which exercises the scenario.

modules/javafx.controls/src/main/java/javafx/scene/chart/Chart.java
```diff
@@ -540,40 +541,31 @@ final boolean isAccessibilityActive() {
...
-                        if (win != null) {
-                            if (accessibilityActive == winProp) {
-                                accessibilityActive = null;
-                            }
+                // set up a subscription to be fired once the chart becomes a part of the scene graph
+                accessibilityActive = sceneProperty()
+                    .flatMap(Scene::windowProperty)
+                    .subscribe((w) -> {
+                        if (w != null) {
+                            // also unsubscribes when appears in a window
```
kevinrushforth, line 563: Suggestion: Add a comment that this can only happen on the FX app thread, since you are relying on that.

modules/javafx.controls/src/main/java/javafx/scene/chart/Chart.java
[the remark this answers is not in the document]
kevinrushforth: While I still think the code would be cleaner with two properties, and the extra memory is not significant, what you have will work, so I won't object.

kevinrushforth (commented):
I'll do some more testing, but this looks OK to me.

kevinrushforth (approved):
All my testing looks good.

modules/javafx.controls/src/main/java/javafx/scene/chart/Chart.java
andy-goryachev-oracle: added comment

modules/javafx.controls/src/main/java/javafx/scene/chart/AreaChart.java
hjohn: @kevinrushforth I don't know, I would think it should be possible to create a property or property wrapper to make it thread-safe.  Just having a version with all its methods synchronized is probably sufficient.  If that saves a lot of complex logic in, what looks like, dozens of controls, then it seems like a far better solution to me.  Of course, that should only apply to special properties; I'm not suggesting making all properties synchronized -- although I think we could as without contention, it would be nearly free, and it would make all kinds of undefined behavior scenario's when used from multiple threads (by accident or otherwise) suddenly deterministic and debuggable.

hjohn: I think the problem is in the callbacks themselves, as they'd be on the wrong thread.  I vaguely remember experimenting with a system where you can provide an Executor for callbacks (like `Platform::runLater`) to mitigate this.

kevinrushforth (approved):

modules/javafx.controls/src/main/java/javafx/scene/chart/AreaChart.java
hjohn: Yes, but that's because there is no synchronization.  Here is a version that does do synchronization; I see no more exceptions (and it runs just as fast really):

```java
public class PropTest {
    Thread mainThread = Thread.currentThread();
    final var prop = new SimpleBooleanProperty(false) {
      @Override
      public synchronized void addListener(ChangeListener<? super Boolean> listener) {
        super.addListener(listener);
      }

      @Override
      public synchronized void set(boolean newValue) {
        super.set(newValue);
      }
    };

    // add listeners in a background thread
    var thr = new Thread(() -> {
        for (int i = 0; i < 1000000; i++) {
            ChangeListener<Boolean> listener = (obs, o, n) -> {
                if (!mainThread.equals(Thread.currentThread())) {
                    System.err.println("***** ERROR: listener called on wrong thread: " + Thread.currentThread());
                }
            };
            prop.addListener(listener);
        }
    });
    thr.start();

    // Fire events in main thread
    for (int jj = 0; jj < 1000000; jj++) {
        prop.set(!prop.get());
    }
  }
}
```

hjohn: Of course for a generic solution, we could provide a wrapper for properties, or custom synchronized properties.

kevinrushforth: Yes, using a custom property, or a wrapper, would solve this particular synchronization problem, although that wouldn't solve all of the problems. 

We would then be left with the problem that if the accessibility change listener did fire -- which necessarily happens on the FX app thread -- while the object was being set up on a background thread, the listener might try to modify properties that are being touched on the background thread. This is just another case of a more general problem we already have with animation, and which Andy fixed by not starting animation in a few places unless and until we are on the FX app thread.

So I think the solution Andy has chosen of deferring adding the listener is better than trying to fix the accessibility property, followed by fixing the listeners that use it, to be thread-safe.

andy-goryachev-oracle: Thank you, @kevinrushforth .

Getting back to the rule of allowing Node construction from a background thread but not "use", should we consider enforcing the thread access rules for global and static objects (like Platform properties)?

In other words, constructing objects that might modify internal properties is ok, but trying to access shared objects is not?

(maybe it's time to move the discussion out of this PR into the mailing list perhaps?)

azuev-java:
I finished testing of the last iteration and i do not see any regressions in the accessibility on Mac or Windows - seems to be working fine on both platforms.

hjohn:
> @hjohn thank you for help with `Subscription`. Do you have any other comments?

No, but admittedly I did not look beyond the subscription issue.  I think Kevin and Alexander have got you covered on this one :)

PR #1698, opened 2025-02-06
8349105: Pagination: exception initializing in a background thread

andy-goryachev-oracle:
## Root Cause
Animation gets started in a background thread, which causes the animation handler to run in the FX application thread, thus creating simultaneous access to the control's fields (list of children in this case).

## Solution
Postpone the animation unless running in the FX application thread.  There is no functional difference if the component is created/used in the FX application thread.

---------
### Progress

### Issue
 * [JDK-8349105](https://bugs.openjdk.org/browse/JDK-8349105): Pagination: exception initializing in a background thread (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1698/head:pull/1698` \
`$ git checkout pull/1698`

`$ git checkout pull/1698` \
`$ git pull https://git.openjdk.org/jfx.git pull/1698/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1698`

`$ git pr show -t 1698`

</details>

<a href="https://git.openjdk.org/jfx/pull/1698.diff">https://git.openjdk.org/jfx/pull/1698.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1698#issuecomment-2641008505)
</details>


kevinrushforth (commented):
The test runs out of heap space for me on Windows with or without your fix using the default heap size. This may point to a leak, although bumping up the memory causes it to pass with your fix.

The fix looks good, although the test doesn't catch the bug most of the time (even when bumping up the heap).

andy-goryachev-oracle:
> The test runs out of heap space for me on Windows

This may be a memory leak, or it could simply be expected - after all, it is a stress test and we do create heavy objects in a tight loop.

I think it's worth investigating in the context of this ticket.  Thank you!

kevinrushforth:
> This may be a memory leak, or it could simply be expected - after all, it is a stress test and we do create heavy objects in a tight loop.
> 
> I think it's worth investigating in the context of this ticket. Thank you!

Yep, it does needs to be investigated for this PR, since we can't enable a test that will cause OOM when running the tests normally.

I'll be interested to know what you find.

andy-goryachev-oracle:
I am unable to reproduce OOM on Windows 11 with the fix (I did see the OOM earlier when it was failing before the fix).

I've tried the unmodified test, as well as the extended version with the DURATION doubled to 10 seconds and added `@RepeatedTest(value = 5)` to the `pagination()` test.  Same extended test passes just fine on macOS.

A cursory look at the Pagination/PaginationSkin code reveals no obvious candidates for memory leaks.  Maybe try running it again, now that it has the text layout fixes as well?

kevinrushforth:
I still see the OOM. Are you running it from gradle? Here is the command line that will reproduce it, after first building with: "`gradle sdk shims`":

```
gradle --info -PTEST_ONLY=true -PFULL_TEST=true -PUSE_ROBOT=true \
    :systemTests:test --tests NodeInitializationStressTest.pagination
```

kevinrushforth:
I also see the OOM on Mac.

andy-goryachev-oracle:
I see the problem with gradle (I was running it in Eclipse).  We apparently set `-Xmx512m` which is simply inadequate.  We should at least double that, or even go to 2GB.
Where is it being set?

kevinrushforth:
> I see the problem with gradle (I was running it in Eclipse). We apparently set -Xmx512m which is simply inadequate. We should at least double that, or even go to 2GB.
> 
> Where is it being set?

The default is set by the gradle installation itself, which is not controlled by us. I know you can bump the memory with a gradle option on the command line (which we do for our CI builds, but we just set it to 512Mb).

We can trivially change our CI test script, but I'd be hesitant to require every developer to set that on the command line. A better solution might be to set the max heap to 1Gb in the system tests project in `build.gradle`.

andy-goryachev-oracle:
> A better solution might be to set the max heap to 1Gb in the system tests project in `build.gradle`.

created https://bugs.openjdk.org/browse/JDK-8349679

kevinrushforth:
A quick test suggests that there is still something up with Pagination that increasing the memory won't fully resolve. Even bumping it up to 2Gb fails sometimes on my Mac (e.g., if I increase the duration to 10 sec). I added `-verbose:gc` and the memory shoots up to almost 2Gb and doesn't really ever go down. None of the other tests come even close to hitting a memory problem, and when a GC happens, the memory in use drops down to about 12-15 Mbytes (not surprising given that the entire test suite runs with the default heap size if pagination is disabled).

More study is needed, so let's pick this up next week.

andy-goryachev-oracle:
it looks like the amount of garbage generated during this test exceeds the ability of gc to collect it.

adding System.gc() to the test allows it to run even after -Xmx500m, though I still would like to increase the heap size in [JDK-8349679](https://bugs.openjdk.org/browse/JDK-8349679)

andy-goryachev-oracle:
You are probably right about `System.gc()` being called too often.  I would say though that the purpose of this test is to uncover initialization issues rather than finding other problems with the controls, so it's probably ok as is. 

For the OOME - there seems to be no leak, since the memory usage drops down to initial values either using `System.gc()` in the loop, or when running in the IDE with the large heap: the memory fills up rather quickly, then drops again down to low value when one hits "Run GC" in the VirtualVM for example.

(we also have no leak detected in the skin test and no memory issues logged against Pagination in JBS)

kevinrushforth:
> You are probably right about `System.gc()` being called too often. I would say though that the purpose of this test is to uncover initialization issues rather than finding other problems with the controls, so it's probably ok as is.

it still might be worth a follow-up bug. As it is, calling `System.gc()` each time through the loop means that you are creating far fewer objects than you would in a tight loop (by an order of magnitude based on a few quick tests I did).

> For the OOME - there seems to be no leak, since the memory usage drops down to initial values either using `System.gc()` in the loop, or when running in the IDE with the large heap: the memory fills up rather quickly, then drops again down to low value when one hits "Run GC" in the VirtualVM for example.

When I run it with `-verbose:gc` I see the memory continually increasing, although clowly. I'll try it with Visual VM and see if I spot anything. In any case, if there is a leak here, it isn't a large one (and isn't directly related to this bug), so would be something to look into as a follow-up.

andy-goryachev-oracle:
Created https://bugs.openjdk.org/browse/JDK-8349750 placeholder for collecting follow-up issues.

tests/system/src/test/java/test/robot/javafx/scene/NodeInitializationStressTest.java
```diff
@@ -408,9 +407,14 @@ public void pagination() {
...
-            c.setPageCount(100);
-            c.setCurrentPageIndex(nextInt(100));
+            int mx = 1 + nextInt(100);
             accessControl(c);
+            c.setPageCount(mx);
+            c.setCurrentPageIndex(nextInt(mx));
+            // this test generates a lot of garbage quickly
+            if (nextBoolean(0.1)) {
+                System.gc();
+            }
```
kevinrushforth, line 417: While investigation the Pagination leak I also took a closer look at this `System.gc()`. The reason this prevents the OOM has nothing to do with the GC per se, but rather it slows the loops down so much that they don't generate nearly as much garbage in the 5 seconds the test runs in. You can verify that yourself by replacing the `System.gc()` with `sleep(10)` or similar (depending on the speed of your system), and get the same effect.

I don't think there is a good reason to call `System.gc()` at all in this loop. Once the leak is fixed, it  should run fine at full speed.

kevinrushforth (commented):
I filed [JDK-8349756](https://bugs.openjdk.org/browse/JDK-8349756) to track the leak. I think the call to `System.gc()` is effective only because it slows down the tight loop.

andy-goryachev-oracle:
Please review #1705 before this one.

andy-goryachev-oracle:
For testing purposes, merged this PR with the memory leak fix https://github.com/openjdk/jfx/pull/1705 .  Once the latter is integrated, unrelated changes will disappear.

kevinrushforth:
> For testing purposes, merged this PR with the memory leak fix #1705 . Once the latter is integrated, unrelated changes will disappear.

Can you merge master so this will be realized?

andy-goryachev-oracle:
> Can you merge master so this will be realized?

done.

kevinrushforth (approved):
All good now.

mstr2 (approved):
Looks good.

PR #1699, opened 2025-02-06
8349098: TabPane: exception initializing in a background thread

andy-goryachev-oracle:
## Root Cause
Animation gets started in a background thread, which causes the animation handler to run in the FX application thread, thus creating simultaneous access to the control's fields (list of children in this case).

## Solution
Skip the animation.

The fix is similar to https://github.com/openjdk/jfx/pull/1698

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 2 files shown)


kevinrushforth:
Reviewer: @arapte 

/reviewers 2

arapte (approved):
LGTM

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
```diff
@@ -522,7 +520,7 @@ private void removeTabs(List<? extends Tab> removedList) {
                     }
                 };
 
-                if (closeTabAnimation.get() == TabAnimation.GROW) {
+                if (Platform.isFxApplicationThread() && (closeTabAnimation.get() == TabAnimation.GROW)) {
```
arapte, line 523: With this check, non Application thread cannot play the animation but in else block the `cleanup` is executed on the same thread.
Can there still be a situation when, this non-Application thread and Application thread be concurrently modifying the tab?

kevinrushforth: If you mean two threads accessing this same TabPaneSkin instance, then that's not a valid case. JavaFX objects are not thread-safe when accessed from multiple threads. This bug (and the other related bugs fixed or under review) is about making sure that multiple threads, including the JavaFX application thread, each concurrently accessing their own instance, don't interfere with each other.

So it would only be a problem if "cleanup" attempted some animation or touched static state or similar, which is doesn't look like it does.

kevinrushforth: Although, having said that, the purpose of the cleanup is to clean up after the animation. So a better fix might be to put the entire if-else inside an `if (Platform.isFxApplicationThread())` test.

kevinrushforth: Actually, no. It also calls requestLayout. I think the fix is correct as is.

kevinrushforth (approved):
The fix looks good. I confirm that the newly-enabled test fails without the fix and passes with the fix.

PR #1708, opened 2025-02-12
8349004: DatePicker: NPE in show() when initialized in a background thread

andy-goryachev-oracle:
## Root Cause

Focus is being requested in show(), even a background thread.

## Solution

Do not request focus if in a background thread.

---------
### Progress

### Issue
 * [JDK-8349004](https://bugs.openjdk.org/browse/JDK-8349004): DatePicker: NPE in show() when initialized in a background thread (**Bug** - P4) ⚠️ Issue is not open.

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1708/head:pull/1708` \
`$ git checkout pull/1708`

`$ git checkout pull/1708` \
`$ git pull https://git.openjdk.org/jfx.git pull/1708/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1708`

`$ git pr show -t 1708`

</details>

<a href="https://git.openjdk.org/jfx/pull/1708.diff">https://git.openjdk.org/jfx/pull/1708.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1708#issuecomment-2654760142)
</details>

### DIFF
--- a/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/DatePickerContent.java
+++ b/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/DatePickerContent.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2013, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -25,24 +25,26 @@
 
 package com.sun.javafx.scene.control;
 
+import static com.sun.javafx.PlatformUtil.isMac;
+import static java.time.temporal.ChronoField.DAY_OF_WEEK;
+import static java.time.temporal.ChronoField.MONTH_OF_YEAR;
+import static java.time.temporal.ChronoUnit.DAYS;
+import static java.time.temporal.ChronoUnit.MONTHS;
+import static java.time.temporal.ChronoUnit.WEEKS;
+import static java.time.temporal.ChronoUnit.YEARS;
 import java.time.DateTimeException;
 import java.time.LocalDate;
+import java.time.YearMonth;
+import java.time.chrono.ChronoLocalDate;
+import java.time.chrono.Chronology;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DecimalStyle;
-import java.time.chrono.Chronology;
-import java.time.chrono.ChronoLocalDate;
 import java.time.temporal.ChronoUnit;
 import java.time.temporal.ValueRange;
 import java.time.temporal.WeekFields;
-import java.time.YearMonth;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
-
-import static java.time.temporal.ChronoField.*;
-import static java.time.temporal.ChronoUnit.*;
-
-import com.sun.javafx.scene.control.skin.*;
 import javafx.application.Platform;
 import javafx.beans.property.ObjectProperty;
 import javafx.beans.property.SimpleObjectProperty;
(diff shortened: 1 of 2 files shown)


kevinrushforth:

/reviewers 2

kevinrushforth (commented):
I'm not sure this fix is sufficient in all cases.

As I mentioned in the bug report, I think we need to consider disallowing calling `ComboBoxBaseSkin::show` (including the overridden methods in the subclasses) from a background thread. At a minimum, we need to ensure that the `ComboBoxBaseSkin::show` implementations never call any of the `PopupWindow::show` method. In looking at the code, it seems possible that there are some cases where it might.

It is already a documented (and enforced) error to call Window::show -- and by extension, the various `PopupWindow::show` methods on a background thread. I note that the `PopupWindow::show` methods are not documented to throw, but they do throw when they call the no-arg `Window::show` method, which they do if the parent Window is visible, so a documented, fail-fast exception check is best.

So I think we need to do one of two things in each of the three overridden `ComboBoxBaseSkin::show` methods (`ComboBoxPopupControl`, `ColorPickerSkin`, `DatePickerSkin`).

1. Change the spec and implementation to throw an exception if `ComboBoxBaseSkin::show` is called on a thread other than the FX Application Thread
2. Go with the currently proposed fix, but also ensure that our implementations  of`ComboBoxBaseSkin::show` don't call any of the `PopupWindow::show` methods on a background thread

II lean toward option 1, since I can't think of a good reason for an app to call show() from a background thread.

Either way, I want to see a follow-up bug filed to document that the three `PopupWindow::show` methods must be called on the FX app thread (with a check to fail fast if they don't so that it is caught in all cases).

kevinrushforth:
I took a quick look at PR #1709 which has a similar problem with its show method, and there is a third option we should consider:

3. Change the implementation of all skin show methods to do nothing (bail out early) if not on the FX application thread. If we do this, I would want to see it documented.

In any case, I'd like to pause the reviews of the individual PRs that touch the skin's `show` method until we resolve the larger question of what to do about `*Skin::show`.

One thing that would help inform the discussion:

Why did you call show in the first place in your NodeInitStressTest (for that matter, why directly call _any_ skin methods, since apps generally don't)? Was it just something you did to stress it, or is it reflective of a normal operation that an app might do during initialization of a node on a background thread? I have a feeling it's the former, but maybe I'm missing something.

kevinrushforth:
Even after considering option 3, I prefer option 1. I can't think of a good reason to allow a show method that will fail in some case (because it is documented that an app can't show a window on a background thread), and a clear exception seems better than silently making it a no-op. @mstr2 said something similar in #1709

andy-goryachev-oracle:
> Why did you call show in the first place in your NodeInitStressTest (for that matter, why directly call _any_ skin methods, since apps generally don't)

1. `show()` javadoc says "display the popup _**aspect**_ of the user interface", which I interpreted as something that could in theory be an acceptable operation.  For example, ColorPicker does not fail in its `show()`.

2. `show()` is not a skin method (see `ComboBoxBase`).  The reason the skin is explicitly set in the test is because (in theory) we ought to support custom controls and/or custom skins.  In other words, calling `Control.setSkin()` from a background thread is also an acceptable operation.

andy-goryachev-oracle:
> 1. Change the spec and implementation to throw an exception if `ComboBoxBaseSkin::show` is called on a thread other than the FX Application Thread

I like this option!

Does it mean ColorPicker.show() should also throw?

kevinrushforth:
> > 1. Change the spec and implementation to throw an exception if `ComboBoxBaseSkin::show` is called on a thread other than the FX Application Thread
> 
> I like this option!
> 
> Does it mean ColorPicker.show() should also throw?

Yes. If we do this -- and I think we should strongly consider it -- all of the `show` methods in Control, PopupWindow, and Skin should throw on a background thread.

kevinrushforth:
I took another look today, and I am convinced that we need to change the spec and implementation of all `show` and `hide` methods to enforce the threading restriction that they must only be called on the JavaFX Application thread. I filed the following bug to do this:

[JDK-8350048](https://bugs.openjdk.org/browse/JDK-8350048): Enforce threading restrictions for show and hide methods in Window, Control, and Skin

Therefore, this PR should be withdrawn and the bug should be closed as a Duplicate of JDK-8350048.

PR #1709, opened 2025-02-12
8349096: Split/MenuButton: exception initializing in a background thread

andy-goryachev-oracle:
## Root Cause

The ContextMenu (PopupWindow) cannot be shown in a background thread.

## Solution

Bail out of `show()` if in a background thread.

---------
### Progress

### Issue
 * [JDK-8349096](https://bugs.openjdk.org/browse/JDK-8349096): Split/MenuButton: exception initializing in a background thread (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1709/head:pull/1709` \
`$ git checkout pull/1709`

`$ git checkout pull/1709` \
`$ git pull https://git.openjdk.org/jfx.git pull/1709/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1709`

`$ git pr show -t 1709`

</details>

<a href="https://git.openjdk.org/jfx/pull/1709.diff">https://git.openjdk.org/jfx/pull/1709.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1709#issuecomment-2654779936)
</details>

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/ContextMenu.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/ContextMenu.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -25,10 +25,7 @@
 
 package javafx.scene.control;
 
-import com.sun.javafx.beans.IDProperty;
-import com.sun.javafx.collections.TrackableObservableList;
-import com.sun.javafx.util.Utils;
-
+import javafx.application.Platform;
 import javafx.beans.property.ObjectProperty;
 import javafx.beans.property.ObjectPropertyBase;
 import javafx.collections.ListChangeListener.Change;
@@ -44,6 +41,9 @@ import javafx.scene.Node;
 import javafx.scene.Scene;
 import javafx.scene.control.skin.ContextMenuSkin;
 import javafx.stage.Window;
+import com.sun.javafx.beans.IDProperty;
+import com.sun.javafx.collections.TrackableObservableList;
+import com.sun.javafx.util.Utils;
 
 /**
  * <p>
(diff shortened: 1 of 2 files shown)


mstr2:
I'm not convinced that this is the right solution.

`Node` says:
```java
 * Node objects may be constructed and modified on any thread as long they are
 * not yet attached to a {@link Scene} in a {@link Window} that is
 * {@link Window#isShowing showing}.
```

It's not clear to me that invoking actions on a node should count as "modifying", and there is no specification that says it's okay to invoke actions on a non-UI thread. I'd rather have actions throw an exception if they are invoked on a background thread, rather than silently not doing anything at all.

It's one thing to configure (i.e. modify) properties of a node, and then show it in the scene graph; it's another thing to tell it to do something it can't do in its current state.

kevinrushforth:
Let's resolve the general question regarding what to do about the various `*Skin::show` methods before reviewing this particular PR. See #1708

/reviewers 2

andy-goryachev-oracle:
> It's not clear to me that invoking actions on a node should count as "modifying"

I agree with this statement.

What threw me off is the way javadoc for `show()` is currently phrased: "popup aspects" implies something else beyond the simple aspect, for example, it could be a visible property on a detail pane or selected tab/page.

Should we also remove the word "aspects" from `ComboBoxBase::show` then?

kevinrushforth:
One more reason to disallow show in this case is that ContextMenu is a descendent of PopupWindow, and whatever else we do, we need to disallow calling `PopupWindow::show` from a background thread.

kevinrushforth:
> What threw me off is the way javadoc for show() is currently phrased: "popup aspects" implies something else beyond the simple aspect, for example, it could be a visible property on a detail pane or selected tab/page.
> 
> Should we also remove the word "aspects" from ComboBoxBase::show then?

Yeah, maybe this could be made clearer at the same time.

kevinrushforth:
I took another look today, and I am convinced that we need to change the spec and implementation of all `show` and `hide` methods to enforce the threading restriction that they must only be called on the JavaFX Application thread. I filed the following bug to do this:

[JDK-8350048](https://bugs.openjdk.org/browse/JDK-8350048): Enforce threading restrictions for show and hide methods in Window, Control, and Skin

Therefore, this PR should be withdrawn and the bug should be closed as a Duplicate of JDK-8350048.

PR #1715, opened 2025-02-19
8340344: The first item in TreeView is not aligned in the beginning

Ziad-Mid:
The issue occurred because items preceding an item with children (items with a disclosure node) had different widths, which led to misalignment. This can be fixed by requesting a cell relayout whenever the disclosure node's width changes.

---------
### Progress

### Issue
 * [JDK-8340344](https://bugs.openjdk.org/browse/JDK-8340344): The first item in TreeView is not aligned in the beginning (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1715/head:pull/1715` \
`$ git checkout pull/1715`

`$ git checkout pull/1715` \
`$ git pull https://git.openjdk.org/jfx.git pull/1715/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1715`

`$ git pr show -t 1715`

</details>

<a href="https://git.openjdk.org/jfx/pull/1715.diff">https://git.openjdk.org/jfx/pull/1715.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1715#issuecomment-2859091766)
</details>


modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeCellSkin.java
```diff
@@ -207,7 +207,7 @@ public final DoubleProperty indentProperty() {
         boolean disclosureVisible = disclosureNode != null && treeItem != null && ! treeItem.isLeaf();
 
         final double defaultDisclosureWidth = maxDisclosureWidthMap.containsKey(tree) ?
-                maxDisclosureWidthMap.get(tree) : 18;   // JDK-8119169: default width of default disclosure node
+                maxDisclosureWidthMap.get(tree) : 20;   // JDK-8119169: default width of default disclosure node
```
andy-goryachev-oracle, line 210: This is probably not a good solution: I think the root cause is not in that the default value is incorrect, but in that two different values are used in the same view (i.e. that it first uses the default value, then the value computed from the disclosure node, if I read the code correctly).

Could you please investigate?

I've linked the JBS ticket mentioned in the comment, take a look at the discussion there https://bugs.openjdk.org/browse/JDK-8119169 , especially the comment about variable width disclosure node.

Ziad-Mid:
.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeCellSkin.java
```diff
@@ -246,6 +255,17 @@ public final DoubleProperty indentProperty() {
         layoutLabelInArea(x, y, w, h);
     }
 
+    VirtualFlow<?> getVirtualFlow() {
+        Parent p = getSkinnable();
+        while (p != null) {
+            if (p instanceof VirtualFlow) {
```
andy-goryachev-oracle, line 261: can be simplified:

```java
if (p instanceof VirtualFlow vf) {
  return vf;
```

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeCellSkin.java
```diff
@@ -219,6 +221,13 @@ public final DoubleProperty indentProperty() {
                 disclosureWidth = disclosureNode.prefWidth(h);
                 if (disclosureWidth > defaultDisclosureWidth) {
                     maxDisclosureWidthMap.put(tree, disclosureWidth);
+
+                    final VirtualFlow<?> flow = getVirtualFlow();
+                    for (IndexedCell cell : flow.cells) {
```
andy-goryachev-oracle, line 226: `getVirtualFlow()` might return null, right?

Ziad-Mid: Yes,
I added a check for null value

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeCellSkin.java
Ziad-Mid: I made the change , thanks

modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeViewTest.java
```diff
@@ -4075,6 +4077,50 @@ public void testRemoveTreeItemFromCollapsedAncestorKeepsSelectedItem() {
...
+        stageLoader = new StageLoader(treeView);
+        Toolkit.getToolkit().firePulse();
+
+        List<Double> values = new ArrayList<>(layoutXMap.values());
+        assertFalse(values.isEmpty(), "Expected layoutX values to be collected");
+
+        double first = values.get(0);
+        for (Map.Entry<String, Double> entry : layoutXMap.entrySet()) {
+            double x = entry.getValue();
+            assertEquals(first, x, "Alignment mismatch for item: " + entry.getKey());
```
andy-goryachev-oracle, line 4120: typically when we compare results of a floating point computation, we should not use exact comparison, due to accumulation of small errors stemming from inexact nature of floating point values.  the amount of allowed difference depends on the magnitude of values and a number of operations.  we usually throw 1e-6 or something like that.

it might be ok here, because the computation goes through the same path and same values each time.

andy-goryachev-oracle (commented):
Good job figuring out the fix and coming up with a test!

- the reproducer in the ticket works correctly after the fix
- the test fails in the master and passes with the fix

left some minor comments, but I think it is in a very good shape.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TreeViewTest.java
Ziad-Mid: that's true, I did use it at first with a tolerance of 0.01 which is minimal and it didn't make a difference in this test since it's not some complex test case. 
But I will add it and use 1e-6.
`assertEquals(first, x, 1e-6, "Alignment mismatch for item: " + entry.getKey());`

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeCellSkin.java
```diff
@@ -222,12 +222,13 @@ public final DoubleProperty indentProperty() {
...
                     maxDisclosureWidthMap.put(tree, disclosureWidth);
 
-                    final VirtualFlow<?> flow = getVirtualFlow();
+                    VirtualFlow<?> flow = getVirtualFlow();
                     if (flow != null) {
                         for (IndexedCell cell : flow.cells) {
-                            if (cell == null || cell.isEmpty()) continue;
-                            cell.requestLayout();
-                            cell.layout();
+                            if (cell != null || !cell.isEmpty()) {
```
andy-goryachev-oracle, line 228: it has to be `&&`

`if (cell != null && !cell.isEmpty()) {`

andy-goryachev-oracle (approved):
excellent work, thanks!

Ziad-Mid:
@arapte could you please take a look?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeCellSkin.java
```diff
@@ -219,6 +221,16 @@ public final DoubleProperty indentProperty() {
...
                     maxDisclosureWidthMap.put(tree, disclosureWidth);
+
+                    VirtualFlow<?> flow = getVirtualFlow();
+                    if (flow != null) {
+                        for (IndexedCell cell : flow.cells) {
+                            if (cell != null && !cell.isEmpty()) {
+                                cell.requestLayout();
+                                cell.layout();
+                            }
+                        }
```
arapte, line 232: This would perform a layout of all cells, re-layout of previous cells and layout of next cells.
but, layout of next cells is performed again as continuation layout of TreeView.
A solution would be to break the loop when `cell` is current cell being layouted, so that at least the layout of next cells is not performed twice.

Can there be a way to avoid re-layout of previous cells, too ? 🤔

arapte: Also noticed a scenario: When a TreeCell with disclosureNode is currently not shown on screen but becomes visible on scrolling. In that case the TreeCells are layout while the scroll is in progress. I see only one way to avoid this that maxDisclosureWidthMap should reflect the correct value before starting the layout.

andy-goryachev-oracle: > In that case the TreeCells are layout while the scroll is in progress.

does this create any visual artifacts?

arapte: > does this create any visual artifacts?

No artifacts, just that one can notice the re-layout and the TreeCell shift to right.

andy-goryachev-oracle: this is the expected behavior though, right?

what we **don't** want is the flicker - when the re-layout is needed but happens in another pulse.  When e.g. scrollbars need to appear/disappear it may even lead to flicker that does not end.

arapte: > This would perform a layout of all cells, re-layout of previous cells and layout of next cells. but, layout of next cells is performed again as continuation layout of TreeView. A solution would be to break the loop when `cell` is current cell being layouted, so that at least the layout of next cells is not performed twice.

Would it be possible to avoid re-layout of the cells that are after the Cell with a disclosureNode ?

Ziad-Mid: > Would it be possible to avoid re-layout of the cells that are after the Cell with a disclosureNode ?

This is actually what is happening, when we scroll down , when we find a TreeCell that has a disclosure node this is when the condition `disclosureWidth > defaultDisclosureWidth` is true and we do a re-layout of the visible previous cells before the disclosure node and the new value is saved in the `maxDisclosureWidthMap` so the cells that are layouted after the disclosure node when scrolling are using the new value of `defaultDisclosureWidth`.

arapte: Yes, the cells after disclosure node use the new `defaultDisclosureWidth`. But their layout happens twice using the new `defaultDisclosureWidth`.

**First time**: The loop `for (IndexedCell cell : flow.cells) {` initiates the layout of all the cells, including the cells that are after the disclosure node.

**Second time**: and layout of those after cells happens again as part of ongoing layout of `TreeView` from this method `Parent.layout()`

andy-goryachev-oracle: as long as both layouts happens within the same pulse, we should be ok.
I did not see any flicker with the last change.

Ziad-Mid: Yes,
I did limit the re-layout in my fix to happen only for the cells before the disclosure node cell.

andy-goryachev-oracle (approved):

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeCellSkin.java
```diff
@@ -222,10 +222,11 @@ public final DoubleProperty indentProperty() {
...
 
+                    int indexWithDisclosureNode = getSkinnable().getIndex();
                     VirtualFlow<?> flow = getVirtualFlow();
                     if (flow != null) {
                         for (IndexedCell cell : flow.cells) {
-                            if (cell != null && !cell.isEmpty()) {
+                            if (cell != null && !cell.isEmpty() && cell.getIndex() < indexWithDisclosureNode ) {
                                 cell.requestLayout();
                                 cell.layout();
                             }
```
arapte, line 232: Just a minor change please.
The for loop could break when `cell.getIndex() >= indexWithDisclosureNode`, so that we avoid looping through the cells that are after disclosureNode.

```
for (IndexedCell cell : flow.cells) {
    if (cell != null) {
        if (cell.getIndex() >= indexWithDisclosureNode) {
            break;
        } else if (!cell.isEmpty()) {
            cell.requestLayout();
            cell.layout();
        }
    }
}
```

Ziad-Mid: Thanks for the suggestion,
I have pushed new changes.

arapte (approved):
LGTM

andy-goryachev-oracle (approved):

Ziad-Mid:
Thanks for reviews!

PR #1716, opened 2025-02-19
8335547: Support multi-line prompt text for TextArea

Ziad-Mid:
Added multi line prompt support for TextArea this will provide the ability to have multiple lines in textArea as expected,
Also fixed tests to meet the new changes

---------
### Progress
- [x] Change requires CSR request [JDK-8352564](https://bugs.openjdk.org/browse/JDK-8352564) to be approved

### Issues
 * [JDK-8335547](https://bugs.openjdk.org/browse/JDK-8335547): Support multi-line prompt text for TextArea (**Enhancement** - P4)
 * [JDK-8352564](https://bugs.openjdk.org/browse/JDK-8352564): Support multi-line prompt text for TextArea (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1716/head:pull/1716` \
`$ git checkout pull/1716`

`$ git checkout pull/1716` \
`$ git pull https://git.openjdk.org/jfx.git pull/1716/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1716`

`$ git pr show -t 1716`

</details>

<a href="https://git.openjdk.org/jfx/pull/1716.diff">https://git.openjdk.org/jfx/pull/1716.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1716#issuecomment-2726214139)
</details>


modules/javafx.controls/src/main/java/javafx/scene/control/TextInputControl.java
```diff
@@ -316,14 +316,17 @@ public String getName() {
...
     private StringProperty promptText = new SimpleStringProperty(this, "promptText", "") {
         @Override protected void invalidated() {
-            // Strip out newlines
             String txt = get();
-            if (txt != null && txt.contains("\n")) {
-                txt = txt.replace("\n", "");
-                if (isBound()) {
-                    unbind();
-                }
+            if (isBound()) {
```
andy-goryachev-oracle, line 320: I think we might want to bypass the whole thing if it is a TextArea.

modules/javafx.controls/src/main/java/javafx/scene/control/TextInputControl.java
```diff
@@ -316,14 +316,17 @@ public String getName() {
...
+                unbind();
+            }
+            if (TextInputControl.this instanceof TextArea) {
```
andy-goryachev-oracle, line 323: Is this the right approach?  One can always ask whether we should rely on the `instanceof TextArea` here, for example in the context of a custom TextInputControl.

Also think of:
- Can a custom control override this behavior?
- Do we want to give this capability to the application?
- Should we create a public API instead?  And if so, which kind?

andy-goryachev-oracle:
Also notice the github actions (GHA) are failing, meaning this PR breaks some tests 

```
TextAreaTest > checkPromptTextPropertyBind() FAILED
TextFieldTest > checkPromptTextPropertyBind() FAILED
```

modules/javafx.controls/src/main/java/javafx/scene/control/TextInputControl.java
Ziad-Mid: Yes I did change the code a bit and now the tests do not fail

modules/javafx.controls/src/main/java/javafx/scene/control/TextInputControl.java
Ziad-Mid: The purpose of this is to remove the stripping of new lines in TextArea only as that is normally how it should be,
but the default behavior will stay as it was before for other classes that extends TextInputControl.
And to give the capability to applications to modify this behavior is not necessary.

Ziad-Mid:
> Also notice the github actions (GHA) are failing, meaning this PR breaks some tests
> 
> ```
> TextAreaTest > checkPromptTextPropertyBind() FAILED
> TextFieldTest > checkPromptTextPropertyBind() FAILED
> ```

andy-goryachev-oracle:
`ServiceLifecycleTest > cancelCalledFromOnFailed() FAILED`
this failure happens intermittently, see https://bugs.openjdk.org/browse/JDK-8088343 and is unrelated to the issue.

also, you can try re-running failed tests (look for a button in the top right corner).

andy-goryachev-oracle:
What do you think of an alternative solution: instead of trying to fix the property value, we could leave it alone but change the way it's interpreted by the skin.

Look at `TextFieldSkin::createPromptNode` method.  We could bind the Text.text property to the `StringBinding` created by `Bindings::createStringBinding`, something along the lines

```
        promptNode.textProperty().bind(Bindings.createStringBinding(() -> {
            String s = getSkinnable().getPromptText();
            // TODO replace newlines here
            return s;
        }, getSkinnable().promptTextProperty());
```

This will remove all the logic from the control and place it where it needs to happen (TextFieldSkin).  The TextAreaSkin is unchanged and will display newlines when all the cruft is removed from `TextInputControl::promptText`.

(We could also change the said property to be lazily initialized since in most cases it's not used).

What do you think?

andy-goryachev-oracle (commented):
The idea is solid - you could make it ready for review (RFR) now.

I've left a number of general formatting suggestions, but the main concern is the decision to convert the tests to be specific to the `TextField` and the `TextArea` (omitting the `PasswordField`).  Since the only difference is the expected values, maybe we can keep the existing tests and introduce a function that computes the expected value based on the (parameter) type.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TextAreaTest.java
```diff
@@ -65,6 +71,15 @@ public void cleanup() {
         removeUncaughtExceptionHandler();
     }
 
+    private void initStage() {
+        //This step is not needed (Just to make sure StubToolkit is loaded into VM)
+        Toolkit tk = Toolkit.getToolkit();
+        root = new StackPane();
+        scene = new Scene(root);
+        stage = new Stage();
+        stage.setScene(scene);
```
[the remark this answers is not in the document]
andy-goryachev-oracle: note: this depends on whether we want to keep the tests in `TextInputControlTest`, and whether a `stage` is needed for the tests.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TextInputControlTest.java
```diff
@@ -107,39 +107,6 @@ public void textDefaultsToEmptyString(Class<?> type) {
         assertEquals("", textInput.getText());
     }
 
-    @ParameterizedTest
-    @MethodSource("parameters")
-    public void bindPromptTextWithoutLineBreaks(Class<?> type) {
```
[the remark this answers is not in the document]
Ziad-Mid: With the new approach the `promptText` property is accepting any value so it's expected to have prompt text with Linebreaks for `TextField` and `PasswordField`, is this the test that I have to keep ?

Else to have a test that uses `stage` in `TextInputControlTest` we will have to check for the Class type in many cases since `TextInputSkinShim::getPromptNode` have two overloaded methods for `TextArea` and `TextField`
It will look like this in the test :

```
        SkinBase<?> skin;
        if (type.equals(TextField.class)) {
            textInput = new TextField();
            skin = new TextFieldSkin((TextField) textInput);
        } else if (type.equals(PasswordField.class)) {
            textInput = new PasswordField();
            skin = new TextFieldSkin((TextField) textInput);
        } else if (type.equals(TextArea.class)) {
            textInput = new TextArea();
            skin = new TextAreaSkin((TextArea) textInput);
        }

//test logic

if (type.equals(TextField.class) || type.equals(PasswordField.class)) {
            promptNode = TextInputSkinShim.getPromptNode((TextField) textInput);
        } else {
            promptNode = TextInputSkinShim.getPromptNode((TextArea) textInput);
        }
```

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
```diff
@@ -734,6 +730,13 @@ private void createPromptNode() {
         promptNode.fontProperty().bind(getSkinnable().fontProperty());
 
         promptNode.textProperty().bind(getSkinnable().promptTextProperty());
+        promptNode.textProperty().bind(Bindings.createStringBinding(() -> {
+            String s = getSkinnable().getPromptText();
+            if (s == null)
+                s = "";
+            s = s.replace("\n", "");
+            return s;
+        }, getSkinnable().promptTextProperty()));
```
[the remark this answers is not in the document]
mstr2: You should consider using a fluent binding here, which is a more modern solution compared to the `Bindings` class. It is also simpler because you don't need to check for `null`:

```java
promptNode.textProperty().bind(getSkinnable().promptTextProperty().map(s -> s.replace("\n", "")));
```

mstr2:
What happens if the text contains line breaks other than `\n`, such as `\r` or `\r\n`? There's `System.lineSeparator()`, but we might want to remove all kinds of line separators regardless.

andy-goryachev-oracle:
> What happens if the text contains line breaks other than `\n`

This is a good question.
I guess it depends on what other characters cause line break in JavaFX.
Wikipedia (https://en.wikipedia.org/wiki/Newline) specifies a few:

```
1. vert tab \u000b ---
2. form feed \u000c ---
3. cr \u000d ---
4. crlf \u000d\u000a ---
5. lf \u000a ---
6. nel \u0085 ---
7. ls \u2028 ---
8. ps \u2029 ---
```

andy-goryachev-oracle:
@Ziad-Mid please make sure we _need_ to change - does the code (with no filtering) render **other** newline characters?

Ziad-Mid:
I did test the mentioned line separators without filtering and the new lines are not rendered as newlines, so there is no need to add filtering for other line separators as they are already handled by default.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
```diff
@@ -26,10 +26,7 @@
 package javafx.scene.control.skin;
 
 import java.util.List;
-import javafx.beans.binding.BooleanBinding;
-import javafx.beans.binding.DoubleBinding;
-import javafx.beans.binding.ObjectBinding;
-import javafx.beans.binding.StringBinding;
+import javafx.beans.binding.*;
```
andy-goryachev-oracle, line 29: please do not use wildcard imports

modules/javafx.controls/src/test/java/test/javafx/scene/control/TextAreaTest.java
```diff
@@ -25,13 +25,10 @@
 
 package test.javafx.scene.control;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertSame;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assertions.fail;
+import static org.junit.jupiter.api.Assertions.*;
```
andy-goryachev-oracle, line 28: please do not use wildcard imports

mstr2: In tests, and especially for assertions, wildcard imports are usually accepted. We use them all over the place. However, I would agree that symbols other than assertions should usually be fully qualified.

andy-goryachev-oracle: good point

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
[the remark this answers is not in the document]
mstr2: Have you considered the suggestion?

mstr2:
I think that we should treat all _usual_ line separators in the same way (that is, `\n`, `\r`, and `\r\n`), as I don’t know if we have a specification that only one kind of line separator visually breaks lines in JavaFX.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TextAreaTest.java
Ziad-Mid: I kept the wildcards for assertions and removed others
thanks for mentioning them

andy-goryachev-oracle:
Confirming that no other character (see earlier comment) other than LF render as newlines, on macOS and Windows.

The ticket mentions a possible "workaround" of using \r (CR) on Windows, but it is not true.

Screenshot 2025-03-21 112855

Ziad-Mid:
> @Ziad-Mid : please do not resolve the conversations without the original commenter confirming that the conversation is resolved (it's really a bug in guthub, as only the originator of the comment can say if the conversation has been resolved). I think there are still some questions left unanswered, please check.
> 
> (I've started to use thumbs-up emoji to mark the comment as addressed, if you want my opinion).
> 
> Other than that, the changes look good to me.

I did unresolved them, and I will check the suggestions whenever possible and answer,
thanks

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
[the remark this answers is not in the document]
Ziad-Mid: Yes I will push the new changes thanks for the suggestion @mstr2

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
```diff
@@ -734,13 +734,7 @@ private void createPromptNode() {
...
 
         promptNode.textProperty().bind(getSkinnable().promptTextProperty());
-        promptNode.textProperty().bind(Bindings.createStringBinding(() -> {
-            String s = getSkinnable().getPromptText();
-            if (s == null) {
-                return "";
-            }
-            return s.replace("\n", "");
-        }, getSkinnable().promptTextProperty()));
+        promptNode.textProperty().bind(getSkinnable().promptTextProperty().map(s -> s.replace("\n", "")));
```
andy-goryachev-oracle, line 737: code in line 736 is not needed (we all missed that)

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
```diff
@@ -30,6 +30,7 @@
 import javafx.beans.binding.DoubleBinding;
 import javafx.beans.binding.ObjectBinding;
 import javafx.beans.binding.StringBinding;
+import javafx.beans.binding.Bindings;
```
andy-goryachev-oracle, line 33: unused import can be removed

Ziad-Mid: removed thanks

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
Ziad-Mid: removed it also thanks

modules/javafx.controls/src/test/java/test/javafx/scene/control/TextInputControlTest.java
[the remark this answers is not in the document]
Ziad-Mid: I guess we will keep it like this as you said since tested is identical between TextField and PasswordField

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
```diff
@@ -733,7 +732,7 @@ private void createPromptNode() {
         promptNode.visibleProperty().bind(usePromptText);
         promptNode.fontProperty().bind(getSkinnable().fontProperty());
 
-        promptNode.textProperty().bind(getSkinnable().promptTextProperty());
+        promptNode.textProperty().bind(getSkinnable().promptTextProperty().map(s -> s.replace("\n", "")));
```
mstr2, line 735: I recommend to replace all usual newline characters with `replaceAll("[\r\n]", "")` as I can't think of a compelling reason to remove only some, but leave others in the string.

Ziad-Mid: The tests show that only LF "\n" is rendered as a new line, there is no need to add more restrictions that is not needed
and the same was tested by @andy-goryachev-oracle previously in the comments and it confirms the same.

mstr2: That doesn't sound like a compelling reason to me. In fact, it makes it seems like a bug in JavaFX that a line break is only rendered with `\n`, but not with `\r\n` or `\r`.

In any case, the goal here is to (semantically) transform a string such that it doesn't contain line breaks, and line breaks come in three different usual forms. Our goal should always be to do the right thing, and not stop half-way and rely on unspecified rendering quirks for the rest.

kevinrushforth: Option 1 is intentionally the status quo, and matches what Swing's JComponent does, although @mstr2 is right that this isn't documented. An RFE to treat `\r` or `\r\n` as a newline could be considered in the future. We wouldn't do that as part of this PR.

So for _this_ PR, the question is what characters should be elided for the prompt text of a `TextField` so that multiple lines as a single line? Limiting this to stripping `\n` is sufficient given the current implementation, unless and until something else changes. Also, it matches what the existing implementation tries to do when it modifies the actual property value.

Ziad-Mid: @kevinrushforth 
This is due to `TextFieldContent::insert` we have a filterInput and the second parameter to strip newlines is set to true in the textField class.
`filterInput(String txt, boolean stripNewlines, boolean stripTabs)`
For `TextAreaContent::insert` the parameter is set to false to allow newlines.

mstr2 (approved):

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
kevinrushforth: Thanks for the explanation. I have no further comments or questions, so feel free to integrate when ready.

PR #1717, opened 2025-02-19
8350048: Enforce threading restrictions for show and hide methods in Window, Control, and Skin

andy-goryachev-oracle:
- enforced fx application thread
- clarify doc where an `IllegalStateException` can get thrown, such as hide() and show()
- clarified `ComboBoxBase::show`
- added a headful test `TestThreadingRestrictions`

---------
### Progress
- [x] Change requires CSR request [JDK-8350962](https://bugs.openjdk.org/browse/JDK-8350962) to be approved

### Issues
 * [JDK-8350048](https://bugs.openjdk.org/browse/JDK-8350048): Enforce threading restrictions for show and hide methods in Window, Control, and Skin (**Bug** - P3)
 * [JDK-8350962](https://bugs.openjdk.org/browse/JDK-8350962): Enforce threading restrictions for show and hide methods in Window, Control, and Skin (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1717/head:pull/1717` \
`$ git checkout pull/1717`

`$ git checkout pull/1717` \
`$ git pull https://git.openjdk.org/jfx.git pull/1717/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1717`

`$ git pr show -t 1717`

</details>

<a href="https://git.openjdk.org/jfx/pull/1717.diff">https://git.openjdk.org/jfx/pull/1717.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1717#issuecomment-2672882947)
</details>


kevinrushforth (commented):
The API doc changes look good, although I noted a couple additional changes that are needed.

The API docs for `Dialog::hide` should also be documented to throw ISE (since it calls `close()` directly the implementation is fine)

The API docs for `Stage::close` should also be documented to throw ISE (since it calls `hide()` directly the implementation is fine)

For `ComboBoxBase`, were you going to get rid of the word "aspect" in the opening sentence "... display the popup aspect of the user interface."? I think that's a little awkward even with that word removed. Maybe it could just be: "... display the popup associated with this control."

As for the implementation, There are a few places where the skin will call hide(), and I can't tell by just looking at the code that they are all guaranteed to be on the FX app thread.

For example, I recommend checking the following:

* `MenuBarSkin` line 799
* `ComboBoxListViewSkin` line 199

There may be others.

modules/javafx.controls/src/main/java/javafx/scene/control/ChoiceBox.java
```diff
@@ -480,15 +479,25 @@ public String getName() {
...
+     * @throws IllegalStateException if this method is called on a thread
+     * other than the JavaFX Application Thread.
      */
     public void show() {
-        if (!isDisabled()) setShowing(true);
+        Toolkit.getToolkit().checkFxUserThread();
+        if (!isDisabled()) {
+            setShowing(true);
+        }
     }
```
arapte, line 491: Would it be a good idea to move the check `Toolkit.getToolkit().checkFxUserThread();` to a new method `show()` in Parent class `Control`? And may be similarly to Parent classes of other classes.

andy-goryachev-oracle:
> For example, I recommend checking the following:
> MenuBarSkin line 799
> ComboBoxListViewSkin line 199

I think it should be sufficient to mention the exception in the actual methods where the check is done, such as `hide()` and `show()`.

kevinrushforth:
> > For example, I recommend checking the following:
> > MenuBarSkin line 799
> > ComboBoxListViewSkin line 199
> 
> I think it should be sufficient to mention the exception in the actual methods where the check is done, such as `hide()` and `show()`.

I'm not talking about docs here. What I meant that if the code in question is ever executed on a background thread such that hide is called, this will now cause an exception at runtime. I don't know whether it is possible, but it isn't immediately obvious that this can't happen on a background thread -- at least not for the two cases I mentioned. In fact, I'm pretty sure that at least the second case (ComboBoxListViewSkin line 199) is possible.

```
        lh.addInvalidationListener(comboBox.sceneProperty(), (o) -> {
            if (((ObservableValue)o).getValue() == null) {
                comboBox.hide();
            }
        });
```

I suspect that the following sequence of operations, all on a background thread, will trigger the exception:

1. Create a ComboBox
2. Create a Skin
3. Add the combobox to a Scene (the scene must not be showing, of course)

A solution might be to check whether the combobox is showing before hiding it.

```
        lh.addInvalidationListener(comboBox.sceneProperty(), (o) -> {
            if (((ObservableValue)o).getValue() == null) {
                if (comboBox.isShowing()) {
                    comboBox.hide();
                }
            }
        });
```

And, if the code in MenuBarSkin is a problem, then something like this might be in order:

```
            if (menuButton.isShowing()) {
                menuButton.hide();
            }
```

Similarly, I think this listener in `MenuBarSkin::rebuildUI` (starting on line 913) might need a check before calling `menuButton.hide()`:

```
            menuButton.menuListener = (observable, oldValue, newValue) -> {
                if (menu.isShowing()) {
                    menuButton.show();
                    menuModeStart(container.getChildren().indexOf(menuButton));
                } else {
                    menuButton.hide();     // <--  check if (menuButton.isShowing()) before calling
                }
            };
```

kevinrushforth:
The updated docs look good. Go ahead and create the CSR.

andy-goryachev-oracle:
> MenuBarSkin line 799
> ComboBoxListViewSkin line 199

ComboBoxListViewSkin - thanks for the suggestion!

MenuBarSkin - is a bit more complex case because of the interaction with the system menu `Toolkit.getToolkit().getSystemMenu().isSupported()` which needs to be called from the FX application thread.

We probably should disallow creating MenuBars in background thread (similar to Window, WebView, and HtmlEditor) in the constructor.  What do you think?

andy-goryachev-oracle:
Created https://bugs.openjdk.org/browse/JDK-8350976 for the MenuBarSkin.

kevinrushforth (approved):
All looks good now.

Have you done a headful test run?

andy-goryachev-oracle:
> Have you done a headful test run?

combined branch (this PR + #1697 + #1727) is green on all platforms except for a single unrelated failure on linux: 

`RegionBackgroundFillUITest > testScenario1() FAILED`

modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
```diff
@@ -1183,6 +1183,8 @@ public void toBack() {
     /**
      * Closes this {@code Stage}.
      * This call is equivalent to {@code hide()}.
+     * @throws IllegalStateException if this method is called on a thread
+     * other than the JavaFX Application Thread.
```
mstr2, line 1187: Minor: add a blank line before `@throws`. I also think it's easier to read docs when multi-line text for a javadoc tag is indented, either by four spaces or lined up with the beginning of the first line of text (i.e. with the beginning of `IllegalStateException`).

mstr2 (approved):

modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
andy-goryachev-oracle: good idea, even though it does not affect generated javadoc.
will do next time, to avoid re-setting the approvals ;-)

mstr2: > will do next time, to avoid re-setting the approvals ;-)

I'll re-approve if you decide to do the changes anyway.

andy-goryachev-oracle: it is not required, and the code is inconsistent (Stage:472) ... but ok.

mstr2: Can you do it for the other changed docs also?

mstr2: You've added the same javadoc tag in 20 different places. All I'm asking is that you do the change that you did in this particular place in all of the other places too.

andy-goryachev-oracle: I misunderstood, sorry.  Done.

mstr2 (approved):

kevinrushforth (approved):

PR #1719, opened 2025-02-21
8327478: Add System test to verify TextSelection issue for webkit-617.1

Gopalora:
There was no test included with the fix for [JDK-8326989](https://bugs.openjdk.org/browse/JDK-8326989),

Hence we are adding a system test now.

Test is written as 
1. Load html content in web view.
2. pick the color of mouse pointer.
3. Perform double click.
4. pick the color again.
> expected bahaviour: colour picked in step 2 and 4 should not match.

Verification:
The test passes with latest webkit source
Also verified that test fails when the fix for [JDK-8326989](https://bugs.openjdk.org/browse/JDK-8326989) is reverted.

---------
### Progress

### Issue
 * [JDK-8327478](https://bugs.openjdk.org/browse/JDK-8327478): Add System test to verify TextSelection issue for webkit-617.1 (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1719/head:pull/1719` \
`$ git checkout pull/1719`

`$ git checkout pull/1719` \
`$ git pull https://git.openjdk.org/jfx.git pull/1719/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1719`

`$ git pr show -t 1719`

</details>

<a href="https://git.openjdk.org/jfx/pull/1719.diff">https://git.openjdk.org/jfx/pull/1719.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1719#issuecomment-2691547461)
</details>

### DIFF
--- a/tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
@@ -39,19 +39,17 @@ import org.junit.jupiter.api.Timeout;
 import test.robot.testharness.RobotTestBase;
 import test.util.Util;
 
-public class TextSelectionTest extends RobotTestBase{
+public class TextSelectionTest extends RobotTestBase {
 
-    private static final String html =
-    """
-    <html>
-    <head></head>
-    <body>&nbsp&nbsp&nbsp&nbsp some text</body>
-    </html>
-    """;
+    private static final String html = """
+        <html>
+        <body><font size="2">&nbsp&nbsp&nbsp&nbsp some text</font></body>
+        </html>
+        """;
 
     private static CountDownLatch webviewLoadLatch = new CountDownLatch(1);
-    private Color colorBefore;
-    private Color colorAfter;
+    private volatile Color colorBefore;
+    private volatile Color colorAfter;
 
     @BeforeEach
     public void beforeEach() {
--- a/tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
@@ -99,7 +99,8 @@ public class TextSelectionTest {
         Util.parkCursor(robot);
         Util.runAndWait(() -> colorBefore = robot.getPixelColor(x, y));
 
-        Util.doubleClick(robot, x, y);
+        Util.runAndWait(() -> robot.mouseMove(x, y));
+        Util.doubleClick(robot);
         Util.sleep(500); // Wait for the selection highlight to be drawn
 
         Util.parkCursor(robot);
--- a/tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
@@ -1,0 +1,123 @@
+/*
+ * Copyright (c) 2025, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package test.robot.javafx.web;
+
+import java.util.concurrent.CountDownLatch;
+
+import javafx.application.Application;
+import javafx.concurrent.Worker;
+import javafx.scene.input.MouseButton;
+import javafx.scene.paint.Color;
+import javafx.scene.robot.Robot;
+import javafx.scene.Scene;
+import javafx.scene.web.WebView;
+import javafx.stage.Stage;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
+import test.util.Util;
+
+public class TextSelectionTest {
+
+    private static final String html = "<html>" +
+        "<head></head> " +
+        "<body>&nbsp&nbsp&nbsp&nbsp some text</body>" +
+        "</html>";
+
+    private static CountDownLatch startupLatch = new CountDownLatch(1);
+    private static CountDownLatch webviewLoadLatch = new CountDownLatch(1);
+    private static Scene scene;
+    private static Robot robot;
+    private Color colorBefore;
+    private Color colorAfter;
+
+    public static class TestApp extends Application {
+        @Override
+        public void start(Stage primaryStage) {
+
+            robot = new Robot();
+            WebView webview = new WebView();
+            scene = new Scene(webview, 400, 300);
+            primaryStage.setScene(scene);
+            primaryStage.setAlwaysOnTop(true);
+
+            webview.getEngine().getLoadWorker().stateProperty().addListener((ov, o, n) -> {
+                if (n == Worker.State.SUCCEEDED) {
+                    webviewLoadLatch.countDown();
+                }
+            });
+            webview.getEngine().loadContent(html);
+            primaryStage.setOnShown(e -> startupLatch.countDown());
+            primaryStage.show();
+        }
+    }
+
+    @BeforeAll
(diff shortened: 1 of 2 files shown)


kevinrushforth:
@Gopalora Please enable GHA tests on your repo.

I'll provide a few comments now, and formally review it once your OCA status is verified.

Gopalora:
Thanks for the review.
I have done the code change as per the suggestions in the code file.
Please have a relook of the changes.

tests/system/src/test/java/test/util/Util.java
```diff
@@ -333,6 +334,20 @@ public static void waitForLatch(CountDownLatch latch, int seconds, String msg) {
         }
     }
 
+    /**
+     * Makes double click of the mouse left button.
+     */
+    public static void doubleClick(Robot robot, int x, int y) {
```
kevinrushforth, line 340: Since this is now a general-purpose utility, I'd prefer to separate out the mouse move from the double-click (with the mouse move being done in the test itself) and just have this be `void doubleClick(Robot robot)`.

Gopalora: Done.

kevinrushforth:
The latest changes look good. I'll formally review it once your OCA status is verified.

tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
```diff
@@ -0,0 +1,112 @@
...
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
+import test.util.Util;
+
+public class TextSelectionTest {
+
+    private static final String html = "<html>" +
```
andy-goryachev-oracle, line 47: we could use new java text block here

```
private static final String html =
    """
    <html>
    <head></head>
    <body>&nbsp&nbsp&nbsp&nbsp some text</body>
    </html>
    """;
```

tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
```diff
@@ -0,0 +1,112 @@
...
+import javafx.stage.Stage;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
+import test.util.Util;
+
+public class TextSelectionTest {
```
andy-goryachev-oracle, line 45: I wonder if we could extend RobotTestBase base class to simplify the code?

tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
```diff
@@ -0,0 +1,112 @@
...
+        Util.shutdown();
+    }
+
+    // ========================== TEST CASE ==========================
+    @Test
+    @Timeout(value=20)
+    public void testTextSelection() throws Exception {
+
+        int x = (int)(scene.getWindow().getX() + scene.getX() + 22);
+        int y = (int)(scene.getWindow().getY() + scene.getY() + 15);
```
andy-goryachev-oracle, line 97: can these offsets (22, 15) be measured instead of hard-coded?

Gopalora: We are not getting the text area from web view. This point is used to simulate the mouse click.

tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
Gopalora: Done.

tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
Gopalora: Done.

tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
+
+import test.robot.testharness.RobotTestBase;
+import test.util.Util;
+
+public class TextSelectionTest extends RobotTestBase{
```
kevinrushforth, line 42: Minor: please add a space before the `{`

kevinrushforth (commented):
Looks good with one minor code style issue.

tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private static final String html =
+    """
+    <html>
+    <head></head>
+    <body>&nbsp&nbsp&nbsp&nbsp some text</body>
+    </html>
+    """;
```
kevinrushforth, line 50: One more minor code style issue: lines 45-50 should be intended

tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
andy-goryachev-oracle: so they are arbitrary, and will work for every font size?

(The default font size varies depending on the platform and other circumstances, see `PrismFontFactory::getSystemFontSize()` - though I am not sure if the default font size is picked up by the `WebView`.

tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private static CountDownLatch webviewLoadLatch = new CountDownLatch(1);
+    private Color colorBefore;
```
andy-goryachev-oracle, line 53: suggestion: make `colorBefore/After` _volatile_ since they are being accessed from different threads without any synchronization.
I think this should be sufficient (no need for `AtomicReference`).

tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
```diff
@@ -0,0 +1,92 @@
...
+        Util.runAndWait(() -> colorBefore = robot.getPixelColor(x, y));
+
+        Util.runAndWait(() -> robot.mouseMove(x, y));
+        Util.doubleClick(robot);
+        Util.sleep(500); // Wait for the selection highlight to be drawn
+
+        Util.parkCursor(robot);
+        Util.runAndWait(() -> colorAfter = robot.getPixelColor(x, y));
+
+        Assertions.assertNotEquals(colorBefore, colorAfter,
```
andy-goryachev-oracle, line 89: should we also test for non-null `colorBefore` ?

tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
andy-goryachev-oracle: https://en.wikipedia.org/wiki/Cache_coherence

kevinrushforth: > https://en.wikipedia.org/wiki/Cache_coherence

That's interesting, but not particularly relevant. What is relevant is the JVM spec, specifically the Java Memory Model, which guarantees a happens-before ordering of actions in two threads that synchronize-with each other.

> edit: you might be right about runAndWait, we know it does separate the access from multiple threads in time, and it might be even smart enough to figure out that the field is being accessed from two different threads and emit the appropriate bytecode (similarly how we don't need the `final` keyword for effectively final variables nowadays).

It isn't about the byte code that is emitted as much as it is about how the code is executed by the two different threads. And it isn't because it need to be "smart enough", it's because that's how the memory model works.

A write by the thread that executes the runnable passed into runAndWait (the FX app thread in this case), happens-before the read done by the test thread that is done after the runAndWait returns because the runAndWait call causes the actions done in the runAndWait's runnable to synchronize-with the wait done by the calling thread.

> But I think that adding `volatile` **_guarantees_** that all the necessary steps are done by the JVM.

Yes, adding `volatile` guarantees the appropriate happens-before relationship. But so does synchronizing.

I don't really care one way or the other in this case.

tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
kevinrushforth: `Robot::getPixelColor` can't return null (although I checked the docs and we don't specify one way or the other). It seems fine either way.

andy-goryachev-oracle: what if the lambdas never run?

kevinrushforth: Then we have far bigger problems than this one test. The reason we use `runAndWait` instead of `runLater` is so we guarantee that the lambda has run before that method returns. It's guaranteed to do so.

Gopalora: Added volatile to the variables colorBefore and colorAfter.

tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
Gopalora: Done.

tests/system/src/test/java/test/robot/javafx/web/TextSelectionTest.java
Gopalora: Done as per one suggestion on "https://docs.oracle.com/en/java/javase/18/text-blocks/index.html"

Gopalora: Removed the line "<head></head>" as this is empty.

andy-goryachev-oracle (approved):
lgtm

PR #1720, opened 2025-02-21
8332947: [macos] java.awt.desktop.OpenURIHandler is not receiving events

pabulaner:
When trying to register an open URI handler when using JavaFX with a native menu, this task fails on Mac.
Either the native menu is not shown or the URIs are not received.

This pull request fixes this issue if the registration takes place inside the start method of an application before calling the show method.

The test for this pull request is non trivial, as the application needs to be installed on the Mac before it can be tested. Therefore the test is provided in a separate repository and it needs to be discussed if the test is necessary to have inside the JFX repo and if so, how it should be integrated.

Co-Author: @floriankirmaier

Link to the test repo: https://github.com/pabulaner/openurifx

---------
### Progress

### Error

### Issue
 * [JDK-8332947](https://bugs.openjdk.org/browse/JDK-8332947): [macos] java.awt.desktop.OpenURIHandler is not receiving events (**Bug** - P4)(⚠️ The fixVersion in this issue is [25] but the fixVersion in .jcheck/conf is jfx25, a new backport will be created when this pr is integrated.)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1720/head:pull/1720` \
`$ git checkout pull/1720`

`$ git checkout pull/1720` \
`$ git pull https://git.openjdk.org/jfx.git pull/1720/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1720`

`$ git pr show -t 1720`

</details>

<a href="https://git.openjdk.org/jfx/pull/1720.diff">https://git.openjdk.org/jfx/pull/1720.diff</a>

</details>

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
@@ -405,6 +405,22 @@ jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved)
     GLASS_CHECK_EXCEPTION(env);
 }
 
+- (void)application:(NSApplication *)application openURLs:(NSArray<NSURL *> *)urls
+{
+    LOG("GlassApplication:application:openURLs");
+
+    GET_MAIN_JENV;
+    jclass appEventHandlerClass = [GlassHelper ClassForName:"com.apple.eawt._AppEventHandler" withEnv:env];
+    jmethodID handleOpenURIMethod = (*env)->GetStaticMethodID(env, appEventHandlerClass, "handleOpenURI", "(Ljava/lang/String;)V");
+
+    for (NSURL *url in urls) {
+        jstring jURL = NSStringToJavaString(jEnv, url.absoluteString);
+
+        (*env)->CallStaticVoidMethod(env, appEventHandlerClass, handleOpenURIMethod, jURL);
+        (*env)->DeleteLocalRef(env, jURL);
+    }
+}
+
 - (void)application:(NSApplication *)theApplication openFiles:(NSArray *)filenames
 {
     LOG("GlassApplication:application:openFiles");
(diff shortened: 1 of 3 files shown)


kevinrushforth:
@pabulaner Once your OCA status is verified, we can formally take a look at this.

I see that the referenced bug is targeting AWT and yet your fix for this is in JavaFX. We don't currently reference anything from `com.apple.eawt` in JavaFX and I would prefer not to add a dependency now. I recommend looking for another solution, possibly in AWT if feasible.

If it does turn out to need a fix in JavaFX, then the JBS bug will need to be retargeted.

kevinrushforth:
> We don't currently reference anything from `com.apple.eawt` ...

Further, this is a non-public class in a non-exported package of `java.desktop`. So this proposed fix is using JNI to break encapsulation. You will need a different solution.

modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
```diff
@@ -405,6 +405,22 @@ - (void)applicationDidUnhide:(NSNotification *)aNotification
     GLASS_CHECK_EXCEPTION(env);
 }
 
+- (void)application:(NSApplication *)application openURLs:(NSArray<NSURL *> *)urls
+{
+    LOG("GlassApplication:application:openURLs");
+
+    GET_MAIN_JENV;
+    jclass appEventHandlerClass = [GlassHelper ClassForName:"com.apple.eawt._AppEventHandler" withEnv:env];
```
kevinrushforth, line 413: Using an internal class of `java.desktop` is not an acceptable solution.

kevinrushforth (changes requested):

kevinrushforth:
I see that this was closed in favor of #1755

pabulaner:
Yes, I am fixing the issue now with an event based approach, so I don't add a direct dependency.

PR #1723, opened 2025-02-22
8350149: VBox ignores bias of child controls when fillWidth is set to false

hjohn:
Fixes the case where `VBox` will ignore the (horizontal) bias of a child when its fill width property is set to `false`.  This manifests itself with labels that have their wrap text property set to `true`, and the container is not wide enough to hold the text on a single line (in other words, the label is potentially far wider, and fill width should have no effect).  No reflow would occur before this fix.

The solution is to ensure the `computeChildAreaMin/Pref/MaxHeight` functions are provided with a `fillWidth` parameter, to be used in the case a horizontally biased control is encountered (several of the parameters to these compute functions are only used for those special cases and ignored otherwise).

With this additional information, the compute functions can decide between the preferred width of a control or the available width of the container.  In the old implementation this decision was made *before* these functions were called, and the available width was reset to -1 to indicate the case that the preferred width should be used.  However, there are three cases, and so setting a single parameter to -1 can't effectively communicate that:

Assuming a control that is horizontally biased, and is resizable there are three cases when calculating a **height**:

1. There is no available width: bias is ignored (as there is no value to use as a dependent value) and the height is then simply calculated ignoring available width (which is -1) and fill width settings (same as unbiased controls).
2. There is an available width and fill width is false; the bias logic must first find a reasonable width before it can call any height function; with fill width false, this reasonable width will be the preferred width of the control, unless it exceeds its maximum width or the available width, in which case the smallest of these three values is used.  The final width calculated is then used to determine the height.
3. There is an available width and fill width is true; this is the same as case 2, except the available width is used as a dependent value (unless its greater than the child's maximum width, in which case the smaller is used).  The final width calculated is then used to determine the height.

All in all, this PR removes several inconsistencies between horizontal and vertical computations. The end result is that the horizontal and vertical bias calculations now more closely mirror each other.

**Note**: some tests have had their values adjusted.  This is because the asymmetries between the compute width/height functions have been removed.  Most of these tests use the `MockBiased` test control that simulates an area of pixels that always covers the same size (ie. it is 10x1000 or 100x100 or 1000x10 = 10000).  One should be aware though that when correctly querying its minimum size, it will say something like 100x100 -- and that's correct, as that's what `MockBiased` will tell the layout system, even though in a single dimension it can return a minimum size of 1 when the other dimension provided is set to say 10000. 

Due to a bug in how content bias was handled before in the height functions (content bias was taken into account when `-1` was passed which is incorrect) some tests saw much smaller minimum sizes.  `HBoxTest` is an example of this, but if you compare it to the equivalent code in `VBoxTest`, you can see that something is clearly off, as the tests don't agree even though they mirror each other.  The original developer tried to rationalize these differences, instead of investigating where they came from.

It may be worth to add tests for a different type of biased control.  `MockBiased` is an example of a control that requires roughly the same area when resized (like a reflowing Label, or a FlowPane).  Another example is a control that tries to maintain a fixed aspect ratio, like a scaling Image.  The latter type is not interested in maintaining an area of roughly the same size, but instead it is interested in maintaining a fixed factor between its width and height.

**Note 2**: Many layout containers will call min/pref/max width/height methods with a dependent size provided (instead of -1) even when the control has no content bias.  Control implementations therefore need to be careful not to use this value when they have no bias or have the opposite bias.  `MockBiased` handles this correctly.

**BorderPaneTest**

BorderPaneTest has had some tests adjusted as its height computations were affected by the bug in the compute height functions where a biased calculation was partially executed when it shouldn't have. The height values make much more sense now instead of being some weird fractional number when a biased control is involved.  I've adjusted the tests accordingly, and tried adding some rationalization of how the values are reached.  Note that `BorderPane` (apparently) always adjusts whatever size it is given during layout to be greater than its minimum size; this is the first time I've seen containers do that, instead of going with whatever they've been given...

---------
### Progress

### Issue
 * [JDK-8350149](https://bugs.openjdk.org/browse/JDK-8350149): VBox ignores bias of child controls when fillWidth is set to false (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1723/head:pull/1723` \
`$ git checkout pull/1723`

`$ git checkout pull/1723` \
`$ git pull https://git.openjdk.org/jfx.git pull/1723/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1723`

`$ git pr show -t 1723`

</details>

<a href="https://git.openjdk.org/jfx/pull/1723.diff">https://git.openjdk.org/jfx/pull/1723.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1723#issuecomment-2677142018)
</details>


modules/javafx.graphics/src/main/java/javafx/scene/layout/GridPane.java
```diff
@@ -1447,8 +1447,13 @@ private CompositeSize computePrefHeights(double[] widths) {
             Node child = managed.get(i);
             int start = getNodeRowIndex(child);
             int end = getNodeRowEndConvertRemaining(child);
-            double childPrefAreaHeight = computeChildPrefAreaHeight(child, isNodePositionedByBaseline(child) ? rowPrefBaselineComplement[start] : -1, getMargin(child),
-                    widths == null ? -1 : getTotalWidthOfNodeColumns(child, widths));
+            double childPrefAreaHeight = computeChildPrefAreaHeight(
```
andy-goryachev-oracle, line 1450: thank you for fixing the indentation!  much more readable now.

andy-goryachev-oracle:
you probably want to start off a new PR from the latest master...

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
```diff
@@ -2036,18 +2007,17 @@ public final double snappedRightInset() {
...
+
+            alt = computeBoundedWidth(child, fillWidth, contentWidth);
             max = child.maxHeight(alt);
         }
         // For explanation, see computeChildPrefAreaHeight
         if (maxBaselineComplement != -1) {
             double baseline = child.getBaselineOffset();
             if (child.isResizable() && baseline == BASELINE_OFFSET_SAME_AS_HEIGHT) {
-                return top + snapSizeY(boundedSize(child.minHeight(alt), child.maxHeight(alt), Double.MAX_VALUE)) + bottom
+                return top + snapSizeY(boundedSize(child.minHeight(alt), max, Double.MAX_VALUE)) + bottom
```
andy-goryachev-oracle, line 2020: here and elsewhere: a wise man once said that the sum of snapped values may not come out snapped.  Should we start snapping the result of any algebraic operation that involves snapped values?

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
```diff
@@ -2058,62 +2028,133 @@ public final double snappedRightInset() {
...
+    /*
+     * Removes the given Margin (if any) from a width which still includes margins
+     * to create a content width.
+     */
+    private double computeContentWidth(Insets margin, double width) {
+        boolean snap = isSnapToPixel();
+        double left = margin != null ? snapSpaceX(margin.getLeft(), snap) : 0;
+        double right = margin != null ? snapSpaceX(margin.getRight(), snap) : 0;
+
+        return width - left - right;
```
andy-goryachev-oracle, line 2093: snap?

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
```diff
@@ -2058,62 +2028,133 @@ public final double snappedRightInset() {
...
+     *
+     * Sizes provided by the container that may be used as a dependent value when
+     * calculating sizes for biased controls. These may be set to -1 to indicate
+     * no such information is available. If given, the sizes include the Margin
+     * of the child. As such the Margin must be removed before passing these
+     * values as a dependent value to min/pref/max width/height functions.
+     *
+     * # content width/heights:
+     *
+     * The space allocated to a child, minus its margins. These are never -1.
```
andy-goryachev-oracle, line 2044: or always `> 0.0` ?

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
```diff
@@ -2058,62 +2028,133 @@ public final double snappedRightInset() {
...
+     *
+     * # bounded width/heights:
+     *
+     * The space allocated to a child, minus its margins, adjusted according to
+     * its constraints (min <= X <= max). These are never -1.
```
andy-goryachev-oracle, line 2049: always `> 0.0` here?

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/RegionTest.java
```diff
@@ -1009,7 +1009,631 @@ public void testComputeChildPrefAreaHeightHonorsMinWidthOverMax() {
...
+        Region c2 = new MockBiased(Orientation.HORIZONTAL, 100, 200);
+        Region c3 = new MockRegion(10, 10, 100, 100, 1000, 1000);
+
+        pane.getChildren().addAll(c1, c2, c3);
+
+        /*
+         * Ensure that no fillHeight/height combinations have effect on controls that are not vertically biased:
+         */
+
+        assertEquals(12, RegionShim.computeChildMinAreaWidth(pane, c2, -1, new Insets(1), -1, false), 1e-100);
```
andy-goryachev-oracle, line 1025: might as well declare a static `computeChildMinAreaWidth()` which delegates to  `RegionShim.computeChildMinAreaWidth`...

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
hjohn: LOL, I think I may have said that :)  However, perhaps the problem is not as bad as I made it out to be.  It's definitely true that any operation on a floating point value may slightly nudge it away from a true snapped value (somewhere in the 10th decimal place) and that this problem gets worse the more operations you perform (ie. a HBox with 10.000 children will have a worse error than one with only a few children).

But there may be some middle ground possible here.  As long as our layout containers are aware that values returned from `computeMinWidth` etc are only "near" snapped (meaning that if you round them to the nearest pixel, you'll always get the right result), it should be fine -- one must take care not to immediately call `ceil` or `floor` on these values.  A thing we may want to look into is how this is rendered on screen by the rendering layer; does the rendering do additional work when values are near snapped, or do they perform some rounding already anyway.  I certainly never noticed these small errors resulting in display artifacts.

In any case, this is probably better addressed in a separate effort, and we should probably write some guidelines first.  I'm hoping to do some of this with the space distributor PR.

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
hjohn: Yeah, or possibly zero included.  What I meant here is that these are **real** values, and don't have a "special" value `-1` meaning absent/unavailable.

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/RegionTest.java
hjohn: Yeah, I can streamline this a bit more.

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
andy-goryachev-oracle: In theory, all these calculations end up being used by the layoutChildren() which always (?) snap the values.

So the question is whether this small error gets accumulated enough to shift the result to a wrong value.  Given a typical number of children (<1000) and screen sizes (<10000), we might conclude that it's unlikely, since

`Math.ulp(1000 * 10_000.0) = 1.862645149230957E-9`

which is much smaller than the distance between pixels even at 1000% scale.

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
andy-goryachev-oracle: A _very quick_ test with the monkey tester showed it never entered `boundedSize()` with zero values, but I can't be sure (I've seen 0's in `layoutChildren()`).

The statement about "never -1" is even more suspect below when margins are subtracted...

I just wanted to point this out because I am not entirely sure that we'll never receive 0 width/height.

andy-goryachev-oracle:
FYI: To help with testing, I've added some containers to the monkey tester:

(AnchorPane, BorderPane, FlowPane) and added more options to the HBox/VBox page and add child function and the childrens' context menus.

kevinrushforth:
I'll review this too.

/reviewers 2

modules/javafx.graphics/src/shims/java/javafx/scene/layout/RegionShim.java
```diff
@@ -53,12 +53,16 @@ public static double computeChildMaxAreaWidth(Region r,
                 baselineComplement, margin, height, fillHeight);
     }
 
+    public static double computeChildMinAreaHeight(Region r,
+        Node child, Insets margin) {
+    return r.computeChildMinAreaHeight(child, margin);
+}
```
andy-goryachev-oracle, line 59: there is something wrong with indentation here.
and ... in this class

since it's a shim, I would rather auto-format the whole thing...

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/BorderPaneTest.java
```diff
@@ -346,9 +346,9 @@ public void testWithBiasedChildren() {
         assertEquals(40/*l*/ + 60/*r*/ + 200/*c*/, borderpane.prefWidth(-1), 1e-100);
         assertEquals(240 /* l + r + c*/, borderpane.prefHeight(-1), 1e-10);
         assertEquals(110, borderpane.minWidth(-1), 1e-100); /* min center + 2x pref width (l, r) */
-        assertEquals(50, borderpane.minHeight(-1), 1e-10);
+        assertEquals(20 /*t*/ + 200 /*c*/ + 20 /*b*/, borderpane.minHeight(-1), 1e-10);
```
andy-goryachev-oracle, line 349: minor: Do you think it'll be easier to define the constants explicitly, line
```
double L = 40;
double C = 200;
```
and use those?

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
```diff
@@ -1933,48 +1923,37 @@ public final double snappedRightInset() {
...
         double bottom = margin != null? snapSpaceY(margin.getBottom(), snap) : 0;
 
         double alt = -1;
-        if (child.isResizable() && child.getContentBias() == Orientation.HORIZONTAL) { // height depends on width
-            double left = margin != null ? snapSpaceX(margin.getLeft(), snap) : 0;
-            double right = margin != null ? snapSpaceX(margin.getRight(), snap) : 0;
-            alt = snapSizeX(boundedSize(
-                    child.minWidth(-1), width != -1 ? width - left - right
-                    : child.prefWidth(-1), child.maxWidth(-1)));
+        if (availableWidth != -1 & child.isResizable() && child.getContentBias() == Orientation.HORIZONTAL) { // height depends on width
```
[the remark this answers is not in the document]
hjohn: Oops, definitely `&&` :)

modules/javafx.graphics/src/main/java/javafx/scene/layout/VBox.java
```diff
@@ -482,24 +474,25 @@ private double growOrShrinkAreaHeights(List<Node>managed, double areaHeights[][]
 
         double[] usedHeights = areaHeights[0];
         double[] temp = areaHeights[1];
+        final boolean isFillWidth = isFillWidth();
```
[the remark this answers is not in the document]
hjohn: Yeah, I think I "copied" the code style surrounding it... you normally wouldn't catch me writing `final` on locals

modules/javafx.graphics/src/shims/java/javafx/scene/layout/RegionShim.java
hjohn: I'll see what I can do

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/BorderPaneTest.java
hjohn: I just copied the style that was being used, but can make any changes desired of course (at the expense of increasing the diff and amount of code to review).

andy-goryachev-oracle: I know it's a general policy not to do unrelated changes or reformatting, but I think in this case 
a) it's a test and
b) you already touched it
so might as well.

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
hjohn: There are a few guards in the code that prevent these values from becoming negative (as some negative values have special meanings).  See for example this code: 

    @Override public final double minWidth(double height) {
        final double override = getMinWidth();
        if (override == USE_COMPUTED_SIZE) {
            return super.minWidth(height);
        } else if (override == USE_PREF_SIZE) {
            return prefWidth(height);
        }
        return Double.isNaN(override) || override < 0 ? 0 : override;
    }

This method is `final`, to prevent anyone from messing with this logic.  And as you can see, the `override` value (which comes from calling `computeMinWidth` for example) is checked against special values, and other values are returned instead.  If it is not any special value, it is guarded against negative (it becomes 0) and against `NaN`.

Whether that truly guarantees we'll never see negative values, I'm not 100% sure off, but I think that's definitely the intent as specific negative values have special meanings.  So I think I'll document these as `cannot be negative`.

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
hjohn: That's absolutely true, however this can change quickly when large values are added or subtracted to/from small values, or when division or multiplication gets involved.  So I'd say its relatively safe to do simple calculations with near snapped values, but one must be careful with functions like `ceil` and `floor` as they can amplify tiny floating-point errors due to their discontinuous nature.  For example, if take a snapped `spacing` and add a snapped `left` and `right` margin, then call ceil on the result, it could go like this:

             1.99999999999999 +
             4.00000000000001 +
             4.00000000000001 =
            10.0000000000001 -> ceil -> 11 (instead of the expected 10)

andy-goryachev-oracle: good point!

This is exactly the reason for the code in ScaledMath:71
```
return Math.ceil(d - Math.ulp(d)) / scale;
```

hjohn: Come to think of it, most the issues here are caused by using functions like `floor` and `ceil`.  It might be an idea to change these functions to bias them slightly towards rounding to the closest value, instead of always doing a straight up `floor` or `ceil`.

For example, let's say I calculate a size as `3.0000001`; ceiling this (with snapSize) to `4` is quite ridiculous; obviously `3` was intended, but the ceiling function won't care.  But what if we subtracted a value (assuming we're dealing with pixels of course, which the `snap` functions are)? We could bias it slightly towards the correct value by using something ridiculously small like 1/10000th of a pixel. For example:

`3.0000001` - 1/10000th of a pixel = `2.9999`. Ceiling this value yields the intended `3`.  We could work much safer with near snapped values, as there is much less risk of a one ulp difference being dramatically amplified by floor/ceil functions.

hjohn: > good point!
> 
> This is exactly the reason for the code in ScaledMath:71
> 
> ```
> return Math.ceil(d - Math.ulp(d)) / scale;
> ```

Yeah, but I think we may want to subtract more than just 1 ulp.  A one ulp difference can be created after any operation (like add/subtract).  Do two of these without resnapping, and the difference will be >1 ulp)

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/BorderPaneTest.java
hjohn: I can do this, but realize that the test is large, and there's comments like this in multiple places, also places I didn't touch.  So, I can fix it here locally, resulting in this method standing out from the rest, or I can do this everywhere causing this PR to have a lot of non-changes that reviewers will need to be aware of and ignore.

Normally I would do this by marking a commit as a "refactor" commit, in which functionality changes are prohibited, and so reviewers only need to look for changes that jump out as not just a minor refactor / rename / move.  However, in the JDK repositories this isn't so easily done, and such commits are not retained (it is squashed together with functional commits), and so to keep them separated you must make a new ticket, new PR, and get new buy-in to get it included.

I'm really trying to avoid to make a commit too large to review and drown out the functional changes with cosmetic changes, even though every time I touch code like `CssStyleHelper` it itches to remove redundant checks or give fields/variables more descriptive names.  Even the changes I did in Region I was already questioning if this wouldn't be too much (there is more that I would have liked to change there, but I refrained as I fear this will just make it too hard to review).  I face a similar dilemma with HBox/VBox modifications... do I do the same modifications twice (for the sake of reviews) or do I introduce an `AbstractBox`, remove all duplicate code, and then apply the modifications there, ensuring that it will be almost impossible to review and seeing what I actually changed...

What we really need IMHO is a way to include refactors with functional changes.  In my day job, we do this by having a specific commit order; we have one or a couple of commits that do minor cleanups or refactors -- no functionality is changed in these commits, leading to quick and easy reviews of those changes.  Then the final commit (we force push to keep commits clean and focused during reviews -- the review tooling we use handles this without any problem) contains **only** the functional changes which, thanks to the earlier refactors, are often tiny changes that just change some parameters or add a function or two -- no noise, and reviewers can focus purely on functionality.

andy-goryachev-oracle: I think the main concern with the reformatting/tangentially related changes coming from the maintainers is the general pain associated with conflicts created when merging and backporting.

In this case, I think we might be ok, specifically in the tests.  I would rather see the cleaner and more maintainable code, even at the small additional expense.  Up to you.

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
andy-goryachev-oracle: I think you are onto something here.  It almost feels like we shouldn't be doing ceil/floor at all, rounding to the closest pixel instead.

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/BorderPaneTest.java
kevinrushforth: > I think the main concern with the reformatting/tangentially related changes coming from the maintainers is the general pain associated with conflicts created when merging and backporting.

It's not just this, it's also about avoiding changes in parts of the code you aren't touching. When you do this, reviewers have to look at a hole bunch of changes that aren't related to the fix.

> In this case, I think we might be ok, specifically in the tests. I would rather see the cleaner and more maintainable code, even at the small additional expense. Up to you.

If a fix touches only one method of a file with many methods, I very much do not want to see the entire file reformatted as part of that fix. So I recommend to either reformat the one method or leave it alone.

andy-goryachev-oracle (approved):
Looks good, and I don't see any regressions.

hjohn:
@kevinrushforth Do you have time to take a look?  It would be good to get this in, as there are more changes I'd like to do in this area (fixing the HBox/VBox rounding errors on high DPI displays).

kevinrushforth:
Yes, sorry for the delay. This one is next on my list.

kevinrushforth (approved):
The changes look correct to me as does my testing. I left a minor question or two, but nothing that needs to be addressed before integration.

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
hjohn: @kevinrushforth Yes, I wasn't intending to do anything about this in this change as it is unrelated.  I understand why `ceil` is used; FX distinguishes between sizes and spaces, where sizes are part of some visual part of a control, and spaces are non-control areas.  Spaces are generally ceil'd, so that even the thinnest control or border would show up as 1 pixel, while positions and spaces are just rounded.  Still, a border that would end up being 0.0001 pixels wide probably is still best hidden completely.

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
hjohn: Yes, sorry, this was addressed in the larger discussion.  I've left this as-is to keep the PR focus'd on one thing.

The calculation here is using 3 snapped values, and one can reasonably assume the result is "nearly" snapped.  If this value is used later with a ceiling function though, then it might ceil to the next higher value if the result is slightly too high due to floating point errors.  This is why it might be a good idea to adjust how our ceiling functions work in all cases; instead of using a tiny epsilon (or ulp), use a much larger value but still tiny in terms of pixels (like 1/10000th of a pixel).  Any "near" snapped values won't accidentally get rounded up to the next higher pixel then when ceil is used.

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
hjohn: 0 is probably still allowed, but it shouldn't be negative.

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
andy-goryachev-oracle: > Any "near" snapped values won't accidentally get rounded up to the next higher pixel

I like this idea!

kevinrushforth: I agree. This sounds promising.

kevinrushforth:
I think all questions have been resolved; if so, this is ready to integrate.

Let's file a follow-on bug to address possible changes to snapping.

Glavo:
Unfortunately, this commit caused our app to behave abnormally :(

I know this is a problem with our app, I will try to fix it. But what I want to say is that this PR does break some real-world applications, and maybe a workaround is really needed to make them work.

hjohn:
@Glavo I'd be interested in what problem you're running into (maybe before/after description?).  Perhaps there is a bug in the fix.  It's also possible you inadvertently have been relying on the behavior where content bias is sometimes ignored, which was a bug.  All FX containers always respect content bias.  In that case the only solution is to modify your layout a bit.  Let me know if I can be of help.

Glavo:
> @Glavo I'd be interested in what problem you're running into (maybe before/after description?). Perhaps there is a bug in the fix. It's also possible you inadvertently have been relying on the behavior where content bias is sometimes ignored, which was a bug. All FX containers always respect content bias. In that case the only solution is to modify your layout a bit. Let me know if I can be of help.

We encountered this problem. The content of this page (excluding the title bar) is a `BorderPane`, and the red box is the content of its center, a `FlowPane`.

image

In OpenJFX 25+9 and earlier, the height of this `FlowPane` was 280 pixels, but since 25+10, its height has grown to 576 pixels, so that the interface layout is broken.

image

It is not difficult for us to solve this problem. Wrapping this `FlowPane` in a `ScrollPane` will restore its height to normal. However, I still don’t understand the specific reason behind it.

hjohn:
The BorderPane shares some of the calculations with VBox, so it's possible it is a bug.  This small program can reproduce the issue:

```css
.main .button {
  -fx-pref-width: 12em;
  -fx-pref-height: 12em;
}
```

```java
package app;
import javafx.scene.Scene;
import javafx.scene.layout.FlowPane;
import javafx.stage.Stage;

public class LayoutBugDemo extends Application {

    @Override
        BorderPane borderPane = new BorderPane();
        FlowPane flowPane = new FlowPane();

        flowPane.getStyleClass().add("main");

        borderPane.setTop(new Label("Instance Name"));
        borderPane.setCenter(flowPane);

        flowPane.getChildren().addAll(
            new Button("Minecraft"),
            new Button("Forge"),
            new Button("NeoForge"),
            new Button("OptiFine"),
            new Button("Fabric"),
            new Button("Fabric API"),
            new Button("Quilt"),
            new Button("SQL/QFAPI")
        );

        borderPane.setBottom(new Button("Install"));

        Scene scene = new Scene(borderPane, 600, 400);

        scene.getStylesheets().add("styles.css");

        stage.setScene(scene);
        stage.setTitle("JavaFX Layout Bug Demo");
        stage.show();
    }

        launch(args);
    }

}

```

hjohn:
I looked a bit further, and I think that the `getAreaHeight` function in `BorderPane` which forwards to `Region`'s `computeChildPrefAreaHeight` (which was given an extra parameter) is the culprit.  It's called with `fillWidth` set to `false` but calling it with `fillWidth` set to `true` seems to restore the old behavior.

The only troubling part of that fix however is that its counterpart (`getAreaWidth`) calls the corresponding region function for horizontal calculations with `fillHeight` set to `false` (which is why I used `false` for `fillWidth` in the height calculation as well).

I'll need to look more closely what is the correct fix here, as it seems to me that having vertical calculations use fill, but not horizontal calculations is also not what one would expect.   Creating an example where the FlowPane is laid out vertically and seeing how it behaves then may give a clue if that breaks in the same way (allocating too much width to it when used as the center of a border pane).

hjohn:
It looks like when using a vertical flow pane, that the left-to-right version of the layout is also quite poor:

image

When I apply the same fix in `BorderPane` for the horizontal calculations, it becomes this:

image

Note how the flow pane now doesn't take an unreasonable amount of space anymore cutting the install button from view.

andy-goryachev-oracle:
It might be better to create a new JBS ticket if this is a new bug (or reference an existing one).

Glavo:
Do you plan to fix this issue before the release of OpenJFX25? I'm wondering if I need to use a workaround to get around it.

kevinrushforth:
There first needs to be a new bug filed. Once there is, and a fix is available, we can see how risky the fix will be and decide  whether to backport it to jfx25.

hjohn:
@kevinrushforth @Glavo bug is filed here: https://bugs.openjdk.org/browse/JDK-8362873

PR #1724, opened 2025-02-25
8281384: Random chars on paste from Windows clipboard

Schmidor:
Windows programs may reuse a clipboard buffer that is larger than the new content. In this case de NUL terminator is not at the end of the buffer, but within it. 
The current implementation copys the whole buffer into a text field, including the NUL terminator and the remaining chars.

The JIRA ticket contains a JNA based sample program, which prefills the buffer for demonstrating this issue.
If this should be added as a unit test, I'm open for advice how to do that.

---------
### Progress

### Issue
 * [JDK-8281384](https://bugs.openjdk.org/browse/JDK-8281384): Random chars on paste from Windows clipboard (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1724/head:pull/1724` \
`$ git checkout pull/1724`

`$ git checkout pull/1724` \
`$ git pull https://git.openjdk.org/jfx.git pull/1724/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1724`

`$ git pr show -t 1724`

</details>

<a href="https://git.openjdk.org/jfx/pull/1724.diff">https://git.openjdk.org/jfx/pull/1724.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1724#issuecomment-2681987830)
</details>


modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinSystemClipboard.java
```diff
@@ -252,7 +252,15 @@ protected final Object popFromSystem(String mimeFull) {
             if (TEXT_TYPE.equals(mime) || URI_TYPE.equals(mime)) {
                 try {
                     // JDK-8118474 - internal Windows data null terminated
-                    return new String(data, 0, data.length - 2, defaultCharset);
+                    // JDK-8281384 - buffer might be larger than data and null terminator not at the end
```
kevinrushforth, line 255: Minor: We generally don't include the bug ID of the bug we are fixing in a comment (and I see no need here).

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinSystemClipboard.java
```diff
@@ -252,7 +252,15 @@ protected final Object popFromSystem(String mimeFull) {
...
+                    int nullTerm = data.length - 2;
+                    for (int i = 0; i < data.length; i += 2) {
+                        if (data[i] == 0) {
```
kevinrushforth, line 258: Since this is UTF-16, don't you need to check that both the even and odd bytes are 0?

```
                        if (data[i] == 0 && data[i+1] == 0) {
```

If you do this, you will want to validate that the length is even (if it isn't already ensured by `popBytes`).

kevinrushforth (commented):
I left a couple comments. As for an automated test, maybe there is something we could do with FFM, but it might be a bit of a project to figure out how to incorporate it into our test framework.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinSystemClipboard.java
kevinrushforth: Although, since the previous line does, I don't mind if you leave it.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinSystemClipboard.java
Schmidor: done and aborting loop now if there's only one byte left.

kevinrushforth: Looks good now. I'll test and then finish my review.

kevinrushforth:
@Schmidor I can't reproduce the original bug on Windows 11. Back when I did reproduce it, I was running Windows 10, so Microsoft might have fixed something on their end. I'll try to find a Windows 10 system to verify the fix on, but it looks like a good fix nonetheless.

lukostyra (commented):
So far this change looks good, I could reproduce it with attached `ClipboardTest` program and `HelloTextArea` on my Windows 11 machine. After the fix it's no longer an issue and the contents are as expected. Tests also did not show any regressions. I do, however, have one question.

I kind of wish we got correct information about how big Clipboard contents are from Windows API, and I'm kind of surprised it only gives us buffer size information and not actual content information. Are you sure we shouldn't fetch this information from Windows API somewhere instead of manually searching for a null terminator?

kevinrushforth:
That's a good question. If Windows does provide the information, then it would be best to use it. If not, maybe it would be better to move this check in the native code anyway, so that the Java code could just use the returned byte array?

hjohn:
How about fixing this in https://github.com/openjdk/jfx/blob/7a7854c947728daab6a924baa0596ccfc2df1ebb/modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp#L383 ?

It already includes special handling for `CF_DROP`, and you could add special handling for text/unicode:

        if (CF_TEXT == cf || CF_UNICODETEXT == cf) {
               ...
        }

I also checked, it seems that Windows does not provide a way to determine the content length.  For (unicode) text it indeed relyings on nul terminators.

kevinrushforth:
That's a good idea, and is along the lines of what I was thinking when I mentioned moving this check into the native code.

modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
```diff
@@ -403,6 +403,13 @@ HRESULT PopMemory(
                 //as well as corrupted format
                 cdata = 0;
             }
+        } else if(CF_TEXT == cf || CF_UNICODETEXT == cf){
```
mstr2, line 406: Instead of doing that, consider always requesting `CF_UNICODETEXT` from the clipboard. Windows will then convert both `CF_TEXT` and `CF_OEMTEXT` to unicode (see [Synthesized Clipboard Formats](https://learn.microsoft.com/en-us/windows/win32/dataxchg/clipboard-formats)).

modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
```diff
@@ -403,6 +403,13 @@ HRESULT PopMemory(
...
+            for(int i = 0; i < cdata - 1; i+=2){
+                jbyte *pos = me.getMem() + offset + i;
+                if(*(pos) == 0 && *(pos + 1) == 0){
```
mstr2, line 409: 1. `offset` is always 0 here.
2. Minor: add spacing around operators and after keywords (`for`, `if`).

modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
Schmidor: As the existing code was already pretty optimistic about 2 byte chars, is it possible that is already handled somewhere else?
I'm not sure whether this is done explicitly somewhere or if CF_UNICODETEXT is just tried first.

mstr2: Could be, but in any case, the way it's impemented right now doesn't seem to be right. We should only assume 2-byte characters if what's being pulled from the clipboard is actually `CF_UNICODETEXT`.

lukostyra: I did a little bit of digging and we will sometimes request `CF_TEXT` and `CF_OEMTEXT` (see `create_mime_stuff()` in this file). Technically we could change the pairings there and always request text data in `CF_UNICODETEXT`, but I would do some more testing and check if that doesn't cause regressions in some surprise spot.

Schmidor: As @mstr2 said, I think the data is always converted to CF_UNICODETEXT. If it wasn't there should have been more visible problems, as the current code always assumes wide chars.

So I just need to know how much we should change within this ticket: Just removing CF_TEXT from here, or also removing flavors from create_mime_stuff().

lukostyra: I think the point is to explicitly request `CF_UNICODETEXT` and, if Clipboard has `CF_TEXT` or `CF_OEMTEXT` inside, it will automatically convert it to Unicode on output. This way we ensure we always get 2-byte Unicode text independently of Clipboard contents. So I would say to change the flavors in `create_mime_stuff()` as well.

kevinrushforth: That makes sense to me, too. I was originally thinking that removing the `CF_TEXT == cf ||` in this `if` statement would be sufficient, but it's certainly safer, and seems more correct, to also ensure that we request `CF_UNICODETEXT`, given that here and in the calling Java code, we assume 2-byte Unicode.

modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
```diff
@@ -204,11 +204,8 @@ _bstr_t getMime(CLIPFORMAT cf)
 int create_mime_stuff()
 {
     addPair(GLASS_TEXT_PLAIN, CF_UNICODETEXT);
-    addPair(GLASS_TEXT_PLAIN_LOCALE, CF_TEXT);
```
lukostyra, line 207: I would not remove them, rather replace CF_TEXT with CF_UNICODETEXT. We still need to support the MIMEs.

modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
```diff
@@ -204,11 +204,8 @@ _bstr_t getMime(CLIPFORMAT cf)
...
     addPair(GLASS_IMAGE, CF_JAVA_BITMAP);
     addPair(GLASS_FILE_LIST, CF_HDROP);
-    addPair(MS_LOCALE, CF_LOCALE);
```
lukostyra, line 210: CF_LOCALE has to stay as CF_LOCALE - it is not a text format.

modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
```diff
@@ -204,11 +204,8 @@ _bstr_t getMime(CLIPFORMAT cf)
...
-    addPair(MS_OEMTEXT, CF_OEMTEXT);
```
lukostyra, line 211: As above - `s/CF_OEMTEXT/CF_UNICODETEXT`

lukostyra (changes requested):

modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
```diff
@@ -204,11 +204,11 @@ _bstr_t getMime(CLIPFORMAT cf)
 int create_mime_stuff()
 {
     addPair(GLASS_TEXT_PLAIN, CF_UNICODETEXT);
-    addPair(GLASS_TEXT_PLAIN_LOCALE, CF_TEXT);
+    addPair(GLASS_TEXT_PLAIN_LOCALE, CF_UNICODETEXT);
```
lukostyra, line 207: Seems like those changes made JFX unaware of text data ready to be pasted from "outside" - running your test program and trying to paste something to a TextBox (ex. in `HelloTextBoxClipboard`) doesn't work anymore (Paste popup menu option is disabled, Ctrl+V does nothing, whereas both were possible to do on master). I checked that these mime stuff changes are at fault.

My guess now would be that these mime_stuff changes have to be reverted and the actual change should happen as we request Clipboard data - probably a condition changing clipboard format `cf` from `CF_TEXT` and `CF_OEMTEXT` to `CF_UNICODETEXT` right before we call `me.Load()` in `PopMemory()`. Please check if that is in fact the case before committing though, I'm not 100% sure if that would work.

If that works, a comment explaining why we do this clipboard format swap would also come in handy for future generations.

lukostyra (changes requested):

modules/javafx.graphics/src/main/native-glass/win/GlassClipboard.cpp
Schmidor: I can't check if replacing cf with CF_UNICODETEXT there works, as probably that flavor is already always prioritized.

My test for the early NUL terminator explicitly only sets CF_TEXT, but is converted to unicode.

lukostyra: I did give it some tries and it did indeed seem to prioritize `CF_UNICODETEXT`. I suspect we could leave it be without any extra conditions or changes then.

kevinrushforth (approved):
The latest version looks good to me. I tested it and it works as expected.

I left a suggestion and minor formatting comment. Will reapprove if you make changes.

kevinrushforth (approved):

lukostyra (approved):
LGTM

PR #1728, opened 2025-03-04
8351067: Enforce Platform threading use

andy-goryachev-oracle:
Summary
-------

Adds a thread check to a number of `Platform` methods:
```
accessibilityActiveProperty()
getPreferences()
isAccessibilityActive()
```
These methods will throw an `IllegalStateException` if called on a thread other than the JavaFX Application Thread.

Problem
-------

JavaFX allows the Nodes and Scenes to be created and modified on any thread as long as they are not yet attached to a `Window` that is showing.

This is allowed in an implicit assumption that the construction code only modifies the properties of the said Nodes and Scenes, but not other static or global entities.  Concurrent multi-threaded access of such entities not only breaks the initialization of the properties, but also causes the failures down the road, if the change to the global properties happens while the Nodes and Scenes are still being constructed.

Even JavaFX platform developers did not avoid tripping over this issue, as can be illustrated by https://bugs.openjdk.org/browse/JDK-8348987 . 

Solution
--------

Fail each method fast with an  `IllegalStateException` if called from a background thread.

While this solution won't prevent other possible abuse, such as getting a reference to the property in the JavaFX Application Thread and later accessing it in a background thread, adding a check and allowing these methods to fail fast should prevent most likely scenarios.

/csr
/reviewers 2

---------
### Progress
- [x] Change requires CSR request [JDK-8351592](https://bugs.openjdk.org/browse/JDK-8351592) to be approved

### Issues
 * [JDK-8351067](https://bugs.openjdk.org/browse/JDK-8351067): Enforce Platform threading use (**Bug** - P4)
 * [JDK-8351592](https://bugs.openjdk.org/browse/JDK-8351592): Enforce Platform threading use (**CSR**)

### Reviewers
 * [Michael Strauß](https://openjdk.org/census#mstrauss) (@mstr2 - **Reviewer**) Review applies to [e9bb3403](https://git.openjdk.org/jfx/pull/1728/files/e9bb3403d3b8d0a2ff2e8711a667fee01e5bad09)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1728/head:pull/1728` \
`$ git checkout pull/1728`

`$ git checkout pull/1728` \
`$ git pull https://git.openjdk.org/jfx.git pull/1728/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1728`

`$ git pr show -t 1728`

</details>

<a href="https://git.openjdk.org/jfx/pull/1728.diff">https://git.openjdk.org/jfx/pull/1728.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1728#issuecomment-2698660570)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/application/Platform.java
+++ b/modules/javafx.graphics/src/main/java/javafx/application/Platform.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -421,8 +421,8 @@ public final class Platform {
      * This property is typically set to true the first time an
      * assistive technology, such as a screen reader, requests
      * information about any JavaFX window or its children.
-     *
-     * <p>This method may be called from any thread.</p>
+     * <p>
+     * This property can be accessed only from the FX application thread.
      *
      * @return the read-only boolean property indicating if accessibility is active
      *
--- a/modules/javafx.graphics/src/main/java/javafx/application/Platform.java
+++ b/modules/javafx.graphics/src/main/java/javafx/application/Platform.java
@@ -413,6 +413,7 @@ public final class Platform {
     private static ReadOnlyBooleanWrapper accessibilityActiveProperty;
 
     public static boolean isAccessibilityActive() {
+        Toolkit.getToolkit().checkFxUserThread();
         return accessibilityActiveProperty == null ? false : accessibilityActiveProperty.get();
     }
 
@@ -426,9 +427,12 @@ public final class Platform {
      *
      * @return the read-only boolean property indicating if accessibility is active
      *
+     * @throws IllegalStateException if this method is called on a thread
+     *     other than the JavaFX Application Thread.
      * @since JavaFX 8u40
      */
     public static ReadOnlyBooleanProperty accessibilityActiveProperty() {
+        Toolkit.getToolkit().checkFxUserThread();
         if (accessibilityActiveProperty == null) {
             accessibilityActiveProperty = new ReadOnlyBooleanWrapper(Platform.class, "accessibilityActive");
             accessibilityActiveProperty.bind(PlatformImpl.accessibilityActiveProperty());
--- a/modules/javafx.graphics/src/main/java/javafx/application/Platform.java
+++ b/modules/javafx.graphics/src/main/java/javafx/application/Platform.java
@@ -422,7 +422,7 @@ public final class Platform {
      * assistive technology, such as a screen reader, requests
      * information about any JavaFX window or its children.
      * <p>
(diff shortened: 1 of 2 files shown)


modules/javafx.graphics/src/main/java/javafx/application/Platform.java
```diff
@@ -421,8 +421,8 @@ public static boolean isAccessibilityActive() {
      * This property is typically set to true the first time an
      * assistive technology, such as a screen reader, requests
      * information about any JavaFX window or its children.
-     *
-     * <p>This method may be called from any thread.</p>
+     * <p>
+     * This property can be accessed only from the FX application thread.
```
kevinrushforth, line 425: "FX application thread" --> "JavaFX Application Thread"

kevinrushforth (commented):
One wording suggestion, but otherwise looks fine.

andy-goryachev-oracle:
@mstr2 : should we also enforce the access rules in `Platform::getPreferences()` ?

mstr2:
> @mstr2 : should we also enforce the access rules in `Platform::getPreferences()` ?

In principle, yes. But just checking at the point of calling `Platform.getPreferences()` doesn't seem to be sufficient. An application could easily call `getPreferences()` on the FX thread, and then unsafely access the returned instance on another thread.

We should probably first come up with a systematic approach on how to solve these kinds of problems, and not do ad-hoc solutions here and there.

andy-goryachev-oracle:
This **is** the time we systematically approach the subject, see https://bugs.openjdk.org/browse/JDK-8348987

The case we are trying to deal with explicitly is initialization of Nodes in the background threads which is permitted.  We are not addressing the misuse of APIs by the applications, only providing a mechanism to fail early in the narrow case when controls/skins/nodes try access the Platform globals.

I believe ::getPReferences() is the only API in Platform left without the explicit check, thus the question.

mstr2:
> We are not addressing the misuse of APIs by the applications, only providing a mechanism to fail early in the narrow case when controls/skins/nodes try access the Platform globals.

Then I diagree with the problem statement. Calling out to `checkFxUserThread` in the property getter is a very superficial fix that's not materially better than simply documenting that we assume the property only be used on the FX thread.

On a related note, I would like to see us move to a more structured form of documentation, for example with custom javadoc tags. This makes it easier to see information at a glance in the generated documentation.

andy-goryachev-oracle:
> move to a more structured form of documentation, for example with custom javadoc tags.

what custom tags?  could you elaborate?

mstr2:
> what custom tags? could you elaborate?

Maybe a custom tag like `@threadAccess JavaFX application thread`. This would then show up in the generated property list instead of the summary text.

andy-goryachev-oracle:
> @threadAccess JavaFX application thread

Out of scope for this PR, but it might be a good idea.

mstr2 (approved):

kevinrushforth (approved):
Looks good. I left one minor comment on the test.

kevinrushforth (approved):

kevinrushforth:
The only change since @mstr2 approved this PR is to the name of a test method. As such, a single re-review should be sufficient, which I've already done, so I'm lowering the (re)review requirement down to 1 to reflect this.

/reviewers 1

PR #1733, opened 2025-03-13
JDK-8351867 No UI changes while iconified

hjohn:
Adds code to trigger a scene update when a Window is restored

This seems to solve https://bugs.openjdk.org/browse/JDK-8351867 and https://bugs.openjdk.org/browse/JDK-8146479

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassWindowEventHandler.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassWindowEventHandler.java
@@ -62,6 +62,16 @@ class GlassWindowEventHandler extends Window.EventHandler implements Supplier<Vo
             case WindowEvent.RESTORE:
                 stage.stageListener.changedIconified(false);
                 stage.stageListener.changedMaximized(false);
+
+                // Force a redraw as the scene may have changed while minimized (and its size won't have)
+                QuantumToolkit.runWithRenderLock(() -> {
+                    GlassScene scene = stage.getScene();
+                    if (scene != null) {
+                        scene.updateSceneState();
+                    }
+                    return null;
+                });
+
                 break;
             case WindowEvent.MOVE: {
                 float wx = window.getX();


beldenfox:
When the window is restored glass calls notifySize with RESTORED and then notifyRepaint with valid dimensions. Why is the repaint being dropped?

hjohn:
> When the window is restored glass calls notifySize with RESTORED and then notifyRepaint with valid dimensions. Why is the repaint being dropped?

Can you give me a bit more detail what you mean here?  I suspect it might be related to the size not actually changing, and so a size update with the same size does nothing?

beldenfox:
> Can you give me a bit more detail what you mean here? I suspect it might be related to the size not actually changing, and so a size update with the same size does nothing?

Sorry, I'm asking a general question about how JavaFX deals with repaint events issued by the OS. I've never understood how JavaFX's painting model interfaces with OS level calls that expect drawing to occur on demand and synchronously. But that conversation should probably be moved to another forum.

hjohn:
> My thought on this bug was to fix it in the scenegraph, rather than forcing a complete repaint in the toolkit. The scenegraph tracks the state of what is dirty and would only need to repaint if something has changed (and probably only those parts that have changed).
> 
> @arapte Can you take a look at this and see what you think? I am not able to look at it for at least a week.

Yeah, I'm sort of mimicking what happens when a RESCALE or MOVE occurs, so it issimilar to what some of the other handlers do (MOVE does `updateSceneState`, and RESCALE does that and `entireSceneNeedsRepaint`).  I'm not that well versed in differences between these two, but it seemed not a bad place to fix this as restores happen very rarely.

andy-goryachev-oracle:
@hjohn this did not fix the issue for me on macOS 15.3.1 M1

kevinrushforth:
Yeah, your solution might be fine. I just wanted to take a look at an alternative (or rather ask Ambarish to).

hjohn:
> @hjohn this did not fix the issue for me on macOS 15.3.1 M1 (the reproducer still shows "Initial State" label)

@andy-goryachev-oracle Hm, I can't test it on Mac, but this solution works on Windows.  I just tested again with your reproducer, and it fails without this change and works with it.  Maybe someone with a Mac can take a closer look?

beldenfox:
I just verified that this PR does not fix the original issue on the Mac.

Note: I could not reproduce the bug on my Mac 15.3.2 M2 Max using the latest master. Then I went to System Settings > Desktop & Dock and turned on "Minimize windows into application icon". Then the problem reproduced. But when I toggled that switch back off and the bug still reproduced.

This may be irrelevant but I did notice a difference between Mac and Windows. After glass sends notifySize RESTORE it sends notifyRepaint on Windows but not on Mac. In fact I don't think the Mac ever sends notifyRepaint.

beldenfox:
Adding `entireSceneNeedsRepaint` fixes this on the Mac.

arapte:
This does fix the issue on windows.
The issue on mac seems intermittent even without this change, and stays intermittent with this change.

The change seems to achieve the intended but the trigger check may not be the best way. `updateSceneState` is used when there are any changes with scene properties.
The redraw should be triggered by checking the change in scene graph. This change just triggers irrespective of that.

beldenfox:
At least on the Mac there seems to be a timing issue. If the MIMIMIZE notification is sent before a scheduled draw occurs the drawing code will clear the scene dirty bit but then skip the actual drawing since the window is minimized. But I've also seen cases where drawing completes before MINIMIZE is sent.

When the window is restored notifyRepaint will be called on Windows and Linux which looks like it will call entireSceneNeedsRepaint to set the scene dirty bit (based on my reading of the code). Windows calls notifyRepaint whenever it gets a WM_PAINT event and I've verified that that happens after a window is restored. Looking at the sources it seems that Linux calls notifyRepaint _only_ when a window is restored (!)

Mac never calls notifyRepaint. There's code to call it but given where it's located I don't think it will ever be executed.

It would be nice to get some clarification on when notifyRepaint should be called and get the platforms to behave more consistently.

hjohn:
> This does fix the issue on windows. The issue on mac seems intermittent even without this change, and stays intermittent with this change.
> 
> The change seems to achieve the intended but the trigger check may not be the best way. `updateSceneState` is used when there are any changes with scene properties. The redraw should be triggered by checking the change in scene graph. This change just triggers irrespective of that.

I think the bug can also appear without the scene graph having changed, when you iconify immediately programmatically (ie. before showing the window).  The change detection would need to make sure it also detects the case where a scene was not drawn yet at all:

```java
public class App extends Application {

        Application.launch(args);
    }

    @Override
        Scene scene = new Scene(new Label("This should be modified when the signal is received"), 500, 500);

        stage.setScene(scene);

        stage.setIconified(true);
        stage.show();
    }
}
```

kevinrushforth:
Reviewers: @arapte @lukostyra

hjohn:
This doesn't look like an easy fix; although it fixes the issue for me on Windows, I don't have access to a Mac to fix the problem on that platform as well.  Perhaps someone else wants to take a look at this issue.

beldenfox:
First an errata: I wrote earlier that notifyRepaint is only called on Linux when restoring a window. I was wrong, it's also called for EXPOSE events.

Anyway, I can reproduce this bug on all three platforms.

JavaFX won't paint to a minimized window. This suggests to me that when we transition out of the minimized state someone needs to call `entireSceneNeedsRepaint`. The glass code on Windows and Linux both issue `notifyRepaint` which eventually calls `entireSceneNeedsRepaint`.  The Mac code never issues `notifyRepaint` and  `entireSceneNeedsRepaint` isn't called.

I can tweak the Mac code to call `notifyRepaint` to match Windows and Linux but presumably that won't fix the bug. It looks like some other bit of state gets out of whack due to the way the system refuses to paint while minimized. I have no idea whether `updateSceneState` is the appropriate call to get the system back into sync or not.

I have verified that on all platforms `updateSceneState` is not being called when the window is restored. I'll also note that with this test case there's no focus ring or flashing caret that needs to be drawn which seems to make a difference.

Reproducing this on macOS 15 is complicated by some weirdness going on under the hood. The address of the main screen changes unexpectedly. This causes glass to send `notifyMoveToAnotherScreen` which resets a whole lot of state. This can happen when I first move a window or when I iconify and de-iconify it. Depending on the "Minimize windows to application icon" setting this can happen every time I de-iconify or only the first time. I have no idea why the system keeps generating new NSScreen objects on an ongoing basis.

beldenfox:
I think this PR is the correct fix on Windows and Linux. The scene state tracks the minimized state of the window. When the window is restored someone needs to tell the scene state to go update the isWindowMinimized flag.

The Mac will require another tweak. Personally I would just modify glass to send notifyRepaint in this specific instance. I suppose that should be treated as a separate bug. Unfortunately it's going to be a bear to test given the strange things happening under the hood with NSScreen (at least on my system).

andy-goryachev-oracle:
I can help with the testing: I have 15.3.2 with one or two external monitors.
What might help is to enumerate scenarios we want to be tested.

beldenfox:
> I can help with the testing: I have 15.3.2 with one or two external monitors. What might help is to enumerate scenarios we want to be tested.

Thanks for the offer. I'm also running 15.3.2. I suspect I'll have to try test on an older version of the OS to avoid all those unexpected screen change notifications.

> Do you want to address the macOS side in a separate PR?

Maybe? It's clear that someone needs to call `entireSceneNeedsRepaint` to trigger the redraw. Currently there are many places in the core code that do this (for example, when the window's size changes). So maybe this is just another instance where the core should handle it. In that case it could be folded into this PR.

The alternative is to update the Mac glass code to more closely match what Windows and Linux are doing. In that case it should be a separate PR. But it's beginning to look like notifyPaint is sort of vestigial. Outside of this bug the Mac is just fine not calling it because the core handles everything internally.

From my newbie perspective it looks like this is just another area where the core code should be calling `entireSceneNeedsRepaint`. But this is not an area I'm familiar with.

tsayao:
I spotted this while working on Linux glass. The test case I was doing was "showing" an iconified fullscreened stage. Linux glass has a repaint call, but it seems wrong.

tsayao:
Hmm, it also relates to something else I was intrigued by:

What happens if the Stage is both Maximized and Iconified? If `RESTORED` assumes that the window is neither Maximized nor Iconified, then when a Maximized and Iconified window is de-iconified (presented), it should not be considered `RESTORED` — and in this case, the bug still persists.

tsayao:
About the comment on the change:

After some experiments, I’d say the size can change because you can call `setFullScreen(true)` or `setMaximized(true)` and then deiconify. While this doesn’t directly set the width and height of the Scene, the actual scene size does change. 

Anyways, this is just a comment for a case to consider, but the change does make the scene repaint in this case.

kevinrushforth:
In reading through all of the comments again, I think that the proposed solution, with the addition of `entireSceneNeedsRepaint`, is a reasonable one. The only other question I have is: Would it be possible to create an automated test that fails (on macOS at least) without the fix and passes on all platforms with the fix?

beldenfox:
I'll work on creating an automated test for this.

beldenfox:
Test file attached. [drawdeiconify.patch](https://github.com/user-attachments/files/20165986/drawdeiconify.patch)

The test fails on all platforms without this PR. With this PR the test succeeds on Linux and Windows for regular windows and fails when de-iconifying to a maximized state. It also fails on Windows for TRANSPARENT stages since we never get a WM_PAINT message and so glass never calls notifyRepaint.

lukostyra:
>The test fails on all platforms without this PR. With this PR the test succeeds on Linux and Windows for regular windows and fails when de-iconifying to a maximized state. It also fails on Windows for TRANSPARENT stages since we never get a WM_PAINT message and so glass never calls notifyRepaint.

I'm trying to figure out how to exactly proceed with this. It would be nice to add this change and to also have the test in the repository, but it does fail for some cases like you mentioned (and I also verified that on my Windows machine). Would those cases be fixable?

beldenfox:
> I'm trying to figure out how to exactly proceed with this. It would be nice to add this change and to also have the test in the repository, but it does fail for some cases like you mentioned (and I also verified that on my Windows machine). Would those cases be fixable?

Sorry, my earlier comment was a bit terse. This is all fixable. When the stage is de-iconified we need to call `updateSceneState` to fix up the state and then `entireSceneNeedsRepaint` to repaint on the next pulse. Currently for the second step we're relying on Glass to call `notifyPaint` but that never happens on Mac and also doesn't happen for TRANSPARENT stages on Windows (I haven't tested Linux). Rather than expecting every Glass implementation to get this right for every stage style we should call `entireSceneNeedsRepaint` in the core code. That should fix all cases including TRANSPARENT stages on Windows.

lukostyra:
I have noticed the tests for maximized stages fail on my system:

```
DrawAfterDeiconifyTest > maximizedStageRedrawsAfterDeiconify(StageStyle) > [1] DECORATED FAILED
    org.opentest4j.AssertionFailedError: expected:rgba(255,105,180,255) but was:rgba(0,255,0,255)
        at app//org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:38)
        at app//org.junit.jupiter.api.Assertions.fail(Assertions.java:138)
        at app//test.robot.testharness.VisualTestBase.assertColorEquals(VisualTestBase.java:168)
        at app//test.robot.javafx.stage.DrawAfterDeiconifyTest.lambda$redrawsAfterDeiconify$6(DrawAfterDeiconifyTest.java:110)

DrawAfterDeiconifyTest > maximizedStageRedrawsAfterDeiconify(StageStyle) > [2] UNDECORATED FAILED
    org.opentest4j.AssertionFailedError: expected:rgba(255,105,180,255) but was:rgba(0,255,0,255)
        at app//org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:38)
        at app//org.junit.jupiter.api.Assertions.fail(Assertions.java:138)
        at app//test.robot.testharness.VisualTestBase.assertColorEquals(VisualTestBase.java:168)
        at app//test.robot.javafx.stage.DrawAfterDeiconifyTest.lambda$redrawsAfterDeiconify$6(DrawAfterDeiconifyTest.java:110)

DrawAfterDeiconifyTest > maximizedStageRedrawsAfterDeiconify(StageStyle) > [3] TRANSPARENT FAILED
    org.opentest4j.AssertionFailedError: expected:rgba(255,105,180,255) but was:rgba(0,255,0,255)
        at app//org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:38)
        at app//org.junit.jupiter.api.Assertions.fail(Assertions.java:138)
        at app//test.robot.testharness.VisualTestBase.assertColorEquals(VisualTestBase.java:168)
        at app//test.robot.javafx.stage.DrawAfterDeiconifyTest.lambda$redrawsAfterDeiconify$6(DrawAfterDeiconifyTest.java:110)
```

This is on Windows 11, the Stage in fact does not change its background. I'm not sure what the problem could be, or if it's fixable.

beldenfox:
> I have noticed the tests for maximized stages fail on my system:

That's expected. This PR fixes the case where the window goes from iconified to restored. When looking at the code I realized the same fix needs to be applied when the window goes from iconified to maximized so I added that to the test.

hjohn:
> > I have noticed the tests for maximized stages fail on my system:
> 
> That's expected. This PR fixes the case where the window goes from iconified to restored. When looking at the code I realized the same fix needs to be applied when the window goes from iconified to maximized so I added that to the test.

Do I need to add additional code for that case?  The fix is now in `WindowEvent.RESTORE`, which I think should be catching both cases, or does de-iconifying to a maximized stage somehow not send a restore event?  The terms are somewhat confusing; if `RESTORE` indicates go to a non-maximized/non-minimized/non-iconified "normal" state, then I suppose I need to add more code in `MAXIMIZE` as well...

Edit: I see that iconified state is explicitly set to `false` in the `MAXIMIZE` branch, so it looks like we'd need to do the same there.  I'll do some tests.

beldenfox:
> Do I need to add additional code for that case? The fix is now in `WindowEvent.RESTORE`, which I think should be catching both cases, or does de-iconifying to a maximized stage somehow not send a restore event?

At least on Windows you can go straight from an iconified state to a maximized state and the only event you'll see is WindowEvent.MAXIMIZE. That should be true on the Mac as well but there are bugs in the way glass handles maximized windows so I'm not entirely sure what the event sequence is.

andy-goryachev-oracle:
@hjohn you might want to add [JDK-8146479](https://bugs.openjdk.org/browse/JDK-8146479) to the PR via `/ issue add` command.

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassWindowEventHandler.java
```diff
@@ -105,15 +107,7 @@ public Void get() {
...
-                // We need to sync the new scales for painting
-                QuantumToolkit.runWithRenderLock(() -> {
-                    GlassScene scene = stage.getScene();
-                    if (scene != null) {
-                        scene.entireSceneNeedsRepaint();
-                        scene.updateSceneState();
-                    }
-                    return null;
-                });
+                forceRepaint();
```
andy-goryachev-oracle, line 110: (github diff is confusing, this code is for `WindowEvent.RESCALE` case)

I suspect this might also fix an issue that I encountered in one of my applications which uses Canvas-based text editor.  After the macOS is woken up after a sleep with an external monitor attached, the text on retina looked blurred.  I could not get a reproducer in time, but I think it was because while I put it to sleep, the scene got re-rendered on the external monitor going from scale 2 to scale 1, but when woken up, the scene was correctly re-rendered for retina but the canvas was not resized, requiring it to be scaled.

I'll try to test this scenario, but it may take a few days.

andy-goryachev-oracle:
Looks good on macOS 15.5 M1.
Could we test this tested on linux please?  I'll add one reviewer temporarily for now just for that.

/reviewers 3

andy-goryachev-oracle (approved):

andy-goryachev-oracle:
I am going to launch a headful CI run, will report in about an hour.

edit: @lukostyra already started one, thanks!

edit2: headful CI run passes (with one unrelated failure with QPathTest on mac-13-aarch64 that we occasionally see).

beldenfox:
Tested on Ubuntu 22.04. Test failed in master branch and succeeded with this PR.

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassWindowEventHandler.java
hjohn: Sure, no hurry to integrate this

andy-goryachev-oracle: actually, the reason I bumped the reviewer count is to let you add the issue and make sure we get the testing results for linux (which we got) and our CI run (which we did).

I may not be able to test my failing scenario before you integrate anyway.  We are good to go.

andy-goryachev-oracle:
thank you @hjohn for fixing this bug!

PR #1735, opened 2025-03-13
8351878: RichTextArea: Pasting from RTF doesn't apply formatting

andy-goryachev-oracle:
Fixed two issues found in importing RTF text:

- Arabic import
- missing font size

The former was caused by my removal of the character decoder code present in the original JDK RTF parser/reader.  Why did I do that?

### DIFF
--- a/modules/jfx.incubator.richtext/src/main/java/com/sun/jfx/incubator/scene/control/richtext/rtf/AttrSet.java
+++ b/modules/jfx.incubator.richtext/src/main/java/com/sun/jfx/incubator/scene/control/richtext/rtf/AttrSet.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2023, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -87,6 +87,7 @@ public class AttrSet {
         return StyleAttributeMap.builder().
             setBold(getBoolean(StyleAttributeMap.BOLD)).
             setFontFamily(getString(StyleAttributeMap.FONT_FAMILY)).
+            setFontSize(getDouble(StyleAttributeMap.FONT_SIZE)).
             setItalic(getBoolean(StyleAttributeMap.ITALIC)).
             setTextColor(getColor(StyleAttributeMap.TEXT_COLOR)).
             setUnderline(getBoolean(StyleAttributeMap.UNDERLINE)).
@@ -109,6 +110,14 @@ public class AttrSet {
         Object v = attrs.get(attr);
         if (v instanceof Color c) {
             return c;
+        }
+        return null;
+    }
+
+    private Double getDouble(Object attr) {
+        Object v = attrs.get(attr);
+        if (v instanceof Number n) {
+            return n.doubleValue();
         }
         return null;
     }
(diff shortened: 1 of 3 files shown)


Ziad-Mid:
The tests run successfully with the fix however there are a few things that do not work as expected I guess :
1- Pasting text from notes app (MacOS) that uses numbered list in some cases shows 0. instead of the item number example: 
- in note app do something like this 
```
Test
1. one
2. two
3. three
Test
```
- Copy and paste it is rendred like this 
```
Test
0. one
0. two
0. three
Test
```
2- Tested pasting text written in outlook and pasting it and it do not apply formatting neither for RTL or LTR text
3- another thing the arabic diacritics text is not rendered correctly (this may not be related to this issue specifically) you can try with this text  السَّلَامُ عَلَيْكُمْ

andy-goryachev-oracle:
@Ziad-Mid : 
interesting, thank you!

1. The list formatting was not explicitly supported, so in a sense it works as designed (all numbered items are represented by `{\listtext	0.	}`.  This will likely be a separate enhancement.

2. RTL/LTR attributes are not yet supported by the demo due to many bugs in RTL.

3. Arabic diacritics rendered with the System font look bad, I agree.  Maybe @prrace can weigh on it, looks like changing the font to one of the Arabic ones does improve the rendering, although the result appears not as good compared to Chrome:

Screenshot 2025-03-24 at 13 58 56

Screenshot 2025-03-24 at 13 59 33

Screenshot 2025-03-24 at 14 02 47

andy-goryachev-oracle:
Weird, even "Noto Sans Arabic" font produces weirdly positioned diacritics in JavaFX, while it's ok in MS Word:

Screenshot 2025-03-24 at 14 19 22

Ziad-Mid:
> 2. RTL/LTR attributes are not yet supported by the demo due to many bugs in RTL.
By RTL and LTR I just meant normal text orientation and arabic one , when writing a text in outlook and copying it then pasting it doesn't keep the formatting

andy-goryachev-oracle:
> doesn't keep the formatting

please be more specific - maybe include a screenshot of the outlook message vs. rich text area?

I can see the pasted Arabic text rendered correctly (that is, it looks similar to the text in outlook and going from right to left).  The RTL _paragraph attribute_ is not yet supported, so the paragraph consisting of only Arabic text might look left-aligned contrary to expectation.

andy-goryachev-oracle:
> The font and size are not applied

Thank you, the screenshot helped!

The problem was in the way HTML text was copied to Outlook (for some reason, Outlook prioritizes HTML over RTF when pasting).  The font sizes in HTML were using points (`pt.`) instead of pixels (`px.`).

In addition to that, copying to HTML did not process boolean attributes correctly.

Thank you for finding these issues!

andy-goryachev-oracle:
Created https://bugs.openjdk.org/browse/JDK-8353003 for the diacritic marks issue.

kevinrushforth:
Reviewers: @lukostyra @Ziad-Mid

lukostyra (approved):
LGTM - did some testing, both with issue-attached Word file and some standard Polish diacritic test sentences and everything seems to work fine.

Question coming from my own curiosity - when you post a sentence containing diacritics (ex. shortest tester for all diacritics in Polish - "Zażółć gęślą jaźń") you can see that some fonts don't support those special characters (ex. Arial Rounded MT Bold on Windows) so as a fallback they are drawn with a different font that supports them. It's quite normal behavior to do so I've seen often, but do you know what is the process for determining which font to use as a replacement in RTA? Is it something we have to worry about?

andy-goryachev-oracle:
> LGTM
than you!


> some fonts don't support those special characters (ex. Arial Rounded MT Bold on Windows) so as a fallback they are drawn with a different font that supports them. It's quite normal behavior to do so I've seen often, but do you know what is the process for determining which font to use as a replacement in RTA?

A good question!  Maybe @prrace can explain the process.  Looking in the Monkey Tester, it is clear that the font substitution algorithm is unable to pick the right fallback font from Arial Rounded MT Bold on macOS (while Swing simply refuses to render anything):

Screenshot 2025-04-04 at 08 21 04

Ziad-Mid (approved):

andy-goryachev-oracle:
Thank you all for reviewing!
/integrate

PR #1737, opened 2025-03-14
8091629: [Mac] Support the automatic hiding of the "java" menu bar item when using the system menu bar

beldenfox:
On the Mac the application menu (the one to the right of the Apple logo menu) is created by JavaFX and there is no public API to customize or localize it. This PR allows a client to get rid of it and replace it with an entirely custom JavaFX Menu.

Prior to OS X 10.6 (Snow Leopard) the application menu required special handling. Nowadays it's just the first item in the NSApp's mainMenu. The only restriction is that the OS ignores the item's title and instead uses the application's name as specified in the CFBundle.

This PR is a work-in-progress/trial balloon. It provides a new property on the MenuBar that controls whether the system will use the default menus. It also supplies the calls necessary to implement the standard menu items "Hide <application>", "Hide Others", and "Show All" which need to be routed to native calls on the NSApplication object. A manual test app is available in tests/manual/controls/DefaultAppMenu.java.

---------
### Progress
- [ ] Change requires a CSR request matching fixVersion jfx25 to be approved (needs to be created)

### Issue
 * [JDK-8091629](https://bugs.openjdk.org/browse/JDK-8091629): [Mac] Support the automatic hiding of the "java" menu bar item when using the system menu bar (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1737/head:pull/1737` \
`$ git checkout pull/1737`

`$ git checkout pull/1737` \
`$ git pull https://git.openjdk.org/jfx.git pull/1737/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1737`

`$ git pr show -t 1737`

</details>

<a href="https://git.openjdk.org/jfx/pull/1737.diff">https://git.openjdk.org/jfx/pull/1737.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1737#issuecomment-2725484205)
</details>


kevinrushforth:
This should be discussed on the mailing list first. I recommend moving it to Draft for now. If it goes forward it will need a CSR.

/reviewers 2 reviewers
/csr

modules/javafx.graphics/src/main/java/javafx/application/ApplicationServices.java
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright (c) 2024, Oracle and/or its affiliates. All rights reserved.
```
OguzOzkeroglu, line 2:
```suggestion
 * Copyright (c) 2025, Oracle and/or its affiliates. All rights reserved.
```

jperedadnr:
@beldenfox I've tested the PR after solving the minor conflicts and it looks promising! 
It would be great if this PR moves forward.

beldenfox:
@jperedadnr Thanks for taking a look! I'll go clean up the conflicts.

My interest in this has cooled quite a bit. It's part of a bundle of other issues and I'm not sure it's worth fixing one without addressing the others. Another problem is that the system menu bar tracks the focused window so when all windows are iconified the menu bar goes away and the old default application menu re-appears. There's some existing logic that could fix that but it would have to be cleaned up and tested.

The biggest problem is that when implicit exit is turned off the default Quit menu item stops working (there's a bug report on that already). Solving that means crafting an API for handling quit requests and that's more than I've been willing to take on right now.

If I were a JavaFX developer on macOS I would probably pick up a third-party package that fixes all of these issues in one go (even if it works by digging around in the JavaFX internals). I'm not sure this isolated PR would be all that compelling.

beldenfox:
Definitely the wrong API for this. A new proposed API is in the works.

PR #1741, opened 2025-03-21
8351733: Crash when creating too many nested event loops

beldenfox:
There is an undocumented limit on nesting calls to CFRunLoopRun (or the equivalent wrapper NSRunLoop methods). When the limit is hit the OS terminates the Java app. The situation arises when a JavaFX app creates too many nested event loops from within  Platform.runLater runnables.

This PR doesn't change the limit (which is 250+ nested loops) but it does throw an exception just before the limit is reached so a JavaFX developer will get a useful Java stack trace instead of an OS crash log.

On the Mac the nested event loop has two stages: first we ask the run loop to run, then we pull an event out and process it. A Platform.runLater runnable is executed in the first stage so if the runnable starts a new nested event loop the system will re-enter CFRunLoopRun. The same isn't true if an input event handler starts a new nested event loop; at that point we're in stage two and are past the call to CFRunLoopRun.

---------
### Progress
- [x] Change requires CSR request [JDK-8353287](https://bugs.openjdk.org/browse/JDK-8353287) to be approved

### Issues
 * [JDK-8351733](https://bugs.openjdk.org/browse/JDK-8351733): Crash when creating too many nested event loops (**Bug** - P3)
 * [JDK-8353287](https://bugs.openjdk.org/browse/JDK-8353287): Crash when creating too many nested event loops (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1741/head:pull/1741` \
`$ git checkout pull/1741`

`$ git checkout pull/1741` \
`$ git pull https://git.openjdk.org/jfx.git pull/1741/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1741`

`$ git pr show -t 1741`

</details>

<a href="https://git.openjdk.org/jfx/pull/1741.diff">https://git.openjdk.org/jfx/pull/1741.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1741#issuecomment-2744438550)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
@@ -26,7 +26,6 @@
 package com.sun.javafx.tk.quantum;
 
 import javafx.application.ConditionalFeature;
-import javafx.application.Platform;
 import javafx.geometry.Dimension2D;
 import javafx.scene.image.Image;
 import javafx.scene.image.PixelBuffer;
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
@@ -26,6 +26,7 @@
 package com.sun.javafx.tk.quantum;
 
 import javafx.application.ConditionalFeature;
+import javafx.application.Platform;
 import javafx.geometry.Dimension2D;
 import javafx.scene.image.Image;
 import javafx.scene.image.PixelBuffer;
@@ -619,10 +620,17 @@ public final class QuantumToolkit extends Toolkit {
         return stage;
     }
 
+    private boolean maxNestedEventLoopsHit() {
+        if (eventLoopMap == null) {
+            return false;
+        }
+        return eventLoopMap.size() >= Platform.MAX_NESTED_EVENT_LOOPS;
+    }
+
     @Override public boolean canStartNestedEventLoop() {
         checkFxUserThread();
 
-        return inPulse == 0 && Application.GetApplication().canStartNestedEventLoop();
+        return inPulse == 0 && !maxNestedEventLoopsHit();
     }
 
     @Override public Object enterNestedEventLoop(Object key) {
@@ -633,8 +641,9 @@ public final class QuantumToolkit extends Toolkit {
         }
 
         if (!canStartNestedEventLoop()) {
-            if (!Application.GetApplication().canStartNestedEventLoop()) {
-                throw new RuntimeException("Exceeded limit on nested event loops");
+            if (maxNestedEventLoopsHit()) {
+                throw new RuntimeException("Exceeded limit on number of nested event loops (" +
+                    Platform.MAX_NESTED_EVENT_LOOPS + ")");
             } else {
                 throw new IllegalStateException("Cannot enter nested loop during animation or layout processing");
             }
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
@@ -641,7 +641,7 @@ public final class QuantumToolkit extends Toolkit {
 
         if (!canStartNestedEventLoop()) {
             if (maxNestedEventLoopsHit()) {
-                throw new RuntimeException("Exceeded limit on number of nested event loops (" +
+                throw new IllegalStateException("Exceeded limit on number of nested event loops (" +
                     PlatformImpl.MAX_NESTED_EVENT_LOOPS + ")");
             } else {
                 throw new IllegalStateException("Cannot enter nested loop during animation or layout processing");
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
@@ -624,7 +624,7 @@ public final class QuantumToolkit extends Toolkit {
         if (eventLoopMap == null) {
             return false;
         }
-        return eventLoopMap.size() >= Platform.MAX_NESTED_EVENT_LOOPS;
+        return eventLoopMap.size() >= PlatformImpl.MAX_NESTED_EVENT_LOOPS;
     }
 
     @Override public boolean canStartNestedEventLoop() {
@@ -643,7 +643,7 @@ public final class QuantumToolkit extends Toolkit {
         if (!canStartNestedEventLoop()) {
             if (maxNestedEventLoopsHit()) {
                 throw new RuntimeException("Exceeded limit on number of nested event loops (" +
-                    Platform.MAX_NESTED_EVENT_LOOPS + ")");
+                    PlatformImpl.MAX_NESTED_EVENT_LOOPS + ")");
             } else {
                 throw new IllegalStateException("Cannot enter nested loop during animation or layout processing");
             }
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
@@ -622,7 +622,7 @@ public final class QuantumToolkit extends Toolkit {
     @Override public boolean canStartNestedEventLoop() {
         checkFxUserThread();
 
-        return inPulse == 0;
+        return inPulse == 0 && Application.GetApplication().canStartNestedEventLoop();
     }
 
     @Override public Object enterNestedEventLoop(Object key) {
@@ -633,7 +633,11 @@ public final class QuantumToolkit extends Toolkit {
         }
 
         if (!canStartNestedEventLoop()) {
-            throw new IllegalStateException("Cannot enter nested loop during animation or layout processing");
+            if (!Application.GetApplication().canStartNestedEventLoop()) {
+                throw new RuntimeException("Exceeded limit on nested event loops");
+            } else {
+                throw new IllegalStateException("Cannot enter nested loop during animation or layout processing");
+            }
         }
 
         if (eventLoopMap == null) {
--- a/modules/javafx.graphics/src/main/java/javafx/application/Platform.java
+++ b/modules/javafx.graphics/src/main/java/javafx/application/Platform.java
@@ -302,7 +302,7 @@ public final class Platform {
      * @throws IllegalStateException if this method is called on a thread
      * other than the JavaFX Application Thread.
      *
-     * @throws RuntimeException if this call would exceed the maximum
+     * @throws IllegalStateException if this call would exceed the maximum
      * number of nested event loops.
      *
      * @return the value passed into the corresponding call to exitEventLoop
--- a/modules/javafx.graphics/src/main/java/javafx/application/Platform.java
+++ b/modules/javafx.graphics/src/main/java/javafx/application/Platform.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -413,6 +413,7 @@ public final class Platform {
     private static ReadOnlyBooleanWrapper accessibilityActiveProperty;
 
     public static boolean isAccessibilityActive() {
+        Toolkit.getToolkit().checkFxUserThread();
         return accessibilityActiveProperty == null ? false : accessibilityActiveProperty.get();
     }
 
@@ -421,14 +422,17 @@ public final class Platform {
      * This property is typically set to true the first time an
      * assistive technology, such as a screen reader, requests
      * information about any JavaFX window or its children.
-     *
-     * <p>This method may be called from any thread.</p>
+     * <p>
+     * This property can be accessed only from the JavaFX Application Thread.
      *
      * @return the read-only boolean property indicating if accessibility is active
      *
+     * @throws IllegalStateException if this method is called on a thread
+     *     other than the JavaFX Application Thread.
      * @since JavaFX 8u40
      */
     public static ReadOnlyBooleanProperty accessibilityActiveProperty() {
+        Toolkit.getToolkit().checkFxUserThread();
         if (accessibilityActiveProperty == null) {
             accessibilityActiveProperty = new ReadOnlyBooleanWrapper(Platform.class, "accessibilityActive");
             accessibilityActiveProperty.bind(PlatformImpl.accessibilityActiveProperty());
@@ -445,12 +449,15 @@ public final class Platform {
      * by JavaFX when the operating system reports that a platform preference has changed.
      *
      * @return the {@code Preferences} instance
+     * @throws IllegalStateException if this method is called on a thread
+     *     other than the JavaFX Application Thread.
      * @see <a href="Platform.Preferences.html#preferences-table-windows">Windows preferences</a>
      * @see <a href="Platform.Preferences.html#preferences-table-macos">macOS preferences</a>
      * @see <a href="Platform.Preferences.html#preferences-table-linux">Linux preferences</a>
      * @since 22
      */
     public static Preferences getPreferences() {
+        Toolkit.getToolkit().checkFxUserThread();
         PlatformImpl.checkPreferencesSupport();
         return PlatformImpl.getPlatformPreferences();
     }
--- a/modules/javafx.graphics/src/main/java/javafx/application/Platform.java
+++ b/modules/javafx.graphics/src/main/java/javafx/application/Platform.java
@@ -272,6 +272,13 @@ public final class Platform {
     }
 
     /**
+     * The maximum number of nested event loops.
+     *
+     * @since 25
+     */
+    public static final int MAX_NESTED_EVENT_LOOPS = 240;
+
+    /**
      * Enter a nested event loop and block until the corresponding
      * exitNestedEventLoop call is made.
      * The key passed into this method is used to
@@ -301,6 +308,9 @@ public final class Platform {
      *
      * @throws IllegalStateException if this method is called on a thread
      * other than the JavaFX Application Thread.
+     *
+     * @throws RuntimeException if this call would exceed the maximum
+     * number of nested event loops.
      *
      * @return the value passed into the corresponding call to exitEventLoop
      *
--- a/modules/javafx.graphics/src/main/java/javafx/application/Platform.java
+++ b/modules/javafx.graphics/src/main/java/javafx/application/Platform.java
@@ -272,13 +272,6 @@ public final class Platform {
     }
 
     /**
-     * The maximum number of nested event loops.
-     *
-     * @since 25
-     */
-    public static final int MAX_NESTED_EVENT_LOOPS = 200;
-
-    /**
      * Enter a nested event loop and block until the corresponding
      * exitNestedEventLoop call is made.
      * The key passed into this method is used to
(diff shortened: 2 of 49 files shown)


modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
```diff
@@ -753,6 +754,11 @@ - (jobject)getPlatformPreferences
     return platformSupport != nil ? [platformSupport collectPreferences] : nil;
 }
 
++ (BOOL)canStartNestedEventLoop
+{
+    return nestedRunLoopRunCount <= 250;
```
kevinrushforth, line 759: How sure are you that 250 is a safe limit? Are there situations where we might not be able to do this many?

I recommend added a comment as to where the magic number "250" comes from. And maybe put it in a static const variable?

kevinrushforth (commented):
If the hard-coded 250 is a robust limit, then the fix looks reasonable to me. I'll test it and finish the review next week.

modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
andy-goryachev-oracle, line 759: would it be possible to determine the source of this limit - a header perhaps?
could it be dependent on os version?

modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
```diff
@@ -762,13 +768,22 @@ + (jobject)enterNestedEventLoopWithEnv:(JNIEnv*)env
...
-        if (event != nil) {
-            [app sendEvent: event];
+    while (shouldKeepRunningNestedLoop) {
+        // The app will crash if we exceed an undocumented limit on the number of nested
+        // CFRunLoopRun calls.
+        nestedRunLoopRunCount += 1;
+        BOOL ran = [theRL runMode:NSDefaultRunLoopMode
+                       beforeDate:[NSDate dateWithTimeIntervalSinceNow:0.010]];
+        nestedRunLoopRunCount -= 1;
+        if (ran) {
```
andy-goryachev-oracle, line 778: I could be wrong reading this code, but doesn't the `nestedRunLoopRunCount` value remain unchanged?  L774 gets always cancelled by L777?

andy-goryachev-oracle: never mind, it's entering and exiting the loop there.

jaybhaskar:
It is better to use dynamically calculated safe limit instead of a hardcoded value like 250. The dynamic calculation should be based on system resource and and 250 would be fallback value. like 
static int getSafeRunLoopLimit() {
    struct rlimit limit;
    if (getrlimit(RLIMIT_STACK, &limit) == 0) {
        return (int)(limit.rlim_cur / (512 * 1024)); 
    }
    return 250; // Fallback to 250 if stack size retrieval fails
}
macOS does not document an exact limit, nested CFRunLoopRun calls can cause stack growth. 512 KB is a safe approximation.

modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
beldenfox: I have not been able to find any header detailing this limit. My guess is that someone at Apple decided to add an internal check for infinite recursion to ensure the app fails early with a clear message and decided 255 levels of nesting should be enough for anyone. The crash log contains these lines (which also show up in the debugger):

> Application Specific Information:
> Too many nested CFRunLoopRuns

The check is new with macOS 15. I tested some standalone code on macOS 13 and 14 and had no problem reaching a nesting level of 500. But I don't think we should make the JavaFX check conditional on OS version number; I wouldn't want a JavaFX app that worked on macOS 14 to fail mysteriously on macOS 15.

beldenfox:
> It is better to use dynamically calculated safe limit instead of a hardcoded value like 250.

The problem doesn't seem to be resource exhaustion. It seems that Apple is using a hard-coded number specifically to detect a potential case of infinite CFRunLoop recursion and fail early with a clear message rather than let the app run wild. This doesn't seem to be tied to any limit I can check dynamically.

modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
andy-goryachev-oracle: I second Kevin's suggestion to add a constant with a good comment, since it's an arbitrarily chosen limit.  It would explain the "why" if/when the things change in macOS.

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
```diff
@@ -633,7 +633,11 @@ void vsyncHint() {
         }
 
         if (!canStartNestedEventLoop()) {
-            throw new IllegalStateException("Cannot enter nested loop during animation or layout processing");
+            if (!Application.GetApplication().canStartNestedEventLoop()) {
+                throw new RuntimeException("Exceeded limit on nested event loops");
```
andy-goryachev-oracle, line 637: should we tell what the limit is (using the static constant mentioned earlier)?
e.g.
"Too many nested event loops (250)"? or something like that.

modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
beldenfox: I added a comment explaining where the number comes from. Since it's only used in this one spot putting it in a static const variable seemed redundant.

I'm not sure that 250 is a safe limit. I know that the JUnit test crashes if this value is set to 252 but succeeds at 251. I don't want to risk disturbing existing apps but since each count represents a runnable waiting for an event loop to exit even 200 seems like a high number. It might be prudent to set this to, say, 246 to give us a bit more headroom.

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
beldenfox: The number is in the Mac-only Glass code and may change over time so it would be a maintenance burden to repeat it here. But if a developer does see this exception I'm sure their first question would be what the limit is. I can update the comment with something vague (like "over > 240") but I'm hoping that the very, very long Java stack trace attached to this exception will be enough.

modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
kevinrushforth: I agree that it might be wise to set it a bit lower, so either 246 or even 240 seems reasonable.

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
kevinrushforth: Unless you are going to track the actual depth, I'd probably leave it as is. When Java throws a StackOverflow it doesn't print the depth.

modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
beldenfox: Set to 244 so I didn't have to alter the system test.

beldenfox:
At the risk of opening a can of worms I ran some manual tests on Linux and Windows. Windows starts behaving erratically when the nesting level goes beyond 280. I've occasionally seen a Java.lang.stackOverflowError but most often it freezes up and then the next click causes it to crash with an access violation. Linux behaves well until it gets beyond 600 and then also starts to freeze up (doesn't seem to crash). I haven't dug into what's going on under the hood.

kevinrushforth (commented):
The fix and test look good. My local testing shows that it works as expected. I confirm that on macOS 15 the new test crashes before your fix and passes with the fix.

Just to be on the safe side, I fired off a headful CI test run. I'll approve once it passes.

kevinrushforth:
> At the risk of opening a can of worms I ran some manual tests on Linux and Windows. Windows starts behaving erratically when the nesting level goes beyond 280. I've occasionally seen a Java.lang.stackOverflowError but most often it freezes up and then the next click causes it to crash with an access violation. Linux behaves well until it gets beyond 600 and then also starts to freeze up (doesn't seem to crash). I haven't dug into what's going on under the hood.

We might want to consider imposing a limit on Linux and Windows, too. It could probably be done as a follow-up...

kevinrushforth:
> We might want to consider imposing a limit on Linux and Windows, too. It could probably be done as a follow-up...

Although doing it as part of this PR would be OK, too. Which would you prefer?

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
andy-goryachev-oracle: what I meant is the exception should indicate the limit that triggered the exception.
because it will likely be the first question the application developer asks when seeing this exception.

andy-goryachev-oracle:
I still not sure why anyone would need to create that many (>200) nested event loops.  So +1 for checking that on win/linux as well, possibly using the same constant (I would like to see a constant, maybe `REASONABLE_NUMBER_OF_NESTED_EVENT_LOOPS` )

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
mstr2: The limit is an implementation detail, and should not be elevated to specification. What application developers should be doing if they want to avoid hitting the limit, is call `Platform.canStartNestedEventLoop()`.

kevinrushforth:
If we go with the same limit on each platform, which does have some advantages, we might consider documenting that limit in the spec.

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
kevinrushforth: This is more like the StackOverflow case where the nesting level has gone too deep. It wouldn't hurt tell the application how deep they are (and from that they could infer what the limit is for the platform), but I'm not sure how useful it would be to the application developer. I don't mind one way or the other.

If we do decide to have a platform-independent nesting level, we would specify that limit. I'm still not sure whether we want to do this, but as I mentioned in another comment, there are some advantages to it.

kevinrushforth:
I think we will eventually settle on one of two solutions. In both cases we will want to document the concept of a limited nesting depth and encourage application developers to check `Platform.canStartNestedEventLoop` if they want to be sure that the platform can start one.

The two options are:

1. Make the limit platform-specific limit. The API docs will therefore not mention any particular limit. Each glass platform would implement `Application.canStartNestedEventLoop` imposing the limit for that platform. We _might_ consider documenting a guaranteed lower limit that is well below any specific platform limit (e.g., 150).
2. Make the limit platform-independent. We could, and probably should, choose to document this limit. We could choose to implement Application.canStartNestedEventLoop in shared code rather than overriding it in the pipeline. We would need to pick a safe "lowest common denominator" (maybe 200).

My gut feel is that we should go with option 1 and keep this platform-specific, which means we would not specify the limit (other than maybe giving a guarantee of "at least 150"), unless there are good arguments in favor of option 2 that we haven't considered.

Comments?

kevinrushforth:
> A centralized solution will probably lose that distinction but I'm not sure that's important. Exceeding 240+ nested events loops is problematic either way.

Agreed.

andy-goryachev-oracle:
I am leaning toward option 2, because I think having too many nested loop in an application is a design problem.  In a well-behaved application the number should never exceed some relatively small number so having a universal and relatively high number (~200) would be a good safeguard.

beldenfox:
For customers creating cross-platform apps having the limit be the same on all platforms would simplify testing. If this was per-platform we would need to publish the limits since developers would want to test on the most restrictive platform. And my guess is that we would set Mac and Windows to be close to the same (around 240) and Linux would be the outlier. So I'm leaning toward a cross-platform limit.

With that said documenting a cross-platform limit is awkward. Why is there a limit and why is it 240? There's one reason on the Mac, a different one on Windows, and Linux is just following the lead of the other two. In the cross-platform version of this PR I included a public constant but didn't even try to explain where the number came from. I'm tempted to make this non-public but have it show up in the exception description.

andy-goryachev-oracle:
Related mailing list discussion:
https://mail.openjdk.org/pipermail/openjfx-dev/2025-March/053011.html

kevinrushforth:
I can see the benefit of a cross-platform limit. And a limit is 240, or even 200, is way more than should be needed for a reasonable application.

mstr2:
> For customers creating cross-platform apps having the limit be the same on all platforms would simplify testing. If this was per-platform we would need to publish the limits since developers would want to test on the most restrictive platform.

Why would an application need to be tested against a specific limit? It seems the only thing that would be tested in this case is whether JavaFX works as advertised. An application that can conceivably start a nested event loop should always check whether it can start the event loop in the first place, and fail gracefully if it can't.

I can see some value in guaranteeing a lower limit for the number of nested event loops, which might be useful for applications that provably don't start more than a certain number of nested event loops.

mstr2:
In any case, this PR should at least add the following bits of information to `Platform.enterNestedEventLoop()`:
1. There is some limit to the number of nested event loops.
2. Applications should always check `Platform.canStartNestedEventLoop()` before invoking `enterNestedEventLoop()`.
3. An exception will be thrown if the limit is exceeded.

beldenfox:
> Why would an application need to be tested against a specific limit?

If you knew that your app could hit a limit and the limit could vary by platform you would want to focus your testing on the platform with the lowest limit. In any case it's always annoying when a program works on one platform and fails on another. Better to have it fail the same way everywhere if possible (and in this case we can make that happen).

> An application that can conceivably start a nested event loop should always check whether it can start the event loop in the first place, and fail gracefully if it can't.

This would require adding checks before any calls that start a nested event loop. In the test case submitted by the customer the check would have to occur before a call to `Stage.showAndWait`. On the Mac a nested event loop is entered whenever the fullscreen state changes. I don't think there's any good way for a developer to know which JavaFX calls might enter a nested event loop.

As far as we know this crash is triggered when the app enters an unexpected state of runaway recursion. I think the goal is to keep the system from crashing long enough to get a Java stack trace to understand how the system got into that state. In the case submitted by the customer the stack trace suggests that JavaFX is throwing an exception that it should not be throwing (I was able to reproduce the recursive state but am still sorting through the rubble).

modules/javafx.graphics/src/main/java/javafx/application/Platform.java
```diff
@@ -271,6 +271,13 @@ public static boolean isSupported(ConditionalFeature feature) {
         return PlatformImpl.isSupported(feature);
     }
 
+    /**
+     * The maximum number of nested event loops.
+     *
+     * @since 25
+     */
+    public static final int MAX_NESTED_EVENT_LOOPS = 200;
```
andy-goryachev-oracle, line 279: I would rather not make this constant public.  This is essentially a safeguard measure, with the main purpose to prevent the crash and give the indication to the application developer that their program has an issue.  From this perspective, the actual number is not important and should not be codified as a public API (but definitely declared as a constant internally).

What do you think?

beldenfox: I don't think it needs to be public. If we privatize this constant I could use some guidance on where to place it so that the system test can access it. Currently there are two system tests that exercise nested event loops (PlatformTest.java and NestedEventLoopTest.java) and they only rely on public API's.

andy-goryachev-oracle: `PlatformImpl` is probably a good choice, and it's already configured to be accessible to the system tests via `tests/system/src/test/addExports`

andy-goryachev-oracle (approved):
Looks good!  The exception message should be clear enough for a developer who is encountering it for the first time.

kevinrushforth:
This needs a second reviewer. It will likely need a CSR since there is a new exception being documented and throws.

modules/javafx.graphics/src/main/java/javafx/application/Platform.java
```diff
@@ -302,6 +302,9 @@ public static boolean isSupported(ConditionalFeature feature) {
      * @throws IllegalStateException if this method is called on a thread
      * other than the JavaFX Application Thread.
      *
+     * @throws RuntimeException if this call would exceed the maximum
```
kevinrushforth, line 305: I wonder if there is a more specific subclass of `RuntimeException` that we could throw? Maybe `IllegalStateException`, which seems the closest in meaning and has the advantage that we already throw it in other cases.

kevinrushforth (commented):
The implementation looks good.

I left one question / suggestion on the type of Exception throws.

I think the suggestion made by @mstr2 to recommend calling `canStartNestedLoop` if the app wants to know whether or not this call will succeed is a good one.

I think we should update the docs for `Stage::showAndWait` documenting the same exception, since the implementation of that method also spins up a nested event loop.

/csr

kevinrushforth:
Since this fix has been expanded to all platforms, let's remove `[macos]` from the title of the JBS issue and PR.

modules/javafx.graphics/src/main/java/javafx/application/Platform.java
beldenfox: I'll have to wait for others to chime in on that question, I don't have much background in this area.

andy-goryachev-oracle: The `IllegalStateException` is the right one, as it states that it _"Signals that a method has been invoked at an illegal or inappropriate time."_

Another positive thing it won't require CSR as this method already throws one.

andy-goryachev-oracle (approved):
Do we still need a CSR @kevinrushforth ?

beldenfox:
> I think the suggestion made by @mstr2 to recommend calling `canStartNestedLoop` if the app wants to know whether or not this call will succeed is a good one.

The docs on `canStartNestedEventLoop` describe its behavior already. I'm not sure what further information would be useful. Would it be enough to add a see also link?

mstr2:
> The docs on `canStartNestedEventLoop` describe its behavior already. I'm not sure what further information would be useful. Would it be enough to add a see also link?

The exception being thrown is a behavior of `enterNestedEventLoop()`, and should therefore be documented with this method (`@see` is not enough). While that is done now with the `@throws` javadoc tag, I think it wouldn't hurt to be more explicit about that; much in the same way as the javadoc summary already contains an explicit description of the preconditions of calling the method.

kevinrushforth:
> Do we still need a CSR @kevinrushforth ?

It's still a good idea, since we are throwing the exception in cases where we didn't use to. It will be a simple CSR with a risk level of "minimal" (since it's in a case where misbehavior is going to happen anyway).

kevinrushforth:
> > The docs on `canStartNestedEventLoop` describe its behavior already. I'm not sure what further information would be useful. Would it be enough to add a see also link?
> 
> The exception being thrown is a behavior of `enterNestedEventLoop()`, and should therefore be documented with this method (`@see` is not enough). While that is done now with the `@throws` javadoc tag, I think it wouldn't hurt to be more explicit about that; much in the same way as the javadoc summary already contains an explicit description of the preconditions of calling the method.

How about adding something like this, right after the other preconditions?

```
* There is a finite limit on the depth of the nested event loop stack. An exception will be thrown
*  if this limit is exceeded. Applications that want to avoid an exception can call
* {@link #canStartNestedEventLoop canStartNestedEventLoop} to see whether it is possible
* to start one.
```

modules/javafx.graphics/src/main/java/javafx/application/Platform.java
```diff
@@ -288,6 +288,12 @@ public static boolean isSupported(ConditionalFeature feature) {
      * It must not be called during animation or layout processing.
      * </p>
      *
+     * There is a finite limit on the depth of the nested event loop stack. An
+     * exception will be thrown if this limit is exceeded. Applications that
+     * want to avoid an exception can call
+     * {@link #canStartNestedEventLoop canStartNestedEventLoop} to check
+     * whether it is possible to start one.
+     *
```
andy-goryachev-oracle, line 296: +1

andy-goryachev-oracle (approved):

kevinrushforth (commented):
Latest version looks good. I've fired off a CI headful test build and will report back once it completes and I finish my testing.

kevinrushforth (approved):
All good.

kevinrushforth:
@beldenfox I also reviewed the CSR and it is ready to be Finalized.

PR #1744, opened 2025-03-26
8314482: TextFlow: TabStopPolicy

andy-goryachev-oracle:
TBD


modules/javafx.graphics/src/main/java/javafx/scene/text/TextFlow.java
```diff
@@ -543,6 +565,96 @@ public final void setTabSize(int spaces) {
         tabSizeProperty().set(spaces);
     }
 
+    /**
+     * {@code TabAdvancePolicy} determines the tab stop positions within this {@code TextFlow}.
```
kevinrushforth: `TabAdvancePolicy` --> `TabStopPolicy`

modules/javafx.graphics/src/main/java/javafx/scene/text/TextFlow.java
```diff
@@ -543,6 +565,96 @@ public final void setTabSize(int spaces) {
...
+                }
+
+                @Override
+                public Object getBean() {
+                    return TextFlow.this;
+                }
+
+                @Override
+                public String getName() {
+                    return "tabAdvancePolicy";
```
kevinrushforth: `"tabStopPolicy"`

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStop.java
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * @since 999 TODO
+ */
+public class TabStop {
+    private final double position;
+
+    /**
+     * Constructs a new tab stop with the specified position.
+     *
+     * @param position the position in pixels
```
kevinrushforth, line 38: Is pixels the right unit here? Should it be points instead? Or should it allow for the specification of units so that you could specify it in "em"s?

@prrace might want to weigh in on this.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStop.java
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * This class encapsulates an immutable single tab stop within the {@link TabStopPolicy}.
+ * <p>
+ * A tab stop is at a specified distance from the
+ * left margin, aligns text in a specified way, and has a specified leader.
+ *
+ * @since 999 TODO
+ */
+public class TabStop {
```
kevinrushforth: Should this be a record? If not, it should at least be a final class.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
```diff
@@ -0,0 +1,101 @@
...
+import javafx.collections.FXCollections;
+import javafx.collections.ObservableList;
+import javafx.scene.layout.Region;
+
+/**
+ * The TabStopPolicy determines the tab stop positions within the text layout.
+ *
+ * @since 999 TODO
+ */
+public class TabStopPolicy {
```
kevinrushforth: Is there a reason this class is not final? It probably should be.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * Constructs a new {@code TabStopPolicy} instance.
+     */
+    public TabStopPolicy() {
+    }
+
+    /**
+     * Specifies the unmodifiable list of tab stops, sorted by position from smallest to largest.
+     * The list can be changed using
```
kevinrushforth: "changed using" ... what?

More importantly, this isn't an unmodifiable list, right? (it isn't implemented that way, and it would not be very useful if it were) So you can't enforce that it is sorted. Rather this should just be an ordinary observable list, like we have in other classes, and if you need a sorted view of it, then have one internally that keeps it sorted.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
```diff
@@ -0,0 +1,101 @@
...
+     *
+     * @return the non-null, unmodifiable list of tab stops, sorted by position
+     */
+    public final ObservableList<TabStop> tabStops() {
+        return tabStops;
+    }
+
+    /**
+     * Provides default tab stops (beyond the last tab stop specified by {@code #tabStops()},
```
kevinrushforth: Having this be a plural when it is a single value (a double) is awkward. I would name the property `defaultStop` and document that the default is used for all stops beyond the last stop.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStop.java
andy-goryachev-oracle: Using the same units used everywhere else (text layout, padding, border, etc.)
I think any conversion should be the responsibility of the application code.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStop.java
andy-goryachev-oracle: No and no.  The application code or the custom component should be able to extend the TabStop class to supply additional options.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
andy-goryachev-oracle: good point, changed to `final`.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
andy-goryachev-oracle: This terminology is borrowed from a well known software product.  I think it is apt as it describes the spacing (and the positions) of multiple tab stops, and also the users might be familiar with the term.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStop.java
kevinrushforth: How would such extensibility work? What would the implementation do with additional options? Or is the intention that something other than TextFlow (e.g., RichTextArea or some third-party component would use that information, but TextFlow would ignore it? If so, that seems OK.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStop.java
kevinrushforth: That seems fine then, as long as this has been thought through.

A loosely related question: what about setting the tab stop policy via CSS? I can see apps that would want to do this. It could be added later if you don't see it as a primary use case.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
kevinrushforth: Hmm. OK then.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * @return the non-null, list of tab stops
+     */
+    public final ObservableList<TabStop> tabStops() {
+        return tabStops;
+    }
+
+    /**
+     * Provides default tab stops (beyond the last tab stop specified by {@code #tabStops()},
+     * as a fixed repeating distance in pixels from the last tab stop position.
```
kevinrushforth: This is a little misleading. The stops are applied after the last tap stop position, but the distance is not fixed "from" that point. The next sentence explains this, but since the first sentence shows up in the method summary, I might reword this a bit to remove the possibility of this misunderstanding.

How about: "as a fixed repeating distance in pixels for tabs after the last tab stop position." or similar?

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStop.java
```diff
@@ -0,0 +1,71 @@
...
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package javafx.scene.text;
+
+/**
+ * This class encapsulates an immutable single tab stop within the {@link TabStopPolicy}.
+ * <p>
+ * A tab stop is at a specified distance from the
+ * left margin, aligns text in a specified way, and has a specified leader.
```
kevinrushforth: Where is the alignment and leader specified?

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
```diff
@@ -0,0 +1,100 @@
...
+     * The position of default tab stops is computed at regular intervals relative to
+     * the leading edge of the {@code TextFlow} this policy is registered with.
+     * <p>
+     * The value of less than or equal 0 disables the default stops.
```
kevinrushforth: Minor: "A value less than or equal ..."

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
```diff
@@ -0,0 +1,100 @@
...
+ * The TabStopPolicy determines the tab stop positions within the text layout.
+ *
+ * @since 999 TODO
+ */
+public final class TabStopPolicy {
+    private final ObservableList<TabStop> tabStops = FXCollections.observableArrayList();
+    private final SimpleDoubleProperty defaultStops = new SimpleDoubleProperty(0.0);
+
+    /**
+     * Constructs a new {@code TabStopPolicy} instance.
```
kevinrushforth: Minor: Maybe add "with an empty list of stops"?

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
```diff
@@ -0,0 +1,100 @@
...
+     */
+    public TabStopPolicy() {
+    }
+
+    /**
+     * The list of tab stops.
+     *
+     * @return the non-null, list of tab stops
+     */
```
kevinrushforth, line 53: Minor: remove the comma.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
```diff
@@ -0,0 +1,100 @@
...
+    /**
+     * Provides default tab stops (beyond the last tab stop specified by {@code #tabStops()},
+     * as a fixed repeating distance in pixels from the last tab stop position.
+     * The position of default tab stops is computed at regular intervals relative to
+     * the leading edge of the {@code TextFlow} this policy is registered with.
+     * <p>
+     * The value of less than or equal 0 disables the default stops.
+     *
+     * @return the default tab stops property
+     * @defaultValue 0
```
kevinrushforth, line 67: Would "8" be a better default value? I think having tabs do something (instead of effectively being ignored) by default when you create a TabStopPolicy would be less surprising to app developers.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStop.java
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * @since 999 TODO
```
kevinrushforth: Change this to either 25 (if you'd like to shoot for getting it into 25) or 26.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
mstr2: This doesn't read like a proper English sentence to me. In Microsoft Word, there's a spinner control titled "Default tab stops:", which I suppose should be read as, for example: "Default tab stops every 0.5 cm".

As for naming the API, I suggest something like `defaultTabWidth` or `defaultWidth` (the word "tab" doesn't need to be repeated, as it's declared in a class named `TabStopPolicy`).

andy-goryachev-oracle: Not sure if defaultTabWidth is a good choice here though, because it's not, it's actually tab stops - think of the typewriter (I know, referring to an ancient technology like floppy disk for the 'save' icon does not age well)

I am open to suggestions.  I am used to "default tab stops", perhaps we could just explain it better?

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStop.java
andy-goryachev-oracle: oops, a leftover from an earlier version.  there is no support currently for the alignment or the leader.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
andy-goryachev-oracle: No, this value is in pixels (unlike the `tabSize` property).  I think it's better to let the application specify it explicitly.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStop.java
andy-goryachev-oracle: I hope to get it to 25, but we'll see.
Will change once we know more.

andy-goryachev-oracle:
> question: what about setting the tab stop policy via CSS?

A very good question!

It might be entirely possible (in a separate PR), though judging by experience from adding tab stops to the rich editor demo https://github.com/openjdk/jfx/pull/1800 it's probably unnecessary, since TextFlow is somewhat of a low-level control.

Another complication is that the policy is a complex entity, with a list of (possibly complex) `TabStop`s, and other fields, so I am not sure how easy it would be to add CSS support.  How would it look in CSS?

andy-goryachev-oracle:
Also, decided to make the `TabStop` class `final`.  We can always remove the `final` keyword if the use case arises.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
kevinrushforth: Getting back to this, there are two naming questions here. First, should "tab" be in the name (it does seem a bit redundant, since the name of the class is TabStopPolicy)? Second, what is the best name for the value that defines a repeating set of stops?

If we drop the word "tab" the list of tab stops would just be named "stops", which seems fine. Here are some possible names for the default:

- defaultStop
- defaultStops
- defaultStopInterval
- Something Else?

I kind of like the last of these, `defaultStopInterval`, as it is the most descriptive. Whatever we do will need to be documented well.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
kevinrushforth: OK, this seems like a good reason to leave it.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
andy-goryachev-oracle: ooh, I like `defaultStopInterval`, thank you.

mstr2: Maybe make it even shorter and do `TabStopPolicy.defaultInterval`. The reasoning for dropping the "stop" is the same as before, it's already the class name.

andy-goryachev-oracle: even better!

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
```diff
@@ -0,0 +1,99 @@
...
+     */
+    public TabStopPolicy() {
+    }
+
+    /**
+     * The list of tab stops.
+     *
+     * @return the non-null list of tab stops
+     */
+    public final ObservableList<TabStop> tabStops() {
```
kevinrushforth, line 54: Did you consider also dropping "tab" from the name here? Either is fine with me, so I'll leave it up to you.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
```diff
@@ -0,0 +1,99 @@
...
+        return tabStops;
+    }
+
+    /**
+     * Specifies the default tab stop interval (beyond the last tab stop provided by {@code #tabStops()}),
+     * as a fixed repeating distance (in pixels) to the next tab stop computed at regular intervals
+     * relative to the leading edge of the {@code TextFlow} node.
```
kevinrushforth: This is a bit of a run-on sentence. Maybe consider breaking it up into two sentences, with the detail about "a fixed, repeating distance..." in the second sentence? Something like:

```
    * Specifies the default tab stop interval for tabs beyond the last stop provided
    * by {@code #tabStops()}. This is a fixed repeating distance (in pixels) to the
    * next tab stop computed at regular intervals relative to the leading edge
    * of the {@code TextFlow} node.
```

modules/javafx.graphics/src/main/java/javafx/scene/text/TextFlow.java
```diff
@@ -509,6 +525,10 @@ public final DoubleProperty lineSpacingProperty() {
      * The size of a tab stop in spaces.
      * Values less than 1 are treated as 1. This value overrides the
      * {@code tabSize} of contained {@link Text} nodes.
+     * <p>
+     * Note that this method should not be used to control the tab placement when multiple {@code Text} nodes
+     * with different fonts are contained within this {@code TextFlow}.
+     * In this case, the {@link #setTabStopPolicy(TabStopPolicy)} should be used instead.
```
kevinrushforth: either drop the word "the" before "setTabStopPolicy" or add "method" after.

modules/javafx.graphics/src/main/java/javafx/scene/text/TextFlow.java
```diff
@@ -543,6 +565,96 @@ public final void setTabSize(int spaces) {
         tabSizeProperty().set(spaces);
     }
 
+    /**
+     * {@code TabStopPolicy} determines the tab stop positions within this {@code TextFlow}.
```
kevinrushforth: I think you can remove `{@code TabStopPolicy}` since that's what this property is. Start with "Determines the ..."

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
```diff
@@ -56,26 +56,25 @@ public final ObservableList<TabStop> tabStops() {
...
-     * A value of less than or equal 0 disables the default stops.
+     * A value of less than or equal 0 disables the default interval.
      *
-     * @return the default tab stops property
+     * @return the default tab interval property
      * @defaultValue 0
      */
-    public final DoubleProperty defaultStopsProperty() {
-        return defaultStops;
+    public final DoubleProperty defaultInterval() {
```
kevinrushforth: Oops. This should be `defaultIntervalProperty()`, which explains the GHA build failure.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
andy-goryachev-oracle: I prefer `tabStops`.

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
andy-goryachev-oracle: much better, thank you!

modules/javafx.graphics/src/main/java/javafx/scene/text/TabStopPolicy.java
```diff
@@ -56,16 +56,17 @@ public final ObservableList<TabStop> tabStops() {
     }
 
     /**
-     * Specifies the default tab stop interval (beyond the last tab stop provided by {@code #tabStops()}),
-     * as a fixed repeating distance (in pixels) to the next tab stop computed at regular intervals
-     * relative to the leading edge of the {@code TextFlow} node.
+     * Specifies the default tab stop interval for tabs beyond the last stop provided
+     * by {@code #tabStops()}.  This is a fixed repeating distance (in pixels) to the
```
kevinrushforth: Change this to a `@link` or remove the `#` and `()` since they show up in the docs.

kevinrushforth (commented):
API / docs looks good with one formatting issue that I noticed when looking at the generated API docs.

I'll review the CSR next.

andy-goryachev-oracle:
created the CSR.

andy-goryachev-oracle:
Good point!

Added comprehensive test, thanks to the latest changes that made the StubTextLayout work essentially like the real one, and the new LayoutInfo API.

andy-goryachev-oracle:
@arapte could you please take a look at this PR?

arapte (approved):
LGTM+

andy-goryachev-oracle:
thank you all

/integrate

PR #1745, opened 2025-03-28
8353163: [XWayland] Disable SwingNodePlatformExitCrashTest on wayland

jayathirthrao:
SwingNodePlatformExitCrashTest hangs while running on wayland, we need to skip this test on wayland until https://bugs.openjdk.org/browse/JDK-8350009 is fixed.

---------
### Progress

### Issue
 * [JDK-8353163](https://bugs.openjdk.org/browse/JDK-8353163): [XWayland] Disable SwingNodePlatformExitCrashTest on wayland (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1745/head:pull/1745` \
`$ git checkout pull/1745`

`$ git checkout pull/1745` \
`$ git pull https://git.openjdk.org/jfx.git pull/1745/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1745`

`$ git pr show -t 1745`

</details>

<a href="https://git.openjdk.org/jfx/pull/1745.diff">https://git.openjdk.org/jfx/pull/1745.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1745#issuecomment-2760609614)
</details>

### DIFF
--- a/tests/system/src/test/java/test/robot/javafx/embed/swing/SwingNodePlatformExitCrashTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/embed/swing/SwingNodePlatformExitCrashTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -31,6 +31,7 @@ import javax.swing.SwingUtilities;
 import javafx.application.Platform;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Assumptions;
 import test.util.Util;
 
 public class SwingNodePlatformExitCrashTest extends SwingNodeBase {
@@ -53,6 +54,7 @@ public class SwingNodePlatformExitCrashTest extends SwingNodeBase {
 
     @BeforeAll
     public static void skipShutDown() {
+        Assumptions.assumeTrue(!Util.isOnWayland());
         // Skip shutdown as Toolkit shutdown is already done in Platform.exit
         // and will cause hang if called
         doShutdown = false;


andy-goryachev-oracle (approved):

andy-goryachev-oracle:
Gradle Wrapper Validation step failed - could you re-start the job?  I suspect github does not like mornings.

tests/system/src/test/java/test/robot/javafx/embed/swing/SwingNodePlatformExitCrashTest.java
```diff
@@ -53,6 +54,7 @@ public void testPlatformExitBeforeShowHoldEDT() throws InvocationTargetException
 
     @BeforeAll
     public static void skipShutDown() {
+        Assumptions.assumeTrue(!Util.isOnWayland());
```
kevinrushforth, line 57: Please add `// JDK-8350009` in a comment, either at the end of this line or on a line by itself before this).

kevinrushforth (commented):
LGTM with one inline suggestion about adding the bug ID being used to skip this test (JDK-8350009).

Regarding [JDK-8350009](https://bugs.openjdk.org/browse/JDK-8350009), I was finally able to test this on my Ubuntu 24.04 VM after resetting the Wayland screencast token (I had some problem with my system earlier). After doing this, the test only hangs for me on JDK 23. When I run it on JDK 24 it now reliably passes for me.

More testing is needed, though, so go ahead and integrate this fix to skip the test on Wayland after adding the bug ID. If we then find out that it runs reliably on multiple systems when using JDK 24, you can fix [JDK-8350009](https://bugs.openjdk.org/browse/JDK-8350009) by adding the check for JDK version so that it will only skip on Wayland for JDK 23 and earllier.

johanvos:
I'm a bit worried with disabling the test, as there is no guarantee the bug will be fixed. Since the issue is clearly identified, I would rather go for fixing JDK-8350009 instead of disabling the test.

kevinrushforth:
The hang we are seeing on Wayland Linux when running on some systems is due to [JDK-8335468](https://bugs.openjdk.org/browse/JDK-8335468), which is fixed in AWT in JDK 24. So 

So the solution to [JDK-8350009](https://bugs.openjdk.org/browse/JDK-8350009), which is a test bug, will almost certainly be to disable the test on Wayland when running the test on JDK 23 (or earlier). Something like this, modifying the currently proposed Assumption so that it only skipped when running an older JDK:

```
        Assumptions.assumeTrue(!Util.isOnWayland() || Runtime.version().feature() >= 24);
```

Maybe it is better, then, to just fix  [JDK-8350009](https://bugs.openjdk.org/browse/JDK-8350009) with the revised Assumption as above.

PR #1746, opened 2025-03-28
8328716: [TestBug] Screen capturing utility for failed tests

andy-goryachev-oracle:
Introduce a facility, in the form of JUnit5 annotation, to allow for capturing a desktop screenshot of a failed test.

The primary intent is to be able to debug an intermittent test case, rather than wholesale addition of the new annotation to all the tests.

A possible change is to output a data URL in the form of
`data:image/png;base64,iVBORw0KGgoAAAANSUhEU...`
so it can be rendered in Safari (Chrome truncates the image)


tests/system/src/test/java/test/util/ScreenCaptureTestWatcher.java
```diff
@@ -0,0 +1,114 @@
...
+ * <p>
+ * For Eclipse users, override dependencies with the following entries:
+ * <pre>{@code
+ * --add-modules=javafx.base,javafx.graphics,javafx.controls,javafx.swing
+ * --add-opens javafx.controls/test.com.sun.javafx.scene.control.test=javafx.base
+ * --add-exports javafx.base/com.sun.javafx=ALL-UNNAMED
+ * -Djava.library.path="../../../../build/sdk/lib"
+ * -ea
+ * }</pre>
+ * <p>
```
kevinrushforth, line 62: I don't much like IDE-specific documentation in our class files.

tests/system/src/test/java/test/util/ScreenCaptureTestWatcher.java
```diff
@@ -0,0 +1,114 @@
...
+ * WARNING: using this utility may significantly increase the size of your logs!
+ * Make sure there is plenty of free disk space.
+ */
+// TODO investigate having a hard-coded or programmable via property limit on the number of
+// captured screenshots.
```
kevinrushforth, line 67: I had the same thought. Given that each test runs in its own JVM, I can't think of a good way to do it other than recording information in a file in tests/system/build.

tests/system/src/test/java/test/util/ScreenCaptureTestWatcher.java
```diff
@@ -0,0 +1,114 @@
...
+public class ScreenCaptureTestWatcher implements TestWatcher {
+    @Override
+    public void testAborted(ExtensionContext extensionContext, Throwable err) {
+        err.printStackTrace();
```
kevinrushforth, line 71: Is this needed or is it just for debugging? We already get a stack trace when a test fails, so it might be redundant.

tests/system/src/test/java/test/util/ScreenCaptureTestWatcher.java
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    @Override
+    public void testFailed(ExtensionContext extensionContext, Throwable err) {
+        err.printStackTrace();
```
kevinrushforth, line 76: Same question as about about printing the stack trace.

tests/system/src/test/java/test/util/ScreenCaptureTestWatcher.java
```diff
@@ -0,0 +1,114 @@
...
+        System.err.println(generateScreenshot("Screenshot:{", "}"));
```
kevinrushforth, line 77: When does the `testFailed` method run? After a failing `@Test` method that throws the exception? What if there is more than one failing test? Ideally what we want is something that runs after all failing tests, but before the `@AfterEach` method.

kevinrushforth: Or... maybe we really do want to take a screen dump for each failing test in a test class.

Also, what happens if one of the lifecycle methods is the one that throws the error (e.g., `Before` `After`, etc)?

tests/system/src/test/java/test/robot/javafx/scene/control/behavior/TextAreaBehaviorRobotTest.java
```diff
@@ -29,12 +29,15 @@
...
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
 import com.sun.javafx.PlatformUtil;
+import test.util.ScreenCaptureTestWatcher;
 
 /**
  * Tests TextArea behavior by exercising every key binding not covered by TextAreaBehaviorTest,
  * since the mapped functions require rendered text.
  */
+@ExtendWith(ScreenCaptureTestWatcher.class)
```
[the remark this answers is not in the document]
andy-goryachev-oracle: I don't think so - it's a simple way to plug in into the junit5 framework.

tests/system/src/test/java/test/util/ScreenCaptureTestWatcher.java
```diff
@@ -0,0 +1,114 @@
...
+import javafx.stage.Screen;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.jupiter.api.extension.TestWatcher;
+
+/**
+ * Standard Test Watcher for Headful Tests (JUnit 5 Only).
+ * <p>
+ * This facility takes a screenshot of any failed test, then logs the base64-encoded screenshot
+ * to {@code stderr}.
```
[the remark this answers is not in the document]
andy-goryachev-oracle: The intent is not to annotate each test, but rather to use this tool to debug the issues in an intermittently failing test.

tests/system/src/test/java/test/util/ScreenCaptureTestWatcher.java
```diff
@@ -0,0 +1,114 @@
...
+ * <pre>{@code
+ * --add-modules=javafx.base,javafx.graphics,javafx.controls,javafx.swing
+ * --add-opens javafx.controls/test.com.sun.javafx.scene.control.test=javafx.base
+ * --add-exports javafx.base/com.sun.javafx=ALL-UNNAMED
+ * -Djava.library.path="../../../../build/sdk/lib"
+ * -ea
+ * }</pre>
+ * <p>
+ * WARNING: using this utility may significantly increase the size of your logs!
+ * Make sure there is plenty of free disk space.
```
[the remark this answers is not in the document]
andy-goryachev-oracle: see the earlier comment about the intermittent tests

tests/system/src/test/java/test/util/ScreenCaptureTestWatcher.java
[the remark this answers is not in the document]
kevinrushforth: Hmm. That isn't what I thought we were doing. I thought the idea was to annotate most (if not all all) screen capture tests, qualified by a system property, enable that property in our CI test runs, so that when we do get intermittent failures, we'll be able to take a look at them.

We _could_ to what you propose, but in that case I wouldn't want _any_ tests annotated as part of this or any other PR. Once we have this annotation on any test in our repo, then my objection about not having it on by default needs to be addressed, at which point you might as well annotate more tests, since we have seen occasional failures on many of them.

tests/system/src/test/java/test/util/ScreenCaptureTestWatcher.java
andy-goryachev-oracle: good question!  A quick test indicates that `@BeforeAll` and `@AfterAll` are treated as the framework errors and not handled by the watcher, but the failures in `@Test`, `@BeforeEach`, and `@AfterEach` are sent to the test watcher.

tests/system/src/test/java/test/util/ScreenCaptureTestWatcher.java
andy-goryachev-oracle: Exactly.  I don't have a good way, and I would rather not write to a file.

tests/system/src/test/java/test/util/ScreenCaptureTestWatcher.java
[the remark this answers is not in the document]
andy-goryachev-oracle: There is more than one way to sk^H^H pet the cat.

We could use a property to disable (or rather, enable) the screenshots, and only enable the capture during the debugging session.  This will prevent us from catching those hard-to-reproduce intermittent tests that fail only occasionally.

The other concern is that in the case of some infrastructure misconfiguration (for example, a sudden loss of screen capture permission) would result in log file size explosion, and I don't think we have `tailwatch`-like facility to halt the test run when this happens.

I think the debug-only annotation is a meaningful compromise, but I would very much welcome other ideas.

tests/system/src/test/java/test/util/ScreenCaptureTestWatcher.java
andy-goryachev-oracle: Good point - it's not needed.  The junit framework outputs a truncated stack trace which should be sufficient:

```
TextAreaBehaviorRobotTest > fail() FAILED
    java.lang.Error
        at test.robot.javafx.scene.control.behavior.TextAreaBehaviorRobotTest.fail(TextAreaBehaviorRobotTest.java:49)
```

kevinrushforth (commented):
I tested this by injecting a failure into one of the Robot screen capture tests, `RectangleTest` (see the patch I applied below):

While it did trigger the screen shot, the rendered window had already been closed before the screen shot was taken, meaning it didn't capture the screen at the point of failure. This suggests that the test watcher doesn't get called until after the `@AfterEach` method has run. Since a well-written test will cleanup after itself in the `@AfterEach` method, this will limit the usefulness of doing it via an annotation. Is there an option or some other annotation that will allow us to intercept at the point of failure?


```java
--- a/tests/system/src/test/java/test/robot/helloworld/RectangleTest.java
+++ b/tests/system/src/test/java/test/robot/helloworld/RectangleTest.java
@@ -34,10 +34,13 @@ import javafx.stage.Stage;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.Timeout;
 import test.robot.testharness.VisualTestBase;
+import org.junit.jupiter.api.extension.ExtendWith;
+import test.util.ScreenCaptureTestWatcher;

 /**
  * Basic visual tests using glass Robot to sample pixels.
  */
+@ExtendWith(ScreenCaptureTestWatcher.class)
 @Timeout(value=15000, unit=TimeUnit.MILLISECONDS)
 public class RectangleTest extends VisualTestBase {

@@ -79,7 +82,7 @@ public class RectangleTest extends VisualTestBase {
         waitFirstFrame();
         runAndWait(() -> {
             Color color = getColor(testScene, WIDTH / 2, HEIGHT / 2);
-            assertColorEquals(Color.CORNFLOWERBLUE, color, TOLERANCE);
+            assertColorEquals(Color.GREEN, color, TOLERANCE);
         });
     }

```

tests/system/src/test/java/test/util/ScreenCaptureTestWatcher.java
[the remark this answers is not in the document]
kevinrushforth: Yes, there are two main approaches we could take:

Option 1: A utility that is available for developers to add to one or more specific tests in a branch in their personal fork when debugging failures in those tests. We would not add it to any test in the mainline repo in this case. Since it is a developer-only utility, it isn't necessary to have a system property to enable it.

Option 2: A utility that is added to all robot tests in the repo, with a flag to enable it (off by default).

There are pros / cons of each approach. The second approach is more likely to catch a failing test that fails rarely and would easily allow us to enable it and run it on a new platform (e.g., for testing a new version of macOS or Ubuntu), although it is also more prone to the "runaway" screenshot problem unless we can come up with a good way to limit it.

I left a couple usability questions that are independent of this. It probably makes sense to address those first and then come back to this.

andy-goryachev-oracle:
> This suggests that the test watcher doesn't get called until after the @AfterEach method has run.

That's entirely possible.  I suppose one can register a Thread::uncaughtExceptionHandler and call a static method, something like ScreenCapture::takeScreenshot() (to be added).

tests/system/src/test/java/test/util/ScreenCaptureTestWatcher.java
```diff
@@ -22,50 +22,41 @@
...
  * Make sure there is plenty of free disk space.
  */
-public class ScreenshotFailedTestWatcher implements TestWatcher {
+public class ScreenCaptureTestWatcher implements TestWatcher {
     @Override
-    public void testFailed(ExtensionContext cx, Throwable e) {
-        System.err.println(generateScreenshot("Screenshot:\ndata:image/png;base64,", "\n"));
+    public void testFailed(ExtensionContext extensionContext, Throwable err) {
+        // can be pasted into Safari address bar
+        System.err.println(generateScreenshot("Screenshot:{\ndata:image/png;base64,", null));
```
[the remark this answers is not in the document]
andy-goryachev-oracle: yes, and you pointed to the problem exactly - not only it creates a difficult task of retrieving files from the remote system, require a [complex] cleanup schedule, as well as run a risk of losing the files after a while and impossibility of finding the right screenshot after the fact (sometimes much later).

at the same time, the suggested approach keeps the screenshot in the context, available for analysis at any point in time.  the only drawback is base-64 inflation of about 33-37% relative to the raw binary.

tests/system/src/test/java/test/util/ScreenCaptureTestWatcher.java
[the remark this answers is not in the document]
andy-goryachev-oracle: I would choose the Option 1 (debug utility), with an addition of a static `ScreenCapture.takeScreenshot()`.

I don't think there is a need to add this tool to the hundreds of tests which work just fine.

andy-goryachev-oracle:
Turns out one can't intercept assertion exceptions with `Thread.setDefaultUncaughtExceptionHandler()` (duh!), so this is likely to be a known limitation.

We might still provide general purpose methods in `ScreenCapture` (new class) like 

```java
    /**
     * Captures a screenshot using JavaFX {@link Robot} in the PNG format.
     * <p>
     * This method can be called from any thread.  If called from a thread other than
     * the JavaFX Application Thread, the current thread will be paused until the screenshot is taken.
     *
     * @return the byte array containing the screenshot
     * @throws IOException when an I/O error occurs
     */
    public static byte[] takeScreenshot() throws IOException {
```

and 

```java
    /**
     * Captures a screenshot using JavaFX {@link Robot} in the PNG format,
     * in the form of a Base-64 encoded {@code String}.
     * <p>
     * This method can be called from any thread.  If called from a thread other than
     * the JavaFX Application Thread, the current thread will be paused until the screenshot is taken.
     *
     * @param prefix the string to append before the base-64 representation, or null
     * @param postfix the string to append after the base-64 representation, or null
     * @return the screenshot in Base-64 encoded PNG, or an error message
     */
    public static String takeScreenshotBase64(String prefix, String postfix) {
```

What do you think?

kevinrushforth:
> We might still provide general purpose methods in ScreenCapture (new class) like

That might work. So the idea would be a utility class not tied to a JUnit5 annotation? The usage would then be adding a call to that while debugging a test wherever it makes sense for your test, possible from an AfterEach method.

>     public static byte[] takeScreenshot() throws IOException {

What would this do? I can see the usefulness of returning a WritableImage (basically calling `Robot::getScreenCapture` on the whole screen) or a base64-encoded PNG file, but what does it mean to return a raw byte array? I probably wouldn't recommend this one.

>     public static String takeScreenshotBase64(String prefix, String postfix) {

This seems useful, although what are "prefix" and "postfix" for?

tests/system/src/test/java/test/util/ScreenCaptureTestWatcher.java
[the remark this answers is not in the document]
kevinrushforth: I would agree with this given the limitations using the annotation (especially the fact that it runs _after_ the `@AfterEach` method) and that we have no good way to throttle it.

If we go this way, the final version of the PR would need to remove the annotation from the existing tests.

andy-goryachev-oracle:
Adding `ScreenshotCapture.writeScreenshot(System.err);` before the failing line in the `RectangleTest` successfully captures the screenshot:

ss-3

andy-goryachev-oracle:
@kevinrushforth do you know why there are two java windows in the screenshot above?

kevinrushforth:
> @kevinrushforth do you know why there are two java windows in the screenshot above?

If you mean two _terminal_ windows, no I don't know. They're probably leftovers.

tests/system/src/test/java/test/util/ScreenshotCapture.java
```diff
@@ -0,0 +1,156 @@
...
+    /**
+     * Captures a screenshot using JavaFX {@link Robot} in the PNG format.
+     * <p>
+     * This method can be called from any thread.  If called from a thread other than
+     * the JavaFX Application Thread, the current thread will be paused until the screenshot is taken.
+     *
+     * @return the byte array containing the screenshot
+     * @throws IOException when an I/O error occurs
+     */
+    public static byte[] takeScreenshot() throws IOException {
```
kevinrushforth, line 60: I have no idea how to interpret this byte array. What is the use case for this method?

tests/system/src/test/java/test/util/ScreenshotCapture.java
```diff
@@ -0,0 +1,156 @@
...
+     * in the form of a Base-64 encoded {@code String}.
+     * <p>
+     * This method can be called from any thread.  If called from a thread other than
+     * the JavaFX Application Thread, the current thread will be paused until the screenshot is taken.
+     *
+     * @param prefix the string to append before the base-64 representation, or null
+     * @param postfix the string to append after the base-64 representation, or null
+     * @return the screenshot in Base-64 encoded PNG, or an error message
+     */
+    public static String takeScreenshotBase64(String prefix, String postfix) {
```
kevinrushforth, line 107: Passing in the `data:image/png;base64,` as a prefix is odd. This method takes a screen shot and turns it into a base64-encoded PNG. Don't make the caller pass in something that has to match this internal behavior of this method. I would recommend removing the prefix and postfix entirely, and having this method return a simple base64-encoded PNG with a ``data:image/png;base64,` data URL. The caller can add any other String before or after this.

tests/system/src/test/java/test/util/ScreenshotCapture.java
```diff
@@ -0,0 +1,156 @@
...
+     * <p>
+     * Example:
+     * <pre>
+     * Screenshot:
+     * data:image/png;base64,iVBORw0KGgoA...</pre>
+     *
+     * @param out the output print stream
+     */
+    public static void writeScreenshot(PrintStream out) {
+        out.println(ScreenshotCapture.takeScreenshotBase64("Screenshot:\ndata:image/png;base64,", null));
```
kevinrushforth, line 93: With the change I am suggesting to `takeScreenshotBase64 `, this would become:

```
        out.println("Screenshot:");
        out.println(takeScreenshotBase64());
```

That makes the `takeScreenshotBase64` API cleaner with a better separation of concerns.

kevinrushforth (commented):
The functionality looks good, and I confirm that there are now no uses of the new utility API or test watcher in our check-in tests. I left a couple comments on the API for the test methods.

kevinrushforth:
One other thing that might be helpful is class docs that describe the two different ways of using it (the watcher annotation and the direct call to the utility method).

kevinrushforth:
> One other thing that might be helpful is class docs that describe the two different ways of using it (the watcher annotation and the direct call to the utility method).

Scratch the first part of this suggestion: the annotation already has this. Maybe add class docs in the utility class showing it its use?

tests/system/src/test/java/test/util/ScreenshotCapture.java
andy-goryachev-oracle: To save to a file, or send to a stream, for example.  This is a general purpose test utility.  The javadoc says "in PNG format".

andy-goryachev-oracle:
> Maybe add class docs in the utility class

both classes have code samples to illustrate the usage - maybe insufficient.  any suggestions?

tests/system/src/test/java/test/util/ScreenshotCapture.java
andy-goryachev-oracle: I wanted to avoid creating multiple multi-megabyte strings.

The method to use is writeScreenshot(), which should be simple enough to use, it's the method mentioned at the class level javadoc.

Related: I was debating whether we should have two specific methods
```
writeScreenshotStdout();
writeSceeenshotStderr();
```
instead of passing `System:out` or `System::err`.  What do you think?

tests/system/src/test/java/test/util/ScreenshotCapture.java
andy-goryachev-oracle: The reason I have prefix and postfix instead of what you are proposing is that in my case only one event is written to the output.  This might be important when both stderr and stdout are redirected to the same log file, in which case it might produce an interleaved result.

Also, prefix/postfix allow for more flexibility, such as outputting a JSON segment like
```
{ screenshot:"..." }
```
or any other representation depending on the situation.

tests/system/src/test/java/test/util/ScreenshotCapture.java
kevinrushforth: My suggestion doesn't change anything about how many or what size strings are allocated in the common case. All it does is moves which method tacks on the needed "data:image/png;base64," piece from the caller to the method that actually creates the PNG.

As for two methods vs one, I like the flexibility as you have it, although really only the System.err version is likely to be used (with gradle as the test runner, writing it to stdout isn't generally as useful).

kevinrushforth:
> > Maybe add class docs in the utility class
> 
> both classes have code samples to illustrate the usage - maybe insufficient. any suggestions?

No, I just failed to notice it. My fault.

tests/system/src/test/java/test/util/ScreenshotCapture.java
kevinrushforth: To be clear, I agree that the caller is the one who should create the representation you show above. What I object to is having the caller pass in the first part of the data URL. That's not a clean separation of concerns.

As for whether to keep the prefix / postfix _without_ the data URL part, I don't care as much. I don't see it as needed, but it's fine if you want to keep it.

andy-goryachev-oracle: hmm, not sure if I share the concern (did I understand the concern?)

the idea is to have a low-level method `takeScreenshotBase64(prefix, postfix)` that can be used in non-standard situations, but also provide the convenience method that will be used in the tests, `writeScreenshotBase64()`.

kevinrushforth: > hmm, not sure if I share the concern (did I understand the concern?)
> 
> the idea is to have a low-level method `takeScreenshotBase64(prefix, postfix)` that can be used in non-standard situations, but also provide the convenience method that will be used in the tests, `writeScreenshotBase64()`.

That part sounds fine. However, even in non-standard situations, the caller of `takeScreenshotBase64` should not be expected to pass `"data:image/png;base64,"` to that method. It is the method itself that knows that the format is PNG, so that method should provide that part of the string when converting it to a base64 screenshot. So, for example:

```
    takeScreenshotBase64(null, null);  <-- returns a data URL of a base64-encoded PNG
    takeScreenshotBase64("{ screenshot:", "}"); <-- return the same with the prefix and postfix
```

So what I'm trying to say is that `"data:image/png;base64,"` is an integral part of the base64-encoded image, not an optional prefix that a caller must pass in.

andy-goryachev-oracle: One thing that happened recently (in the past 10 years or so) is that JSON became a frequently used format for logs.  It's not perfect, but it's easy to parse.  With that, json-oriented log viewers came.

We don't have a consistent way to log data (or show data in `.toString()`).  A good log viewer can pretty print an object in the log file, decode a hex- or base-64 encoded string, or even run a query on a log file.

We are not ready for JSON logs, I admit, but this was the rationale behind the design of this class:

- a low-level method that returns byte[]
- a base-64 encoding method that allows for custom prefix/suffix to be able to do a data url or a json
- a convenience method to use in tests

Also, considering what you said about gradle and stderr, perhaps `writeScreenshot()` should always emit to stderr.

andy-goryachev-oracle: Not a "good" log viewer, but to illustrate:

Screenshot 2025-04-04 at 14 41 54

kevinrushforth: > We are not ready for JSON logs, I admit, but this was the rationale behind the design of this class:
> 
> * a low-level method that returns byte[]
> * a base-64 encoding method that allows for custom prefix/suffix to be able to do a data url or a json
> * a convenience method to use in tests

Are you saying that in a JSON image, you wouldn't have the `""data:image/png;base64,"` prefix?

> Also, considering what you said about gradle and stderr, perhaps `writeScreenshot()` should always emit to stderr.

That seems reasonable. We can always add an overload that takes a PrintStream if needed later.

andy-goryachev-oracle: > in a JSON image, you wouldn't have the `""data:image/png;base64,"` prefix?

correct, the prefix would be something else, like
`takeScreenshotBase64("{screenshot:","}");`

kevinrushforth: Are you sure? How would it know it was PNG (and not JPG or BMP)?

In any case, if there is a possibility that the caller might not want the base64-encoded image to always be prefixed with "data:image/png;base64," then passing it in is OK. Honestly, though, this is a test utility, so I don't image it will ever get used.

kevinrushforth:
I'll do a quick re-test then approve.

andy-goryachev-oracle:
Thank you for an interesting discussion!

kevinrushforth (approved):
LGTM with one very minor comment on a comment. I'll reapprove if you fix.

tests/system/src/test/java/test/util/ScreenCaptureTestWatcher.java
```diff
@@ -42,7 +42,7 @@
 public class ScreenCaptureTestWatcher implements TestWatcher {
     @Override
     public void testFailed(ExtensionContext extensionContext, Throwable err) {
-        // can be pasted into Safari address bar
-        ScreenshotCapture.writeScreenshot(System.err);
+        // the data url can be pasted into Safari address bar to view the screenshot
```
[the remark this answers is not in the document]
kevinrushforth: Although apparently not Chrome. :)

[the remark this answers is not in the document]
andy-goryachev-oracle: Actually, Safari is the browser which will show the complete screenshot.  Chrome will only show a truncated screenshot, and I don't have Firefox to test.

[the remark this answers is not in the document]
andy-goryachev-oracle: I was planning to add a decoding tool to the MonkeyTester anyway.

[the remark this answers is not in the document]
kevinrushforth: > Actually, Safari is the browser which will show the complete screenshot. Chrome will only show a truncated screenshot, and I don't have Firefox to test.

I did all my testing on Firefox, both Windows and macOS. It also works. Chrome truncates it so small that it is useless.

> I was planning to add a decoding tool to the MonkeyTester anyway.

Good.

kevinrushforth:
Since this is now just an optional test utility for debugging, and isn't wired up to anything, it doesn't need two reviewers.

/reviewers 1

PR #1748, opened 2025-03-29
8353223: [Linux] - Maximized windows do not retain correct size when resized programmatically

tsayao:
When a window is maximized, if only the width or height is programmatically set, the window will retain its current maximized dimension (either height or width) as is, unless the other dimension is also explicitly set.

This is consistent with what Gtk does.

---------
### Progress

### Issue
 * [JDK-8353223](https://bugs.openjdk.org/browse/JDK-8353223): [Linux] - Maximized windows do not retain correct size when resized programmatically (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1748/head:pull/1748` \
`$ git checkout pull/1748`

`$ git checkout pull/1748` \
`$ git pull https://git.openjdk.org/jfx.git pull/1748/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1748`

`$ git pr show -t 1748`

</details>

<a href="https://git.openjdk.org/jfx/pull/1748.diff">https://git.openjdk.org/jfx/pull/1748.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1748#issuecomment-2763341878)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
+++ b/modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
@@ -1069,6 +1069,17 @@ void WindowContextTop::set_bounds(int x, int y, bool xSet, bool ySet, int w, int
     geometry.gravity_x = gravity_x;
     geometry.gravity_y = gravity_y;
 
+    // When the window is programmatically resized while maximized
+    if (is_maximized) {
+        if (w < 0 && cw < 0) {
+            cw = gdk_window_get_width(gdk_window);
+        }
+
+        if (h < 0 && ch < 0) {
+            ch = gdk_window_get_height(gdk_window);
+        }
+    }
+
     if (w > 0) {
         geometry.final_width.type = BOUNDSTYPE_WINDOW;
         geometry.final_width.value = w;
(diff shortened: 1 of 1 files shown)


tsayao:
I initially pasted the wrong JBS issue on the title because I was multitasking :)

tsayao:
I'm having second thoughts about this, begin that resize should be ignored if the window is maximized.

beldenfox:
I did a quick test on Mac and Windows. I modified the original test case so I could visually track what's going on; instead of reducing the width of the window by 1 unit I reduce it by 30.

Mac - when I click the button the window is resized losing 30 units on the right side.

Windows 11 - the window is immediately resized losing 30 units on the right side. The title bar is resized but not as much (?) so it sticks out and is missing the usual rounded corners. The icon in the title bar indicate that Windows still thinks the window is maximized.

For what it's worth on Windows and Mac the setWidth() call doesn't seem to be changing the un-maximized size tracked by the OS. If I click on the un-maximize button it looks like it's going down to the original 640x480.

tsayao:
The documentation states about fullscreen (on `setWidth` / `setHeight`):

> If this Window is an instance of Stage, changing this attribute will not visually affect the Window while [fullScreen](https://openjfx.io/javadoc/21/javafx.graphics/javafx/stage/Stage.html#fullScreenProperty()) is true, but will be honored by the Window once [fullScreen](https://openjfx.io/javadoc/21/javafx.graphics/javafx/stage/Stage.html#fullScreenProperty()) becomes false.

It would make sense if Maximized followed the same behavior.

GTK ignores all `gtk_window_resize `calls when the window is maximized and restores it to its original size when unfullscreened.

If macOS and Windows handle fullscreen the same way as GTK, it would be better to update the documentation accordingly.

tsayao:
I will do a more broad fix.

PR #1751, opened 2025-04-01
8352982: gradle TEST_SDK_PATH param doesn't work with relative paths

arapte:
Issue:
The test execution fails when a relative path is specified a relative path for `TEST_SDK_PATH`.
For example:
1. gradle -PTEST_ONLY=true -PTEST_SDK_PATH=**..**/jfx1/build test 
2. gradle -PTEST_ONLY=true -PTEST_SDK_PATH=**~**/jfx1/build test

Solution:
Convert the relative path to an absolute path.

More about fix:
The property TEST_SDK_PATH belongs to the root project rt in build.gradle.
If we modify this property in build.gradle, it does not reflect in the child projects. The child projects for example graphics, controls would still have the value of TEST_SDK_PATH before modification.
To solve this, I Introduced a new variable in build.gradle `TEST_SDK_DIR`. It would hold the modified value of `TEST_SDK_PATH` and gets correctly reflected across all sub-projects.

Verified that both relative paths and absolute path work fine after this change.

---------
### Progress

### Warning
&nbsp;⚠️ Found leading lowercase letter in issue title for `8352982: gradle TEST_SDK_PATH param doesn't work with relative paths`

### Issue
 * [JDK-8352982](https://bugs.openjdk.org/browse/JDK-8352982): gradle TEST_SDK_PATH param doesn't work with relative paths (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1751/head:pull/1751` \
`$ git checkout pull/1751`

`$ git checkout pull/1751` \
`$ git pull https://git.openjdk.org/jfx.git pull/1751/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1751`

`$ git pr show -t 1751`

</details>

<a href="https://git.openjdk.org/jfx/pull/1751.diff">https://git.openjdk.org/jfx/pull/1751.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1751#issuecomment-2768329277)
</details>

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -725,11 +725,7 @@ if (hasProperty("TEST_SDK_PATH")) {
     if (!IS_TEST_ONLY) {
         fail("TEST_ONLY must be set to true when using an external TEST_SDK_PATH for running tests: include -PTEST_ONLY=true")
     }
-    String testSdkPath = "${TEST_SDK_PATH}"
-    if (testSdkPath.startsWith(".")) {
-        testSdkPath = file(testSdkPath).absolutePath
-    }
-    ext.TEST_SDK_DIR = testSdkPath
+    ext.TEST_SDK_DIR = file("${TEST_SDK_PATH}").absolutePath
     def sdkShimsPath = file("${TEST_SDK_DIR}")
     def sdkPath = file("${TEST_SDK_DIR}/sdk")
     def shimsPath = file("${TEST_SDK_DIR}/shims")
@@ -742,7 +738,6 @@ if (hasProperty("TEST_SDK_PATH")) {
 } else {
     ext.TEST_SDK_DIR = "${rootProject.buildDir}"
 }
-println "TEST_SDK_PATH: " + TEST_SDK_DIR
 
 // Make sure JDK_HOME/bin/java exists
 if (!file(JAVA).exists()) throw new Exception("Missing or incorrect path to 'java': '$JAVA'. Perhaps bad JDK_HOME? $JDK_HOME")
@@ -1503,6 +1498,9 @@ logger.quiet("OS_NAME: $OS_NAME")
 logger.quiet("OS_ARCH: $OS_ARCH")
 logger.quiet("JAVA_HOME: $JAVA_HOME")
 logger.quiet("JDK_HOME: $JDK_HOME")
+if (IS_TEST_SDK) {
+    logger.quiet("TEST_SDK_PATH: ${TEST_SDK_DIR}")
+}
 
 Version javaVersionInfo = Runtime.version()
 logger.quiet("java.runtime.version: " + javaVersionInfo)
--- a/build.gradle
+++ b/build.gradle
@@ -719,22 +719,34 @@ defineProperty("TEST_ONLY", "false")
 ext.IS_TEST_ONLY = Boolean.parseBoolean(TEST_ONLY) || !buildSdkForTest
 
 ext.IS_TEST_SDK = false
+ext.TEST_SDK_DIR = ""
+
 if (hasProperty("TEST_SDK_PATH")) {
     if (!IS_TEST_ONLY) {
         fail("TEST_ONLY must be set to true when using an external TEST_SDK_PATH for running tests: include -PTEST_ONLY=true")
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -719,22 +719,34 @@ defineProperty("TEST_ONLY", "false")
...
     }
-    def sdkShimsPath = file("${TEST_SDK_PATH}")
-    def sdkPath = file("${TEST_SDK_PATH}/sdk")
-    def shimsPath = file("${TEST_SDK_PATH}/shims")
-    if (!sdkShimsPath.exists() || !sdkShimsPath.isDirectory() ||
-             !sdkPath.exists() || !sdkPath.isDirectory() ||
-           !shimsPath.exists() || !shimsPath.isDirectory()) {
-        fail("The provided TEST_SDK_PATH=${TEST_SDK_PATH} is invalid")
+    String testSdkPath = "${TEST_SDK_PATH}"
+    if (testSdkPath.startsWith("~")) {
```
kevinrushforth, line 729: This isn't related to handling relative paths at all. Rather it is related to how bash works. The `~` character is handled by the shell and it is a known limitation that you can't use it in the middle of a string.

kevinrushforth (commented):
I'll test and review this later, but one quick comment:

> 2. gradle -PTEST_ONLY=true -PTEST_SDK_PATH=~/jfx1/build test

This isn't a bug, but is a known limitation in the way shell expansion of `~` works. If we were going to deal with this, we would want to do it consistently for all path names, in all of our build scripts, not just this one path name in `build.gradle`. Let's not.

build.gradle
arapte: Thanks Kevin, I shall remove the part related to `~`

build.gradle
```diff
@@ -719,22 +719,30 @@ defineProperty("TEST_ONLY", "false")
...
+
 if (hasProperty("TEST_SDK_PATH")) {
     if (!IS_TEST_ONLY) {
         fail("TEST_ONLY must be set to true when using an external TEST_SDK_PATH for running tests: include -PTEST_ONLY=true")
     }
-    def sdkShimsPath = file("${TEST_SDK_PATH}")
-    def sdkPath = file("${TEST_SDK_PATH}/sdk")
-    def shimsPath = file("${TEST_SDK_PATH}/shims")
+    String testSdkPath = "${TEST_SDK_PATH}"
+    if (testSdkPath.startsWith(".")) {
```
kevinrushforth, line 729: This will exclude relative paths underneath the jfx directory you are in unless you explicitly add `./` before the directory. It's a rare case, but seems worth handling. Perhaps you can remove the `if` test entirely and always use `file(testSdkPath).absolutePath`?

build.gradle
```diff
@@ -719,22 +719,30 @@ defineProperty("TEST_ONLY", "false")
...
              !sdkPath.exists() || !sdkPath.isDirectory() ||
            !shimsPath.exists() || !shimsPath.isDirectory()) {
         fail("The provided TEST_SDK_PATH=${TEST_SDK_PATH} is invalid")
     }
     ext.IS_TEST_SDK = true
 } else {
-    defineProperty("TEST_SDK_PATH", "${rootProject.buildDir}")
+    ext.TEST_SDK_DIR = "${rootProject.buildDir}"
 }
+println "TEST_SDK_PATH: " + TEST_SDK_DIR
```
kevinrushforth, line 745: Minor: Consider moving this down to the block where most other build properties like this are logged?

kevinrushforth (commented):
Relative paths that start with a `.` now work for me. I noted one more case that should be handled and also left one minor comment.

build.gradle
arapte: Moved to use `logger.quiet` along with other properties.

build.gradle
arapte: > always use file(testSdkPath).absolutePath?

This sounds better, changed accordingly.

kevinrushforth (approved):
LGTM

PR #1754, opened 2025-04-01
8207333: [Linux, macOS] Column sorting is triggered always after context menu request on table header

jperedadnr:
Note: The JBS issue [JDK-8207333](https://bugs.openjdk.org/browse/JDK-8207333) refers to Linux, but it happens on macOS too. 

When a TableColumn has a ContextMenu, if the user right clicks on the tableColumnHeader, the ContextMenu shows up, but, as described on the issue, on macOS and Linux, the table gets sorted as well.

Currently, `TableColumnHeader#mouseReleasedHandler` checks for `MouseEvent::isPopupTriggered`, but it doesn't have a check on `mousePressed`. However,  it can be seen that a right click on the header has the following values for `MouseEvent:: isPopupTriggered` on the different platforms and mouse pressed and released events:

| isPopupTriggered on: | Windows  | macOS | Linux |
| ------------- | ------------- | ------------- | ------------- |
| mousePressed  | false  | true | true |
| mouseReleased  | true  | false | false |

Also, the JavaDoc for `MouseEvent::isPopupTriggered` clearly states that:

> **Note**: Popup menus are triggered differently on different systems. Therefore, `popupTrigger` should be checked in both `onMousePressed` and `mouseReleased` for proper cross-platform functionality.

Therefore, this PR adds a check to `MouseEvent::isPopupTrigger` in the mouse pressed event, that can be used later on to cancel the header sorting when the mouse released event happens.

Also a system test has been added. It fails on macOS and Linux, and passes on Windows before this PR, and passes on the three platforms with this PR.

---------
### Progress

### Issue
 * [JDK-8207333](https://bugs.openjdk.org/browse/JDK-8207333): [Linux, macOS] Column sorting is triggered always after context menu request on table header (**Bug** - P4)

### Reviewers
 * [Ziad El Midaoui](https://openjdk.org/census#zelmidaoui) (@Ziad-Mid - Author)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1754/head:pull/1754` \
`$ git checkout pull/1754`

`$ git checkout pull/1754` \
`$ git pull https://git.openjdk.org/jfx.git pull/1754/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1754`

`$ git pr show -t 1754`

</details>

<a href="https://git.openjdk.org/jfx/pull/1754.diff">https://git.openjdk.org/jfx/pull/1754.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1754#issuecomment-2770211003)
</details>


andy-goryachev-oracle:
I'll take a look.
Should we update the JBS and this PR title to remove `[linux]` as it applies to both platforms?

andy-goryachev-oracle:
I can only test it on macOS and Windows.  Could someone with a linux system help please?
/reviewers 2

andy-goryachev-oracle (approved):
thank you for fixing this bug, @jperedadnr !

testing with the monkey tester looks good on macOS 15.3.2 and windows 11.

andy-goryachev-oracle:
@Ziad-Mid could you be a second reviewer?

Ziad-Mid:
> @Ziad-Mid could you be the second reviewer?

I can't test on Linux , I will check same for macOS

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
```diff
@@ -243,6 +245,7 @@ public TableColumnHeader(final TableColumnBase tc) {
         }
 
         if (me.isConsumed()) return;
+        popupTriggeredOnMousePressed = me.isPopupTrigger();
         me.consume();
 
         header.getTableHeaderRow().columnDragLock = true;
```
beldenfox, line 251: On Mac a context menu can be invoked using the right button or the left button in conjunction with the Control key. In the second case the pop trigger flag will be set and the primary button will be down. If I'm reading the code correctly this could get the header stuck in column re-ordering mode.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
```diff
@@ -271,7 +274,10 @@ public TableColumnHeader(final TableColumnBase tc) {
         TableColumnHeader header = (TableColumnHeader) me.getSource();
         header.getTableHeaderRow().columnDragLock = false;
```
beldenfox, line 275: On Ubuntu 22.04 the table column header is not receiving a mouse released event if a context menu is shown. Not sure where the released event is going but it's not to any node on the primary stage. The docs are vague on this but this looks like a separate bug that's making it hard to test this PR.

andy-goryachev-oracle: what happens in the master branch (without this PR's changes)?

beldenfox: I was having trouble reproducing the original bug on the master branch; the sorting is done when the mouse is released and most of the time that event never arrives (I think it does every now and then).

andy-goryachev-oracle: this is strange - the bug is immediately apparent.  which macOS version do you have?  although I don't think the version matters.

I am testing with an external mouse, with control-left-click, and with the double-finger-click gesture on the trackpad.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
jperedadnr: control+left-click works as expected for me too.

When the context menu is showing, clicking anywhere outside of it, closes the popup (because of the autohide property), but doesn't consume the event. If you click on a button, for instance, it will be fired. In any case, this is the same before and after this PR, so maybe an issue for a later discussion.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
jperedadnr: On Linux I have some issues too: the test that I attached to the JBS issue doesn't fail for me on Linux, but the test that I added to this PR does fail without the fix.

andy-goryachev-oracle: Can you try the monkey tester?  it allows setting properties on the table view / table column at runtime...

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
beldenfox: > control+left-click works as expected for me too.

For control+left-click it looks like the code will call `columnReordingStarted` but on the released event it won't call `columnReorderingComplete`. I have no idea if this will lead to problems or not.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
beldenfox: I just verified that on Linux the released event usually goes to the popup window. Basically it's whatever window is under the cursor at the time; if I move the mouse fast enough it may go to a node on the primary stage. I don't think this is our doing, it's how GTK is delivering the event.

On the JavaFX side I'm not sure this counts as a bug. The docs state that entered and released events should arrive in pairs during drag and drop operations but doesn't address this case.

jperedadnr: The Monkey tester app works for me on Linux (Ubuntu 22.04, either mouse or pad), before and after this PR (so I can't reproduce the issue there, as I mentioned, though the test added to the PR fails without the patch).

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java
jperedadnr: @beldenfox Do you think we should address that issue (ctrl+left click + columnReordering) in this PR? Else, do we need to do anything else to move it forward?

beldenfox: You should be good with a few tweaks. If the press event is a popup trigger don't set the columnDragLock or call columnReorderingStarted (since you might not get a release event). In the drag handler check popupTriggeredOnMousePressed and do nothing if it's set. Your released handler looks fine as-is. I think that should cover it.

By the way, in a perfect world you would have more predictable behavior. When the popup menu is shown it would grab the mouse pointer and you would never see the release event. That would allow the user to click to bring up the popup and start dragging across the menu items without having to release the mouse.

jperedadnr: Thanks, @beldenfox! I've included your suggestions,  and also added the ctrl+left click macOS case to the test.

andy-goryachev-oracle:
Some tests in TableColumnHeaderTest are failing in CI.

jperedadnr:
Indeed... this test fails: `TableColumnHeaderTest::testColumnRightClickDoesAllowResizingWhenConsumed`.

It has this JavaDoc:
```
/**
     * When a right click is done on a table column header and consumed by a self added event handler, the column
     * drag lock should be set to true in the pressed handler, but still to false again in the released handler.<br>
     * By that we guarantee, that the column resizing still works.
     */
  ```
This was changed based on @beldenfox suggestion. I'll change it back.

andy-goryachev-oracle:
much better, thank you.  btw, you can re-start failing tasks - I hope it's just an intermittent failure.

jperedadnr:
It has failed three times in a row. Most likely this has already been fixed with JDK-8354337, so I'll sync from head.

jperedadnr:
It is green now (https://github.com/jperedadnr/jfx/actions/runs/14715013268) after I ran again the failed gradle wrapper validation.

beldenfox:
> This was changed based on @beldenfox suggestion. I'll change it back.

If the mouse pressed event triggers a context menu you might never see the mouse released event. So you have to consider what it means if column drag lock gets set and then never cleared.

The failing test strikes me as odd, it is explicitly testing a bit of internal state to verify that a right-click drag operation will work. Users don't expect this to work and overloading right-click makes life complicated so right-click drags are generally disabled even on Windows. I understand that this test case was added because the table header was getting into a persistent bad state but unless I'm missing something (and I certainly might be) the bug should have been resolved by turning right-click drag into a no-op. (It doesn't help that the test uses the MouseEventFirer which sets the isPopupTrigger flag on both pressed and release which doesn't match the behavior of any platform.)

jperedadnr:
@beldenfox I understand your concerns about `columnDragLock` and the related tests `testColumnRightClickDoesAllowResizing()` and `testColumnRightClickDoesAllowResizingWhenConsumed()`, but if I get it right, this PR doesn't really modify the behaviour of such lock.

So I suggest we take this on a follow up issue, revisiting the need to change the lock from true to false when the popup is triggered, and modifying or removing those tests accordingly.

What do you think?

jperedadnr:
@andy-goryachev-oracle About this edge case, on macOS it seems better behaviour after this PR: before the highlighting jumps unexpectedly, while after it, the highlighted area is updated smoothly. And I don't see the popup flashing.
About Windows, do you mean we should try to reproduce the same behaviour as on macOS? I'm not sure this should be part of this PR (the issue was only on macOS/Linux).

andy-goryachev-oracle:
My immediate thought was we should finish the drag to reorder gesture and avoid showing the popup if the right mouse button is pressed during the drag operation, regardless of the platform (this will change the behavior but I think it makes more sense).  I would recommend doing it as a part of this PR, but it's up to you.

Doing this will also avoid the issue we saw on windows where the highlighted column being dragged jumps suddenly to the original position which is simply wrong.

What do you think?

beldenfox:
I think this PR should probably be accepted as-is. It fixes an obvious problem and doesn't introduce any new ones. Someone should enter a follow-up PR citing some of the other problems but most of those have been around for years and only affect users who are looking for trouble (most wouldn't even consider using the right mouse button to drag).

Here you have separate behavior on click vs drag-click and have to deal with two different approaches to context menus. I've tackled this sort of thing before and it usually takes two or four or six tries to get right. To reduce the complexity you have be aggressive at ignoring irrelevant events (no right-click drag!) and avoid setting up state related to dragging until the user actually starts dragging. So this code needs a modest re-write but, again, it doesn't need to happen in this PR.

kevinrushforth:
> I think this PR should probably be accepted as-is. It fixes an obvious problem and doesn't introduce any new ones. Someone should enter a follow-up PR citing some of the other problems but most of those have been around for years and only affect users who are looking for trouble ...

I tend to agree, as long as any remaining problems are preexisting and not likely to be encountered in the primary use case.

andy-goryachev-oracle:
The windows problem is pre-existing.

The after-the-fix behavior is slightly better.  I'll create the follow-up ticket.

andy-goryachev-oracle (approved):

andy-goryachev-oracle:
created https://bugs.openjdk.org/browse/JDK-8355939

Ziad-Mid (approved):
LGTM, test fail in master and work with the fix

andy-goryachev-oracle:
This PR is ready to integrate.

PR #1755, opened 2025-04-02
8354631: [macos] OpenURIHandler events not received by AWT when JavaFX is primary toolkit

pabulaner:
When trying to register an open URI handler when using JavaFX with a native menu, this task fails on Mac.
Either the native menu is not shown or the URIs are not received.

This pull request fixes this issue if AWT is registered after JavaFX, so that AWT runs embedded inside JavaFX.
It fixes this by introducing a native event to AWT, which can be used by JavaFX to forward events such as an openURL event.

The test for this pull request is non trivial, as the application needs to be installed on the Mac before it can be tested. Therefore the test is provided in a separate repository and it needs to be discussed if the test is necessary to have inside the JFX repo and if so, how it should be integrated.

JDK Pull Request: https://github.com/openjdk/jdk/pull/24379
Co-Author: @FlorianKirmaier

Link to the test repo: https://github.com/pabulaner/openurifx

---------
### Progress

### Issue
 * [JDK-8354631](https://bugs.openjdk.org/browse/JDK-8354631): [macos] OpenURIHandler events not received by AWT when JavaFX is primary toolkit (**Bug** - P4)

### Reviewers
 * [Harshitha Onkar](https://openjdk.org/census#honkar) (@honkar-jdk - no project role)

### Contributors
 * Florian Kirmaier `<fkirmaier@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1755/head:pull/1755` \
`$ git checkout pull/1755`

`$ git checkout pull/1755` \
`$ git pull https://git.openjdk.org/jfx.git pull/1755/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1755`

`$ git pr show -t 1755`

</details>

<a href="https://git.openjdk.org/jfx/pull/1755.diff">https://git.openjdk.org/jfx/pull/1755.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1755#issuecomment-2802979217)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
@@ -474,6 +474,21 @@ jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved)
     return YES;
 }
 
+- (void) application:(NSApplication *)theApplication openURLs:(NSArray<NSURL *> *)urls
+{
+    for (NSURL* url in urls) {
+         NSDictionary *userInfo = @{
+            @"name": @"openURL",
+            @"url": url.absoluteString
+        };
+
+        [[NSNotificationCenter defaultCenter]
+                postNotificationName:@"EmbeddedEvent"
+                object:nil
+                userInfo:userInfo];
+    }
+}
+
 - (BOOL)applicationShouldOpenUntitledFile:(NSApplication *)sender
 {
     LOG("GlassApplication:applicationShouldOpenUntitledFile");
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
@@ -78,6 +78,11 @@ static NSString* JavaRunLoopMode = @"AWTRunLoopMode";
 // don't deadlock.
 static NSArray<NSString*> *runLoopModes = nil;
 
+// Custom event that is provided by AWT to allow libraries like
+// JavaFX to forward native events to AWT even if AWT runs in
+// embedded mode.
+static NSString* awtEmbeddedEvent = @"AWTEmbeddedEvent";
+
 #ifdef STATIC_BUILD
 jint JNICALL JNI_OnLoad_glass(JavaVM *vm, void *reserved)
 #else
@@ -483,7 +488,7 @@ jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved)
         };
 
         [[NSNotificationCenter defaultCenter]
-                postNotificationName:@"EmbeddedEvent"
+                postNotificationName:awtEmbeddedEvent
                 object:nil
                 userInfo:userInfo];
     }


kevinrushforth:
@pabulaner We'll take a look once your OCA is recorded.

/reviewers 2

kevinrushforth:
@pabulaner In order for your OCA status to be verified, you need to respond to the [bot's instructions](#issuecomment-2772685035) to confirm your OCA status by entering `/covered`, if you are covered by your employer's OCA, or `/signed`, if you signed the OCA yourself.

pabulaner:
Are there any news on my OCA status?

kevinrushforth:
> Are there any news on my OCA status?

@robilad Can you provide a status update?

robilad:
OCA processed and account now marked as verified in Skara

kevinrushforth:
We'll need some sort of standalone test, even if it can't be automated, that we can use to test the fix. Something that isn't a full-blown app. The test program should be able to reproduce the problem without the fix and can then be used to verify the fix.

When testing, we will want to consider the following scenarios:

1. Pure JavaFX app (even though there isn't a `java.awt.desktop.OpenURIHandler` in this case, we need to make sure that it doesn't misbehave)
2. JavaFX + Swing app -- JavaFX Toolkit initialized first (I think this is the main case you want to handle)
3. JavaFX + Swing app -- AWT Toolkit initialized first (I presume this case already works, so the main thing is to ensure no regressions)

kevinrushforth:
Also, since the proposed fix is in JavaFX, the JBS bug needs to be updated to the correct component/subcomponent/version. I can do that and also assign the bug to @FlorianKirmaier since you list him as a co-author of this fix. If you wish to have Skara record Florian as a co-author, you can use Skara's `/contributor add` command.

kevinrushforth:
> Also, since the proposed fix is in JavaFX, the JBS bug needs to be updated to the correct component/subcomponent/version.

I see that it's not quite that simple, since there is a corresponding AWT fix.

You cannot use the same bug ID for both a JDK fix and a JavaFX fix. File a separate bug for the JavaFX half of the fix, and use that new bug for this JavaFX PR. Perhaps @FlorianKirmaier can do this?

The two fixes will need to be done and tested both independently and together. It is OK if the bug isn't fully fixed without both halves of the fix. However, it is not OK for it to misbehave in some other way (exception, crash, etc) if you run a JavaFX without the fix with an JDK with the fix or vice versa.

So there now needs to be a matrix of testing that includes:

(JDK without/with fix) X (JavaFX without/with fix)

For each of those 4 cases you will need to test:

* JavaFX + Swing app -- JavaFX Toolkit initialized first
* JavaFX + Swing app -- AWT Toolkit initialized first

pabulaner:
And thank You for the quick reply, I will look into it and ask @FlorianKirmaier for his support.

FlorianKirmaier:
I've created a new ticket for the JFX Part of the fix:
https://bugs.openjdk.org/browse/JDK-8354631
@pabulaner Can you change the title and ticket number in the PR, so it matches with the new ticket?

Let me know, if i should change anything, in the ticket.

pabulaner:
So I adapted the test to the new requirements. 
The test can be found here:

https://github.com/pabulaner/openurifx

Currently I just put the JDK and JFX builds right into the repo, but they can be self build and easily changed inside the test script.
Maybe You can give me feedback if that is the test You expected or if it was something else.

When I tested it, it didn't seem to give any errors and worked without problems.

pabulaner:
I refactored the test repository to make more transparent which version is used and how it is build.

pabulaner:
@kevinrushforth could You please take a look again now?

kevinrushforth:
Will take a look tomorrow.

kevinrushforth (commented):
I verified that, in connection with the AWT fix, this fix allows the AWT OpenURIHandler to receive events when JavaFX is the primary toolkit.

I also verified that in the case where AWT is the primary toolkit, the code in question isn't ever called. So that mode works equally well with or without this fix.

I left a comment with a requested change, both to the name of the custom embedded event, and a request that it be defined as a static constant (put it near the top of the file).

kevinrushforth:
I updated the JBS title slightly for (I hope) better clarity. Please update the PR title to match.

pabulaner:
I implemented the feedback and changed the title.

kevinrushforth (commented):
The updated fix looks good. I verified that it works as expected.

I'll approve it early next week if no issues are found during testing.

prrace:
The JDK bug https://bugs.openjdk.org/browse/JDK-8332947 says this used to work on macOS 13.x but was broken by new requirements on 14.x ... it isn't spelled out enough for me to understand. Did it really used to work ?

Also today, FX has nothing that allows an app to  handle these events - ie no Desktop class.
if/when it does, then I presume that if ALSO the FX has registered to handle these events that in such a case FX would not forward them to FX.

Also if/when FX has a way to handle these events, wouldn't you want AWT to do the same .. and pass them on to FX if no AWT handler is registered ?

And its a bit more complicated then than "I'm the hosting toolkit, so I don't forward", there needs to be some check for the app having registered a handler.

pabulaner:
Interesting feedback.

I was told that it did work on an older version, but I don't remember if I tested it or not. Apparently MacOS changed something in their event handling logic from one version to another so that only one event handler could be registered at a time.

If JavaFX would have its own Desktop class, then the NativeEvent handler would not need to forward those events anymore, as FX would handle it.

If JavaFX would handle some events that AWT doesn't, then I guess it would make sense to also allow AWT to send native events. Since I didn't found any such events yet, I didn't look into it.

I don't understand Your last point. If no event handler is registered, JavaFX still forwards them, but they won't do anything unless a handler is registered at some point. So what exactly do You mean?

So is there something You want me to do in the code?
Or did I answer Your questions?

prrace:
> Interesting feedback.

Most of it wasn't really directed at you .. more at Kevin :-)
The main question I had for you was about did it ever really work and I guess you aren't sure.

> I don't understand Your last point. If no event handler is registered, JavaFX still forwards them, but they won't do anything unless a handler is registered at some point. So what exactly do You mean?

Well we don't expect FX to know if an AWT event handler is registered - especially one at the application level, ie via
https://docs.oracle.com/en/java/javase/21/docs/api/java.desktop/java/awt/Desktop.html#setOpenURIHandler(java.awt.desktop.OpenURIHandler)

What I mean is that if FX acquired a similar API, and was the hosting toolkit, I would expect it to check if there was a "javafx.desktop.Desktop" URI Handler installed and forward only if that was "null". And vice versa for AWT.

prrace (approved):

kevinrushforth:
> Most of it wasn't really directed at you .. more at Kevin :-)

Those are good questions as to how we might need to evolve this if and when JavaFX adds desktop-like APIs. We might, indeed, want some sort of "forward these events to the other toolkit if and only if they aren't for you" on both ends.

> The main question I had for you was about did it ever really work and I guess you aren't sure.

Based on my testing of this, I very much doubt anything in macOS has changed in a way to affect this. I strongly suspect that it may have (accidentally) worked prior to JDK 22 / JavaFX 22 when we fixed the bug where AWT was taking over the NSApplicationDelegate even if it was embedded in JavaFX. See [JDK-8318854](https://bugs.openjdk.org/browse/JDK-8318854) and [JDK-8319669](https://bugs.openjdk.org/browse/JDK-8319669).

PR #1763, opened 2025-04-04
8234153: [TEST_BUG] Rewrite Popup_parentWindow_Test

lukostyra:
This change rewrites `Popup_parentWindow_Test` after a-long-time-ago-done changes to Popup public API.

Popup used to have an `owner` and `parentWindow` properties. I couldn't track how far back that change happened so I can't fully say what these Properties actually were, but to my knowledge both properties were replaced by `ownerNode` and `ownerWindow` Properties respectively. New Properties are read-only and are set while calling respective `Popup.show()` - `Popup.show(Node, ...)` sets both Properties, while `Popup.show(Window, ...)` sets only `ownerWindow` Property.

A lot of original test scenarios were cut out because with current Popup API they make little to no sense. Original `Popup_parentWindow_Test` extended `PropertiesTestBase` and ran the regular Property access checks. Those are parametrized and split into four test methods:
- `testGetBean` - confidence check for whether we can get the Property bean, unnecessary since new tests access the Properties directly
- `testGetName` - Properties in `PropertiesTestBase` are referred to by String-provided name and Reflection, so this was a confidence check whether those Properties actually exist in the object - unnecessary, since we access the Properties directly
- `testBinding` - I deemed those not doable, since the Properties we try to access are read-only and cannot be bound to
- `testBasicAccess` - This is the only test I found that was workable into a reimplementation, and so it is manually reimplemented as `Popup_owner_Test`

Moreover, because of lack of `Popup.setParent()` API I also deemed more complex test scenarios (cases referring to TestScene's `_window` Property) unnecessary. Since there is no `Popup.show(Scene, ...)` those cases were also eliminated. This leaves current test pool into three separate parameters for one, similar test case.

---------
### Progress

### Issue
 * [JDK-8234153](https://bugs.openjdk.org/browse/JDK-8234153): [TEST_BUG] Rewrite Popup_parentWindow_Test (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1763/head:pull/1763` \
`$ git checkout pull/1763`

`$ git checkout pull/1763` \
`$ git pull https://git.openjdk.org/jfx.git pull/1763/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1763`

`$ git pr show -t 1763`

</details>

<a href="https://git.openjdk.org/jfx/pull/1763.diff">https://git.openjdk.org/jfx/pull/1763.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1763#issuecomment-2778434099)
</details>


modules/javafx.graphics/src/test/java/test/javafx/stage/Popup_owner_Test.java
```diff
@@ -0,0 +1,261 @@
...
+            this.counter = 0;
+            this.expected = expected;
+        }
+
+        public void reset() {
+            counter = 0;
+        }
+
+        public void assertCalled() {
+            if (counter != expected) {
```
andy-goryachev-oracle, line 245: minor: this can probably be replaced by `assertEquals()`

modules/javafx.graphics/src/test/java/test/javafx/stage/Popup_owner_Test.java
```diff
@@ -0,0 +1,261 @@
...
+                fail(String.format("Listener has not been called, or was not called enough times! Called %d times, expected %d", counter, expected));
+            }
+        }
+
+        public void assertNotCalled() {
+            if (counter != 0) {
```
andy-goryachev-oracle, line 251: minor: `assertTrue()` here?

modules/javafx.graphics/src/test/java/test/javafx/stage/Popup_owner_Test.java
```diff
@@ -0,0 +1,261 @@
...
+import test.com.sun.javafx.test.objects.TestStage;
+
+
+/**
+ * This test specifically checks for ownerNode and ownerWindow properties.
+ * Because those properties are set via different show() methods instead of
+ * regular setter method, we can't use PropertiesTestBase for this check.
+ * Instead we rely on checking this manually.
+ */
+public final class Popup_owner_Test {
```
andy-goryachev-oracle, line 55: minor: maybe `Popup_Owner_Test`?

andy-goryachev-oracle (approved):
lgtm, with some very minor comments.

modules/javafx.graphics/src/test/java/test/javafx/stage/Popup_owner_Test.java
mstr2: We do seem to have this funny-looking `PascalCase_camelCase_Test` convention in several places.

lukostyra: Yeah I mostly followed the convention of `Popup_parentWindow_Test`. I assumed the original followed `ClassToTest_propertyName_Test` convention and since this one tests `ownerWindow` and `ownerNode` properties at the same time, the name got reduced to `owner`.

lukostyra:
Since it's a rewrite I'm going to assume this needs two reviewers (forgot to do that when submitting)

/reviewers 2

andy-goryachev-oracle (approved):

modules/javafx.graphics/src/test/java/test/javafx/stage/Popup_owner_Test.java
```diff
@@ -0,0 +1,257 @@
+/*
+ * Copyright (c) 2011, 2025, Oracle and/or its affiliates. All rights reserved.
```
arapte, line 2: Either we should use `git mv <oldFile> <newFile>` here to retain the file commit history.
or
As the test changed a lot from before, we could choose not to do `git mv`, in the case the copyright in new file should be only 2025

I would recommend to `git mv` as a general good practice.

arapte (changes requested):
Test looks good. We can have another bug to address the naming convention to change other similar classes too.
Suggesting one minor change.

modules/javafx.graphics/src/test/java/test/javafx/stage/Popup_owner_Test.java
lukostyra: This was done with `git mv`. The problem is that the similarity between the old file and new file was too little for Git to consider this to be an actual rename, even when using `git mv`. I believe the similarity index threshold in Git is 50%.

The only way to achieve this would be to split it into two separate commits - rename, then rewrite - but that will still disappear since Skara will squish all the commits in a PR into one.

arapte (approved):
lgtm

lukostyra:
Thanks for review everyone!

/integrate

PR #1769, opened 2025-04-10
8088343: Intermittent unit test failure in javafx.concurrent.ServiceLifecycleTest

andy-goryachev-oracle:
The code should not set the `Task.state` value to `CANCELLED` if the said task is already `SUCCEEDED` or `FAILED`.

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/concurrent/Task.java
+++ b/modules/javafx.graphics/src/main/java/javafx/concurrent/Task.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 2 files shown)


kevinrushforth:

/reviewers 2

modules/javafx.graphics/src/main/java/javafx/concurrent/Task.java
```diff
@@ -1014,12 +1016,31 @@ protected void failed() { }
             // state flag will not be readable immediately after this call. However,
             // that would be the case anyway since these properties are not thread-safe.
             if (isFxApplicationThread()) {
+                switch (getState()) {
+                case FAILED:
+                case SUCCEEDED:
```
kevinrushforth, line 1021: Avoiding transitioning from a one of the completion states to canceled is a good thing. My question, though is: how did we get here? Is this masking some other problem?

modules/javafx.graphics/src/main/java/javafx/concurrent/Task.java
```diff
@@ -1014,12 +1016,31 @@ protected void failed() { }
...
+                    case SUCCEEDED:
+                        // a finished or failed task retains its state
+                        rv.set(false);
+                        break;
+                    default:
+                        setState(Worker.State.CANCELLED);
+                        break;
+                    }
+                });
+                return rv.get();
```
kevinrushforth, line 1041: This is ineffective since you don't wait for the runLater to execute (and waiting could lead to deadlock, so I wouldn't recommend that). I don't think there is anything better than unconditionally returning `flag` when not running on the FX app thread. That's what the current code does (and is what your proposed fix will do most of the time).

kevinrushforth (commented):
Avoiding the case where we transition from `FAILED` or `SUCCEEDED` to `CANCELLED` is good, and should fix the test failures (although I'll do some additional testing), but I can't help wondering if underlying problem is something else. The question I have is: _how_ do we get into a state where Worker.State of the Task one of the completion states (`FAILED` or `SUCCEEDED`), whereas the parent FutureTask is not completed, and so enters a canceled state. You can see this by instrumenting the code and calling `super.isCancelled()`.

I wonder if the custom service created by `TestServiceFactory` is causing, or contributing to, the problem?

modules/javafx.graphics/src/main/java/javafx/concurrent/Task.java
andy-goryachev-oracle: I think the variations in the test timing uncovered this issue.  I don't think it's a problem with the execution of the task, but rather with the reporting of its final state.

modules/javafx.graphics/src/main/java/javafx/concurrent/Task.java
andy-goryachev-oracle: good point!

In this particular case, returning `true` from `cancel()` when state will not be changed is acceptable.

kevinrushforth (approved):
LGTM. I instrumented the code and after a few test runs was able to spot a couple cases where this would have failed without the fix.

andy-goryachev-oracle:
@arapte or @Ziad-Mid : could one of you be the second reviewer?

arapte (approved):
LGTM... Could observe test failure without this change.

PR #1771, opened 2025-04-10
8354318: freetype.c: compilation error: 'incompatible pointer type' with gcc 14

kevinrushforth:
This PR fixes a bug where a local variable was declared to be the wrong type, causing an error at compilation time when compiling with gcc 14. gcc 14 rejects assignments of incompatible pointer types without an explicit cast. In this case, the solution is declare the variable to be of the correct type in the first place.


kevinrushforth:
Reviewer: @lukostyra

lukostyra (approved):
LGTM

PR #1772, opened 2025-04-11
8296554: MouseLocationOnScreenTest sometime fails when system is busy

Gopalora:
There was a Assertion fail issue in mouse location test case JDK-8296554,
Reason: We felt the one mili second delay time for the Robot test may be insufficient in few OS.
Solution: We Changed the delay time to three mili second.

Verification:
Tested in Windows 11, and the Assert fail issue is not found.

---------
### Progress

### Issue
 * [JDK-8296554](https://bugs.openjdk.org/browse/JDK-8296554): MouseLocationOnScreenTest sometime fails when system is busy (**Bug** - P5)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1772/head:pull/1772` \
`$ git checkout pull/1772`

`$ git checkout pull/1772` \
`$ git pull https://git.openjdk.org/jfx.git pull/1772/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1772`

`$ git pr show -t 1772`

</details>

<a href="https://git.openjdk.org/jfx/pull/1772.diff">https://git.openjdk.org/jfx/pull/1772.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1772#issuecomment-2796401103)
</details>

### DIFF
--- a/tests/system/src/test/java/test/robot/javafx/scene/MouseLocationOnScreenTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/scene/MouseLocationOnScreenTest.java
@@ -43,7 +43,7 @@ import test.util.Util;
 public class MouseLocationOnScreenTest {
     static CountDownLatch startupLatch = new CountDownLatch(1);
     static Robot robot;
-    private static int DELAY_TIME = 3;
+    private static int DELAY_TIME = 1;
 
     public static class TestApp extends Application {
 
@@ -118,15 +118,22 @@ public class MouseLocationOnScreenTest {
      * returned by robot are same
      */
     static void validate(Robot robot, int x, int y) {
-        Assertions.assertEquals(x, (int) robot.getMouseX());
-        Assertions.assertEquals(y, (int) robot.getMouseY());
+        try {
+            Util.sleep(DELAY_TIME);
+            Assertions.assertEquals(x, (int) robot.getMouseX());
+            Assertions.assertEquals(y, (int) robot.getMouseY());
+        }
+        catch(AssertionError e) {
+            Util.sleep(DELAY_TIME + 1); // 3 MS Delay
+            Assertions.assertEquals(x, (int) robot.getMouseX());
+            Assertions.assertEquals(y, (int) robot.getMouseY());
+        }
     }
 
     private static void edge(Robot robot, int x1, int y1, int x2, int y2) {
         for (int x = x1; x <= x2; x++) {
             for (int y = y1; y <= y2; y++) {
                 robot.mouseMove(x, y);
-                Util.sleep(DELAY_TIME);
                 validate(robot, x, y);
             }
         }
@@ -138,14 +145,12 @@ public class MouseLocationOnScreenTest {
         double dy = (y1 - y0) / dmax;
 
         robot.mouseMove(x0, y0);
-        Util.sleep(DELAY_TIME);
         validate(robot, x0, y0);
 
         for (int i = 1; i <= dmax; i++) {
             int x = (int) (x0 + dx * i);
             int y = (int) (y0 + dy * i);
             robot.mouseMove(x, y);
-            Util.sleep(DELAY_TIME);
             validate(robot, x, y);
         }
     }
--- a/tests/system/src/test/java/test/robot/javafx/scene/MouseLocationOnScreenTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/scene/MouseLocationOnScreenTest.java
@@ -43,7 +43,7 @@ import test.util.Util;
 public class MouseLocationOnScreenTest {
     static CountDownLatch startupLatch = new CountDownLatch(1);
     static Robot robot;
-    private static int DELAY_TIME = 1;
+    private static int DELAY_TIME = 3;
 
     public static class TestApp extends Application {
 
--- a/tests/system/src/test/java/test/robot/javafx/scene/MouseLocationOnScreenTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/scene/MouseLocationOnScreenTest.java
@@ -44,6 +44,7 @@ public class MouseLocationOnScreenTest {
     static CountDownLatch startupLatch = new CountDownLatch(1);
     static Robot robot;
     private static int DELAY_TIME = 1;
+    private static int VALIDATE_COUNT = 5;
 
     public static class TestApp extends Application {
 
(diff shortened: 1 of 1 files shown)


tests/system/src/test/java/test/robot/javafx/scene/MouseLocationOnScreenTest.java
```diff
@@ -43,7 +43,7 @@
 public class MouseLocationOnScreenTest {
     static CountDownLatch startupLatch = new CountDownLatch(1);
     static Robot robot;
-    private static int DELAY_TIME = 1;
+    private static int DELAY_TIME = 3;
```
andy-goryachev-oracle, line 46: minor: `Util.sleep()` could be moved to `validate()`

question: even though the value is 3x the original, should it be longer to account for any hiccups?  maybe 10?  what do you think?

kevinrushforth: If we go with this solution, I would recommend leaving it at 3ms. It fails very, very rarely with the existing 1ms, so 3x the time ought to be enough. Tripling it again would just make the test take that much longer without any real assurance that it is needed.

kevinrushforth (commented):
I observe the following behavior:

* On my Mac laptop, this makes the test take nearly three times as long -- 48 sec (use to take 17 sec), which is expected given that `sleep(DELAY)` is done in the inner loop.
* On my Windows laptop, this makes the test take twice as long -- 56 sec (use to take 28 sec)

This is getting to be about as much time as I would want this test to take, so I wouldn't increase it beyond 3ms without some evidence that it is needed.

What testing have you done to ensure that this is sufficient? Have you seen this fail on a particular system with a given load, and then observed that under the same conditions the updated test now passes?

andy-goryachev-oracle:
The other alternative (just a suggestion) is to modify the `validate()` method as follows:

- wait 1 ms
- check the coordinates
- if wrong, wait for longer (3? 5?)
- check again
- if coordinates are wrong, fail

what do you think?

Gopalora:
We may think to check pixels at 10 pixel gap. Each pixel makes the test process busy.

tests/system/src/test/java/test/robot/javafx/scene/MouseLocationOnScreenTest.java
```diff
@@ -118,15 +118,22 @@ public void testMouseLocation() throws Exception {
      * returned by robot are same
      */
     static void validate(Robot robot, int x, int y) {
-        Assertions.assertEquals(x, (int) robot.getMouseX());
-        Assertions.assertEquals(y, (int) robot.getMouseY());
+        try {
+            Util.sleep(DELAY_TIME);
+            Assertions.assertEquals(x, (int) robot.getMouseX());
```
andy-goryachev-oracle, line 123: this works also, though I'd simply use `if`.

A bigger question is why are we using `int` coordinates?  Wouldn't it make the test depend on the scale (and also on the screen resolution and window position)?

Should we instead use the double coordinates and `Assertions.equals(double, double, double)` ?

arapte: I would recommend to use a for loop as below. 
```
    private static int VALIDATE_COUNT = 3;
    static void validate(Robot robot, int x, int y) {
        for (int i = 0; i < VALIDATE_COUNT; i++) {
            Util.sleep(DELAY_TIME);
            if (x == (int)robot.getMouseX() &&
                y == (int)robot.getMouseY()) {
                break;
            }
        }
        Assertions.assertEquals(x, (int) robot.getMouseX());
        Assertions.assertEquals(y, (int) robot.getMouseY());
    }
```

kevinrushforth: Good idea. As a (minor) suggestion, I might loop "5" times before calling it a failure.

kevinrushforth: > A bigger question is why are we using `int` coordinates? Wouldn't it make the test depend on the scale (and also on the screen resolution and window position)?

I run this frequently on my Windows laptop with 125% and have never seen a problem. Unless we do, this seems unrelated to the bug being fixed.

mstr2: I'd recommend waiting for a few short times (1, 3, 5ms seems good), and then an additional long time just in case the machine has experienced a hiccup at that very moment. There's no harm done in waiting a second for a test to fail, as that would not slow down tests in most cases.

kevinrushforth: @Gopalora if you adopt Michael's addition to Ambarish's suggestion, I recommend leaving `DELAY_TIME` at 1, looping for 5 times, and then having one final test for 500 or 1000 msec (0.5 or 1.0 seconds).

tests/system/src/test/java/test/robot/javafx/scene/MouseLocationOnScreenTest.java
```diff
@@ -118,15 +119,33 @@ public void testMouseLocation() throws Exception {
...
+        if(equalValue == true) {
+            Assertions.assertEquals(x, (int) robot.getMouseX());
+            Assertions.assertEquals(y, (int) robot.getMouseY());
+        }
+        else {
+            //Making Delay and Check after 500 ms.
+            Util.sleep(DELAY_TIME * 500);
+            Assertions.assertEquals(x, (int) robot.getMouseX());
+            Assertions.assertEquals(y, (int) robot.getMouseY());
+        }
```
arapte, line 142: We could avoid repeat of assert statement with this slight changed code:
```
      if (!equalValue) {
            // Delay for 500ms more
            Util.sleep(500);
        }    
        Assertions.assertEquals(x, (int) robot.getMouseX());
        Assertions.assertEquals(y, (int) robot.getMouseY());
```

andy-goryachev-oracle (approved):
complicated, but it works.
I'll re-approve once you fix the whitespace.

kevinrushforth (commented):
The updated test looks good to me. I noted one unnecessary boxed primitive.

tests/system/src/test/java/test/robot/javafx/scene/MouseLocationOnScreenTest.java
```diff
@@ -118,6 +119,21 @@ public void testMouseLocation() throws Exception {
      * returned by robot are same
      */
     static void validate(Robot robot, int x, int y) {
+        Boolean equalValue = false;
```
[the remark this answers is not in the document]
andy-goryachev-oracle: `eagle-eye.png`!

andy-goryachev-oracle (approved):
thank you for the fix and addressing the comments!

kevinrushforth (approved):

PR #1773, opened 2025-04-11
8353599: TabPaneSkin: 'overflowMenuDecorator' property

andy-goryachev-oracle:
This change fixes missing graphic in the `TabPane` overflow menu, using a different approach.  Instead of providing a graphic override property, it now uses a "decorator" pattern to allow for wider customization of the overflow menu (following @Maran23 's suggestion).

### Summary of Changes

- minor `TabPaneSkin` constructor javadoc clarification
- added the `overflowMenuDecorator` property
- changed popup menu to be created on demand
- avoid adding the popup reference to the `TabHeaderSkin.properties` (I think it was done for testing purposes, but I could not find any references to it in the platform code)

### Question for the Reviewers

1. do we want to bind the `MenuItem.text` property if decorator is set?

# Overflow Menu Decorator Property in the TabPaneSkin

## Summary

Add the `overflowMenuDecorator` property in the `TabPaneSkin` class to allow for customization of
the `TabPane` overflow menu, including support for custom graphic.

## Problem

The original issue described impossibility of supporting custom graphic (such as `Path` or `Canvas`)
in the `TabPane` overflow menu.
This happens because
the overflow menu must duplicate the the graphic, and the current implementation is limited to duplicating
either `ImageView` or `Label` with an `ImageView` graphic.

Going beyond the original complaint, it is also impossible to customize the overflow menu, when, for example,
the application requirements call for text-only overflow menu (with tabs that include the graphic), or when
the overflow menu needs to contain different text/graphic, or apply different rules to individual menu items.

## Solution

The solution is to allow the application to set a "decorator" which would be used to customize the menu items
in the overflow menu (the actual menu item instances are of certain type created by the skin).

The `overflowMenuDecorator` property holds the decorator instance.  The default `null` value makes the `TabPane`
work exactly as it works now.

Example:

```java
public class TabPaneGraphicFactoryExample {
    public void example() {
        Tab tab1 = new Tab("Tab1");
        tab1.setGraphic(...);

        Tab tab2 = new Tab("Tab2");
        tab2.setGraphic(...);

        TabPane tabPane = new TabPane();
        tabPane.getTabs().addAll(tab1, tab2);

        TabPaneSkin skin = new TabPaneSkin(tabPane);
        // set overflow menu decorator
        skin.setOverflowMenuDecorator(this::decorateMenu);
        tabPane.setSkin(skin);
    }

    // custom overflow menu decorator
    private void decorateMenu(Tab tab, MenuItem menu) {
        switch (tab.getText()) {
        case "Tab1":
            menu.setGraphic(new Circle(10));
            break;
        case "Tab2":
            menu.setText(null);
            menu.setGraphic(new Canvas(10, 10));
        }
    }
}
```

## Specification

In the `TabPaneSkin` class:

```java
    /**
     * This property allows to customize the overflow menu items.  When this property is not {@code null},
     * the {@link MenuItem} text will get initialized to be the same as the {@link Tab} text and {@code null}
     * graphic prior to invocation of the decorator.
     * <p>
     * When this property is {@code null}, the menu item is initialized with the text and the graphic
     * obtained from the corresponding {@link Tab}.  For the graphic, either an {@link ImageView}
     * or a {@link Label} with an {@link ImageView} will be used.
     * <p>
     * Changing this property while the menu is shown has no effect.
     *
     * @since 28
     * @defaultValue null
     */

    public final ObjectProperty<BiConsumer<Tab, MenuItem>> overflowMenuDecoratorProperty()

    public final BiConsumer<Tab, MenuItem> getOverflowMenuDecorator()

    public final void setOverflowMenuDecorator(BiConsumer<Tab, MenuItem> d)
```

---------

---------
### Progress
- \[ \] Change requires CSR request [JDK-8390003](https://bugs.openjdk.org/browse/JDK-8390003) to be approved

### Issues
 * [JDK-8353599](https://bugs.openjdk.org/browse/JDK-8353599): TabPaneSkin: 'overflowMenuDecorator' property (**Enhancement** - P4)
 * [JDK-8390003](https://bugs.openjdk.org/browse/JDK-8390003): TabPaneSkin: 'overflowMenuDecorator' property (**CSR**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1773/head:pull/1773` \
`$ git checkout pull/1773`

`$ git checkout pull/1773` \
`$ git pull https://git.openjdk.org/jfx.git pull/1773/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1773`

`$ git pr show -t 1773`

</details>

<a href="https://git.openjdk.org/jfx/pull/1773.diff">https://git.openjdk.org/jfx/pull/1773.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1773#issuecomment-2810336717)
</details>


andy-goryachev-oracle:
> then have null mean "I want no overflow menu"

An interesting suggestion.

This would be a much larger change, since we would have to remove the button as well.  Also, the `menuGraphicFactory` seems rather unrelated to the overflow menu button - an RFE with a boolean property might be better choice.

(you can probably tell I am against this idea ;-)

andy-goryachev-oracle:
> default factory should not be a `null` value.

oh, I see.  I think it makes the life easier, since there is no need to publish the pointer to the deficient `DEFAULT_FACTORY`.

This property can be qualified with "override" to make it clearer (though longer).

`menyGraphicFactoryOverrideProperty`

mstr2:
> > default factory should not be a `null` value.
> 
> oh, I see. I think it makes the life easier, since there is no need to publish the pointer to the deficient `DEFAULT_FACTORY`.
> 
> This property can be qualified with "override" to make it clearer (though longer).
> 
> `menyGraphicFactoryOverrideProperty`

On the other hand, if you have a `DEFAULT_FACTORY`, you also have an explicit API element to explain the behavior of the default factory, instead of rolling the documentation into `menuGraphicFactory`.

kevinrushforth:
Reviewers: @arapte @kevinrushforth

andy-goryachev-oracle:
/touch

@arapte @kevinrushforth  please review

arapte:
With this change, For a same tab the graphic in TabHeader and graphic in overflow menu can be different.
Here is a screenshot from modified version of the sample program.
Screenshot 2025-09-24 at 18 04 50

Of course it is in programmer's control.
But, can we fix on a behavior that if a graphic is set on a tab, then the same graphic would be used in overflow menu and a graphic from menuGraphicFactory will be neglected ?

andy-goryachev-oracle:
> But, can we fix on a behavior that if a graphic is set on a tab, then the same graphic would be used in overflow menu and a graphic from menuGraphicFactory will be neglected ?

Good question!

We can't, if the graphic set on a tab is some sort of a custom `Node` and not the usual `ImageView` or `Label`.  In this case, we can't easily clone that custom graphic such that it appears in the tab **and** in the overflow menu.

Ideally, the solution would be to replace setting the graphic on a tab with setting a Supplier<Node>.  This way the graphic can be supplied multiple times as needed, but it would be a breaking change and something that is not needed in 99.9% of the cases.

mstr2:
> > But, can we fix on a behavior that if a graphic is set on a tab, then the same graphic would be used in overflow menu and a graphic from menuGraphicFactory will be neglected ?
> 
> Good question!
> 
> We can't, if the graphic set on a tab is some sort of a custom `Node` and not the usual `ImageView` or `Label`. In this case, we can't easily clone that custom graphic such that it appears in the tab **and** in the overflow menu.
> 
> Ideally, the solution would be to replace setting the graphic on a tab with setting a Supplier. This way the graphic can be supplied multiple times as needed, but it would be a breaking change and something that is not needed in 99.9% of the cases.

WPF has [VisualBrush](https://learn.microsoft.com/en-us/dotnet/api/system.windows.media.visualbrush?view=windowsdesktop-10.0), which is like a JavaFX `Paint` that, instead of filling a shape with a color, fills it with the rendered representation of another node. Maybe there's something similar that could be done for JavaFX, which would make it easier to solve problems where a node in one part of the scene graph should be projected into another part of the scene graph.

andy-goryachev-oracle:
> rendered representation of another node

I think this idea might be counter-productive: the app may want to style the overflow icons differently (different size, for example, or using different CSS selectors).  The original developers just did not consider scenarios that go beyond commonly used labels or images.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
```diff
@@ -256,6 +257,45 @@ public TabPaneSkin(TabPane control) {
...
+     * <p>
+     * When this property is {@code null}, the menu provides only the basic graphic copied from the corresponding
+     * {@link Tab} - either an {@link ImageView} or a {@link Label} with an {@link ImageView} as its graphic.
+     * <p>
+     * Changing this property while the menu is shown has no effect.
+     *
+     * @since 27
+     * @defaultValue null
+     */
+    private ObjectProperty<Function<Tab, Node>> menuGraphicFactory;
```
Maran23: In JavaFX, `Callback` is always used for this kind of factories. In my opinion, we should be consistent with that

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
```diff
@@ -485,27 +525,28 @@ private static int getRotation(Side pos) {
...
+            return f.apply(t);
         }
-        if (n instanceof ImageView) {
-            ImageView iv = (ImageView) n;
+
+        Node n = t.getGraphic();
+        return extractGraphic(n);
+    }
+
+    private Node extractGraphic(Node n) {
```
Maran23: Should be renamed to `createGraphicCopy` or something like that. This is not an extract operation, but rather a try to copy the existing graphic

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
```diff
@@ -485,27 +525,28 @@ private static int getRotation(Side pos) {
...
     }
 
-    /**
-     * VERY HACKY - this lets us 'duplicate' Label and ImageView nodes to be used in a
-     * Tab and the tabs menu at the same time.
-     */
-    private static Node clone(Node n) {
-        if (n == null) {
-            return null;
+    private Node prepareGraphic(Tab t) {
```
Maran23: Should also be around `createGraphic`. Does not prepare anything, but creates a graphic whether by factory or by trying to create a copy.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
```diff
@@ -1953,35 +1994,34 @@ private void showPopupMenu() {
...
 
         public Tab getTab() {
             return tab;
         }
 
+        // is this really necessary?
         public void dispose() {
             textProperty().unbind();
-            tab.disableProperty().removeListener(weakDisableListener);
+            disableProperty().unbind();
```
Maran23, line 2038: This seems unrelated.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
```diff
@@ -1953,35 +1994,34 @@ private void showPopupMenu() {
...
-            tab.disableProperty().addListener(weakDisableListener);
+            disableProperty().bind(tab.disableProperty());
             textProperty().bind(tab.textProperty());
         }
 
         public Tab getTab() {
             return tab;
         }
 
+        // is this really necessary?
```
Maran23: Pretty sure it is, otherwise we probably get a memory leak. I don't really like spreading random comments in the codebase which will be there for the next 20 years without any benefit.

I would recommend checking if this is needed with a Profiler / `JMemoryBuddy`. But as far as I can see, this is indeed needed.

modules/javafx.controls/src/test/java/test/javafx/scene/control/TabPaneTest.java
```diff
@@ -1338,4 +1342,49 @@ private void scrollTabPane(Side side, double deltaX, double deltaY) {
...
+        assertNotNull(menu);
+        assertEquals(3, menu.getItems().size());
+        return menu;
+    }
+
+    @Test
+    public void menuGraphicFactory() {
+        ContextMenu menu = setupMenuGraphicFactory();
+        for (MenuItem mi : menu.getItems()) {
+            assertTrue(mi.getGraphic() instanceof Path);
```
Maran23: It would be good to check if the right graphic was created. Now we just check if it is `Path`. 
If there is a bug where the wrong graphic for a `Tab` is used (thinnk about `Tab` 1 has the graphic of `Tab` 2), we would not catch that regression.

I recommend something like creating a `Label` with the `Tab` text, and check that:
- `Tab` with text "1" creates a graphic `Label` with text "1"
- `Tab` with text "2" creates a graphic `Label` with text "2"

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
```diff
@@ -1509,7 +1550,6 @@ public void handle(MouseEvent me) {
             });
 
             getProperties().put(Tab.class, tab);
-            getProperties().put(ContextMenu.class, tab.getContextMenu());
```
Maran23, line 1523: Seems unrelated - does that have any consequences?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
```diff
@@ -162,10 +163,10 @@ private enum TabAnimationState {
 
     /**
      * Creates a new TabPaneSkin instance, installing the necessary child
-     * nodes into the Control {@link Control#getChildren() children} list, as
+     * nodes into the Control {@link TabPane#getChildren() children} list, as
```
Maran23: Seems unrelated, and `Control` is fine, so why this change?

Maran23 (changes requested):

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
andy-goryachev-oracle: related: I've replaced `weakDisableListener` L1967 with binding.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
andy-goryachev-oracle: the popup logic has been moved to L1985

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
```diff
@@ -162,10 +163,10 @@ private enum TabAnimationState {
 
     /**
      * Creates a new TabPaneSkin instance, installing the necessary child
-     * nodes into the Control {@link Control#getChildren() children} list, as
+     * nodes into the Control {@link javafx.scene.control.Control#getChildren() children} list, as
```
Maran23: I would vote +1 to change this in all Skins, but obviously very minor

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
Maran23: ok will do a deeper check

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
andy-goryachev-oracle: I think I removed the `Control` import and it broke javadoc, which I only saw in GHA.

Actually, it should be Parent.getChildren() , and maybe say 
`nodes into the control's {@link javafx.scene.Parent#getChildren() children} list`

Maran23: For me, this is always red and inaccessible, with `Control`, `Parent`, whatever it is

andy-goryachev-oracle: this is weird, navigates to `Parent.getChildren()` in eclipse.

what's more important, generates correct javadoc -

Screenshot 2026-03-24 at 15 59 52

did you rebuild?

andy-goryachev-oracle:
🎂 Happy 1st birthday, PR #1773!

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
```diff
@@ -256,6 +257,45 @@ public TabPaneSkin(TabPane control) {
...
+            };
+        }
+        return menuGraphicFactory;
+    }
+
+    public final Callback<Tab,Node> getMenuGraphicFactory() {
+        return menuGraphicFactory == null ? null : menuGraphicFactory.get();
+    }
+
+    public final void setMenuGraphicFactory(Callback<Tab,Node> f) {
```
Maran23: Minor: Space is missing between `<Tab,Node>`

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
```diff
@@ -2347,7 +2386,7 @@ private void stopAnim(Animation anim) {
 
     // For testing purpose.
     ContextMenu test_getTabsMenu() {
-        return tabHeaderArea.controlButtons.popup;
+        return tabHeaderArea.controlButtons.test_getTabsMenu();
```
Maran23: I wonder if we could do instead:

```
        if (tabHeaderArea.controlButtons.popup == null) {
            tabHeaderArea.controlButtons.setupPopupMenu();
        }
        return tabHeaderArea.controlButtons.popup;
```

That would mean we'd have one less method that's only used in tests. Would be my preferred solution, though.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
Maran23: This seems fine. `showControlButtons()` could probably not call `setupPopupMenu` eagerly and instead let `showPopupMenu` do the work while here.

Maran23 (commented):
Minor comments, looks good otherwise to me.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
Maran23: Yes, my IDE still says: `Symbol 'getChildren()' is inaccessible from here`. Is that link clickable in Javadoc? If that works, I'm fine with the change.

mstr2:
I agree with the shape of the `menuGraphicFactory` property, but I disagree that this should be thought of as an "override" rather than what it says it is: It's a factory that takes in a tab and produces an overflow menu graphic, just like the default factory does. There's no material difference between the default factory and any other custom factory.

This is why, when you give the overflow-menu factory an API, you should also give the default factory an API element. Treating `null` as a toggle for an actual factory implementation is just not a good choice, and we shouldn't do that.

Maran23:
> This is why, when you give the overflow-menu factory an API, you should also give the default factory an API element. Treating `null` as a toggle for an actual factory implementation is just not a good choice, and we shouldn't do that.

Makes sense and I agree. This should follow the pattern we already have in e.g. `TableColumn`:
https://github.com/openjdk/jfx/blob/697a534bf494d82bf8e0ecacfebaba003dcc6c7a/modules/javafx.controls/src/main/java/javafx/scene/control/TableColumn.java#L202-L211

Note:
There is also this pattern we have in JavaFX:
https://github.com/openjdk/jfx/blob/697a534bf494d82bf8e0ecacfebaba003dcc6c7a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkin.java#L208-L216

which is similar to what Andy did. But a default factory in our case makes more sense.

andy-goryachev-oracle:
Thank you all for the comments!

As far as using a `null` to indicate the default: I would very much prefer that.
The reason is to allow an easy revert: set the factory, and if not needed anymore, set back to `null`.

Otherwise, the revert gets complicated: either we need to get and save the object somewhere, or declare a public constant or a method to provide one - both with dubious value.  

But the main reason is that there is absolutely no need to give the application access to a pointer to the default factory, none whatsoever.  Or am I missing something?

Maran23:
> As far as using a `null` to indicate the default: I would very much prefer that. The reason is to allow an easy revert: set the factory, and if not needed anymore, set back to `null`.

I wonder if `null` could also indicate: I don't want tab graphics, so I set the factory to `null`.

I think developers usually would not change the tab menu graphic factory just to revert it back at one point. If that is needed and we have a default tab menu graphic factory, we could just save it in case we really need to restore it (which I don't think will happen).

So:

```
var defaultFactory = getMenuGraphicFactory();
```

To restore it later.

andy-goryachev-oracle:
> I wonder if `null` could also indicate: I don't want tab graphics, so I set the factory to `null`.

no, the javadoc says it provides the default functionality

> I think developers usually would not change the tab menu graphic factory just to revert it back at one point.

I wouldn't be so sure (maybe in this particular corner case).  Several times I had to deal with default values (often some obscure non-public classes to boot) when the UI calls for runtime selection via e.g. "Preferences" menu.  `null` is convenient in this way, and there is nothing the default factory itself can provide...

mstr2:
We seem to disagree on a fundamental point, not on whether `null` is a useful default. For you, the default behavior of producing a `Node` for a given `Tab` is not a factory, and only the user-provided function is one. I think this distinction is not justifiable, as there is no difference in principle between the default factory and a user-provided factory. Both take in a `Tab` and produce a `Node` that can be used as a graphical representation for the tab. It doesn't matter if the default factory is limited in functionality, the purpose remains the same.

If you are introducing an API for a concept that's already there, the default behavior should be represented using the new API, and it should be given a name and corresponding documentation (which in our case means a public final constant). To make it explicitly clear: I think that `null` is not a useful name for a factory, it's the absence of one; and treating `null` as a sentinel for an actual, real, factory implementation is not a good API design. Whether a `null` sentinel makes it "easier" to revert the factory back to its default is completely besides the point (and it's not even true, a named constant is just as easy to use as a `null` sentinel). 

It's also a bit of a distraction, because it is exceedingly rare that developers revert control configuration; it is so rare that it should definitely not drive API design. I've read all of the supporting materials and discussions that you provided, and nowhere did anyone (including yourself) say that _reverting_ control configuration is what motivates this proposal. This point just randomly appeared in the discussion a few days ago.

The only justifiable API design that would, in my option, warrant using a `null` sentinel is if you named it `menuGraphicFactoryOverride`: in this case, the null value would actually mean that no override is present. But I would not be in favor of adding such a property, as the better alternative is to just give a name to the factory that produces a node for a given tab.

Maran23:
I tend to agree with Michael. 
A descriptive name/default value is easier to understand and is also consistent with `TableColumn`. 
Another nice advantage: As in `TableColumn`, we can add javadoc explaining the default factory behavior and also the limitations it has. Like in `TableColumn`:

https://github.com/openjdk/jfx/blob/697a534bf494d82bf8e0ecacfebaba003dcc6c7a/modules/javafx.controls/src/main/java/javafx/scene/control/TableColumn.java#L202-L211

Also, I can think of setting this property to null to simply say: I don't want tab menu graphics at all.

andy-goryachev-oracle:
`menuGraphicFactoryOverride` might be a better name, I agree - that what it is designed to be, an override.

There is one possible use case which favors a factory over an override: when the application is too lazy to implement the whole factory, wanting instead to handle the custom tabs, delegating the other kinds to the default factory.  Not the use case I initially envisioned, but it is a possibility that suggests exposing a `DEFAULT_MENU_GRAPHIC_FACTORY` might be a better approach. 

One more point I wanted to bring up is the use of `Callback` instead of a dedicated interface - we had this problem when adding table column resize policies.  A `Callback` cannot be evolved, while a dedicated interface can, using default methods.  I don't think this should be the case here since the functionality is very limited, but I want to bring it to your attention.

Given the popular feedback, I'll update the PR to expose the `DEFAULT_MENU_GRAPHIC_FACTORY`.

mstr2:
> `menuGraphicFactoryOverride` might be a better name, I agree - that what it is designed to be, an override.
> 
> There is one possible use case which favors a factory over an override: when the application is too lazy to implement the whole factory, wanting instead to handle the custom tabs, delegating the other kinds to the default factory. Not the use case I initially envisioned, but it is a possibility that suggests exposing a `DEFAULT_MENU_GRAPHIC_FACTORY` might be a better approach.

I didn't think of this use case, but it seems reasonably useful for developers who just want to keep the default factory implementation, but just add one more thing to it.

> One more point I wanted to bring up is the use of `Callback` instead of a dedicated interface - we had this problem when adding table column resize policies. A `Callback` cannot be evolved, while a dedicated interface can, using default methods. I don't think this should be the case here since the functionality is very limited, but I want to bring it to your attention.

While `Callback` can't be evolved, it can conceivably be extended by another interface. But I agree that this probably isn't relevant for this use case.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
andy-goryachev-oracle: I'll change it to 

```
     * Creates a new TabPaneSkin instance, installing the necessary child
     * nodes into the Control {@link javafx.scene.Parent#getChildren() children} list, as
     * well as the necessary input mappings for handling key, mouse, etc events.
```

which still works correctly in Eclipse and in the resulting javadoc.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
```diff
@@ -283,13 +291,25 @@ public Object getBean() {
...
+
+                @Override
+                protected void invalidated() {
+                    var v = get();
+                    if (v == null) {
+                        if (isBound()) {
+                            unbind();
+                        }
+                        set(lastValue);
+                        throw new NullPointerException("cannot set factory to null");
```
Maran23: I think we should allow `null` and just return a `null` graphic then. No need to do any handling here, less logic. 
Just:
```
    private Node createGraphic(Tab t) {
        Callback<Tab, Node> f = getMenuGraphicFactory();
        if (f == null) {
            return null;
        }
        return f.call(t);
    }
```

And this is also aligned with other components in JavaFX, like e.g. `null` SelectionModel.

andy-goryachev-oracle: No, this is **not** a SelectionModel.  This is a (missing) part of the basic functionality, inability to handle a custom graphic.  We typically don't notice because it's a corner case, but a corner case where there is no workaround using the public API.

I don't quite get the objections here, and I don't particularly like exposing half-baked internal implementations in the form of the default factory.  I also don't completely buy the argument about "set to null if the application does not want graphic" because when the application does not want a graphic, it does not set a graphic on the tab, right?

What we ended up with after the last commit is too many public APIs for a rare corner case, a negative gain.  I really don't understand your logic here, why make it more complicated?

Maran23: This PR is about customizing the menu graphic, right?
So it could very well be a usecase that the menu should have no graphics, but the tabs should have. 
In fact, this is what happens now as well, as the graphic cloning is veryyy limited to `ImageView` and `Label` only. 
If you e.g. use shapes (which I did in the past), no graphic is shown in the menu. 
Now, If I mix Shapes and `ImageView`s, some menu items have a graphic while others have not - so I might want to extend it or completely remove it. 

Or some Tabs have a graphic (Like the 'Home' Tab) while all others have not - then it looks very weird that just one tab has a graphic in the menu (and the space is reserved) while the remaining have not. This is actually something from my real world experience - therefore something I have interest in.
 
My other point is that I really want us to stick to the existing model that is used in JavaFX. 
That is: Null is allowed and will often just kind of 'disable' the functionality. Which in this case makes sense. 
And IMO, we are making things more complicated by handling `null`, overwriting `invalidated` and even tracking the `lastValue` instead of just writing `f == null ? null : f.call(t)` and call it a day.

andy-goryachev-oracle: fx uses multiple metaphors, there is no single prescribed manner, I think.  For example, a null `HBox.alignment` is functionally equivalent to `Pos.TOP_LEFT` (HBox:336)

The design principle that I would consider more important is `frequent use cases should be easy, everything else should be possible`.  In your case, all you need with this PR is to install a factory that produces a `null` `Node`, and this is not what I would think is the most frequent case, which is just to show the same icons that the application set on tabs.

Also, I really dislike the DEFAULT_MENU_GRAPHIC_FACTORY idea - exposing something internal for no purpose whatsoever.  Maybe I should just close this PR.

mstr2: > Also, I really dislike the DEFAULT_MENU_GRAPHIC_FACTORY idea - exposing something internal for no purpose whatsoever. Maybe I should just close this PR.

It isn't internal, it's user-facing behavior and therefore not internal by definition. Your own proposal adds documentation for the default behavior, which we wouldn't do if it was an implementation detail.

andy-goryachev-oracle: It _was_ internal in the original proposal.  In fact, we `could` improve the default behavior by adding, let's say, support for arbitrary `Path`s, but with the public factory it's set in stone.

I think I should revert the public factory and go back to `null` signifying the default behavior, whatever that is.

mstr2: > It _was_ internal in the original proposal. In fact, we `could` improve the default behavior by adding, let's say, support for arbitrary `Path`s, but with the public factory it's set in stone.
> 
> I think I should revert the public factory and go back to `null` signifying the default behavior, whatever that is.

This is the text of your original proposal:

```
    /**
     * ...
     * When this property is {@code null}, the menu provides only the basic graphic copied from the corresponding
     * {@link Tab} - either an {@link ImageView} or a {@link Label} with an {@link ImageView} as its graphic.
     * ...
     */
    public final ObjectProperty<Function<Tab, Node>> menuGraphicFactoryProperty() {
```

An observable behavior that is described in documentation can _never_ be an internal implementation detail; it is part of the public contract of the class. Whether it has its own API element (like a constant field) isn't relevant.

andy-goryachev-oracle: Oh, I agree with the sentiment: it's great when the behavior is properly documented.

But if we can get back to the initial sticky wicket, the use of `null` to signify the default behavior, I think using `null` is a cleaner solution in this particular corner case.  In most of the use cases, the default is acceptable and no one cares.  But if not, there is a way to override it.  That's it.

I am going to revert the factory and call the property `menuGraphicFactoryOverrideProperty` to make clear it's an override.  How does that sound?

Maran23:
Sorry for the delay. 

Today I set up the project I've worked on in the past (but not anymore), and where this feature would have been really helpful (back when I implemented this). 

In this case, tiles open tabs. 
Only the Home tab has an icon but no name (and shows all the tiles).
When clicking on a Tile, a Tab opens but they never have an icon, only text. 

And that's where I had the idea to simply set the menu graphic to `null` so there would be no icon at all, saving space entirely. At the same time, however, it would be useful to be able to customize the text there as well so that 'Home' is shown there while the Tab has no text.

image

Overflow:

image

andy-goryachev-oracle:
> At the same time, however, it would be useful to be able to customize the text there as well so that 'Home' is shown there while the Tab has no text.

so this **_is_** possible with the new API, correct?

edit:
by the "existing API", I meant everything that's available in JavaFX, but not the menu graphic override property.  For example, you could set the "Home" text on the tab, and, assuming you disable reordering of the tabs, you could set max size of the first normal tab to 0 via CSS, or something similar.

Otherwise, the property needs to be redesigned to create the menu item instead of the simple graphic, and it gets more involved for such a corner case.  Still possible, I suppose.

Is that what you meant?

Maran23:
Yes.

I was referring to the idea of removing graphics completely by setting `null` (kind of not wasting any CPU cycle at all for that)
and that at least in that usecase, I would also need to customize the text. So I can see a usecase for that. (Without relying on hacks or try to do sizing stuff)

That made me wonder if we may want to have some kind of `Tab` decorator or something. Or maybe we should instead allow full customization, similar as we do for the Table Column Menu? That would be the most flexible, as you can just create whatever `Menu` you want. In my Table Column Menu example, this is especially helpful if you want to open a customization dialog instead. I saw that quite often, where people can have a whole table config that is saved to the DB.

Tabs are usually not that complex - but this way we allow my usecase and potentially every other one. And you already made the right step by creating the Popup more lazily - which is required to not create unnecessary objects.

andy-goryachev-oracle:
I agree, allowing to fully customize the context menu is probably a better idea anyway.

The problem with context menus in javafx is that they made the same mistake many people do in Swing: they think it's something that should be created early, or created and set to a property.  A much better approach, in my opinion, is to create context menus upon request.  

In other words, instead of the `contextMenu` property in `Control`, `Tab`, `TableColumnBase` it should have been a `contextMenuSupplier` property.

Of course it is possible to use the supplier, at the price of a little dance, see for instance
https://github.com/andy-goryachev/MP3Player/blob/8b0ff12460e19850b783b961f214eacf5e1cdaf8/src/goryachev/fx/FX.java#L1251

I am going to set this PR back to Draft state to explore the alternative you suggested.  Thanks!

Maran23:
> In other words, instead of the `contextMenu` property in `Control`, `Tab`, `TableColumnBase` it should have been a `contextMenuSupplier` property.

This is a valid point. 
I thought the same thing when looking at `Tooltip` too. 

We create a `Tooltip` and it might never ever be visible. Although in the `Tooltip` case, we need the show delay to know how long we need to wait until we can show it. But only when the mouse actually moved over the `Node`. So if a `Node` was never mouse hovered (or the user navigates only with the keyboard), the `Tooltip`, similar to the `ContextMenu` is just created and never used.

Maran23:
One thing I wonder: The decorator vs. the possibility to customize the whole overflow menu, like you can do for the table (menu). So a `protected` method where you can on your own iterate through all tabs and do what you want to do.

andy-goryachev-oracle:
> customize the whole overflow menu

I agree with the sentiment - there should be a way to override things that need further customization, but in this case it's probably not necessary.  On top of that, the skin creates its own `TabMenuItem`, which limits our choices.

PR #1776, opened 2025-04-13
8354478: Improve StageStyle documentation

tsayao:
Improve StageStyle Documentation

- Update `StageStyle.UTILITY`:
Clarified that UTILITY stages may impose platform-specific restrictions on window states, such as preventing maximize, minimize (iconify), and fullscreen operations.

- Update `StageStyle.UNDECORATED`:
Improved the description to clarify that although UNDECORATED removes standard window decorations (title bar, borders, and controls), window operations like minimize, maximize, and fullscreen may still be allowed programmatically or via platform-specific functions such as key shortcuts or menu options.

- Remove mention of solid white background:
This seems to be not true anymore, even withou a `Scene` (or there's a bug)

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/stage/StageStyle.java
+++ b/modules/javafx.graphics/src/main/java/javafx/stage/StageStyle.java
@@ -32,12 +32,16 @@ package javafx.stage;
 public enum StageStyle {
 
     /**
-     * Defines a normal {@code Stage} style with a solid white background and platform decorations.
+     * Defines a normal {@code Stage} style with a solid background and platform decorations.
      */
     DECORATED,
 
     /**
-     * Defines a {@code Stage} style with a solid white background and no decorations.
+     * Defines a {@code Stage} style with no window decorations, such as a title bar,
+     * borders, or window controls.
+     * This style allows window operations such as resize, minimize, maximize
+     * and fullscreen to be either programmatically controlled or achieved through
+     * platform-specific functions, such as key shortcuts or menu options.
      */
     UNDECORATED,
 
@@ -51,8 +55,11 @@ public enum StageStyle {
     TRANSPARENT,
 
     /**
-     * Defines a {@code Stage} style with a solid white background and minimal
-     * platform decorations used for a utility window.
+     * Defines a lightweight {@code Stage} with minimal decorations, intended for
+     * supporting tasks such as tool palettes.
+     * Utility stages may restrict window operations like maximize, minimize,
+     * and fullscreen depending on the platform. They are designed to float above
+     * primary windows without acting as a main application stage.
      */
     UTILITY,
 


modules/javafx.graphics/src/main/java/javafx/stage/StageStyle.java
```diff
@@ -32,12 +32,16 @@
...
      */
     DECORATED,
 
     /**
-     * Defines a {@code Stage} style with a solid white background and no decorations.
+     * Defines a {@code Stage} style with no window decorations, such as a title bar,
+     * borders, or window controls.
+     * This style allows window operations such as resize, minimize, maximize
+     * and fullscreen to be either programmatically controlled or achieved through
+     * platform-specific functions, such as key shortcuts or menu options.
```
hjohn, line 44: It looks like you intend the 2nd part to be a new paragraph:
```suggestion
     * Defines a {@code Stage} style with no window decorations, such as a title bar,
     * borders, or window controls.
     * <p>
     * This style allows window operations such as resize, minimize, maximize
     * and fullscreen to be either programmatically controlled or achieved through
     * platform-specific functions, such as key shortcuts or menu options.
```

andy-goryachev-oracle: +1

modules/javafx.graphics/src/main/java/javafx/stage/StageStyle.java
```diff
@@ -32,12 +32,16 @@
 public enum StageStyle {
 
     /**
-     * Defines a normal {@code Stage} style with a solid white background and platform decorations.
+     * Defines a normal {@code Stage} style with a solid background and platform decorations.
```
andy-goryachev-oracle, line 35: Should we say the background color is platform-specific, or perhaps identify why it is different?

modules/javafx.graphics/src/main/java/javafx/stage/StageStyle.java
```diff
@@ -51,8 +55,11 @@ public enum StageStyle {
     TRANSPARENT,
 
     /**
-     * Defines a {@code Stage} style with a solid white background and minimal
-     * platform decorations used for a utility window.
+     * Defines a lightweight {@code Stage} with minimal decorations, intended for
```
nlisker, line 58: Is there a reason you removed use the word "style" here? The other ones use it.

modules/javafx.graphics/src/main/java/javafx/stage/StageStyle.java
```diff
@@ -32,12 +32,16 @@
...
     /**
-     * Defines a normal {@code Stage} style with a solid white background and platform decorations.
+     * Defines a normal {@code Stage} style with a solid background and platform decorations.
      */
     DECORATED,
 
     /**
-     * Defines a {@code Stage} style with a solid white background and no decorations.
+     * Defines a {@code Stage} style with no window decorations, such as a title bar,
+     * borders, or window controls.
```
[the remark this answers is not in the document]
tsayao: I restored it, but I'm still curious about when the white background should actually be visible. I noticed that on Linux, if a stage is shown without a scene, it behaves oddly—it doesn't render because there's nothing being painted. To fix this, I added a white background as part of the "Linux glass fixes" package I'm working on. I also observed that when a scene is present, a white background gets set automatically. But under what conditions is it actually visible?

modules/javafx.graphics/src/main/java/javafx/stage/StageStyle.java
tsayao: I restored it, I think it's a Linux glass specific bug.

modules/javafx.graphics/src/main/java/javafx/stage/StageStyle.java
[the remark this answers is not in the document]
tsayao: A scene with a TRANSPARENT Parent is one scenario.

modules/javafx.graphics/src/main/java/javafx/stage/StageStyle.java
andy-goryachev-oracle: If it's a bug, could you please create a JBS ticket?

andy-goryachev-oracle (approved):
thank you

modules/javafx.graphics/src/main/java/javafx/stage/StageStyle.java
tsayao: Sure. I'll look into it.

hjohn: To be honest, I don't understand why we specify this at all.  "white" surely won't be the default when running on dark theme... I'm pretty sure on Windows it isn't even white, it's some light gray.

tsayao: When there's no scene attached, it seems to fall back to the default behavior defined by the OS. However, once a scene is set, a white background is sent, although this gets overridden by the root node of the scene.

So I’d say this behavior isn’t really tied to Stage state or StageStyle, but rather to the presence and characteristics of the Scene itself.

lukostyra (approved):

PR #1779, opened 2025-04-16
8354795: DialogPane show details button wipes out base style class "hyperlink"

hjohn:
The "show details" hyperlink button in an alert dialog that has an expandable detail area wipes out its base style class by overwriting all styles. This means styling in modena.css that targets `.hyperlink` is no longer applied, like the default text fill colors.

The culprit is this code in DialogPane:

        InvalidationListener expandedListener = o -> {
            final boolean isExpanded = isExpanded();
            detailsButton.setText(isExpanded ? lessText : moreText);
            detailsButton.getStyleClass().setAll("details-button", (isExpanded ? "less" : "more")); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
        };

Here it uses `setAll` to set styles, wiping out the default `.hyperlink` style from "Hyperlink detailsButton = new HyperLink()"

---------
### Progress

### Issue
 * [JDK-8354795](https://bugs.openjdk.org/browse/JDK-8354795): DialogPane show details button wipes out base style class "hyperlink" (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1779/head:pull/1779` \
`$ git checkout pull/1779`

`$ git checkout pull/1779` \
`$ git pull https://git.openjdk.org/jfx.git pull/1779/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1779`

`$ git pr show -t 1779`

</details>

<a href="https://git.openjdk.org/jfx/pull/1779.diff">https://git.openjdk.org/jfx/pull/1779.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1779#issuecomment-2808877086)
</details>

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/DialogPane.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/DialogPane.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2014, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -34,7 +34,6 @@ import java.util.WeakHashMap;
 import com.sun.javafx.scene.control.skin.Utils;
 import com.sun.javafx.scene.control.skin.resources.ControlResources;
 import javafx.beans.DefaultProperty;
-import javafx.beans.InvalidationListener;
 import javafx.beans.property.BooleanProperty;
 import javafx.beans.property.ObjectProperty;
 import javafx.beans.property.SimpleBooleanProperty;
@@ -818,16 +817,15 @@ public class DialogPane extends Pane {
         final Hyperlink detailsButton = new Hyperlink();
         final String moreText = ControlResources.getString("Dialog.detail.button.more"); //$NON-NLS-1$
         final String lessText = ControlResources.getString("Dialog.detail.button.less"); //$NON-NLS-1$
+        final ObservableList<String> styleClasses = detailsButton.getStyleClass();
 
-        InvalidationListener expandedListener = o -> {
-            final boolean isExpanded = isExpanded();
-            detailsButton.setText(isExpanded ? lessText : moreText);
-            detailsButton.getStyleClass().setAll("details-button", (isExpanded ? "less" : "more")); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
-        };
+        styleClasses.add("details-button");  //$NON-NLS-1$
 
-        // we call the listener immediately to ensure the state is correct at start up
-        expandedListener.invalidated(null);
-        expandedProperty().addListener(expandedListener);
+        expandedProperty().subscribe(expanded -> {
+            styleClasses.remove(expanded ? "more" : "less");  //$NON-NLS-1$ //$NON-NLS-2$
+            styleClasses.add(expanded ? "less" : "more");  //$NON-NLS-1$ //$NON-NLS-2$
+            detailsButton.setText(expanded ? lessText : moreText);
+        });
 
         detailsButton.setOnAction(ae -> setExpanded(!isExpanded()));
         return detailsButton;


modules/javafx.controls/src/main/java/javafx/scene/control/DialogPane.java
```diff
@@ -818,16 +817,15 @@ protected Node createDetailsButton() {
...
         final String moreText = ControlResources.getString("Dialog.detail.button.more"); //$NON-NLS-1$
         final String lessText = ControlResources.getString("Dialog.detail.button.less"); //$NON-NLS-1$
+        final ObservableList<String> styleClasses = detailsButton.getStyleClass();
 
-        InvalidationListener expandedListener = o -> {
-            final boolean isExpanded = isExpanded();
-            detailsButton.setText(isExpanded ? lessText : moreText);
-            detailsButton.getStyleClass().setAll("details-button", (isExpanded ? "less" : "more")); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
-        };
+        styleClasses.add("details-button");  //$NON-NLS-1$
```
andy-goryachev-oracle, line 822: do we use these `//$NON-NLS-1$` in jfx?

andy-goryachev-oracle:
An SCCE would have been nice, to help with the review/testing.

modules/javafx.controls/src/main/java/javafx/scene/control/DialogPane.java
hjohn: I do not know if we do, as there is also non-public code that I don't have access to.  I've kept them as-is to make sure the change is focused and doesn't break anything else.

If I had to guess then probably not, if you can ask around and find out I'll be happy to remove them.

hjohn:
> An SCCE would have been nice, to help with the review/testing.
> 
> edit: alternatively, I'll add the (missing) Dialog/DialogPane page to the monkey tester. 🐒🔬

If you still need one, I can provide a bit of code.  It basically is just changing the hyperlink style, and then seeing it doesn't affect the Alert dialog when it has an extended area.  After the change, it should affect it.  I did include a screenshot -- I may find more of these bugs as I continue developing with modena.css with dark theme colors :)

andy-goryachev-oracle:
> If you still need one, I can provide a bit of code.

Thanks, that's ok, I would rather add the Dialog page to the MT, as it will allow for more extensive testing.

modules/javafx.controls/src/main/java/javafx/scene/control/DialogPane.java
andy-goryachev-oracle: I'll double check and will let you know tomorrow.

andy-goryachev-oracle:
I've added the Dialog page to the monkey tester

Looking at the stock behavior, I don't really see the problem.  The .details-button should not _look_ a hyperlink (despite being implemented using Hyperlink control), so the code seems to be correct in that it replaces the hyperlink style with something else.

What we probably need for the dark mode is to update the stylesheet as you described in the ticket.

What do you think?

andy-goryachev-oracle:
Screenshot 2025-04-17 at 08 59 10

hjohn:
> I've added the Dialog page to the monkey tester https://github.com/andy-goryachev-oracle/MonkeyTest
> 
> Looking at the stock behavior, I don't really see the problem. The .details-button should not _look_ a hyperlink (despite being implemented using Hyperlink control), so the code seems to be correct in that it replaces the hyperlink style with something else.
> 
> What we probably need for the dark mode is to update the stylesheet as you described in the ticket.
> 
> What do you think?

I think it _should_ look like a hyperlink, as the choices of potential clickable controls that users will recognize as such is limited; it can be either a button, checkbox or hyperlink.  You can't just define something new and expect users to understand it at a glance; and why even do so for an alert dialog?  This is clearly intended to mimic other standard alert dialogs that use the same kind of button to display details.

There are so many things dependent on the base style in modena, not having a base style is highly unusual and in this case IMHO a clear mistake in how the styles are being applied.  It just wasn't caught in testing as the default color matches close enough, but becomes rather obvious when changing the base colors significantly.

andy-goryachev-oracle:
> I think it _should_ look like a hyperlink

It might look like a hyperlink, but it has a different semantics (semantics of a hyperlink is to navigate to some other place).   When application wants to change the hyperlink styling, we don't want this to propagate to unrelated controls such as dialog buttons, or we get this:

https://media2.giphy.com/media/v1.Y2lkPTc5MGI3NjExN29yNmd3dGRlNW0wNXMxaWpmM25pdTUyOGZrcmV1dm9oemFkaWZucSZlcD12MV9pbnRlcm5hbF9naWZfYnlfaWQmY3Q9Zw/13FrpeVH09Zrb2/giphy.gif

hjohn:
In Windows, and also our color chooser, hyperlinks are used with some regularity to avoid a button where it would look out of place -- in small windows, buttons often are seen as a "closing" action, which show details clearly is not intended to be.  The Windows copy dialog for example has a show details text that's clickable, and when keyboard focused displays the classic hyperlink dashed rectangular border.

I've looked in FX code, and wiping out the base style is AFAICS not done anywhere.  There are of course many cases of `getStyleClass().setAll( ... )` but all of the ones I checked do this on non-Controls (StackPane most often).  The big difference here is that such containers like StackPane don't have a default style so there is nothing that would be overwritten.  Only controls have these, and wiping it out is, and I'm repeating myself, highly unusual as it will kill many lines of default styling in modena leaving you to have to repeat all those lines.  That didn't happen, and neither was this code commented as "intentionally removing the default style class for this control".

If this indeed was the intent, then it would have made **much** more sense to not wipe out the base style, and only make adjustments specifically for hyperlinks that are also styled with `.details-button`.

mstr2:
It seems a bit unusual to remove the default style class of a control. If you use a `Hyperlink` control, you also want the result to be hyperlinky and inherit all the hyperlinky styles. This is probably the reason why this control was used here in the first place.

If you don't want your control to be hyperlinky (and also not buttony), the cleaner solution would be to not use `Hyperlink`, but derive a new control from the look-less `ButtonBase` instead, which doesn't come with default styling.

modules/javafx.controls/src/main/java/javafx/scene/control/DialogPane.java
andy-goryachev-oracle: I can confirm that we don't use these `//$NON-NLS-*` comments in javaFX.

`<rant>`
These comments were invented as a way to solve the localization problem.  The problem is real, but the "solution" was exactly the opposite.  Instead of marking the strings that don't need to be localized, what should have happened is to mark the string that need to be localized - with the information containing the context!

example:
```java
// verb, as in "execute this test"
String run = "Run";
```

What I've done with some other projects was 

```java
String run = TXT.get("Class.LocalizedResourceID.verb, as in execute this test", "Run");
```

`</rant>`

andy-goryachev-oracle:
> If you use a `Hyperlink` control, you also want the result to be hyperlinky and inherit all the hyperlinky styles

not necessarily, if you want all the functionality but different visuals.

BTW, the circular button in the DialogPane looks nothing like other modena buttons.  I suspect the code was transplanted from some other project.  May be the DialogPane should have used the TitledPane for the expandable content, to be visually consistent with the rest of javaFX?

In any case, I think it is wrong to retain the hyperlink style in this case.  What I agree with is that it should not use `setAll()` to wipe out the existing styles, because the `createDetailsButton()` is a protected method and the application can override it to set some other base style.

So my suggestion would be:
- remove the `hyperlink` style
- set `details-button` style
- add/remove `less`/`more` styles as proposed in this PR

mstr2:
> > If you use a `Hyperlink` control, you also want the result to be hyperlinky and inherit all the hyperlinky styles
> 
> not necessarily, if you want all the functionality but different visuals.

Yes, but the only thing `Hyperlink` adds from a functionality standpoint is the `visited` property, and that's not even used in this case.

andy-goryachev-oracle:
> `visited` property,

agree.  2014 was like, ten thousand years ago ;-)

modules/javafx.controls/src/main/java/javafx/scene/control/DialogPane.java
hjohn: Okay, I can remove them.

hjohn:
> > If you use a `Hyperlink` control, you also want the result to be hyperlinky and inherit all the hyperlinky styles
> 
> not necessarily, if you want all the functionality but different visuals.
> 
> BTW, the circular button in the DialogPane looks nothing like other modena buttons. I suspect the code was transplanted from some other project. May be the DialogPane should have used the TitledPane for the expandable content, to be visually consistent with the rest of javaFX?
> 
> In any case, I think it is wrong to retain the hyperlink style in this case. What I agree with is that it should not use `setAll()` to wipe out the existing styles, because the `createDetailsButton()` is a protected method and the application can override it to set some other base style.
> 
> So my suggestion would be:
> 
> * remove the `hyperlink` style
> * set `details-button` style
> * add/remove `less`/`more` styles as proposed in this PR

So... you want me to add more styles so it then ends up looking like a correctly colored hyperlink?

Wouldn't it be easier to do this the other way around, leave the hyperlink style, then apply any specific styling that is needed to make a details-button?

andy-goryachev-oracle:
> So... you want me to add more styles so it then ends up looking like a correctly colored hyperlink?

No.  I would like to retain the existing functionality where the styles alternate between
`[ "details-button", "less" ]` and `[ "details-button", "more" ]`

If you look at the existing code, the setAll() in line 825 gets invoked at the moment the details button is created because the listener gets called in line 829.

So basically, your new code needs one change in (new) line 822:

`styleClasses.setAll("details-button");`

the rest is good.

modules/javafx.controls/src/main/java/javafx/scene/control/DialogPane.java
andy-goryachev-oracle: line 822 should be

`styleClasses.setAll("details-button");`

hjohn:
> > So... you want me to add more styles so it then ends up looking like a correctly colored hyperlink?
> 
> No. I would like to retain the existing functionality where the styles alternate between `[ "details-button", "less" ]` and `[ "details-button", "more" ]`
> 
> If you look at the existing code, the setAll() in line 825 gets invoked at the moment the details button is created because the listener gets called in line 829.
> 
> So basically, your new code needs one change in (new) line 822:
> 
> `styleClasses.setAll("details-button");`
> 
> the rest is good.

Then we can close this, as that's what the original code does already.

andy-goryachev-oracle:
> Then we can close this, as that's what the original code does already.

Exactly (but I like the changes you made to add/remove "more"/"less" styles, since they won't erase any styles that the application might add during construction).

Let me ask about the original problem, which I read is difficulty to implement the dark theme.  Should we make any changes to `modena.css` in this PR instead (with or without the style manipulation)?

hjohn:
> > Then we can close this, as that's what the original code does already.
> 
> Exactly (but I like the changes you made to add/remove "more"/"less" styles, since they won't erase any styles that the application might add during construction).

The details button is private and not documented, so it is hard to reach, and even if you could it would be unsupported. 

> Let me ask about the original problem, which I read is difficulty to implement the dark theme. Should we make any changes to `modena.css` in this PR instead (with or without the style manipulation)?

The only change needed is to not wipe out the hyperlink style, then it will correctly follow the Modena specified colors which can be set to support a dark theme, see here how I do that:

```css
    -fx-accent: #165693;
    -fx-focus-color: -fx-accent;
    -fx-base: #16181c;
    -fx-background: #22252b;
    -fx-control-inner-background: derive(-fx-base, 20%);
    -fx-control-inner-background-alt: -fx-control-inner-background;

    -fx-outer-border: derive(-fx-color, 23%);
    -fx-selection-bar-non-focused: -fx-base;

    /* Lighter than -fx-background and used to provide a small highlight when
     * needed on top of -fx-background. This is never a shadow in Modena but
     * keep -fx-shadow-highlight-color name to be compatible with Caspian.
     */
    -fx-shadow-highlight-color: ladder(
        -fx-background,
        rgba(0,0,0,0.07) 0%,
        rgba(0,0,0,0.07) 20%,
        rgba(0,0,0,0.07) 70%,
        rgba(0,0,0,0.7) 90%,
        rgba(0,0,0,0.75) 100%
      );

    /* A gradient that goes from a bit lighter than -fx-color on the top to
     * a little darker at the bottom.  Typically is the third color in the
     * -fx-background-color list as a thin highlight inside the outer border.
     * Insets are typically 1.
     */
    -fx-inner-border: linear-gradient(to bottom,
                ladder(
                    -fx-color,
                    derive(-fx-color,-30%) 0%,
                    derive(-fx-color,-20%) 40%,
                    derive(-fx-color,-25%) 60%,
                    derive(-fx-color,-55%) 80%,
                    derive(-fx-color,-55%) 90%,
                    derive(-fx-color,-75%) 100%
                ),
                ladder(
                    -fx-color,
                    derive(-fx-color,-20%) 0%,
                    derive(-fx-color,-10%) 20%,
                    derive(-fx-color,-5%) 40%,
                    derive(-fx-color,2%) 60%,
                    derive(-fx-color,5%) 100%
                ));
```

This can be combined with a native API call (on Windows) to give your Window a dark title bar as well (if you want that code, let me know).

Because the style is wiped out however, I have to work-around it and copy the hyperlink styles specifically for details-button, like so:

```css
/* copied from .hyperlink styles for each pseudo state */
.dialog-pane > .button-bar > .container > .details-button {
    -fx-text-fill: -fx-accent;
}
.dialog-pane > .button-bar > .container > .details-button:armed,
.dialog-pane > .button-bar > .container > .details-button:visited,
.dialog-pane > .button-bar > .container > .details-button:hover:armed {
    -fx-text-fill: -fx-text-background-color;
}
```

andy-goryachev-oracle:
> The details button is private and not documented

it is not private: the method that creates it in the DialogPane:817 is protected, so the application can override it and get a pointer to the button, set styles, etc.

`protected Node createDetailsButton()`

andy-goryachev-oracle:
>  I have to work-around it and copy the hyperlink styles specifically for details-button

Adding these styles to `modena.css` seems to be a better fix.  What do you think?

hjohn:
> > The details button is private and not documented
> 
> it is not private: the method that creates it in the DialogPane:817 is protected, so the application can override it and get a pointer to the button, set styles, etc.
> 
> `protected Node createDetailsButton()`

Sorry, I missed that.  So its returning a `Hyperlink` that has been mangled to not be a hyperlink anymore.

> Adding these styles to modena.css seems to be a better fix. What do you think?

Unfortunately, I think we're going in circles now. I've already given my opinion on this several times and how to most correctly resolve this.

Anyone else have an opinion on how this should be resolved?

Maran23:
I think wiping default style classes should not be done.

Maybe somewhat related and only my understanding:
There are quite some places in JavaFX Controls where e.g. a `StackPane` is used to somewhat mimic a `Control` (often a `Button`).

Example: The `TitleRegion` is a `StackPane`, but behaves like a `Button`. This may was done to avoid focus issues, but I can also imagine that this was done to not wipe default style classes / break styling. 
In the example the `title` style class is set there (If it would be a `Button`, it would be wiped as well there, which might be expected or not for some developers).

So to come back, this looks to be the only place where this is done, right? I agree that is should keep the `hyperlink` style class and not wipe them. Or something else must be used, which is probably not feasable.

hjohn:
> I think wiping default style classes should not be done.
> 
> Maybe somewhat related and only my understanding: There are quite some places in JavaFX Controls where e.g. a `StackPane` is used to somewhat mimic a `Control` (often a `Button`).

I think that when you use an inheritance based system for controls, that you should always pick a starting point that does not yet have any unwanted behaviors or styles, instead of going for a move evolved starting point and then removing what you don't want.  The latter is harder as future changes may add new behaviors that make sense for the control you branched off from but not for your use case.

So if you don't want a Button that's clearly delineated for the user, that can have short cuts, can be hovered over, can be armed, disabled, highlighted, and participate as default or cancel action, then Button is not for you.  So using Button to create a clickable Tab is not a good idea.  The safest choice will be one of the container types as these don't have behaviors, nor do they have styles and skins.
 
> Example: The `TitleRegion` is a `StackPane`, but behaves like a `Button`. This may was done to avoid focus issues, but I can also imagine that this was done to not wipe default style classes / break styling. In the example the `title` style class is set there (If it would be a `Button`, it would be wiped as well there, which might be expected or not for some developers).

I think it was done to not have to de-evolve Button to make it into a Tab.  Tab could be its own control, with its own styles and skins, but I think it is so integrated that it would rarely serve a purpose if it was available separately from TabPane.
 
> So to come back, this looks to be the only place where this is done, right? I agree that is should keep the `hyperlink` style class and not wipe them. Or something else must be used, which is probably not feasable.

I didn't find any other instances where this is done.  If a Hyperlink is not what is desired here, then I think it is a bad idea to use Hyperlink and try to second guess what its skin is doing in combination with the default style.  Removing the hyperlink style means you'd need to style it yourself, which requires knowledge of the skin.  So either:

- It should not be a Hyperlink but something new, then create a new Control, or start from a Container type or any other type where you don't need to "remove" functionality or styling
- It should be a Hyperlink, but just look a bit different, then add additional styles or replace its skin

PR #1781, opened 2025-04-16
8354813: Parent.isNeedsLayout() may return wrong value in property listener

mstr2:
A listener that is added to `Parent.needsLayoutProperty()` may see a wrong value when calling `Parent.isNeedsLayout()` from the callback. The fix is to apply the value before notifying the listeners.

/reviewers 2

---------
### Progress

### Issue
 * [JDK-8354813](https://bugs.openjdk.org/browse/JDK-8354813): Parent.isNeedsLayout() may return wrong value in property listener (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1781/head:pull/1781` \
`$ git checkout pull/1781`

`$ git checkout pull/1781` \
`$ git pull https://git.openjdk.org/jfx.git pull/1781/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1781`

`$ git pr show -t 1781`

</details>

<a href="https://git.openjdk.org/jfx/pull/1781.diff">https://git.openjdk.org/jfx/pull/1781.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1781#issuecomment-2809163869)
</details>


modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
```diff
@@ -991,10 +991,13 @@ boolean isPerformingLayout() {
     private double minHeightCache = -1;
 
     void setLayoutFlag(LayoutFlags flag) {
+        // Needs to be set before needsLayout is updated, as otherwise a listener that
+        // calls isNeedsLayout() might see the old value.
+        layoutFlag = flag;
```
andy-goryachev-oracle, line 996: This is the right change, but I suspect it might cause regression.

The JavaFX entities which call `isNeedsLayout()`:

`VirtualFlow::layoutChildren`
`VirtualFlow::setCellIndex`

We may need to focus on List/Table/Tree/TableViews during testing.

(Anecdata: I've seen continuous layout calls in the TableView before)

mstr2: It's very unlikely to cause regressions, because `needsLayoutProperty()` is not used anywhere in JavaFX. The only thing that is different now is that a listener added to this property will see the correct value when calling `isNeedsLayout()`.

andy-goryachev-oracle: you are probably right, re:javafx.  applications might listen to this property, though it's rather unlikely.

mstr2: Yes, but I don't think there's anything we can do about it short of not fixing the bug...

andy-goryachev-oracle (approved):
I see no ill effects in the TableView.

Just to be on the safe side, I tried to re-create the scenario with the canvas-based cell and row factories in the monkey tester, and I did not see any issues.

hjohn (approved):

PR #1787, opened 2025-04-21
8089080: [TextArea] Caret disappear after pressing backspace to clear the content

Ziad-Mid:
Fixed Caret disappear after unfocus then focus on the TextArea, the issue was happening because the caret position was set to -1 when we focus on the TextArea with empty text and caused the caret to disappear from at the skin level when the position is set to -1

---------
### Progress

### Issue
 * [JDK-8089080](https://bugs.openjdk.org/browse/JDK-8089080): [TextArea] Caret disappear after pressing backspace to clear the content (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1787/head:pull/1787` \
`$ git checkout pull/1787`

`$ git checkout pull/1787` \
`$ git pull https://git.openjdk.org/jfx.git pull/1787/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1787`

`$ git pr show -t 1787`

</details>

<a href="https://git.openjdk.org/jfx/pull/1787.diff">https://git.openjdk.org/jfx/pull/1787.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1787#issuecomment-2876372803)
</details>


azuev-java (approved):
Looks good.

andy-goryachev-oracle:
Looks good!

Do you think it is possible to write a test for this case that would fail without the fix and pass with the fix?

Ziad-Mid:
> Do you think it is possible to write a test for this case that would fail without the fix and pass with the fix?

I have tried to make a test for it but I cannot get the information about whether the caret is visible or the caret position as the expected I have tried the available methods like `isCaretVisible` and `getCaretPosition`, so not sure if that is possible.
Also for me the issue is not reproducible in the test case from the ticket but in Monkey tester by default the issue happens when focusing on the `TextArea` , with the fix the issue do not happen anymore.

andy-goryachev-oracle (approved):
I see.  You are right - the fix is trivial enough, but the test is not.  I verified that failure is fixed in the monkey tester on macOS 15.4.1

Please wait the customary 24 hrs before integrating in case anyone else has  additional feedback.

PR #1789, opened 2025-04-22
8354943: [Linux] Simplify and update glass gtk backend: window sizing, positioning, and state management issues

tsayao:
This is a continuation to [JDK-8236651](https://bugs.openjdk.org/browse/JDK-8236651) and it aims to stabilize the linux glass gtk backend.

Overall, it has been made more robust within its scope, particularly in terms of sizing, positioning, and state management.

List of changes:
1. It embraces the asynchronous nature of X11 by reporting geometry changes only upon receiving a configure event, rather than immediately as before. This is because it merely requests changes from the window manager, which may or may not honor them. However, it still reports changes immediately in certain special cases, such as when the window has not yet been realized (i.e., when the window has not actually been created yet). One scenario where this behavior is evident is when we request the window to move to position (0, 0), but the window manager instead places it in the top-right corner where panels converge.
2. FullScreen now keeps track of geometry changes and apply them on restore as documented on Stage.java. No geometry changes affects the FullScreen state;
3. States (fullscreen, maximized and iconify) are now reported back to Java when it actually happens rather than immediately (except when not realized);
4. When a window is maximized, it will ignore geometry changes and restore to the geometry it had prior to being maximized. After some testing, I believe this is the best behavior for platform compatibility;
5. Unifies the WindowContext class: previously, there were three separate classes—two of which (for applets and Java Web Start) were removed, leaving only one. However, the supporting infrastructure was still there partially. [Unify WindowContext in glass-gtk](https://bugs.openjdk.org/browse/JDK-8305768)
6. `setAlpha` and `setBackground` were removed because they no longer function correctly due to the lack of shared rendering. It's not possible to paint a background using GTK and then render custom content on top of it, unless the rendering is done by Gtk (it is not). It is possible to keep the background by painting with cairo, and make it work with software rendering, but that's a very remote scenario;
7. Tests were added and re-enabled to ensure everything works correctly. The stage tests now cover various StageStyle configurations, as I found that `DECORATED` stages often behave differently from `UNDECORATED` or `UTILITY` stages;
8. Added Logs for debugging. Enable it with ` -PCONF=DebugNative`;
9. It no longer keeps track of geometry internally, as GTK already manages this - except in special cases, such as when the window has not yet been realized. Previously, it used BoundsType to distinguish between "Window Bounds" and "Content Bounds." Now, it consistently uses "Content Bounds" and converts them to "Window Bounds" by adding extents (decoration sizes) when necessary;
[...]

PR #1797, opened 2025-04-28
8176813: Mac: Failure to exit full-screen programmatically in some cases

beldenfox:
On macOS the system animates the transition into and out of fullscreen and this animation runs asynchronously. JavaFX tries to make the setFullScreen call appear synchronous by running a nested event loop while the transition is going on. But this means that runLater runnables can fire during a call to setFullScreen.

This can also occur during a call to Window.hide() if the window is in fullscreen mode. During the setView call glass tries to take the window out of fullscreen mode which fires up a nested event loop and, again, runLater runnables (like pulses) start firing.

In this PR GlassRunnables that try to run during the fullscreen transition are instead placed in a deferral list. When the fullscreen event loop exits they are re-scheduled.

---------
### Progress

### Issue
 * [JDK-8176813](https://bugs.openjdk.org/browse/JDK-8176813): Mac: Failure to exit full-screen programmatically in some cases (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1797/head:pull/1797` \
`$ git checkout pull/1797`

`$ git checkout pull/1797` \
`$ git pull https://git.openjdk.org/jfx.git pull/1797/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1797`

`$ git pr show -t 1797`

</details>

<a href="https://git.openjdk.org/jfx/pull/1797.diff">https://git.openjdk.org/jfx/pull/1797.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1797#issuecomment-2836036234)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
@@ -99,9 +99,12 @@ jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved)
 }
 
 - (id)initWithRunnable:(jobject)runnable;
+- (void)maybeRun;
 - (void)run;
 
 @end
+
+static NSMutableArray<GlassRunnable*> *deferredRunnables = nil;
 
 @implementation GlassRunnable
 
@@ -132,6 +135,27 @@ jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved)
     [pool drain];
 }
 
+- (void)maybeRun
+{
+    if (isFullScreenExitingLoop) {
+        if (deferredRunnables == nil) {
+            deferredRunnables = [[NSMutableArray alloc] initWithCapacity: 2];
+        }
+        [deferredRunnables addObject: self];
+    } else {
+        [self run];
+    }
+}
+
++ (void)rescheduleDeferredRunnables
+{
+    if (deferredRunnables != nil) {
+        for (GlassRunnable *runnable in deferredRunnables) {
+            [runnable performSelectorOnMainThread:@selector(maybeRun) withObject:nil waitUntilDone:NO modes:runLoopModes];
+        }
+        [deferredRunnables removeAllObjects];
+    }
+}
 @end
 
 @implementation NSApplicationFX
@@ -805,6 +829,7 @@ jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved)
         (*env)->ExceptionClear(env);
     }
     isFullScreenExitingLoop = NO;
+    [GlassRunnable rescheduleDeferredRunnables];
 }
 
 + (void)leaveFullScreenExitingLoopIfNeeded
@@ -1113,7 +1138,7 @@ JNIEXPORT void JNICALL Java_com_sun_glass_ui_mac_MacApplication__1submitForLater
     if (jEnv != NULL)
     {
         GlassRunnable *runnable = [[GlassRunnable alloc] initWithRunnable:(*env)->NewGlobalRef(env, jRunnable)];
-        [runnable performSelectorOnMainThread:@selector(run) withObject:nil waitUntilDone:NO modes:runLoopModes];
+        [runnable performSelectorOnMainThread:@selector(maybeRun) withObject:nil waitUntilDone:NO modes:runLoopModes];
     }
 }
 
--- a/tests/system/src/test/java/test/javafx/scene/RestoreSceneSizeTest.java
+++ b/tests/system/src/test/java/test/javafx/scene/RestoreSceneSizeTest.java
@@ -90,9 +90,8 @@ public class RestoreSceneSizeTest {
 
     @Test
     public void testUnfullscreenSize() throws Exception {
-        // Disable on Mac until JDK-8176813 is fixed
         // Disable on Linux until JDK-8353556 is fixed
-        assumeTrue(!(PlatformUtil.isMac() || PlatformUtil.isLinux()));
+        assumeTrue(!PlatformUtil.isLinux());
 
         Thread.sleep(200);
         final double w = (Math.ceil(WIDTH * scaleX)) / scaleX;
--- a/tests/system/src/test/java/test/javafx/stage/RestoreStagePositionTest.java
+++ b/tests/system/src/test/java/test/javafx/stage/RestoreStagePositionTest.java
@@ -92,7 +92,7 @@ public class RestoreStagePositionTest {
         double y = stage.getY();
 
         Platform.runLater(() -> stage.setFullScreen(true));
-        Thread.sleep(400);
+        Thread.sleep(800);
         Assertions.assertTrue(stage.isFullScreen());
         CountDownLatch latch = new CountDownLatch(2);
 
--- a/tests/system/src/test/java/test/javafx/stage/RestoreStagePositionTest.java
+++ b/tests/system/src/test/java/test/javafx/stage/RestoreStagePositionTest.java
@@ -49,7 +49,7 @@ public class RestoreStagePositionTest {
         initFX();
         try {
             RestoreStagePositionTest test = new RestoreStagePositionTest();
-            test.testUfullscreenPosition();
+            test.testUnfullscreenPosition();
             test.testDemaximizedPosition();
         } catch (Throwable e) {
             e.printStackTrace();
@@ -83,7 +83,7 @@ public class RestoreStagePositionTest {
     }
 
     @Test
-    public void testUfullscreenPosition() throws Exception {
+    public void testUnfullscreenPosition() throws Exception {
         Thread.sleep(200);
         Assertions.assertTrue(stage.isShowing());
         Assertions.assertFalse(stage.isFullScreen());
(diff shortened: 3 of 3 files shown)


kevinrushforth:
This looks like a reasonable solution. I'll put it on my review queue and also do a full headful test run later this week.

A couple quick questions:

* Do you think this will increase the possibility of deadlock?
* Will there be any problem one of the deferred Runnables causes an exitFullScreen (e.g., on a different Stage in a dual screen case)? This might be worth testing.

@andy-goryachev-oracle You might want to test this as well, if you have time.

andy-goryachev-oracle:
the proposed solution makes sense.

question: could this deferral of runnables cause some kind of race condition in respect to other synchronous changes?  Like, for example, trying to hide the window while is being transitioned to or from full screen?

andy-goryachev-oracle:
Using the reproducer in the ticket, I could not reproduce the issue with `testDemaximizedPosition()` - it passed in the master branch.

(Just in case, I am going to attach the exact file)

[Stage_RestorePosition_8176813.java.zip](https://github.com/user-attachments/files/19949842/Stage_RestorePosition_8176813.java.zip)

andy-goryachev-oracle:
Could we also rename `RestoreStagePositionTest::testUfullscreenPosition`, it might be a typo.

beldenfox:
> * Do you think this will increase the possibility of deadlock?

This PR shifts the execution of runLater runnables out to the event loop in effect when setFullScreen() was called. This matches where Windows and Linux would process them. So if we encounter new deadlock conditions we should be seeing them on other platforms.

> * Will there be any problem one of the deferred Runnables causes an exitFullScreen (e.g., on a different Stage in a dual screen case)? This might be worth testing.

Could you provide more details on what you want tested? I'm not that familiar with how fullscreen works on dual screen setups.

beldenfox:
> question: could this deferral of runnables cause some kind of race condition in respect to other synchronous changes? Like, for example, trying to hide the window while is being transitioned to or from full screen?

One of the goals of this PR is to try to avoid making calls on the NSWindow during the transition. In general the OS either rejects the call or it triggers a bug of some sort.

I can only think of two ways for an NSWindow change request to come in during the transition. One is a runLater runnable getting fired by the event loop. This PR fixes that. The other route is if the OS sends us a notification in the middle of the transition. A JavaFX change listener or event handler could make a call that affects the NSWindow. Luckily it looks like the OS takes mercy on us and doesn't send us notifications mid-transition. For example, it doesn't notify us that the NSWindow's size or position has changed until the transition is done. So I think this PR would make it very hard to even attempt to do something like hide the window during the transition.

Beyond that this PR just shifts the execution of these runnables to a later point which matches when they would be executed on Windows or Linux. So if there's an issue with the order in which modifications to the stage are executed I think it would be similar across platforms.

beldenfox:
> Using the reproducer in the ticket, I could not reproduce the issue with `testDemaximizedPosition()` - it passed in the master branch.

You're right, it should work. The test is still disabled on macOS even though the blocking bug was fixed a while ago.

andy-goryachev-oracle:
> testDemaximizedPosition()

sorry, I meant with the standalone reproducer, not the actual test.  see here
https://github.com/andy-goryachev-oracle/Test/blob/main/src/goryachev/bugs/Stage_RestorePosition_8176813.java

beldenfox:
> > testDemaximizedPosition()
> 
> sorry, I meant with the standalone reproducer, not the actual test.

The standalone reproducer is just a slightly tweaked version of the system test. Both should work fine. The system test should have been re-enabled years ago when JDK-8089230 was resolved.

andy-goryachev-oracle:
I modified the reproducer to invoke each test 10 times:

```
    public static void launch() throws Exception {
        initFX();
        try {
            Stage_RestorePosition_8176813 test = new Stage_RestorePosition_8176813();
            t(() -> test.testUnfullscreenPosition());
            t(() -> test.testDemaximizedPosition());
        } catch (Throwable e) {
            e.printStackTrace();
        } finally {
            teardown();
        }
    }
    
    interface Ru {
        public void run() throws Exception;
    }

    static void t(Ru r) throws Exception {
        for (int i = 0; i < 10; i++) {
            r.run();
        }
    }
```

got this:

```
java.lang.AssertionError: Assertion failed: expected true, was false
	at andy_test/goryachev.bugs.Stage_RestorePosition_8176813.fail(Stage_RestorePosition_8176813.java:70)
	at andy_test/goryachev.bugs.Stage_RestorePosition_8176813.assertTrue(Stage_RestorePosition_8176813.java:75)
	at andy_test/goryachev.bugs.Stage_RestorePosition_8176813.testUnfullscreenPosition(Stage_RestorePosition_8176813.java:129)
	at andy_test/goryachev.bugs.Stage_RestorePosition_8176813.lambda$0(Stage_RestorePosition_8176813.java:50)
	at andy_test/goryachev.bugs.Stage_RestorePosition_8176813.t(Stage_RestorePosition_8176813.java:65)
	at andy_test/goryachev.bugs.Stage_RestorePosition_8176813.launch(Stage_RestorePosition_8176813.java:50)
	at andy_test/goryachev.apps.AppTestLauncher.main(AppTestLauncher.java:15)
```

andy-goryachev-oracle:
it's possible that 200/400 ms is just too short.

andy-goryachev-oracle (changes requested):
my testing looks good.

I'd suggest to double the timeout in `RestoreStagePositionTest::testUnfullscreenPosition()` (line 95) to 800ms since I was getting exceptions during testing.

kevinrushforth:
I'm doing a headful test run on our CI system with this PR + PR #1799 (so I don't have to do two runs). I'll also review it later.

kevinrushforth:
During the headful test run, the following test failed on all four of our test systems, so it is likely related to your fix.

```
RestoreStagePositionTest > testDemaximizedPosition() FAILED
    org.opentest4j.AssertionFailedError: Window was moved ==> expected: <400.0> but was: <424.0>
        at app//org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1024)
        at app//test.javafx.stage.RestoreStagePositionTest.testDemaximizedPosition(RestoreStagePositionTest.java:152)
```

I note that the x64 systems had a different actual value (450), which is probably related to screen resolution or visible bounds of the system.

This was on the maximize / demaximize test. There is a full screen test in the same test class.

beldenfox:
> During the headful test run, the following test failed on all four of our test systems, so it is likely related to your fix.

I think it's unrelated. During testing @andy-goryachev-oracle noted that the demaximized position test was no longer failing. According to the JBS the issue that was blocking that test was fixed years ago. So I (foolishly) re-enabled it and verified that it worked fine on my system (and apparently the system @andy-goryachev-oracle is using).

I've disabled it again. I should not have enabled it since it's irrelevant to the original bug or the changes in this PR. Sorry about that.

Still, it's puzzling that this is failing on the test systems. It's maximizing a 300x400 stage and then de-maximizing it. That should work unless the screen is tiny. I've been looking at the maximization code and it can get the size of the window wrong but it shouldn't get the x,y position wrong. I'll take a closer look.

beldenfox:
For what it's worth the failure of testDemaximizedPosition is due to old behavior captured in [JDK-8355990](https://bugs.openjdk.org/browse/JDK-8355990). During the maximize transition glass sends out a slew of position updates. This test can capture any one of these notifications and trigger the latch indicating that the transition is done. I've already put together a fix and will re-visit this test when I submit that PR.

kevinrushforth:
Thanks for the explanation. I didn't actually look at the tests in this PR or I would have seen that the failure was in a formerly-disabled test.

andy-goryachev-oracle (approved):

kevinrushforth:
> > * Will there be any problem one of the deferred Runnables causes an exitFullScreen (e.g., on a different Stage in a dual screen case)? This might be worth testing.
> 
> Could you provide more details on what you want tested? I'm not that familiar with how fullscreen works on dual screen setups.

My thought was to do something like this on a dual-monitor system:

1. Create two stages, one on each screen
2. Programmatically enter full-screen in both Windows from the same block (e.g., the same `runLater`)
3. Sleep until animation finishes
4. Programmatically exit full-screen in both Windows from the same block (e.g., the same `runLater`)
5. Sleep until animation finishes
6. Repeat steps 2-5 but this time from different runLater runnables
7. Programmatically enter full-screen in stage1
8. Sleep until animation finishes
9. Programmatically enter full-screen in stage2
10. While stage2 is still animating its full-screen-enter, programmatically exit full screen in stage1.

If there is a problem, it likely exists today. I doubt this PR will make it worse, but I wanted to at least bring it up as a thought experiment.

kevinrushforth (approved):
The code changes look good and all my testing looks good. I left a question about some possible additional testing, but I don't consider it directly related to this PR. It could be done as a follow-up if there is a concern.

beldenfox:
New stage attribute tests are being written for PR #1789. I'll get some dual-monitor tests added and try to ensure they push the system a bit (like trying to trigger this bug).

Just realized that some of the existing tests expect that when a stage is taken out of fullscreen or maximized mode you will see one and only one update to the size and position properties. I'm not entirely sure that's a requirement we should be testing for.

andy-goryachev-oracle:
> I'm not entirely sure that's a requirement we should be testing for.

I agree.  We should test that it settles to the expected values eventually.

PR #1799, opened 2025-04-30
8353314: macOS: Inconsistent fullscreen behavior

beldenfox:
macOS will allow any window to enter fullscreen mode but won't expand the window's size if the resizable bit isn't set in the window's style mask. For undecorated stages the code has to set this bit before entering fullscreen and restore the old value after exiting fullscreen.

The old code was taking a pointer to an NSWindow and casting it to a pointer to an unrelated type (GlassWindow). It was also making an unnecessary check. windowWillEnterFullScreen stashes away the state of the resizable bit before setting it and windowDidExitFullScreen restores the old state. There's no need for setResizableForFullscreen to check anything, it just needs to do what it's told.

System tests for this case are underway as part of PR #1789.

You might have difficulty reproducing the bug in the master branch.  The old code was doing a bogus pointer cast and then dereferencing it to check a state flag so the code sometimes worked and sometimes didn't.

---------
### Progress

### Issue
 * [JDK-8353314](https://bugs.openjdk.org/browse/JDK-8353314): macOS: Inconsistent fullscreen behavior (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1799/head:pull/1799` \
`$ git checkout pull/1799`

`$ git checkout pull/1799` \
`$ git pull https://git.openjdk.org/jfx.git pull/1799/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1799`

`$ git pr show -t 1799`

</details>

<a href="https://git.openjdk.org/jfx/pull/1799.diff">https://git.openjdk.org/jfx/pull/1799.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1799#issuecomment-2842846925)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassViewDelegate.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassViewDelegate.m
@@ -1234,15 +1234,13 @@ static jstring convertNSStringToJString(id aString, int length)
 - (void)setResizableForFullscreen:(BOOL)resizable
 {
     NSWindow* window =  [self->nsView window];
-    if (!((GlassWindow*) window)->isResizable) {
-        NSUInteger mask = [window styleMask];
-        if (resizable) {
-            mask |= NSResizableWindowMask;
-        } else {
-            mask &= ~(NSUInteger)NSResizableWindowMask;
-        }
-        [window setStyleMask: mask];
+    NSUInteger mask = [window styleMask];
+    if (resizable) {
+        mask |= NSResizableWindowMask;
+    } else {
+        mask &= ~(NSUInteger)NSResizableWindowMask;
     }
+    [window setStyleMask: mask];
 }
 
 /*
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassViewDelegate.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassViewDelegate.m
@@ -1236,9 +1236,9 @@ static jstring convertNSStringToJString(id aString, int length)
     NSWindow* window =  [self->nsView window];
     NSUInteger mask = [window styleMask];
     if (resizable) {
-        mask |= NSResizableWindowMask;
+        mask |= NSWindowStyleMaskResizable;
     } else {
-        mask &= ~(NSUInteger)NSResizableWindowMask;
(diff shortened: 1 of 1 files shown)


andy-goryachev-oracle:
Thank you for taking on this issue!

Does this PR depend on https://github.com/openjdk/jfx/pull/1789 in any way (system tests?).  In other words, should we wait for 1789 before looking at this, or this PR is sufficiently independent?

beldenfox:
> Does this PR depend on #1789 in any way (system tests?). In other words, should we wait for 1789 before looking at this, or this PR is sufficiently independent?

This is a small enough PR that I was hoping to get it in before #1789 which might take a while to go through. Still, it needs to be tested with all of the different stage styles and #1789 has all the system tests for that. I've been testing it using a manual test app, the one I attached to [JDK-8355990](https://bugs.openjdk.org/browse/JDK-8355990). #1789 also has a manual test case you can grab (it's called TestStage.java). Let me know if you want one of these attached here or to the JBS ticket.

Never sure how to deal with this chicken-and-egg problem. I've been trying to get fixes in before the tests in #1789 are submitted since that seems less complicated than disabling the tests in that PR and then having to go back and re-enable them piecemeal later.

andy-goryachev-oracle:
I see, thanks.
We can try to process this independently of #1789, since we do have a reproducer in the ticket.

/reviewers 2

kevinrushforth:
The code changes look good. I'm doing a headful test run on our CI system with this PR + PR #1797 (so I don't have to do two runs). I'll finish my review later.

andy-goryachev-oracle (approved):
Oh yes, the reproducer works as expected on macOS M1 15.4.1:

```
scene.width=150.0
scene.height=17.0
stage.width=150.0
stage.height=17.0
scene.width=1800.0
scene.height=1126.0
stage.width=1800.0
stage.height=1126.0
scene.width=150.0
scene.height=17.0
stage.width=150.0
stage.height=17.0
```

no ill effects observed in the monkey tester transitioning to/from fullscreen with iconified / maximized / normal / on top flags.

kevinrushforth (approved):
The fix looks good. I can confirm the problem using a simple variant of HellofullScreenToggle and an UNDECORATED stage. It works as expected now.

PR #1800, opened 2025-05-01
8356042: RichTextArea: tab stops attributes

andy-goryachev-oracle:
### Summary

This PR adds support for controlling tab stops: the `TAB_STOPS` paragraph attribute and the `defaultTabStops` property in the `RichTextModel`.  

While adding the paragraph attribute is a trivial process, adding a model-wide property requires adding of a mechanism to support document properties (something that was originally omitted in the first incubating release).  Using the document properties, we can now persist not only the default tab stops, but also provide the version umber for the storage format provided by the `RichTextFormatHandler`, which will enable support the format evolution in the future [0].

To showcase the feature, the `RichEditorDemoApp` gains a visual ruler and additional dialogs and menus.

Screenshot 2026-03-04 at 14 00 30

### Paragraph Attribute

Paragraph-specific tab stops are enabled by:

- `StyleAttributeMap.TAB_STOPS` constant
- `StyleAttributeMap.Builder.setTabStops(double ... positions)`

The tab stops are represented by the `TabStops` class which is basically an immutable list (we can't use List<TabStops> because it makes it impossible to offer compile-time type safety of `StyleAttribute`).

### Default Tab Stops

After the last paragraph tab stop, or when no paragraph tab stops is set, the document provides a way to set default tab stops via the model's `defaultTabStops` property:

These changes support the new property and other document properties:

- document-wide properties support in the `StyledTextModel` base class
- `defaultTabStops` property in the `RichTextModel` and `RichTextFormatHandler`
- document properties `VERSION` and `DEFAULT_TAB_STOPS`
- `StyledSegment`: `ofDocumentProperties()` factory, `getDocumentProperties()`

### Other Improvements

A number of other improvements were made along with the tab stop related changes:

- `character()`, `paragraph()`, and `document()` factory methods in the `StyleAttribute` class
- `isCharacterAttribute()`, `isParagraphAttribute()`, and `isDocumentAttribute()` methods in the `StyleAttribute` class
- `RichTextArea`: `documentArea` read-only property

### Testing

In addition to the `RichEditorDemoApp`, the standalone Monkey Tester [1] provides additional options via context menu in the `CodeArea` and `RichTextArea` pages.

### References

[0] [Rich Text Area (Incubator) Data Format Version 2](https://github.com/andy-goryachev-oracle/jfx/blob/8356042.ruler/doc-files/controls/RichTextArea/RichTextArea_DataFormat_Incubator.md)

[1] https://github.com/andy-goryachev-oracle/MonkeyTest/tree/8356042.tabs

---------

---------
### Progress
- \[x\] Change requires CSR request [JDK-8382308](https://bugs.openjdk.org/browse/JDK-8382308) to be approved

### Warnings
&nbsp;⚠️ Patch contains a binary file (doc-files/controls/RichTextArea/demo-text2.png)
&nbsp;⚠️ Patch contains a binary file (doc-files/controls/RichTextArea/rich-editor-example.png)

### Issues
 * [JDK-8356042](https://bugs.openjdk.org/browse/JDK-8356042): RichTextArea: tab stops attributes (**Enhancement** - P4)
 * [JDK-8382308](https://bugs.openjdk.org/browse/JDK-8382308): RichTextArea: tab stops attributes (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1800/head:pull/1800` \
`$ git checkout pull/1800`

`$ git checkout pull/1800` \
`$ git pull https://git.openjdk.org/jfx.git pull/1800/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1800`

`$ git pr show -t 1800`

</details>

<a href="https://git.openjdk.org/jfx/pull/1800.diff">https://git.openjdk.org/jfx/pull/1800.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1800#issuecomment-4000542330)
</details>


modules/javafx.graphics/src/main/java/javafx/scene/text/TabStop.java
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * This class encapsulates an immutable single tab stop within the {@link TabStopPolicy}.
+ * <p>
+ * A tab stop is at a specified distance from the
+ * left margin, aligns text in a specified way, and has a specified leader.
+ *
+ * @since 999 TODO
+ */
+public class TabStop {
```
Maran23, line 35: If this immutable, why not use a `record` instead?

andy-goryachev-oracle: good question!

I see no value here in making it a record, and I think there is a possibility for an application to extend TabStop in order to provide additional attributes.

andy-goryachev-oracle: (this thing is still in Draft, but thank you for checking up on it!)

kevinrushforth:
> should the informational note [0] be added to the repo, and where? doc-files/notes/27? or maybe under doc-files/info/RichTextArea ?

If this is part of the spec, then it should go in the javadoc. Otherwise, somewhere under `doc-files` might bbe reasonable -- like we do for the UI behavior docs (but not doc-files/notes, which is for supporting data for release notes).

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextModel.java
```diff
@@ -58,6 +66,40 @@ public RichTextModel() {
...
 
+    /**
+     * Specifies the default tab stop interval for tabs beyond the last stop provided
+     * by the paragraph's
+     * {@link jfx.incubator.scene.control.richtext.model.StyleAttributeMap#TAB_STOPS StyleAttributeMap.TAB_STOPS}
+     * attribute, if any.
+     * This is a fixed repeating distance (in pixels) to the
+     * next tab stop computed at regular intervals relative to the document content leading edge.
+     * <p>
+     * A value of equal or less than 0 disables the default interval.
```
kevinrushforth: And what do tabs do in this case? Fall back to tab stops using characters and render spaces to get to the next stop (like Text and TextFlow)? Render nothing? Render a single space?

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttribute.java
```diff
@@ -34,23 +34,72 @@
  * @see StyleAttributeMap
  * @since 24
  */
-public final class StyleAttribute<T> {
+public abstract class StyleAttribute<T> {
```
kevinrushforth: Is there is a compelling reason to make this class abstract? It seem better to keep this as a final concrete class. Subclasses aren't needed (they are overkill) for what you are doing; you can do the same thing with constructor parameters and final fields.

If there is some good reason to use subclasses that I'm missing, then I recommend making it a sealed class with named, non-public concrete classes. It isn't meant to be subclassed by applications.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttributeMap.java
```diff
@@ -349,6 +356,14 @@ public Double getSpaceRight() {
         return get(SPACE_RIGHT);
     }
 
+    /**
+     * This convenience method returns the value of {@link #TAB_STOPS} attribute, or null.
+     * @return the paragraph alignment attribute value
+     */
+    public TabStop[] getTabStops() {
```
kevinrushforth: `@since 27`

And would it be better to return a `List`?

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledSegment.java
```diff
@@ -114,6 +117,13 @@ public enum Type {
      */
     public abstract StyledSegment subSegment(int start, int end);
 
+    /**
+     * Returns the document properties or {@code null} if the segment type
+     * is not {@link Type#DOCUMENT_PROPERTIES}.
+     * @return the properties or null
+     */
+    public Map<String,String> getDocumentProperties() { return null; }
```
kevinrushforth, line 136: `@since 27`

kevinrushforth (commented):
Looking at the API, my biggest question is: why is there no `tabStopPolicy` property in the RichTextArea control, which would then be forwarded on to the model? This seems the more natural place to me, unless I'm missing something.

I left a few other brief comments on the API inline.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttribute.java
andy-goryachev-oracle: Good point!

`sealed`, in my opinion, should work with inline classes, but it does not.  This fails:

```
    public static <P> StyleAttribute<P> character(String name, Class<P> type) {
        return new StyleAttribute<P>(name, type) {
            @Override
            public boolean isCharacterAttribute() {
                return true;
            }
        };
    }
```

I'll change to named inner classes, you are absolutely right.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttributeMap.java
andy-goryachev-oracle: There might be benefits, but this would not compile:

`public static final StyleAttribute<List<TabStop>> TAB_STOPS = StyleAttribute.paragraph("TAB_STOPS", List<TabStop>.class);`

The idea here is to make the declaration and usage of it completely type-safe.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextFormatHandler.java
```diff
@@ -586,6 +664,36 @@ private String decodeText() throws IOException {
...
 
+        private static String decode(String text) throws IOException {
+            StringBuilder sb = null;
+            int sz = text.length();
+            for (int i = 0; i < sz; i++) {
+                char c = text.charAt(i);
+                switch (c) {
+                case '%':
+                    // decoding hex
+                    if(sb == null) {
```
kevinrushforth: Minor: space after `if`.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextFormatHandler.java
```diff
@@ -586,6 +664,36 @@ private String decodeText() throws IOException {
...
+                        sb = new StringBuilder(sz);
+                        if(i > 0) {
```
kevinrushforth: Ditto

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttribute.java
```diff
@@ -72,18 +106,92 @@ public final Class<T> getType() {
...
 
     /**
-     * Returns true for a paragraph attribute, false for a character attribute.
-     *
-     * @return true for a paragraph attribute, false for a character attribute
+     * Returns true if this instance is a character attribute.
+     * @return true for a character attribute
+     * @since 27
+     */
+    public boolean isCharacterAttribute() {
```
kevinrushforth, line 113: Have you considered a 3-valued enum for attribute type rather than 3 separate boolean attributes, which are mutually exclusive?

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttribute.java
```diff
@@ -72,18 +106,92 @@ public final Class<T> getType() {
...
+                name.equals(a.name);
+        }
+        return false;
+    }
+
+    @Override
+    public int hashCode() {
+        int h = getClass().hashCode();
+        h = 31 * h + name.hashCode();
+        h = 31 * h + type.hashCode();
```
kevinrushforth, line 158: Should you also add the attribute type (isCharacterAttribute(), etc)? It seems more bullet-proof, although in practice, it seems unlikely that there would be two attributes with the same name where one was a character attr and the other was a paragraph attr.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttributeMap.java
kevinrushforth: That's too bad, since List<TabStop> seems like a cleaner API. It might worth filing a follow-on RFE to see if there is a way to make it work. Perhaps ask Joe about it when you file the CSR?

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledSegment.java
```diff
@@ -49,6 +50,8 @@
 public abstract class StyledSegment {
     /** StyledSegment type */
     public enum Type {
+        /** Identifies a segment which contains the document properties */
+        DOCUMENT_PROPERTIES,
```
kevinrushforth, line 57: This needs `@since 27`.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/RichTextArea.java
```diff
@@ -2551,4 +2561,39 @@ public Object queryAccessibleAttribute(AccessibleAttribute attribute, Object...
...
+
+    public final Bounds getDocumentArea() {
+        if (documentArea == null) {
+            return null;
+        }
+        return documentArea.get();
+    }
+
+    private final void setDocumentArea(double minX, double minY, double width, double height) {
+        if (documentArea != null) {
```
kevinrushforth, line 2594: Typically a setter calls the property to create it if not already created. Is there a reason to instead ignore the set call in that case?

modules/jfx.incubator.richtext/src/test/java/test/jfx/incubator/scene/control/richtext/model/TestRichExport.java
```diff
@@ -0,0 +1,287 @@
...
+
+    @Test
+    public void emptyModel() {
+        verify("{}{!}");
+    }
+
+    @Test
+    public void simpleTextModel() {
+        append("1/n2");
+        verify("{}1/n2{!}");
```
kevinrushforth: Did you mean to test newlines? If so, then you have a typo: `/n` --> `\n`. If you did mean forward-slash + `n` then ok.

modules/jfx.incubator.richtext/src/test/java/test/jfx/incubator/scene/control/richtext/model/TestRichTextFormatHandler.java
```diff
@@ -277,4 +283,35 @@ private StyleAttributeMap normalize(StyleAttributeMap a) {
         }
         return a;
     }
+
+    @Test
+    public void save() throws IOException {
+        RichTextModel m = new RichTextModel();
+        m.setDefaultTabStops(155);
```
kevinrushforth, line 290: It would be helpful to also test saving / loading tab stops.

kevinrushforth (commented):
Generally looks good. I tested it using the app and it is working as expected. I left a few comments and questions inline.

The next step is to file the CSR.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttribute.java
andy-goryachev-oracle: No, this would be a wrong pattern.

Here, `isXXX()` indicates a trait, which may or may not map exactly to the type, and we certainly don't want the callers to write `if(getType == SOMETIME)`, but rather write `if(isXXX())`.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttribute.java
andy-goryachev-oracle: This one is interesting, it actually points to a deeper problem: currently it is possible to create two different `StyleAttribute`s with the same name and different types.  

They will work fine in hash tables and `StyleAttributeMap`s, but they will fail when marshaling to external storage, for example, by the `RichTextModel`, because it stores/looks up the attributes by name.  

This poses a number of problems:
1. we might need to ensure uniqueness by inventing a static registry.
2. even with a registry, we might end up with a compatibility issue down the line.  For example, the application might declare an attribute with the name that will be clobbered by the next version of RichTextArea.
3. there another, addmittedly more theoretical, possibility, when the application and let's say a third party library both decide to declare two different attributes with the same name.

One possible solution is to ensure that system attributes use some special character in their name (!b) and disallow that character in custom attributes.  This will decouple custom attributes from ours.

What do you think?

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/RichTextArea.java
andy-goryachev-oracle: Yes, there is a reason.  The document area gets updated frequently (resizing etc.) but we don't want to create the property unless it was requested.  When it's requested, the value is initialized with the current value and then follows the changes (externally it is a read-only property).

modules/jfx.incubator.richtext/src/test/java/test/jfx/incubator/scene/control/richtext/model/TestRichExport.java
andy-goryachev-oracle: good catch, thanks!

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttribute.java
kevinrushforth: This seems worth considering further. I recommend filing a follow-up issue so that we don't hold up this PR.

doc-files/controls/RichTextArea/RichTextArea_DataFormat_Incubator.md
```diff
@@ -0,0 +1,186 @@
...
+
+This format (as a part of an incubator module) is likely to change once the decision is made to integrate it
+into the JavaFX core.
+
+
+## Example
+
+In short, the format is plain text which contains a sequence of segments representing:
+
+- document properties
```
kevinrushforth: I recommend adding a header with a version number before the document properties section, instead of making the version number be just another document property. It is more future-proof, and easier to sanity check.

doc-files/controls/RichTextArea/RichTextArea_DataFormat_Incubator.md
```diff
@@ -0,0 +1,186 @@
...
+{ff|System}{fs|12.0}{tc|000000}monospaced{!}
+{8}paragraph attributes{!bg|FFFF4D}{!bullet}{!dir|L}{!spaceAbove|3.0}{!spaceBelow|4.0}{!spaceLeft|1.0}{!spaceRight|2.0}{!alignment|C}
+```
+
+In this example, the document properties are enclosed in `{# }`, each paragraph attribute is enclosed in `{! }`,
+each character attribute is enclosed in `{ }`.
+
+
+
+## Description
```
kevinrushforth: Should this format be part of the JavaFX API specification? If so, we will eventually need a normative definition that is part of the javadoc-generated API docs.

Related to this, you might consider defining the syntax and grammar of this file in some standardized way (e.g., BNF), especially if this becomes part of the specification.

doc-files/controls/RichTextArea/RichTextArea_DataFormat_v2.md
```diff
@@ -0,0 +1,186 @@
+# Rich Text Area (Incubator) Data Format Version 2
+
+Andy Goryachev
+
+<andy.goryachev@oracle.com>
```
kevinrushforth: We generally don't include email addresses in docs committed to the repo.

doc-files/controls/RichTextArea/RichTextArea_DataFormat_v2.md
```diff
@@ -0,0 +1,186 @@
...
+Example: 
+
+`{#tabs|99.5|version|v2}`
+
+specifies the following properties:
+
+|Key     |Value      |Comment
+|:-------|:----------|:-----------
+|tabs    |99.5       |default tab stops in pixels
+|version |v2         |format version, currently `v2`
```
kevinrushforth: See earlier comment.

kevinrushforth (commented):
I left a couple comments on the newly added RTA format doc. The rest looks good.

kevinrushforth:
@Ziad-Mid can you be the second reviewer?

doc-files/controls/RichTextArea/RichTextArea_DataFormat_Incubator.md
andy-goryachev-oracle: This is a good question!

Should it be documented as a part of `RichTextModel` javadoc, or similarly to the way `cssref.html` (inheriting all its formatting and maintenance shortcomings)?

BNF is a good idea too, albeit I'd rather postpone it until the moment we decide to add this incubating feature to the core.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttribute.java
andy-goryachev-oracle: Filed https://bugs.openjdk.org/browse/JDK-8382874

doc-files/controls/RichTextArea/RichTextArea_DataFormat_Incubator.md
andy-goryachev-oracle: I like this idea!

So the question is - another revision is coming with https://bugs.openjdk.org/browse/JDK-8365949

what would make more sense - add the version header to this PR or in the next one?

andy-goryachev-oracle: Added the format string in this PR.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextFormatHandler.java
```diff
@@ -55,9 +55,16 @@
  * The handler uses a simple text-based format:<p>
  * (*) denotes an optional element.
  * <pre>
+ * VERSION_STRING
  * DOCUMENT_PROPERTIES
  * PARAGRAPH[]
  *
+ * VERSION_STRING*: {
```
kevinrushforth, line 61: I presume this is optional to support backward compatibility with saved files prior to adding the version? When the RichTextArea API is finalized, this should be a required element.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledSegment.java
```diff
@@ -240,6 +253,36 @@ public String toString() {
         };
     }
 
+    /**
+     * Creates a StyledSegment that contains the document version.
+     * @param version the document version
+     * @return the StyledSegment instance
+     * @since 27
+     */
+    public static StyledSegment ofVersion(String version) {
```
kevinrushforth, line 264: What is the advantage of making the version number be a `StyledSegment` (as opposed to a separate type, e.g., `Header` or `Version`)?

modules/jfx.incubator.richtext/src/test/java/test/jfx/incubator/scene/control/richtext/model/TestRichTextFormatHandler.java
kevinrushforth: Unless I'm missing something, this is still pending.

kevinrushforth (commented):
Added a couple more comments / questions. I'll look at the CSR soon.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextFormatHandler.java
andy-goryachev-oracle: correct, it will be.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledSegment.java
andy-goryachev-oracle: _"Entities must not be multiplied beyond necessity"_

I think there is no need in separate classes (and the corresponding `instanceof` operation), any required logic can be implemented as a switch on `.getType()`

modules/jfx.incubator.richtext/src/test/java/test/jfx/incubator/scene/control/richtext/model/TestRichTextFormatHandler.java
andy-goryachev-oracle: nope, it's there: saving is L290 vs L297, loading is L315

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/TabStops.java
```diff
@@ -0,0 +1,197 @@
...
+import java.util.List;
+import java.util.ListIterator;
+import javafx.scene.text.TabStop;
+
+/**
+ * Encapsulates the tab stop positions within a paragraph.
+ *
+ * @since 27
+ */
+public class TabStops implements List<TabStop> {
```
kevinrushforth: This seems like a good solution to allow using a list of stops (rather than the array you used to have). Since the list of stops is unmodifiable after constructions, perhaps the docs should say that?

Suggestion: Make this class `final` unless you have a a compelling reason to extend it.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/TabStops.java
```diff
@@ -0,0 +1,197 @@
...
+    // JDK should have a public ImmutableList class.
+    private final List<TabStop> stops;
+
+    /**
+     * Constructor.
```
kevinrushforth: I would use the same language as the `of` method -- it's a little terse as it stands/ Something like:

```
Creates a {@code TabStops} instance from tab stop positions.
```

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/TabStops.java
```diff
@@ -0,0 +1,197 @@
...
+    public TabStops(List<TabStop> stops) {
+        this.stops = List.copyOf(stops);
+    }
+
+    /**
+     * Creates a {@code TabStops} instance from tab stop positions.
+     * @param positions the tab stop positions
+     * @return the new instance
+     */
+    public static TabStops of(double ... positions) {
```
kevinrushforth: Minor: we usually omit the space before the `...`

doc-files/controls/RichTextArea/RichTextArea_DataFormat_Incubator.md
kevinrushforth: Can you file a follow-up issue to resolve these questions?

modules/jfx.incubator.richtext/src/main/java/com/sun/jfx/incubator/scene/control/richtext/Converters.java
```diff
@@ -196,16 +198,19 @@ private static String fromTabStops(TabStop[] ts) {
         return sb.toString();
     }
 
-    private static TabStop[] toTabStops(String text) {
+    private static TabStops toTabStops(String text) {
+        if (text.length() == 0) {
+            return null;
```
kevinrushforth, line 203: An empty list might be cleaner unless there is a need for `null` (e.g., if `null` is logically distinct from an empty list).

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextModel.java
```diff
@@ -47,6 +47,21 @@
  */
 public class RichTextModel extends StyledTextModel {
 
+    /**
+     * This value of the {@link #defaultTabStopsProperty()} results in a legacy behavior
+     * where the tab spacing corresponds to 8 spaces.
+     * This may result in milasigned columns when different text segments use different fonts.
+     * @since 27
+     */
+    public static final double DEFAULT_TAB_STOPS_8 = 0.0;
```
kevinrushforth: When I suggested using named static final fields (constants), I hadn't realized that this was a floating-point attribute. It still seems OK to do this, but we should be careful checking for equality (other than equality with 0).

I don't much care for the use of `8` in the name. That seems odd. Maybe `DEFAULT_TAB_STOPS_FIXED` or similar?

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextFormatHandler.java
```diff
@@ -463,16 +531,34 @@ public StyledSegment nextSegment() {
...
+                    case '#':
+                        // document properties
+                        index += 2;
+                        Map<String,String> dp = parseDocumentProperties();
+                        return StyledSegment.ofDocumentProperties(dp);
+                    case '@':
+                        // version string
+                        index += 2;
+                        // TODO check if already exists
+                        String ver = parseString();
```
kevinrushforth, line 544: In order for it to be effective, the version needs to be the first thing in the file in addition to there being only one of them (and it eventually needs to be mandatory). I'm still not sold on treating the version string as "just another section", but at a minimum, you need to check whether this is the first thing seen (not just the first version block), and fail if it isn't.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledTextModel.java
```diff
@@ -742,6 +786,9 @@ private final TextPos replace(StyleResolver resolver, TextPos start, TextPos end
                 offset += len;
                 btm += len;
                 break;
+            case DOCUMENT_PROPERTIES:
+                handleDocumentProperties(seg.getDocumentProperties(), completeReplacement);
+                break;
```
kevinrushforth, line 791: Doesn't this need to handle VERSION (in any case, this seems like another reason not to treat version as just another attribute type).

doc-files/controls/RichTextArea/RichTextArea_DataFormat_Incubator.md
```diff
@@ -0,0 +1,206 @@
...
+
+The document is persisted as a UTF-8 encoded plain text file which contains a sequence of segments.
+The first segment is the document properties segment, followed by segments representing the paragraphs.
```
kevinrushforth: You need to say something about the version here, which is before the document properties.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttributeMap.java
```diff
@@ -644,6 +659,21 @@ public Builder setStrikeThrough(boolean on) {
...
+         * Sets the tab stops paragraph attribute.
+         * @param positions the tab stop positions
+         * @return this Builder instance
+         * @since 27
+         */
+        public Builder setTabStops(double ... positions) {
+            TabStop[] ts = new TabStop[positions.length];
+            for (int i = 0; i < positions.length; i++) {
+                ts[i] = new TabStop(positions[i]);
+            }
```
kevinrushforth: The `ts` variable is unused and unneeded now, so you can removed these lines.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttributeMap.java
```diff
@@ -349,6 +356,14 @@ public Double getSpaceRight() {
         return get(SPACE_RIGHT);
     }
 
+    /**
+     * This convenience method returns the value of {@link #TAB_STOPS} attribute, or null.
+     * @return the paragraph alignment attribute value
+     */
```
kevinrushforth, line 362: Missing `@since 27`

kevinrushforth (commented):
The updated API looks good with a couple naming suggestions. I left a few doc comments as well.

I also left a couple implementation comments. The main one being how the version number is treated. As I mentioned in one of the inline comments, I'm not sold on the notion that the Version string should be treated as "just another section in the docs" and as a type of StyleAttribute. That latter especially seems odd, since the version is only something you should encounter when reading or writing a file. It has relevance to the I/O of the file, not to the processing of the document once read.

I think this could be handled as a follow-up if you prefer, perhaps at the same time the Version string is made mandatory.

modules/jfx.incubator.richtext/src/main/java/com/sun/jfx/incubator/scene/control/richtext/Converters.java
andy-goryachev-oracle: no, in this case we rely on `null` being returned.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextModel.java
andy-goryachev-oracle: FIXED is good, thanks!

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextFormatHandler.java
andy-goryachev-oracle: It is sort of byproduct of the serialization mechanism, in a sense that it emits a series of well-defined segments.  I suspect we might need to revisit the overall scheme once most of the requested attributes are in, and perhaps to change the way the model is being serialized altogether (for example, provide a more efficient binary format).

I'd like to leave this until later, with your permission.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledTextModel.java
andy-goryachev-oracle: (see my earlier response re:format)

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledSegment.java
kevinrushforth: This feels more like a "round peg in a square hole" to me, so I don't buy your argument. The Version number is a concern of the transport layer not the runtime document. I'm willing to defer it, but I want to revisit it and make a deliberate decision. This feels very unnatural from an API point of view.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextFormatHandler.java
kevinrushforth: Later is fine. Make sure we have a follow-on bug for it.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/TabStops.java
```diff
@@ -0,0 +1,197 @@
...
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.List;
+import java.util.ListIterator;
+import javafx.scene.text.TabStop;
+
+/**
+ * Encapsulates the tab stop positions within a paragraph.
```
kevinrushforth: Can you add a sentence about the list of stops being unmodifiable after construction? I recommend linking to `List#unmodifiable`. You might also copy the part about "calling any mutator method on the List will always cause UnsupportedOperationException to be thrown". That way it is clear what to expect if an app calls such a method.

doc-files/controls/RichTextArea/RichTextArea_DataFormat_Incubator.md
```diff
@@ -0,0 +1,200 @@
...
+
+`{@RichText-v2-incubator}`
+
+
+### Document Properties Segment
+
+The second segment is a document properties segment which contains a pipe-delimited key-value pairs enclosed in `{# ... }`.
+There is only one document properties segment per file.
+
+Example:
```
kevinrushforth: Minor: trailing whitespace

doc-files/controls/RichTextArea/RichTextArea_DataFormat_Incubator.md
```diff
@@ -0,0 +1,200 @@
...
+
+To avoid repeating the same attributes over and over, the format employs a form of compression, where the duplicate
+attributes are replaced by a numeric tokens:
+
+```
+{number}
+{!number}
+```
+
+where `number` is the index of the duplicate attribute map in the document.
```
kevinrushforth: Minor: trailing whitespace

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledTextModel.java
```diff
@@ -534,6 +544,38 @@ public final void export(TextPos start, TextPos end, StyledOutput out) throws IO
...
     }
 
+    /**
+     * This method is called by {@link #export(TextPos, TextPos, StyledOutput)}
+     * to export the model's version string, if any.
+     *
+     * @return the model version string, or null
+     * @since 27
+     */
+    protected String versionString() {
```
kevinrushforth, line 554: I don't understand why this is declared here in StyledTextModel, but not actually defined as a useful (non-null) value except in the RichTextModel subclass. This is another thing to add to the follow-up to figure out how version should be exposed in the API.

kevinrushforth (commented):
A few more doc comments, and one more issue to add to the follow-up issue around the version string.

Go ahead and update the CSR and move it to "Proposed".

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledTextModel.java
andy-goryachev-oracle: The main idea here is that the StyledTextModel handles the export, so it must know how to decompose the content into a series of `StyledSegment` fragments to be consumed by `StyledInput`. 

The model may not have the version or the document properties (a view-only model, for instance), but the base class knows how to deal with it if these things are present.

But I do get your point about the follow-up, so I filed

https://bugs.openjdk.org/browse/JDK-8386850

andy-goryachev-oracle:
Updated and "proposed" the CSR.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttributeMap.java
```diff
@@ -349,6 +355,15 @@ public Double getSpaceRight() {
         return get(SPACE_RIGHT);
     }
 
+    /**
+     * This convenience method returns the value of {@link #TAB_STOPS} attribute, or null.
+     * @return the paragraph alignment attribute value
```
kevinrushforth: This looks like a copy/paste error. That should be something like: `@return the tab stops attribute value`.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttributeMap.java
```diff
@@ -644,6 +659,17 @@ public Builder setStrikeThrough(boolean on) {
             return this;
         }
 
+        /**
+         * Sets the tab stops paragraph attribute.
+         * @param positions the tab stop positions
+         * @return this Builder instance
+         * @since 27
+         */
+        public Builder setTabStops(double ... positions) {
```
kevinrushforth: Minor: there is still an unneeded space before the `...`

kevinrushforth (commented):
Two more quick API doc issues I spotted.

kevinrushforth (approved):
LGTM

kevinrushforth:
@Ziad-Mid Can you also review?

Ziad-Mid (approved):
LGTM , latest changes are fine and works properly as expected

PR #1801, opened 2025-05-01
8355012: JavaFX modena.css -fx-highlight-text-fill bug

Ziad-Mid:
The issue was happening because `-fx-highlight-text-fill` was set to white , so when the background color is bright (>60% brightness ) the text is not visible , the solution is to change the value to be dynamically set depending on the level of brightness of the background this is done by `-fx-focused-text-base-color` .

---------
### Progress

### Issue
 * [JDK-8355012](https://bugs.openjdk.org/browse/JDK-8355012): JavaFX modena.css -fx-highlight-text-fill bug (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1801/head:pull/1801` \
`$ git checkout pull/1801`

`$ git checkout pull/1801` \
`$ git pull https://git.openjdk.org/jfx.git pull/1801/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1801`

`$ git pr show -t 1801`

</details>

<a href="https://git.openjdk.org/jfx/pull/1801.diff">https://git.openjdk.org/jfx/pull/1801.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1801#issuecomment-2846026196)
</details>


mstr2:
I don't think that styling details of the Modena theme should be unit-tested in `TextInputControlTest`. First of all, this has nothing to do with the `TextInputControl` implementation. Second, it seems a bit pointless to me to unit-test individual styling aspects of one particular CSS theme, it's not what unit tests are meant to do.

andy-goryachev-oracle:
> I'm interested in a second opinion on whether we should do this kind of testing, and what in particular it should accomplish.

We already have `fontSetFromCSS()` test there, so I am not against adding a test in this file.

You do bring a good point though, which is whether we should do in respect to CSS.  Perhaps we should add a set of tests that verify that CSS properties are loaded from 
- inline style
- scene stylesheet
- user agent stylesheet

(unless we have these tests already)

mstr2:
> We already have `fontSetFromCSS()` test there, so I am not against adding a test in this file.

That's not my point. `fontSetFromCSS` tests whether a font can be set via CSS, which concerns the implementation of `TextInputControl`. In contrast to that, the test in this PR doesn't test any aspect of `TextInputControl` at all. It tests the Modena theme, which doesn't belong here.
 
> You do bring a good point though, which is whether we should do in respect to CSS. Perhaps we should add a set of tests that verify that CSS properties are loaded from
> 
> * inline style
> * scene stylesheet
> * user agent stylesheet
> 
> (unless we have these tests already)

This seems a bit too much, since we already have tests for the CSS infrastructure that applies values to styleable properties. I don't think we need to repeat that for each individual property, at least not manually.

andy-goryachev-oracle:
> I don't think we need to repeat that for each individual property, at least not manually.

Maybe not manually, but a reflection-based test that verifies each `Styleable` property (and also verify that these properties are listed in the cssref.html somehow).

Ziad-Mid:
I can put the test in the `StylesheetTest` class, which might be more appropriate for this case. what do you think ?

mstr2:
> I can put the test in the `StylesheetTest` class, which might be more appropriate for this case. what do you think ?

No, because you're not testing the `Stylesheet` class either. What you are doing is testing the Modena theme, for which we have no appropriate testing methodology in JavaFX. I would rather just remove the test entirely, because this new test is going to be another person's future work. It's also more of a change tracker than a unit test, as a style is not a meaningful unit of code. If we want to start testing stylesheets, we should first have a discussion on goals and methodology.

kevinrushforth:
I agree with @mstr2 and @Maran23 that this test should not be in any existing test class. It isn't testing the CSS engine or any particular aspect of stylesheets in general. Rather it is testing an aspect of the `modena.css` stylesheet in particular. I recommend creating a new test class, perhaps under `test.com.sun.javafx.scene.control.skin.modena`.

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/skin/modena/ModenaTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+    private static Collection<Class> parameters() {
+        return List.of(
+                TextField.class,
+                PasswordField.class,
+                TextArea.class
+        );
+    }
+
+    private TextInputControl textInput;
```
andy-goryachev-oracle, line 33: Since this test tests `modena.css` aspects of the `TextInputControl` class hierarchy, should this test be renamed then (`TextInputControlModena` or something like that)?

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/skin/modena/ModenaTest.java
```diff
@@ -0,0 +1,98 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+import java.util.Collection;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class ModenaTest {
```
andy-goryachev-oracle, line 23: perhaps add a javadoc describing what this file is supposed to test?

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/skin/modena/ModenaTest.java
```diff
@@ -0,0 +1,98 @@
+package test.com.sun.javafx.scene.control.skin.modena;
```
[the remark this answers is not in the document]
Ziad-Mid: Added it, thanks

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/skin/modena/ModenaTest.java
Ziad-Mid: Yes found that more convenient and changed the name to TextInputControlModena, we can use this class to test modena.css for `TextInputControl`

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/skin/modena/ModenaTest.java
Ziad-Mid: I did add it

andy-goryachev-oracle (approved):

kevinrushforth (approved):
LGTM

Ziad-Mid:
Thank you all for your reviews and feedback!

andy-goryachev-oracle:
good job!

PR #1804, opened 2025-05-13
8351357: Resolve system test focus issues

lukostyra:
Originally this issue was supposed to resolve problems with some system tests (`MenuDoubleShortcutTest`, `TextAreaBehaviorTest` and others) failing on my machine. In the process of figuring this out I found out the problem is Windows `::SetForegroundWindow()` API refusing to give focus to JFX Stage upon calling `Stage.show()`.

The problem happened only when running system tests via Gradle, and with more investigation it turned out the culprit is actually running tests via a Gradle Daemon, which is the default behavior. According to [SetForegroundWindow API remarks](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setforegroundwindow) there is a list of conditions a process must meet to be granted a privilege of receiving focus, which is supposed to prevent focus stealing. While we do meet the required conditions, we don't meet "one of" additional conditions listed in the reference:
- Gradle daemon is a background process, so tests started by it do not meet `The calling process is the foreground process.` and `The calling process was started by the foreground process.` conditions
- We most probably run the tests from the terminal, so `There is currently no foreground window, and thus no foreground process.` condition fails - the foreground window is the Terminal itself.
- Each test has fresh-started JFX stage so `The calling process received the last input event.` condition cannot be met and would require either Robot workarounds or manual interaction before each test case.
- There is no debugger involved in the process (at least most of the time) so `Either the foreground process or the calling process is being debugged.` is also not met.

As such, Windows refuses to grant JFX Stage focus, which fails some system tests relying on it.

While we cannot remedy these conditions in-code (outside of hacky solutions I found with `AttachThreadInput` API which I am not a fan of) the only solution seems to be running the tests on Windows via either `gradle --no-daemon` or by setting `org.gradle.daemon=false` property somewhere in `gradle.properties`.

In the process of debugging this problem I wrote a canary test to detect whether a Stage receives focus right after calling `show()`. I ran this test on all (accessible to me) platforms (Windows, Linux, macOS) - on both Linux and macOS the test passes regardless of whether the Gradle deamon is used or not. On my Windows machine (Win 11 24H2) it fails when testing through Gradle Daemon (among tests mentioned in the first paragraph) and succeeds when `--no-daemon` flag is used.

I figured the test could be added to the repository and serve as a helper if someone else encounters similar problems.

### DIFF
--- a/tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright (c) 2025, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+package test.robot.javafx.stage;
+
+import java.util.concurrent.CountDownLatch;
+
+import javafx.application.Application;
+import javafx.application.Platform;
+import javafx.scene.Group;
+import javafx.scene.Scene;
+import javafx.scene.input.KeyCode;
+import javafx.scene.paint.Color;
+import javafx.scene.robot.Robot;
+import javafx.stage.Stage;
+import javafx.stage.WindowEvent;
+import test.util.Util;
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class StageFocusTest {
+
+    static CountDownLatch startupLatch;
+
+    static final double STAGE_SIZE = 200;
+
+    static final double STAGE_X = 100;
+    static final double STAGE_Y = 100;
+
+    static boolean receivedEvent = false;
+
+    Robot robot;
+
+    public static class TestApp extends Application {
+        @Override
+        public void start(Stage stage) {
+            Platform.setImplicitExit(false);
+
+            Group root = new Group();
+            Scene scene = new Scene(root, STAGE_SIZE, STAGE_SIZE);
+            scene.setFill(Color.LIGHTGREEN);
+            scene.setOnKeyPressed(e -> {
+                if (e.getCode() == KeyCode.A)
+                    receivedEvent = true;
+            });
+
+            stage.setScene(scene);
+            stage.addEventHandler(WindowEvent.WINDOW_SHOWN, e -> {
+                Platform.runLater(() -> {
+                    stage.setX(STAGE_X);
+                    stage.setY(STAGE_Y);
+                    startupLatch.countDown();
+                });
+            });
+            stage.show();
+        }
+    }
+
+    @BeforeAll
+    public static void setupOnce() throws Exception {
+        startupLatch = new CountDownLatch(1);
+        Util.launch(startupLatch, TestApp.class);
+    }
+
+    @AfterAll
+    public static void teardown() {
+        Util.shutdown();
+    }
+
+    @BeforeEach
+    public void setup() {
+        Util.runAndWait(() -> robot = new Robot());
+    }
+
+    /**
+     * Checks whether Stage is actually shown when calling show()
+     *
+     * Meant as a "canary" test of sorts to ensure other tests relying on
+     * Stage being actually shown and on foreground work fine.
+     */
+    @Test
+    public void testStageHasFocusAfterShow() {
+        Util.sleep(250);
+        Util.runAndWait(() -> {
+            robot.keyPress(KeyCode.A);
+        });
+        assertTrue(receivedEvent, "Expected key press has NOT been received! Stage did not have focus after showing. Some tests might fail because of this." +
+                                  "If that is the case, try re-running the tests with '--no-daemon' flag in Gradle.");
+    }
+}


lukostyra:
Renamed the issue title to better reflect what the change actually does

tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
```diff
@@ -0,0 +1,116 @@
...
+     * Meant as a "canary" test of sorts to ensure other tests relying on
+     * Stage being actually shown and on foreground work fine.
+     */
+    @Test
+    public void testStageHasFocusAfterShow() {
+        Util.sleep(250);
+        Util.runAndWait(() -> {
+            robot.keyPress(KeyCode.A);
+        });
+        assertTrue(receivedEvent, "Expected key press has NOT been received! Stage did not have focus after showing. Some tests might fail because of this." +
```
kevinrushforth: Do you need to sleep before the assertion to ensure that the event has been delivered?

tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
```diff
@@ -0,0 +1,116 @@
...
+public class StageFocusTest {
+
+    static CountDownLatch startupLatch;
+
+    static final double STAGE_SIZE = 200;
+
+    static final double STAGE_X = 100;
+    static final double STAGE_Y = 100;
+
+    static boolean receivedEvent = false;
```
kevinrushforth: Should this be volatile? It is set on one thread and read from another; unless you know that this is synchronously set by the robot key press call, I don't think you can be sure that the write `happens-before` the read.

kevinrushforth (commented):
The test looks good and runs as expected on my Windows 11 system. I left two related questions about the delivery of the event. If the key press event might not be delivered synchronously from the robot key press call, then the test could fail spuriously. You might consider a couple changes to make the test more robust.

tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
andy-goryachev-oracle: Perhaps we ought to have a utility (latch?) to ensure the sequence?

tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
andy-goryachev-oracle: or better an `AtomicBoolean`

tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * Checks whether Stage is actually shown when calling show()
+     *
+     * Meant as a "canary" test of sorts to ensure other tests relying on
+     * Stage being actually shown and on foreground work fine.
+     */
+    @Test
+    public void testStageHasFocusAfterShow() {
+        Util.sleep(250);
+        Util.runAndWait(() -> {
```
beldenfox, line 215: If the stage isn't focused after the sleep the robot's key press might go to some other app entirely. You might want to add an assert after the sleep checking that stage.isFocused() returns true.

tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
beldenfox: A delay of some sort is necessary to give the event loop time to process the event.

kevinrushforth: Right. I think Andy was suggesting a latch (with a timeout) that would be triggered (counted down to 0) when the event was received. This would also remove the concern I raised about the flag not being volatile.

tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
lukostyra: I checked this and it actually is a bug, `isFocused()` returns true in this case (when the Stage is not focused). We don't capture the failure of `SetForegroundWindow()`, so JFX assumes we are in focus. It doesn't necessarily change the outcome of the test, the key press check will then fail, but it would be good for the flag to also reflect that.

I'll add the assertion and file a separate issue for that.

tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
lukostyra: With a latch I think we don't even need the boolean value (we can simply wait for latch to trigger) so I removed it

tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
andy-goryachev-oracle: I wonder if we could paint the window in some unusual color and check the screen pixels to make sure the window is on top, in addition to being focused?

tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
```diff
@@ -75,53 +88,43 @@ public void start(Stage stage) {
...
-     */
-    @Test
-    public void testStageHasFocusAfterShow() throws InterruptedException {
-        // ensures UI finished showing the Stage
-        Util.sleep(250);
-        assertTrue(theStage.isFocused());
+        // check if window is on top
+        Util.runAndWait(() -> {
+            Color color = getColor(STAGE_SIZE / 2, STAGE_SIZE / 2);
+            assertColorEquals(Color.LIGHTGREEN, color, TOLERANCE);
```
andy-goryachev-oracle: I wonder if this implementation is a reliable test: the stage in question may be overlapped by another window somewhere in the corner, right?

What would be a reliable test?  Strictly speaking, we must check every pixel in the scene, though I wonder if checking each pixel in a grid (maybe 20 x 20, since we don't expect any reasonable window to be less than that) should be enough?

lukostyra: The most reliable way would be to probably scan the whole grid, I agree... I could change this - the stage is not that big after all. Maybe even turn the stage to undecorated to make this more reliable across platforms. I expect the very edge of the Stage to be not exactly the same color, but if we leave ex. 1px margin around the edges it should still be fine. I'll play around with this and update soon.

andy-goryachev-oracle: I thought about that too - but why wouldn't the client area, the scene itself, be of different color?  The decorations and possible effects should be outside of the stage, shouldn't they?

lukostyra: Turning the scene to undecorated would just make the math easier, without having to count in the window decorations which almost certainly would be of different color.

As for the margin thing, I had to do a bit of history searching and I came back with this old PR: https://github.com/openjdk/jfx/pull/1242

I think when writing the above my blurry memory recalled one of the attempts at solving this which involved just skipping the 1px margin of the Stage on HiDPI systems. I agree, it should be fine just going through the entire Stage.

lukostyra: Actually a small correction, using `VisualTestBase` already assumes the Stage we get is undecorated. So no changes need to be made there.

tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
```diff
@@ -0,0 +1,141 @@
...
+                if (e.getCode() == KeyCode.A) {
+                    eventReceivedLatch.countDown();
+                }
+            });
+
+            theStage.setScene(scene);
+            theStage.addEventHandler(WindowEvent.WINDOW_SHOWN, e -> {
+                Platform.runLater(() -> {
+                    theStage.setX(STAGE_X);
+                    theStage.setY(STAGE_Y);
```
andy-goryachev-oracle: setting Stage (x,y) coordinates in runLater makes the window jump.
it's probably better to move LL98-99 before runLater, to L96

with the robot disabled, the test fails with expected
```
org.opentest4j.AssertionFailedError: Event received latch timed out! Stage most probably did not have focus after showing. Some tests might fail because of this. If that is the case, try re-running the tests with '--no-daemon' flag in Gradle. ==> expected: <true> but was: <false>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
```

tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
```diff
@@ -0,0 +1,141 @@
...
+            }
+        });
+
+        // check if we actually have focus and key presses are registered by the app
+        Util.runAndWait(() -> {
+            getRobot().keyPress(KeyCode.A);
+        });
+        assertTrue(
+            eventReceivedLatch.await(TIMEOUT, TimeUnit.MILLISECONDS),
+            "Event received latch timed out! Stage most probably did not have focus after showing. Some tests might fail because of this. " +
```
andy-goryachev-oracle: I'd suggest to replace "Event received latch timed out!", or re-phrase completely, maybe

"The Stage did not received the key stroke generated by Robot.  This might happen if the stage did not receive focus.  Some tests might fail because of this.
Try re-running the tests with '--no-daemon' flag in Gradle.".

or something along these lines.

tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
```diff
@@ -0,0 +1,141 @@
...
+
+    static final int STAGE_X = 100;
+    static final int STAGE_Y = 100;
+
+    static final int TIMEOUT = 2000; // ms
+    static final double TOLERANCE = 0.07;
+
+    static final Color SCENE_COLOR = Color.LIGHTGREEN;
+
+    private Stage theStage = null;
```
andy-goryachev-oracle: null is unnecessary

tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
```diff
@@ -0,0 +1,141 @@
...
+        // give UI a bit of time to finish showing transition
+        // ex. on Windows above latch is set despite the UI still "animating" the show
+        sleep(500);
+
+        // check if window is on top
+        Util.runAndWait(() -> {
+            WritableImage capture = getRobot().getScreenCapture(null, STAGE_X, STAGE_Y, STAGE_SIZE, STAGE_SIZE, false);
+            PixelReader captureReader = capture.getPixelReader();
+            for (int x = 0; x < STAGE_SIZE; ++x) {
+                for (int y = 0; y < STAGE_SIZE; ++y) {
```
andy-goryachev-oracle, line 204: I was thinking of testing ever 5th or 10th pixel, but this code is fast enough.
If I quickly click on some window underneath the test window, the test correctly fails with 

org.opentest4j.AssertionFailedError: expected:rgba(144,238,144,255) but was:rgba(234,234,234,255)

andy-goryachev-oracle (commented):
Looks good!  Left some minor suggestions.

beldenfox:
I'm having an issue with this test. I can't get it to fail on my Windows 11 box but I can get other tests to fail due to focus issues.

I'm working on a system keyboard test (see attached patch for a pared down version). On my Windows 11 box the focus does not go to the stage unless I invoke the Robot mouse-click workaround. If I don't use that workaround the key event is going to the terminal window instead of the stage.

In that same environment the StageFocusTest in this PR always succeeds. I've never seen it fail. I've run these tests from both a Cygwin terminal and a MSYS2 MINGW terminal using bash and sh and StageFocusTest always passes.

I also notice that even when my test case fails the window is still on top, it just doesn't have focus. So the color testing in this PR seems unnecessary (?) Are you seeing cases where the stage is below other windows when it shows up?

[ShortcutKeyboardTest.patch](https://github.com/user-attachments/files/21026158/ShortcutKeyboardTest.patch)

lukostyra:
> I'm having an issue with this test. I can't get it to fail on my Windows 11 box but I can get other tests to fail due to focus issues.

I started to see this problem as well while working on [JDK-8359899 (the Stage.isFocused discrepancy)](https://bugs.openjdk.org/browse/JDK-8359899). I have a suspicion that switching to VisualTestBase caused some sort of difference - VisualTestBase creates new Stages separately and doesn't take into account the primary Stage that you would get via overriding Application.start(). I noticed your test does not use VisualTestBase so it might be it.

I have not looked into a solution yet outside of the obvious - reverting to not using VisualTestBase. I will get that figured out very soon with Andy's review suggestions.

> I also notice that even when my test case fails the window is still on top, it just doesn't have focus. So the color testing in this PR seems unnecessary (?) Are you seeing cases where the stage is below other windows when it shows up?

I did have situations where the Stage would appear under currently active window, sometimes even under the terminal window that runs the test.

lukostyra:
I did a couple of extra changes in addition to the regular fixes:
- `extend VisualTestBase` was removed, it was indeed the cause of the test always passing. The "Windows forbids granting focus" scenario seems to only happen when we try and show the Stage provided by `Application.start()`
- Tests are now parameterized and check both the primaryStage delivered by `Application.start()` as well as a separately created Stage. Figured it might be worth testing as well, since it seems to behave slightly differently and can affect other test.

I did notice this morning that on a fresh system start, the test _can_ pass even with `--daemon` gradle flag. This happens when it is the first invocation of Gradle that run - then the used Gradle daemon instance is a child of current Terminal process (which is a foreground process) which makes Windows treat it as a foreground process as well and grant focus to its children. The most reliable way I found of reproducing this and getting the test to fail on my Windows 11 box (also has to be done after calling `gradle --stop`):
- Run one build using `gradle --daemon`
- Close the Terminal window - that will cut the daemon's association with the Terminal and make Windows treat it as a background process
- Re-open the terminal and re-run the tests using `gradle --daemon`

For me the test then fails when checking the primary Stage delivered by `Application.start()` at the keystroke part of the test.

tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
```diff
@@ -0,0 +1,224 @@
...
+            "Some tests might fail because of this. Try re-running the tests with '--no-daemon' flag in Gradle."
+        );
+
+        // give UI a bit of time to finish showing transition
+        // ex. on Windows above latch is set despite the UI still "animating" the show
+        Thread.sleep(500);
+
+        // check if window is on top
+        Util.runAndWait(() -> {
+            WritableImage capture = robot.getScreenCapture(null, STAGE_X, STAGE_Y, STAGE_SIZE, STAGE_SIZE, false);
```
beldenfox, line 201: If the color test fails the developer doesn't see the helpful message telling them to use `--no-deamon` to patch things up.

lukostyra:
>What does it mean to integrate this? We would be adding a test designed to fail with the default testing environment. After failure a developer would have to re-run with the no daemon flag set. Shouldn't that become the default?

I agree that this way of running the tests should be the default but there is no way (as far as I know) to do this from the project level. The only path would be to add a `gradle.properties` file in the project root but we don't have one and it is ignored via gitignore. We only have a template which you need to copy-paste and rename to `gradle.properties`. As I understand it, different developers might need different ways of configuring gradle for their own needs, which should not be pushed to the repository or tracked by Git as meaningful changes.

According to [Gradle manual](https://docs.gradle.org/current/userguide/gradle_daemon.html#sec:disabling_the_daemon) other ways to disable the daemon are user-specific (`--no-daemon` flag, user-wide or system-wide property).

>If not, can we ensure this test runs early so it truly serves as a canary test?

I am not sure if that is possible, the only thing coming to my mind is executing that test specifically via `build.gradle` before running the whole suite. I'd be happy to get suggestions if anyone else knows something (ex. from JUnit) that could help it.

>Should folks developing new tests assume that they don't have to add the usual workarounds like sending a mouse click to the window?

Some tests did that, some tests did not. This whole ordeal started because `MenuDoubleShortcutTest` seemingly crashed mid-execution, later I found out the problem is Windows' focus stealing policy when you run the tests - test window was not in focus, the terminal was, so when the test program sent Ctrl+C for test purposes it was delivered to the terminal with obvious results. Before that was the case (and judging by how native calls like `_setVisible` behave) the assumption was that showing the Stage means it gets focus.

I have a patch queued up after this gets introduced which fixes the problem we talked about earlier with `Window.isFocused()` API being incorrect. Ideally I would hope that tests requiring sending some inputs to the Stage will call that function and fail early (like some other tests already do).

beldenfox:
> Some tests did that, some tests did not. This whole ordeal started because `MenuDoubleShortcutTest` seemingly crashed mid-execution […]

Ack! I wrote that test! And I should know better. I'll make sure it gets focus and stops using Ctrl+C as a test case.

I'm sure I ran MenuDoubleShortcutTest on Windows before submitting it. I just tried it again and it worked fine. Once. The second time it failed. I wish we could find a formula which guarantees these tests always work OR that they always fail.

lukostyra:
> I'm sure I ran MenuDoubleShortcutTest on Windows before submitting it.

This also started being a problem at some later point in time; on my machine tests used to do a full run with no failures with the Gradle daemon being enabled. I can't unfortunately pinpoint when the problems started happening, I only found out after the fact that `SetForegroundWindow()` enforces conditions mentioned in the PR description.

AFAIK `--no-daemon` consistently helps (I use MinGW on my Windows box). @kevinrushforth also mentioned offline that he disables the Gradle daemon in his machine-wide `gradle.properties` so he never encountered these problems.

I also submitted a (for now draft) PR for the `isFocused()` API fix - https://github.com/openjdk/jfx/pull/1849. It is independent from this test but the test was helpful with checking the behavior while writing the test. Feedback is welcome!

kevinrushforth:
Yes, I always disable the gradle daemon in my `$HOME/.gradle/gradle.properties`. Our CI builds and test runs also do this (our Jenkins scripts pass `--no-daemon` to all invocations of gradle). We've had way too many problems over the years, and this is just another example.

tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
```diff
@@ -206,7 +205,10 @@ public void testStageHasFocusAfterShow(Stage stage) throws InterruptedException
             for (int x = 0; x < STAGE_SIZE; ++x) {
                 for (int y = 0; y < STAGE_SIZE; ++y) {
                     Color color = captureReader.getColor(x, y);
-                    assertColorEquals(SCENE_COLOR, color, TOLERANCE);
+                    assertTrue(checkColorEquals(SCENE_COLOR, color, TOLERANCE),
```
beldenfox, line 206: It looks like this assertion is executed in the JavaFX thread instead of the JUnit testing thread. Is this officially supported? Seems to be working in my testing.

kevinrushforth: Yes, this is supported. The `Util.runAndWait` method catches all Throwables and re-throws them to the calling thread.

lukostyra:
> BTW, this test also fails on macOS 16 unless `--no-daemon` is specified. I'm also seeing failures on the MenuDoubleShortcutTest that are new. So it looks like macOS has also tightened up security related to background processes stealing focus.

Thanks for checking! I don't have access to macOS 16 yet, so it's good to know this ahead of time.

tests/system/src/test/java/test/robot/javafx/stage/StageFocusTest.java
```diff
@@ -0,0 +1,226 @@
...
+            testStages.add(new Stage());
+            launchLatch.countDown();
+        }
+    }
+
+
+    @BeforeAll
+    public static void setupOnce() throws InterruptedException {
+        Util.launch(launchLatch, TestApp.class);
+        assertTrue(launchLatch.await(TIMEOUT, TimeUnit.MILLISECONDS), "Timeout waiting for test stage to be shown");
```
kevinrushforth: Minor: This is redundant, since `Util.launch` already waits for the latch to countdown (and if it hadn't already been done, a 2 second wait for toolkit startup is too short).

kevinrushforth (approved):

PR #1805, opened 2025-05-14
8356652: Input field ignores custom input source characters

beldenfox:
Under the hood the Keyman input method appears as a US English keyboard layout. The characters attached to an NSEvent are always US English Roman even if the selected Keyman layout is, say, Hebrew or Dvorak. Keyman sends the correct Hebrew or Dvorak character to insertText:replacementRange: instead.

This PR special-cases the Keyman layout, detecting it using the same method that AWT does. When Keyman is active Glass records the insertText: character and uses that when sending out KeyEvents.

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassView3D.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassView3D.m
@@ -247,6 +247,7 @@
         self->imEnabled = NO;
         self->handlingKeyEvent = NO;
         self->didCommitText = NO;
+        self->insertTextChar = 0;
 
         lastKeyEvent = nil;
     }
@@ -513,6 +514,7 @@
 
     handlingKeyEvent = YES;
     didCommitText = NO;
+    insertTextChar = 0;
     BOOL hadMarkedText = (nsAttrBuffer.length > 0);
     BOOL inputContextHandledEvent = (imEnabled && [self.inputContext handleEvent:theEvent]);
     handlingKeyEvent = NO;
@@ -532,7 +534,7 @@
         ;
     } else if (!inputContextHandledEvent || (nsAttrBuffer.length == 0)) {
         [GlassApplication registerKeyEvent:theEvent];
-        wasConsumed = [self->_delegate sendJavaKeyEvent:theEvent isDown:YES];
+        wasConsumed = [self->_delegate sendJavaKeyEvent:theEvent isDown:YES character:insertTextChar];
     }
 
     return wasConsumed;
@@ -547,7 +549,7 @@
 - (void)keyUp:(NSEvent *)theEvent
 {
     KEYLOG("keyUp");
-    [self->_delegate sendJavaKeyEvent:theEvent isDown:NO];
+    [self->_delegate sendJavaKeyEvent:theEvent isDown:NO character:insertTextChar];
 }
 
 - (void)flagsChanged:(NSEvent *)theEvent
@@ -785,6 +787,20 @@
     if ([self->nsAttrBuffer length] > 0 || [aString length] > 1) {
         self->didCommitText = YES;
         [self commitString: aString];
+    }
+
+    // The Keyman input method sends different characters to keyDown than to
+    // insertText. If the layout is Hebrew the NSEvent sent to keyDown will
+    // contain Roman characters but Hebrew will be sent to insertText. Like
+    // AWT we special-case this layout.
+    if ([self.inputContext.selectedKeyboardInputSource containsString: @"keyman"]) {
+        if ([aString isKindOfClass: [NSString class]]) {
+            NSString* nsString = (NSString*)aString;
+            // Longer strings are sent out above as commits.
+            if (nsString.length == 1) {
+                self->insertTextChar = [nsString characterAtIndex: 0];
+            }
+        }
     }
 
     // If a user tries to enter an invalid character using a dead key
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassView3D.h
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassView3D.h
@@ -53,7 +53,7 @@
     BOOL handlingKeyEvent;
     BOOL didCommitText;
     BOOL isHiDPIAware;
-
+    unichar insertTextChar;
     NSEvent *lastKeyEvent;
 }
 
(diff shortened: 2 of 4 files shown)


andy-goryachev-oracle:
/reviewers 2

azuev-java (approved):
Looks reasonable.

andy-goryachev-oracle:
Noticed a bit of a problem.

To reproduce, install Himyarit Musnad keyboard https://keyman.com/keyboards/himyarit_musnad

typing 'h' causes this character to appear: 𐩱
typing 'n' followed by 'h' generates garbage input rather than expected 𐩱:

Screenshot 2025-05-15 at 08 51 18

(I am using the monkey tester, but it will work with any application that uses a TextArea)

azuev-java:
> typing 'h' causes this character to appear: 𐩱 typing 'n' followed by 'h' generates garbage input rather than expected 𐩱:

The same happens without the fix. For some reason events for the keys not mapped on the active keyman keyboard still pass to the component - at least partially. That affects the character composition.

andy-goryachev-oracle:
You are right, @azuev-java !  Looks like this is a separate bug, where the same text involving surrogate pairs is rendered differently by the TextArea:

Screenshot 2025-05-15 at 10 48 11

andy-goryachev-oracle (approved):
The change fixes the immediate problem with keyman, I'll create a separate ticket for the rendering bug.

beldenfox:
The N key in the Himyarit Musnad layout shouldn't generate a character (just a beep). Under the hood glass is seeing a keyDown: event for the N key (same as it would for a U.S. English layout) so it generates a KeyEvent. I can't figure out a good way of avoiding this, any fix I can think of would also affect keys like Tab  where we *have* to generate a KeyEvent. I'll continue to think about this but for now this PR gets Keyman working at some minimal level and I have my doubts that we can do much better.

andy-goryachev-oracle:
Created https://bugs.openjdk.org/browse/JDK-8357070 for the rendering bug.

andy-goryachev-oracle:
> The N key in the Himyarit Musnad layout shouldn't generate a character (just a beep)

Do you think it's a bug in keyman or our code?  I head the beep (correctly) but then n appears, despite the keyman onscreen keyboard showing an empty key.  If it is our code, should there be some additional logic?

srl295:
FYI @Sgschantz

beldenfox:
> Do you think it's a bug in keyman or our code? I head the beep (correctly) but then n appears, despite the keyman onscreen keyboard showing an empty key. If it is our code, should there be some additional logic?

Keyman is definitely sending us a keyDown: for the N key. I'm still trying to figure out how to ignore it without also dropping other key strokes that we need.

srl295:
@beldenfox i mentioned the mac lead who should be able to comment. 

Does the Devanagari case in https://bugs.openjdk.org/browse/JDK-8195675?focusedId=14148577&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14148577 work ? That's the one I was testing for the corresponding jdk issue

andy-goryachev-oracle:
@srl295 
ISIS Devanagari is no longer available:
https://keyman.com/keyboards/isis_devanagari

srl295:
It’s still there if you click through.
https://keyman.com/go/package/download/isis_devanagari?version=2.1.1&tier=stable

srl295 (approved):
Looks reasonable to me

beldenfox:
A second shot at getting the Keyman input method working. In this version we detect that Keyman is active at the very beginning of key down handling. If Keyman is active we drop any key event that doesn’t lead to a call to one of the NSTextInputClient methods. Character-generating keys are caught inside `insertText:replacementRange:`. All other keys (like Home, Backspace, Delete, function keys, etc.) are caught in `doCommandBySelector:`. Note: AWT uses a different heuristic; it captures keys like Home in `performKeyEquivalent:`.

Under the hood Keyman is using some Roman layout (for me it’s US English) and all keyboard shortcuts will be based on that layout and not the Keyman layout. So if you test with Dvorak shortcuts like Cmd+X require you to hit X in the US English (QWERTY) location, not the Dvorak one. This is a general problem with Keyman that affects all applications.

A keystroke that generate more than one UTF-16 code unit will be processed as a JavaFX InputMethodEvent rather than a KeyEvent. This is old logic that’s been in JavaFX for a long time that I didn’t want to re-visit. You’ll see this with the Himyarit Musnad layout.

I tried to isolate the Keyman logic as much as possible. I’ve done some testing with Himyarit Musnad, Hebrew, and ISIS-Devanagari. It won’t work the newer Guatami Devanagari since it keeps sending backspaces to erase previous characters. That layout doesn’t seem to work well anywhere.

srl295:
> keeps sending backspaces to erase previous characters

I do know that keyboards in general need to be able to interact with the existing context, that is, to convert h + i into ひ erasing the "h". I'm not sure of the correct mechanism here but just wanted to note that.

andy-goryachev-oracle:
As far as I can tell, this version works as expected.

I found one irregularity with the ISIS Devanagari keyboard: unmapped 'w' produces w symbol, unmapped punctuation keys produce corresponding symbols, but unmapped 'x' produces no input.  This could be the keyman problem though.

It is weird that we have to add third-party specific code to JavaFX as a workaround - why doesn't macOS provide a general purpose API for this?

modules/javafx.graphics/src/main/native-glass/mac/GlassView3D.m
```diff
@@ -777,6 +803,9 @@ - (void)doCommandBySelector:(SEL)aSelector
     // According to Apple an NSResponder will send this up the responder chain
     // but a text input client should not. So we ignore this which avoids an
     // annoying beep.
+    if (keymanActive) {
```
andy-goryachev-oracle, line 806: I noticed the latest code does not beep.  Were the beeps produced earlier expected for unmapped keys?

beldenfox: I'm not sure where the beeps are coming from. I still hear a beep when I press N on the Himyarit Musnad layout but I don't hear a beep when I press X on the ISIS-Devanagari layout. Both are unused keys but one produces a beep and the other doesn't.

beldenfox:
> It is weird that we have to add third-party specific code to JavaFX as a workaround - why doesn't macOS provide a general purpose API for this?

Normally an input method is designed for a specific language and is bundled with a collection of keyboard layouts. The input method is an active blob of code but the keyboard layouts are basically big passive tables mapping hardware key positions to characters. It appears that Keyman is trying to use one input method that works with all of its layouts. It's not producing the corresponding data tables; the OS just sees the same Roman table when it generates events. We're basically working around that mis-match.

To really do this right Keyman would either have to create new layout tables on-the-fly (assuming they can, I've never tried it) or use a low-level event tap to re-write the keyboard events (assuming they can, I've never tried it). I suspect the API's are available but are probably a devil to get right.

andy-goryachev-oracle (approved):
Thank you for clarification!

This PR fixes the issues with multiple keyman keyboards without introducing regressions with the normal keyboard or the IME.

PR #1809, opened 2025-05-15
8342530: Specifying "@Nx" scaling level in ImageStorage should only load that specific level

lukostyra:
This follow-up change finishes the earlier changes to `ImageStorage.loadAll()` and adds support for loading specific scale requested in the input.

`loadAll()` will now first check if the input path ends with a scaling level specified, and if that is the case it will attempt creating a Stream. If requested resource does not exist it will throw an Exception, skipping the rest of the load process. If the resource does _not_ have a scaled name in its path, it will continue loading as normal - looking for all scale levels, trying to load the main resource and falling back to trying to load "@1x" variant.

Added tests to check the new `ImageTools.hasScaledName()` method + new behavior.

---------
### Progress

### Issue
 * [JDK-8342530](https://bugs.openjdk.org/browse/JDK-8342530): Specifying "@Nx" scaling level in ImageStorage should only load that specific level (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1809/head:pull/1809` \
`$ git checkout pull/1809`

`$ git checkout pull/1809` \
`$ git pull https://git.openjdk.org/jfx.git pull/1809/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1809`

`$ git pr show -t 1809`

</details>

<a href="https://git.openjdk.org/jfx/pull/1809.diff">https://git.openjdk.org/jfx/pull/1809.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1809#issuecomment-2883497884)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/iio/common/ImageTools.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/iio/common/ImageTools.java
@@ -52,7 +52,7 @@ public class ImageTools {
     /**
      * Regex pattern for hasScaledName
      */
-    private static final Pattern SCALED_FILE_PATTERN = Pattern.compile(".*@[1-9][0-9]?x(\\.[^\\.]+)?");
+    private static final Pattern SCALED_FILE_PATTERN = Pattern.compile(".*@[1-9][0-9]?x(?:\\.[^\\.]+)?");
 
     /**
      * See the general contract of the <code>readFully</code>
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/iio/common/ImageTools.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/iio/common/ImageTools.java
@@ -146,6 +146,24 @@ public class ImageTools {
         return result.toString();
     }
 
+    public static boolean hasScaledName(String path) {
+        int slash = path.lastIndexOf('/');
+        String name = (slash < 0) ? path : path.substring(slash + 1);
+        int at = name.lastIndexOf('@');
+        if (at < 0) return false;
+
+        int dot = name.lastIndexOf('.');
+        String scale = (dot < 0) ? name.substring(at + 1) : name.substring(at + 1, dot);
+        if (!scale.endsWith("x")) return false;
+
+        try {
+            Integer.valueOf(scale.substring(0, scale.length() - 1));
+        } catch (NumberFormatException ignored) {
+            return false;
+        }
+        return true;
+    }
+
     public static InputStream createInputStream(String input) throws IOException {
         InputStream stream = null;
 
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/iio/common/ImageTools.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/iio/common/ImageTools.java
@@ -35,6 +35,8 @@ import java.io.InputStream;
 import java.net.URL;
 import java.nio.ByteBuffer;
 import java.util.Locale;
+import java.util.regex.Pattern;
+import java.util.regex.Matcher;
 
 /**
  * A set of format-independent convenience methods useful in image loading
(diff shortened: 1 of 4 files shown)


mstr2 (approved):
Looks good.

modules/javafx.graphics/src/main/java/com/sun/javafx/iio/common/ImageTools.java
```diff
@@ -146,6 +146,24 @@ public static String getScaledImageName(String path, int scaleFactor) {
...
+        if (!scale.endsWith("x")) return false;
+
+        try {
+            Integer.valueOf(scale.substring(0, scale.length() - 1));
+        } catch (NumberFormatException ignored) {
+            return false;
+        }
+        return true;
+    }
+
```
hjohn, line 166: 1. Is looking for a slash going to be compatible on all platforms?  Where is the path string coming from?
2. Catching `NumberFormatException` to "check" if something is a number is bad form
3. It will allow `@0x` and `@-1x` etc...
4. Consider using a regular expression, it is much more concise and intended for this kind of matching

Here's a regular expression for this:

    Pattern SCALED_PATTERN = Pattern.compile(".*@[1-9][0-9]?x(\\.[^\\.]+)?");
    
The above will match any path that ends with `@` followed by a number from 1 to 99, followed by an `x`, optionally followed by an extension that does not contain a dot.

hjohn (changes requested):

modules/javafx.graphics/src/main/java/com/sun/javafx/iio/common/ImageTools.java
hjohn: If you want you can even return the scale with a slightly altered pattern:

     Pattern SCALED_PATTERN = Pattern.compile(".*@([1-9][0-9]?)x(?:\\.[^\\.]+)?");
     
Then do:

     Matcher matcher = SCALED_PATERN.matcher(path);
     
     if (matcher.matches()) {
          return Integer.parseInt(matcher.group(1));  // can't throw NumbefFormatException, number is validated by pattern
     }
     
     return 0;  // there was no scale

lukostyra: >Is looking for a slash going to be compatible on all platforms? Where is the path string coming from?

My assumption was that other parts of ImageTools use this assumption so here it would be okay as well. `getScaledImageName` right above this method assumes so (although there it might not matter as much).

Regex might be a better fit for this, I agree.

hjohn (approved):
Looks good; you could add tests to check if `@0x` and `@-1x` are now indeed not recognized as scaled names.

modules/javafx.graphics/src/main/java/com/sun/javafx/iio/common/ImageTools.java
```diff
@@ -47,6 +49,11 @@ public class ImageTools {
      */
     public static final int PROGRESS_INTERVAL = 5;
 
+    /**
+     * Regex pattern for hasScaledName
+     */
+    private static final Pattern SCALED_FILE_PATTERN = Pattern.compile(".*@[1-9][0-9]?x(\\.[^\\.]+)?");
```
mstr2, line 55: You can use a non-capturing group instead of a capturing group if you don't intend to retrieve its value:
```suggestion
    private static final Pattern SCALED_FILE_PATTERN = Pattern.compile(".*@[1-9][0-9]?x(?:\\.[^\\.]+)?");
```

kevinrushforth:
@lukostyra Can you rerun your GitHub action build? It failed with an intermittent network error when trying to run the gradle-wrapper verification task.

lukostyra:
All checks passed so I'll merge this one. Thanks for review everyone!

/integrate

PR #1810, opened 2025-05-15
8357067: Platform preference change can emit multiple notifications

mstr2:
Some platform preference changes can trigger the emission of multiple notifications. For example, when switching from a high-contrast theme on Windows to the regular theme, the following notifications are emitted (log can be viewed in `PlatformPreferencesChangedTest`):

```
changed:
	Windows.UIColor.Accent=0x0078d4ff
	Windows.SysColor.COLOR_HIGHLIGHTTEXT=0xffffffff
	Windows.SysColor.COLOR_WINDOW=0xffffffff
	Windows.UIColor.AccentLight1=0x0091f8ff
	Windows.SysColor.COLOR_3DFACE=0xf0f0f0ff
	Windows.SysColor.COLOR_HOTLIGHT=0x0066ccff
	Windows.SysColor.COLOR_WINDOWTEXT=0x000000ff
	Windows.SysColor.COLOR_BTNTEXT=0x000000ff
	Windows.UIColor.Foreground=0x000000ff
	Windows.UIColor.AccentDark1=0x0067c0ff
	Windows.UIColor.Background=0xffffffff
	Windows.UIColor.AccentLight3=0x99ebffff
	Windows.UIColor.AccentLight2=0x4cc2ffff
	Windows.SysColor.COLOR_GRAYTEXT=0x6d6d6dff
	Windows.SysColor.COLOR_HIGHLIGHT=0x0078d7ff
	Windows.UIColor.AccentDark2=0x003e92ff
	Windows.UIColor.AccentDark3=0x001a68ff

changed:
	-Windows.SPI.HighContrastColorScheme
	Windows.SPI.HighContrast=false

changed:
	Windows.UIColor.Foreground=0xffffffff
	Windows.UIColor.Background=0x000000ff
```

Notably, the values for Windows.UIColor.Foreground/Background are inconsistent between the notifications (although they are eventually correct). In general, only a single notification should be emitted that includes all of the changed preferences.

This artifact is only visible on Windows. The reason is that changing some system settings (like high-contrast theme) causes a number of different window messages to be sent to the application. We should wait for all window messages to come in, and then aggregate all of the changed preferences into a single notification.

In order to minimize the delay between changing a system setting and sending out the notification in JavaFX, the implementation only waits when instructed to do so by the native toolkit. This allows us to instantly send out notifications for most changes, but selectively wait a bit for changes where the native toolkit knows that more changes might be coming in.

---------
### Progress

### Issue
 * [JDK-8357067](https://bugs.openjdk.org/browse/JDK-8357067): Platform preference change can emit multiple notifications (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1810/head:pull/1810` \
`$ git checkout pull/1810`

`$ git checkout pull/1810` \
`$ git pull https://git.openjdk.org/jfx.git pull/1810/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1810`

`$ git pr show -t 1810`

</details>

<a href="https://git.openjdk.org/jfx/pull/1810.diff">https://git.openjdk.org/jfx/pull/1810.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1810#issuecomment-2884633802)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/glass/ui/Application.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/glass/ui/Application.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -84,7 +84,7 @@ public abstract class Application {
         // currently used only on Mac OS X
         public void handleQuitAction(Application app, long time) {
         }
-        public void handlePreferencesChanged(Map<String, Object> preferences) {
+        public void handlePreferencesChanged(Map<String, Object> preferences, int suggestedDelayMillis) {
         }
     }
 
@@ -255,10 +255,10 @@ public abstract class Application {
         }
     }
 
-    protected void notifyPreferencesChanged(Map<String, Object> preferences) {
+    protected void notifyPreferencesChanged(Map<String, Object> preferences, int suggestedDelayMillis) {
         EventHandler handler = getEventHandler();
         if (handler != null) {
-            handler.handlePreferencesChanged(preferences);
+            handler.handlePreferencesChanged(preferences, suggestedDelayMillis);
         }
     }
 
--- a/modules/javafx.graphics/src/main/native-glass/win/PlatformSupport.h
+++ b/modules/javafx.graphics/src/main/native-glass/win/PlatformSupport.h
@@ -61,14 +61,19 @@ public:
      * Collect the specified platform preferences and notify the JavaFX application when a preference has changed.
      * The change notification includes all specified preferences, not only the changed preferences.
      */
-    bool updatePreferences(PreferenceType) const;
+    bool updatePreferences(PreferenceType, bool delayedChangesExpected) const;
 
     /**
      * Handles the WM_SETTINGCHANGE message.
-    */
+     */
     bool onSettingChanged(WPARAM, LPARAM) const;
 
 private:
+    /**
+     * Suggested aggregation delay for changes that come in over a period of time.
+     */
+    static constexpr int SUGGESTED_DELAY_MILLIS = 1000;
+
     JNIEnv* env;
     jobject application;
     bool initialized;
@@ -88,6 +93,7 @@ private:
     void querySystemColors(jobject properties) const;
     void querySystemParameters(jobject properties) const;
     void queryUISettings(jobject properties) const;
+    void queryUIColors(jobject properties) const;
     void queryNetworkInformation(jobject properties) const;
 
     void putString(jobject properties, const char* key, const char* value) const;
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/application/PlatformImpl.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/application/PlatformImpl.java
@@ -28,6 +28,7 @@ package com.sun.javafx.application;
 import com.sun.javafx.PlatformUtil;
 import com.sun.javafx.PreviewFeature;
 import com.sun.javafx.SecurityUtil;
+import com.sun.javafx.application.preferences.DelayedChangeAggregator;
 import com.sun.javafx.application.preferences.PlatformPreferences;
 import com.sun.javafx.application.preferences.PreferenceMapping;
 import com.sun.javafx.css.StyleManager;
@@ -57,6 +58,7 @@ import javafx.application.Application;
 import javafx.application.ConditionalFeature;
 import javafx.beans.property.BooleanProperty;
 import javafx.beans.property.SimpleBooleanProperty;
+import javafx.collections.MapChangeListener;
 import javafx.scene.Scene;
 
 public class PlatformImpl {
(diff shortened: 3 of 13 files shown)


modules/javafx.graphics/src/main/java/com/sun/glass/ui/Application.java
```diff
@@ -84,7 +84,7 @@ public void handleOpenFilesAction(Application app, long time, String files[]) {
         // currently used only on Mac OS X
         public void handleQuitAction(Application app, long time) {
         }
-        public void handlePreferencesChanged(Map<String, Object> preferences) {
+        public void handlePreferencesChanged(Map<String, Object> preferences, int suggestedDelayMillis) {
```
andy-goryachev-oracle, line 87: I wonder if we should always debounce the changes with a short delay?
A delay of 100-150 ms will be acceptable from the user perspective.
What do you think?

modules/javafx.graphics/src/main/native-glass/win/PlatformSupport.h
```diff
@@ -61,14 +61,19 @@ class PlatformSupport final
...
      * Handles the WM_SETTINGCHANGE message.
-    */
+     */
     bool onSettingChanged(WPARAM, LPARAM) const;
 
 private:
+    /**
+     * Suggested aggregation delay for changes that come in over a period of time.
+     */
+    static constexpr int SUGGESTED_DELAY_MILLIS = 1000;
```
andy-goryachev-oracle, line 75: 1 second seems too long.
What is the typical range for the high contrast train of changes?

mstr2: Quite a lot, Windows takes around 1-2 seconds to change to and from a high-contrast theme (there's even a full-frame wait screen that pops up). The suggested delay of 1 second works reliably on my machine.

andy-goryachev-oracle: If it takes that long to update all the colors, maybe we should not add complexity to this subsystem and just send the events as they come?

What exactly is the problem the user experiences right now?

andy-goryachev-oracle:
I am not comfortable with this change, as it falls under application requirements.

The function of the JavaFX platform is to communicate the OS events as faithfully as it can, in my opinion.  If the application requirements call for some debouncing, especially that with a one second delay, then the application is free to implement that.

If, however, the application requirements call for no delay, this change makes it impossible.

mstr2:
> I am not comfortable with this change, as it falls under application requirements.
> 
> The function of the JavaFX platform is to communicate the OS events as faithfully as it can, in my opinion. If the application requirements call for some debouncing, especially that with a one second delay, then the application is free to implement that.

The platform preferences API is an abstraction, and nowhere do we specify that it replicates the exact stream of OS events. What we _do_ (implicitly) specify, however, is that the observable state of platform preferences accurately reflects the OS settings. Do you really think that it is okay for an application to be able to observe inconsistent transient states? In this example, applications can observe inconsistent `foregroundColor` and `backgroundColor` values.

> If, however, the application requirements call for no delay, this change makes it impossible.

An application can require all it wants, but "inconsistent transient states are okay" is not an acceptable trade-off. There's no delay for uncorrelated changes, which is as good as it gets.

drmarmac:
I'm in favor of abstracting this away as good as possible, i.e. trying to aggregate such events if some sufficiently universal timing parameters (or something better than time-based logic) can be found. As an application developer I often choose javafx especially because it keeps OS-implementation dependent behavior to a minimum and reduces the likelihood of me "using things the wrong way", like having inconsistent states that might need to be handled.  And who knows what the next version of windows will do...

modules/javafx.graphics/src/main/native-glass/win/PlatformSupport.h
hjohn: Are you sure a delay is a good way to solve this?  It won't guarantee that no inconsistent states will be observed (what if Windows is slower, or a change is reverted within the delay period -- you may still see "half" changes).

mstr2: I don't think that there's a perfect way to solve this. Yes, this can theoretically happen, and then you'll see a visual glitch. I can imagine that this might also be the reason why Windows shows a full-screen popup when you switch to and from a high-contrast theme: to hide glitchy applications during the transition period.

hjohn:
> I am not comfortable with this change, as it falls under application requirements.
> 
> The function of the JavaFX platform is to communicate the OS events as faithfully as it can, in my opinion

The function of JavaFX is to abstract the platform it runs on, preferably in such a way that one can't distinguish between platforms at all, just like about every other API the JDK offers (ie. NIO, Threads, Swing, etc).  This is one of the reasons we don't expose window handles or DirectX/GL API's to users.

modules/javafx.graphics/src/main/native-glass/win/PlatformSupport.h
hjohn: Alright, just making sure, as you mentioned in another thread seeing inconsistent transient states isn't an acceptable trade-off.  They can still occur, and so users can't rely on them.  In that respect the situation is similar to before this change, other than it being less likely to occur (and perhaps never or rarely in practice).

mstr2: Yes, I think having inconsistent transient states is not an acceptable trade-off, that's why I think reducing them by 99,9% (from guaranteed to happen, to almost hever happening) is sensible. If it does happen anyway, it shouldn't crash your application but show up as a visual glitch (I mean, we're talking about colors). In that sense, you can "rely" on it.

PR #1811, opened 2025-05-16
8357157: Exception thrown from AnimationTimer freezes application

mstr2:
When an exception is thrown from `AnimationTimer::handle`, the JavaFX application freezes. The reason is that the user exception will bubble up into framework code, preventing the normal operation of JavaFX.

The following program demonstrates the defect:

```java
public class FailingAnimationTimer extends Application {
    @Override
        var button = new Button("start timer");
        button.setOnAction(_ -> {
            var timer = new AnimationTimer() {
                @Override
                public void handle(long l) {
                    throw new RuntimeException("foo");
                }
            };

            timer.start();
        });

        var root = new HBox();
        root.getChildren().add(new TextField("test"));
        root.getChildren().add(button);
        stage.setScene(new Scene(root));
        stage.show();
    }
}
```

The solution is to not allow user exceptions to bubble up into animation framework code. If an exception occurs, it is instead sent to the current thread's uncaught exception handler. This is the same thing that we already do for exceptions thrown by invalidation listeners and change listeners.

In addition to that, a failing animation timer has the potential to spam logs, which is why I introduced a cut-off value at 100 exceptions for each individual timer, after which no further exceptions from this particular timer are sent to the uncaught exception handler. After reaching the cut-off value, the following warning is logged:

`WARNING: Too many exceptions thrown by AnimationTimer, ignoring further exceptions. The cut-off number can be set with the system property com.sun.scenario.animation.failingTimerThreshold (current = 100).`

/reviewers 2

---------
### Progress

### Issue
 * [JDK-8357157](https://bugs.openjdk.org/browse/JDK-8357157): Exception thrown from AnimationTimer freezes application (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1811/head:pull/1811` \
`$ git checkout pull/1811`

`$ git checkout pull/1811` \
`$ git pull https://git.openjdk.org/jfx.git pull/1811/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1811`

`$ git pr show -t 1811`

</details>

<a href="https://git.openjdk.org/jfx/pull/1811.diff">https://git.openjdk.org/jfx/pull/1811.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1811#issuecomment-2886926623)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/scenario/animation/AbstractPrimaryTimer.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/scenario/animation/AbstractPrimaryTimer.java
@@ -63,10 +63,9 @@ public abstract class AbstractPrimaryTimer {
     private final int PULSE_DURATION_NS = getPulseDuration(1000000000);
     private final int PULSE_DURATION_TICKS = getPulseDuration((int)TickCalculation.fromMillis(1000));
 
-    // This property is used to control the number of exceptions that can be thrown by a timer callback
-    // before we stop logging them to prevent spamming the log.
-    private static final String FAILING_TIMER_THRESHOLD_PROP = "com.sun.scenario.animation.failingTimerThreshold";
-    public static final int FAILING_TIMER_THRESHOLD = Settings.getInt(FAILING_TIMER_THRESHOLD_PROP, 100);
+    // The number of exceptions that can be thrown by a timer callback before we stop sending
+    // them to the uncaught exception handler to prevent spamming the log.
+    public static final int FAILING_TIMER_THRESHOLD = 100;
 
     // This PropertyChangeListener is added to Settings to listen for changes
     // to the nogap and fullspeed properties.
@@ -386,9 +385,7 @@ public abstract class AbstractPrimaryTimer {
 
                 if (Logging.getJavaFXLogger().isLoggable(System.Logger.Level.WARNING)) {
                     Logging.getJavaFXLogger().warning(
-                        "Too many exceptions thrown by " + type().getSimpleName() + ", ignoring further exceptions." +
-                        " The cut-off number can be set with the system property " + FAILING_TIMER_THRESHOLD_PROP +
-                        " (current = " + FAILING_TIMER_THRESHOLD + ").");
+                        "Too many exceptions thrown by " + type().getSimpleName() + ", ignoring further exceptions.");
                 }
             }
         }
--- a/modules/javafx.graphics/src/main/java/com/sun/scenario/animation/AbstractPrimaryTimer.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/scenario/animation/AbstractPrimaryTimer.java
@@ -63,6 +63,8 @@ public abstract class AbstractPrimaryTimer {
     private final int PULSE_DURATION_NS = getPulseDuration(1000000000);
     private final int PULSE_DURATION_TICKS = getPulseDuration((int)TickCalculation.fromMillis(1000));
 
+    // This property is used to control the number of exceptions that can be thrown by a timer callback
+    // before we stop logging them to prevent spamming the log.
     private static final String FAILING_TIMER_THRESHOLD_PROP = "com.sun.scenario.animation.failingTimerThreshold";
     public static final int FAILING_TIMER_THRESHOLD = Settings.getInt(FAILING_TIMER_THRESHOLD_PROP, 100);
 
@@ -371,8 +373,14 @@ public abstract class AbstractPrimaryTimer {
         void handleException(Throwable e) {
             if (exceptionsThrown < FAILING_TIMER_THRESHOLD) {
                 exceptionsThrown++;
-                Thread thread = Thread.currentThread();
-                thread.getUncaughtExceptionHandler().uncaughtException(thread, e);
+
+                try {
+                    Thread thread = Thread.currentThread();
+                    thread.getUncaughtExceptionHandler().uncaughtException(thread, e);
+                } catch (Throwable ignored) {
+                    // The uncaught exception handler shouldn't throw exceptions, but if it does,
+                    // we will swallow it to prevent it from bubbling up.
+                }
             } else if (exceptionsThrown == FAILING_TIMER_THRESHOLD) {
                 exceptionsThrown++;
 
--- a/modules/javafx.graphics/src/main/java/com/sun/scenario/animation/AbstractPrimaryTimer.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/scenario/animation/AbstractPrimaryTimer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2007, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2007, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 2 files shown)


modules/javafx.graphics/src/main/java/com/sun/scenario/animation/AbstractPrimaryTimer.java
```diff
@@ -308,29 +315,74 @@ protected void timePulseImpl(long now) {
...
+            try {
+                animationTimersSnapshot[i].receiver().handle(now);
+            } catch (Throwable e) {
+                animationTimersSnapshot[i].handleException(e);
             }
-        } finally {
-            animationTimersLocked = false;
         }
+
+        animationTimersLocked = false;
```
andy-goryachev-oracle, line 347: same here

modules/javafx.graphics/src/main/java/com/sun/scenario/animation/AbstractPrimaryTimer.java
```diff
@@ -308,29 +315,74 @@ protected void timePulseImpl(long now) {
...
+            try {
+                receiversSnapshot[i].receiver().timePulse(TickCalculation.fromNano(now));
+            } catch (Throwable e) {
+                receiversSnapshot[i].handleException(e);
             }
-        } finally {
-            receiversLocked = false;
         }
+
+        receiversLocked = false;
```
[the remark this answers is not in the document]
mstr2: I've changed the `handleException` method so it will not throw.

modules/javafx.graphics/src/main/java/com/sun/scenario/animation/AbstractPrimaryTimer.java
```diff
@@ -308,29 +317,80 @@ protected void timePulseImpl(long now) {
...
+        int exceptionsThrown;
+
+        void handleException(Throwable e) {
+            if (exceptionsThrown < FAILING_TIMER_THRESHOLD) {
+                exceptionsThrown++;
+
+                try {
+                    Thread thread = Thread.currentThread();
+                    thread.getUncaughtExceptionHandler().uncaughtException(thread, e);
+                } catch (Throwable ignored) {
```
andy-goryachev-oracle, line 380: any reason not to wrap the whole method in `try`/`catch`?

modules/javafx.graphics/src/main/java/com/sun/scenario/animation/AbstractPrimaryTimer.java
```diff
@@ -79,13 +86,15 @@ public abstract class AbstractPrimaryTimer {
         return null;
     };
 
-    private PulseReceiver receivers[] = new PulseReceiver[2];
+    @SuppressWarnings("unchecked")
+    private ReceiverRecord<PulseReceiver>[] receivers = new ReceiverRecord[2];
```
andy-goryachev-oracle, line 90: weird that there is no way to do it cleanly without `@SuppressWarnings("unchecked")`

BTW, Eclipse shows 5,608 hits if I enable "unchecked generic type operation" warning.  Pretty useless, if you ask me.

modules/javafx.graphics/src/main/java/com/sun/scenario/animation/AbstractPrimaryTimer.java
mstr2: There isn't anything else in there that could throw, except for the logger itself. I don't think it makes sense to try to account for failing loggers, as basically everything will break if that's the case.

modules/javafx.graphics/src/main/java/com/sun/scenario/animation/AbstractPrimaryTimer.java
```diff
@@ -308,29 +317,80 @@ protected void timePulseImpl(long now) {
...
+                animationTimersSnapshot[i].handleException(e);
             }
-        } finally {
-            animationTimersLocked = false;
         }
+
+        animationTimersLocked = false;
     }
 
+    private static abstract class ReceiverRecord<T> {
```
andy-goryachev-oracle, line 352: All this does look complicated, just to limit the number of messages in the log (a noble cause).

What if we just print the first message and ignore the rest?  It might be relatively unlikely that two different failures appeared at the same time, each with its own root cause?

If we just show the first one, we'll remove all this complexity and the system property, and possibly get back to the simpler original code?

What do you think?

mstr2: We're throwing away all exceptions after a certain point (either after the first, or some number of exceptions after that). A potential use case would be debugging a large application that for some reason throws lots of different exceptions from timers. There can't be many applications that do this, because right now that would most likely just freeze the application. I'm okay either way (logging the first exception, or logging the first 100 exceptions with a user-configurable threshold).

mstr2: We could also just keep it with a fixed threshold, but remove the system property.

andy-goryachev-oracle: It just seems unnecessary to add all this complexity.  The first exception is all we need, really - we can skip the 2nd, 3rd, ...  This might rule out the property.

It might be unlikely that two or more completely unrelated failures would happen in the field in a well-tested application [citation needed], I think there is a little value in trying to add this complex logic per timer.  A static boolean might suffice.

mstr2: On the other hand, AnimationTimer is a public API that allows for arbitrary code execution, which always has the potential to fail for any number of reasons (think of calling an unstable API from the timer callback). We can’t rule that out entirely. This is only a fallback in any case, as a well-tested application shouldn’t throw exceptions from timer callbacks in the first place.

mstr2: I've removed the system property, so there's a hard-coded limit of 100 exceptions now.

andy-goryachev-oracle: 100 still looks excessive to me - it's unlikely that we'll have two different scenarios when this happens, so most of the time the log will have 100 identical traces.

I mean, one is probably enough, we could have 2 or 4 just to drive the message home.  Anything beyond 8 is simply annoying, don't you think?

mstr2: In general, we shouldn't suppress any exceptions that would otherwise be logged. The only thing that makes this case a bit different is that with a periodic timer, there's the potential for an inflated log that grows without bounds. I think that defining a reasonable upper limit is enough to prevent that; in particular, I think we shouldn't try to assume that any particular number of logged failures (like 1 or 4) is "enough" for application developers.

andy-goryachev-oracle: well, I still think it's way too high of a number, but the expectation is that once the dev sees a 100 traces in the log they'll fix it right away.

And if some company finds that all of a sudden their client logs exploded in size, they should have invested more in QA.

azuev-java (approved):
Looks reasonable although not having a way to override a threshold is strange. OTOH 100 is a generous enough for the application.

Maran23:
A bit late, but I checked the PR and also had mixed feelings about the error limit stuff (rest is fine!). 
But I actually found out, that there is something similar in Java already: `-Xmaxwarns` and `-Xmaxerrs`, which is also 100 by default (I saw that recently in a Gradle build). 

Which made me feel like this solution is fine and the solution is similar to things which are in Java already for years. At least we do not do something completely new.

PR #1813, opened 2025-05-20
8357393: RichTextArea: fails to properly save text attributes

andy-goryachev-oracle:
Fixing a bug that breaks proper saving of character attributes.

This, unfortunately, makes a breaking change in the RichTextArea native format [0].

## References

[0] https://github.com/andy-goryachev-oracle/Test/blob/main/doc/RichTextArea/RichTextArea_DataFormat_V2.md

---------
### Progress
- [x] Change requires CSR request [JDK-8357505](https://bugs.openjdk.org/browse/JDK-8357505) to be approved

### Issues
 * [JDK-8357393](https://bugs.openjdk.org/browse/JDK-8357393): RichTextArea: fails to properly save text attributes (**Bug** - P4)
 * [JDK-8357505](https://bugs.openjdk.org/browse/JDK-8357505): RichTextArea: fails to properly save text attributes (**CSR**)

### Reviewers
 * [Ziad El Midaoui](https://openjdk.org/census#zelmidaoui) (@Ziad-Mid - Author)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1813/head:pull/1813` \
`$ git checkout pull/1813`

`$ git checkout pull/1813` \
`$ git pull https://git.openjdk.org/jfx.git pull/1813/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1813`

`$ git pr show -t 1813`

</details>

<a href="https://git.openjdk.org/jfx/pull/1813.diff">https://git.openjdk.org/jfx/pull/1813.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1813#issuecomment-2899192427)
</details>


modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttributeMap.java
```diff
@@ -428,7 +428,7 @@ private static void initAccessor() {
         StyleAttributeMapHelper.setAccessor(new StyleAttributeMapHelper.Accessor() {
             @Override
             public StyleAttributeMap filterAttributes(StyleAttributeMap ss, boolean isParagraph) {
-                return ss.filterAttributes(isParagraph);
+                return (ss == null ? null : ss.filterAttributes(isParagraph));
```
kevinrushforth, line 431: You could skip the null check if you disallowed a null map.

modules/jfx.incubator.richtext/src/main/java/com/sun/jfx/incubator/scene/control/richtext/StyleAttributeMapHelper.java
```diff
@@ -52,20 +52,39 @@ public static void setAccessor(Accessor a) {
     /**
      * Returns a new StyleAttributeMap instance which contains only the character attributes,
      * or null if no character attributes found.
+     *
      * @param ss the style attribute map
-     * @return the instance
+     * @return the instance of StyleAttributeMap, or null
```
kevinrushforth, line 57: Is there a good reason to allow null here? Unless there is a semantic difference between null and the empty map, it might be easier to make this non-nullable. This could be done later, since the fact that it can return null is preexisting.

kevinrushforth (approved):
The fix looks reasonable to me. I left a suggestion about whether we really need to allow null, but that's preexisting and could be considered in a follow-up.

Another thing that needs to be done in a follow-up is that you will need to add versioning to the internal rich text format. While this is still an incubating API we can break backward compatibility, but at some point, we will need a format that can evolve compatibility with newer runtimes able to read older files.

kevinrushforth:
@Ziad-Mid Can you be the second reviewer?

/reviewers 2

modules/jfx.incubator.richtext/src/main/java/com/sun/jfx/incubator/scene/control/richtext/StyleAttributeMapHelper.java
andy-goryachev-oracle: I've recorded this in the feedback document - this might be worth doing.

andy-goryachev-oracle:
> While this is still an incubating API we can break backward compatibility, but at some point, we will need a format that can evolve compatibility with newer runtimes able to read older files.

You are absolutely right: we must provide a mechanism for versioning as well as properly document the format.  The draft spec is 
https://github.com/andy-goryachev-oracle/Test/blob/main/doc/RichTextArea/RichTextArea_DataFormat_V2.md

but it is guaranteed to change to enable versioning (and addition of tab stop attributes in #1800)

Ziad-Mid (approved):
LGTM. I’ve tested it on macOS, and the issue is resolved with this fix. It is working as expected.

PR #1814, opened 2025-05-21
8334137: Marlin: replace sun.misc.Unsafe memory access methods with FFM

kevinrushforth:
PR to replace the use of sun.misc.Unsafe memory access methods in the Marlin rasterizer with FFM.

I broke this up into the following commits. The bulk of the work is in the first two:

1. Encapsulate all off-heap access in OffHeapArray -- All of the memory allocation and management was already done in the OffHeapArray class, but the users of OffHeapArray, primarily Renderer and RendererNoAA, called Unsafe methods directly. I encapsulated all of the calls to Unsafe in OffHeapArray. The main impact on calling code is that the base address is no longer accessible or needed. Instead, the `(put|get)(Byte|Int)` methods take only an offset. This commit was straight refactoring with no behavioral changes.
2. Initial FFM implementation -- I changed the memory management and access methods to use FFM. Each OffHeap array uses a shared Arena to manage the single memory segment allocated at construction time. The resize method creates a new Arena and memory segment, copying the data from the old and then closing it
3. Set `used` to 0 in `dispose()` -- While testing and instrumenting the code, I discovered that the Renderer dispose methods resize the edges array back to the default size without clearing the "used" field. The used field will be cleared before the next time it is accessed, but clearing it in dispose allows optimizing resize to not copy any data.
4. Remove '--sun-misc-unsafe-memory-access=allow' from test and app execution, since it is no longer needed. This also enables `-Werror` for the `javafx.graphics` module.
5. ~~Temporary debug prints that will be removed before making this "rfr"~~

Additional commits address review comments.

---------
### Progress

### Issue
 * [JDK-8334137](https://bugs.openjdk.org/browse/JDK-8334137): Marlin: replace sun.misc.Unsafe memory access methods with FFM (**Bug** - P2)

### Reviewers
 * [Laurent Bourgès](https://openjdk.org/census#lbourges) (@bourgesl - **Reviewer**) Review applies to [6c38c604](https://git.openjdk.org/jfx/pull/1814/files/6c38c6040d1abd9b84f3648f892215573a8b1275)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1814/head:pull/1814` \
`$ git checkout pull/1814`

`$ git checkout pull/1814` \
`$ git pull https://git.openjdk.org/jfx.git pull/1814/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1814`

`$ git pr show -t 1814`

</details>

<a href="https://git.openjdk.org/jfx/pull/1814.diff">https://git.openjdk.org/jfx/pull/1814.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1814#issuecomment-2949050618)
</details>


modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -26,82 +26,164 @@
...
-            final Field field = Unsafe.class.getDeclaredField("theUnsafe");
-            field.setAccessible(true);
-            UNSAFE = (Unsafe) field.get(null);
-        } catch (Exception e) {
-            throw new InternalError("Unable to get sun.misc.Unsafe instance", e);
-        }
-
-        SIZE_INT = Unsafe.ARRAY_INT_INDEX_SCALE;
+        // KCR FIXME: get this from FFM
+        SIZE_INT = 4;
```
kevinrushforth, line 50: I will address this.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -26,82 +26,164 @@
...
     }
 
     /* members */
-    long address;
-    long length;
-    int  used;
+    private MemorySegment segment;
+//    private long address;
```
kevinrushforth, line 55: I'll remove this (it's a leftover).

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -26,82 +26,164 @@
...
-import java.lang.reflect.Field;
-import sun.misc.Unsafe;
+import java.lang.foreign.Arena;
+import java.lang.foreign.MemorySegment;
+import java.lang.foreign.ValueLayout;
+import java.nio.ByteOrder;
+
+// KCR: BEGIN DEBUG
+import java.util.concurrent.atomic.AtomicInteger;
+// KCR: END DEBUG
```
kevinrushforth, line 36: I'll revert all of the debugging code.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -26,82 +26,164 @@
...
 
-    // unsafe reference
-    static final Unsafe UNSAFE;
+    private Arena arena;
+
     // size of int / float
     static final int SIZE_INT;
+    // FFM stuff
+    private static final ValueLayout.OfByte BYTE_LAYOUT = ValueLayout.JAVA_BYTE;
+    private static final ValueLayout.OfInt INT_LAYOUT = ValueLayout.JAVA_INT.withOrder(ByteOrder.BIG_ENDIAN);
```
kevinrushforth, line 46: @minborg Is this the best layout to use? I chose it thinking that, since we don't need to access this memory from native code, it might perform better to use Java's big endian byte order. Is this correct?

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -26,82 +26,164 @@
...
+                    ", thread = " + Thread.currentThread());
+        }
+        // KCR: END DEBUG
+
+        arena = Arena.ofShared();
+
         // note: may throw OOME:
-        this.address = UNSAFE.allocateMemory(len);
+        // KCR FIXME: Set a MemoryLayout
+        this.segment = arena.allocate(len);
```
kevinrushforth, line 78: I'll choose a memory layout and also specify 4-byte alignment.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -26,82 +26,164 @@
...
+        // KCR: END DEBUG
+
+        Arena newArena = Arena.ofShared();
+        MemorySegment newSegment = newArena.allocate(len);
+
+        // KCR: Double-check this
+        // If there are any bytes in use, copy them to the newly reallocated array
+        if (this.used > 0) {
+            MemorySegment.copy(segment, 0, newSegment, 0, Math.min(this.used, len));
+        }
```
kevinrushforth, line 145: @bourgesl I presume that there is never a need to copy more than `used` bytes?

modules/javafx.graphics/src/main/java/com/sun/marlin/Renderer.java
```diff
@@ -650,8 +645,12 @@ public void dispose() {
             }
         }
 
+        // KCR: Double-check this
+        // Clear used bytes in edges array
+        edges.setUsed(0);
```
kevinrushforth, line 650: @bourgesl I presume that clearing `used` is correct here?

modules/javafx.graphics/src/main/java/com/sun/marlin/MaskMarlinAlphaConsumer.java
```diff
@@ -105,28 +101,22 @@ public void setMaxAlpha(int maxalpha) {
...
         // AA:
         byte[] _ALPHA_MAP = buildAlphaMap(MarlinConst.MAX_AA_ALPHA);
         ALPHA_MAP = _ALPHA_MAP; // Keep alive the OffHeapArray
         ALPHA_MAP_UNSAFE = new OffHeapArray(ALPHA_MAP, ALPHA_MAP.length); // 1K
 
-        long addr = ALPHA_MAP_UNSAFE.address;
-
         for (int i = 0; i < _ALPHA_MAP.length; i++) {
-            _unsafe.putByte(addr + i, _ALPHA_MAP[i]);
+            ALPHA_MAP_UNSAFE.putByte(i, _ALPHA_MAP[i]);
```
andy-goryachev-oracle, line 110: I know you are trying to keep the diffs to a minimum, but since there is no more Unsafe, can we replace the "UNSAFE" with something more meaningful?

kevinrushforth: Thanks for the reminder. I actually do want to change this. I had originally put a "FIXME" comment to do that, but removed it during the cleanup prior to the draft PR.

My thought was to use something like "OFF_HEAP" rather than a bake the implementation of the off-heap memory into the name.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
bourgesl: True.

modules/javafx.graphics/src/main/java/com/sun/marlin/Renderer.java
bourgesl: Yes. (I skipped that code as edges were left dirty until the next init() call to reset it unconditionally, resize was just a realloc to trim array)

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
bourgesl: It may depend on hardward arch ???

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
bourgesl: Yes 4-bytes min or more like 16 if it helps the compiler to use avx instructions for fill / copy ?

bourgesl:
Few mlnor comments. It looks good to me !

nlisker:
If it helps at this point, we did an initial prototype in the sandbox: https://github.com/openjdk/jfx-sandbox/tree/Marlin_FFM with several commits and reverts, so might need to go through them. What @bourgesl and I weren't sure about was which `Arena` to use. I thought `Confined` was more suitable than `Shared` because there's no multithreaded access.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: Based on some additional checking I did, there is no good reason to set the byte order explicitly, so I'll just use `ValueLayout.JAVA_INT`.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: Reverted.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: I'll set it to 16 then.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: Thanks for confirming.

modules/javafx.graphics/src/main/java/com/sun/marlin/Renderer.java
kevinrushforth: Thanks for confirming.

kevinrushforth:
> If it helps at this point, we did an initial prototype in the sandbox: https://github.com/openjdk/jfx-sandbox/tree/Marlin_FFM with several commits and reverts, so might need to go through them.

Thanks. I took a look at it, and one of the things you did could be good for a follow-up: you use a memory layout that allows accessing data in a more structured manner than as a raw array of bytes or ints. For the initial conversion, I wanted to make minimal changes to the callers of OffHeapMemory, keeping the logic identical. This way everything about the implementation of OffHeap memory is encapsulated in the OffHeapMemory class.

> What @bourgesl and I weren't sure about was which `Arena` to use. I thought `Confined` was more suitable than `Shared` because there's no multithreaded access.

I wondered about that as well. I changed it to `ofConfined()` and don't see any problems so far. I'll run a full set of CI tests and if there are no problems, leave it at that.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -128,17 +105,9 @@ void incrementUsed(int used) {
...
      */
     void resize(final long len) {
-        // KCR: BEGIN DEBUG
-        if (LOG_OFF_HEAP_MALLOC) {
-            System.err.println("OffHeapArray::resize : instance = " + instance + " len = " + len +
-                    " [was: " + this.length + ", used = " + used + "]");
-        }
-        // KCR: END DEBUG
-
         Arena newArena = Arena.ofShared();
```
kevinrushforth: I missed changing this to `ofConfined()`, so will do that.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -25,83 +25,133 @@
...
-    long length;
-    int  used;
+    private MemorySegment segment;
+    private long length;
+    private int used;
 
     OffHeapArray(final Object parent, final long len) {
+        arena = Arena.ofConfined();
+
         // note: may throw OOME:
```
Maran23, line 70: This note is probably outdated? 
You removed it at least below, so probably can do here as well?

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -25,83 +25,133 @@
...
+                                + len + " for segment = " + this.segment);
         }
 
         // Register a cleaning function to ensure freeing off-heap memory:
         MarlinUtils.getCleaner().register(parent, this::free);
     }
 
+    /**
+     * Gets the length of this array.
+     *
```
Maran23: Minor: The javadoc below has no empty line between the description and the return, so maybe remove here as well?

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -25,83 +25,133 @@
...
+     * Sets the number of bytes currently being used. Always <= length
+     * @param used number of used bytes
+     */
+    void setUsed(int used) {
+        this.used = used;
+    }
+
+    /**
+     * Increments the number of bytes currently being used.
+     * Curr used + incr used must be <= length
```
Maran23: Typo? Can be named to: `Current used + increment must be <= length` maybe.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: It's still relevant, so I added back the comment I inadvertently removed.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: done

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: Fixed. I also changed the name of the argument to be `increment` for clarity.

nlisker:
Do you think it's worth asking someone on the Panama team to take a look? It's the first FFM introduction into JavaFX, they might have worthwhile comments. Memory leaks can hide in native memory in a way that's hard to find, such as from improper closure of resources.

bourgesl (approved):
Looks good to me, as it mimics the former Unsafe usage.

kevinrushforth:
> Do you think it's worth asking someone on the Panama team to take a look? It's the first FFM introduction into JavaFX, they might have worthwhile comments. Memory leaks can hide in native memory in a way that's hard to find, such as from improper closure of resources.

That would probably be helpful. I had tagged @minborg earlier and will reach out directly.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -25,83 +25,134 @@
...
+    private static final int ALIGNMENT = 16;
+    private static final ValueLayout.OfByte BYTE_LAYOUT = ValueLayout.JAVA_BYTE;
+    private static final ValueLayout.OfInt INT_LAYOUT = ValueLayout.JAVA_INT;
 
     /* members */
-    long address;
-    long length;
-    int  used;
+    private MemorySegment segment;
+    private long length;
```
andy-goryachev-oracle: just curious: why `long`?  are we expecting the length to exceed 2 billion?

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -25,83 +25,134 @@
...
+     */
+    long getLength() {
+        return length;
+    }
+
+    /**
+     * Gets the number of bytes currently being used. Always <= length
+     * @return number of used bytes
+     */
+    int getUsed() {
```
andy-goryachev-oracle, line 98: wait, length is `long`, but used is `int`?

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -25,83 +25,134 @@
...
+        this.used = used;
+    }
+
+    /**
+     * Increments the number of bytes currently being used.
+     * Current used + increment must be <= length
+     * @param increment number of used bytes to increment
+     */
+    void incrementUsed(int increment) {
+        this.used += increment;
```
andy-goryachev-oracle, line 116: if used > 2B, it will overflow, right?

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -25,83 +25,134 @@
...
+    }
+
+    byte getByte(long offset) {
+        return segment.get(BYTE_LAYOUT, offset);
+    }
+
+    int getInt(long offset) {
+        return segment.get(INT_LAYOUT, offset);
     }
+
```
andy-goryachev-oracle: extra newline?

modules/javafx.graphics/src/main/java/com/sun/marlin/Renderer.java
```diff
@@ -399,9 +395,8 @@ private void addLine(double x1, double y1, double x2, double y2) {
         }
 
 
-        final Unsafe _unsafe = OffHeapArray.UNSAFE;
         final long SIZE_INT = 4L;
-        long addr   = _edges.address + edgePtr;
+        long addr   = edgePtr;
```
andy-goryachev-oracle, line 399: minor: extra spaces?

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: This is preexisting. The goal was to keep the same interface that the users of OffHeapMemory currently use, so I do not want to change or revisit this. Other than encapsulation, I made no changes to the OffHeap interface.

modules/javafx.graphics/src/main/java/com/sun/marlin/RendererNoAA.java
```diff
@@ -372,15 +368,15 @@ private void addLine(double x1, double y1, double x2, double y2) {
...
 
         // use substraction to avoid integer overflow:
-        if (_edges.length - edgePtr < _SIZEOF_EDGE_BYTES) {
+        if (_edges.getLength() - edgePtr < _SIZEOF_EDGE_BYTES) {
             // suppose _edges.length > _SIZEOF_EDGE_BYTES
             // so doubling size is enough to add needed bytes
             // note: throw IOOB if neededSize > 2Gb:
             final long edgeNewSize = ArrayCacheConst.getNewLargeSize(
-                                        _edges.length,
+                                        _edges.getLength(),
```
andy-goryachev-oracle, line 379: L377 is unclear - is this a TODO to throw IOOBE or a description of what would happen?
(this is a separate issue from FFM changes, really)

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: I noticed this too. This is preexisting, so I don't want to change it as part of this PR (to minimize changes). It might or might not be worth a follow-up.

andy-goryachev-oracle:
I am getting this printed to stderr when launching the monkey tester:

```
WARNING: java.lang.System::load has been called by com.sun.glass.utils.NativeLibLoader in module javafx.graphics (file:/Users/angorya/Projects/jfx3/jfx/rt/modules/javafx.graphics/bin/)
```

why?

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: Yes. It seems worth checking and throw an exception. Other invariants could also be enforced such as used <= length.

I'll file a follow-up issue for this, since this behavior is preexisting.

modules/javafx.graphics/src/main/java/com/sun/marlin/Renderer.java
kevinrushforth: Preexisting. I don't want to reformat it.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -25,83 +25,134 @@
...
+        this.used += increment;
+    }
+
     /*
      * As realloc may change the address, updating address is MANDATORY
      * @param len new array length
      * @throws OutOfMemoryError if the allocation is refused by the system
      */
     void resize(final long len) {
+        Arena newArena = Arena.ofConfined();
```
minborg, line 129: The comment "updating address is MANDATORY" can be removed.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
minborg: Maybe we can remove the `length` field and instead use `segment.byteSize()`?

modules/javafx.graphics/src/main/java/com/sun/marlin/RendererNoAA.java
kevinrushforth: Yes, this is preexisting and not something to look at as part of this PR.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -25,83 +25,134 @@
...
                                 + ": OffHeapArray.reallocateMemory = "
-                                + len + " to addr = " + this.address);
+                                + len + " for segment = " + this.segment);
         }
     }
 
     void free() {
-        UNSAFE.freeMemory(this.address);
-        if (LOG_UNSAFE_MALLOC) {
+        arena.close();
```
andy-goryachev-oracle, line 151: since we don't assign arena to `null`, I assume that the code expects:

1. `free()` will be only called once
2. no access to this OffHeapArray happens after `free()` is called

right?

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: removed

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: Correct. Note that `free` is only called by the cleaner once the user of the offheap array is no longer using it (i.e., when its cleaner object is unreachable).

An exception would occur if `free` were called more than once or if there was an attempt to access the memory segment after the call to `free`.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: Thanks. I've done this.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: Good catch. I've remove this and added a better description.

kevinrushforth:
> ```
> ...
> WARNING: Use --enable-native-access=javafx.graphics to avoid a warning for callers in this module
> WARNING: Restricted methods will be blocked in a future release unless native access is enabled
> ```
> 
> why?

Because you forgot to pass `--enable-native-access=javafx.graphics`

See:
https://github.com/openjdk/jfx/blob/master/doc-files/release-notes-24.md#javafx-applications-must-use---enable-native-access

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
andy-goryachev-oracle: Question: can the number of bytes to be copied depend on memory layout (BE vs LE)?

andy-goryachev-oracle:
> --enable-native-access=javafx.graphics

I did indeed forget.  The `sun.misc.Unsafe` warning is no longer printed.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: No

andy-goryachev-oracle (approved):
Thank you for clarifications, Kevin.

LGTM, apart from the possible int/long overflow, which is also in the original code (pre-existing, as you said).

modules/javafx.graphics/src/main/java/com/sun/marlin/MaskMarlinAlphaConsumer.java
```diff
@@ -92,41 +88,35 @@ public MaskData getMaskData() {
...
         byte[] _ALPHA_MAP = buildAlphaMap(MarlinConst.MAX_AA_ALPHA);
         ALPHA_MAP = _ALPHA_MAP; // Keep alive the OffHeapArray
-        ALPHA_MAP_UNSAFE = new OffHeapArray(ALPHA_MAP, ALPHA_MAP.length); // 1K
-
-        long addr = ALPHA_MAP_UNSAFE.address;
+        ALPHA_MAP_OFF_HEAP = new OffHeapArray(ALPHA_MAP, ALPHA_MAP.length); // 1K
 
         for (int i = 0; i < _ALPHA_MAP.length; i++) {
-            _unsafe.putByte(addr + i, _ALPHA_MAP[i]);
+            ALPHA_MAP_OFF_HEAP.putByte(i, _ALPHA_MAP[i]);
```
prrace, line 110: I guess this is only done once, but following multiple layers of calculation I think this array defaults to 64K long, in which case a bulk copy would be better, but it is perhaps out-of-scope for a 1:1 replacement

https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/lang/foreign/MemorySegment.html#copy(java.lang.Object,int,java.lang.foreign.MemorySegment,java.lang.foreign.ValueLayout,long,int)

kevinrushforth: Thanks. This seems like a good follow-up RFE. I'll file it.

kevinrushforth:
Thanks for the reviews. I'll leave this until tomorrow in case there are other comments.

prrace:
> > If it helps at this point, we did an initial prototype in the sandbox: https://github.com/openjdk/jfx-sandbox/tree/Marlin_FFM with several commits and reverts, so might need to go through them.
> 
> Thanks. I took a look at it, and one of the things you did could be good for a follow-up: you use a memory layout that allows accessing data in a more structured manner than as a raw array of bytes or ints. For the initial conversion, I wanted to make minimal changes to the callers of OffHeapMemory, keeping the logic identical. This way everything about the implementation of OffHeap memory is encapsulated in the OffHeapMemory class.
> 
> > What @bourgesl and I weren't sure about was which `Arena` to use. I thought `Confined` was more suitable than `Shared` because there's no multithreaded access.
> 
> I wondered about that as well. I changed it to `ofConfined()` and don't see any problems so far. I'll run a full set of CI tests and if there are no problems, leave it at that.

Multi-threaded access in the sense of the not being used concurrently by different threads ?
What FFM means is that the thread that creates the Arena is the only thread ever allowed to use the Arena or any memory allocated by it.
I don't quite see how you guarantee a static field  of a class is never accessed by another thread.
And all it takes is for some other thread to have initialize the class and boom

And how in fact is Marlin ensuring only one thread is ever used to run Marlin ?
I didn't think it did anything of the kind. Is this just a case of the FX thread always runs it ?

kevinrushforth:
> Multi-threaded access in the sense of the not being used concurrently by different threads ? What FFM means is that the thread that creates the Arena is the only thread ever allowed to use the Arena or any memory allocated by it. I don't quite see how you guarantee a static field of a class is never accessed by another thread. And all it takes is for some other thread to have initialize the class and boom
> 
> And how in fact is Marlin ensuring only one thread is ever used to run Marlin ? I didn't think it did anything of the kind. Is this just a case of the FX thread always runs it ?

The QuantumRenderer thread uses Marlin during rendering and the JavaFX application thread uses it during picking.

@bourgesl can provide a more complete answer, but from what I see, all accesses to OffHeapArray are done via an instance of the RendererContext class. Marlin ensures that a RendererContext instance is created and accessed on the same thread by using thread local state and using the rendering context for the thread in question (or creating it if it is the first time). See [DMarlinRenderingEngine::getRendererContext](https://github.com/openjdk/jfx/blob/b22ed6d202f7bb3765fa402a748d455f6316f7c9/modules/javafx.graphics/src/main/java/com/sun/marlin/DMarlinRenderingEngine.java#L254).

prrace:
> > Multi-threaded access in the sense of the not being used concurrently by different threads ? What FFM means is that the thread that creates the Arena is the only thread ever allowed to use the Arena or any memory allocated by it. I don't quite see how you guarantee a static field of a class is never accessed by another thread. And all it takes is for some other thread to have initialize the class and boom
> > And how in fact is Marlin ensuring only one thread is ever used to run Marlin ? I didn't think it did anything of the kind. Is this just a case of the FX thread always runs it ?
> 
> The QuantumRenderer thread uses Marlin during rendering and the JavaFX application thread uses it during picking.
> 
> @bourgesl can provide a more complete answer, but from what I see, all accesses to OffHeapArray are done via an instance of the RendererContext class. Marlin ensures that a RendererContext instance is created and accessed on the same thread by using thread local state and using the rendering context for the thread in question (or creating it if it is the first time). See [DMarlinRenderingEngine::getRendererContext](https://github.com/openjdk/jfx/blob/b22ed6d202f7bb3765fa402a748d455f6316f7c9/modules/javafx.graphics/src/main/java/com/sun/marlin/DMarlinRenderingEngine.java#L254).

I'd looked at that but it was not obvious. And it definitely is not obvious that it will be the same thread that initialises the MaskMarlinAlphaConsumer.java class

andy-goryachev-oracle:
> I don't quite see how you guarantee a static field of a class is never accessed by another thread.

Wouldn't a WrongThreadException be thrown if that happens?

kevinrushforth:
> > > Multi-threaded access in the sense of the not being used concurrently by different threads ? What FFM means is that the thread that creates the Arena is the only thread ever allowed to use the Arena or any memory allocated by it. I don't quite see how you guarantee a static field of a class is never accessed by another thread. And all it takes is for some other thread to have initialize the class and boom
> > > And how in fact is Marlin ensuring only one thread is ever used to run Marlin ? I didn't think it did anything of the kind. Is this just a case of the FX thread always runs it ?
> > 
> > 
> > The QuantumRenderer thread uses Marlin during rendering and the JavaFX application thread uses it during picking.
> > @bourgesl can provide a more complete answer, but from what I see, all accesses to OffHeapArray are done via an instance of the RendererContext class. Marlin ensures that a RendererContext instance is created and accessed on the same thread by using thread local state and using the rendering context for the thread in question (or creating it if it is the first time). See [DMarlinRenderingEngine::getRendererContext](https://github.com/openjdk/jfx/blob/b22ed6d202f7bb3765fa402a748d455f6316f7c9/modules/javafx.graphics/src/main/java/com/sun/marlin/DMarlinRenderingEngine.java#L254).
> 
> I'd looked at that but it was not obvious. And it definitely is not obvious that it will be the same thread that initialises the MaskMarlinAlphaConsumer.java class

Yes, I see what you are saying. The `ALPHA_MASK_XXX` arrays are static fields of `MaskMarlinAlphaConsumer`, and the renderer context class has an instance field of `MaskMarlinAlphaConsumer`. So it does seem possible that `MaskMarlinAlphaConsumer` could be initialized by a thread other than the prism renderer thread. Hmmm.

kevinrushforth:
> > I don't quite see how you guarantee a static field of a class is never accessed by another thread.
> 
> Wouldn't a WrongThreadException be thrown if that happens?

Yes. If we had 100% complete test coverage for all possible use cases, the fact that we don't see such an exception would give us confidence, but we don't. So we either need to prove that it can't happen, or we will need to go back to using a shared arena -- at least for the two static off-heap arrays.

kevinrushforth:
> > I'd looked at that but it was not obvious. And it definitely is not obvious that it will be the same thread that initialises the MaskMarlinAlphaConsumer.java class
> 
> Yes, I see what you are saying. The `ALPHA_MASK_XXX` arrays are static fields of `MaskMarlinAlphaConsumer`, and the renderer context class has an instance field of `MaskMarlinAlphaConsumer`. So it does seem possible that `MaskMarlinAlphaConsumer` could be initialized by a thread other than the prism renderer thread. Hmmm.

Fortunately, the presence of following field in RendererConext does not cause the MaskMarlinAlphaConsumer class to be initialized:

```
    public MaskMarlinAlphaConsumer consumer = null;
```

I instrumented the code, adding something that _did_ force initialization and was able to provoke a WrongThreadException. Without my modification, MaskMarlinAlphaConsumer doesn't get initialized until rendering.

This seems _very_ fragile, though. At best, presuming we can prove that initialization and access to MaskMarlinAlphaConsumer always happens on the Prism rendering thread, this is an accident waiting to happen.

@bourgesl Do you have any thoughts on this? Using the shared arena is definitely safer, but there might be a performance penalty.

andy-goryachev-oracle:
the `MaskMarlinAlphaConsumer` field in `RendererContext` is created/resized in DMarlinRasterizer:102

perhaps we could refactor this code to hide the public field and ensure the single threaded access?

kevinrushforth:
> the `MaskMarlinAlphaConsumer` field in `RendererContext` is created/resized in DMarlinRasterizer:102
> 
> perhaps we could refactor this code to hide the public field and ensure the single threaded access?

That would get into the sort of changes I'd prefer not to make, at least not as part of the initial Unsafe --> FFM conversion, although Laurent could comment on that.

Other possibilities include: using thread-local storage for the ALPHA_MAP_XXX off heap arrays (in practice, they are only used from the Quantum renderer thread, so there would only ever be one copy) or dispensing with off-heap arrays entirely for the ALPHA_MAP, and using plain Java arrays. Given how these static arrays are used, I'm not sure using off-heap memory is really buying much in terms of performance (as opposed to the dynamic edges array, which is the primary use of off-heap memory).

Really, though, the simplest thing to do for now is to use shared (rather than confined) access for these two arrays and file a follow-up to optimize it if needed.

andy-goryachev-oracle:
> Really, though, the simplest thing to do for now is to use shared

I agree.

kevinrushforth:
It turns out there is an even simpler solution: defer the initialization of the static OffHeapArray objects to the constructor of `MaskMarlinAlphaConsumer`. Only the QuantumRenderer thread will construct an instance of `MaskMarlinAlphaConsumer`.
 I verified that even in the case where I force the class to be initialized on the wrong thread, it works correctly.

I'll do some additional testing and then push a fix.

kevinrushforth:
@bourgesl @andy-goryachev-oracle I pushed the fix. Please re-review.

bourgesl (approved):

andy-goryachev-oracle (approved):

bourgesl:
> > > I'd looked at that but it was not obvious. And it definitely is not obvious that it will be the same thread that initialises the MaskMarlinAlphaConsumer.java class
> > 
> > Yes, I see what you are saying. The `ALPHA_MASK_XXX` arrays are static fields of `MaskMarlinAlphaConsumer`, and the renderer context class has an instance field of `MaskMarlinAlphaConsumer`. So it does seem possible that `MaskMarlinAlphaConsumer` could be initialized by a thread other than the prism renderer thread. Hmmm.
> 
> Fortunately, the presence of following field in RendererConext does not cause the MaskMarlinAlphaConsumer class to be initialized:
> 
> ```
>     public MaskMarlinAlphaConsumer consumer = null;
> ```
> 
> I instrumented the code, adding something that _did_ force initialization and was able to provoke a WrongThreadException. Without my modification, MaskMarlinAlphaConsumer doesn't get initialized until rendering.
> 
> This seems _very_ fragile, though. At best, presuming we can prove that initialization and access to MaskMarlinAlphaConsumer always happens on the Prism rendering thread, this is an accident waiting to happen.
> 
> @bourgesl Do you have any thoughts on this? Using the shared arena is definitely safer, but there might be a performance penalty.

These alpha masks are constant (static), so reused in marlin java2d where multiple threads & rendering contexts can work in parallel.
With FFM, the global arena could be used, but the current patch is good enough as javafx uses only 1 thread to perform rendering.
I hope using the shared arena does not hurt performance in this case (1 thread).

kevinrushforth:
> These alpha masks are constant (static), so reused in marlin java2d where multiple threads & rendering contexts can work in parallel. With FFM, the global arena could be used, but the current patch is good enough as javafx uses only 1 thread to perform rendering. I hope using the shared arena does not hurt performance in this case (1 thread).

By deferring the creation to the first time a MaskMarlinAlphaConsumer object is constructed, I was able to continue using the confined arena, so there is no use of a shared arena in Marlin for JavaFX.

I had also thought about the global arena, since the alpha objects are global (static) objects that are never released. That might be a better choice, and is probably what you will need to do anyway for Java2D. If there is no performance penalty for accessing data from a memory segment created using the global arena, then we could switch to that for JavaFX as well.

@minborg Do you have any advice regarding this?

minborg:
> If there is no performance penalty for accessing data from a memory segment created using the global arena, then we could switch to that for JavaFX as well.

Using memory from the Global arena can actually be _faster_ as we do not have to perform any liveness checks for such segments. So, I think this is a good move. Also, the memory does not have to be acquired/released if the address is conveyed to native code (as the segment's scope never closes).

kevinrushforth:
I was going to file a follow-on issue for switching to the global arena for the alpha maps, but I now think it is better to do that now. The changes are simple. My initial testing shows that it works quite well. I'll make the change in this PR shortly, after some further testing.

nlisker:
> > If there is no performance penalty for accessing data from a memory segment created using the global arena, then we could switch to that for JavaFX as well.
> 
> Using memory from the Global arena can actually be _faster_ as we do not have to perform any liveness checks for such segments. So, I think this is a good move. Also, the memory does not have to be acquired/released if the address is conveyed to native code (as the segment's scope never closes).

In this case, you'll need to change [`free()`](https://github.com/kevinrushforth/jfx/blob/2dd7a35d2eec16b04de1f4cb7aeba31be5d98a55/modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java#L126) to remove the closure of the arena, but I think that then you'll have to manually clear the memory segments it allocates. This requires non-trivial modifications with `MarlinUtils`'s cleaner. When @bourgesl and I discussed these changes, we were unsure what the correct arena for this situation is.

kevinrushforth:
> > > If there is no performance penalty for accessing data from a memory segment created using the global arena, then we could switch to that for JavaFX as well.
> > 
> > 
> > Using memory from the Global arena can actually be _faster_ as we do not have to perform any liveness checks for such segments. So, I think this is a good move. Also, the memory does not have to be acquired/released if the address is conveyed to native code (as the segment's scope never closes).
> 
> In this case, you'll need to change [`free()`](https://github.com/kevinrushforth/jfx/blob/2dd7a35d2eec16b04de1f4cb7aeba31be5d98a55/modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java#L126) to remove the closure of the arena, but I think that then you'll have to manually clear the memory segments it allocates. This requires non-trivial modifications with `MarlinUtils`'s cleaner. When @bourgesl and I discussed these changes, we were unsure what the correct arena for this situation is.

The alpha map objects are never freed. They are static arrays that are created and filled once, in the static initializer of the `MaskMarlinAlphaConsumer` class. The change I am thinking to propose is [here](https://github.com/kevinrushforth/jfx/commit/6c38c6040d1abd9b84f3648f892215573a8b1275) if you want to take a look before I push it to this PR branch.

I could still integrate this PR "as is" and file a follow-on for using the global arena for the alpha map arrays depending on what @nlisker and @bourgesl think.

modules/javafx.graphics/src/main/java/com/sun/marlin/MaskMarlinAlphaConsumer.java
```diff
@@ -92,41 +92,35 @@ public MaskData getMaskData() {
...
+    static OffHeapArray ALPHA_MAP_OFF_HEAP_NO_AA;
 
+    static void initAlphaMaps() {
         // AA:
         byte[] _ALPHA_MAP = buildAlphaMap(MarlinConst.MAX_AA_ALPHA);
         ALPHA_MAP = _ALPHA_MAP; // Keep alive the OffHeapArray
-        ALPHA_MAP_UNSAFE = new OffHeapArray(ALPHA_MAP, ALPHA_MAP.length); // 1K
-
-        long addr = ALPHA_MAP_UNSAFE.address;
+        ALPHA_MAP_OFF_HEAP = new OffHeapArray(ALPHA_MAP, ALPHA_MAP.length); // 1K
```
andy-goryachev-oracle: L110: is `ALPHA_MAP` needed to keep the `parent` reference in `OffHeapArray`?  if so, why not use a single `Object` instead of an array?

nlisker:
> > > > If there is no performance penalty for accessing data from a memory segment created using the global arena, then we could switch to that for JavaFX as well.
> > > 
> > > 
> > > Using memory from the Global arena can actually be _faster_ as we do not have to perform any liveness checks for such segments. So, I think this is a good move. Also, the memory does not have to be acquired/released if the address is conveyed to native code (as the segment's scope never closes).
> > 
> > 
> > In this case, you'll need to change [`free()`](https://github.com/kevinrushforth/jfx/blob/2dd7a35d2eec16b04de1f4cb7aeba31be5d98a55/modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java#L126) to remove the closure of the arena, but I think that then you'll have to manually clear the memory segments it allocates. This requires non-trivial modifications with `MarlinUtils`'s cleaner. When @bourgesl and I discussed these changes, we were unsure what the correct arena for this situation is.
> 
> The alpha map objects are never freed. They are static arrays that are created and filled once, in the static initializer of the `MaskMarlinAlphaConsumer` class. The change I am thinking to propose is [here](https://github.com/kevinrushforth/jfx/commit/6c38c6040d1abd9b84f3648f892215573a8b1275) if you want to take a look before I push it to this PR branch.
> 
> I could still integrate this PR "as is" and file a follow-on for using the global arena for the alpha map arrays depending on what @nlisker and @bourgesl think.

Ah, you're using **2** arenas for the different allocations. That's something I didn't think of and seems like a decent solution.

I'm fine with any solution that keeps the rendering correct (obviously) and addresses the concerns of memory deallocation, which as I said, can be tricky to do with native memory.

modules/javafx.graphics/src/main/java/com/sun/marlin/MaskMarlinAlphaConsumer.java
kevinrushforth: It's actually unused (in practice), but I didn't want to change the logic any more than I had to. It doesn't matter enough to care about, so I didn't want to think about whether the reference could be eliminated.

kevinrushforth:
I think I'll proceed with this solution, using a global arena for the static alpha maps (never freed), and the confined arena for everything else.

bourgesl (approved):

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -25,83 +25,173 @@
...
                                 + ": OffHeapArray.reallocateMemory = "
-                                + len + " to addr = " + this.address);
+                                + len + " for segment = " + this.segment);
         }
     }
 
     void free() {
-        UNSAFE.freeMemory(this.address);
-        if (LOG_UNSAFE_MALLOC) {
+        if (global) {
```
nlisker: Trying to close the global arena results in an UnsupportedOperationException according to the docs. Not sure doing this manually adds anything.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -46,8 +47,40 @@ final class OffHeapArray {
...
+     * Creates an OffHeapArray of the specified length using a confined arena.
+     * All access to this array must be done on the same thread that constructed it.
+     *
+     * @param parent the object that will be used to register a cleaner to
+     * free the off-heap array when {@code parent} becomes phantom reachable
+     * @param len the number of bytes to allocate
+     */
     OffHeapArray(final Object parent, final long len) {
-        arena = Arena.ofConfined();
+        this(parent, len, false);
```
andy-goryachev-oracle: I would suggest to remove this constructor.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
prrace: true

kevinrushforth: Right. I could remove it if you like, although it isn't hurting anything leaving it there.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: Since global is the special case, it seemed easier to keep the existing constructor and only pass "global=true" in the one place that's needed -- the static initializer of `MaskMarlinAlphaConsumer`.

kevinrushforth: Although... there is only one use now of the default constructor, so it would be trivial to change it and dispense with the first constructor.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
```diff
@@ -25,83 +25,173 @@
...
+     * confined arena
+     */
+    OffHeapArray(final Object parent, final long len, boolean global) {
+        this.global = global;
+
+        if (global) {
+            arena = Arena.global();
+        } else {
+            arena = Arena.ofConfined();
+        }
```
nlisker: If you prefer, this can be a ternary `arena = global ? ...`.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
andy-goryachev-oracle: it's a minor thing;  we could do it later if you plan a follow-up - I more dislike the idea of having an unused array...

andy-goryachev-oracle:
My (limited) testing on macOS 15.5 M1 shows no ill effects.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
nlisker: It doesn't really matter, but personally I try to use the built in functionality.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: I'll file a few cleanup bugs before I integrate and note them in comments.

kevinrushforth:
I ran a full CI test build after the latest changes to use the global arena for the alpha maps. No unexpected failures.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: ~~I decided to fix the constructor now (simple cleanup without any behavior change).~~

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: Done.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: I removed the explicit check.

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: I decided to fix the constructor now (simple cleanup without any behavior change).

modules/javafx.graphics/src/main/java/com/sun/marlin/RendererNoAA.java
kevinrushforth: [JDK-8359259](https://bugs.openjdk.org/browse/JDK-8359259)

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: [JDK-8359259](https://bugs.openjdk.org/browse/JDK-8359259)

modules/javafx.graphics/src/main/java/com/sun/marlin/OffHeapArray.java
kevinrushforth: [JDK-8359259](https://bugs.openjdk.org/browse/JDK-8359259)

modules/javafx.graphics/src/main/java/com/sun/marlin/MaskMarlinAlphaConsumer.java
kevinrushforth: [JDK-8359260](https://bugs.openjdk.org/browse/JDK-8359260)

modules/javafx.graphics/src/main/java/com/sun/marlin/MaskMarlinAlphaConsumer.java
kevinrushforth: [JDK-8359260](https://bugs.openjdk.org/browse/JDK-8359260)

kevinrushforth:
I filed the following two cleanup issues:

[JDK-8359259](https://bugs.openjdk.org/browse/JDK-8359259): Missing range checks in Marlin for JavaFX
[JDK-8359260](https://bugs.openjdk.org/browse/JDK-8359260): Optimizations for Marlin FFM in JavaFX

kevinrushforth:
Thanks for the reviews.

/integrate

samypr100:
@kevinrushforth More of a follow-up question rather than a bug report yet (apologies if this is not the right avenue). I found usage of [sun.misc.BASE64Encoder](https://github.com/openjdk/jfx/blob/c77c2335856a967907aaacc9546f44943c069add/modules/javafx.web/src/android/java/javafx/scene/web/WebEngine.java#L478) in `javafx.web` module as well as [jdk.unsupported](https://github.com/openjdk/jfx/blob/c77c2335856a967907aaacc9546f44943c069add/modules/javafx.graphics/src/main/java/module-info.java#L45) still referenced in javafx.graphics.

Would those need to be addressed as well (assuming the end goal was to remove all sun.misc references)? I'm also naively assuming the requires on `jdk.unsupported` can be dropped after these changes unless I'm mistaken.

kevinrushforth:
> More of a follow-up question rather than a bug report yet (apologies if this is not the right avenue).

I'll give a brief answer here. Any follow-up discussion can happen on the mailing list.

> I found usage of [sun.misc.BASE64Encoder](https://github.com/openjdk/jfx/blob/c77c2335856a967907aaacc9546f44943c069add/modules/javafx.web/src/android/java/javafx/scene/web/WebEngine.java#L478) in `javafx.web` module

The `src/android` directory under the `javafx.web` module is unused by the JavaFX desktop build, and requires additional stuff outside the jfx repo to build it (I've never built it myself). Since `sun.misc.BASE64Encoder` no longer exists in the JDK -- and hasn't since it was removed in JDK 9 -- that copy of `WebEngine.java` cannot possibly compile unmodified, at least not with an unmodified JDK.

> as well as [jdk.unsupported](https://github.com/openjdk/jfx/blob/c77c2335856a967907aaacc9546f44943c069add/modules/javafx.graphics/src/main/java/module-info.java#L45) still referenced in javafx.graphics.
> 
> Would those need to be addressed as well (assuming the end goal was to remove all sun.misc references)? I'm also naively assuming the requires on `jdk.unsupported` can be dropped after these changes unless I'm mistaken.

Thank you for pointing this out. This was an oversight on my part. I should have removed the `requires jdk.unsupported` as part of this PR. I filed [JDK-8371087](https://bugs.openjdk.org/browse/JDK-8371087) as a follow-on to do this.

PR #1817, opened 2025-05-27
8357594: Additional geometry-based Text/TextFlow APIs

andy-goryachev-oracle:
## Summary
This change adds new methods to the `TextFlow` which work correctly in the presence of non-empty insets (borders/padding). For backward compatibility, the old buggy methods are getting deprecated (not for removal). Also, adds new methods in Text which provide missing functionality.

## Problem
A number of methods in `TextFlow` fail to return correct values in the presence of non-empty insets (i.e. when either padding or border are set):
- caretShape 
- hitTest 
- rangeShape

Additionally, the current API fail to provide strike-through shape, and account for line spacing in the range shape, a problem shared between the `TextFlow` and the `Text` classes.

## Solution
The solution is two-fold: 
1) deprecate the buggy methods (not for removal), adding explanations in their javadoc comments 
2) add new methods that behave correctly in the presence of non-empty insets and/or implementing the missing functionality.

The proposed solution retains the buggy methods for the purposes of backward compatibility in applications which employ the workarounds, while providing new APIs with additional parameters similar to those offered by the new TextLayout API https://bugs.openjdk.org/browse/JDK-8341670

## Testing

Additional visualization of the data returned by the new APIs is available in the Monkey Tester using the following branch (in the Text and TextFlow pages):

https://github.com/andy-goryachev-oracle/MonkeyTest/tree/text.insets.corrected

---------
### Progress
- [x] Change requires CSR request [JDK-8358000](https://bugs.openjdk.org/browse/JDK-8358000) to be approved

### Issues
 * [JDK-8357594](https://bugs.openjdk.org/browse/JDK-8357594): Additional geometry-based Text/TextFlow APIs (**Enhancement** - P4)
 * [JDK-8358000](https://bugs.openjdk.org/browse/JDK-8358000): Additional geometry-based Text/TextFlow APIs (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1817/head:pull/1817` \
`$ git checkout pull/1817`

`$ git checkout pull/1817` \
`$ git pull https://git.openjdk.org/jfx.git pull/1817/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1817`

`$ git pr show -t 1817`

</details>

<a href="https://git.openjdk.org/jfx/pull/1817.diff">https://git.openjdk.org/jfx/pull/1817.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1817#issuecomment-2980975565)
</details>


andy-goryachev-oracle:
@Ziad-Mid and @arapte could you take a look at this PR please?

Ziad-Mid:
I tried it, the tests pass successfully however when trying in MonkeyTester I am not able to show characters and caret path when I check the box it throws an error :
`Exception in thread "JavaFX Application Thread" java.lang.NullPointerException: Cannot invoke "com.oracle.tools.fx.monkey.util.TextShapeLogic.ordinal()" because "logic" is null`

(https://github.com/andy-goryachev-oracle/MonkeyTest/tree/text.insets.corrected)

andy-goryachev-oracle:
> when trying in MonkeyTester

Please update (git pull) the monkey tester and retry (or just select a value in the "API:" combo box.

mstr2:
I'm a bit skeptical with this approach. You describe the existing methods as "buggy". If that's the case, being good stewards of the API should point us towards fixing the bugs instead of adding new methods that seem confusingly similar.

The JBS ticket includes the following information:

> This bug might present a compatibility risk for existing applications which tried to compensate for it by getting the borders and padding from the corresponding Nodes and adding it to each path element (or setting translateX/Y).

Is this just a guess, or do you have any data to corroborate the problem? Messing up the API and baking "buggy" implementations into JavaFX forever should require quite a bit of justification.

Have alternatives been discussed? For example, we could have a system property to fall back to the old behavior.

andy-goryachev-oracle:
> Have alternatives been discussed?

Yes, internally.  This PR has been out for quite a while, too.

The priority here is not to break the existing applications that might have coded a workaround.  We offer a new set of (better named) APIs withe the right behavior, while deprecating (not for removal) the existing ones.

Adding a system property seems a worse choice since it will break the apps and force the change.

mstr2:
> > Have alternatives been discussed?
> 
> Yes, internally. This PR has been out for quite a while, too.

Okay... the OpenJFX project states in CONTRIBUTING.md, section "New features / API additions":
_Discuss the proposed feature on the [openjfx-dev](https://mail.openjdk.org/mailman/listinfo/openjfx-dev) mailing list._

I'm not sure how "internal" discussions qualify. Oracle is not the only stakeholder in this project.

> The priority here is not to break the existing applications that might have coded a workaround.

My question was whether this is an educated guess, or whether this problem is corroborated by data.

andy-goryachev-oracle:
Point taken.  Posted to the mailing list.

> My question was whether this is an educated guess

It's an educated guess.

modules/javafx.graphics/src/main/java/javafx/scene/text/TextFlow.java
```diff
@@ -193,11 +193,15 @@ private void checkOrientation() {
 
     /**
      * Maps local point to {@link HitInfo} in the content.
+     * <p>
+     * NOTE: this method may return incorrect value with non-empty border or padding.
```
kevinrushforth: You should be more explicit, and say that it will not take border or padding into account.

modules/javafx.graphics/src/main/java/javafx/scene/text/Text.java
```diff
@@ -1088,14 +1088,32 @@ public final PathElement[] caretShape(int charIndex, boolean caretBias) {
 
     /**
      * Returns the shape for the range of the text in local coordinates.
+     * <p>
+     * NOTE: this shapes returned do not include line spacing.
```
kevinrushforth: I would remove the "NOTE:", making it a simple statement (and probably no need for the `<p`> tag), since this is just what the method does.

Question: should this method be deprecated to be consistent with TextFlow?

kevinrushforth (commented):
I left a couple doc comments inline. I'm left wondering if we should do the same thing in Text that you are doing in TextFlow. Namely, deprecate the following four methods, adding getXxxxx methods as replacements:

- hitTest
- caretShape
- rangeShape
- underlineShape

Unlike TextFlow, there is no functional reason we need to do it. So the question is one of consistency. I note that you are already adding one of the four, getRangeShape (to add the lineSpacing parameter). What do you think?

andy-goryachev-oracle:
> Question: should this method be deprecated to be consistent with TextFlow?

> Namely, deprecate the following four methods, adding getXxxxx methods as replacements

I am open to suggestions, but currently leaning toward leaving `Text` as is.  As you correctly mentioned, it is not functionally required, and would only add more noise.

I feel making `Text` such a heavy object (with properties and all) might have been a design mistake: it would have been much better to use `TextFlow` everywhere, including `Labels`, `TextFields`, and `TextAreas`, and use `Text` as immutable objects similar to `GlyphList` to store runs of characters with the same font and attributes.  Doing so might have allowed easy introduction of rich text into these controls too, but oh well.

modules/javafx.graphics/src/main/java/javafx/scene/text/TextFlow.java
```diff
@@ -193,11 +193,15 @@ private void checkOrientation() {
 
     /**
      * Maps local point to {@link HitInfo} in the content.
+     * <p>
+     * NOTE: this method does not take border or padding into account.
      *
      * @param point the specified point to be tested
      * @return a {@code HitInfo} representing the character index found
      * @since 9
+     * @deprecated replaced by {@code getHitInfo()}
```
kevinrushforth: I recommend `@link` here.

modules/javafx.graphics/src/main/java/javafx/scene/text/TextFlow.java
```diff
@@ -210,42 +214,132 @@ public final HitInfo hitTest(javafx.geometry.Point2D point) {
...
     /**
      * Returns shape of caret in local coordinates.
+     * <p>
+     * NOTE: this method does not take border or padding into account.
      *
      * @param charIndex the character index for the caret
      * @param leading whether the caret is biased on the leading edge of the character
      * @return an array of {@code PathElement} which can be used to create a {@code Shape}
      * @since 9
+     * @deprecated replaced by {@code getCaretShape()}
```
kevinrushforth: `@link` (here and in. other deprecated methods).

kevinrushforth (commented):
The API and docs look good to me now. I left a minor formatting suggestion for the deprecated methods (use `@link` to point to the replacement methods).

Update the CSR to match, and I'll Review it.

andy-goryachev-oracle:
Made https://bugs.openjdk.org/browse/JDK-8357594 to be the main issue, closed https://bugs.openjdk.org/browse/JDK-8341438 as a duplicate of JDK-8357594 .

modules/javafx.graphics/src/main/java/javafx/scene/text/TextFlow.java
```diff
@@ -368,9 +462,18 @@ public boolean usesMirroring() {
         inLayout = false;
     }
 
-    private PathElement[] getRange(int start, int end, int type) {
+    private PathElement[] getRange(int start, int end, int type, boolean accountForInsets, double lineSpacing) {
+        double dx;
+        double dy;
+        if(accountForInsets) {
```
kevinrushforth: Minor: space after `if`

kevinrushforth (approved):

mstr2 (approved):

PR #1819, opened 2025-06-01
8358255: Factor out boilerplate code of EventHandler properties in Scene and Window

mstr2:
`EventHandler` property implementations in `Scene` and `Window` use anonymous classes derived from `ObjectPropertyBase`. We can remove about 650 lines of boilerplate code by using a common property class instead.

/reviewers 2

---------
### Progress

### Issue
 * [JDK-8358255](https://bugs.openjdk.org/browse/JDK-8358255): Factor out boilerplate code of EventHandler properties in Scene and Window (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1819/head:pull/1819` \
`$ git checkout pull/1819`

`$ git checkout pull/1819` \
`$ git pull https://git.openjdk.org/jfx.git pull/1819/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1819`

`$ git pr show -t 1819`

</details>

<a href="https://git.openjdk.org/jfx/pull/1819.diff">https://git.openjdk.org/jfx/pull/1819.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1819#issuecomment-2927949769)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
@@ -4410,23 +4410,8 @@ public class Scene implements EventTarget {
 
     public final ObjectProperty<EventHandler<? super ContextMenuEvent>> onContextMenuRequestedProperty() {
         if (onContextMenuRequested == null) {
-            onContextMenuRequested = new ObjectPropertyBase<>() {
-
-                @Override
-                protected void invalidated() {
-                    setEventHandler(ContextMenuEvent.CONTEXT_MENU_REQUESTED, get());
-                }
-
-                @Override
-                public Object getBean() {
-                    return Scene.this;
-                }
-
-                @Override
-                public String getName() {
-                    return "onContextMenuRequested";
-                }
-            };
+            onContextMenuRequested = new EventHandlerProperty<>(
+                "onContextMenuRequested", ContextMenuEvent.CONTEXT_MENU_REQUESTED);
         }
         return onContextMenuRequested;
     }
@@ -4453,23 +4438,7 @@ public class Scene implements EventTarget {
 
     public final ObjectProperty<EventHandler<? super MouseEvent>> onMouseClickedProperty() {
         if (onMouseClicked == null) {
-            onMouseClicked = new ObjectPropertyBase<>() {
-
-                @Override
-                protected void invalidated() {
-                    setEventHandler(MouseEvent.MOUSE_CLICKED, get());
-                }
-
-                @Override
-                public Object getBean() {
-                    return Scene.this;
-                }
-
-                @Override
-                public String getName() {
-                    return "onMouseClicked";
-                }
-            };
+            onMouseClicked = new EventHandlerProperty<>("onMouseClicked", MouseEvent.MOUSE_CLICKED);
         }
         return onMouseClicked;
     }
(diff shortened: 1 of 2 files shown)


kevinrushforth:
@arapte can be the second reviewer

nlisker:
I've tried to do something similar for controls by replacing a lot of the anonymous classes with concrete ones. I can't find the issue/discussion right now, but Kevin measured a non-negligible increase in memory usage. I assume it's because constant folding is doable for the constants in the methods of the anonymous class, but not for the final fields in the concrete property (because they are not really final when considering reflection).

Because an application doesn't have a lot of windows and scenes, it's possible that these changes won't have detrimental effects, but I suggest measuring.

andy-goryachev-oracle:
To clarify, the difference is as follows:
- on per-class basis, we are saving a few kilobytes by replacing a bunch of class data structures with one (with two, one for `Scene` and one for `Window`)
- on per-instance basis, we are adding 2 pointers multiplied by the number of properties (39 for `Scene`), multiplied by the number of instances of `Scene`/`Windows`.

So depending on the application, it might be a net win for a small app, or more memory used when the app devs create hundreds of windows and keep them hidden (I've seen those).

Still, I think it's a good change.

mstr2:
> I've tried to do something similar for controls by replacing a lot of the anonymous classes with concrete ones. I can't find the issue/discussion right now, but Kevin measured a non-negligible increase in memory usage. I assume it's because constant folding is doable for the constants in the methods of the anonymous class, but not for the final fields in the concrete property (because they are not really final when considering reflection).
> 
> Because an application doesn't have a lot of windows and scenes, it's possible that these changes won't have detrimental effects, but I suggest measuring.

I don't think there's any constant folding here, because methods like `getName()` are not constant expressions. You can always alias a property instance dynamically by referencing, which is why the compiler can't inline the value into the calling expression.

That being said, of course the nominal class stores two additional fields that the anonymous class didn't. But it doesn't store any more than that: the name strings come from literal expressions and are therefore interned, and the event types are references to static instances.

mstr2:
>  when the app devs create hundreds of windows and keep them hidden (I've seen those)

:dizzy_face:

hjohn:
> I've tried to do something similar for controls by replacing a lot of the anonymous classes with concrete ones. I can't find the issue/discussion right now, but Kevin measured a non-negligible increase in memory usage. I assume it's because constant folding is doable for the constants in the methods of the anonymous class, but not for the final fields in the concrete property (because they are not really final when considering reflection).
> 
> Because an application doesn't have a lot of windows and scenes, it's possible that these changes won't have detrimental effects, but I suggest measuring.

I think that's not really needed; we'd be crippling any FX improvements if the criteria is that we can't use more memory than before?  The better question is, is having less classes (improving the **common** case), and less code an acceptable trade-off versus the case where you have hundreds or thousands of scenes and windows?  I think it is.  These classes weigh-in at thousands of bytes already (and far more once their peer is created); we're not doing anyone any favors by being overly cautious here.

kevinrushforth:
> > when the app devs create hundreds of windows and keep them hidden (I've seen those)
> 
> 😵

Window and Scene are heavy-weight classes that are reasonably limited in quantity, so I'm not at all worried about this. The concern I raised earlier (as I recall) was about properties in nodes.

hjohn (approved):
I think this is a huge improvement, and a net win overall.

nlisker:
> > > when the app devs create hundreds of windows and keep them hidden (I've seen those)
> > 
> > 
> > 😵
> 
> Window and Scene are heavy-weight classes that are reasonably limited in quantity, so I'm not at all worried about this. The concern I raised earlier (as I recall) was about properties in nodes.

Yes, I'm thinking the same thing.

PR #1822, opened 2025-06-06
8358820: Allow interpolation outside of range [0,1]

mstr2:
JavaFX unnecessarily restricts interpolation in the following ways:
1. `Interpolatable` implementations often clamp intermediate values to the interpolation factor range [0,1].
2. `SplineInterpolator` doesn't accept Y coordinates outside of [0,1] for its control points. While this was probably done so that the computed interpolation factor doesn't exceed [0,1], the restriction goes far beyond that. For example, the following function is not representable, even though its values are all within the [0,1] range:<br>
   <br>
   The following function is also not representable, but would be very useful for [bouncy animations](https://easings.net/#easeOutBack):<br>
   

Fortunately, there is no technical reason why JavaFX can't support the full range of animations that can be represented with a cubic Beziér interpolation function.

This PR includes the following changes:
1. The specification of `Interpolatable` is changed to require implementations to accept interpolation factors outside of [0,1].
2. All implementations of `Interpolatable` now correctly return intermediate values outside of [0,1].
3. `SplineInterpolator` now accepts control points with any Y coordinate.

Here's how the result looks like for the previously unrepresentable interpolation function `cubic-bezier(0.34, 2.2, 0.64, 1)`:<br>

/reviewers 2
/csr

---------
### Progress
- [x] Change requires CSR request [JDK-8360047](https://bugs.openjdk.org/browse/JDK-8360047) to be approved

### Issues
 * [JDK-8358820](https://bugs.openjdk.org/browse/JDK-8358820): Allow interpolation outside of range [0,1] (**Enhancement** - P4)
 * [JDK-8360047](https://bugs.openjdk.org/browse/JDK-8360047): Allow interpolation outside of range [0,1] (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1822/head:pull/1822` \
`$ git checkout pull/1822`

`$ git checkout pull/1822` \
`$ git pull https://git.openjdk.org/jfx.git pull/1822/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1822`

`$ git pr show -t 1822`

</details>

<a href="https://git.openjdk.org/jfx/pull/1822.diff">https://git.openjdk.org/jfx/pull/1822.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1822#issuecomment-2951218687)
</details>

### DIFF
--- a/tests/manual/graphics/CssTransitionsTest.java
+++ b/tests/manual/graphics/CssTransitionsTest.java
@@ -117,8 +117,8 @@ public class CssTransitionsTest extends Application {
             new RectInfo("#rect2", "rect2", Color.WHITE),
             new RectInfo("#rect3", "rect3", Color.WHITE),
             new RectInfo("#rect4", "rect4", Color.WHITE),
-            new RectInfo("#rect4", "rect5", Color.WHITE),
-            new RectInfo("#rect5", "rect6", Color.WHITE));
+            new RectInfo("#rect5", "rect5", Color.WHITE),
+            new RectInfo("#rect6", "rect6", Color.WHITE));
     }
 
     private Region createBackgroundTransitionsTab() {


nlisker:
Noting that this proposed change supersedes the previous changes that were done for for https://bugs.openjdk.org/browse/JDK-8226911. The proposed behavioral change makes sense.

mstr2:
> Noting that this proposed change supersedes the previous changes that were done for for https://bugs.openjdk.org/browse/JDK-8226911. The proposed behavioral change makes sense.

Can I interest you in a review of the API and/or the implementation?

modules/javafx.graphics/src/main/java/javafx/animation/Interpolatable.java
```diff
@@ -39,10 +39,9 @@
  *                 {@link #interpolate(Object, double)} method.</td>
  *         </tr>
  *         <tr><td style="vertical-align: top"><a id="linear" style="white-space: nowrap">linear</a></td>
- *             <td>Two components are combined by linear interpolation such that {@code t = 0} produces
- *                 the start value, {@code t = 1} produces the end value, and {@code 0 < t < 1} produces
- *                 {@code (1 - t) * start + t * end}. This interpolation type is usually applicable for
- *                 numeric components.</td>
+ *             <td>Two components are combined by linear interpolation such that the intermediate value is
+ *                 produced by computing {@code (1 - t) * start + t * end}. This interpolation type is usually
+ *                 applicable for numeric components.</td>
```
nlisker: I think that the use of "start" and "end" is confusing now since they don't correspond to actual start and end. You can get values that are "after the end". It might be seen more as an extrapolation at this point. In general, this interface now also functions as an extrapolator.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/layout/region/BorderImageSlices.java
```diff
@@ -59,11 +59,11 @@ public BorderImageSlices(BorderWidths widths, boolean filled) {
     public BorderImageSlices interpolate(BorderImageSlices endValue, double t) {
         Objects.requireNonNull(endValue, "endValue cannot be null");
 
-        if (t <= 0 || equals(endValue)) {
+        if (t == 0 || equals(endValue)) {
```
nlisker, line 62: `==` on floating point could produce unexpected results since there's +0 and -0. Here it doesn't seem to matter because it's an optimization, but I see more of these in the new code, so might be worth double checking.

mstr2: +0 and -0 seem to be equal when using the `==` operator:
```java
double a = 0.0;
double b = -0.0;
System.out.println(a == b); // prints "true"
System.out.println(a < b); // prints "false"
System.out.println(a > b); // prints "false"
```

If the sign is relevant, one must use `Double.compare`:
```
double a = 0.0;
double b = -0.0;
System.out.println(Double.compare(a, b)); // prints "1"
System.out.println(Double.compare(b, a)); // prints "-1"
```

modules/javafx.graphics/src/main/java/javafx/animation/Interpolatable.java
mstr2: I've added the following sentence:
`Note that this formula produces values less than {@code start} for {@code t < 0} and values greater than {@code end} for {@code t > 1}.`

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/layout/region/BorderImageSlices.java
jddarcy: Yes, IEEE 754 (and Java) require that -0.0 == +0.0 to evaluate to true.

palexdev:
Currently the workaround I use for this is to have a custom Interpolator implementation like:
```java
public class Cubic extends Interpolator {
    private final double x1;
    private final double y1;
    private final double x2;
    private final double y2;

    private static final double CUBIC_ERROR_BOUND = 0.001;

    public Cubic(double x1, double y1, double x2, double y2) {
        this.x1 = x1;
        this.y1 = y1;
        this.x2 = x2;
        this.y2 = y2;
    }

    double elevateCubic(double a, double b, double m) {
        return 3 * a * (1 - m) * (1 - m) * m +
               3 * b * (1 - m) * m * m +
               m * m * m;
    }

    @Override
    public double curve(double t) {
        double start = 0.0;
        double end = 1.0;
        while (true) {
            final double midpoint = (start + end) / 2;
            final double estimate = elevateCubic(x1, x2, midpoint);
            if (Math.abs(t - estimate) < CUBIC_ERROR_BOUND) {
                return elevateCubic(y1, y2, midpoint);
            }
            if (estimate < t) {
                start = midpoint;
            } else {
                end = midpoint;
            }
        }
    }
}
```
Which allows me to define such curves:
```java
    public static final MotionPreset EXPRESSIVE_FAST_SPATIAL = MotionPreset.of(
        new Cubic(0.42, 1.67, 0.21, 0.90),
        MEDIUM3 // Duration constant
    );
```
I believe I copied this from Flutter.
(The new Material 3 Expressive makes use a lot of springs, see [Material 3 Docs](https://m3.material.io/styles/motion/overview/specs#1b299695-5822-4738-ae56-ef9389b412d2))


Unfortunately, I need to use such curves in CSS too, so I kinda need this commit to be merged

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/layout/region/Margins.java
```diff
@@ -76,11 +76,11 @@ public Margins(double top, double right, double bottom, double left, boolean pro
     public Margins interpolate(Margins endValue, double t) {
         Objects.requireNonNull(endValue, "endValue cannot be null");
 
-        if (t <= 0 || equals(endValue)) {
+        if (t == 0 || equals(endValue)) {
```
andy-goryachev-oracle, line 79: the class javadoc says

```
 * If proportional is true, then the values represent fractions or percentages
 * and are in the range 0..1, although this is not enforced.
```

should this be changed/clarified?

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/layout/region/Margins.java
```diff
@@ -89,10 +89,10 @@ public Margins interpolate(Margins endValue, double t) {
         }
 
         return new Margins(
-            InterpolationUtils.interpolate(top, endValue.top, t),
-            InterpolationUtils.interpolate(right, endValue.right, t),
-            InterpolationUtils.interpolate(bottom, endValue.bottom, t),
-            InterpolationUtils.interpolate(left, endValue.left, t),
+            Math.max(0, InterpolationUtils.interpolate(top, endValue.top, t)),
```
andy-goryachev-oracle, line 92: what is the reason for clamping this at 0, but not 1.0 ?

also, why clamp here but not in `Insets` ?

modules/javafx.graphics/src/main/java/javafx/scene/paint/RadialGradient.java
```diff
@@ -234,17 +234,21 @@ public final List<Stop> getStops() {
...
      * @param centerY the Y coordinate of the center point of the gradient's circle
-     * @param radius the radius of the circle defining the extents of the color gradient
+     * @param radius the radius of the circle defining the extents of the color gradient,
+     *               must be greater than or equal to zero
      * @param proportional whether the coordinates and sizes are proportional
-     * to the shape which this gradient fills
+     *                     to the shape which this gradient fills
      * @param cycleMethod cycle method applied to the gradient
      * @param stops the gradient's color specification
+     * @throws IllegalArgumentException if {@code focusDistance} or {@code radius} is negative
```
andy-goryachev-oracle, line 251: can we use the `@since` tag applicable to `@throws` only?
(also below)

modules/javafx.graphics/src/main/java/javafx/scene/paint/RadialGradient.java
```diff
@@ -585,8 +606,12 @@ public static RadialGradient valueOf(String value) {
         if ("focus-distance".equals(tokens[0])) {
             GradientUtils.Parser.checkNumberOfArguments(tokens, 1);
             distance = GradientUtils.Parser.parsePercentage(tokens[1]);
-
             parser.shift();
+
+            if (distance < 0) {
+                throw new IllegalArgumentException(
+                    "Invalid gradient specification: focus-distance cannot be negative");
```
andy-goryachev-oracle: why wrap here?
also below

modules/javafx.graphics/src/test/java/test/javafx/scene/paint/RadialGradientTest.java
```diff
@@ -284,6 +284,22 @@ public void testValueOfIllegalNotations() {
         } catch (IllegalArgumentException iae) {
             // expected
         }
+
+        // radius can't be negative
+        try {
+            RadialGradient.valueOf("radial-gradient(radius -100, red 0%, blue 30%, black 100%)");
+            fail("IllegalArgument should have been thrown.");
```
andy-goryachev-oracle, line 291: nice: these tests do more than a simple `assertThrows` would do.

modules/javafx.graphics/src/test/java/test/javafx/scene/paint/RadialGradientTest.java
```diff
@@ -284,6 +284,22 @@ public void testValueOfIllegalNotations() {
...
+            assertTrue(iae.getMessage().contains("radius cannot be negative"));
+        }
+
+        // focusDistance can't be negative
+        try {
+            RadialGradient.valueOf("radial-gradient(focus-distance -100%, red 0%, blue 30%, black 100%)");
+            fail("IllegalArgument should have been thrown.");
+        } catch (IllegalArgumentException iae) {
+            assertTrue(iae.getMessage().contains("focus-distance cannot be negative"));
+        }
```
andy-goryachev-oracle, line 302: `RadialGradient` adds `checkInvariants` in three places, but only two tests are added.
Should we add one more?

andy-goryachev-oracle (approved):
looks good; a few minor comments.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/layout/region/Margins.java
mstr2: `Margins` is used for border widths, which can't be negative according to specification. However, they can be any positive size, therefore we don't clamp here.

modules/javafx.graphics/src/main/java/javafx/scene/paint/RadialGradient.java
mstr2: I removed the line break in this place and others.

modules/javafx.graphics/src/test/java/test/javafx/scene/paint/RadialGradientTest.java
mstr2: No, but I've removed the `checkInvariants()` call in the private constructor, as it is only called from the `interpolate()` method and thus always operates with valid gradients (no need to verify what has already been verified).

modules/javafx.graphics/src/main/java/javafx/scene/paint/RadialGradient.java
mstr2: I don't think this is possible.

andy-goryachev-oracle (approved):

modules/javafx.graphics/src/test/java/test/javafx/scene/paint/RadialGradientTest.java
hjohn: You can do this actually as `assertThrows` returns the exception thrown:

            assertEquals("message", assertThrows(
                IllegalArgumentException.class, 
                () -> RadialGradient.valueOf("radial-gradient(radius -100, red 0%, blue 30%, black 100%)")
            ).getMessage());

mstr2: That's true, but I don't want to refactor all of the tests here at this time.

kevinrushforth:
The API spec changes look good. I left a comment in the CSR that you need to add the changes made to `RadialGradient` (it looks good otherwise).

I spot checked the implementation and it all looks like what I expect. I'll finish reviewing and testing today.

tests/manual/graphics/CssTransitionsTest.java
```diff
@@ -94,28 +94,31 @@ private Region createTransitionDelayTab() {
...
-            new RectInfo("#rect2", "rect2"),
-            new RectInfo("#rect3", "rect3"),
-            new RectInfo("#rect4", "rect4"),
-            new RectInfo("#rect5", "rect5"));
+            new RectInfo("#rect1", "rect1", Color.WHITE),
+            new RectInfo("#rect2", "rect2", Color.WHITE),
+            new RectInfo("#rect3", "rect3", Color.WHITE),
+            new RectInfo("#rect4", "rect4", Color.WHITE),
+            new RectInfo("#rect4", "rect5", Color.WHITE),
+            new RectInfo("#rect5", "rect6", Color.WHITE));
```
kevinrushforth: Typo in the first string arg in the above two lines: they should be 5 and 6, respectively.

andy-goryachev-oracle (approved):

kevinrushforth (approved):
Tests pass. All good.

PR #1826, opened 2025-06-13
8359445: GHA: Update gradle wrapper-validation action to v4

kevinrushforth:
As noted in JBS, our GitHub Actions workflow uses `gradle/actions/wrapper-validation@v3`. We should update to the latest version, which is v4.

In addition to it being a good idea to use the latest, we've seen a lot of intermittent failures in gradle validation, and I hope this might help.

### DIFF
--- a/.github/workflows/submit.yml
+++ b/.github/workflows/submit.yml
@@ -59,7 +59,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


andy-goryachev-oracle (approved):
lgtm, the GHA builds and runs, apparently.

Maran23 (approved):
Looks good. I also hope it helps with intermittent validation failures (Got that at least >10 times in the last month).

PR #1829, opened 2025-06-15
8359598: [TestBug] VirtualFlowTestUtils should not create a temporary Stage

Maran23:
Currently, the VirtualFlowTestUtils used ONLY for tests has many utility methods to get and do something with the VirtualFlow of Virtualized Controls.

It has one flaw: It may creates a temporary Stage with the StageLoader, when the Control was never inside a Scene (and Stage) yet. This is done to get the VirtualFlow, which is not possible otherwise.

Take the following test code:
```
VirtualFlowTestUtils.clickOnRow(tableView, 2);
VirtualFlowTestUtils.clickOnRow(tableView, 4);
```

What it does it to create a Stage for the first method and dispose it after. And create another Stage for the second method and dispose it after.

This does not test a realistic scenario and the chance is quite high that developers used that methods without even knowing that it contains such logic.

So the idea is to remove the StageLoader code from VirtualFlowTestUtils and rather use it in the Test code before calling the Util methods.

For the example above, this would result in:
```
stageLoader = new StageLoader(tableView);
VirtualFlowTestUtils.clickOnRow(tableView, 2);
VirtualFlowTestUtils.clickOnRow(tableView, 4);
```

The stageLoader should be disposed in an @AfterEach.

Note: Nearly all touched tests are indeed much older test code. New tests are not affected and already use it correcty.
Sometimes a call to `Toolkit.getToolkit().firePulse();` was needed. Previously, creating a new Stage for every call to the util methods did that implicitly.

---------
### Progress

### Issue
 * [JDK-8359598](https://bugs.openjdk.org/browse/JDK-8359598): [TestBug] VirtualFlowTestUtils should not create a temporary Stage (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1829/head:pull/1829` \
`$ git checkout pull/1829`

`$ git checkout pull/1829` \
`$ git pull https://git.openjdk.org/jfx.git pull/1829/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1829`

`$ git pr show -t 1829`

</details>

<a href="https://git.openjdk.org/jfx/pull/1829.diff">https://git.openjdk.org/jfx/pull/1829.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1829#issuecomment-2974611407)
</details>

### DIFF
--- a/modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/infrastructure/VirtualFlowTestUtils.java
+++ b/modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/infrastructure/VirtualFlowTestUtils.java
@@ -24,11 +24,8 @@
  */
 package test.com.sun.javafx.scene.control.infrastructure;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertFalse;
-import static org.junit.jupiter.api.Assertions.assertNotNull;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-import java.util.List;
+import com.sun.javafx.scene.control.LabeledText;
+import com.sun.javafx.scene.control.VirtualScrollBar;
 import javafx.geometry.Orientation;
 import javafx.scene.Node;
 import javafx.scene.control.Cell;
@@ -46,8 +43,13 @@ import javafx.scene.control.skin.TableHeaderRow;
 import javafx.scene.control.skin.TableViewSkinBase;
 import javafx.scene.control.skin.VirtualFlow;
 import javafx.util.Callback;
-import com.sun.javafx.scene.control.LabeledText;
-import com.sun.javafx.scene.control.VirtualScrollBar;
+
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class VirtualFlowTestUtils {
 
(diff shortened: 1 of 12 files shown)


modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/infrastructure/VirtualFlowTestUtils.java
```diff
@@ -346,10 +343,6 @@ public static VirtualFlow<?> getVirtualFlow(Control control) {
 
         flow = (VirtualFlow<?>)control.lookup("#virtual-flow");
 
-        if (sl != null) {
```
andy-goryachev-oracle, line 349: should we have a check here with a helpful message (which recommends to use `StageLoader`), instead of returning `null`?

Maran23: I thought about this as well. No strong opinion from my side.

azuev-java: Depends on the amount of noise it will produce during the test run but in general i would also prefer some kind of message here.

andy-goryachev-oracle: there should be no noise: it's more of a question of instructing the developer why their test failed.  if there is no message, the NPE will happen in the context of the failing test which makes it rather obvious.  I agree with @Maran23 , either way is fine.

andy-goryachev-oracle (approved):
Sorry for a long delay reviewing this PR.

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/infrastructure/VirtualFlowTestUtils.java
Maran23: Since they were two similar opinions about this, I added an exception with a clear message why the flow might be null. :)

PR #1830, opened 2025-06-15
8359599: Calling refresh() for all virtualized controls recreates all cells instead of refreshing the cells

Maran23:
When calling `refresh()` on virtualized Controls (`ListView`, `TreeView`, `TableView`, `TreeTableView`), all cells will be recreated completely, instead of just refreshing them.

This is because `recreateCells()` of the `VirtualFlow` is called when `refresh()` was called. This is not needed, since refreshing the cells can be done much cheaper with `rebuildCells()`.

This will reset all cells (`index = -1`), add them to the pile and fill them back in the viewport with an index again. This ensures `updateItem()` is called.

The contract of `refresh()` is also a big vague, stating:
```
Calling {@code refresh()} forces the XXX control to recreate and repopulate the cells 
necessary to populate the visual bounds of the control.
In other words, this forces the XXX to update what it is showing to the user. 
This is useful in cases where the underlying data source has changed in a way that is not observed by the XXX itself.
```

As written above, recreating is not needed in order to fulfull the contract of updating what is shown to the user in case the underlying data source changed without JavaFX noticing (e.g. calling a normal Setter without any Property and therefore listener involved).

---------
### Progress
- [ ] Change requires a CSR request matching fixVersion jfx26 to be approved (needs to be created)

### Issue
 * [JDK-8359599](https://bugs.openjdk.org/browse/JDK-8359599): Calling refresh() for all virtualized controls recreates all cells instead of refreshing the cells (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1830/head:pull/1830` \
`$ git checkout pull/1830`

`$ git checkout pull/1830` \
`$ git pull https://git.openjdk.org/jfx.git pull/1830/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1830`

`$ git pr show -t 1830`

</details>

<a href="https://git.openjdk.org/jfx/pull/1830.diff">https://git.openjdk.org/jfx/pull/1830.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1830#issuecomment-2973953731)
</details>

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/TableView.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/TableView.java
@@ -1763,8 +1763,8 @@ public class TableView<S> extends Control {
     }
 
     /**
-     * Calling {@code refresh()} forces the TableView control to recreate and
-     * repopulate the cells necessary to populate the visual bounds of the control.
+     * Calling {@code refresh()} forces the TableView control to repopulate the
+     * cells necessary to populate the visual bounds of the control.
      * In other words, this forces the TableView to update what it is showing to
      * the user. This is useful in cases where the underlying data source has
      * changed in a way that is not observed by the TableView itself.
@@ -1772,7 +1772,7 @@ public class TableView<S> extends Control {
      * @since JavaFX 8u60
      */
     public void refresh() {
-        getProperties().put(Properties.RECREATE, Boolean.TRUE);
+        getProperties().put(Properties.REBUILD, Boolean.TRUE);
     }
 
 
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
@@ -152,7 +152,6 @@ public abstract class TableViewSkinBase<M, S, C extends Control, I extends Index
 
     private int visibleColCount;
 
-    boolean needCellsRecreated = true;
     boolean needCellsReconfigured = false;
 
     private int itemCount = -1;
@@ -297,19 +296,18 @@ public abstract class TableViewSkinBase<M, S, C extends Control, I extends Index
         });
 
         final ObservableMap<Object, Object> properties = control.getProperties();
-        properties.remove(Properties.REFRESH);
         properties.remove(Properties.RECREATE);
+        properties.remove(Properties.REBUILD);
         lh.addMapChangeListener(properties, (c) -> {
             if (!c.wasAdded()) {
                 return;
             }
-            if (Properties.REFRESH.equals(c.getKey())) {
-                refreshView();
-                getSkinnable().getProperties().remove(Properties.REFRESH);
-            } else if (Properties.RECREATE.equals(c.getKey())) {
-                needCellsRecreated = true;
-                refreshView();
+            if (Properties.RECREATE.equals(c.getKey())) {
+                requestRecreateCells();
                 getSkinnable().getProperties().remove(Properties.RECREATE);
+            } else if (Properties.REBUILD.equals(c.getKey())) {
+                requestRebuildCells();
+                getSkinnable().getProperties().remove(Properties.REBUILD);
             }
         });
 
@@ -425,13 +423,10 @@ public abstract class TableViewSkinBase<M, S, C extends Control, I extends Index
 
         super.layoutChildren(x, y, w, h);
 
-        if (needCellsRecreated) {
-            flow.recreateCells();
-        } else if (needCellsReconfigured) {
+        if (needCellsReconfigured) {
             flow.reconfigureCells();
         }
 
-        needCellsRecreated = false;
         needCellsReconfigured = false;
 
         final double baselineOffset = table.getLayoutBounds().getHeight() / 2;
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableRowSkinTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableRowSkinTest.java
@@ -352,20 +352,6 @@ public class TreeTableRowSkinTest {
         assertEquals(5, row.getChildrenUnmodifiable().stream().filter(TreeTableCell.class::isInstance).count());
     }
 
-    /** TreeTableView.refresh() must release all discarded cells JDK-8307538 */
-    @Test
-    public void cellsMustBeCollectableAfterRefresh() {
-        IndexedCell<?> row = VirtualFlowTestUtils.getCell(treeTableView, 0);
-        assertNotNull(row);
-        WeakReference<Object> ref = new WeakReference<>(row);
-        row = null;
-
-        treeTableView.refresh();
-        Toolkit.getToolkit().firePulse();
-
-        JMemoryBuddy.assertCollectable(ref);
-    }
-
     /** TreeTableView.setRowFactory() must release all discarded cells JDK-8307538 */
     @Test
     public void cellsMustBeCollectableAfterRowFactoryChange() {
--- a/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/Properties.java
+++ b/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/Properties.java
@@ -82,12 +82,12 @@ public class Properties {
 
     /***************************************************************************
      *
-     * ListView, TableView
+     * Virtualized controls
      *
      **************************************************************************/
 
-    public static final String REFRESH = "refreshKey";
     public static final String RECREATE = "recreateKey";
+    public static final String REBUILD = "rebuildKey";
 
 
 
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/ListView.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/ListView.java
@@ -1032,8 +1032,8 @@ public class ListView<T> extends Control {
     }
 
     /**
-     * Calling {@code refresh()} forces the ListView control to recreate and
-     * repopulate the cells necessary to populate the visual bounds of the control.
+     * Calling {@code refresh()} forces the ListView control to repopulate the
+     * cells necessary to populate the visual bounds of the control.
      * In other words, this forces the ListView to update what it is showing to
      * the user. This is useful in cases where the underlying data source has
      * changed in a way that is not observed by the ListView itself.
@@ -1041,7 +1041,7 @@ public class ListView<T> extends Control {
      * @since JavaFX 8u60
      */
     public void refresh() {
-        getProperties().put(Properties.RECREATE, Boolean.TRUE);
+        getProperties().put(Properties.REBUILD, Boolean.TRUE);
     }
 
 
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/TreeTableView.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/TreeTableView.java
@@ -2079,8 +2079,8 @@ public class TreeTableView<S> extends Control {
     }
 
     /**
-     * Calling {@code refresh()} forces the TreeTableView control to recreate and
-     * repopulate the cells necessary to populate the visual bounds of the control.
+     * Calling {@code refresh()} forces the TreeTableView control to repopulate the
+     * cells necessary to populate the visual bounds of the control.
      * In other words, this forces the TreeTableView to update what it is showing to
      * the user. This is useful in cases where the underlying data source has
      * changed in a way that is not observed by the TreeTableView itself.
@@ -2088,7 +2088,7 @@ public class TreeTableView<S> extends Control {
      * @since JavaFX 8u60
      */
     public void refresh() {
-        getProperties().put(Properties.RECREATE, Boolean.TRUE);
+        getProperties().put(Properties.REBUILD, Boolean.TRUE);
     }
 
 
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/TreeView.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/TreeView.java
@@ -1109,8 +1109,8 @@ public class TreeView<T> extends Control {
     }
 
     /**
-     * Calling {@code refresh()} forces the TreeView control to recreate and
-     * repopulate the cells necessary to populate the visual bounds of the control.
+     * Calling {@code refresh()} forces the TreeView control to repopulate the
+     * cells necessary to populate the visual bounds of the control.
      * In other words, this forces the TreeView to update what it is showing to
      * the user. This is useful in cases where the underlying data source has
      * changed in a way that is not observed by the TreeView itself.
@@ -1118,7 +1118,7 @@ public class TreeView<T> extends Control {
      * @since JavaFX 8u60
      */
     public void refresh() {
-        getProperties().put(Properties.RECREATE, Boolean.TRUE);
+        getProperties().put(Properties.REBUILD, Boolean.TRUE);
     }
 
 
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ListViewSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ListViewSkin.java
@@ -101,7 +101,6 @@ public class ListViewSkin<T> extends VirtualContainerBase<ListView<T>, ListCell<
 
     private ObservableList<T> listViewItems;
 
-    private boolean needCellsRebuilt = true;
     private boolean needCellsReconfigured = false;
 
     private int itemCount = -1;
@@ -117,10 +116,9 @@ public class ListViewSkin<T> extends VirtualContainerBase<ListView<T>, ListCell<
 
     private MapChangeListener<Object, Object> propertiesMapListener = c -> {
         if (! c.wasAdded()) return;
-        if (Properties.RECREATE.equals(c.getKey())) {
-            needCellsRebuilt = true;
-            getSkinnable().requestLayout();
-            getSkinnable().getProperties().remove(Properties.RECREATE);
+        if (Properties.REBUILD.equals(c.getKey())) {
+            requestRebuildCells();
+            getSkinnable().getProperties().remove(Properties.REBUILD);
         }
     };
 
@@ -230,7 +228,7 @@ public class ListViewSkin<T> extends VirtualContainerBase<ListView<T>, ListCell<
         control.itemsProperty().addListener(weakItemsChangeListener);
 
         final ObservableMap<Object, Object> properties = control.getProperties();
-        properties.remove(Properties.RECREATE);
+        properties.remove(Properties.REBUILD);
         properties.addListener(weakPropertiesMapListener);
 
         // Register listeners
@@ -281,13 +279,10 @@ public class ListViewSkin<T> extends VirtualContainerBase<ListView<T>, ListCell<
                                             final double w, final double h) {
         super.layoutChildren(x, y, w, h);
 
-        if (needCellsRebuilt) {
-            flow.rebuildCells();
-        } else if (needCellsReconfigured) {
+        if (needCellsReconfigured) {
             flow.reconfigureCells();
         }
 
-        needCellsRebuilt = false;
         needCellsReconfigured = false;
 
         if (getItemCount() == 0) {
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeViewSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeViewSkin.java
@@ -102,9 +102,9 @@ public class TreeViewSkin<T> extends VirtualContainerBase<TreeView<T>, TreeCell<
 
     private MapChangeListener<Object, Object> propertiesMapListener = c -> {
         if (! c.wasAdded()) return;
-        if (Properties.RECREATE.equals(c.getKey())) {
+        if (Properties.REBUILD.equals(c.getKey())) {
             requestRebuildCells();
-            getSkinnable().getProperties().remove(Properties.RECREATE);
+            getSkinnable().getProperties().remove(Properties.REBUILD);
         }
     };
 
@@ -184,7 +184,7 @@ public class TreeViewSkin<T> extends VirtualContainerBase<TreeView<T>, TreeCell<
         flow.getHbar().addEventFilter(MouseEvent.MOUSE_PRESSED, ml);
 
         final ObservableMap<Object, Object> properties = control.getProperties();
-        properties.remove(Properties.RECREATE);
+        properties.remove(Properties.REBUILD);
         properties.addListener(propertiesMapListener);
 
         // init the behavior 'closures'
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualContainerBase.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualContainerBase.java
@@ -191,4 +191,8 @@ public abstract class VirtualContainerBase<C extends Control, I extends IndexedC
         flow.rebuildCells();
     }
 
+    void requestRecreateCells() {
+        flow.recreateCells();
+    }
+
 }
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/ListViewTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/ListViewTest.java
@@ -25,25 +25,14 @@
 
 package test.javafx.scene.control;
 
-import static javafx.collections.FXCollections.observableArrayList;
-import static org.junit.jupiter.api.Assertions.assertArrayEquals;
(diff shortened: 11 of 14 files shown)


kevinrushforth:
This is a risky change and will need to be carefully tested.

Reviewers: @andy-goryachev-oracle @johanvos 

/reviewers 2 reviewers

johanvos:
It is indeed true that refresh() is often a very expensive operation. Whenever `VirtualFlow.recreateCells()` is called, most of the internal state of the VirtualFlow is destroyed, and everything is recalculated from scratch. I believe that is not a problem, as long as recreateCells() is not called (by the controls) unless really needed. In that light, this PR seems interesting, as it will limit the number of rebuild-from-scratch cases.
I ran the basic controls tests after applying the PR, and (a bit to my surprise) they all passed, which is great.
However, it is very likely that some code out there implicitly rely on the rebuild-from-scratch logic, and that code will then work different after applying this PR. I believe it would be good to find such a case where the behavior (which, I agree, is often a bit vaguely defined) changes, so that we can discuss this with a concrete example.

Hence, while I like the idea here (avoiding unneeded heavy-cost operations in VirtualFlow), I would like to avoid a number of follow-up issues once this is merged -- driven by a change in "expected" behavior.

Maran23:
> Hence, while I like the idea here (avoiding unneeded heavy-cost operations in VirtualFlow), I would like to avoid a number of follow-up issues once this is merged -- driven by a change in "expected" behavior.

I completely agree. We need to be careful with such changes.

> However, it is very likely that some code out there implicitly rely on the rebuild-from-scratch logic, and that code will then work different after applying this PR.

Since all rows (and cells) are reset and then updated, all changes that were not taken into account by the control are taken into account in any case then.
On a side note here: In any JavaFX project, I have overwritten the `refresh` method (since it is not final) and always implemented the lightweight method as proposed here. I never found any problem.

But I think there is one concrete case which breaks now.
Take the following example (slightly modified version of the code in the ticket):
<details>
<summary>Example</summary>

```
import java.util.concurrent.atomic.AtomicBoolean;

import javafx.beans.property.SimpleStringProperty;
import javafx.geometry.Insets;
import javafx.scene.Scene;
import javafx.scene.control.TableRow;
import javafx.scene.layout.Priority;
import javafx.stage.Stage;

public class TableRefresh extends Application {

        launch(args);
    }

    @Override
        TableView<String> tv = new RefreshTableView<>();
        tv.setEditable(true);
        AtomicBoolean alternate = new AtomicBoolean(false);

        tv.setRowFactory(e -> {
            if (alternate.get()) {
                System.out.println("Creating alternative row");
                return new TableRow<>() {
                    {
                        setPadding(new Insets(5));
                    }
                };
            }

            System.out.println("Creating row");
            return new TableRow<>();
        });
        TableColumn<String, String> col = new TableColumn<>("Name");
        col.setCellValueFactory(i -> new SimpleStringProperty(i.getValue()));

        tv.getColumns().addAll(col);

        tv.setItems(FXCollections.observableArrayList("A", "B"));
        VBox.setVgrow(tv, Priority.ALWAYS);
        Button btn = new Button("Refresh");
        btn.setOnAction(e -> {
            alternate.set(true);
            tv.refresh();
        });
        Scene scene = new Scene(new VBox(tv, btn));

        stage.setScene(scene);
        stage.setTitle("Table Refresh");
        stage.show();
    }
}
```
</details>

Here, I'm aware that `refresh()` is recreating all rows, and update a boolean flag before calling `refresh()`, which leads to another path that is picked up and therefore other rows. 
In this case, it is much better to just set a new `TableRow` factory. This will do the same as what `refresh()` is doing right now (but not after this PR).
In never saw this code in the wild though.

But that is still a valid risk that we need to consider. This is the only problem I see right now.

andy-goryachev-oracle:
My main concern with this PR is that we might get a minute performance improvement, while risking regression.  Would it be possible to get some measurements using a modern system?

What do you think?

johanvos:
I think the performance improvements due to this PR can be pretty significant. The problem is indeed the risk on regression. I believe we need improvements in 2 areas before we can safely merge this:
1. More functional regression tests
2. Performance tests.

hjohn:
I think a CSR is needed if we're changing the documentation.  It specifically says it recreates the cells (although the purpose of that eludes me, aside from badly written cells).  

Since cells are supposed to be updated, and not "retain" anything from previous contents, I think only only buggy cells would benefit from recreating.  Such buggy cells however would probably have subtle problems already when they're never recreated, like listener leaks.

So although I agree that recreating is not needed as cell functionality is defined by their `update` methods, I do think this then requires a CSR.

kevinrushforth:
> So although I agree that recreating is not needed as cell functionality is defined by their `update` methods, I do think this then requires a CSR.

I agree.

/csr needed

Maran23:
/open

Will do the CSR + try to benchmark when I have more time!

Maran23:
I did some benchmarks and attached them to the description. Results look very good and is expected from what I measured myself when I implemented that. 

Since we do not need to throw away all `TableRow`s and `TableCell`s, and then recreate them, we get a massive boost. Especially when we have many items (-> rows) and many columns. 
If `TableCell`s have special logic, like showing a graphic or styling, this will be even bigger.

BlueGoliath:
@Maran23 Not a developer but I'd just like to say thanks for this. For the longest time I thought it was because of lazy coding on my part(new `ReadOnlyObjectWrapper` instances instead of caching) but it turns out that even if you optimize your code, `TableView`'s `refresh `method *still* performs terribly.

I'd like also to point out that the terrible performance isn't just from having 100s of rows. Comparing the CPU usage difference between viewing many updating `Label`(s) at once vs a single `TableView` shows a CPU usage increase(0.01%-0.03%ish to 0.06%-0.08%ish) on relatively extremely powerful modern hardware with just about a dozen rows. 

But what's even worse is the garbage allocation rate. `TableView` in my application allocates entire MEGABYTES of garbage in an application that has a relatively low garbage allocation rate. This not only results in more GCs but could result in expanding the heap, resulting in more app memory usage.

Can this please get more attention?

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableRowSkinTest.java
```diff
@@ -352,20 +352,6 @@ public void cellsShouldBeAddedInRowFixedCellSize() {
...
-        assertNotNull(row);
-        WeakReference<Object> ref = new WeakReference<>(row);
-        row = null;
-
-        treeTableView.refresh();
-        Toolkit.getToolkit().firePulse();
-
-        JMemoryBuddy.assertCollectable(ref);
-    }
-
```
hjohn, line 368: Is there another test that verifies that cells are garbage collectable?  For example, in the case where a table / list / tree table becomes smaller visually, I think that it should then perhaps discard some cells that then should be collectable?

hjohn (commented):
I think the changes look good.  I'm a bit confused in the performance table with what is meant with the `50 ms -> 0 ms` in the "after" cases though?

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableRowSkinTest.java
Maran23: I think there are some when switching the `TableRow`, as this should remove all old rows and gc them at one point.

Other than that, I think there is no case where we gc cells. When we change the viewport width/height, all rows (cells if a `fixedCellSize` is set) will be piled / cached, but not destroyed.

Maran23:
> I think the changes look good. I'm a bit confused in the performance table with what is meant with the `50 ms -> 0 ms` in the "after" cases though?

Every `refresh()` will trigger 2 layouts for some reason, where the second one does nothing as nothing is dirty, so basically a noop. I can have a look into that (maybe as a follow up?) but I remember that this happens sometimes in general for the `VirtualFlow` and we should check that generally at one point. 

@andy-goryachev-oracle solved that problem in the `RichTextArea` by isolating the method which should be called e.g. two times (due to e.g. `ScrollBar`s) instead of a real relayout.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TreeTableRowSkinTest.java
hjohn: Alright, as long as there are still some GC checks I think this would be fine.  Not sure if I agree with keeping cells/rows around that exceed the amount that are currently displayed, but that's how it already works (my own virtualized controls will only reference the cells displayed, but with non-fixed size cells/rows this may not be optimal).

hjohn:
> > I think the changes look good. I'm a bit confused in the performance table with what is meant with the `50 ms -> 0 ms` in the "after" cases though?
> 
> Every `refresh()` will trigger 2 layouts for some reason, where the second one does nothing as nothing is dirty, so basically a noop. I can have a look into that (maybe as a follow up?) but I remember that this happens sometimes in general for the `VirtualFlow` and we should check that generally at one point.

No need to address that in this PR,  I was just confused what the numbers meant (shouldn't the `before` column than not also have `X ms -> 0 ms`?).  So it seems like quite a good performance improvement.

As a side note, even 30-40 ms seems incredibly slow, that's bound to create noticeable input lag or frame skips :/  How many cells were visible?  1000 or 100x1000?  If the latter, than 30-40 ms seems okayish.

hjohn (approved):

Maran23:
> As a side note, even 30-40 ms seems incredibly slow, that's bound to create noticeable input lag or frame skips :/ How many cells were visible? 1000 or 100x1000? If the latter, than 30-40 ms seems okayish.

I agree. One problem here is, that all cells will be updated (via `updateItem`) of a `TableRow`, even if not visible. 

I counted all `updateItem` calls, results below.

Code from Benchmark above:
- `TableRow` `updateItem`: 78
- `TableCell` `updateItem`: 7800

39 rows are displayed, and they are updated twice (first with `-1` to reset, then with the actual index). And all rows have 100 cells.

A `TableCell` `updateItem` without any code (no `setText`, no `setGraphic`) is indeed faster, around 10-20ms.
Looking into the code, there are some unnecessary `requestLayout` calls as well.

modules/javafx.controls/src/main/java/javafx/scene/control/TableView.java
```diff
@@ -1763,16 +1763,16 @@ public void sort() {
     }
 
     /**
-     * Calling {@code refresh()} forces the TableView control to recreate and
-     * repopulate the cells necessary to populate the visual bounds of the control.
+     * Calling {@code refresh()} forces the TableView control to repopulate the
+     * cells necessary to populate the visual bounds of the control.
```
andy-goryachev-oracle: the word 'repopulate' is used twice (here and elsewhere)

I would suggest to rephrase these comments to indicate what happens exactly, possibly borrowing text from the `VirtualFlow`:

- recreate: a layout pass should be done, and that the cell factory has changed. All cells in the viewport are recreated with the new cell factory.
- rebuild: a layout pass should be done, and cell contents have changed. All cells are removed and then added properly in the viewport

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
```diff
@@ -297,19 +296,18 @@ public TableViewSkinBase(final C control) {
...
             if (!c.wasAdded()) {
                 return;
             }
-            if (Properties.REFRESH.equals(c.getKey())) {
-                refreshView();
-                getSkinnable().getProperties().remove(Properties.REFRESH);
-            } else if (Properties.RECREATE.equals(c.getKey())) {
-                needCellsRecreated = true;
-                refreshView();
+            if (Properties.RECREATE.equals(c.getKey())) {
```
andy-goryachev-oracle, line 305: I just have to raise this concern.

It looks weird to me to use properties as a roundabout way to invoke a hidden method in the skin.  Node properties, a public facility, can be easily mutated by unrelated code, making it easy to break the intended functionality.

Why not make these two methods explicitly a part of the public API by replacing `requestRebuildCells` and `requestRecreateCells` with

```
protected VirtualContainerBase.rebuildCells()
protected VirtualContainerBase.recreateCells()
```

? 

(requestXXX is a misnomer - it might suggest an async nature, whereas it these are synchronous methods)

BlueGoliath:
What needs to happen for this to get merged? Testing?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
kevinrushforth: Changing this is out of scope for this PR.

kevinrushforth: Having said that, I've always disliked using user properties for this sort of thing. It feels like a bit of a hack, so it might be worth filing a follow-up to evaluate a different approach in general (not just for this one instance of the pattern). I don't see this as a high priority, though.

kevinrushforth:
@Maran23 Can you add the performance test program you ran to validate the performance numbers to this PR? A new directory under `tests/performance` seems a good place for it.

I'll review the CSR.

@johanvos had some comments earlier, so allow time for him to respond as well.

modules/javafx.controls/src/main/java/javafx/scene/control/TableView.java
kevinrushforth: At a minimum, replace the first occurrence of "repopulate" with "rebuild".

```
     * Calling {@code refresh()} forces the TableView control to rebuild the
     * cells necessary to populate the visual bounds of the control.
```

I wouldn't over-specify this by saying what `VirtualFlow` will do, but if you want to add a sentence saying that this will request a layout that would be fine:

```
     * Calling {@code refresh()} forces the TableView control to rebuild the
     * cells necessary to populate the visual bounds of the control.
     * This will request a layout of the TableView cells.
```

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
Maran23: Adding to what @kevinrushforth said, this is used everywhere. All layout constraints are implemented by using `Properties`. Changing anything there can break the layout.

If we just talking about the properties calling a method in the skin, I dislike the current way as well, but have no other idea really. Maybe a good topic for the mailing list then.

Maran23:
> @Maran23 Can you add the performance test program you ran to validate the performance numbers to this PR? A new directory under `tests/performance` seems a good place for it.

Sure. I don't know how easy it is, but it might be worth to consider using `JMH` in the future. No idea if this is easy to wire up with JavaFX though.

johanvos:
> Every `refresh()` will trigger 2 layouts for some reason, where the second one does nothing as nothing is dirty, so basically a noop. I can have a look into that (maybe as a follow up?) but I remember that this happens sometimes in general for the `VirtualFlow` and we should check that generally at one point.

This is something that worries me. One of the main issues I see with e.g. VirtualFlow, is that some methods can be invoked both during a rendering pulse as well as (indirectly) via explicit invocations, most often by code invoked with Platform.runLater().
Depending on whether code is called during a pulse or not, the behavior can be very different. A major problematic consequence of a method that used to be called outside the pulse, and that is now for some reason (e.g. due to concurrency, as it depends whether the Runnable submitted to Platform.runLater() is executed before or after the pulse) called during a pulse, is that this can lead to flickering. If `requestNextPulse` is called during the layout phase, it is very well possible that a "wrong" layout is rendered briefly before the correct layout is shown.

I am not saying that this PR makes the situation worse or better, but I think there is a reasonable chance that some applications will behave differently (and show flickering). Having said that, I don't think that this PR can be blamed in case there is a different behavior.

modules/javafx.controls/src/main/java/javafx/scene/control/TableView.java
Maran23: Changed to rebuild. I did not add the request layout line, in case we may want to change this later. Since as @hjohn and @johanvos mentioned, it is rather weird right now.

Maran23:
> This is something that worries me. One of the main issues I see with e.g. VirtualFlow, is that some methods can be invoked both during a rendering pulse as well as (indirectly) via explicit invocations, most often by code invoked with Platform.runLater(). Depending on whether code is called during a pulse or not, the behavior can be very different. A major problematic consequence of a method that used to be called outside the pulse, and that is now for some reason (e.g. due to concurrency, as it depends whether the Runnable submitted to Platform.runLater() is executed before or after the pulse) called during a pulse, is that this can lead to flickering. If `requestNextPulse` is called during the layout phase, it is very well possible that a "wrong" layout is rendered briefly before the correct layout is shown.

I'm also not happy with the current situation. There are also some cases where cells request a layout while they are layouted. There is quite some room to optimize several scenarios here.
I would like to check this out and optimize at one point, but really afraid that there might be regressions. So this will probably take a while, but I have a far better understanding of the `VirtualFlow` and surrounding classes than e.g. a year ago + we have far more tests as well!

Maran23:
There seems to be an intermediate test failure, maybe a race condition unrelated to this change:

```
TaskEventTest > cancelledCalledAfterHandler() FAILED
        at app//test.javafx.concurrent.TaskEventTest.cancelledCalledAfterHandler(TaskEventTest.java:410)
```

kevinrushforth:
> There seems to be an intermediate test failure, maybe a race condition unrelated to this change:

Yes, it is already filed and tracked by [JDK-8357459](https://bugs.openjdk.org/browse/JDK-8357459).

modules/javafx.controls/src/main/java/javafx/scene/control/TableView.java
andy-goryachev-oracle: Thank you, though I would insist on actually explaining what "rebuild" means, as it is not clear from the context.

`VirtualFlow` offers more detailed explanation, so perhaps we should borrow that.

kevinrushforth: Or maybe link to `VirtualFlow`? I'm not sure the details are important enough to repeat.

andy-goryachev-oracle: I would rather explain it at the top level, since it is the class the app devs use, instead of referring to some part of the skin in another package.

And what if the control uses a custom skin that is not based on the `VirtualFlow`?

In fact, the public method in this Control is the normative one, so it must define the expected behavior for all the implementors to follow.

kevinrushforth: > In fact, the public method in this Control is the normative one, so it must define the expected behavior for all the implementors to follow.

You're absolutely right.

Maran23: > And what if the control uses a custom skin that is not based on the `VirtualFlow`?

But that can also be used against using the description from `VirtualFlow`. Right now, I would insist to not add any more details, as the rebuild implementation may change later on. 

Right now, we remove and add them to the viewport, but this is not really needed. What we want to achieve is that `updateItem` is called, and this way ensures that. But there are other, better and faster ways to do so. Something I might revisit at one point in the future.

And since the `In other words` explains why this useful, I think we are good here.

andy-goryachev-oracle: I still don't know what "rebuild" means in this context.

kevinrushforth: We should be careful not to over-specify this. The "in other words" part already suggests what it does. Maybe it could be made more explicit by adding "to match the underlying data source", like this?

```
     * In other words, this forces the TableView to update what it is showing to
     * the user to match the underlying data source.
     * This is useful in cases where the underlying data source has
     * changed in a way that is not observed by the TableView itself.
```

Maran23: I like this idea.
If we want to explain what rebuild does, I would probably mention that this means the cells are updated. So maybe instead of `rebuild`, we could use something like `re-update`, maybe?

kevinrushforth: Yes, mentioning that the cells are updated is good (and is what I was trying to add). I'd stick with "rebuild" in the first sentence, though, since that is used in VirtualFlow.

andy-goryachev-oracle: I don't understand what "rebuild" is, **in this context**.
There is no mention of VirtualFlow here.

Discard cells and create new?  Calls `updateItem()` on them?  Something else?

kevinrushforth: I think I see what you are getting at now. Let's take a step back.

What we want is to give the app developer enough information to know what the purpose of calling this method is, and what the effect will be. What we don't want to do is say _how_ that is done. So you are right that we shouldn't appeal to VirtualFlow since it isn't really relevant in this context. Likewise, we don't want to constrain it with implementation details.

The `refresh()` method was added by [JDK-8098085](https://bugs.openjdk.org/browse/JDK-8098085) to allow an app developer with a custom cell to say, in effect, "even if it doesn't look like anything is changed, get the contents of the cells anyway". If that's what `updateItem()` does, then yes we can say that. Otherwise, we need to find some other way to say it.

andy-goryachev-oracle: exactly, thank you @kevinrushforth

andy-goryachev-oracle: Discussed with Kevin yesterday, what do you think of this version:
```
     * Calling {@code refresh()} forces the XXX to update what it is showing to
     * the user. This is useful in cases where the underlying data source has
     * changed in a way that is not observed by the XXX itself.
```

kevinrushforth: That looks good. You might even simplify it further and remove `Calling {@code refresh()}`, since it's redundant, and just say `Forces the XXX ...`

kevinrushforth: @Maran23 Can you update the javadoc for the four methods as Andy suggested above (possibly with my simplification) and then update the CSR Specification section to match? I think that's the only remaining thing needed to move this forward.

kevinrushforth:
@Maran23 In case you missed it, this is pending changes from you.

1. I noted some final doc changes in [this comment above](#discussion_r2455868260).
2. I had added an [earlier comment](#issuecomment-3393245504) about adding your perf test to the PR somewhere under `tests/performance`.

Maran23:
> @Maran23 In case you missed it, this is pending changes from you.
> 
>     1. I noted some final doc changes in [this comment above](#discussion_r2455868260).
> 
>     2. I had added an [earlier comment](#issuecomment-3393245504) about adding your perf test to the PR somewhere under `tests/performance`.

On my list, thanks for the reminder! Did not had any time the last days, but I saw all comments and suggestions!

modules/javafx.controls/src/main/java/javafx/scene/control/TableView.java
Maran23: > Discussed with Kevin yesterday, what do you think of this version:

I like it! Will update today.

Maran23:
I was confused when rerunning the benchmarks because suddenly I got 3 `layoutChildren` calls without the fix.
Turns out that this is due to the fix made in https://github.com/openjdk/jfx/pull/1879 (JDK-8360940).

I made new benchmarks and edited the description:

Before fix & without JDK-8360940 | Before fix (master) | With fix | With fix & with JDK-8370498 |
| - | - | - | - |

It looks like with https://github.com/openjdk/jfx/pull/1945, the startup got a bit faster (which I think might be expected since we can save unnecessary layouts)

kevinrushforth (commented):
The latest version of the API docs looks good. The CSR needs to be updated with the diffs between master and the latest version.

tests/performance/control/src/table/TableRefreshTest.java
```diff
@@ -0,0 +1,108 @@
...
+            }
+
+            BorderPane root = new BorderPane();
+            root.setPadding(new Insets(4));
+            root.setCenter(tv);
+
+            Button refreshBtn = new Button("Refresh");
+            refreshBtn.setOnAction(_ -> tv.refresh());
+            Button recreateBtn = new Button("Recreate");
+            recreateBtn.setOnAction(_ -> tv.getProperties().put("recreateKey", true));
```
kevinrushforth, line 77: Question: Can this be done with public API? It's OK as is, since it is a test, but I figured I'd ask.

Maran23: Not like this, but setting a new `TableRow` factory will essentially trigger and do the same:
https://github.com/openjdk/jfx/blob/a4e5451373beed93942e0bf2b6c8b81d3e6ad66d/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java#L331-L337

kevinrushforth: OK, thanks. I would leave the test as you have it.

PR #1605, opened 2024-10-20
8313424: JavaFX controls in the title bar (Preview)

mstr2:
Implementation of [`StageStyle.EXTENDED`](https://gist.github.com/mstr2/0befc541ee7297b6db2865cc5e4dbd09).

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change requires CSR request [JDK-8356115](https://bugs.openjdk.org/browse/JDK-8356115) to be approved
- [x] Change must be properly reviewed (2 reviews required, with at least 2 [Reviewers](https://openjdk.org/bylaws#reviewer))

### Issues
 * [JDK-8313424](https://bugs.openjdk.org/browse/JDK-8313424): JavaFX controls in the title bar (Preview) (**Enhancement** - P4)
 * [JDK-8356115](https://bugs.openjdk.org/browse/JDK-8356115): JavaFX controls in the title bar (Preview) (**CSR**)

### Reviewers
 * [Andy Goryachev](https://openjdk.org/census#angorya) (@andy-goryachev-oracle - **Reviewer**)
 * [Markus Mack](https://openjdk.org/census#mmack) (@drmarmac - Author)
 * [Kevin Rushforth](https://openjdk.org/census#kcr) (@kevinrushforth - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1605/head:pull/1605` \
`$ git checkout pull/1605`

Update a local copy of the PR: \
`$ git checkout pull/1605` \
`$ git pull https://git.openjdk.org/jfx.git pull/1605/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 1605`

View PR using the GUI difftool: \
`$ git pr show -t 1605`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jfx/pull/1605.diff">https://git.openjdk.org/jfx/pull/1605.diff</a>

</details>
<details><summary>Using Webrev</summary>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1605#issuecomment-2424357745)
</details>

PR #1831, opened 2025-06-15
8359601: Fix window button states of an extended stage

mstr2:
The window button states (disabled/hidden) of extended stages with a `HeaderButtonOverlay` or custom header buttons are inconsistent with what we would expect from the OS (Windows and Linux). To figure out what we would expect, I started with gathering some data. The following table shows the button states of system-decorated windows on various platforms:

#### Windows 11

| Window attributes | Iconify | Maximize | Close |
|---|---|---|---|
| resizable, not modal | visible | visible | visible |
| not resizable, not modal | visible | visible, disabled | visible |
| resizable, modal | visible, disabled | visible | visible |
| not resizable, modal | hidden | hidden | visible, utility-style |

#### Ubuntu 24 / Fedora 41 (GNOME)

| Window attributes | Iconify | Maximize | Close |
|---|---|---|---|
| resizable, not modal | visible | visible | visible |
| not resizable, not modal | visible | hidden | visible |
| resizable, modal | visible, _not working_ | visible, _not working_ | visible |
| not resizable, modal | visible, _not working_ | hidden | visible |

#### Kubuntu 24 (KDE)

| Window attributes | Iconify | Maximize | Close |
|---|---|---|---|
| resizable, not modal | visible | visible | visible |
| not resizable, not modal | visible | hidden | visible |
| resizable, modal | visible, _not working_ | visible | visible |
| not resizable, modal | visible, _not working_ | hidden | visible |

## Observations

1. On Windows, buttons are generally disabled when their operation is not possible with the given window attributes.
   * Exception: modal/non-resizable windows look like utility windows (iconify and maximize are hidden)
2. On GNOME and KDE, buttons are generally hidden when their operation is not possible.
   * Exception: iconify and maximize on modal windows is not hidden, but seems to simply not do anything (bug?)

## Permitted window button operations

Given the gathered observations and some simple logic, this is the table of operations that are permitted for all combinations of modal and resizable window attributes:

| Window attributes | Iconify | Maximize | Close |
|---|---|---|---|
| resizable, not modal | yes | yes | yes |
| not resizable, not modal | yes | no | yes |
| resizable, modal | no | yes | yes |
| not resizable, modal | no | no | yes |

## Fixes

This PR includes the following changes:
1. Unused code relating to window modality is removed.
2. The disabled states of `HeaderButtonOverlay` as well as `HeaderButtonBehavior` are changed to match the table above.
3. The stylesheets for GNOME and KDE are changed such that disabled buttons are hidden.
4. The stylesheet for Windows is changed such that a modal/non-resizable window looks like a utility window.

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/glass/ui/HeaderButtonOverlay.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/glass/ui/HeaderButtonOverlay.java
@@ -26,7 +26,6 @@
 package com.sun.glass.ui;
 
 import com.sun.glass.events.MouseEvent;
-import com.sun.javafx.binding.ObjectConstant;
 import com.sun.javafx.util.Utils;
 import javafx.beans.property.DoubleProperty;
 import javafx.beans.property.ObjectProperty;
@@ -264,12 +263,15 @@ public class HeaderButtonOverlay extends Region {
     private final ButtonRegion maximizeButton = new ButtonRegion(HeaderButtonType.MAXIMIZE, "-FX-INTERNAL-maximize-button", 1);
     private final ButtonRegion closeButton = new ButtonRegion(HeaderButtonType.CLOSE, "-FX-INTERNAL-close-button", 2);
     private final Subscription subscriptions;
+    private final boolean modal;
     private final boolean utility;
     private final boolean rightToLeft;
 
     private Node buttonAtMouseDown;
 
-    public HeaderButtonOverlay(ObservableValue<String> stylesheet, boolean utility, boolean rightToLeft) {
+    public HeaderButtonOverlay(ObservableValue<String> stylesheet, boolean modal,
+                               boolean utility, boolean rightToLeft) {
+        this.modal = modal;
         this.utility = utility;
         this.rightToLeft = rightToLeft;
 
@@ -310,10 +312,7 @@ public class HeaderButtonOverlay extends Region {
         getStyleClass().setAll("-FX-INTERNAL-header-button-container");
 
         if (utility) {
-            iconifyButton.managedProperty().bind(ObjectConstant.valueOf(false));
-            maximizeButton.managedProperty().bind(ObjectConstant.valueOf(false));
             getChildren().add(closeButton);
-            getStyleClass().add(UTILITY_STYLE_CLASS);
         } else {
             getChildren().addAll(iconifyButton, maximizeButton, closeButton);
         }
@@ -450,6 +449,9 @@ public class HeaderButtonOverlay extends Region {
     }
 
     private void onResizableChanged(boolean resizable) {
+        boolean utilityStyle = utility || (modal && !resizable);
+        toggleStyleClass(this, UTILITY_STYLE_CLASS, utilityStyle);
+        iconifyButton.setDisable(utility || modal);
         maximizeButton.setDisable(!resizable);
     }
 
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/scene/layout/HeaderButtonBehavior.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/scene/layout/HeaderButtonBehavior.java
@@ -31,10 +31,12 @@ import javafx.css.PseudoClass;
 import javafx.event.EventHandler;
 import javafx.scene.Node;
 import javafx.scene.Scene;
+import javafx.scene.input.MouseButton;
 import javafx.scene.input.MouseEvent;
 import javafx.scene.layout.HeaderButtonType;
 import javafx.stage.Modality;
 import javafx.stage.Stage;
+import javafx.stage.StageStyle;
 import javafx.util.Subscription;
 import java.util.Objects;
 
@@ -54,19 +56,22 @@ public final class HeaderButtonBehavior implements EventHandler<MouseEvent> {
             .flatMap(Scene::windowProperty)
             .map(w -> w instanceof Stage s ? s : null);
 
-        if (type == HeaderButtonType.MAXIMIZE) {
-            subscription = Subscription.combine(
-                stage.flatMap(Stage::resizableProperty).subscribe(this::onResizableChanged),
-                stage.flatMap(Stage::fullScreenProperty).subscribe(this::onFullScreenChanged),
-                stage.flatMap(Stage::maximizedProperty).subscribe(this::onMaximizedChanged),
-                () -> node.removeEventHandler(MouseEvent.MOUSE_RELEASED, this)
-            );
-        } else {
-            subscription = Subscription.combine(
-                stage.flatMap(Stage::fullScreenProperty).subscribe(this::onFullScreenChanged),
-                () -> node.removeEventHandler(MouseEvent.MOUSE_RELEASED, this)
-            );
+        Subscription subscription = Subscription.combine(
+            stage.flatMap(Stage::fullScreenProperty).subscribe(this::onFullScreenChanged),
+            () -> node.removeEventHandler(MouseEvent.MOUSE_RELEASED, this));
+
+        if (type != HeaderButtonType.CLOSE) {
+            subscription = Subscription.combine(subscription,
+                stage.subscribe(this::onStageChanged),
+                stage.flatMap(Stage::resizableProperty).subscribe(this::onResizableChanged));
         }
+
+        if (type == HeaderButtonType.MAXIMIZE) {
+            subscription = Subscription.combine(subscription,
+                stage.flatMap(Stage::maximizedProperty).subscribe(this::onMaximizedChanged));
+        }
+
+        this.subscription = subscription;
 
         node.addEventHandler(MouseEvent.MOUSE_RELEASED, this);
 
@@ -81,7 +86,7 @@ public final class HeaderButtonBehavior implements EventHandler<MouseEvent> {
 
     @Override
     public void handle(MouseEvent event) {
-        if (!node.getLayoutBounds().contains(event.getX(), event.getY())) {
+        if (!node.getLayoutBounds().contains(event.getX(), event.getY()) || event.getButton() != MouseButton.PRIMARY) {
             return;
         }
 
@@ -114,9 +119,24 @@ public final class HeaderButtonBehavior implements EventHandler<MouseEvent> {
         return scene.getWindow() instanceof Stage stage ? stage : null;
     }
 
-    private void onResizableChanged(Boolean resizable) {
+    private void onStageChanged(Stage stage) {
+        if (stage != null) {
+            boolean utility = stage.getStyle() == StageStyle.UTILITY;
+            boolean modal = stage.getModality() != Modality.NONE;
+            updateDisableState(utility, modal, stage.isResizable());
+        }
+    }
+
+    private void onResizableChanged(Boolean unused) {
+        onStageChanged(getStage());
+    }
+
+    private void updateDisableState(boolean utility, boolean modal, boolean resizable) {
         if (!node.disableProperty().isBound()) {
-            node.setDisable(resizable == Boolean.FALSE);
+            switch (type) {
+                case ICONIFY -> node.setDisable(utility || modal);
+                case MAXIMIZE -> node.setDisable(!resizable);
+            }
         }
     }
 
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassWindow.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassWindow.m
@@ -1404,69 +1404,6 @@ JNIEXPORT void JNICALL Java_com_sun_glass_ui_mac_MacWindow__1toBack
     GLASS_CHECK_EXCEPTION(env);
 }
 
-
-/*
- * Class:     com_sun_glass_ui_mac_MacWindow
- * Method:    _enterModal
- * Signature: (J)V
- */
-JNIEXPORT void JNICALL Java_com_sun_glass_ui_mac_MacWindow__1enterModal
-(JNIEnv *env, jobject jWindow, jlong jPtr)
-{
-    LOG("Java_com_sun_glass_ui_mac_MacWindow__1enterModal");
-    if (!jPtr) return;
-
-    GLASS_ASSERT_MAIN_JAVA_THREAD(env);
-    GLASS_POOL_ENTER;
-    {
-        GlassWindow *window = getGlassWindow(env, jPtr);
-        [NSApp runModalForWindow:window->nsWindow];
-    }
-    GLASS_POOL_EXIT;
-    GLASS_CHECK_EXCEPTION(env);
-}
-
-/*
- * Class:     com_sun_glass_ui_mac_MacWindow
- * Method:    _enterModalWithWindow
- * Signature: (JJ)V
- */
-JNIEXPORT void JNICALL Java_com_sun_glass_ui_mac_MacWindow__1enterModalWithWindow
-(JNIEnv *env, jobject jWindow, jlong jDialogPtr, jlong jWindowPtr)
-{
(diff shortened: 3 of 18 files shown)


credmond:
Most of this seems to be working well!

But I have found a likely bug. Hopefully fairly self-explanatory from the code.

Minimise icon is normally disabled, but still enabled when EXTENDED is used & when dialog is resizable. 

Just flip comment on `dialog.initStyle(StageStyle.EXTENDED)` to see difference.


```
import javafx.scene.Scene;
import javafx.scene.control.ButtonType;
import javafx.scene.control.Dialog;
import javafx.scene.layout.StackPane;
import javafx.stage.Modality;
import javafx.stage.Stage;
import javafx.stage.StageStyle;

public class MinimiseIconEnabledBug extends Application {
    @Override
        Button button = new Button("Click");
        button.setOnAction(e -> {
            final Dialog<Object> dialog = new Dialog<>();
            dialog.initOwner(primaryStage);
            // Existing behaviour: minimise icon disabled when commented
            // Possible "bug": Minimise icon enabled when uncommented
            // dialog.initStyle(StageStyle.EXTENDED);
            dialog.initModality(Modality.NONE);
            dialog.setResizable(true); // This is important
            dialog.setTitle("My Dialog");
            dialog.setContentText("Lorem ipsum dolor sit amet, consectetur adipiscing...");
            dialog.getDialogPane().getButtonTypes().addAll(ButtonType.OK);
            dialog.show();
        });

        StackPane root = new StackPane(button);
        Scene scene = new Scene(root, 300, 200);

        primaryStage.setTitle("My First JavaFX App");
        primaryStage.setScene(scene);
        primaryStage.show();
    }

        launch(args);
    }
}
```

credmond:
Another suspected issue (not really related to this PR), close request handlers are not called when using EXTENDED. E.g.:

```
        stage.setOnCloseRequest(event -> {
            System.out.println("Never called...");;
        });
```

Same for dialogs...

mstr2:
Good catch, a non-modal owned window also can't be iconified. I've added code and tests for this scenario.

credmond:
> Another suspected issue (not really related to this PR), close request handlers are not called when using EXTENDED. E.g.:
> 
> ```
>         stage.setOnCloseRequest(event -> {
>             System.out.println("Never called...");;
>         });
> ```
> 
> Same for dialogs...

Sample:

```
import javafx.scene.Scene;
import javafx.scene.control.ButtonType;
import javafx.scene.control.Dialog;
import javafx.scene.layout.StackPane;
import javafx.stage.Modality;
import javafx.stage.Stage;
import javafx.stage.StageStyle;

public class MinimiseIconEnabledBug extends Application {
    @Override
        Button button = new Button("Click");
        button.setOnAction(e -> {
            final Dialog<Object> dialog = new Dialog<>();
            dialog.initOwner(primaryStage);
            // dialog.initStyle(StageStyle.EXTENDED); // uncomment for bug
            dialog.initModality(Modality.NONE);
            dialog.setResizable(true); // This is important
            dialog.getDialogPane().getButtonTypes().addAll(ButtonType.OK);
            dialog.setOnCloseRequest(dialogEvent -> {
                System.out.println("Close the dialog pane. This is not called when EXTENDED is used.");
            });
            dialog.show();
        });

        StackPane root = new StackPane(button);
        Scene scene = new Scene(root, 300, 200);
        primaryStage.setScene(scene);
        primaryStage.show();
    }

        launch(args);
    }
}
```

mstr2:
I've filed another ticket for that: [JDK-8359763](https://bugs.openjdk.org/browse/JDK-8359763).

credmond:
> Good catch, a non-modal owned window also can't be iconified. I've added code and tests for this scenario.

Looks good, can't find anymore issues...

tsayao:
JavaFX modal windows are not native modals, as evidenced by the code removal on this PR (it's not used) . On GNOME,  native modal windows automatically remove the minimize and maximize buttons. 

image

I did an experiment and found that it's possible to get native modal windows (at least with GTK Glass) by passing the modality flag from `WindowStage` to `createWindow`. I haven’t done extensive testing, but it seems to work - likely because the modality can’t be changed afterward.

mstr2:
> JavaFX modal windows are not native modals, as evidenced by the code removal on this PR (it's not used) . On GNOME, native modal windows automatically remove the minimize and maximize buttons.
> 
> I did an experiment and found that it's possible to get native modal windows (at least with GTK Glass) by passing the modality flag from `WindowStage` to `createWindow`. I haven’t done extensive testing, but it seems to work - likely because the modality can’t be changed afterward.

That's a good idea for an enhancement.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/HeaderButtonOverlay.java
```diff
@@ -264,12 +263,15 @@ protected void invalidated() {
...
     private final ButtonRegion closeButton = new ButtonRegion(HeaderButtonType.CLOSE, "-FX-INTERNAL-close-button", 2);
     private final Subscription subscriptions;
+    private final boolean modal;
     private final boolean utility;
     private final boolean rightToLeft;
 
     private Node buttonAtMouseDown;
 
-    public HeaderButtonOverlay(ObservableValue<String> stylesheet, boolean utility, boolean rightToLeft) {
+    public HeaderButtonOverlay(ObservableValue<String> stylesheet, boolean modal,
```
drmarmac: it's a bit confusing to have `isModal` in `Window` and `getModality` in `Stage` on one hand, and a `modal` field / parameter here, which is set to `isModal() || getOwner() != null` by callers. Maybe clarify by renaming to something like `effecitvelyModal` or `modalOrOwned`?

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/layout/HeaderButtonBehavior.java
```diff
@@ -114,9 +136,32 @@ private Stage getStage() {
...
+        if (getStage() instanceof Stage stage) {
+            updateButtonDisableState(stage);
+        }
+    }
+
+    private void updateButtonDisableState(Stage stage) {
         if (!node.disableProperty().isBound()) {
-            node.setDisable(resizable == Boolean.FALSE);
+            boolean utility = stage.getStyle() == StageStyle.UTILITY;
+            boolean modal = stage.getOwner() != null || stage.getModality() != Modality.NONE;
```
drmarmac: in case `HeaderButtonOverlay.modal` is renamed, this variable could be renamed in the same way.

modules/javafx.graphics/src/test/java/test/com/sun/glass/ui/HeaderButtonOverlayTest.java
```diff
@@ -322,19 +324,67 @@ void customButtonOrder_rightToLeft() {
...
+
+        ButtonDisabledStateTest(boolean resizable, boolean modal, boolean iconifyDisabled, boolean maximizeDisabled) {
+            this.resizable = resizable;
+            this.modal = modal;
+            this.iconifyDisabled = iconifyDisabled;
+            this.maximizeDisabled = maximizeDisabled;
+        }
+
+        final boolean resizable;
+        final boolean modal;
```
drmarmac: same here

drmarmac (commented):
The updated behavior looks good on windows 11, I tested various combinations of modal/owned/resizable/StageStyle settings, including modifying the resizeable bit when a stage is already showing.

I also checked the behavior in macOS Sequoia 15.5. There, a window where `setResizable(false)` is called before showing the stage, unexpectedly has a working maximize button even without `StageStyle.EXTENDED`.
Looks like this was not the case prior to the integration of #1695, so this behavior was probably introduced there and may need a fix in this PR or elsewhere.

I don't see any code using the java and native enterModal/exitModal methods which were removed, so this looks like a safe change.
Finally, I left some comments regarding potentially confusing definitions of `modal`, otherwise the code looks good.

credmond:
> I've filed another ticket for that: [JDK-8359763](https://bugs.openjdk.org/browse/JDK-8359763).

Just curious, why a separate bug for this instead of fixing in this PR and avoiding the headaches/conversations of more PRs -- is it more complicated that it sounds?

mstr2:
> > I've filed another ticket for that: [JDK-8359763](https://bugs.openjdk.org/browse/JDK-8359763).
> 
> Just curious, why a separate bug for this instead of fixing in this PR and avoiding the headaches/conversations of more PRs -- is it more complicated that it sounds?

Shorter PRs tend to get more timely reviews. That being said, since it is closely related, I've added the fix to this PR.

/issue add JDK-8359763

mstr2:
> I also checked the behavior in macOS Sequoia 15.5. There, a window where `setResizable(false)` is called before showing the stage, unexpectedly has a working maximize button even without `StageStyle.EXTENDED`. Looks like this was not the case prior to the integration of #1605, so this behavior was probably introduced there and may need a fix in this PR or elsewhere.

The implementation relied on the `_setResizable` JNI method being called in a very particular way that has since changed (internally, it would call another method with the same name, but the implementation would actually _toggle_ the resizable state, not _set_ it to a given value). I've fixed this problem.

credmond:
I have found another bug/issue, which is again related to wrapping a root node in a StackPane (something that is common in, for example, ControlsFx). Although the observed behaviour is probably strictly logically correct, I think this should be prevented (e.g, HeaderBar has/gets special meaning/treatment).

```
public class DisappearingHeaderBarNodesBug extends Application {

    @Override
        BorderPane root = new BorderPane();
        Button button = new Button("Click for \"bug()\"!");
        button.setOnAction(_ -> bug(root));
        root.setCenter(new VBox(new Label("Something above"), button, new Label("Something below")));
        root.setTop(getHeaderBar());

        primaryStage.initStyle(StageStyle.EXTENDED);
        primaryStage.setScene(new Scene(root, 300, 200));
        primaryStage.show();
    }

    // Wrapping the root node in a stackpane is common
    private void bug(Pane root) {
        StackPane stackPane = new StackPane();
        root.getScene().setRoot(stackPane);
        stackPane.getChildren().addFirst(root);
    }

    private HeaderBar getHeaderBar() {
        HeaderBar headerBar = new HeaderBar();
        headerBar.setCenter(new Label("!!! HeaderBar !!!"));
        headerBar.setLeading(new Label("L"));
        return headerBar;
    }

        launch(args);
    }
}
```

As you can see, after you click the button, and resize, the HeaderBar can disappear completely:

https://github.com/user-attachments/assets/93a320a5-0010-4941-a2a4-9e0e65cbd57d

mstr2:
@credmond We should keep this PR focused, let's use the mailing list or JBS to discuss other issues.

credmond:
FYI to any reviewer: I've been using this branch's EXTENDED/HeaderBar extensively now without any new issues...

mstr2:
I think this fix should make the 25 release if possible, as it would be unfortunate to preview the new feature with distracting bugs included. The fix is also quite well-tested at this point.

@kevinrushforth @andy-goryachev-oracle Do you have any review capacity to spare for this within the current RDP?

credmond:
> I think this fix should make the 25 release if possible, as it would be unfortunate to preview the new feature with distracting bugs included. The fix is also quite well-tested at this point.
> 
> @kevinrushforth @andy-goryachev-oracle Do you have any review capacity to spare for this within the current RDP?

Agreed. It wouldn't take long until others come across this and raise the same issue.

kevinrushforth:
> I think this fix should make the 25 release if possible, as it would be unfortunate to preview the new feature with distracting bugs included. The fix is also quite well-tested at this point.
> 
> @kevinrushforth @andy-goryachev-oracle Do you have any review capacity to spare for this within the current RDP?

Yes, I'll review it this week. It seems a good candidate for backporting to jfx25.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/layout/HeaderButtonBehavior.java
```diff
@@ -54,20 +58,24 @@ public HeaderButtonBehavior(Node node, HeaderButtonType type) {
...
-        } else {
-            subscription = Subscription.combine(
-                stage.flatMap(Stage::fullScreenProperty).subscribe(this::onFullScreenChanged),
-                () -> node.removeEventHandler(MouseEvent.MOUSE_RELEASED, this)
-            );
+                () -> { if (getStage() instanceof Stage s) s.removeEventFilter(WindowEvent.WINDOW_SHOWING, this); });
+        }
+
+        if (type == HeaderButtonType.MAXIMIZE) {
+            subscription = Subscription.combine(subscription,
```
andy-goryachev-oracle, line 73: here and on L66 the `subscription` pointer (set in L61) gets overwritten.  does it mean some subscription(s) won't be cancelled in `dispose()`?

edit: please disregard, I see it's combining subscriptions.

modules/javafx.graphics/src/main/native-glass/mac/GlassWindow+Java.m
```diff
@@ -212,34 +212,30 @@ - (void)_grabFocus
     s_grabWindow = window;
 }
 
-- (void)_setResizable
+- (void)_setResizable:(bool)resizable
```
kevinrushforth, line 215: Good. It definitely seems cleaner to pass in resizable as a parameter rather than having this method be a "toggle" resizable as it was previously.

modules/javafx.graphics/src/main/native-glass/mac/GlassWindow.m
```diff
@@ -1161,10 +1162,7 @@ static jlong _createWindowCommonDo(JNIEnv *env, jobject jWindow, jlong jOwnerPtr
     GLASS_POOL_ENTER;
     {
         GlassWindow *window = getGlassWindow(env, jPtr);
-        if (window->isResizable != jResizable)
-        {
-            [window performSelectorOnMainThread:@selector(_setResizable) withObject:nil waitUntilDone:YES];
-        }
+        [window _setResizable:jResizable];
```
kevinrushforth, line 1165: I presume you removed the `performSelectorOnMainThread` because the calling method always calls this on the main thread?

kevinrushforth (commented):
The code looks good. I submitted a set of headful test jobs and will report when done. I want to do a little light manual testing, especially on macOS where there are a few changes in the non-extended case (although the code looks equivalent).

modules/javafx.graphics/src/main/native-glass/mac/GlassWindow.m
mstr2: Yes, this is a call that always happens on the FX thread as a result of calling `javafx.stage.Window.show()`. This is verified just a few lines earlier with `GLASS_ASSERT_MAIN_JAVA_THREAD` (that's always been there, I didn't add it).

kevinrushforth: Also, the calling Java code does a thread check. So we're good.

kevinrushforth:
The headful CI tests passed on Linux and macOS (I will need to manually run it on Windows).

kevinrushforth:
Headful tests passed on Windows.

kapilkumar9976:
Hi @kapilkumar9976, thanks for making a comment in an OpenJDK project!

All comments and discussions in the OpenJDK Community must be made available under the OpenJDK [Terms of Use](https://openjdk.java.net/legal/tou/). If you already are an OpenJDK [Author](https://openjdk.java.net/bylaws#author), [Committer](https://openjdk.java.net/bylaws#committer) or [Reviewer](https://openjdk.java.net/bylaws#reviewer), please click [here](https://bugs.openjdk.java.net/secure/CreateIssue.jspa?pid=11300&issuetype=1) to open a new issue so that we can record that fact. Please Use "Add GitHub user kapilkumar9976" for the summary.

PR #1832, opened 2025-06-17
8358802: Update boot JDK to 24.0.1

arapte:
Support for JDK 24 was added in Gradle 8.14.2. Hence update bootjdk to JDK 24.0.1 and gradle to 8.14.2
No changes to gradle build script were needed for this upgrade.

/issue add JDK-8358800


kevinrushforth:
Reviewers: @kevinrushforth @tiainen or @johanvos 

/reviewers 2

gradlew
```diff
@@ -115,7 +114,7 @@ case "$( uname )" in                #(
   NONSTOP* )        nonstop=true ;;
 esac
 
-CLASSPATH=$APP_HOME/gradle/wrapper/gradle-wrapper.jar
+CLASSPATH="\\\"\\\""
```
kevinrushforth, line 117: This is from the gradle distribution? It looks odd...

kevinrushforth:
Please merge in the latest master to get the updated gradle wrapper-validation action (since you are updating the gradle wrapper, I'd like to see a GHA run with the newer action).

kevinrushforth (approved):
LGTM

gradlew
arapte: Yes, it seems to be an empty string, added when this command is run: `sh gradlew wrapper --gradle-version 8.14.2`. Earlier the gradle-wrapper.jar was used in classpath, but now it is used with -jar option. so CLASSPATH is kept empty.

PR #1834, opened 2025-06-18
8357584: [XWayland] [OL10] Robot.mousePress() is delivered to wrong place

azvegint:
This changeset introduces an adapted version of [the OpenJDK fix](https://github.com/openjdk/jdk/commit/2dfbf41d2a3dbcd44f9ed9a58a1b0932d7536977) for mouse and keyboard interactions with Robot.
More info about the issue itself is available in the PR [description](https://github.com/openjdk/jdk/pull/25265#issue-3068640753)

In short, the currently used XTest for keyboard and mouse interactions may not be suitable for automated testing at some point, as it may require user confirmation to control the mouse or keyboard from time to time.

This fix adds support for the [Remote Desktop XDG portal](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.RemoteDesktop.html).
This allows us to control the keyboard and mouse with Robot on Wayland, even outside the XWayland server (e.g., clicking on window headers and Wayland native apps).

----

* The remote desktop for Robot is enabled by default on GnomeShell 47+
* It can be enabled manually by setting the `javafx.robot.screenshotMethod` system property to `dbusRemoteDesktop` (e.g. it works on Ubuntu 24.04 with GnomeShell 46)
* The key handling might still have bugs.
* The main goal is to add this new functionality.  If there are small issues that can't be solved right away,  I will prefer to fix them in follow up fixes.

---------
### Progress

### Issue
 * [JDK-8357584](https://bugs.openjdk.org/browse/JDK-8357584): [XWayland] [OL10] Robot.mousePress() is delivered to wrong place (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1834/head:pull/1834` \
`$ git checkout pull/1834`

`$ git checkout pull/1834` \
`$ git pull https://git.openjdk.org/jfx.git pull/1834/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1834`

`$ git pr show -t 1834`

</details>

<a href="https://git.openjdk.org/jfx/pull/1834.diff">https://git.openjdk.org/jfx/pull/1834.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1834#issuecomment-2984742915)
</details>


kevinrushforth:

/reviewers 2

kevinrushforth:
@arapte Would you also be able to look into this in case I am not able to?

azvegint:
> I see most checks relate to Gnome and its version - what would be the behavior if we ran the tests on a different desktop env like KDE?

This question has already been raised on the OpenJDK PR. The short answer is that it's an unsupported configuration (I am unsure about JFX), so I didn't test it. If some tuning is required to support it, I would prefer to address that as a follow-up fix.

Please see quote from https://github.com/openjdk/jdk/pull/25265#discussion_r2093639881 below:

---
> BTW how it is expected to work on KDE + wayland?

It is not a supported configuration.However, when I tested it a while ago, it only worked in a single-screen scenario for the Screencast.
There was an issue with the screen streams; they did not specify their `x` and `y` coordinates.
This prevented the correct arrangement of multiple screens.
I don't know if this has been fixed, but the issue is addressed in our code here:

https://github.com/openjdk/jdk/blob/3dd34517000e4ce1a21619922c62c025f98aad44/src/java.desktop/unix/native/libawt_xawt/awt/screencast_portal.c#L132-L155

In this case, the Screencast is simply disabled.

The Remote Desktop uses the same streams from Screencast portal to control the mouse position.

modules/javafx.graphics/src/main/native-glass/gtk/screencast_pipewire.c
```diff
@@ -28,8 +28,73 @@
...
+#include "glass_key.h"
+
+#define JNU_CHECK_EXCEPTION_RETURN(env, y)      \
+    do {                                        \
+        if ((*env)->ExceptionCheck(env)) {      \
+            return (y);                         \
+        }                                       \
+    } while (0)
+
+struct pw_buffer *(*fp_pw_stream_dequeue_buffer)(struct pw_stream *stream);
```
lukostyra, line 42: These PipeWire declarations have been just moved here from the original header. Is this move necessary?

Also (a general thought towards past implementation, not just to your change) I think we should not define those manually at all. There are PipeWire headers defining those that we should include (and I believe we do already include some in `screencast_pipewire.h`

azvegint: > These PipeWire declarations have been just moved here from the original header. Is this move necessary?

It is about clang 17 warning, it was [fixed](https://github.com/openjdk/jdk/pull/23236) on the jdk side.
I am trying to keep the code similar between OpenJDK and OpenJFX, so I brought this fix here to make maintenance easier.

> Also (a general thought towards past implementation, not just to your change) I think we should not define those manually at all. There are PipeWire headers defining those that we should include (and I believe we do already include some in screencast_pipewire.h

The point is basically the same: ease of maintenance between OpenJDK and OpenJFX.
OpenJDK can also be built on very old Linux distributions that lack the necessary headers.

lukostyra (approved):
LGTM

arapte (approved):
LGTM, 
Verified that no test fail on ubuntu 24.04. and changes are very similar to that of JDK.

PR #1836, opened 2025-06-19
8324941: POC for Headless platform for JavaFX

johanvos:
After spending a year in the sandbox repository, the Headless Platform is now ready to be reviewed in the main repository.

### the Headless Platform
The Headless Platform is a top-level com.sun.glass.ui platform that replaces the second-level Monocle-Headless subplatform, that is part of the top-level Monocle platform. 
The platform can be used like any other platform, especially for running headless JavaFX applications, or for running tests (e.g. on CI systems)

### changes
The code for the Headless Platform is in a new package com.sun.glass.ui.headless in the javafx.graphics module, and it does not require a code change in other packages.
This PR adds a simple change in the `build.gradle` file, to make the Headless Platform the standard when running headless tests (instead of using Monocle/Headless)

### enable the Headless Platform
Setting the system property `glass.platform` to `Headless` will select the Headless Platform instead of the default one (either gtk, mac or win).

### testing
`gradlew --info -PHEADLESS_TEST=true -PFULL_TEST=true :systemTests:cleanTest :systemTests:test`
runs all the system tests, apart from the robot tests. There are 2 failing tests, but there are valid reasons for those to fail.

### robot tests
Most of the robot tests are working on headless as well. add `-PUSE_ROBOT` to test those.

---------
### Progress

### Issue
 * [JDK-8324941](https://bugs.openjdk.org/browse/JDK-8324941): POC for Headless platform for JavaFX (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1836/head:pull/1836` \
`$ git checkout pull/1836`

`$ git checkout pull/1836` \
`$ git pull https://git.openjdk.org/jfx.git pull/1836/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1836`

`$ git pr show -t 1836`

</details>

<a href="https://git.openjdk.org/jfx/pull/1836.diff">https://git.openjdk.org/jfx/pull/1836.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1836#issuecomment-2987229433)
</details>


modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
```diff
@@ -0,0 +1,406 @@
...
+        return getGlassEventButton(buttons[0]);
+    }
+
+    int getGlassEventButton(MouseButton button) {
+        if (button == MouseButton.SECONDARY) return MouseEvent.BUTTON_RIGHT;
+        if (button == MouseButton.PRIMARY) return MouseEvent.BUTTON_LEFT;
+        if (button == MouseButton.MIDDLE) return MouseEvent.BUTTON_OTHER;
+        if (button == MouseButton.BACK) return MouseEvent.BUTTON_BACK;
+        if (button == MouseButton.FORWARD) return MouseEvent.BUTTON_FORWARD;
+        return MouseEvent.BUTTON_NONE;
```
mstr2: Use an expression to get compile-time exhaustiveness checks:
```suggestion
        return switch (button) {
            case NONE -> MouseEvent.BUTTON_NONE;
            case PRIMARY -> MouseEvent.BUTTON_LEFT;
            case MIDDLE -> MouseEvent.BUTTON_OTHER;
            case SECONDARY -> MouseEvent.BUTTON_RIGHT;
            case BACK -> MouseEvent.BUTTON_BACK;
            case FORWARD -> MouseEvent.BUTTON_FORWARD;
        };
```

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
```diff
@@ -0,0 +1,406 @@
...
+                case SECONDARY:
+                    modifiers |= KeyEvent.MODIFIER_BUTTON_SECONDARY;
+                    break;
+                case BACK:
+                    modifiers |= KeyEvent.MODIFIER_BUTTON_BACK;
+                    break;
+                case FORWARD:
+                    modifiers |= KeyEvent.MODIFIER_BUTTON_FORWARD;
+                    break;
+            }
```
mstr2: Using an expression can help increase readibility and correctness:
```suggestion
            modifiers |= switch (buttons[i]) {
                case NONE -> KeyEvent.MODIFIER_NONE;
                case PRIMARY -> KeyEvent.MODIFIER_BUTTON_PRIMARY;
                case MIDDLE -> KeyEvent.MODIFIER_BUTTON_MIDDLE;
                case SECONDARY -> KeyEvent.MODIFIER_BUTTON_SECONDARY;
                case BACK -> KeyEvent.MODIFIER_BUTTON_BACK;
                case FORWARD -> KeyEvent.MODIFIER_BUTTON_FORWARD;
            };
```

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
```diff
@@ -0,0 +1,406 @@
...
+        int code = keyCode.getCode();
+        processSpecialKeys(code, true);
+        char[] keyval = getKeyChars(code);
+        int mods = getKeyModifiers();
+        if (view != null) {
+            view.notifyKey(KeyEvent.PRESS, code, keyval, mods);
+            if (keyval.length > 0) {
+                view.notifyKey(KeyEvent.TYPED, 0, keyval, mods);
+            }
+
```
mstr2: Minor: empty line

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
```diff
@@ -0,0 +1,406 @@
...
+        return c == '\000' ? NO_CHAR : new char[] { c };
+    }
+
+
+    private int getKeyModifiers() {
+        int answer = 0;
+        if (this.specialKeys.keyControl) answer = answer | KeyEvent.MODIFIER_CONTROL;
+        if (this.specialKeys.keyShift) answer = answer | KeyEvent.MODIFIER_SHIFT;
+        if (this.specialKeys.keyCommand) answer = answer | KeyEvent.MODIFIER_COMMAND;
+        if (this.specialKeys.keyAlt) answer = answer | KeyEvent.MODIFIER_ALT;
```
mstr2: You can remove four utterances of the word "answer" by using the `|=` operator.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessApplication.java
```diff
@@ -0,0 +1,217 @@
...
+        return 0;
+    }
+
+    @Override
+    protected int staticTimer_getMaxPeriod() {
+        return 1_000_000;
+    }
+
+    @Override
+    protected CommonDialogs.FileChooserResult staticCommonDialogs_showFileChooser(Window owner, String folder, String filename, String title, int type, boolean multipleMode, CommonDialogs.ExtensionFilter[] extensionFilters, int defaultFilterIndex) {
```
mstr2: This is an extremely long line...

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
```diff
@@ -0,0 +1,370 @@
...
+            newX = this.originalX;
+            newY = this.originalY;
+        }
+        notifyResizeAndMove(newX, newY, newWidth, newHeight);
+        return full;
+    }
+
+    private void notifyResizeAndMove(int x, int y, int width, int height) {
+        HeadlessView view = (HeadlessView) getView();
+        //   if (getWidth() != width || getHeight() != height) {
```
mstr2: Why is this code commented out?

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
```diff
@@ -0,0 +1,370 @@
...
+            view.notifyResize(width, height);
+        }
+        //  }
+        if (getX() != x || getY() != y) {
+            notifyMove(x, y);
+        }
+    }
+
+    public Color getColor(int lx, int ly) {
+        int mx = lx;// + getX();
```
mstr2: Why is this code commented out?

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
```diff
@@ -0,0 +1,370 @@
...
+
+        IntBuffer intBuffer = (IntBuffer) pixels.getBuffer();
+
+        for (int i = 0; i < pixels.getHeight(); i++) {
+            int rowIdx = offsetY + i;
+            for (int j = 0; j < pixels.getWidth(); j++) {
+                int idx = rowIdx * stride + offsetX + j;
+                int val = intBuffer.get(i * pixels.getWidth() + j);
+                if (val != 0) {
+                }
```
mstr2: The `if` statement has an empty body.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/NestedRunnableProcessor.java
```diff
@@ -0,0 +1,79 @@
...
+
+        activeRunLoops.push(entry);
+
+        entry.active = true;
+        while (entry.active) {
+            try {
+                runnableQueue.take().run();
+            } catch (Throwable e) {
+                e.printStackTrace();
+                Application.reportException(e);
```
mstr2, line 57: Why do you print the exception, _and_ send it to `Application.reportException()`? The latter already sends it to the uncaught exception handler, where it is printed by default.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
```diff
@@ -0,0 +1,370 @@
...
+        throw new UnsupportedOperationException("Not supported yet.");
+    }
+
+    @Override
+    protected void _exitModal(long ptr) {
+        throw new UnsupportedOperationException("Not supported yet.");
+    }
+
+    @Override
+    protected void _requestInput(long ptr, String text, int type, double width, double height, double Mxx, double Mxy, double Mxz, double Mxt, double Myx, double Myy, double Myz, double Myt, double Mzx, double Mzy, double Mzz, double Mzt) {
```
mstr2: Another very long line.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
```diff
@@ -0,0 +1,406 @@
...
+            }
+            return;
+        }
+        int wx = activeWindow.getX();
+        int wy = activeWindow.getY();
+
+        if (activeWindow != oldWindow) {
+            HeadlessView view = (HeadlessView)activeWindow.getView();
+            int modifiers = 0;
+            view.notifyMouse(MouseEvent.ENTER, MouseEvent.BUTTON_NONE, (int)mouseX-wx, (int)mouseY-wy, (int)mouseX, (int)mouseY, modifiers, true, true);
```
mstr2: Minor: spacing around operators

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
```diff
@@ -0,0 +1,406 @@
...
+            if (oldWindow != null) {
+                HeadlessView oldView = (HeadlessView)oldWindow.getView();
+                if (oldView != null) {
+                    int owx = oldWindow.getX();
+                    int owy = oldWindow.getY();
+                    oldView.notifyMouse(MouseEvent.EXIT, MouseEvent.BUTTON_NONE, (int)mouseX-owx, (int)mouseY-owy, (int)mouseX, (int)mouseY, modifiers, true, true);
```
mstr2: Minor: spacing around operators

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
```diff
@@ -0,0 +1,370 @@
...
+
+    private boolean resizable;
+    private boolean visible;
+    private boolean isFocusable;
+    private boolean enabled;
+    private boolean closed;
+    private float bg_r, bg_g, bg_b;
+    private float alpha;
+    private Pixels icon;
+    private Cursor cursor;
```
mstr2: The following fields are never used, other than assigning a value:
* `ptr`
* `resizable`
* `visible`
* `isFocusable`
* `enabled`
* `closed`
* `bg_r`, `bg_g`, `bg_b`
* `alpha`
* `icon`
* `cursor`

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
```diff
@@ -0,0 +1,370 @@
...
+        int newY = 0;
+        int newWidth = 0;
+        int newHeight = 0;
+        if (maximize && !wasMaximized) {
+            this.originalHeight = this.height;
+            this.originalWidth = this.width;
+            this.originalX = this.x;
+            this.originalY = this.y;
+            newX = 0;
+            newY = 0;
```
mstr2: `newX` and `newY` already have the value 0.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
```diff
@@ -0,0 +1,370 @@
...
+    protected void _requestInput(long ptr, String text, int type, double width, double height, double Mxx, double Mxy, double Mxz, double Mxt, double Myx, double Myy, double Myz, double Myt, double Mzx, double Mzy, double Mzz, double Mzt) {
+        throw new UnsupportedOperationException("Not supported yet.");
+    }
+
+    @Override
+    protected void _releaseInput(long ptr) {
+        throw new UnsupportedOperationException("Not supported yet.");
+    }
+
+    boolean setFullscreen(boolean full) {
```
mstr2: None of the two callers use the return value of this method.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessApplication.java
```diff
@@ -0,0 +1,217 @@
...
+    private final NestedRunnableProcessor processor = new NestedRunnableProcessor();
+    private final HeadlessWindowManager windowManager = new HeadlessWindowManager();
+    private Screen[] screens = null;
+    private HeadlessCursor cursor;
+    private HeadlessRobot activeRobot = null;
+    ByteBuffer frameBuffer;
+
+    private final int MULTICLICK_MAX_X = 20;
+    private final int MULTICLICK_MAX_Y = 20;
+    private final long MULTICLICK_TIME = 500;
```
mstr2: These fields should be `static`.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessApplication.java
```diff
@@ -0,0 +1,217 @@
...
+import java.io.File;
+import java.nio.ByteBuffer;
+import java.nio.IntBuffer;
+
+public class HeadlessApplication extends Application {
+
+    private final NestedRunnableProcessor processor = new NestedRunnableProcessor();
+    private final HeadlessWindowManager windowManager = new HeadlessWindowManager();
+    private Screen[] screens = null;
+    private HeadlessCursor cursor;
```
mstr2: This field is not used in this class, other than assigning a value.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessPlatformFactory.java
```diff
@@ -0,0 +1,128 @@
...
+            } else if (Clipboard.DND.equals(clipboardName)) {
+                return new HeadlessDnDClipboard();
+            } else {
+                throw new IllegalArgumentException("No support for " + clipboardName + " clipboard in headless");
+            }
+        };
+
+    }
+
+    class HeadlessSystemClipboard extends SystemClipboard {
```
mstr2: This class can be `static`.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessPlatformFactory.java
```diff
@@ -0,0 +1,128 @@
...
+            return Clipboard.ACTION_NONE;
+        }
+
+        @Override
+        protected String[] mimesFromSystem() {
+            return new String[0];
+        }
+    }
+
+    class HeadlessDnDClipboard extends SystemClipboard {
```
mstr2: This class can be `static`.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessTimer.java
```diff
@@ -0,0 +1,52 @@
...
+
+import com.sun.glass.ui.Timer;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+
+public class HeadlessTimer extends Timer {
+
+    private static ScheduledThreadPoolExecutor scheduler;
+    private ScheduledFuture task;
```
mstr2: You can use the parameterized type `ScheduledFuture<?>`.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessView.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    @Override
+    protected int _getY(long ptr) {
+        return y;
+    }
+
+    @Override
+    protected void _setParent(long ptr, long parentPtr) {
+        parentPtr = parentPtr;
```
mstr2: You're assigning a variable to itself. But even if you qualify `this.parentPtr`, the assignment seems pointless.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessView.java
```diff
@@ -0,0 +1,130 @@
...
+
+public class HeadlessView extends View {
+
+    private Map capabilities;
+    private int x = 0;
+    private int y = 0;
+    private long parentPtr = 0;
+    private Pixels pixels;
+
+    private boolean imeEnabled;
```
mstr2: The following fields are not read, but only assigned to in this class:
* `capabilities`
* `x`
* `y`
* `imeEnabled`

The field `parentPtr` is neither assigned nor read.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessView.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    @Override
+    protected void _uploadPixels(long ptr, Pixels pixels) {
+        HeadlessWindow window = (HeadlessWindow)this.getWindow();
+        this.pixels = pixels;
+        window.blit(pixels);
+    }
+
+    Pixels getPixels() {
```
mstr2: This method is never used.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/NestedRunnableProcessor.java
```diff
@@ -0,0 +1,79 @@
...
+    private final LinkedList<RunLoopEntry> activeRunLoops = new LinkedList<>();
+
+    private final BlockingQueue<Runnable> runnableQueue = new LinkedBlockingQueue<>();
+
+    @Override
+    public void run() {
+        newRunLoop();
+    }
+
+    void invokeLater(final Runnable r) {
```
mstr2, line 43: The following methods are never used:
* `invokeLater`
* `runLater`
* `invokeAndWait`
* `stopProcessing`

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindowManager.java
```diff
@@ -0,0 +1,31 @@
...
+        for (Window win : windows) {
+            if (win.isVisible() && (!win.isMinimized())) {
+                HeadlessWindow hw = (HeadlessWindow) win;
+                HeadlessView hv = (HeadlessView) hw.getView();
+                hv.notifyRepaint(hw.getX(), hw.getY(), hw.getWidth(), hw.getHeight());
+            }
+        }
+    }
+
+    private HeadlessWindow getFocusedWindow() {
```
mstr2: This private method is never used.

mstr2 (commented):
My testing shows that the headless platform works as expected.

Some general comments:
1. All files need a copyright header.
2. Very long lines can benefit from a few line breaks.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
johanvos: suggestion is merged

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessApplication.java
johanvos: changed to the same format as com.sun.glass.ui.Application

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
johanvos: done

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
johanvos: fixed

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
johanvos: There was a failing test, so it turns out there is an implicit contract that even if a resize is requested without different w/h parameters, at least the even should still be generated.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
johanvos: the lx/ly coordinates are already expected to be taking into account the window X/Y

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
johanvos: removed

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/NestedRunnableProcessor.java
johanvos: removed

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
johanvos: aligned with com.sun.glass.ui.Window

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
johanvos: formatted

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
johanvos: removed

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessApplication.java
johanvos: done

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessPlatformFactory.java
johanvos: done

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessPlatformFactory.java
johanvos: done

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessTimer.java
johanvos: done

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessView.java
johanvos: Added this -- I agree it looks pointless now, but preparing for follow-up review comments where the parentPtr is prepared.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindowManager.java
johanvos: removed

johanvos:
> My testing shows that the headless platform works as expected.
> 
> Some general comments:
> 
> 1. All files need a copyright header.
> 2. Very long lines can benefit from a few line breaks.

Thanks for the comments, those make the PR much cleaner!
I addressed most of them. There are a few comments I didn't address yet (e.g. about fields that are assigned but never read) because there might be follow-up reviews that require those values to be readable.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessApplication.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    @Override
+    protected double staticScreen_getVideoRefreshPeriod() {
+        return 0.;
+    }
+
+    @Override
+    protected Screen[] staticScreen_getScreens() {
+        final int screenWidth = 1000;
```
Maran23, line 160: Should this be something that can be configured by the developer? E.g. a simple system property with a default value of 1000?

johanvos: This is a good question, and I believe it is part of a broader discussion. There are a number of places (e.g. here) where I use hardcoded, default values which should be fine for most usecases I'm aware of. However, I believe there is definitely value in having more configurable values (e.g. in screen dimensions, but also number of screens, default window sizes, multiclick values etc).
The challenge with this is that we need some way to have developers manipulate these values, for example using system properties (as you state as well). This is then adding a feature that doesn't exist yet, so it would require a deeper discussion on how to make the Headless platform configurable. I'm absolutely +1 on that, but I believe it would be easier if we do that as a follow-up?

andy-goryachev-oracle: I was going to suggest a system property like
```
-Dheadless.screens="1920x1024"
-Dheadless.screens="1920x1024x16"
```
to be able to configure the screen resolution and the bit depth (and maybe a scale as well).

I don't know whether supporting multiple screens makes sense here, but perhaps for testing?

And, if the code cannot parse the property it would default to something reasonable like 1280x768 or slightly larger?

Maran23: > I'm absolutely +1 on that, but I believe it would be easier if we do that as a follow-up?

I agree, makes sense. This way, we can also gradually add properties as (if) needed.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessApplication.java
```diff
@@ -0,0 +1,218 @@
...
+import com.sun.glass.ui.Window;
+import java.io.File;
+import java.nio.ByteBuffer;
+import java.nio.IntBuffer;
+
+public class HeadlessApplication extends Application {
+
+    private final NestedRunnableProcessor processor = new NestedRunnableProcessor();
+    private final HeadlessWindowManager windowManager = new HeadlessWindowManager();
+    private Screen[] screens = null;
```
andy-goryachev-oracle: minor: no need to assign null, also L24

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessApplication.java
```diff
@@ -0,0 +1,218 @@
...
+    protected void runLoop(Runnable launchable) {
+        processor.invokeLater(launchable);
+        Thread eventThread = new Thread(processor);
+        setEventThread(eventThread);
+        eventThread.start();
+    }
+
+    @Override
+    protected void _invokeAndWait(Runnable runnable) {
+        throw new UnsupportedOperationException("Not supported yet.");
```
andy-goryachev-oracle, line 64: is there a follow-up bug for this, or the list of things to do?

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessApplication.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    @Override
+    public Cursor createCursor(int x, int y, Pixels pixels) {
+        throw new UnsupportedOperationException("Not supported yet.");
+    }
+
+    @Override
+    protected void staticCursor_setVisible(boolean visible) {
+        throw new UnsupportedOperationException("Not supported yet.");
```
andy-goryachev-oracle, line 114: should it be a no-op instead?  will cursor be supported at all?

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
```diff
@@ -0,0 +1,398 @@
...
+        int wy = activeWindow.getY();
+        int buttonEvent = MouseEvent.BUTTON_NONE;
+        int mouseEvent = MouseEvent.MOVE;
+        if (mouseState.pressedButtons.size() > 0) {
+            MouseButton button = mouseState.pressedButtons.stream().findFirst().get();
+            buttonEvent = getGlassEventButton(button);
+            mouseEvent = MouseEvent.DRAG;
+        }
+        int modifiers = 0;
+        view.notifyMouse(mouseEvent, buttonEvent, (int)mouseX-wx, (int)mouseY-wy, (int)mouseX, (int)mouseY, modifiers, false, false);
```
andy-goryachev-oracle, line 133: should the coordinates be rounded instead of floor'ed (cast to int)?
(also in other places below)

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
```diff
@@ -0,0 +1,398 @@
...
+        checkWindowFocused();
+
+        final int dff = wheelAmt > 0 ? -1 : 1;
+        HeadlessView view = (HeadlessView) activeWindow.getView();
+
+        int wx = activeWindow.getX();
+        int wy = activeWindow.getY();
+        int repeat = Math.abs(wheelAmt);
+        for (int i = 0; i < repeat; i++) {
+            view.notifyScroll((int) mouseX, (int) mouseY, wx, wy, 0, dff, 0, 0, 0, 0, 0, multiplierX, multiplierY);
```
andy-goryachev-oracle, line 184: are all parameters correct?  `modifiers` and `lines`, specifically?

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
```diff
@@ -0,0 +1,398 @@
...
+
+    @Override
+    public WritableImage getScreenCapture(WritableImage image, double x, double y, double width, double height, boolean scaleToFit) {
+        return super.getScreenCapture(image, x, y, width, height, scaleToFit);
+    }
+
+    @Override
+    public void getScreenCapture(int x, int y, int width, int height, int[] data, boolean scaleToFit) {
+        checkWindowFocused();
+        ((HeadlessWindow)activeWindow).getScreenCapture(x, y, width, height, data, scaleToFit);
```
andy-goryachev-oracle: maybe the `activeWindow` can be declared as `HeadlessWindow` to avoid casting

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
```diff
@@ -0,0 +1,398 @@
...
+    }
+
+    private HeadlessWindow getTopWindow() {
+        List<Window> windows = Window.getWindows().stream()
+                .filter(win -> win.getView() != null)
+                .filter(win -> !win.isClosed())
+                .filter(win -> !win.isMinimized()).toList();
+        if (windows.isEmpty()) return null;
+        if (windows.size() == 1) return (HeadlessWindow)windows.get(0);
+        return (HeadlessWindow)windows.get(windows.size() -1);
```
andy-goryachev-oracle, line 253: call me old fashioned, but would a straightforward reverse `for` loop be better in this case?  no unnecessary memory allocations required.

here and in two methods below?

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessView.java
andy-goryachev-oracle: x, y are read in View::getX/Y

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
```diff
@@ -0,0 +1,369 @@
...
+import java.nio.Buffer;
+import java.nio.ByteBuffer;
+import java.nio.IntBuffer;
+import java.util.concurrent.atomic.AtomicInteger;
+import javafx.scene.paint.Color;
+
+public class HeadlessWindow extends Window {
+
+    private static final AtomicInteger ptrCount = new AtomicInteger(0);
+    private long ptr;
```
andy-goryachev-oracle, line 42: since `ptr` is long, should `ptrCount` be `AtomicLong`?

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
andy-goryachev-oracle: should we either add a comment or clean up (or both) then?

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
```diff
@@ -0,0 +1,369 @@
...
+        int mx = lx;
+        int my = ly;
+        int idx = 1000 * my + mx;
+        int rgba = frameBuffer.asIntBuffer().get(idx);
+        int a = (rgba >> 24) & 0xFF;
+        int r = (rgba >> 16) & 0xFF;
+        int g = (rgba >> 8) & 0xFF;
+        int b = rgba & 0xFF;
+
+        Color color = Color.color(
```
andy-goryachev-oracle: could use `Color::rgb(int,int,int,double)`;

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
```diff
@@ -0,0 +1,369 @@
...
+                a / 255.0
+        );
+        return color;
+    }
+
+    public void getScreenCapture(int x, int y, int width, int height, int[] data, boolean scaleToFit) {
+        for (int i = 0; i < height; i++) {
+            for (int j = 0; j < width; j++) {
+                int idx = i * width + j;
+                int fidx = (y + i) * 1000 + x + j;
```
andy-goryachev-oracle: where does 1000 come from?

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
```diff
@@ -0,0 +1,369 @@
...
+            }
+        }
+    }
+
+    void blit(Pixels pixels) {
+        int pW = pixels.getWidth();
+        int pH = pixels.getHeight();
+        int offsetX = this.getX();
+        int offsetY = this.getY();
+        int stride = 1000;
```
andy-goryachev-oracle: same magic 1000 here

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
```diff
@@ -0,0 +1,369 @@
...
+            for (int j = 0; j < pixels.getWidth(); j++) {
+                int idx = rowIdx * stride + offsetX + j;
+                int val = intBuffer.get(i * pixels.getWidth() + j);
+                frameBuffer.asIntBuffer().put(idx, val);
+            }
+        }
+    }
+
+    void clearRect(int x0, int w0, int y0, int h0) {
+        int stride = 1000;
```
andy-goryachev-oracle: and here...

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/NestedRunnableProcessor.java
```diff
@@ -0,0 +1,78 @@
...
+            try {
+                r.run();
+            } finally {
+                latch.countDown();
+            }
+        });
+        try {
+            latch.await();
+        } catch (InterruptedException e) {
+            e.printStackTrace();
```
andy-goryachev-oracle: should we print here or send to `Application.reportException(e)` ?

andy-goryachev-oracle:
Added a little code to the monkey tester to take a snapshot of the scene.
Here is the result launching it on the headless platform on macOS 15.5:



Good job!

Question: While the screen size is limited in this version to 1000x1000, the size of the snapshot exceeds that.  Is that expected?

andy-goryachev-oracle:
WebView works:

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
johanvos: I cleaned it up. Typically, I'm +1 on adding more comments in the code, but not everyone likes that. In this case, it would create a precedent where we would then add tons of comments mainly to document what the tests are doing (or expecting).

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
johanvos: I had those since I think they might be useful "later", when we add more configuration options, or in case there are tests that rely on them. But I believe it is better indeed to remove them, and re-introduce them when needed.

johanvos: ptr and visible are used, I removed the others.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
johanvos: fixed

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
johanvos: changed return type to void

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessApplication.java
johanvos: removed and changed createCursor method

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessView.java
johanvos: removed capabilities and imeEnabled

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/NestedRunnableProcessor.java
johanvos: invokeLater is used from HeadlessApplication
Others are removed now.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessApplication.java
johanvos: removed.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessApplication.java
johanvos: I think it's best to keep this here, because in case there is somehow an invocation of this method, we want to be notified about it, and then come up with the most appropriate way to implement it.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessApplication.java
johanvos: There is a Headless Cursor, but the setVisible method doesn't seem to be used in any of the tests currently.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
johanvos: good idea, done

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
johanvos: I'm not a fan of using Streams if it's not really needed, but in this case I think there might be benefits in eliminating the non-candidates in 1 statement -- I might be wrong though, didn't look yet at how this is compiled.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
johanvos: it's more elegant indeed, so I changed it.
It's not really needed though, since contrary to native implementations, the pointer we return here is not a pointer to some memory, hence we don't need 64 bits. The method declaration for _createWindow requires a long though, so we need to return a long.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
johanvos: done

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
johanvos: it's the stride used in the ByteBuffer holding the frame. I've moved it to a final int and removed all hard occurrences of 1000 .

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
johanvos: it's the stride used in the ByteBuffer holding the frame. I've moved it to a final int and removed all hard occurrences of 1000 .

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
johanvos: it's the stride used in the ByteBuffer holding the frame. I've moved it to a final int and removed all hard occurrences of 1000 .

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
```diff
@@ -0,0 +1,398 @@
...
+            if (view == null) {
+                return;
+            }
+        }
+        int wx = activeWindow.getX();
+        int wy = activeWindow.getY();
+        int modifiers = getModifiers(buttons);
+        view.notifyMouse(MouseEvent.DOWN, getGlassEventButton(buttons), (int)mouseX-wx, (int)mouseY-wy, (int)mouseX, (int)mouseY, modifiers, true, true);
+        int buttonCode = getGlassEventButton(buttons);
+        view.notifyMouse(MouseEvent.DOWN, buttonCode, (int)mouseX-wx, (int)mouseY-wy, (int)mouseX, (int)mouseY, modifiers, true, true);
```
jperedadnr: This code (`view.notifyMouse(...`) is duplicated, and there are two notifications of mouse down while there should be just one.

The first line can be removed.

johanvos: Right, good catch!
I removed that first line.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/NestedRunnableProcessor.java
johanvos: the method has been removed as it wasn't used.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
```diff
@@ -284,27 +286,22 @@ private void notifyResizeAndMove(int x, int y, int width, int height) {
...
         int g = (rgba >> 8) & 0xFF;
         int b = rgba & 0xFF;
 
-        Color color = Color.color(
-                r / 255.0,
-                g / 255.0,
-                b / 255.0,
-                a / 255.0
-        );
+        Color color = Color.rgb(r, g, b, a);
```
jperedadnr: `Color::rgb` requires opacity between 0 and 1, therefore, you need:

```
Color color = Color.rgb(r, g, b, a / 255.0);
```
Minor: you can inline the variable.

johanvos: Good catch, thanks!
Fixed.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessApplication.java
```diff
@@ -0,0 +1,216 @@
+package com.sun.glass.ui.headless;
```
kevinrushforth, line 25: All of the new files need a standard copyright header.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessApplication.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    @Override
+    protected Screen[] staticScreen_getScreens() {
+        final int screenWidth = 1000;
+        final int screenHeight = 1000;
+        if (this.screens == null) {
+            float scaleX = 1.f;
+            float scaleY = 1.f;
+            String scale = System.getProperty("glass.gtk.uiScale");
```
kevinrushforth: "gtk" ?

kevinrushforth (commented):
I left one inline comment about the need for copyright headers.

I did a quick first pass, and this looks like a reasonable replacement for the Headless monocle platform to support running headless tests. It has the advantage of not needing to manually add monocle, which is not shipped on desktop platforms.

This is effectively an experimental feature used for testing. Since there is no API, and it is only enabled by setting a (as yet undocumented) system property, it seems OK that it isn't complete and doesn't have an associated CSR.

Longer term, we will want to think about how this might evolve. The first question is: Does this have any use case beyond testing? AWT has a headless mode that developers can use for server-side rendering. We have discussed the possibility of adding a similar feature for JavaFX, although it is less clear how it would be used, since we don't have image APIs that would be expected for such applications.

Even if this remains a testing feature, how should it eventually be configured? Using system properties? Some sort of config file? Something else?

How should this interact with the AWT headless toolkit? Or does it?

This requires the prism SW pipeline. How / where should be document this restriction?

johanvos:
> The first question is: Does this have any use case beyond testing?

Good question, and the answer is "Yes, absolutely". One of the main drivers for this was a project where PDF prints need to be generated server-side, starting from `Node.snapShot(... WritableImage)`.

hjohn:
> > The first question is: Does this have any use case beyond testing?
> 
> Good question, and the answer is "Yes, absolutely". One of the main drivers for this was a project where PDF prints need to be generated server-side, starting from `Node.snapShot(... WritableImage)`.

@johanvos 

Did you happen to see my proposal to use the software renderer to provide a Canvas#GraphicsContext like API for writable images?  Snapshot is a nice feature, but it feels like a rather roundabout way to first use the GPU to draw things, then copy it back to main memory for pixel manipulation or say PDF creation...

See here: https://www.mail-archive.com/openjfx-dev@openjdk.org/msg20618.html

I'm curious what you think of the proposal (and I'm curious if the headless platform would need to support it or interfere with it :))

craigraw:
Hi @craigraw, thanks for making a comment in an OpenJDK project!

All comments and discussions in the OpenJDK Community must be made available under the OpenJDK [Terms of Use](https://openjdk.java.net/legal/tou/). If you already are an OpenJDK [Author](https://openjdk.java.net/bylaws#author), [Committer](https://openjdk.java.net/bylaws#committer) or [Reviewer](https://openjdk.java.net/bylaws#reviewer), please click [here](https://bugs.openjdk.java.net/secure/CreateIssue.jspa?pid=11300&issuetype=1) to open a new issue so that we can record that fact. Please Use "Add GitHub user craigraw" for the summary.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
johanvos: From a logical point, I would say yes. However, if you look at the Robot implementations for e.g. GTK, there are casts to int as well. So in order to be consistent, I think we should cast to int (and floor, indeed).

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
johanvos: It is a bit complex to easily determine if those parameters have relevance at all. There are no failing tests due to this, so I see this as a good opportunity to add tests in case issues (unexpected behavior) is detected.

johanvos:
> Question: While the screen size is limited in this version to 1000x1000, the size of the snapshot exceeds that. Is that expected?

Good question.
Do you refer to the Node.snapshot()?
The definition of "screen size" is a bit vague on headless, also in relation to window size and view size. I think that a snapshot can in theory be bigger than the screen size, in case multiple screens are used (something we currently don't do).
I believe we can use more input from concrete use cases before we nail down constraints and boundary conditions like this.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessApplication.java
johanvos: whoops. I took a similar approach here as the one used in GtkApplication, but it is not in line with my own suggestion to keep configuration out of this PR.
I removed this property check, as I believe configurations (via properties or not) should be discussed globally, and having one property already being used in the code, would make that discussion harder.

johanvos:
> > > The first question is: Does this have any use case beyond testing?
> > 
> > 
> > Good question, and the answer is "Yes, absolutely". One of the main drivers for this was a project where PDF prints need to be generated server-side, starting from `Node.snapShot(... WritableImage)`.
> 
> @johanvos
> 
> Did you happen to see my proposal to use the software renderer to provide a Canvas#GraphicsContext like API for writable images? Snapshot is a nice feature, but it feels like a rather roundabout way to first use the GPU to draw things, then copy it back to main memory for pixel manipulation or say PDF creation...
> 
> See here: https://www.mail-archive.com/openjfx-dev@openjdk.org/msg20618.html
> 
> I'm curious what you think of the proposal (and I'm curious if the headless platform would need to support it or interfere with it :))

I see the benefits from your proposal, and I believe it makes sense. I don't think it interferes with the headless platform, since it is more in the prism-side of things, but there is always some overlap. The Robot.getScreenCapture() is one of those methods where the implementation would depend on it.

johanvos:
> Even if this remains a testing feature, how should it eventually be configured? Using system properties? Some sort of config file? Something else?

This was discussed in a number of comments as well, and the suggested approach is to leave configuration for follow-up PR's. There are a number of options (e.g. system properties, but also an API), and some of them would require a CSR and/or more formal work. Getting more input from usecases would be helpful here.

> How should this interact with the AWT headless toolkit? Or does it?

There is no interaction yet. I believe there might be value in more direct interactions with AWT though, e.g. with the AWT Print API.
 
> This requires the prism SW pipeline. How / where should be document this restriction?

Surprisingly (or not), most of the tests I've ran (e.g. the OpenJFX system tests) do not require the prism SW pipeline. But tests where one wants to check what pixels are rendered on screen require prism SW indeed. 
It's a good question where this needs to be documented. Since there is no new API, it can't be in the new API docs. One of the options is to add it in a property overview document, e.g. like the one from Abhinay at https://github.com/abhinayagarwal/abhinayagarwal.github.io/blob/master/_posts/2022-10-03-OpenJFX%20flags.markdown . Since the headless platform is enabled by setting a System Property, it could be useful to add the associated options along that system property.

andy-goryachev-oracle:
> Do you refer to the Node.snapshot()?

In my case, it was Scene::snapshot(), but basically the question is where the screen size has any effect?  Robot?  

It does not seem to impact Scene::snapshot.

(I've used the standalone Monkey Tester with an auto snapshot added, see https://github.com/andy-goryachev-oracle/MonkeyTest/tree/headless )

kevinrushforth:
> > Do you refer to the Node.snapshot()?
> 
> In my case, it was Scene::snapshot(), but basically the question is where the screen size has any effect? Robot?
> 
> It does not seem to impact Scene::snapshot.

The screen size is not relevant for snapshot (intentionally so). It's a good question as to where it would make a difference.

kevinrushforth:
I'll review the code soon. I ran a full set of tests, excluding the robot tests, and see two failures (below), which you mentioned are expected. Presumably these are the same two failures you see?

It would probably be best to skip these two tests when running the Headless glass platform. This could either be done as part of this PR or in a follow-up "Skip failing tests on Headless glass platform" bug.

1. MacPasteboardTest

```
MacPasteboardTest > testValidLocalImageURLMacPasteboard() FAILED
    java.lang.NullPointerException: Cannot invoke "com.sun.glass.ui.mac.MacPasteboardShim.pushMacPasteboard(java.util.HashMap)" because "test.com.sun.glass.ui.mac.MacPasteboardTest.macPasteboardShim" is null
        at test.com.sun.glass.ui.mac.MacPasteboardTest.lambda$testValidLocalImageURLMacPasteboard$0(MacPasteboardTest.java:73)

MacPasteboardTest > testDataBase64ImageMacPasteboard() FAILED
    java.lang.NullPointerException: Cannot invoke "com.sun.glass.ui.mac.MacPasteboardShim.pushMacPasteboard(java.util.HashMap)" because "test.com.sun.glass.ui.mac.MacPasteboardTest.macPasteboardShim" is null
        at test.com.sun.glass.ui.mac.MacPasteboardTest.lambda$testDataBase64ImageMacPasteboard$0(MacPasteboardTest.java:89)

MacPasteboardTest > testNotAnImageURLMacPasteboard() FAILED
    java.lang.NullPointerException: Cannot invoke "com.sun.glass.ui.mac.MacPasteboardShim.pushMacPasteboard(java.util.HashMap)" because "test.com.sun.glass.ui.mac.MacPasteboardTest.macPasteboardShim" is null
        at test.com.sun.glass.ui.mac.MacPasteboardTest.lambda$testNotAnImageURLMacPasteboard$0(MacPasteboardTest.java:100)
```

The tests in this class assume we are running the Mac glass platform, so they should all be skipped when running the Headless glass platform. One way to do this would be to add a `PlatformUtil::isHeadless` method and change all of the `assumeTrue(PlatformUtil.isMac())` calls to `assumeTrue(PlatformUtil.isMac() && !PlatformUtil.isHeadless())`. The assumeTrue wouldn't be annotated with a bug ID, since this expresses a fundamental requirement of the test, not a bug that needs to be fixed.


2. ClipboardTest::testCopyUTF8String

```
ClipboardTest > testCopyUTF8String() FAILED
    org.opentest4j.AssertionFailedError: expected: < Jump to: 😃💁 Jump> but was: <HEADLESS_TEST>
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
        at app//org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1145)
        at app//test.javafx.scene.input.ClipboardTest.testCopyUTF8String(ClipboardTest.java:94)
```

This looks like a bug in the Headless glass platform code (not the test), so a bug tracking this headless clipboard failure should be filed. Then add `assumeTrue(!PlatformUtil.isHeadless())` along with a comment above the assumeTrue with the bug ID of the headless clipboard tracking bug.

Separately, another bug should be filed for the failing robot tests (I don't have a list, since I didn't run it with `-PUSE_ROBOT=true` for this test run).

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessPlatformFactory.java
```diff
@@ -0,0 +1,152 @@
...
+
+    @Override
+    public Application createApplication() {
+        return new HeadlessApplication();
+    }
+
+    @Override
+    public MenuBarDelegate createMenuBarDelegate(MenuBar menubar) {
+        throw new UnsupportedOperationException("Not supported yet.");
+    }
```
kevinrushforth, line 50: Suggestion: file one or more follow-on enhancements to add missing features (here and elsewhere).

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindowManager.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class HeadlessWindowManager {
+
+    void repaintAll() {
+        List<Window> windows = Window.getWindows().stream()
+                .filter(win -> win.getView() != null)
+                .filter(win -> !win.isClosed())
+                .filter(win -> win.isVisible()).toList();
+        for (Window win : windows) {
+            if (win.isVisible() && (!win.isMinimized())) {
```
kevinrushforth, line 38: Minor: since you are already using streams, you could use a `.forEach` as the terminal operator rather than collecting it into a list to iterate over again. Separately, you could also convert the `!win.isMinimized()` test into a filter operation (note that the test for `win.isVisible()` is redundant since you already filter out non-visible windows).

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
```diff
@@ -0,0 +1,421 @@
...
+    private final SpecialKeys specialKeys = new SpecialKeys();
+    private final MouseState mouseState = new MouseState();
+    private final char[] NO_CHAR = { };
+
+    public HeadlessRobot(HeadlessApplication application) {
+        this.application = application;
+    }
+
+    void windowAdded(HeadlessWindow window) {
+        if (this.activeWindow == null) activeWindow = window;
```
kevinrushforth, line 58: I think this means that if an app creates two windows, then closes the second, robot won't be able to capture anything from the first. Or will this method be called again on the first window when the second closes?

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessRobot.java
kevinrushforth: I think this is fine as is. I do note that the following is redundant:

```
        if (windows.size() == 1) return (HeadlessWindow)windows.get(0);
```

The general case will handle this case (and should be just as efficient).

modules/javafx.graphics/src/main/java/com/sun/glass/ui/headless/HeadlessWindow.java
```diff
@@ -24,20 +47,14 @@ public class HeadlessWindow extends Window {
...
-    private float bg_r, bg_g, bg_b;
-    private float alpha;
-    private Pixels icon;
-    private Cursor cursor;
+
     private final ByteBuffer frameBuffer;
     private HeadlessView currentView;
     private HeadlessRobot robot;
 
+    private final int stride = 1000;
```
andy-goryachev-oracle, line 56: I see you've created [JDK-8364201](https://bugs.openjdk.org/browse/JDK-8364201) for the configuration options, thank you.

andy-goryachev-oracle (approved):

jperedadnr (approved):

johanvos:
@kevinrushforth I made one more final change, can you re-approve if you agree?

kevinrushforth:
Oh, I thought I already had.

PR #1747, opened 2025-03-28
8319555: [TestBug] Utility for creating instruction window for manual tests

andy-goryachev-oracle:
Provides the base class for manual tests which displays the test instructions, the UI under test, and the Pass/Fail buttons.

Uses `EmojiTest` to illustrate the use of the new class.

Example:

```
public class ManualTestExample extends ManualTestWindow {
    public ManualTestExample() {
        super(
            "Manual Test Example",
            """
            Instructions:
            1. you will see a button named "Test"
            2. press the button
            3. verify that the button can be pressed""",
            400, 250
        );
     }

     public static void main(String[] args) throws Exception {
         launch(args);
     }

     @Override
     protected Node createContent() {
         return new Button("Test");
     }
}
```

Screenshot 2025-03-28 at 11 50 16

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [ ] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Warning
&nbsp;⚠️ Patch contains a binary file (tests/manual/util/doc/ManualTestWindow.png)

### Issue
 * [JDK-8319555](https://bugs.openjdk.org/browse/JDK-8319555): [TestBug] Utility for creating instruction window for manual tests (**Enhancement** - P4)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1747/head:pull/1747` \
`$ git checkout pull/1747`

Update a local copy of the PR: \
`$ git checkout pull/1747` \
`$ git pull https://git.openjdk.org/jfx.git pull/1747/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 1747`

View PR using the GUI difftool: \
`$ git pr show -t 1747`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jfx/pull/1747.diff">https://git.openjdk.org/jfx/pull/1747.diff</a>

</details>
<details><summary>Using Webrev</summary>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1747#issuecomment-2762243970)
</details>

PR #1837, opened 2025-06-30
8360886: Cmd + plus shortcut does not work reliably.

beldenfox:
The Mac platform code figures out where characters are on the keyboard as the user types. The character table is updated on every key press by calling a registerKeyEvent: in GlassApplication. This character table is used to resolve KeyCharacterCombination accelerators like Cmd + "+".

On a US English layout when the user types Cmd + "+" on the main keyboard they're actually typing Cmd + "=". There's special handling in macOS for this combination that can cause two NSEvents to be sent and so there's special handling in the Glass code so we don't process both events. When this special case is invoked registerKeyEvent: isn't being called to update the character table. This bug was introduced when code was consolidated in PR #1528.

The fix is a simple one-liner. I've added a test for this in the KeyboardTest.java program. It's an isolated test because it requires the Robot to send events to hold down a modifier while a character key is pressed. I also updated some obsolete comments and tweaked it to test all KeyCharacterCombinations since they should now work reliably on all platforms.

### DIFF
--- a/tests/manual/events/KeyboardTest.java
+++ b/tests/manual/events/KeyboardTest.java
@@ -28,6 +28,7 @@ import java.util.Locale;
 import java.util.Optional;
 import java.util.Timer;
 import java.util.TimerTask;
+import java.util.concurrent.atomic.AtomicReference;
 import javafx.application.Application;
 import javafx.application.Platform;
 import javafx.event.EventHandler;
@@ -38,6 +39,7 @@ import javafx.scene.control.ChoiceBox;
 import javafx.scene.control.TextArea;
 import javafx.scene.input.KeyCharacterCombination;
 import javafx.scene.input.KeyCode;
+import javafx.scene.input.KeyCombination;
 import javafx.scene.input.KeyEvent;
 import javafx.scene.layout.HBox;
 import javafx.scene.layout.Priority;
@@ -530,10 +532,9 @@ public class KeyboardTest extends Application {
     }
 
     /*
-     * KeyCharacterCombinations should really work on the numeric keypad but
-     * currently don't on Windows and Linux. The tests can exclude combinations
-     * entirely, exclude just the numeric keypad, or cover both the main
-     * keyboard and the keypad.
+     * A test run can exclude KeyCharacterCombinations entirely, process only
+     * those on the main keyboard (excluding the numeric keypad), or cover
+     * the entire layout.
      */
     private enum CombinationScope {
         NONE("without combinations"),
@@ -622,8 +623,6 @@ public class KeyboardTest extends Application {
 
         private void start(Runnable atEnd) {
             runAtEnd = atEnd;
-
-            log.clear();
 
             Optional<Boolean> capsLockOn = Platform.isKeyLocked(KeyCode.CAPS);
             Optional<Boolean> numLockOn = Platform.isKeyLocked(KeyCode.NUM_LOCK);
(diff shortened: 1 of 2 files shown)


tests/manual/events/KeyboardTest.java
```diff
@@ -808,6 +807,105 @@ private void cleanup() {
         }
     }
 
+    // Here we try to find one case where a shortcut references a symbol that
+    // is normally generated using Shift. In particular we're testing whether
+    // Cmd + "+" works as expected on the US English layout on the Mac. The
+    // OS special-cases this combination and so does JavaFX. In any case
+    // Shortcut + "+" is a very common shortcut and deserves a test.
+    private void testShiftedShortcut(Layout layout, Node focusNode, Logging log) {
```
andy-goryachev-oracle: Question: is it possible to convert this manual test into an automated headful one?

beldenfox: I could write a weaker system test. The problem sequence is CMD + EQUALS. There are two possible outcomes depending on the keyboard layout; either the EQUALS yields nothing or it generates a PRESSED event that matches against KeyCharacterCombination("=", KeyCode.SHORTCUT_DOWN). It's not quite the same test but it would have caught this bug. I can verify that it would work on all the keyboard layouts that Apple ships with macOS 15.

I'm less sure if the test would work cross-platform. It should but I don't have a good way of enumerating and testing all the layouts on Windows and Linux.

andy-goryachev-oracle: I was thinking more in terms of exhaustive checking of all the available keys (so we won't encounter another scenario when a bug is introduced similar to #1528, if that is possible).

I know we can't really control the keyboard layout, so perhaps what I am asking may not be possible since these tests might fail on other layouts.

So let me ask
- is the problem limited to cmd+= on the US keyboard?
- could we have a different behavior with a different kb layout (German? French?)
- what happens when the user attaches an external USB or bluetooth keyboard?

beldenfox: The issue here is limited to cmd+=  and cmd+period. These receive special treatment by macOS. To counter-act this they also receive special treatment in the Glass code.

The macOS special handling of cmd+= only applies to certain keyboard layouts (for example, English but not German). But that's not true for Glass, the work-around in glass applies to all layouts.

I will look into writing up a system test. I think I can put one together that tests more keys and works on all layouts. It would be nice not to rely solely on this manual test.

External USB or bluetooth keyboards shouldn't make a difference.

(In case you're curious, Apple wants to make sure you can invoke Cmd + "+" by holding down Cmd and pressing any key that has a "+" symbol printed on it. On a US English layout that means Cmd + "=" should also work. So they send us Cmd + "=" and if that doesn't trigger a shortcut they send us Cmd + "+". Glass tries to ensure that that second event doesn't turn into another KeyEvent. I think this only happens with keys where "+" is Shift + "=" which is why it happens on US English but not German.)

andy-goryachev-oracle: thank you for the detailed explanations!

andy-goryachev-oracle:
SKARA warns about trailing period in the title.  I've corrected the JIRA ticket, please fix the PR.

tests/manual/events/KeyboardTest.java
```diff
@@ -38,6 +39,7 @@
 import javafx.scene.control.TextArea;
 import javafx.scene.input.KeyCharacterCombination;
 import javafx.scene.input.KeyCode;
+import javafx.scene.input.KeyCombination;
```
andy-goryachev-oracle: I wonder if it is possible to add a panel to the test stage with explanations that are currently in the class-level javadoc?  Specifically, the fact that you need to configure a particular layout.

I had some work done to simplify the boilerplate in #1747, though it did not make it yet.

beldenfox:
> SKARA warns about trailing period in the title. I've corrected the JIRA ticket, please fix the PR.

Title updated. Thanks for fixing the ticket.

beldenfox:
Well, now I’ve gone and re-written this PR.

As stated earlier sometimes a single key press can lead to multiple calls to performKeyEquivalent. During testing I discovered that the way Glass checks for this doesn’t always work. Currently it tries to guess which key event might lead to a second one but there are cases where it guesses wrong (e.g. the Hebrew layout). I’ve decided to take a different approach, detecting and discarding the second event. The logic is:

- Is the new key event not the current event (theEvent != NSApp.currentEvent)?
- Are we still working on an earlier event (NSApp.currentEvent == lastKeyEvent)?

If both these conditions are true we can ignore the new key event.

I’ve reverted the changes to the manual test and added a system test. It covers a smattering of carefully selected keys including the combinations involved in this bug. Currently it’s restricted to macOS and Latin keyboards because the Hebrew keyboard on macOS is crazy. I will expand this to other platforms when I have time to do the testing. I’ve run this test on macOS using US English, French, German, Spanish, Hebrew, Devanagari, and a few other layouts.

(About that Hebrew layout. When you hold down Command all the keys start behaving like a US English layout, including the punctuation keys. So the key you press to type a period is entirely different from the key that generates Cmd+period.)

mstr2 (approved):
Looks good.

beldenfox:
Heads up, I just submitted a PR (#1848) that fixes this bug in an entirely different way. The fix in this PR is much simpler and safer (I would not be afraid to back port it) and so I think it should go forward even if it turns out to be an interim fix. In any case it might take a while for #1848 to get vetted and there's the non-zero chance it will get rejected. But if anyone reviews both you'll see the same area of code affected in GlassView3D.m (the performKeyEquivalent: handling).

andy-goryachev-oracle (approved):
Sorry, I've dropped the ball with the re-review.  lgtm.

kevinrushforth:
@beldenfox In case you missed it, this is ready for integration.

beldenfox:
> @beldenfox In case you missed it, this is ready for integration.

Thanks. Lately I haven't been getting e-mail notifications when my pull requests transition to the ready state.

PR #1838, opened 2025-06-30
8341286: TreeView: visual artifacts when setting new root with null value

Ziad-Mid:
The issue is that when we use null value for root item in a TreeView and then expand the root or change it dynamically there are phantom disclosure nodes that appear incorrectly. 
The issue was caused by this line `if (getSkinnable().isEmpty()) return;` in `TreeCellSkin#updateDisclosureNode()` as it prevents the method from running unless the cell's value is not empty and it caused the creation of disclosure nodes visually without any logic behind.

---------
### Progress

### Issue
 * [JDK-8341286](https://bugs.openjdk.org/browse/JDK-8341286): TreeView: visual artifacts when setting new root with null value (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1838/head:pull/1838` \
`$ git checkout pull/1838`

`$ git checkout pull/1838` \
`$ git pull https://git.openjdk.org/jfx.git pull/1838/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1838`

`$ git pr show -t 1838`

</details>

<a href="https://git.openjdk.org/jfx/pull/1838.diff">https://git.openjdk.org/jfx/pull/1838.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1838#issuecomment-3128242088)
</details>


andy-goryachev-oracle:
Hmm, I am not sure this is the right solution: remember, the tree works just fine if you change the root in the form of `TreeItem<>(new Entry(null))`, but breaks when you set `TreeItem<>(null)`.
 
This means one thing: the logic is there, but the `null` content of TreeItem messes it up.
 
Can you try one thing: run two instances of the app in the debugger side by side and step through the relevant parts of the code, until you find the difference?  I suspect some condition breaks when it encounters null.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TreeCellSkin.java
```diff
@@ -359,7 +359,8 @@ private void updateTreeItem() {
     }
 
     private void updateDisclosureNode() {
-        if (getSkinnable().isEmpty()) return;
+//        JDK-8341281 : Root TreeItem with null value breaks TreeTableView
+//        if (getSkinnable().isEmpty()) return;
```
andy-goryachev-oracle: I think you've got it!
I can't reproduce the issue with your fix - both the reproducer and the monkey tester work correctly.

You don't need to keep the commented out code, please remove it.

Ziad-Mid: Yes this is what caused the issue and the disclosure nodes to appear randomly
I will remove it

Ziad-Mid: Yes this is what caused the issue and the disclosure nodes to appear randomly
I will remove it

kevinrushforth:
@Ziad-Mid Can you add an evaluation to this PR (you can add it to the Description)? This may indeed be the right fix, but a good evaluation should explain the problem and why this is the right fix.

PR #1839, opened 2025-07-02
8357714: AudioClip.play crash on macOS when loading resource from jar

johanvos:
After JDK-8287822 (https://bugs.openjdk.org/browse/JDK-8287822), mpeg file content is no longer played via GSTPlatform but via OSXPlatform. 
We need to correctly handle data in case the source is a file inside a jar, in which case CJavaInputStreamCallbacks is used.

The 2 changes I made are:
1. Use ConnectionHolder instead of Locator for CJavaInputStreamCallbacks in case a jar resource is used.
2. In case of random access datasource, use readBlock instead of readNextBlock in AVMediaPlayer

### DIFF
--- a/modules/javafx.media/src/main/native/jfxmedia/platform/osx/OSXMediaPlayer.mm
+++ b/modules/javafx.media/src/main/native/jfxmedia/platform/osx/OSXMediaPlayer.mm
@@ -327,6 +327,7 @@ JNIEXPORT void JNICALL Java_com_sun_media_jfxmediaimpl_platform_osx_OSXMediaPlay
     if ([scheme caseInsensitiveCompare:@"jar"] == NSOrderedSame ||
         [scheme caseInsensitiveCompare:@"jrt"] == NSOrderedSame) {
         CJavaInputStreamCallbacks *callbacks = new (nothrow) CJavaInputStreamCallbacks();
+        jobject jConnectionHolder = CLocator::CreateConnectionHolder(env, jLocator);
         if (callbacks == NULL) {
             [mediaURL release];
             LOGGER_WARNMSG("OSXMediaPlayer: Unable to create CJavaInputStreamCallbacks\n");
@@ -335,7 +336,7 @@ JNIEXPORT void JNICALL Java_com_sun_media_jfxmediaimpl_platform_osx_OSXMediaPlay
             return;
         }
 
-        if (!callbacks->Init(env, jLocator)) {
+        if (!callbacks->Init(env, jConnectionHolder)) {
             [mediaURL release];
             delete callbacks;
             LOGGER_WARNMSG("OSXMediaPlayer: callbacks->Init() failed\n");
--- a/modules/javafx.media/src/main/native/jfxmedia/platform/osx/avf/AVFMediaPlayer.mm
+++ b/modules/javafx.media/src/main/native/jfxmedia/platform/osx/avf/AVFMediaPlayer.mm
@@ -803,8 +803,16 @@ static CVReturn displayLinkCallback(CVDisplayLinkRef displayLink,
         }
 
         NSMutableData* readData = nil;
+        bool isRandomAccess = locatorStream->GetCallbacks()->IsRandomAccess();
+        int64_t pos = 0;
+        int size = 65536;
         while (requestedLength > 0) {
-            unsigned int blockSize = locatorStream->GetCallbacks()->ReadNextBlock();
+            unsigned int blockSize = -1;
+            if (isRandomAccess) {
+                blockSize = locatorStream->GetCallbacks()->ReadBlock(pos, size);
+            } else {
+                blockSize = locatorStream->GetCallbacks()->ReadNextBlock();
+            }
             if (blockSize <= 0) {
                 break;
             }
@@ -818,6 +826,7 @@ static CVReturn displayLinkCallback(CVDisplayLinkRef displayLink,
             [loadingRequest.dataRequest respondWithData:readData];
 
             requestedLength -= readSize;
+            pos += readSize;
         }
 
         [loadingRequest finishLoading];


johanvos:
@sashamatveev can you review this?

modules/javafx.media/src/main/native/jfxmedia/platform/osx/avf/AVFMediaPlayer.mm
```diff
@@ -803,8 +803,16 @@ - (BOOL)resourceLoader:(AVAssetResourceLoader *)resourceLoader
...
 
         NSMutableData* readData = nil;
+        bool isRandomAccess = locatorStream->GetCallbacks()->IsRandomAccess();
+        int64_t pos = 0;
+        int size = 65536;
         while (requestedLength > 0) {
-            unsigned int blockSize = locatorStream->GetCallbacks()->ReadNextBlock();
+            unsigned int blockSize = -1;
+            if (isRandomAccess) {
+                blockSize = locatorStream->GetCallbacks()->ReadBlock(pos, size);
```
andy-goryachev-oracle: would it make sense to declare a local variable for `locatorStream->GetCallbacks()` ?
or do we assume the compiler is smart enough to do that?

andy-goryachev-oracle:
Would it be possible to come up with a system test?

modules/javafx.media/src/main/native/jfxmedia/platform/osx/OSXMediaPlayer.mm
```diff
@@ -327,6 +327,7 @@ - (void) stop
     if ([scheme caseInsensitiveCompare:@"jar"] == NSOrderedSame ||
         [scheme caseInsensitiveCompare:@"jrt"] == NSOrderedSame) {
         CJavaInputStreamCallbacks *callbacks = new (nothrow) CJavaInputStreamCallbacks();
+        jobject jConnectionHolder = CLocator::CreateConnectionHolder(env, jLocator);
         if (callbacks == NULL) {
```
sashamatveev: Missing `NULL` check for  `jConnectionHolder`. `if (callbacks == NULL || jConnectionHolder == NULL)`

If `callbacks` is not `NULL` it needs to be deleted.

modules/javafx.media/src/main/native/jfxmedia/platform/osx/avf/AVFMediaPlayer.mm
```diff
@@ -803,8 +803,16 @@ - (BOOL)resourceLoader:(AVAssetResourceLoader *)resourceLoader
         }
 
         NSMutableData* readData = nil;
+        bool isRandomAccess = locatorStream->GetCallbacks()->IsRandomAccess();
```
sashamatveev, line 802: `URIConnectionHolder` always return `false` for "jar" or "jrt" protocol. Why it was added?

modules/javafx.media/src/main/native/jfxmedia/platform/osx/avf/AVFMediaPlayer.mm
sashamatveev: Not sure, but do not see a need from performance point view.

sashamatveev:
Looks good overall.
- Undo changed in `AVFMediaPlayer.mm`. There is no need for it and this code will never execute.
- Add NULL pointer check and deletion of `callbacks` if not NULL.
- Change title for issue to something like `MediaPlayer and AudioClip can crash when playing media files over "jar" and "jrt" protocols via OSXPlatform`.

modules/javafx.media/src/main/native/jfxmedia/platform/osx/avf/AVFMediaPlayer.mm
johanvos: In case of jar, the ConnectionHolder is a `MemoryConnectionHolder` (because Locator already has a cacheEntry) which has true for isRandomAccess. 
The cacheEntry is created by the `NativeMediaAudioClip<init>` which is indirectly called by the AudioClip constructor.

In case `readNextBlock` is used on a MemoryConnectionHolder, it seems all chuncks are asked  and played simultaneously (bug in MemoryConnectionHolder.readNextBlock?). With `readBlock` the order and sequence seem to be ok.

Alternatively, we might try to use the already-created ConnectionHolder instead of creating a new one in OSXMediaPlayer.mm.

sashamatveev: Yes, you right `MemoryConnectionHolder` in case of jar, but only for AudioClip. MediaPlayer does not use it. I tested AudioClip with jar with and without random access. Without random access I only hear some clicking sounds, so looks like it does not work completely. With random access I can only hear audio for like 30 seconds and then it loops, so looks like it is broken as well.

Can you play audio file correctly via AudioClip?

sashamatveev: I figure it out. Seek is ignored for random access, so setting `pos = 0` will always read from beginning of the file. We should read from `dataRequest.requestedOffset`. I am attaching proposed patch which I tested and works fine.

sashamatveev:
[8357714.0.patch](https://github.com/user-attachments/files/21113478/8357714.0.patch)
- Patch which does not do seek with random access and it will read data starting from `dataRequest.requestedOffset`.
- For size `requestedLength` is used. I do not see a need to use anything else. It will be limited to requested length or 1 Megabyte.
- Added `NULL` checks and `callbacks` release.

johanvos:
@sashamatveev Thanks a lot. That all makes sense, and it seems to work indeed with the corrected position/length calculations.

modules/javafx.media/src/main/native/jfxmedia/platform/osx/OSXMediaPlayer.mm
```diff
@@ -327,15 +327,17 @@ - (void) stop
     if ([scheme caseInsensitiveCompare:@"jar"] == NSOrderedSame ||
         [scheme caseInsensitiveCompare:@"jrt"] == NSOrderedSame) {
         CJavaInputStreamCallbacks *callbacks = new (nothrow) CJavaInputStreamCallbacks();
-        if (callbacks == NULL) {
+        jobject jConnectionHolder = CLocator::CreateConnectionHolder(env, jLocator);
+        if (callbacks == NULL  || jConnectionHolder == NULL) {
```
sashamatveev: Extra space after `callbacks == NULL`.

johanvos: fixed

johanvos:
> Would it be possible to come up with a system test?

It would be great to have a system test, but we currently have no framework for testing media in the system tests. It would be good to create this, but I expect it to be lots of work (dealing with native code and different frameworks based on platform), so I think it's better in a separate issue

kevinrushforth:
> > Would it be possible to come up with a system test?
> 
> It would be great to have a system test, but we currently have no framework for testing media in the system tests. It would be good to create this, but I expect it to be lots of work (dealing with native code and different frameworks based on platform), so I think it's better in a separate issue

Correct. We do not have such a framework, and it would be out of scope for this issue.

andy-goryachev-oracle:
Got a reproducer for you: added AudioClip page to the Monkey Tester

Build a jar and run it.  Select AudioClip page, then pressing "Play XX" button would play .wav or .mp3 file from the said jar.

edit: it works!

andy-goryachev-oracle:
Unrelated to this PR: 
- Changing the rate property for mp3 has no effect (but does change the rate for a .wav).  Changing the volume property does work.  I wonder if this should be documented.
- AudioClip properties specify no default values, we may need to address that.

I'll create a separate ticket later.

andy-goryachev-oracle:
With JDK24 crashes on the first try:
```
#  SIGSEGV (0xb) at pc=0x00000001047cf18c, pid=37494, tid=30723
#
# JRE version: OpenJDK Runtime Environment (24.0+36) (build 24+36-3646)
# Java VM: OpenJDK 64-Bit Server VM (24+36-3646, mixed mode, sharing, tiered, compressed oops, compressed class ptrs, g1 gc, bsd-aarch64)
# Problematic frame:
# V  [libjvm.dylib+0x53718c]  jni_IsInstanceOf+0x1a8
```

andy-goryachev-oracle:
My apologies - forgot to build with `-PCOMPILE_MEDIA=true`.

The fixed code plays .mp3 from a jar without crashing.

andy-goryachev-oracle:
Should we test with other audio formats, or everything is piped through the same code path?

sashamatveev:
> Should we test with other audio formats, or everything is piped through the same code path?

I tested with all supported formats.

sashamatveev (approved):

PR #1842, opened 2025-07-11
8361713: JavaFX API docs overview is missing an intro section

kevinrushforth:
This PR adds an introductory section to the JavaFX 24 API docs similar to that of the JDK docs.

### DIFF
--- a/overview.html
+++ b/overview.html
@@ -1,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+  <body>
+    <h1 id="javafx-overview" class="title">JavaFX Software Development Kit<br>
+    Version @JAVADOC_VERSION@ API Specification</h1>
+
+    The JavaFX API is composed of the following groups of modules:
+
+    <ul>
+      <li>Core JavaFX Modules: These APIs are in modules whose names start with <code>javafx</code>. They are part of the core JavaFX platform.</li>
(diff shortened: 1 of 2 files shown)


kevinrushforth:
Here is what the JavaFX API docs index page will look like after this fix:

javafx-docs-overview

kevinrushforth:
Reviewers: @arapte and either @johanvos or @tiainen 

@irisclark Can you also take a look?

/reviewers 2

overview.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+  <body>
+    <h1 id="javafx-overview" class="title">JavaFX Software Development Kit<br>
+    Version @JAVADOC_VERSION@ API Specification</h1>
+
+    The JavaFX API is composed of the following groups of modules:
+
+    <ul>
```
andy-goryachev-oracle: minor: JDK uses `<blockquote>`, `<dt>`, and `<dd>` instead of `<ul>`.  should we do the same?

kevinrushforth: Good idea. Done.

kevinrushforth: I updated the screenshot to show the results.

andy-goryachev-oracle (approved):
looks good, thanks!

johanvos (approved):
+1

PR #1843, opened 2025-07-11
8141391: DnD doesn't work for JFXPanel

prsadhuk:
Issue is when `JFXPanel `is used to embed some JavaFX components and when we drag something from swing or an external application and try to drop on a JavaFX component inside the JFXPanel, it doesn't accept the drop showing "block" sign suggesting DnD is blocked. 

The problem is in both drag and drop side of operations where `acceptDrag `and `acceptDrop `on `DropTargetDragEvent `is not called because `DragEvent constructor` called from `Scene `
[which is called from `SwingDnD `via `EmbeddedSceneDTInterface.handleDragEnter` and `handleDragDrop] 
`
is not storing `state.acceptedTransferMode` when eventType is not `DragEvent.DRAG_DROPPED` or `DragEvent.DRAG_DONE` but Scene creates DragEvent with only `DragEvent.ANY`
 so `handleDragEnter `and `handleDragDrop `returns null as `DragEvent.getAcceptedTransferMode` returns null

Fix is made so that instead of relying on return value of handleDragEnter/handleDragDrop  which seems to be always null for reason mentioned above, we rely on `isDataFlavorSupported `check for the drag and drop as is being done in JDK side of DnD scheme.
Also, to have a proper drop of the data, we get the focus owner from the JFXPanel scene and insert the text in the location pointed to by the mouse cursor.

Regression test is already present in tests/manual/swing/DragDropOntoJavaFXControlInJFXPanelTest.java

### DIFF
--- a/modules/javafx.swing/src/main/java/com/sun/javafx/embed/swing/SwingDnD.java
+++ b/modules/javafx.swing/src/main/java/com/sun/javafx/embed/swing/SwingDnD.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -71,6 +71,8 @@ final public class SwingDnD {
 
     private final DragSource dragSource;
     private final DragSourceListener dragSourceListener;
+
+    private final DropTarget dropTarget;
 
     // swingDragSource and fxDropTarget are used when DnD is initiated from
     // Swing or external process, i.e. this SwingDnD is used as a drop target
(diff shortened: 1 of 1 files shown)


kevinrushforth:

/reviewers 2

modules/javafx.swing/src/main/java/com/sun/javafx/embed/swing/SwingDnD.java
```diff
@@ -194,23 +196,51 @@ public void drop(final DropTargetDropEvent e) {
                             dropActionToTransferMode(e.getDropAction()));
                     try {
                         applyDropResult(lastTransferMode, e);
+                        Transferable transferable = e.getTransferable();
+                        String droppedText = (String) transferable.getTransferData(DataFlavor.stringFlavor);
+                        javafx.embed.swing.JFXPanel jfxPanel = (javafx.embed.swing.JFXPanel)comp;
+
+                        if (jfxPanel.getScene().getFocusOwner() instanceof javafx.scene.control.TextField tf) {
```
andy-goryachev-oracle: This cannot possibly work:

a) does not build
b) what if the drop target is not a TextField, but some other editable text component (a TextArea or a rich text editor of some sort)?

kevinrushforth: Yes, this the wrong approach. The javafx.swing module does not (and should not) depend on javafx.controls. Drag and Drop is a scene graph operation and can be targeted to any node, even nodes that are not Controls.

prsadhuk: a) Yes, couple of module-info file was changed but not pushed in
b) I could use TextInputControl or the parent

Any other suggestion as clearly we cannot use Scene.DragEvent return value?

andy-goryachev-oracle: What code path does it take when pasting from an FX component?  Maybe we can look how it differs and try to redirect to that path, so the content gets pasted correctly.

beldenfox: Am I missing something? I can't find any evidence that a JavaFX text input control processes drag-and-drop in any way or that it ever did (though I could only easily look back to jfx14). Did the manual test referenced here ever work?

modules/javafx.swing/src/main/java/module-info.java
```diff
@@ -36,6 +36,7 @@
 module javafx.swing {
     requires java.datatransfer;
     requires javafx.base;
+    requires javafx.controls;
```
kevinrushforth: This should be reverted. I can think of no reason for the `javafx.swing` module to depend on `javafx.controls`. Drag and drop is not a Controls action. It is a scene graph action.

modules/javafx.swing/src/main/java/com/sun/javafx/embed/swing/SwingDnD.java
kevinrushforth: @beldenfox No, you aren't missing something. The manual test program as written cannot work. There is no automatic drag-and-drop handling for JavaFX controls like TextField.

If I modify the test program to handle drag events in the target TextField, it works as expected. Here are the added lines needed to make it work. I'll attach the modified test program to the bug report.

```
                        textField.setOnDragOver(event -> {
                            if (event.getGestureSource() != textField &&
                                    event.getDragboard().hasString()) {
                                event.acceptTransferModes(TransferMode.COPY_OR_MOVE);
                            }

                            event.consume();
                        });

                        textField.setOnDragDropped(event -> {
                            Dragboard db = event.getDragboard();
                            boolean success = false;
                            if (db.hasString()) {
                                textField.setText(db.getString());
                                success = true;
                            }
                            event.setDropCompleted(success);

                            event.consume();
                        });

                    }
                });
```

kevinrushforth (changes requested):
This fix is not correct. The problem is that the test program doesn't handle DragOver and DragDropped events. I suspect that the bug should be closed as "Not an issue".

prsadhuk:
Ok..This is what was missing as mentioned in the PR description "Scene creates DragEvent with only DragEvent.ANY
so handleDragEnter and handleDragDrop returns null as DragEvent.getAcceptedTransferMode returns null"
wherein acceptedTransferMode was null..
I didn't know we need to set the transfermode on DRAG_OVER (so that DragEvent gets DRAG_OVER eventType instead of DragEvent.ANY) and setDropCompleted from the test itself and I was trying to do in the code along with setText assuming test was correct.
Thanks for the clarification.

I have updated the test

kevinrushforth (approved):
Looks good. I changed the title of the JBS issue to reflect that this is a problem with the test. You will need to change this PR title to match before you integrate.

andy-goryachev-oracle:
Maybe also update the description in this PR and the ticket?

kevinrushforth:
> Maybe also update the description in this PR and the ticket?

I added a note to the JBS Description.

@prsadhuk Can you update the PR description when you fix the title?

prsadhuk:
> > Maybe also update the description in this PR and the ticket?
> 
> I added a note to the JBS Description.
> 
> @prsadhuk Can you update the PR description when you fix the title?

@kevinrushforth  Do you want me to overwrite the entire description and write it's a problem with the test?
I guess when I wrote that description I believed it was a product issue and subsequent discussion revealed it was a test issue....I thought so because when I added those tests in 2019 in [JDK-8211248](https://bugs.openjdk.org/browse/JDK-8211248) it was reviewed and no such case was highlighted that test was wrong at that time..
If I overwrite the entire description, it will suggest it was known from start it was a test issue which is not correct..so I added as an endnote that it was found to be a test issue...is it ok?

kevinrushforth:
@prsadhuk What you did with the description seems OK.

PR #1844, opened 2025-07-12
8358450: Viewport characteristics media features

mstr2:
Implementation of [viewport characteristics media features](https://www.w3.org/TR/mediaqueries-5/#mf-viewport-characteristics):
* `width`
* `height`
* `aspect-ratio`: width / height
* `orientation`: `portrait`, `landscape`
* `display-mode`: `fullscreen`, `standalone` (note: `browser` and `minimal-ui` are not supported in JavaFX)

Here's a small example how the new media features can be used:
```java
    @Override
        var button = new Button("Toggle full-screen");
        button.setOnAction(_ -> stage.setFullScreen(!stage.isFullScreen()));
        var label = new Label();
        var root = new BorderPane(button, null, null, label, null);
        var scene = new Scene(root, 650, 200);
        BorderPane.setAlignment(label, Pos.CENTER);
        label.textProperty().bind(scene.widthProperty().map(v -> String.format("Width: %.2f", v.doubleValue())));
        scene.getStylesheets().add("data:text/css;base64," + Base64.getEncoder().encodeToString("""
            @media (max-width: 500) {
                .button {
                    -fx-background-color: red;
                }
            }

            @media (600 < width <= 700) {
                .button {
                    -fx-background-color: green;
                }
            }

            @media (min-width: 800) {
                .button {
                    -fx-background-color: yellow;
                }
            }

            @media (display-mode: fullscreen) {
                .button {
                    -fx-background-color: black !important;
                }
            }
            """.getBytes(StandardCharsets.UTF_8)));

        stage.initStyle(StageStyle.DECORATED);
        stage.setScene(scene);
        stage.show();
    }
```

/reviewers 2
/csr

---------
### Progress
- [x] Change requires CSR request [JDK-8363983](https://bugs.openjdk.org/browse/JDK-8363983) to be approved

### Issues
 * [JDK-8358450](https://bugs.openjdk.org/browse/JDK-8358450): Viewport characteristics media features (**Enhancement** - P4)
 * [JDK-8368508](https://bugs.openjdk.org/browse/JDK-8368508): Missing info in JavaFX CSS Reference Guide (**Bug** - P4)
 * [JDK-8363983](https://bugs.openjdk.org/browse/JDK-8363983): Viewport characteristics media features (**CSR**)

### Reviewers
 * [Markus Mack](https://openjdk.org/census#mmack) (@drmarmac - Author)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1844/head:pull/1844` \
`$ git checkout pull/1844`

`$ git checkout pull/1844` \
`$ git pull https://git.openjdk.org/jfx.git pull/1844/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1844`

`$ git pr show -t 1844`

</details>

<a href="https://git.openjdk.org/jfx/pull/1844.diff">https://git.openjdk.org/jfx/pull/1844.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1844#issuecomment-3064698446)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
+++ b/modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
@@ -877,7 +877,7 @@
             <li>The <strong>interval form</strong> consists of a feature name, nested between two comparison
                 operators and two values. For example:<br>
                 <code>
-                    (400px &gt;= width &gt;= 600px)<br>
+                    (600px &gt;= width &gt;= 400px)<br>
                     (10em &lt; height &lt;= 20em)<br>
                 </code>
         </ol>


kevinrushforth:

kevinrushforth:
It would be nice to get this in for JavaFX 26, so will look at it soon.

andy-goryachev-oracle:
A CSS example (either in the ticket or the PR) would be nice, for ease of reviewing.

Here, I am trying to see if the button background will change when I move the monkey tester window from one monitor to another (one is 1900 px wide, the other is 2560).  I see no change:

```
    @media (width < 2000) {
        .button {
            -fx-background-color: red;
        }
    }
```

What am I doing wrong?

mstr2:
> What am I doing wrong?

All of the new media features query the scene, not the screen. If you resize the window, you should see the background color change.

mstr2:
> A CSS example (either in the ticket or the PR) would be nice, for ease of reviewing.

Added an example to the description of this PR.

andy-goryachev-oracle:
Having both `aspect-ratio` and `orientation` is a bit redundant, isn't it?  Is it just for convenience?

mstr2:
> Having both `aspect-ratio` and `orientation` is a bit redundant, isn't it? Is it just for convenience?

It almost seems like there could be historical reasons, like one preceded the other or something like that. But as far as I can see, both media features appear at the [same time](https://www.w3.org/TR/2008/WD-css3-mediaqueries-20081015/) around 2008, and they've been around ever since. So I guess it's mostly convenience.

andy-goryachev-oracle:
Why do I get stderr output with this:
```
WARNING: CSS Error parsing '    @media (aspect-ratio > 0.5) {
        .button {
            -fx-background-color: red;
        }
    }
: Invalid value <0.5> for media feature <aspect-ratio> at [1,30]
```

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
```diff
@@ -898,39 +931,79 @@ <h4>@media</h4>
...
+            </tr>
+            <tr>
+                <td class="value">orientation</td>
+                <td class="value">portrait | landscape</td>
+                <td>discrete</td>
+                <td>portrait if height &gt;= width, landscape otherwise</td>
+            </tr>
+            <tr>
+                <td class="value">display-mode</td>
+                <td class="value">fullscreen | standalone</td>
```
andy-goryachev-oracle, line 970: what does "standalone" mean?

also, "fullscreen" applies to a window, but earlier it's mentioned that the media queries apply to Scene.

I don't see the fullscreen styles change when I use the monkey tester Stage page (the one that allows me to set the fullscreen `Stage` bit) with the following CSS:

```
    @media (display-mode: fullscreen) {
        .button {
            -fx-background-color: red;
        }
    }
```

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaFeatures.java
```diff
@@ -105,8 +213,46 @@ private static <T extends Enum<T>> T enumValue(Function<String, T> func, String
...
+    private static String lowerCaseTextValue(String text) {
+        return text != null ? text.toLowerCase(Locale.ROOT).intern() : null;
+    }
+
+    private static Size lengthValue(String featureName, String lowerCaseText) {
+        int unitIndex = -1;
+
+        for (int i = 0; i < lowerCaseText.length(); i++) {
+            if (!Character.isDigit(lowerCaseText.charAt(i))) {
+                unitIndex = i;
```
andy-goryachev-oracle: I think this is the place where the parser fails with fractional numbers...

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/SizeQueryType.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public String getFeatureName() {
+        return featureName;
+    }
+
+    public SizeSupplier getSupplier() {
+        return supplier;
+    }
+
+    private static final SizeQueryType[] VALUES = values();
```
andy-goryachev-oracle, line 114: is this really needed?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/parser/CssLexer.java
```diff
@@ -49,6 +49,8 @@ public final class CssLexer {
     public static final int RAD = 25;
     public static final int TURN = 26;
     public static final int GREATER = 27;
+    public static final int LESS = 48;
+    public static final int EQUALS = 49;
```
andy-goryachev-oracle: this **will** cause problems.  could we arrange these by value please (add new after `AT_KEYWORD`)?

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
```diff
@@ -64,9 +64,10 @@ public class Stylesheet {
      * binary stream changes. This number does not correlate with JavaFX versions.
      * Version 5: persist @font-face
      * Version 6: converter classes moved to public package
-     * Version 7: media queries
+     * Version 7: user-preference media queries
+     * Version 8: viewport characteristics media queries
      */
-    final static int BINARY_CSS_VERSION = 7;
+    final static int BINARY_CSS_VERSION = 8;
```
andy-goryachev-oracle, line 70: Does this impact backward compatibility with .bss files?
Should it be mentioned in the CSR?

modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
```diff
@@ -2765,8 +2771,11 @@ public void changedLocation(float x, float y) {
 
         @Override
         public void changedSize(float w, float h) {
-            if (w != Scene.this.getWidth()) Scene.this.setWidth(w);
-            if (h != Scene.this.getHeight()) Scene.this.setHeight(h);
+            boolean widthChanged = w != Scene.this.getWidth();
+            boolean heightChanged = h != Scene.this.getHeight();
+            if (widthChanged) Scene.this.setWidth(w);
```
andy-goryachev-oracle: it might be my personal preference, but I would rather see one statement per line.  use case: stepping through in the debugger.  can we add { }'s please?

modules/javafx.graphics/src/test/java/test/javafx/css/CssParser_mediaQuery_Test.java
```diff
@@ -439,7 +475,204 @@ void missingClosingCurlyBraceEmitsParserError() {
...
+    void parseRangeValue_asDiscrete() {
+        Stylesheet stylesheet = new CssParser().parse("""
+            @media (width: 100px) {
+                .foo { bar: baz; }
+            }
+            """);
+
+        assertEquals(1, stylesheet.getRules().size());
+        assertEquals(
+            List.of(EqualExpression.of(SizeQueryType.WIDTH, new Size(100, SizeUnits.PX))),
```
andy-goryachev-oracle: Could we add a test that verifies that the CSS parser accepts fractional decimals?  With or without the units specified?

Also, does it support scientific notation line `1e2` for `100` ?

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaFeatures.java
```diff
@@ -105,8 +213,46 @@ private static <T extends Enum<T>> T enumValue(Function<String, T> func, String
...
+            case "px" -> SizeUnits.PX;
+            case "em" -> SizeUnits.EM;
+            case "ex" -> SizeUnits.EX;
+            case "cm" -> SizeUnits.CM;
+            case "mm" -> SizeUnits.MM;
+            case "in" -> SizeUnits.IN;
+            case "pt" -> SizeUnits.PT;
+            case "pc" -> SizeUnits.PC;
+            default -> throw new IllegalArgumentException(
+                String.format("Invalid value <%s> for media feature <%s>", lowerCaseText, featureName));
```
andy-goryachev-oracle: is this a new code?
could we add a unit test for all the combinations of decimal values (integer, fractional, scientific), with and without the units?

andy-goryachev-oracle (changes requested):

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
mstr2: "Standalone" matches when "fullscreen" doesn't match. The relevant part of the [specification](https://www.w3.org/TR/mediaqueries-5/#display-modes) says that it is a window with "[...] standard system UI elements, such as window decorations, a system status bar, and/or a system back button [...]".

"Fullscreen" matches when none of that is available and the scene extends across the entire display, which is the case when the scene is hosted in a full-screen window.

Note that media queries don't change CSS precedence rules. In your example, you could check whether another rule that sets `-fx-background-color` has a higher precedence. You could try adding `!important` to quickly figure out whether the rule has any effect at all.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/SizeQueryType.java
mstr2: It's not strictly needed. If we remove it, then the enum will allocate a new array on every iteration over its `values()`.

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
mstr2: BSS files are not specified to be backwards-compatible in the sense that an older version of JavaFX can read a newer BSS file. It is mentioned in the CSR, though.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaFeatures.java
mstr2: Yes. It turns out that we need to differentiate `<size>` and `<number>` values, as the latter don't have a unit (and can never have one). I've refactored the code a bit to handle both types of ranges.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaFeatures.java
mstr2: I've removed the custom parsing, and instead delegate to `CssParser` to parse sizes. This means we won't have to worry about any of this as part of this PR; whatever `CssParser` is doing right or wrong, media queries will use that.

modules/javafx.graphics/src/test/java/test/javafx/css/CssParser_mediaQuery_Test.java
mstr2: Even though the CSS accepts scientific notation, `CssLexer` apparently does not. This could be a future enhancement.
I've removed all custom code with regards to parsing numbers, and just delegate to `CssParser`.

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
andy-goryachev-oracle: What I meant is to explain the meaning in the cssref.html.  We can't really refer to w3.org pages for JavaFX, in my opinion.  So, if you want to keep the standalone term in the values, I think you ought to explain what it is, using JavaFX or platform terminology.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/SizeQueryType.java
andy-goryachev-oracle: unless this code is in hot path, I'd suggest to remove the static.

modules/javafx.graphics/src/main/java/javafx/css/Stylesheet.java
andy-goryachev-oracle: I was expecting it in the description - I could be wrong though.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/expression/RangeExpression.java
```diff
@@ -74,15 +101,21 @@ public final boolean equals(Object obj) {
...
+        return Objects.hash(getClass(), featureType, sizeValue, numberValue);
+    }
+
+    final String getFormattedValue() {
+        return sizeValue != null ? sizeValue.toString() : Double.toString(numberValue);
     }
 
     public interface Supplier {
-        RangeExpression rangeExpression(SizeQueryType featureType, Size featureValue);
+        RangeExpression getSizeExpression(SizeQueryType featureType, Size sizeValue);
```
andy-goryachev-oracle, line 118: minor: could we use a name other than Supplier to avoid clashing with java.util.function one (it is confusing).

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaFeatures.java
```diff
@@ -49,28 +48,27 @@ private MediaFeatures() {}
...
      */
-    public static MediaQuery discreteQueryExpression(String featureName, String featureValue) {
+    public static MediaQuery discreteQueryExpression(String featureName, Token featureValue) {
         String lowerCaseFeatureName = lowerCaseTextValue(featureName);
-        String lowerCaseFeatureValue = lowerCaseTextValue(featureValue);
 
         return switch (lowerCaseFeatureName) {
             // Discrete min-/max-features are just features in a range context in disguise.
-            case "min-width" -> rangeQueryExpression(SizeQueryType.WIDTH, lowerCaseFeatureValue,
+            case "min-width" -> rangeQueryExpression(SizeQueryType.WIDTH, featureValue,
```
andy-goryachev-oracle: very minor suggestion:
```
case "min-width" ->
    rangeQueryExpression(SizeQueryType.WIDTH, featureValue, ComparisonOp.GREATER_OR_EQUAL.getExpressionSupplier());
```

andy-goryachev-oracle (commented):
Thank you for making the changes!

However, `display-mode` does not seem to work on macOS 26.1.  To reproduce:

- run the standalone monkey tester https://github.com/andy-goryachev-oracle/MonkeyTest
- open Tools -> CSS Playground, copy and paste the following stylesheet:

```
    @media (display-mode: fullscreen) {
        .button {
            -fx-background-color: red;
        }
    }
``` 
- click Update
- select the Button page
- use Window -> Fullscreen to enter the fullscreen
I expect the buttons to turn red, but this does not happen (the earlier test used the Stage page, same result).

Alternatively, if I use `display-mode: standalone ` query, the buttons immediately turn red, but remain unchanged in fullscreen.

mstr2:
> However, `display-mode` does not seem to work on macOS 26.1. To reproduce: [...]

Fixed that. I was using an `InvalidationListener` that would not be called again if the value of the observable was not actually read.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaFeatures.java
mstr2: I've moved the code more towards the front of the line, similarly to how it's already done for the user-preference features later in the `switch`.

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
mstr2: I've added a description to cssref.html that says what the two values mean in JavaFX terms.

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
```diff
@@ -962,13 +962,13 @@ <h5>Combining Media Features</h5>
...
                 <td>discrete</td>
-                <td>portrait if height &gt;= width, landscape otherwise</td>
+                <td><code>portrait</code> if height &gt;= width, <code>landscape</code> otherwise</td>
             </tr>
             <tr>
                 <td class="value">display-mode</td>
                 <td class="value">fullscreen | standalone</td>
                 <td>discrete</td>
-                <td></td>
+                <td><code>fullscreen</code> if <code>Stage.isFullScreen()</code>, <code>standalone</code> otherwise</code></td>
```
andy-goryachev-oracle: This description is perfectly clear now, thanks!

One question though: at the beginning of the chapter, you say "A media query is a method of testing certain aspects of the [Scene](https://github.com/openjdk/jfx/javafx/scene/Scene.html)" but this is not strictly true, since the Stage is also involved.

Do you think we should mention "and/or Stage" or something like that?

Moreover, what if other media queries are added that depend on some other aspects?  Maybe we should specify whose properties we are looking at in the comments, like "width" -> "Scene width" ?  What do you think?

andy-goryachev-oracle (approved):
The feature looks good on macOS 26.1, excellent work!

This needs to be tested on Windows and Linux[es].

Side note: I realized the CSS Playground does not apply the stylesheet to newly created windows, only to already present ones.  Should be ok to test though, with newly added Windows -> Fullscreen etc. menus.

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
mstr2: Good idea, I've added that to the description.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/expression/RangeExpression.java
mstr2: I've thought about that, but then I never use that type unqualified (i.e. always as `RangeExpression.Supplier`), and because it's so localized, I don't want to make it a top-level interface. A more verbose name like `RangeExpression.ExpressionSupplier` looks very chatty...

andy-goryachev-oracle: ESupplier? RProvider?  but if the access is always qualified, we are good.

andy-goryachev-oracle:
for other reviewers: I've updated the monkey tester's CSS Playground to apply the stylesheets to the newly created windows.

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
```diff
@@ -842,21 +847,50 @@ <h4>@media</h4>
...
+                and a value. For example:<br>
+                <code>
+                    (width &gt; 600px)<br>
+                    (500px &lt;= height)<br>
+                </code>
+                <br>
+            <li>The <strong>interval form</strong> consists of a feature name, nested between two comparison
+                operators and two values. For example:<br>
+                <code>
+                    (400px &gt;= width &gt;= 600px)<br>
```
drmarmac: This should be `(400px &lt;= width &lt;= 600px)`

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaQuery.java
```diff
@@ -42,7 +43,16 @@ public sealed interface MediaQuery
...
+                NegationExpression,
+                RangeExpression {
+
+    /**
+     * Gets the context awareness flags of this media query, indicating which aspects of the
+     * media query context are probed by the query.
+     *
+     * @return the context awareness flags
+     */
+    int getContextAwareness();
```
drmarmac, line 55: Have you considered using an EnumSet here for stronger typing?

mstr2: Yes, it also was an `EnumSet` in an earlier revision. However, these expressions are immutable, which an `EnumSet` is not. So I'd have to wrap the enum set in an unmodifiable set, which adds indirections on a hot path (or accept that these structures are not truly immutable). A primitive `int` is immutable by nature...

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
mstr2: @kevinrushforth This is a typo that affects an already-approved CSR, but fixing it doesn't meaningfully change the specification. Given that I'd like to get this PR in for JFX 26, how do we continue here?

mstr2:
> When horizontally resizing the window around the width threshold, sometimes the transition jumps or even affects one of the buttons differently than the other. Is such a configuration supposed to be supported?

It looks like it should be supported. I'll file a follow-up ticket to track this bug.

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
kevinrushforth: Since it is not a material change that would, by itself, need a CSR, go ahead and proceed with the integration.

kevinrushforth: To clarify a bit, you can fix this now if you like (it will require a re-review of the PR) or file a follow-up bug and fix it. It's probably easiest to fix it now.

kevinrushforth: If you do fix it now, I recommend adding a comment to the CSR with the diff.

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
```diff
@@ -877,7 +877,7 @@ <h5>Evaluating Media Features in a Range Context</h5>
             <li>The <strong>interval form</strong> consists of a feature name, nested between two comparison
                 operators and two values. For example:<br>
                 <code>
-                    (400px &gt;= width &gt;= 600px)<br>
+                    (600px &gt;= width &gt;= 400px)<br>
```
andy-goryachev-oracle, line 880: it's probably customary to go from small value to large
`400 >= width >= 600`

mstr2: Yes, but I think it's better to show both versions here, since otherwise people might think that only one form is supported.

andy-goryachev-oracle (approved):

mstr2:
Thanks @andy-goryachev-oracle and @drmarmac for the reviews!

PR #1845, opened 2025-07-13
8362091: Window title bar should reflect scene color scheme

mstr2:
Currently, the color scheme of a system-decorated stage is as follows:
* On Windows, the title bar is always light (even if the OS color scheme is dark).
* On macOS and Linux, the title bar is light or dark depending on the OS color scheme.

The expected behavior is that the title bar matches the color scheme of the `Scene`.
If an application doesn't specify a color scheme, the title bar color should match the OS color scheme.

This PR fixes the behavior for Windows and macOS, but not for Linux (there's no good way to do that).
Depending on how you look at it, this is either a bug fix or an enhancement.

/reviewers 2

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/win/GlassWindow.cpp
+++ b/modules/javafx.graphics/src/main/native-glass/win/GlassWindow.cpp
@@ -1258,6 +1258,55 @@ void GlassWindow::SetIcon(HICON hIcon)
     m_hIcon = hIcon;
 }
 
+void GlassWindow::SetDarkFrame(bool dark)
+{
+    // The value of the DWMWA_USE_IMMERSIVE_DARK_MODE constant may be different depending on the OS version.
+    // We are going to query the file version of dwmapi.dll to make sure we use the right constant, or the
+    // value 0 to indicate that we don't support this feature.
+    // See: https://github.com/MicrosoftDocs/sdk-api/commit/c19f1c8a148b930444dce998d3c717c8fb7751e1
+    static const DWORD DWMWA_USE_IMMERSIVE_DARK_MODE = []() {
+        DWORD ignored;
+        DWORD infoSize = GetFileVersionInfoSizeEx(FILE_VER_GET_NEUTRAL, L"dwmapi.dll", &ignored);
+        if (infoSize <= 0) {
+            return 0;
+        }
+
+        std::vector<char> buffer(infoSize);
+        if (!GetFileVersionInfoEx(FILE_VER_GET_NEUTRAL, L"dwmapi.dll", ignored,
+                                  static_cast<DWORD>(buffer.size()), &buffer[0])) {
+            return 0;
+        }
+
+        UINT size = 0;
+        VS_FIXEDFILEINFO* fileInfo = nullptr;
+        if (!VerQueryValue(buffer.data(), L"\\", reinterpret_cast<LPVOID*>(&fileInfo), &size)) {
+            return 0;
+        }
+
+        WORD major = HIWORD(fileInfo->dwFileVersionMS);
+        WORD minor = LOWORD(fileInfo->dwFileVersionMS);
+        WORD build = HIWORD(fileInfo->dwFileVersionLS);
+
+        // Windows 10 before build 10.0.17763: not supported
+        if (major < 10 || (major == 10 && minor == 0 && build < 17763)) {
+            return 0;
+        }
+
+        // Windows 10 build 10.0.17763 until 10.0.18985
+        if (major == 10 && minor == 0 && build >= 17763 && build < 18985) {
+            return 19;
+        }
+
+        // Windows 10 build 10.0.18985 or later
+        return 20;
+    }();
+
+    if (DWMWA_USE_IMMERSIVE_DARK_MODE) {
+        BOOL darkMode = dark;
+        DwmSetWindowAttribute(GetHWND(), DWMWA_USE_IMMERSIVE_DARK_MODE, &darkMode, sizeof(darkMode));
+    }
+}
+
 void GlassWindow::ShowSystemMenu(int x, int y)
 {
     WINDOWPLACEMENT placement;
@@ -1453,6 +1502,10 @@ JNIEXPORT jlong JNICALL Java_com_sun_glass_ui_win_WinWindow__1createWindow
                     ::EnableMenuItem(hSysMenu, SC_CLOSE,
                             MF_BYCOMMAND | MF_DISABLED | MF_GRAYED);
                 }
+            }
+
+            if (mask & com_sun_glass_ui_Window_DARK_FRAME) {
+                pWindow->SetDarkFrame(true);
             }
         }
 
@@ -1692,6 +1745,28 @@ JNIEXPORT jboolean JNICALL Java_com_sun_glass_ui_win_WinWindow__1setBackground2
     PERFORM();
 
     return JNI_TRUE;
+}
+
+/*
+ * Class:     com_sun_glass_ui_win_WinWindow
+ * Method:    _setDarkFrame
+ * Signature: (JZ)V
+ */
+JNIEXPORT void JNICALL Java_com_sun_glass_ui_win_WinWindow__1setDarkFrame
+    (JNIEnv *env, jobject jThis, jlong ptr, jboolean dark)
+{
+    ENTER_MAIN_THREAD()
+    {
+        GlassWindow *pWindow = GlassWindow::FromHandle(hWnd);
+        if (pWindow) {
+            pWindow->SetDarkFrame(dark);
+        }
+    }
+    jboolean dark;
+    LEAVE_MAIN_THREAD_WITH_hWnd;
+
+    ARG(dark) = dark;
+    PERFORM();
 }
 
 /*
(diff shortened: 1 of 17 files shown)


modules/javafx.graphics/src/main/native-glass/win/GlassWindow.cpp
```diff
@@ -1258,6 +1258,55 @@ void GlassWindow::SetIcon(HICON hIcon)
     m_hIcon = hIcon;
 }
 
+void GlassWindow::SetDarkFrame(bool dark)
+{
+    // The value of the DWMWA_USE_IMMERSIVE_DARK_MODE constant may be different depending on the OS version.
+    // We are going to query the file version of dwmapi.dll to make sure we use the right constant, or the
+    // value 0 to indicate that we don't support this feature.
+    // See: https://github.com/MicrosoftDocs/sdk-api/commit/c19f1c8a148b930444dce998d3c717c8fb7751e1
+    static const DWORD DWMWA_USE_IMMERSIVE_DARK_MODE = []() {
```
Maran23, line 1268: This implementation looks correct. For comparison, I checked out the implementation in `Godot`.

Version check: https://github.com/godotengine/godot/blob/master/platform/windows/display_server_windows.cpp#L6853 
Set the attribute: https://github.com/godotengine/godot/blob/master/platform/windows/display_server_windows.cpp#L4708

Note that the `build` numbers differ a little bit from here. But checking your linked commit, your numbers seem to be the right one! Will do some testing later (on Windows 11).

bundgaard:
Hi @bundgaard, thanks for making a comment in an OpenJDK project!

All comments and discussions in the OpenJDK Community must be made available under the OpenJDK [Terms of Use](https://openjdk.java.net/legal/tou/). If you already are an OpenJDK [Author](https://openjdk.java.net/bylaws#author), [Committer](https://openjdk.java.net/bylaws#committer) or [Reviewer](https://openjdk.java.net/bylaws#reviewer), please click [here](https://bugs.openjdk.java.net/secure/CreateIssue.jspa?pid=11300&issuetype=1) to open a new issue so that we can record that fact. Please Use "Add GitHub user bundgaard" for the summary.

mstr2:
> Is the extra code (beside looking at the version) so that I can toggle the darkbar from Java?

We have two places in JavaFX that are relevant to dark mode:
1. `Platform.Preferences.colorScheme`, which tells you the color scheme of the operating system (i.e. what you select in PC settings)
2. `Scene.Preferences.colorScheme`, which defaults to the OS color scheme, but is overridable for each scene.

This PR proposes to match the OS-provided title bar color to the value of `Scene.Preferences.colorScheme`, which allows developers to choose how their application should be presented:
* If you don't override the scene color scheme, the window title bar will match the OS by default (light or dark).
* However, if your application uses a fixed light or dark theme (i.e. it doesn't adapt its control theme colors to dark mode), then overriding the scene color scheme ensures consistent presentation independent of OS settings.

modules/javafx.graphics/src/main/native-glass/win/GlassWindow.cpp
```diff
@@ -1258,6 +1258,55 @@ void GlassWindow::SetIcon(HICON hIcon)
...
+        DWORD ignored;
+        DWORD infoSize = GetFileVersionInfoSizeEx(FILE_VER_GET_NEUTRAL, L"dwmapi.dll", &ignored);
```
kevinrushforth: Since you are explicitly passing a wide string constant, shouldn't you also explicitly use the `GetFileVersionInfoSizeExW` variant of the function? (and yes, I know we define `/UNICODE` but it still seems cleaner to be explicit in this case).

modules/javafx.graphics/src/main/native-glass/win/GlassWindow.cpp
```diff
@@ -1258,6 +1258,55 @@ void GlassWindow::SetIcon(HICON hIcon)
...
+        if (infoSize <= 0) {
+            return 0;
+        }
+
+        std::vector<char> buffer(infoSize);
+        if (!GetFileVersionInfoEx(FILE_VER_GET_NEUTRAL, L"dwmapi.dll", ignored,
```
kevinrushforth: Use `GetFileVersionInfoExW`?

modules/javafx.graphics/src/main/native-glass/win/GlassWindow.cpp
```diff
@@ -1258,6 +1258,55 @@ void GlassWindow::SetIcon(HICON hIcon)
...
+                                  static_cast<DWORD>(buffer.size()), &buffer[0])) {
+            return 0;
+        }
+
+        UINT size = 0;
+        VS_FIXEDFILEINFO* fileInfo = nullptr;
+        if (!VerQueryValue(buffer.data(), L"\\", reinterpret_cast<LPVOID*>(&fileInfo), &size)) {
```
kevinrushforth: `VerQueryValueW`?

kevinrushforth (commented):
The changes look reasonable to me. I left a few comments. I still need to test the enhancement.

Speaking of testing, this causes a failure in one of our closed white-box tests. It's trivial for me to fix it, but it will mean that down the road, when this is ready to go in, we will need to coordinate the timing of the integration with you.

kevinrushforth (approved):
My testing looks good.

Suggestion: consider a manual test for this (maybe adding something to MonkeyTester?). We don't have a test program in our repo that sets `Scene.Preference.colorScheme`.

kevinrushforth:
This is pending a second reviewer.

@mstr2 I have updated our closed whitebox test as mentioned above, so it will be ready to go in when your PR is ready. Check with me as to the timing of integrating this.

mstr2:
> My testing looks good.
> 
> Suggestion: consider a manual test for this (maybe adding something to MonkeyTester?). We don't have a test program in our repo that sets `Scene.Preference.colorScheme`.

I was planning to do something like this for [System Colors](https://gist.github.com/mstr2/afac42ab9587c1040a6cf7b091cee99f) on the CSS features arc, which includes all kinds of color-related system properties.

andy-goryachev-oracle:
> Suggestion: consider a manual test for this (maybe adding something to MonkeyTester?). We don't have a test program in our repo that sets `Scene.Preference.colorScheme`.

This could be added to the `StagePage` there (I know there is a Stage Tester tool, but it does not remember user choices and probably can be replaced with the `StagePage`).

andy-goryachev-oracle:
I've added scene preferences to the monkey tester

but for some reason, the DARK color scheme does not show up on my mac when set via properties.
Did I do something wrong?
https://github.com/andy-goryachev-oracle/MonkeyTest/blob/33702ae9f753a4d4857dbb38258c2e42b61eaecc/src/com/oracle/tools/fx/monkey/pages/StagePage.java#L161

mstr2:
> but for some reason, the DARK color scheme does not show up on my mac when set via properties. Did I do something wrong?

~~It seems like you never read the scene color scheme, you're only setting it.~~
Never mind, I think I misunderstood what you're saying. I'll have a look at it.

kevinrushforth:
FWIW, here is the test program I used to verify this on Windows and macOS.
[HelloDarkMode.java](https://github.com/user-attachments/files/22645745/HelloDarkMode.java)

kevinrushforth:
The important bit is:

```java
        scene.getPreferences().setColorScheme(ColorScheme.DARK);
```

andy-goryachev-oracle:
> It seems like you never read the scene color scheme, you're only setting it.

The Platform.Preferences are read in L114 to initialize the UI properties, then the Scene.Preferences are set in L162.  After I change the option via UI to DARK the window shows up in LIGHT.

kevinrushforth:
I ran the MonkeyTester Andy referred above and it works for me. With the fix, I see a dark title bar when setting Scene.Preferences.colorScheme to DARK and a lighter title bar otherwise.

I'm still running on macOS 14 (14.8 to be exact).

mstr2:
I checked out the [MonkeyTest](https://github.com/andy-goryachev-oracle/MonkeyTest) repo, and it also works for me on macOS 26.

andy-goryachev-oracle:
rebooted my machine.  works as expected on macos 15.7 M1

andy-goryachev-oracle:
btw, should the rest of the ui come out in dark mode when `Scene.Preferences.colorSceme=DARK` is set?

Screenshot 2025-10-01 at 11 15 06

mstr2:
> btw, should the rest of the ui come out in dark mode when `Scene.Preferences.colorSceme=DARK` is set?

Do you mean the controls? Not for now, because Modena doesn't have a dark variant. But aspirationally, sure. We should continue the discussion of a Modena Dark theme or an entirely new theme.

kevinrushforth:
> /integrate

I guess you missed my earlier comment about coordinating on the timing? I'll get the closed side in ASAP, so it won't actually be a problem.

andy-goryachev-oracle:
I approved the closed part, ready to go.

mstr2:
> I guess you missed my earlier comment about coordinating on the timing? I'll get the closed side in ASAP, so it won't actually be a problem.

I read your comment and noted it, but then it somehow slipped my mind. Sorry about that.

kevinrushforth:
No worries. I got the closed half in quickly enough that we won't have even a momentary bit of noise in our CI builds.

Maran23:
Had no time to further check this unfortunately, but did a lot of testing today and found no issues on Windows 11! 
Tested also with an application where I implemented a dark/light mode switch with the new `@media` queries and `prefers-color-scheme`.

https://github.com/user-attachments/assets/48fc2b82-1cbe-4067-a7a2-b21ce3052e7c

As mentioned above, Godot has the same logic as we have in the Windows code, so the Windows part looked good to me.

PR #1850, opened 2025-07-22
8362873: Regression in BorderPane after JDK-8350149

hjohn:
This PR fixes a regression introduced in #1723

It reverts the changes made in the BorderPaneTest with new comments explaining why the values are what they are.  It only fixes the regression that was introduced, but it can be said that the calculation is probably also incorrect when a vertical biased control is encountered (the `getAreaWidth` and `getAreaHeight` in `BorderPane` were, and are still not using the same way to calculate their respective sizes).  See the ticket for more details and screenshots.

Let me know if I should only address the regression introduced, or also include the fix for the bias calculation for vertical bias controls.

---------
### Progress

### Issue
 * [JDK-8362873](https://bugs.openjdk.org/browse/JDK-8362873): Regression in BorderPane after JDK-8350149 (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1850/head:pull/1850` \
`$ git checkout pull/1850`

`$ git checkout pull/1850` \
`$ git pull https://git.openjdk.org/jfx.git pull/1850/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1850`

`$ git pr show -t 1850`

</details>

<a href="https://git.openjdk.org/jfx/pull/1850.diff">https://git.openjdk.org/jfx/pull/1850.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1850#issuecomment-3100034731)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/layout/BorderPane.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/layout/BorderPane.java
@@ -626,8 +626,8 @@ public class BorderPane extends Pane {
     private double getAreaHeight(Node child, double width, boolean minimum) {
         if (child != null && child.isManaged()) {
             Insets margin = getNodeMargin(child);
-            return minimum ? computeChildMinAreaHeight(child, -1, margin, width, false):
-                                   computeChildPrefAreaHeight(child, -1, margin, width, false);
+            return minimum ? computeChildMinAreaHeight(child, -1, margin, width, true):
+                                   computeChildPrefAreaHeight(child, -1, margin, width, true);
         }
         return 0;
     }
--- a/modules/javafx.graphics/src/test/java/test/javafx/scene/layout/BorderPaneTest.java
+++ b/modules/javafx.graphics/src/test/java/test/javafx/scene/layout/BorderPaneTest.java
@@ -343,12 +343,25 @@ public class BorderPaneTest {
         MockBiased bottom = new MockBiased(Orientation.HORIZONTAL, 200, 20); // 300 x 13.333
         borderpane.setBottom(bottom);
 
+        /*
+         * Note: there is a mix of horizontal and vertical biased children. The BorderPane
+         * will favor the bias of the center node above all (see its implementation). This
+         * means the VERTICAL biased controls will have their bias ignored as they differ
+         * from the center's bias.
+         */
+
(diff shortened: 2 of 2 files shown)


hjohn:
@Glavo this should fix the regression

@kevinrushforth since the regression was introduced in the version we're about to release, we may want to consider this for immediate backport to avoid shipping it in the final version.

modules/javafx.graphics/src/main/java/javafx/scene/layout/BorderPane.java
```diff
@@ -626,8 +626,8 @@ private double getAreaWidth(Node child, double height, boolean minimum) {
     private double getAreaHeight(Node child, double width, boolean minimum) {
         if (child != null && child.isManaged()) {
             Insets margin = getNodeMargin(child);
-            return minimum ? computeChildMinAreaHeight(child, -1, margin, width, false):
-                                   computeChildPrefAreaHeight(child, -1, margin, width, false);
+            return minimum ? computeChildMinAreaHeight(child, -1, margin, width, true):
+                                   computeChildPrefAreaHeight(child, -1, margin, width, true);
```
hjohn, line 630: Note: before the last boolean parameter was introduced in the `computeChildXXXAreaHeight` functions in `Region`, the default behavior was to have fill width set to `true`.  The reason I picked `false` is because the calculation above in `getAreaWidth` was using `false`, but it looks now that this was never correct.

kevinrushforth:
> @kevinrushforth since the regression was introduced in the version we're about to release, we may want to consider this for immediate backport to avoid shipping it in the final version.

Agreed. This is a good candidate to backport to jfx25.

> Let me know if I should only address the regression introduced, or also include the fix for the bias calculation for vertical bias controls.

Since we want to backport this to jfx25, it would be safer to just fix the regression introduced in 25.

/reviewers 2

kevinrushforth:
Reviewers: @kevinrushforth @andy-goryachev-oracle

andy-goryachev-oracle:
This PR fixes the vertical layout case (`verticalLayout = true;`).

Do we have a ticket for the incorrect computation for the horizontal layout?  If not, could you please create a follow-up one?

Screenshot 2025-07-22 at 09 08 16

andy-goryachev-oracle (approved):

PR #1856, opened 2025-07-25
8364049: ToolBar shows overflow menu with fractional scale

andy-goryachev-oracle:
This is a very localized fix for the issue described in https://bugs.openjdk.org/browse/JDK-8364049 which resulted from comparing snapped and non-snapped values.  The issue seems to happen only with fractional scale, that is, on Windows at 125%, 150%, 175% etc. scales.

While looking at the `ToolBarSkin` code, I noticed a general pattern related to snapping, that might cause similar issues in the tool bar skin and elsewhere.  Here is an example in `ToolBarSkin::computePrefWidth()` in L411:

```
        if (toolbar.getOrientation() == Orientation.HORIZONTAL) {
            for (Node node : toolbar.getItems()) {
                if (!node.isManaged()) continue;
                prefWidth += snapSizeX(node.prefWidth(-1)) + getSpacing();
            }
            prefWidth -= getSpacing();
        } else {
```

the general issue, in my opinion, is that doing `prefWidth += snapSizeX(node.prefWidth(-1)) + getSpacing();` results in the `prefWidth` value differ from its snapped value.  In other words, whenever computation involves snapped values, the result must be snapped as well - and that includes the case when all the parts of the computation are snapped.

Another, related, topic is how to properly snap the values in the computation.  I would say ideally it should be done like this:

```
snappedResult = snap(snap(value1) .OP. snap(value2) .OP. ... snap (valueN))
```

It might be possible to skip the snapping of intermediary values, and only snap the result, but one must be careful not to accumulate errors.

Getting back to the ToolBarSkin, one can see the issue on LL392, 399, 411, 417, 425, 436, 530, and so on.

I decided not to fix the snapping for the purpose of making this PR narrow in scope with the goal to backport it to jfx25, but I did want to describe the issue.

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -686,6 +686,11 @@ public class ToolBarSkin extends SkinBase<ToolBar> {
      * @return the index of the first node that does not fit in the toolbar, or the size of the items list else
      */
     private int getOverflowNodeIndex(double length) {
+        if (getSkinnable().getOrientation() == Orientation.VERTICAL) {
+            length = snapSizeY(length);
+        } else {
+            length = snapSizeX(length);
+        }
         ObservableList<Node> items = getSkinnable().getItems();
         int overflowIndex = items.size();
         double x = 0;


kevinrushforth:
@arapte Can you review this?

/reviewers 2

andy-goryachev-oracle:
@mstr2 could you also take a look at this please?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
```diff
@@ -689,6 +689,11 @@ private double getToolbarLength(ToolBar toolbar) {
      * @return the index of the first node that does not fit in the toolbar, or the size of the items list else
      */
     private int getOverflowNodeIndex(double length) {
+        if (getSkinnable().getOrientation() == Orientation.VERTICAL) {
```
mstr2: Snapping the input argument of a method like `getOverflowNodeIndex` doesn't seem right to me, because the length is computed incorrectly in the first place. Have you considered fixing the snapping in `getToolbarLength`? This also solves the problem.

andy-goryachev-oracle: Yes, but I wanted to do a very localized fix, to be included in jfx25.  There are many places where we should be more careful with snapping mentioned in description, and I suspect even more elsewhere.

I am still not sure what would be the best approach to address them all.  We could probably create an umbrella task and fix individual controls and containers (maybe even start with containers, in continuation of John's work with HBox and VBox).

I think this PR is still good, because a) it provides a fix for the specific issue and b) minimizes regression, but I agree with you that a comprehensive fix is needed.

What do you think?

mstr2: I understand that you want this to be a localized fix. I'm just proposing to change `getToolbarLength` instead, which is even less changed code than your change in `getOverflowNodeIndex`, and seems to be addressing the source of the incorrect length instead.

andy-goryachev-oracle: Upon closer inspection, your suggestion is better: even though the change would impact more code paths including the layout, the actual value produced gets fed into the same `getOverflowNodeIndex()` function.

Thanks!

mstr2:
I think it should be possible to create a unit test here. What should be asserted is that the overflow menu doesn't depend on various scale factors.

andy-goryachev-oracle:
I can't get it to fail with a unit test, and I don't want a headful test here.

Here is a test which passes in master though it should fail (a part of ToolbarTest):

```
    @Test
    public void noOverflowAtScales() {
        // typical windows scales
        double[] scales = {
            1.0,
            1.25,
            1.5,
            1.75,
            2.0,
            2.25
        };

        Button b = new Button("Create Schema");
        b.setFont(Font.font("System", 13));
        toolBar.getItems().add(b);

        BorderPane bp = new BorderPane();
        bp.setTop(new HBox(toolBar));

        stage = new Stage();
        stage.setScene(new Scene(bp, 600, 600));
        stage.show();
        tk.firePulse();

        toolBar.setOrientation(Orientation.HORIZONTAL);
        tk.firePulse();
        for (double scale : scales) {
            toolBar.setScaleX(scale);
            toolBar.setScaleY(scale);
            
            double w = bp.prefWidth(-1);
            bp.resize(w, 100);
            
            tk.firePulse();
            tk.firePulse();
            assertOverflowNotShown();
            System.out.println(toolBar.getScaleX());
        }

        toolBar.setOrientation(Orientation.VERTICAL);
        tk.firePulse();
        for (double scale : scales) {
            toolBar.setScaleX(scale);
            toolBar.setScaleY(scale);
            tk.firePulse();
            assertOverflowNotShown();
        }
    }
```

mstr2:
Here's a test that fails without, and passes with the fix:

```java
    @Test
    public void overflowMenuNotShowingWithDifferentRenderScales() {
        double[] renderScales = {
            1.0,
            1.25,
            1.5,
            1.75,
            2.0,
            2.25
        };

        var rect = new Rectangle(100, 100);
        var toolBar = new ToolBar(rect);
        toolBar.setSkin(new ToolBarSkin(toolBar));

        for (var orientation : Orientation.values()) {
            toolBar.setOrientation(orientation);

            for (double scale : renderScales) {
                Stage stage = new Stage();
                stage.renderScaleXProperty().bind(DoubleConstant.valueOf(scale));
                stage.renderScaleYProperty().bind(DoubleConstant.valueOf(scale));
                stage.setScene(new Scene(new HBox(toolBar), 600, 600));
                stage.show();
                assertOverflowNotShown(toolBar);
                stage.close();
            }
        }
    }

    private void assertOverflowNotShown(ToolBar toolBar) {
        Pane overflowButton = (Pane)toolBar.queryAccessibleAttribute(AccessibleAttribute.OVERFLOW_BUTTON);
        assertNotNull(overflowButton);
        assertFalse(overflowButton.isVisible());
    }
```

I suggest to place this test in `ToolBarSkinTest` (not in `ToolBarTest`), because it tests the implementation of the skin and not the implementation of the control.

andy-goryachev-oracle:
> Here's a test

Many thanks!  The key difference seems to be the binding of the scale properties.

mstr2 (approved):

andy-goryachev-oracle:
@kevinrushforth please take a look

kevinrushforth (approved):
Looks good. I ran both the test case attached to the bug report and the new unit test. As expected they fail without the fix and pass with the fix.

PR #1857, opened 2025-07-25
8364088: ToolBarSkin: NPE in select()

andy-goryachev-oracle:
Fixed an NPE in the ToolBar, looks like a corner case.

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ToolBarSkin.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


mstr2:
Is this unit-testable?

andy-goryachev-oracle:
> Is this unit-testable?

I would say it's probably not worth it, but I'll try come up with a headless test.  I certainly don't want a headful test for what looks like a programmer's oversight.

azuev-java (approved):
Change looks good but just for my curiosity - how do we get to this point? Any minimal test case?

andy-goryachev-oracle:
> Is this unit-testable?
> Any minimal test case?

Github ate my previous response!  Restoring from memory.

The bug came out of testing JDK-8364049.  I could not create a headless unit test: it looks like the focus subsystem is involved which is not easily reproducible with a headless test.  I don't want to create a headful test for this particular scenario.

For the record, the test I tried to write:

```
    @Test
    public void testNPE_8364088() {
        BorderPane bp = new BorderPane();
        Button b = new Button("Create Schema Overflow");
        bp.setTop(new HBox(new ToolBar(b)));
        Stage stage = new Stage();
        stage.setScene(new Scene(bp, 600, 400));
        stage.setTitle(getClass().getSimpleName());
        stage.show();

        runLater(() -> {
            b.requestFocus();
            runLater(() -> {
                stage.setWidth(50);
                runLater(() -> {
                    stage.setWidth(40);
                    System.out.println(stage.getWidth());
                    stage.hide();
                });
            });
        });
    }
    
    private static void runLater(Runnable r) {
        AtomicInteger counter = new AtomicInteger(10);
        Platform.runLater(new Runnable() {
            @Override
            public void run() {
                if (counter.decrementAndGet() > 0) {
                    Platform.runLater(this);
                } else {
                    r.run();
                }
            }
        });
    }

```

andy-goryachev-oracle:
@kevinrushforth I think we should backport this to 25, along with #1856 . what do you think?
is this the right command: `/ backport :jfx25` ?

kevinrushforth:
How likely is this bug to occur? It's a safe enough fix that I don't object to backporting it if you think it's worth doing. As for the syntax, yes, you have it right.

andy-goryachev-oracle:
I just really dislike these small issues (and I too like to add random stuff to the toolbars).

/backport :jfx25

PR #1860, opened 2025-07-29
8355990: [macOS] Restoring a maximized stage does not update the window size

beldenfox:
This PR fixes numerous bugs in the handling of setMaximized() on macOS and also cleans up some issues seen when the user changes the maximized state manually.

- After setMaximized(false) a notifyResize event was never sent so the width and height tracked by JavaFX was wrong.

- During maximize and restore a series of notifyMove events were issued causing the maximized property to flip-flop during the animation.

- Transparent windows were being maximized by passing native screen coordinates to a routine that expected JavaFX screen coordinates so the Y axis was flipped and the window was positioned incorrectly.

- The restored frame is in native screen coordinates which was sent to a routine that expected flipped JavaFX coordinates. That routine corrected this by directly inspecting the call stack to see which routine was calling it.

- When the user maximizes or restores the window a notifyResize event was always sent immediately stating that the window was maximized even when it was not. Then a series of notifyMove events were issued causing the maximized flag to flip-flop during the animation.

This PR cleans all of this up. When using the setMaximized API only one notifyMove and one notifyResize event are sent and transparent windows are positioned correctly. When the user maximizes or restores manually we still see a series of notifyMove and notifyResize events but at least they all report the window’s state correctly.

System tests are currently being written as part of #1789. It’s hard to create a system test that catches things like the mis-alignment of maximized transparent windows so that needs to be tested manually. The reproducer attached to the bug report can be used to verify this and also to see what happens when the user toggles the maximization state manually.

Note that when the test programs tags something as “unexpected” it just means it saw a property change outside of any JavaFX API call. This is actually expected if the user manipulates the window manually.


kevinrushforth:
@beldenfox It might be better to close [JDK-8179980](https://bugs.openjdk.org/browse/JDK-8179980) as a duplicate of this bug (JDK-8355990) rather than add it as a separate issue, since it is a natural step when fixing the underlying product bug to re-enable any skipped tests.

modules/javafx.graphics/src/main/native-glass/mac/GlassWindow.m
```diff
@@ -1037,32 +1037,45 @@ static jlong _createWindowCommonDo(JNIEnv *env, jobject jWindow, jlong jOwnerPtr
...
 
-        if ((maximize == JNI_TRUE) && (isZoomed == JNI_FALSE))
-        {
-            window->preZoomedRect = [window->nsWindow frame];
+        NSRect oldFrame = window->nsWindow.frame;
+        int eventType = com_sun_glass_events_WindowEvent_RESTORE;
 
-            if ([window->nsWindow styleMask] != NSWindowStyleMaskBorderless)
-            {
+        if ((maximize == JNI_TRUE) && (isZoomed == JNI_FALSE)) {
```
andy-goryachev-oracle: changing this code to K&R style messes up the diff alignment...  could we retain the original style?

tests/system/src/test/java/test/javafx/stage/RestoreStagePositionTest.java
```diff
@@ -130,7 +127,7 @@ public void testDemaximizedPosition() throws Exception {
         double y = stage.getY();
 
         Platform.runLater(() -> stage.setMaximized(true));
-        Thread.sleep(200);
+        Thread.sleep(800);
```
andy-goryachev-oracle: Would the `test.util.Util.waitForIdle()` be of use here?  Or `runAndWait()` ?

kevinrushforth:
> @beldenfox It might be better to close [JDK-8179980](https://bugs.openjdk.org/browse/JDK-8179980) as a duplicate of this bug (JDK-8355990) rather than add it as a separate issue, since it is a natural step when fixing the underlying product bug to re-enable any skipped tests.

Thanks for closing [JDK-8179980](https://bugs.openjdk.org/browse/JDK-8179980) as a duplicate. Please remove it from the list of solved issues with `/issue remove 8179980` before you integrate.

tests/system/src/test/java/test/javafx/stage/RestoreStagePositionTest.java
```diff
@@ -119,18 +119,15 @@ public void testUnfullscreenPosition() throws Exception {
...
         Assertions.assertTrue(stage.isShowing());
         Assertions.assertFalse(stage.isMaximized());
 
         double x = stage.getX();
         double y = stage.getY();
 
-        Platform.runLater(() -> stage.setMaximized(true));
-        Thread.sleep(200);
+        Util.runAndWait(() -> stage.setMaximized(true));
+        Util.waitForIdle(stage.getScene());
```
kevinrushforth, line 130: I know that @andy-goryachev-oracle suggested its use, but for tests that do various windowing operations (maximize, full screen, restore, etc), `waitForIdle` is too short (its on the order of 160 msec). To make the test more robust, I recommend reverting this back to the `sleep(800)` you had earlier.

Arguably, we should have a utility method for this sort of "wait for windowing changes to settle down", but we don't.

kevinrushforth (commented):
I did some initial testing and ran a headful test through our CI system and it looks good. I left one comment on the test and will review the native glass code changes next.

tests/system/src/test/java/test/javafx/stage/RestoreStagePositionTest.java
kevinrushforth: Yes on the runAndWait (always a good idea in cases like this where you aren't otherwise latching), but no on the change to use waitForIdle (see my other comment).

tests/system/src/test/java/test/javafx/stage/RestoreStagePositionTest.java
andy-goryachev-oracle: There is `Util::waitForIdle(scene, pulseCount)`.

The real question is how to account for the platform-dependent transitions (which may take some time) reliably.  Waiting for a set amount of time might work, but we have to pick a good timeout value.  Is there any other possibility?

I am ok with 800ms - it feels like it should be long enough for any transition to take place.

beldenfox: Changes to the maximized state are synchronous on all platforms. By the time `runAndWait` returns the window should be maximized. On Mac and Windows the JavaFX window state properties should be up-to-date because the OS notifications are also synchronous. I added waitForIdle because on Linux the configure events might come in later but we only need to wait long enough for the event loop to receive the events. That should be far less than 800ms.

> The real question is how to account for the platform-dependent transitions (which may take some time) reliably.

I can't think of any window attribute change that performs an asynchronous transition (but maybe I'm forgetting one). runAndWait should be enough on Mac and Windows. On Linux you need to give the event loop a chance to spin a few times.

The only real issue is (was?) fullscreen on macOS. The wait isn't really for the transition (it's done by the time setFullScreen returns) but to avoid scheduling deferred Runnables that might execute during the transition. That's less of an issue since #1797 was integrated.

kevinrushforth: Yes, full-screen is the one I know does (or did) have some async behavior. If maximize is reliably synchronous then OK to keep the test the way you have it.

PR #1864, opened 2025-08-05
8364687 : Headless allow lowercase for -Dglass.platform

CodeSimcoe:
Basically allow -Dglass.platform=headless to enable headless, not only -Dglass.platform=Headless.
This makes it consistent with other processings (macosx, windows, linux, gtk and ios) as well was less error-prone for enabling it.

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/glass/ui/Platform.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/glass/ui/Platform.java
@@ -32,6 +32,7 @@ final class Platform {
     public static final String WINDOWS = "Win";
     public static final String GTK = "Gtk";
     public static final String IOS = "Ios";
+    public static final String HEADLESS = "Headless";
     public static final String UNKNOWN = "unknown";
 
     static private String type = null;
@@ -53,6 +54,8 @@ final class Platform {
                    type = GTK;
                 else if (userPlatform.equals("ios"))
                    type = IOS;
+                else if (userPlatform.equals("headless"))
+                    type = HEADLESS;
                 else
                    type = userPlatform;
                 return type;


kevinrushforth:
Reviewer: @johanvos

johanvos:
I'll check later, but I want to raise this already:
there are 2 places (in 2 different modules) where the platform checks are done: com.sun.glass.ui.platform does one (discussed here), but there is also com.sun.javafx.PlatformUtil in the javafx.base method, which now has a method `isHeadless` which returns true on either "headless" or "Headless".
I do understand the differences between the Platform.java and the PlatformUtil.java checks, but there is some overlap that I'm not sure how to address.

CodeSimcoe:
`com.sun.javafx.PlatformUtil::isHeadless` returns `true` for any case matching `"headless"`, which is not what i provided in my original suggestion.
So maybe we could reuse it :

```
            if (userPlatform != null) {
                if (userPlatform.equals("macosx"))
                    type = MAC;
                else if (userPlatform.equals("windows"))
                    type = WINDOWS;
                else if (userPlatform.equals("linux"))
                    type = GTK;
                else if (userPlatform.equals("gtk"))
                    type = GTK;
                else if (userPlatform.equals("ios"))
                    type = IOS;
                else
                    if (PlatformUtil.isHeadless())
                        type = HEADLESS;
                    else
                        type = userPlatform;
                return type;
            }
```

Makes the code a bit clumsy to me, but has the benefit of reusing logic.

johanvos:
Both your original code as the patch you mention above work. I'm unsure what is the best (most elegant and most consistent approach).

The code in `com.sun.glass.ui.Platform` first checks if the property *"glass.platform"* is set, and only if that is not set, there is a fallback to `PlatformUtil`. 

`PlatformUtil` relies mainly on the value of the System Property *"os.name"*. 

But I believe there is an inconsistency in `PlatformUtil`. Some of the code uses the "glass.platform" as well, and I think that goes against the idea that "glass.platform" is glass-specific, where glass is part of the `javafx.graphics` module, hence it should be used in the `javafx.graphics` module, and *not* in the `javafx.base` module.
That would require a major refactory, where the overlap between the javafx.base `PlafformUtil` and the javafx.graphics `Platform` is removed.
@kevinrushforth Do you have any thoughts about this?

Since the current proposed PR is simple and useful (using `headless` instead of `Headless` should work indeed), I am ok with this PR, unless there is an easy solution that provides more consistency?

kevinrushforth:
> But I believe there is an inconsistency in `PlatformUtil`. Some of the code uses the "glass.platform" as well, and I think that goes against the idea that "glass.platform" is glass-specific, where glass is part of the `javafx.graphics` module, hence it should be used in the `javafx.graphics` module, and _not_ in the `javafx.base` module. That would require a major refactory, where the overlap between the javafx.base `PlafformUtil` and the javafx.graphics `Platform` is removed. @kevinrushforth Do you have any thoughts about this?

Yes it does seems a bit odd for something in `javafx.base` to use a system property that conceptually belongs to `javafx.graphics`. For that matter, methods like `useEGL`,  `isEmbedded`, and `isHeadless` really don't belong in `javafx.base`. I checked and there are no references to `PlatformUtil` in `javafx.base` at all. Perhaps the whole class should move to `javafx.graphics` along with addressing the issue of having two inconsistent ways to do the same thing? That might be a reasonable clean-up bug for the future.

> Since the current proposed PR is simple and useful (using `headless` instead of `Headless` should work indeed), I am ok with this PR, unless there is an easy solution that provides more consistency?

I also think the current patch as it stands is good (and better than the second alternative).

johanvos:
I filed https://bugs.openjdk.org/browse/JDK-8365309 as a follow-up.

johanvos:
@CodeSimcoe Feel free to integrate this PR at your convenience.

PR #1868, opened 2025-08-13
8350479: SW pipeline should use default pipeline in Glass

jayathirthrao:
We have Metal pipeline also now in macOS apart from OpenGL pipeline for accelerated hardware pipelines.

Currently when Software pipeline is used we continue to use OpenGL(ES2) pipeline on glass side, which is the default hardware pipeline in macOS. If we want to switch to Metal pipeline as default hardware pipeline in prism, we should be able to use Metal pipeline in glass also when software pipeline is used. For this to happen we need to set some information when SWPipeline is getting initialized and pass that information to glass.

This PR adds new flag `useMTLInGlass` in device details when SWPipeline is getting initialized and then passes it to glass code. When SWPipeline is used, based on this flag we will use appropriate hardware pipeline in glass.

Patch is tested when useMTLInGlass is set to true/false and then using -Dprism.order=es2/mtl/sw and it picks proper GlassView/Layer.

---------
### Progress

### Issue
 * [JDK-8350479](https://bugs.openjdk.org/browse/JDK-8350479): SW pipeline should use default pipeline in Glass (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1868/head:pull/1868` \
`$ git checkout pull/1868`

`$ git checkout pull/1868` \
`$ git pull https://git.openjdk.org/jfx.git pull/1868/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1868`

`$ git pr show -t 1868`

</details>

<a href="https://git.openjdk.org/jfx/pull/1868.diff">https://git.openjdk.org/jfx/pull/1868.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1868#issuecomment-3183884499)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/prism/impl/PrismSettings.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/prism/impl/PrismSettings.java
@@ -50,7 +50,7 @@ public final class PrismSettings {
     public static final boolean cacheComplexShapes;
     public static final boolean useNewImageLoader;
     public static final List<String> tryOrder;
-    public static final String macDefaultPipeline;
+    public static final String defaultPipeline;
     public static final int prismStatFrequency;
     public static final RasterizerType rasterizerSpec;
     public static final String refType;
@@ -201,28 +201,26 @@ public final class PrismSettings {
         /* Force GPU, if GPU is PS 3 capable, disable GPU qualification check. */
         forceGPU = getBoolean(systemProperties, "prism.forceGPU", false);
 
-        String order = systemProperties.getProperty("prism.order");
         String[] tryOrderArr;
-        String[] macDefaultOrderArr = new String[] { "es2", "mtl", "sw" };
-        // macDefaultPipeline flag will be used to determine whether to
-        // use es2/mtl pipeline in glass, when sw pipeline is used in prism
-        macDefaultPipeline = macDefaultOrderArr[0];
+        if (PlatformUtil.isWindows()) {
+            tryOrderArr = new String[] { "d3d", "sw" };
+        } else if (PlatformUtil.isMac()) {
+            tryOrderArr = new String[] { "es2", "mtl", "sw" };
+        } else if (PlatformUtil.isIOS()) {
+            tryOrderArr = new String[] { "es2" };
+        } else if (PlatformUtil.isAndroid()) {
+                tryOrderArr = new String[] { "es2" };
+        } else if (PlatformUtil.isLinux()) {
+            tryOrderArr = new String[] { "es2", "sw" };
+        } else {
+            tryOrderArr = new String[] { "sw" };
+        }
+
+        defaultPipeline = tryOrderArr[0];
+
+        String order = systemProperties.getProperty("prism.order");
         if (order != null) {
             tryOrderArr = split(order, ",");
-        } else {
-            if (PlatformUtil.isWindows()) {
-                tryOrderArr = new String[] { "d3d", "sw" };
-            } else if (PlatformUtil.isMac()) {
-                tryOrderArr = macDefaultOrderArr;
-            } else if (PlatformUtil.isIOS()) {
-                tryOrderArr = new String[] { "es2" };
-            } else if (PlatformUtil.isAndroid()) {
-                    tryOrderArr = new String[] { "es2" };
-            } else if (PlatformUtil.isLinux()) {
-                tryOrderArr = new String[] { "es2", "sw" };
-            } else {
-                tryOrderArr = new String[] { "sw" };
-            }
         }
 
         tryOrder = List.of(tryOrderArr);
--- a/modules/javafx.graphics/src/main/java/com/sun/prism/impl/PrismSettings.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/prism/impl/PrismSettings.java
@@ -50,6 +50,7 @@ public final class PrismSettings {
     public static final boolean cacheComplexShapes;
     public static final boolean useNewImageLoader;
     public static final List<String> tryOrder;
+    public static final String macDefaultPipeline;
     public static final int prismStatFrequency;
     public static final RasterizerType rasterizerSpec;
     public static final String refType;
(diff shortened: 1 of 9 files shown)


modules/javafx.graphics/src/main/java/com/sun/prism/impl/PrismSettings.java
```diff
@@ -202,13 +203,17 @@ dirtyOptsEnabled && getBoolean(systemProperties,
 
         String order = systemProperties.getProperty("prism.order");
         String[] tryOrderArr;
+        String[] macDefaultOrderArr = new String[] { "es2", "mtl", "sw" };
+        // macDefaultPipeline flag will be used to determine whether to
+        // use es2/mtl pipeline in glass, when sw pipeline is used in prism
+        macDefaultPipeline = macDefaultOrderArr[0];
```
Maran23: Since this is a platform independent file, should we name this property in a generic way (even if it is only used for MacOS)?
Something like `defaultPipeline?`

kevinrushforth: Since `macDefaultOrderArr` is already mac-specific, I'm not sure how much this helps without also refactoring the rest of the logic.

One possibility would be to move the if-then-else block up (before the check for the `"prism.order"` system property) and then assign the default pipeline after the tryOrder is assigned. I'll leave it up to Jay as whether he wants to do this.

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWPipeline.java
```diff
@@ -69,6 +73,15 @@ public int getAdapterOrdinal(Screen screen) {
         if (factory == null) {
             factory = new SWResourceFactory(screen);
             factories.put(index, factory);
+            if (PlatformUtil.isMac()) {
+                HashMap devDetails =
+                    (HashMap)SWPipeline.getInstance().getDeviceDetails();
```
kevinrushforth: You're already in an instance method of SWPipeline, so the `SWPipeline.getInstance().` is not needed, and seems odd.

modules/javafx.graphics/src/main/native-glass/mac/GlassLayer.m
```diff
@@ -45,18 +45,22 @@ - (id)initGlassLayer:(NSObject*)ctx
...
         withIsSwPipe:(BOOL)isSwPipe
+       useMTLForBlit:(BOOL)useMTLInGlass
 {
     LOG("GlassLayer initGlassLayer]");
     self = [super init];
     if (self != nil)
     {
-        if (mtlCommandQueuePtr != 0l) { // MTL
+        if (mtlCommandQueuePtr != 0l ||
+            useMTLInGlass) { // MTL
```
kevinrushforth: Minor: perhaps this would look better with the "if" test all on the same line?

modules/javafx.graphics/src/main/native-glass/mac/GlassLayer.h
```diff
@@ -36,7 +36,8 @@
     andClientContext:(NSObject*)clCtx
          mtlQueuePtr:(long)mtlCommandQueuePtr
       withHiDPIAware:(BOOL)HiDPIAware
-        withIsSwPipe:(BOOL)isSwPipe;
+        withIsSwPipe:(BOOL)isSwPipe
+       useMTLForBlit:(BOOL)useMTLInGlass;
```
kevinrushforth: Minor question on the naming: Since this is only ever set to true when using the SW pipeline, do you think having "Sw" somewhere in the name would be helpful? Given the name, it could be confusing to notice that it is set to false when using the MTL pipeline.

modules/javafx.graphics/src/main/native-glass/mac/GlassViewCGL.m
```diff
@@ -173,7 +174,8 @@ - (void)_initialize3dWithJproperties:(jobject)jproperties
                                     andClientContext:(NSObject*)clientCGL
                                          mtlQueuePtr:0l
                                       withHiDPIAware:isHiDPIAware
-                                        withIsSwPipe:isSwPipe];
+                                        withIsSwPipe:isSwPipe
+                                       useMTLForBlit:useMTLInGlass];
```
kevinrushforth: Minor: indentation isn't consistent.

kevinrushforth (approved):
Looks good with a few minor comments. I'll reapprove if you make any changes as a result.

kevinrushforth:
@arapte can you be the second reviewer?

modules/javafx.graphics/src/main/java/com/sun/prism/impl/PrismSettings.java
jayathirthrao: if-then-else block requires "order" value to determine whether user is explicitly setting any value.

In case where user is explicitly setting `-Dprism.order=sw`, tryOrderArr will have only "sw" value. So we need macOS prism default order values outside of the if-then-else block.

Since we need this mac specific default order values, i think its okay to keep this mac specific flag.
I would like to keep the code as it is.

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWPipeline.java
jayathirthrao: Thanks, updated.

modules/javafx.graphics/src/main/native-glass/mac/GlassLayer.m
jayathirthrao: Sure updated.

modules/javafx.graphics/src/main/native-glass/mac/GlassLayer.h
jayathirthrao: I also had doubts about how to include "SW" in the name. Because it would be very long name.

I have removed the references to "Glass" in the name when we are already in glass code and updated these flags to use "SW" in the name.

modules/javafx.graphics/src/main/native-glass/mac/GlassViewCGL.m
jayathirthrao: We are following https://google.github.io/styleguide/objcguide.html#method-invocations & https://google.github.io/styleguide/objcguide.html#method-declarations-and-definitions. So i have also updated the code where it is inconsistent and we are touching in this PR.

Please correct me if i am wrong.

modules/javafx.graphics/src/main/java/com/sun/prism/impl/PrismSettings.java
jayathirthrao: Looks like we are talking about `if-then-else-if` block. That will work and it will help us to remove mac specific reference also. Will go ahead and make that change.

modules/javafx.graphics/src/main/native-glass/mac/GlassViewCGL.m
kevinrushforth: OK

PR #1869, opened 2025-08-13
8322486: ColorPicker: blurry popup

Ziad-Mid:
The issue occurs when displaying the ColorPicker on a second monitor with scale=1 the ColorPicker appears blurry,
This is due to snapping and changing the value of `-fx-snap-to-pixel` for the `ColorPalette` fixes the issue


modules/javafx.controls/src/main/resources/com/sun/javafx/scene/control/skin/modena/modena.css
```diff
@@ -1979,7 +1979,7 @@ is being used to size a border should also be in pixels.
 }
 .color-palette > .color-picker-grid {
     -fx-padding: 0.5px;
-    -fx-snap-to-pixel: false;
+    -fx-snap-to-pixel: true;
```
Maran23: I guess we can completely remove this line then?

kevinrushforth:

kevinrushforth:
`.color-picker-grid` is the only place in `modena.css` that sets snap-to-pixel to false. It was set to false during the initial development of `modena.css` in commit ee4c8a0f95e6148772b0839c95e4348ac8f9e3e0 without any explanation. I also note that the color picker is the only place where a fractional pixel padding is used, so this will need to be tested and carefully inspected visually with different scales to make sure that it looks good after being snapped (with no gaps or visible asymmetry).

If all looks OK, then the change seems reasonable (although @Maran23 is right that it should simply be removed), but I want to know what @andy-goryachev-oracle thinks.

modules/javafx.controls/src/main/resources/com/sun/javafx/scene/control/skin/modena/modena.css
andy-goryachev-oracle: Good point - I wonder if the whole block can be removed.
We need to test this on windows with fractional scale though.

andy-goryachev-oracle:
Good job finding the culprit, @Ziad-Mid !

andy-goryachev-oracle:
I've tested with the _whole selector removed_, on Windows at scales 100%, 125%, 150%, and 175%.

I think it looks good:

Screenshot 2025-08-18 160538
Screenshot 2025-08-18 160627
Screenshot 2025-08-18 160703
Screenshot 2025-08-18 160748

modules/javafx.controls/src/main/resources/com/sun/javafx/scene/control/skin/modena/modena.css
andy-goryachev-oracle: I suggest we remove the whole thing `.color-palette > .color-picker-grid { ... }`

andy-goryachev-oracle:
/reviewers 2

I'd like someone else to look at this on Linux please (looks good to me on macOS/Windows)

andy-goryachev-oracle (approved):

Ziad-Mid:
@arapte could you be second reviewer ?

Maran23 (approved):
Looks good to me too!
But you may still want to wait for someone testing this on Linux (I did not).

andy-goryachev-oracle:
@lukostyra could you take a look how it works on linux please?

lukostyra:
LGTM, I can confirm on Ubuntu 24.04.2 LTS VM this fixes the blurriness of ColorPicker.

Ziad-Mid:
Thank you all for the reviews

PR #1870, opened 2025-08-14
8364547: Window size may be incorrect when constrained to min or max

beldenfox:
When changing the width and height of a window the platform code is responsible for enforcing the min and max size constraints. If the final width and height don't match the width and height passed into setBounds the platform needs to call notifyResize to correct the window's properties. This happens naturally if the window size actually changes since that will trigger the OS to send size change notifications. If the platform window size doesn't change the OS notifications won't trigger. We need to catch that case and send notifyResize anyway.

This PR does this for Mac and Windows. Linux is being handled in PR #1789 which also includes the system tests for these bugs.

---------
### Progress

### Issue
 * [JDK-8364547](https://bugs.openjdk.org/browse/JDK-8364547): Window size may be incorrect when constrained to min or max (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1870/head:pull/1870` \
`$ git checkout pull/1870`

`$ git checkout pull/1870` \
`$ git pull https://git.openjdk.org/jfx.git pull/1870/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1870`

`$ git pr show -t 1870`

</details>

<a href="https://git.openjdk.org/jfx/pull/1870.diff">https://git.openjdk.org/jfx/pull/1870.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1870#issuecomment-3189088984)
</details>


modules/javafx.graphics/src/main/java/com/sun/glass/ui/win/WinWindow.java
```diff
@@ -151,6 +155,10 @@ public void setBounds(float x, float y, boolean xSet, boolean ySet,
             pfReqWidth = (int) Math.ceil(fxReqWidth * platformScaleX);
             pfReqHeight = (int) Math.ceil(fxReqHeight * platformScaleY);
             _setBounds(getRawHandle(), px, py, xSet, ySet, pw, ph, 0, 0, xGravity, yGravity);
+
+            if (minMaxEnforced) {
+                notifyResize(WindowEvent.RESIZE, pw, ph);
+            }
```
kevinrushforth: Could this lead to two resize commands in some cases? I can see why this new logic is needed for the case where the window was already at max (or min) width and height, but if it wasn't already constrained, wouldn't the resize event have already happened?

tsayao: The issue is that the Java side updates the property first, and only afterward requests the Glass native side to apply the change. I've found this to cause many problems and I fixed the same way in  #1789 - when the change can't be applied, it notifies back.

beldenfox: > Could this lead to two resize commands in some cases?

Yes, I was trying not to be too clever with my checks. I figured if the second notification wasn't necessary it would be benign. At the very least it won't trigger invalidation of the window's width and height properties.

I will tighten this up since I have to tweak the code a bit anyway. I just verified that on Windows you can alter the size of a maximized window and the OS will keep it in the MAXIMIZED state (as it resizes glass calls notifyResize with WindowEvent.MAXIMIZED). This PR can kick the window out of the MAXIMIZED or MINIMIZED state incorrectly. Unfortunately it puts the window in the wrong internal state without updating the maximized or iconified properties so it's not easy to write a test to detect this.

kevinrushforth: This is pending work from you, right? I'll put it back on my review queue once you make the changes you mentioned above.

beldenfox: I've made the changes. It should be ready for review.

kevinrushforth: I completely missed that you had already done this. I'll review it then.

kevinrushforth (approved):
The fix looks good.

Windows: I ran `SizingTest` from PR #1789 (removing the `assumeTrue` that skips the test on Windows), and it fails for me without this fix and passes with this fix.

macOS: I ran `SizingTest` from PR #1789, and I get 8 test failures without this fix and 4 failures with this fix. The 4 failures are all using `StageStyle.EXTENDED`, which might be a different bug.

@beldenfox Can you merge in the latest master (I did when testing and there were no issues)? And can you coment on the `StageStyle.EXTENDED` failures on macOS? I think it would be fine to treat them as a separate bug unless it is easy to fix.

I'll reapprove if you make changes.

kevinrushforth:
@lukostyra or @andy-goryachev-oracle Can one of you be the second reviewer?

beldenfox:
Thanks for the review. I'm traveling so won't be able to pick this up again until November.

andy-goryachev-oracle (approved):
tested on macOS 15.7.1 M1
the reproducer works
found no ill effects in the monkey tester.

beldenfox:
There was a request to merge in the latest master but this PR is also marked as ready. Should I go ahead with the merge?

The SizingTest failures on macOS with StageStyle.EXTENDED are due to some weirdness in the OS. When we set the minimum size on an EXTENDED window the OS simply adds 16 units to the minimum height (I can verify this by writing the minSize and then immediately reading it back). I'm still trying to figure out where this magic number comes from. For EXTENDED we add a toolbar but the minimum size of a toolbar is 28 units. In any case I think this is a separate issue. I'll do a bit more research and file a bug.

andy-goryachev-oracle:
there are no merge conflicts, you can integrate.

kevinrushforth:
A primary reason for merging master in a PR whose source branch is out of date is to catch any problems that might arise due to semantic conflicts. Skara and GitHub do a pretty good job of catching merge conflicts, but there can be semantically conflicting changes even if there are no git merge conflicts.

This is 70 commits behind, so merging master is a good idea. It will trigger a GHA run, which isn't conclusive, but is at least a first pass indicator. If it is a clean merge, there will be no need to re-review.

Having said that, I did my testing using a branch in my repo where I merged in master as of two weeks ago (I always do that when I run CI tests) and didn't see any problems, so if you're confident that there won't be a problem, you can integrate.

PR #1873, opened 2025-08-16
8365635: Add MOUSE_DRAG_DONE event type

nlisker:
Adds `MOUSE_DRAG_DONE` event type to `MouseDragEvent` and appropriate handlers.

/reviewers 2
/csr

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -7707,6 +7707,23 @@ public abstract sealed class Node
         return getEventHandlerProperties().onMouseDragExitedProperty();
     }
 
+    public final void setOnMouseDragDone(EventHandler<? super MouseDragEvent> value) {
+        onMouseDragDoneProperty().set(value);
+    }
+
+    public final EventHandler<? super MouseDragEvent> getOnMouseDragDone() {
+        return (eventHandlerProperties == null) ? null : eventHandlerProperties.getOnMouseDragDone();
+    }
+
+    /// Defines a function to be called when a full press-drag-release gesture ends with this node as its source.
+    ///
+    /// @return the event handler that is called when a full press-drag-release finishes
+    /// @see MouseDragEvent#MOUSE_DRAG_DONE
+    /// @since 26
+    public final ObjectProperty<EventHandler<? super MouseDragEvent>> onMouseDragDoneProperty() {
+        return getEventHandlerProperties().onMouseDragDoneProperty();
+    }
+
 
     /* *************************************************************************
      *                                                                         *
--- a/modules/javafx.graphics/src/main/java/javafx/scene/input/MouseDragEvent.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/input/MouseDragEvent.java
@@ -66,14 +66,13 @@ public final class MouseDragEvent extends MouseEvent{
             new EventType<>(MouseEvent.ANY, "MOUSE-DRAG");
 
     /**
-     * This event occurs when the gesture progresses within this node.
+     * This event occurs when the gesture progresses within a node.
      */
     public static final EventType<MouseDragEvent> MOUSE_DRAG_OVER =
             new EventType<>(MouseDragEvent.ANY, "MOUSE-DRAG_OVER");
 
     /**
-     * This event occurs when the gesture ends (by releasing mouse button)
-     * on this node.
+     * This event occurs when the gesture is released over a node (by releasing the mouse button).
      */
     public static final EventType<MouseDragEvent> MOUSE_DRAG_RELEASED =
             new EventType<>(MouseDragEvent.ANY, "MOUSE-DRAG_RELEASED");
@@ -127,6 +126,11 @@ public final class MouseDragEvent extends MouseEvent{
     public static final EventType<MouseDragEvent> MOUSE_DRAG_EXITED =
             new EventType<>(MouseDragEvent.MOUSE_DRAG_EXITED_TARGET,
                     "MOUSE-DRAG_EXITED");
+
+    /// This event occurs when the gesture ends. It is delivered exactly once to the source node/scene, and is always the
+    /// last event in the full press-drag-release process. It is delivered even when the mouse is outside the application.
+    public static final EventType<MouseDragEvent> MOUSE_DRAG_DONE =
+            new EventType<>(MouseDragEvent.ANY, "MOUSE-DRAG_DONE");
 
     /**
      * Constructs new MouseDragEvent event.
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/scene/EventHandlerProperties.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/scene/EventHandlerProperties.java
@@ -497,6 +497,21 @@ public final class EventHandlerProperties {
         return onMouseDragExited;
     }
 
+
+    private EventHandlerProperty<MouseDragEvent> onMouseDragDone;
+
+    public final EventHandler<? super MouseDragEvent> getOnMouseDragDone() {
+        return (onMouseDragDone == null) ? null : onMouseDragDone.get();
+    }
+
+    public ObjectProperty<EventHandler<? super MouseDragEvent>> onMouseDragDoneProperty() {
+        if (onMouseDragDone == null) {
+            onMouseDragDone = new EventHandlerProperty<>(bean, "onMouseDragDone", MouseDragEvent.MOUSE_DRAG_DONE);
+        }
+        return onMouseDragDone;
+    }
+
+
     private EventHandlerProperty<KeyEvent> onKeyPressed;
 
     public final EventHandler<? super KeyEvent> getOnKeyPressed() {
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
@@ -4169,6 +4169,12 @@ public class Scene implements EventTarget {
                             fullPDRSource, result));
                 }
             }
+            if (e.getEventType() == MouseEvent.MOUSE_RELEASED) {
+                Event.fireEvent(fullPDRSource, MouseEvent.copyForMouseDragEvent(e,
+                        fullPDRSource, eventTarget,
+                        MouseDragEvent.MOUSE_DRAG_DONE,
+                        fullPDRSource, result));
+            }
         }
 
         private void updateCursor(Cursor newCursor) {
@@ -4732,6 +4738,27 @@ public class Scene implements EventTarget {
         return onMouseDragExited;
     }
 
+
+    /// Defines a function to be called when a full press-drag-release gesture ends.
+    ///
+    /// @see MouseDragEvent#MOUSE_DRAG_DONE
+    /// @since 26
+    private ObjectProperty<EventHandler<? super MouseDragEvent>> onMouseDragDone;
+
+    public final void setOnMouseDragDone(EventHandler<? super MouseDragEvent> value) {
+        onMouseDragDoneProperty().set(value);
+    }
+
+    public final EventHandler<? super MouseDragEvent> getOnMouseDragDone() {
+        return onMouseDragDone == null ? null : onMouseDragDone.get();
+    }
+
+    public final ObjectProperty<EventHandler<? super MouseDragEvent>> onMouseDragDoneProperty() {
+        if (onMouseDragDone == null) {
+            onMouseDragDone = new EventHandlerProperty<>("onMouseDragDone", MouseDragEvent.MOUSE_DRAG_DONE);
+        }
+        return onMouseDragDone;
+    }
 
     /* *************************************************************************
      *                                                                         *
--- a/modules/javafx.graphics/src/test/java/test/javafx/scene/input/MouseDragEventTest.java
+++ b/modules/javafx.graphics/src/test/java/test/javafx/scene/input/MouseDragEventTest.java
@@ -478,7 +478,7 @@ public class MouseDragEventTest {
     }
 
     @Test
-    public void endingFullPDRShouldProduceReleaseEvents() {
+    public void endingFullPDRShouldProduceReleaseAndDoneEvents() {
         World w = new World(false, false);
 
         // Initiate the gesture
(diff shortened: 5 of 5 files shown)


nlisker:
I noticed that the `@see` documentation tag doesn't work on properties. I'm not sure if it's a bug in javadoc.

nlisker:
Forgot to attach a small manual test application I used for testing full DPR and DnD. Might be helpful for reviewers.
```java
package main;

import javafx.scene.Node;
import javafx.scene.Scene;
import javafx.scene.input.ClipboardContent;
import javafx.scene.input.Dragboard;
import javafx.scene.input.TransferMode;
import javafx.scene.paint.Color;
import javafx.stage.Stage;

public class MouseTest extends Application {

        launch(args);
    }

    @Override
    public void start(@SuppressWarnings("exports") Stage stage) throws Exception {
        var sourceMouse = createSourceMouse();
        var sourceDrag = createSourceDrag();
        var target = createTarget();

        var scene = createScene(sourceMouse, sourceDrag, target);
        stage.setScene(scene);
        stage.show();
    }

    private Rectangle createSourceMouse() {
        var sourceMouse = new Rectangle(50, 50);
        sourceMouse.setId("sourceMouse");
        sourceMouse.setFill(Color.RED);
        sourceMouse.setOnDragDetected(e -> {
            sourceMouse.startFullDrag();
            System.out.println(e.getEventType() + " " + sourceMouse.getId());
            e.consume();
        });
        addDragListeners(sourceMouse);
        return sourceMouse;
    }

    private Rectangle createSourceDrag() {
        var sourceDrag = new Rectangle(50, 50);
        sourceDrag.setId("sourceDrag");
        sourceDrag.setFill(Color.GREEN);
        sourceDrag.setOnDragDetected(e -> {
            Dragboard db = sourceDrag.startDragAndDrop(TransferMode.ANY);
            var content = new ClipboardContent();
            content.putString("");
            db.setContent(content);
            System.out.println(e.getEventType() + " " + sourceDrag.getId());
            e.consume();
        });
        addDragListeners(sourceDrag);
        return sourceDrag;
    }

    private Rectangle createTarget() {
        var target = new Rectangle(100, 50);
        target.setId("target");
        target.setFill(Color.BLUE);
        target.setOnDragDetected(MouseEvent::consume);
        addDragListeners(target);
        return target;
    }

    private Scene createScene(Rectangle sourceMouse, Rectangle sourceDrag, Rectangle target) {
        var scene = new Scene(new VBox(10, new HBox(sourceMouse, sourceDrag), target), 300, 200);
        scene.setOnDragDetected(e -> {
            scene.startFullDrag();
            System.out.println(e.getEventType() + " scene");
        });

        scene.setOnDragDone(e -> {
            e.acceptTransferModes(TransferMode.ANY);
            System.out.println(e.getEventType() + " scene");
        });
        scene.setOnMouseDragReleased(e -> System.out.println(e.getEventType() + " scene"));
        scene.setOnMouseDragDone(e -> System.out.println(e.getEventType() + " scene"));
        return scene;
    }

    private static void addDragListeners(Node node) {
        node.setOnDragOver(e -> {
            e.acceptTransferModes(TransferMode.ANY);
            System.out.println(e.getEventType() + " " + node.getId());
        });
        node.setOnMouseDragOver(e -> System.out.println(e.getEventType() + " " + node.getId()));

        node.setOnDragDropped(e -> {
            e.acceptTransferModes(TransferMode.ANY);
            e.setDropCompleted(true);
            System.out.println(e.getEventType() + " " + node.getId());
        });
        node.setOnMouseDragReleased(e -> System.out.println(e.getEventType() + " " + node.getId()));

        node.setOnDragDone(e -> {
            e.acceptTransferModes(TransferMode.ANY);
            System.out.println(e.getEventType() + " " + node.getId());
        });
        node.setOnMouseDragDone(e -> System.out.println(e.getEventType() + " " + node.getId()));
    }
}
```

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -7707,6 +7707,23 @@ public final EventHandler<? super MouseDragEvent> getOnMouseDragExited() {
...
 
+    public final void setOnMouseDragDone(EventHandler<? super MouseDragEvent> value) {
+        onMouseDragDoneProperty().set(value);
+    }
+
+    public final EventHandler<? super MouseDragEvent> getOnMouseDragDone() {
+        return (eventHandlerProperties == null) ? null : eventHandlerProperties.getOnMouseDragDone();
+    }
+
+    /// Defines a function to be called when a full press-drag-release gesture ends with this node as its source.
```
mstr2: This is the first instance of markdown-style javadoc comments in JavaFX. Do we need a discussion surrounding their use?

nlisker: They were added in JDK 23, which we can use. I don't know if there's anything special to discuss. The JEP gives standard usage examples, which I've followed, so there's no "feature abuse" here either.

mstr2: Well, they look very different from the other style. Usually matters of style are settled once in a style guide, which we don't have. But I think we should at least consciously agree that we're fine with having two different styles of documentation in the same project. Personally, I value uniformity of style more than the (little) added value that markdown comments bring.

arapte: +1 @mstr2 ..
The markdown style is quite different than traditional HTML base style. With increased use of markdown, the comments will look very different. We should keep the uniformity, unless we want to use any feature from markdown style that is not available in traditional style.

andy-goryachev-oracle: I think markdown style is ok, since it results in a more readable source.

The support in other tools might be lacking initially (for example, Eclipse does not refresh its javadoc pane with markdown as it does with the usual style)

modules/javafx.graphics/src/main/java/javafx/scene/input/MouseDragEvent.java
```diff
@@ -128,6 +127,11 @@ public final class MouseDragEvent extends MouseEvent{
             new EventType<>(MouseDragEvent.MOUSE_DRAG_EXITED_TARGET,
                     "MOUSE-DRAG_EXITED");
 
+    /// This event occurs when the gesture ends. It is delivered exactly once to the source node/scene, and is always the
+    /// last event in the full press-drag-release process. It is delivered even when the mouse is outside the application.
+    public static final EventType<MouseDragEvent> MOUSE_DRAG_DONE =
```
andy-goryachev-oracle, line 136: `@since 26` ?

andy-goryachev-oracle (commented):
I must be doing something wrong - I don't see MouseDragEvent.MOUSE_DRAG_DONE, or in fact MouseDragEvent.ANY sent to the listeners, with this version of the monkey tester:

https://github.com/andy-goryachev-oracle/MonkeyTest/tree/mouse.drag.done

The test page in question is
https://github.com/andy-goryachev-oracle/MonkeyTest/blob/mouse.drag.done/src/com/oracle/tools/fx/monkey/pages/DnDPage.java

nlisker:
> I must be doing something wrong - I don't see MouseDragEvent.MOUSE_DRAG_DONE, or in fact MouseDragEvent.ANY sent to the listeners, with this version of the monkey tester:
> 
> https://github.com/andy-goryachev-oracle/MonkeyTest/tree/mouse.drag.done
> 
> The test page in question is https://github.com/andy-goryachev-oracle/MonkeyTest/blob/mouse.drag.done/src/com/oracle/tools/fx/monkey/pages/DnDPage.java

Did you try the sample program I posted?

andy-goryachev-oracle:
> Did you try the sample program

no, intentionally.

nlisker:
In https://github.com/andy-goryachev-oracle/MonkeyTest/blob/mouse.drag.done/src/com/oracle/tools/fx/monkey/pages/DnDPage.java, I don't see you starting a press-drag-release process, only a DnD one.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
nlisker: > Eclipse does not refresh its javadoc pane with markdown as it does with the usual style

Was fixed months ago in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2051. Not sure if a stable version was released yet.

andy-goryachev-oracle:
Yes, calling `startFullDrag()` did the trick:

```
{event=MOUSE_DRAGGED, x/y=(-785, -106), screen=(19, 57), scene=-360, -67)}
{event=MOUSE-DRAG_DONE, x/y=(-785, -106), screen=(19, 57), scene=-360, -67)}
{event=MOUSE_RELEASED, x/y=(-785, -106), screen=(19, 57), scene=-360, -67)}
```

Question: should we explain in greater detail that the new event is coming during full drag only?

andy-goryachev-oracle:
This is only tangentially related, but why in the drag and drop mode the coordinates for DRAG_DONE event are completely bogus when drop happens over a non-accepting target?

```
{event=DRAG_EXITED, x/y=(52, 32), screen=(1022, -1340), scene=(477, 71)}
{event=DRAG_EXITED_TARGET, x/y=(52, 32), screen=(1022, -1340), scene=(477, 71)}
{event=DRAG_DONE, x/y=(0, 0), screen=(0, 0), scene=(0, 0)}
```

nlisker:
> Question: should we explain in greater detail that the new event is coming during full drag only?

It's an event type of `MouseDragEvent`, which is the event used in full PDR, so I don't think there's an ambiguity here. I have a future docs change in the works around mouse events that clears up a lot of things.

nlisker:
> This is only tangentially related, but why in the drag and drop mode the coordinates for DRAG_DONE event are completely bogus when drop happens over a non-accepting target?

I don't know why it was done this way, but the coordinates are explicitly set to 0 from `GlassScene.createDragboard` and `GlassSceneDnDEventHandler.handleDragEnd`.

andy-goryachev-oracle:
> I have a future docs change in the works around mouse events that clears up a lot of things.

Is it going to be a separate PR or a part of this one?

nlisker:
> > I have a future docs change in the works around mouse events that clears up a lot of things.
> 
> Is it going to be a separate PR or a part of this one?

Separate. It will encompass mouse events and drag events, though I hit an obstacle with testing touch events.

hjohn (approved):

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
kevinrushforth: I started a (belated) [discussion on this on the mailing list](https://mail.openjdk.org/pipermail/openjfx-dev/2025-October/056578.html). As mentioned in that message, I think using markdown comments is fine for new APIs and docs that are modified in a way that markdown would be helpful.

kevinrushforth: After reading Michael and John's comments, I informally polled a few folks on the core libs team. I amend my proposal to include a recommendation to avoid mixing styles in a class like Node where there are already a lot of existing properties and you are adding a new one.

For this PR, there doesn't seem to be a good reason to use markdown-style doc comments. I recommend using old-style `/**` block comments for consistency.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -7707,6 +7707,23 @@ public final EventHandler<? super MouseDragEvent> getOnMouseDragExited() {
         return getEventHandlerProperties().onMouseDragExitedProperty();
     }
 
+    public final void setOnMouseDragDone(EventHandler<? super MouseDragEvent> value) {
+        onMouseDragDoneProperty().set(value);
+    }
+
+    public final EventHandler<? super MouseDragEvent> getOnMouseDragDone() {
```
kevinrushforth, line 7714: We normally put the property methods that will inherit the docs below the method (or private property as the case may be) that has the javadoc comments.

I recommend to move these two methods below the `onMouseDragDoneProperty` method.

nlisker: All of the properties above it have the same arrangement.

modules/javafx.graphics/src/main/java/javafx/scene/input/MouseDragEvent.java
```diff
@@ -128,6 +127,15 @@ public final class MouseDragEvent extends MouseEvent{
             new EventType<>(MouseDragEvent.MOUSE_DRAG_EXITED_TARGET,
                     "MOUSE-DRAG_EXITED");
 
+    /**
+     * This event occurs when the gesture ends. It is delivered exactly once to the source node/scene, and is always the
+     * last event in the full press-drag-release process. It is delivered even when the mouse is outside the application.
```
mstr2: 1. `It is delivered exactly once to the source node/scene`: you could probably remove `/scene`, it is implied by the way events work. Maybe be a bit more precise: `It is delivered exactly once to the node on which the press-drag-release gesture was started`.
2. `outside the application` is a bit vague, as applications don't have a spatial extent. How about: `It is delivered even when the mouse was released outside of the window`.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
kevinrushforth: Yes, you're right.

modules/javafx.graphics/src/main/java/javafx/scene/input/MouseDragEvent.java
nlisker: 1. I'm not sure why scene is implied from node considering it's not a node - the language often used is "gesture source". This is also why I don't understand why remove that term and replace it with its meaning.
2. Changed.

mstr2: Why do you use the term "node" then, if it shouldn't mean `Node`?

`Scene` is always implied, as you can't receive input events for event targets that are not part of a scene graph.

Maybe use "gesture source", if that's what is often used here. The wording "node/scene" throws me off: is it one, or the other, or both? It's unnecessarily vague.

nlisker: Changed to "gesture source".

kevinrushforth:
@nlisker I'll review the CSR after you've updated the spec to match the latest doc changes.

nlisker:
> @nlisker I'll review the CSR after you've updated the spec to match the latest doc changes.

Updated.

nlisker:
Thanks for the reviews everyone.

PR #1874, opened 2025-08-17
JDK-8365637 Unmanaged nodes are not added to the Scene's dirty layout list

hjohn:
This PR adds the necessary logic to automatically relayout unmanaged nodes when they're dirty as specified by the documentation. Relevant sections (emphasis mine):

### Parent::requestLayout

> If this parent is either a layout root or unmanaged, then it will be added directly to the **scene's dirty layout list**, otherwise requestParentLayout will be invoked.

### Node::managedProperty

> If a managed node's layoutBounds changes, it will **automatically trigger relayout** up the scene-graph to the **nearest layout root**

And:

> If an **unmanaged node** is of type `Parent`, it will act as a **"layout root"**

From this I conclude that:
- A layout root is any node that is unmanaged (and of type `Parent`), or a Scene's root node
- Changes requiring layout in managed children that have an unmanaged parent somewhere in their ancestry should cause layout to be triggered for the affected areas
- Automatic relayout is triggered either by scheduling a pulse (for scene roots) or by adding a layout root to the Scene's dirty layout list

Note that Scene currently does not have a dirty layout list.

Note also that the documentation for `requestLayout` probably meant to say "**scene** root or unmanaged" instead of "**layout** root or unmanaged."

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -8651,6 +8651,12 @@ public abstract sealed class Node
     }
 
     void markDirtyLayoutBranch() {
+        if (this instanceof Parent p && p.layoutRoot) {  // is this node a layout root?
+            markAsDirtyLayoutRoot(p);
+
+            return;  // No need to propagate flags any further up the hierarchy
+        }
+
         Parent p = getParent();
         while (p != null && p.layoutFlag == LayoutFlags.CLEAN) {
             p.setLayoutFlag(LayoutFlags.DIRTY_BRANCH);
@@ -8660,9 +8666,23 @@ public abstract sealed class Node
                     getSubScene().setDirtyLayout(p);
                 }
             }
+            else if (p.layoutRoot) {
+                markAsDirtyLayoutRoot(p);
+
+                return;  // No need to propagate flags any further up the hierarchy
+            }
+
             p = p.getParent();
         }
+    }
 
+    private void markAsDirtyLayoutRoot(Parent p) {
+        Toolkit.getToolkit().requestNextPulse();
+        Scene s = getScene();
+
+        if (s != null) {
+            s.addToDirtyLayoutList(p);
+        }
     }
 
     private boolean isWindowShowing() {
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
@@ -1021,7 +1021,7 @@ public abstract non-sealed class Parent extends Node {
      * rendered. This is batched up asynchronously to happen once per
      * "pulse", or frame of animation.
      * <p>
-     * If this parent is either a layout root or unmanaged, then it will be
+     * If this parent is either a scene root or unmanaged, then it will be
      * added directly to the scene's dirty layout list, otherwise requestParentLayout
      * will be invoked.
      * @since JavaFX 8.0
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
@@ -645,6 +645,21 @@ public class Scene implements EventTarget {
         clearInitialCssStateNodes.remove(node);
     }
 
+    /**
+     * A set of nodes that are layout roots and are in need of layout. Any unmanaged
+     * {@link Parent} is considered a layout root. If a child of such a node
+     * requests layout, the layout root will be added to this set, to be laid
+     * out on the next layout pass.
+     *
+     * @see Node#managedProperty()
+     * @see Parent#requestLayout()
+     */
+    private final Set<Parent> dirtyLayoutRoots = new HashSet<>();
+
+    void addToDirtyLayoutList(Parent layoutRoot) {
+        dirtyLayoutRoots.add(layoutRoot);
+    }
+
     void doLayoutPass() {
         if (peer != null) {
             peer.layoutOverlay();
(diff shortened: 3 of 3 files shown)


hjohn:
There is a tiny change in the documentation, but it doesn't change the meaning (a scene root is a layout root, and unmanaged nodes are layout roots; there are no other types of layout roots).  Can do a CSR though if deemed needed.

andy-goryachev-oracle:
@hjohn would it be possible to add a unit test?

andy-goryachev-oracle:
There is a single word change in the javadoc.  Do we need a CSR?

nlisker:
> There is a single word change in the javadoc. Do we need a CSR?

No. Even for larger doc fixes I never needed one.

nlisker:
If the node is unmanaged, why does it need a relayout in practice?

hjohn:
> If the node is unmanaged, why does it need a relayout in practice?

The unmanaged node does not need relayout, but its potentially managed children do.  Just like a Scene does not need layout, but its root does.  A node being unmanaged affects its **parent** (the parent must skip it for layout calculations), but does not free it from doing layout on its own children (if I interpret the documentation correctly).

Let's take this hierarchy:

    Region
         StackPane (unmanaged)
               ImageView

The Region will not lay out the stack pane as it is unmanaged, so I position the unmanaged child manually with `resizeRelocate`.  The StackPane now has a defined position and size.  When I add children to it, the StackPane's `layout` (and conversely `layoutChildren`) should still be called to ensure the children nicely fill the StackPane's space.

JavaFX currently will do steps 1 and 2, but does not do step 3:

1. Another child is added to the StackPane (an overlay for the ImageView) this triggers a requestLayout on that child which marks it as needing layout
2. This propagates to the parent, which is also marked as needing layout, but it is considered to be a layout root, so no further propagation occurs
3. Since a layout root was encountered, it should be added to the list of dirty layout roots (implemented in this PR).  This will clear all the needs layout flags (otherwise they stay `true` and never go back to `false` again).

The documentation however seems quite clear that step 3 is also supposed to be done.

kevinrushforth:
@arapte Can you review this?

nlisker:
Does `SubScene`'s root also count as a layout root?

hjohn:
> Does `SubScene`'s root also count as a layout root?

After investigation, it seems that SubScene's are considered **scene roots**, so that would make it a layout root as well.  This code in `Parent` (which monitors Scene/SubScene properties) convinces me that's the case:

        sceneRoot = (newSubScene != null && newSubScene.getRoot() == this) ||
                    (newScene != null && newScene.getRoot() == this);
        layoutRoot = !isManaged() || sceneRoot;

palexdev:
> > If the node is unmanaged, why does it need a relayout in practice?
> 
> The unmanaged node does not need relayout, but its potentially managed children do. Just like a Scene does not need layout, but its root does. A node being unmanaged affects its **parent** (the parent must skip it for layout calculations), but does not free it from doing layout on its own children (if I interpret the documentation correctly).
> 
> Let's take this hierarchy:
> 
> ```
> Region
>      StackPane (unmanaged)
>            ImageView
> ```
> 
> The Region will not lay out the stack pane as it is unmanaged, so I position the unmanaged child manually with `resizeRelocate`. The StackPane now has a defined position and size. When I add children to it, the StackPane's `layout` (and conversely `layoutChildren`) should still be called to ensure the children nicely fill the StackPane's space.
> 
> JavaFX currently will do steps 1 and 2, but does not do step 3:
> 
>     1. Another child is added to the StackPane (an overlay for the ImageView) this triggers a requestLayout on that child which marks it as needing layout
> 
>     2. This propagates to the parent, which is also marked as needing layout, but it is considered to be a layout root, so no further propagation occurs
> 
>     3. Since a layout root was encountered, it should be added to the list of dirty layout roots (implemented in this PR).  This will clear all the needs layout flags (otherwise they stay `true` and never go back to `false` again).
> 
> 
> The documentation however seems quite clear that step 3 is also supposed to be done.

So, as of now unmanaged parents do not layout their children? I thought they did.

immagine
Am I getting something wrong?

hjohn:
> > The documentation however seems quite clear that step 3 is also supposed to be done.
> 
> So, as of now unmanaged parents do not layout their children? I thought they did.

@palexdev You are right, I missed something... thanks for taking an extensive look.

My testing branch for this problem has a fix applied for [JDK-8360940](https://bugs.openjdk.org/browse/JDK-8360940), but it now looks like that fix is incomplete.

What **should** happen when an unmanaged node is encountered is that changes in any of its children should not require re-layout of its parents (as an unmanaged node, all container parents ignore it and don't take it into account into their computations, so there is no point to re-layout the parents).

However, due to a bug, code in `Node` will always walk up the layout tree to the scene root and will mark all ancestors as needing layout.  The code that does this is not very subtle and if a layout is in progress it can sometimes cause the layout flags to go into a bad state (this is what [JDK-8360940](https://bugs.openjdk.org/browse/JDK-8360940) fixes).  With that fix applied however, `Node` no longer always walks up the entire tree (and it shouldn't, as an unmanaged node is not affecting its parents).  This however causes unmanaged parents to no longer be laid out at all, and this PR is supposed to fix that in turn.

So, what it looks like is that this PR and the fix for [JDK-8360940](https://bugs.openjdk.org/browse/JDK-8360940) need to be a single PR.

Sorry for the bad PR, I will move this to draft and provide a complete fix.

hjohn:
I've submitted PR #1879 to address the root cause of this issue.  It is a much smaller change, but does expose a flaw in how layout flags are managed... see that PR for more details.

hjohn:
This PR has become irrelevant, will close it for now.

PR #1875, opened 2025-08-18
8361286: Allow enabling of background loading for images loaded from an InputStream

hjohn:
Support background loading of raw input streams

- Fixed generics (mix up of two ImageLoader types)
- Removed unused code for handling headers, methods, request parameters
- Use `long` for progress as streams may exceed 2 GB
- Improved documentation of Image regarding background loading

---------
### Progress
- [x] Change requires CSR request [JDK-8369572](https://bugs.openjdk.org/browse/JDK-8369572) to be approved

### Issues
 * [JDK-8361286](https://bugs.openjdk.org/browse/JDK-8361286): Allow enabling of background loading for images loaded from an InputStream (**Enhancement** - P4)
 * [JDK-8369572](https://bugs.openjdk.org/browse/JDK-8369572): Allow enabling of background loading for images loaded from an InputStream (**CSR**)

### Reviewers
 * [Jayathirth D V](https://openjdk.org/census#jdv) (@jayathirthrao - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1875/head:pull/1875` \
`$ git checkout pull/1875`

`$ git checkout pull/1875` \
`$ git pull https://git.openjdk.org/jfx.git pull/1875/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1875`

`$ git pr show -t 1875`

</details>

<a href="https://git.openjdk.org/jfx/pull/1875.diff">https://git.openjdk.org/jfx/pull/1875.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1875#issuecomment-3197082303)
</details>


kevinrushforth:
Reviewers: @kevinrushforth @jayathirthrao 

/reviewers 2

kevinrushforth:
This will need a CSR. The newly proposed constructors should have an `@since` tag.

/csr needed

modules/javafx.graphics/src/main/java/javafx/scene/image/Image.java
```diff
@@ -710,8 +727,29 @@ public Image(@NamedArg("is") InputStream is) {
...
+     * If loading in the background is requested, then the progress property can
+     * be monitored for loading progress. Otherwise, the image loading is performed
+     * immediately and is completed after this constructor returns.
+     *
+     * @param is the stream from which to load the image
+     * @param backgroundLoading indicates whether the image
+     *      should be loaded in the background
+     * @throws NullPointerException if input stream is null
+     */
+    public Image(@NamedArg("is") InputStream is, @NamedArg("backgroundLoading") boolean backgroundLoading) {
```
andy-goryachev-oracle, line 747: suggestion: swap the arguments to avoid explicit typecasting in the case of `(null, boolean)`

hjohn: `null` isn't allowed, so this is only in the test.  Also, the cast is only to select the correct method overload, there won't actually be a cast.  I also think it is better to keep the same order (there is another constructor that takes an `InputStream` + parameters, and there are similar constructors for URL + parameters)

nlisker:
Do we still use `@NamedArg`? I thought it's a remnant of pre-Java 8 where parameter names weren't saved.

hjohn:
> Do we still use `@NamedArg`? I thought it's a remnant of pre-Java 8 where parameter names weren't saved.

I don't know about the `NamedArg` specifics, but I do know that saving parameter names is optional and I think still off by default (ie. you need to pass a compiler option to save the names).  Other than that, I just copied what was there and kept the same format.

mstr2:
> Do we still use `@NamedArg`? I thought it's a remnant of pre-Java 8 where parameter names weren't saved.

`FXMLoader` uses the `@NamedArg` annotation to associate XML attribute values with constructor arguments.

andy-goryachev-oracle:
> `FXMLoader` uses the `@NamedArg` annotation to associate XML attribute values with constructor arguments.

I wonder if this should be explained in the @NamedArg javadoc.

hjohn:
> > `FXMLoader` uses the `@NamedArg` annotation to associate XML attribute values with constructor arguments.
> 
> I wonder if this should be explained in the @NamedArg javadoc.

It probably should, I never use FXML, but I should have guessed it was for that purpose :)

modules/javafx.graphics/src/main/java/com/sun/javafx/runtime/async/AbstractRemoteResource.java
```diff
@@ -29,14 +29,7 @@
...
 import java.io.InterruptedIOException;
-import java.io.OutputStream;
-import java.net.HttpURLConnection;
-import java.net.URL;
-import java.net.URLConnection;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import java.util.Objects;
```
andy-goryachev-oracle, line 32: all modified files need the copyright year updated

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
```diff
@@ -813,10 +818,42 @@ private float getMaxRenderScale() {
...
+            AsyncOperationListener<ImageLoader> listener, String url,
             double width, double height, boolean preserveRatio, boolean smooth) {
-        return new PrismImageLoader2.AsyncImageLoader(listener, url, width, height, preserveRatio, smooth);
+
+        SizedStreamSupplier sizedStreamSupplier = () -> {
+            URL u = new URL(url);
+            String protocol = u.getProtocol();
+
+            if(protocol.equals("http") || protocol.equals("https")) {
+                HttpURLConnection conn = (HttpURLConnection) u.openConnection();
```
andy-goryachev-oracle, line 831: I know this code has been copied from `AbstractAsyncOperation`, but would it make more sense (anticipating HTTP/3 https://openjdk.org/jeps/517) to open a connection and then check if it is `instanceof HttpURLConnection` ?

modules/javafx.graphics/src/main/java/javafx/scene/image/Image.java
```diff
@@ -671,6 +681,10 @@ public Image(@NamedArg("url") String url, @NamedArg("requestedWidth") double req
     /**
      * Constructs an {@code Image} with content loaded from the specified URL
      * using the specified parameters.
+     * <p>
+     * If loading in the background is requested, then the progress property can
```
andy-goryachev-oracle: suggestion (also in L734, L776):

```
* If loading in the background is requested, then the {@link #progressProperty() progress} property can
```

modules/javafx.graphics/src/main/java/javafx/scene/image/Image.java
```diff
@@ -1065,9 +1132,19 @@ public void cancel() {
         }
 
         private AsyncOperation constructPeer() {
-            return loadImageAsync(this, url,
-                                  requestedWidth, requestedHeight,
-                                  preserveRatio, smooth);
+            if(inputSource == null) {
+                return loadImageAsync(
```
andy-goryachev-oracle: minor: wrapping is not necessary in this method

modules/javafx.graphics/src/test/java/test/javafx/scene/image/ImageTest.java
```diff
@@ -154,6 +155,35 @@ public void loadImageAsyncProgressTest() {
         assertTrue(p3 == p4);
     }
 
+    @Test
```
andy-goryachev-oracle, line 158: Do you think it would make sense to add a test that actually loads a valid image?

andy-goryachev-oracle (commented):
Looks very good!

Noticed it does not close the input stream with a synchronous constructor (it does with the async one).

+ some minor comments and suggestions.

modules/javafx.graphics/src/main/java/com/sun/javafx/runtime/async/AbstractRemoteResource.java
hjohn: Isn't this done in batches, like in #1878 ?

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
hjohn: If you want to pull that into the scope of this feature. I would think doing this with a dedicated ticket (and then fixing it in all places, not just here) would be more aligned with trying to keep changes small and focused.

modules/javafx.graphics/src/test/java/test/javafx/scene/image/ImageTest.java
hjohn: I don't think that can be done with the stub toolkit  This test is specifically stubbing/mocking everything out to verify that  `Image` works correctly.  However, there are many areas where (real) images are loaded already, and all the code involved (excluding the new parts in `Image`) was already there in one way or another as the URL variants that allowed background loading is also just an `InputStream` with some extra steps to determine their size (for progress).

modules/javafx.graphics/src/main/java/com/sun/javafx/runtime/async/AbstractRemoteResource.java
kevinrushforth: Yes, updating the copyright year can be deferred if you prefer.

modules/javafx.graphics/src/main/java/javafx/scene/image/Image.java
hjohn: Fixed

modules/javafx.graphics/src/main/java/javafx/scene/image/Image.java
hjohn: Added this for the 4 constructors that have the `backgroundLoading` parameter. Thanks :)

modules/javafx.graphics/src/test/java/test/javafx/scene/image/ImageTest.java
```diff
@@ -539,6 +569,13 @@ public void createImageFromNullInputStreamTest() {
         });
     }
 
+    @Test
+    public void createImageAsynFromNullInputStreamTest() {
```
danielpeintner:
```suggestion
    public void createImageAsyncFromNullInputStreamTest() {
```

andy-goryachev-oracle:
> @NamedArg

FYI Created https://bugs.openjdk.org/browse/JDK-8365864 but if someone wants to do it, I can send it their way.

modules/javafx.graphics/src/main/java/com/sun/javafx/runtime/async/AbstractRemoteResource.java
```diff
@@ -46,72 +39,39 @@
...
+        SizedStream sizedStream = sizedStreamSupplier.get();
+
+        setProgressMax(sizedStream.size);
 
-        try {
+        try (ProgressInputStream stream = new ProgressInputStream(sizedStream.stream)) {
             return processStream(stream);
         }
-        finally {
-            stream.close();
```
jayathirthrao, line 113: Do we have to continue closing the stream when background loading is happening with URL?

modules/javafx.graphics/src/main/java/com/sun/javafx/runtime/async/AbstractRemoteResource.java
```diff
@@ -46,72 +39,39 @@
...
+    public interface SizedStreamSupplier {
+        SizedStream get() throws IOException;
     }
 
-    protected AbstractRemoteResource(String url, String method, AsyncOperationListener<T> listener) {
-        this(url, method, null, listener);
-    }
+    private final SizedStreamSupplier sizedStreamSupplier;
 
-    protected AbstractRemoteResource(String url, String method, String outboundContent, AsyncOperationListener<T> listener) {
```
jayathirthrao, line 64: I see that outboundContent calls are no where used. Is this the reason we are cleaning up this code?

hjohn: Yeah, I generally remove code that I see is unused, if it otherwise would have needed to be adjusted.

andy-goryachev-oracle:
@hjohn in case you missed my earlier comment:

Noticed it does not close the input stream with a synchronous constructor (it does with the async one).

you can try it with the monkey tester from this branch https://github.com/andy-goryachev-oracle/MonkeyTest/tree/image.from.stream

or simply

```
            byte[] b = ImageTools.writePNG(src);
            ByteArrayInputStream in = new ByteArrayInputStream(b) {
                @Override
                public void close() throws IOException {
                    System.out.println("Closed " + Thread.currentThread());
                }
            };

            // requires JDK-8361286: Allow enabling of background loading for images loaded from an InputStream
            Image im = new Image(in, loadInBackground);
            imageView.setImage(im);
        } catch (IOException e) {
            e.printStackTrace();
        }
```

hjohn:
> @hjohn in case you missed my earlier comment:
> 
> Noticed it does not close the input stream with a synchronous constructor (it does with the async one).
> 
> you can try it with the monkey tester from this branch https://github.com/andy-goryachev-oracle/MonkeyTest/tree/image.from.stream
> 
> or simply
> 
> ```
>             byte[] b = ImageTools.writePNG(src);
>             ByteArrayInputStream in = new ByteArrayInputStream(b) {
>                 @Override
>                 public void close() throws IOException {
>                     System.out.println("Closed " + Thread.currentThread());
>                 }
>             };
> 
>             // requires JDK-8361286: Allow enabling of background loading for images loaded from an InputStream
>             Image im = new Image(in, loadInBackground);
>             imageView.setImage(im);
>         } catch (IOException e) {
>             e.printStackTrace();
>         }
> ```

A good question would be what the behavior should be.  For the URL variants, since the Image class is creating the stream in both cases, it makes sense that it closes that stream.

For `InputStream` parameters, the standard practice is to leave closing up to the caller, unless it is a wrapper (like `BufferedInputStream`).  So, I think how the synchronous variant does it is probably the correct approach.

However, when consuming the stream asynchronously, it is very hard for users to know when to safely close the stream as you basically handed ownership of the stream to another thread (you can't use the stream for anything while background loading is in progress...)  The only way to know when the other thread is done with the stream is to look at the progress property.

So, I think there are a few options:
1. Document that we'll be closing the stream in the background loading case; effectively, since you're giving this InputStream to be used by a different thread, the ownership of the stream transfers from the user to Image.
2. Document that you need to hook into the progress property to know when to close the stream (yikes!)
3. Don't provide the background loading option for `InputStream` (ie. don't integrate this PR, or only the documentation and clean-up bits)
4. Provide alternative API that takes `Supplier<InputStream>`, very powerful, and it would be clear who owns the stream and who closes it, and this would work for both sync/async.

A prime motivation for allowing input streams for background loading was that the URL API is limited (you can't do authentication, use POST, or provide a body).  In those cases you're forced to `InputStream` but then forego the option to load in the background, meaning that you can't have a progress indicator for large images, or show progressive images like with PJPEG.

andy-goryachev-oracle:
re: closing stream

I wonder if we should close the stream in both sync and async variants (and document that).  I would say there should be no reason not to close - suggesting that if the application wants not to close it's surely a bad choice for the application.

For example, one can think of a situation where multiple images are stored within the same stream, but there should be a reliable way for the application to specify the size of each image (and then the app can provide a derived InputStream which EOFs when the end of an image is reached).

Another alternative is to add another boolean `closeStream`.

What do you think?

hjohn:
> re: closing stream
> 
> I wonder if we should close the stream in both sync and async variants (and document that). I would say there should be no reason not to close - suggesting that if the application wants not to close it's surely a bad choice for the application.
> 
> For example, one can think of a situation where multiple images are stored within the same stream, but there should be a reliable way for the application to specify the size of each image (and then the app can provide a derived InputStream which EOFs when the end of an image is reached).
> 
> Another alternative is to add another boolean `closeStream`.
> 
> What do you think?

It's unusual to close the stream when you're not the owner (although some API's do that, they are often obscure and not used that much).  Changing this behavior for the sync variant would be a breaking change, so I'm not sure we want to even go there.

Adding another `boolean` is of course an option, but I think that if we're going for that route, going for a `Supplier` would be more clear API and more powerful (in the internal code, we basically already have that abstraction as it can abstract over both the URL and InputStream variants with just one API).

If we go the `Supplier` route, we can probably also unify the URL and InputStream constructors to go through the same route, moving the URL opening logic into Image class and removing it from the internal code.

andy-goryachev-oracle:
I like the `Supplier` idea, though I am not sure how the async path would work (unless you add the `closeStream` parameter).

kevinrushforth:
We need to leave the existing synchronous constructors alone, other than to document that they do not close the input stream (which is what we do in the `Font::loadFont` methods that take an `InputStream`).

As for the new constructors, I think the options are:

1. Keep it as you propose, where the async variant _always_ closes the stream (the rationale being that the caller has transferred ownership of the stream)
2. Change the new constructors to never close the stream; this would be OK as long as we provide an easy way for the caller to know when they should do it
3. Add a Supplier to the constructors (I'm a little fuzzy on the details of this one)

andy-goryachev-oracle:
I think at least the convenience constructor `Image(InputStream)` should close the stream.

kevinrushforth:
No, that would be an incompatible change.

andy-goryachev-oracle:
You are right, please disregard the comment.

I think we should clarify the behavior in each constructor that takes `InputStream`, similar to `Font::loadFont()`.

A boolean `closeStream` parameter might be the lightest solution from the API perspective then.

mstr2:
A new parameter to toggle whether the stream will be closed should have at least a moderately strong use case to justify its existence. I can't think of any. If you turn over a stream to an asynchronous process, you effectively relinquish ownership of that stream, as you don't know when (if ever) it will be processed.

But even then, you can very easily achieve the non-closing behavior without new JavaFX API, just by using existing stream APIs:
```java
var streamThatWillNotCloseUnderlyingStream = new FilterInputStream(myStream) {
    @Override
    public void close() {
        // don't call super.close(), so underlying stream won't be closed
    }
}
```

modules/javafx.graphics/src/main/java/com/sun/javafx/runtime/async/AbstractRemoteResource.java
jayathirthrao: I see that the resource is getting closed.

jayathirthrao:
If user is passing stream its better to keep the decision of closing the stream with user for synchronous loading.

Asynchronous URL loading : Current changes takes care of closing the stream and follows already present behaviour. Its better to update the documentation about it.
Asynchronous InputStream loading : This change now explicitly closes the Input Stream which is fine but this needs to be captured in the documentation.

Giving an option to user about how to close the stream in case of asynchronous loading can be taken up as future task, if needed. CI testing is green with current code update.

hjohn:
> I like the `Supplier` idea, though I am not sure how the async path would work (unless you add the `closeStream` parameter).

When you have a `Supplier<InputStream>` instead of `URL` or plain `InputStream` it will look like this:

     new Image(() -> new FileInputStream("xyz"));
     
Since the supplier must create a new stream each time (as it can't know how often it will be called), it makes sense that the caller of the supplier is also the one to close it, whether or not that is sync or async.

hjohn:
Thanks for all the input.  I think there are now two ways to proceed:

1. Leave it as is, including the new `backgroundLoading` variants for `InputStream`, but document clearly for both the old and new constructors when the stream will be closed (ie. only when `backgroundLoading` is `true` we will close the stream).

2. Don't add these new constructors, but add a more obvious variant with `Supplier<InputStream>`, where the ownership is more clear, with the caller of the supplier owning the stream created.

I get the impression the consensus leans towards option 1.  I don't see a direct compelling need for the `Supplier` variant, although I'm happy to provide an implementation if option 2 is favored.

kevinrushforth:
I prefer option 1.

andy-goryachev-oracle:
I agree with Kevin: option 1 is clearer.

andy-goryachev-oracle:
this needs a CSR

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
```diff
@@ -813,10 +818,42 @@ private float getMaxRenderScale() {
...
+    @Override public AbstractRemoteResource<ImageLoader> loadImageAsync(
+            AsyncOperationListener<ImageLoader> listener, String url,
             double width, double height, boolean preserveRatio, boolean smooth) {
-        return new PrismImageLoader2.AsyncImageLoader(listener, url, width, height, preserveRatio, smooth);
+
+        SizedStreamSupplier sizedStreamSupplier = () -> {
+            URL u = new URL(url);
+            String protocol = u.getProtocol();
+
+            if(protocol.equals("http") || protocol.equals("https")) {
```
kevinrushforth: Minor: add a space after `if`.

modules/javafx.graphics/src/main/java/javafx/scene/image/Image.java
```diff
@@ -620,6 +620,9 @@ public String getName() {
 
     /**
      * Constructs an {@code Image} with content loaded from the specified URL.
+     * <p>
+     * The image loading is performed immediately and is completed after this
+     * constructor returns.
```
kevinrushforth: I think it would it be clearer to say that it is completed "when" this constructor returns, as you do in a couple other places. As it is "after ... returns" could mislead someone into thinking that the operation might not be complete upon return.

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
andy-goryachev-oracle: minor suggestion: `"http".equals()`

kevinrushforth:
@hjohn In case you missed it, the spec changes for this are almost ready, pending your reply to my feedback about the wording in a few places and the creation of the CSR.

modules/javafx.graphics/src/main/java/javafx/scene/image/Image.java
hjohn: You're right, that is clearer, fixed.

hjohn:
Just to be sure, the CSR required here is for the newly added constructors?  The documentation changes on the existing constructors are only clarifications, no functionality was changed.

kevinrushforth:
> Just to be sure, the CSR required here is for the newly added constructors? The documentation changes on the existing constructors are only clarifications, no functionality was changed.

Technically not needed, but it wouldn't hurt to include them in the CSR as long as you are filing a CSR anyway.

The doc changes look good to me now.

One more thing: Can you merge in the lastest master? Your source branch is quite out of date.

kevinrushforth:
> @kevinrushforth requested review from ... @andy-goryachev-oracle now

You can ignore this. My request crossed your re-review

andy-goryachev-oracle:
> You can ignore this. My request crossed your re-review

race condition: the review was loading in the background from an InputStream /jk

kevinrushforth:
@hjohn The last piece is the CSR. I'll review it once it is ready.

hjohn:
> @hjohn The last piece is the CSR. I'll review it once it is ready.

@kevinrushforth Sorry for the delay, I created the CSR now and moved it to proposed.

kevinrushforth:
> > @hjohn The last piece is the CSR. I'll review it once it is ready.
> 
> @kevinrushforth Sorry for the delay, I created the CSR now and moved it to proposed.

No worries. I just know how easy it is to miss notifications and wanted to make sure you were aware.

I reviewed the CSR.

kevinrushforth:
@hjohn Now that the CSR has been finalized, this PR will be marked as `ready` once the CSR is approved by the CSR chair.

As @jayathirthrao noted in [this comment](#pullrequestreview-3150607807), he has a corresponding change in a closed-side whitebox test that needs to go in at roughly the same time as this PR. Please coordinate the timing of integration with Jay.

jayathirthrao:
@hjohn Since this PR is ready, please delegate the integration of this PR using "/integrate delegate" command.(https://wiki.openjdk.org/display/SKARA/Pull+Request+Commands#PullRequestCommands-/integrate).

This will enable me to check-in this PR and closed-side whitebox test change at same time.

PR #433, opened 2021-03-19
8276671: Iterative layout algorithm

mstr2:
This PR fixes a long-standing issue with baseline layout computation in JavaFX as mentioned in [JDK-8090261](https://bugs.openjdk.java.net/browse/JDK-8090261) and makes baseline layouts more consistent and dependable. There may be other filed bugs or enhancement requests that relate to the same fundamental issue.

### How the baseline offset is computed
Generally, the value returned from `Node.getBaselineOffset()` is one of the following:
1. the actual baseline offset
2. the special constant `BASELINE_OFFSET_SAME_AS_HEIGHT`

The reason for this is that there is a circular dependency between the height of a resizable control and its baseline: in order to know the baseline offset, one must know the size of the control; and in order to know the size of the control, one must know the baseline offset. The special constant `BASELINE_OFFSET_SAME_AS_HEIGHT` basically means: we don't know the baseline offset at this time, but when it comes to layouting the node, we just use whatever the height of the resizable node turns out to be.

However, some resizable controls (`Labeled` being the most prominent) must report an actual baseline value in order to work meaningfully, but this value can't be computed consistently for the reasons mentioned (this is the cause of JDK-809261).

### Step 1: solving `Labeled`
Since we can't know the actual baseline offset of a resizable `Labeled` before the control is laid out, we schedule another layout pass when the `layoutY` property of the `Text` node was changed during layout. Repeatedly layouting and measuring the resulting baseline offset will quickly converge to the true baseline offset of the `Labeled` control. This is the code sample taken from [JDK-8090261](https://bugs.openjdk.java.net/browse/JDK-8090261) before and after the change:
JDK-8090261

### Step 2: making baseline layouts more consistent
Even with `Labeled` solved, the more general issue is that baseline layouts often yield inconsistent results depending on which layout containers are used to compose the scene graph. In order to fix this, we need to ensure that the baseline offset of any given layout container meaningfully reflects the baseline of the nodes placed within it.

This is achieved by adding `Node.isTextBaseline()`, which returns `true` if the node or any of its children contain text (and thus should be the primary source when determining a baseline). Currently, `Labeled`, `Text` and `TextInputControl` return `true` by default.

Additionally, a new boolean property `Node.prefBaselineProperty()` is introduced. This flag can be used to override the baseline source selection.

The default behavior of `Parent.getBaselineOffset()` is changed such that the baseline is derived from the first managed child for which

1. `Node.isPrefBaseline()` returns `true`, or if there is no such child,
[...]

PR #1879, opened 2025-08-23
8360940: Layout stops updating when using Parent#setNeedsLayout(true) due to incorrect state management

hjohn:
Ensures proper propagation of layout flags when using `forceParentLayout = true`.

This was the root cause of issue #1874

### Note

Apparently it is still quite easy to mess up the layout flags. Basically, the layout flag tracked by Parent should always either be `CLEAN` for any scene graph branch, or `!CLEAN` + a layout pulse is scheduled on the corresponding `Scene`.

However, with careful use of the public API `requestLayout` one can get these flags in a bad state still:

Let's say I have a branch `A (root node under Scene) -> B -> C`, **and** a layout is in progress, **and** we're currently in the `layoutChildren` method of `C`. The flag `performingLayout` will be `true` for all nodes in the branch `A` ->  `B` -> `C`.  The `layout` method will set the layout flag to `CLEAN` as its first action, so when we're at `C::layoutChildren`, all flags have been reset to `CLEAN` already.  See the `Parent::layout` method for how all this works.

Now, to mess up the flags, all you need to do is call `requestLayout` on `B` or `C` from the `layoutChildren` of `C` (or indirectly by changing something and something is listening to this and schedules a layout on something somewhere in this branch); note that `requestLayout` is not documented to be illegal to call during layout, and some classes in FX will do so (ScrollPaneSkin, NumberAxis, etc..) risking the flags getting in a bad state... -- usually you get away with this, as there are many ways that layout is triggered, and eventually the flags will get overwritten and reset to a consistent state.

The bad state occurs because this code path is followed (all code from `Parent`):

    public void requestLayout() {
        clearSizeCache();
        markDirtyLayout(false, forceParentLayout);
    }
    
Calls to `markDirtyLayout(false, false)`:

    private void markDirtyLayout(boolean local, boolean forceParentLayout) {
        setLayoutFlag(LayoutFlags.NEEDS_LAYOUT);
        if (local || layoutRoot) {
            if (sceneRoot) {
                Toolkit.getToolkit().requestNextPulse();
                if (getSubScene() != null) {
                    getSubScene().setDirtyLayout(this);
                }
            } else {
                markDirtyLayoutBranch();
            }
        } else {
            requestParentLayout(forceParentLayout);
        }
    }
    
Before going into the `else` (as none of the nodes is a layout root, and `local` was set to `false`) it will do **setLayoutFlag(LayoutFlags.NEEDS_LAYOUT)** -- this will set a flag on some node; to eventually end up in a consistent state, it must mark all ancestors as well and schedule a pulse (with `Toolkit.getToolkit().requestNextPulse()`)... but:

    void requestParentLayout(boolean forceParentLayout) {
        if (!layoutRoot) {
            final Parent p = getParent();
            if (p != null && (!p.performingLayout || forceParentLayout)) {

                /*
                 * The forceParentLayout flag must be propagated to mark all ancestors
                 * as needing layout. Failure to do so while performingLayout is true 
                 * would stop the propagation mid-tree. This leaves some nodes as needing 
                 * layout, while its ancestors are clean, which is an inconsistent state.
                 */

                p.requestLayout(forceParentLayout);
            }
        }
    }
    
Here there is a guard `!p.isPerformingLayout`, blocking propagation up the tree.  As said, this flag is `true` for all nodes during a layout of the same branch.  The end result thus is that some nodes have their layout flag changed to `NEEDS_LAYOUT`, but it was not propagated, nor was a pulse scheduled...

---------
### Progress

### Issue
 * [JDK-8360940](https://bugs.openjdk.org/browse/JDK-8360940): Layout stops updating when using Parent#setNeedsLayout(true) due to incorrect state management (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1879/head:pull/1879` \
`$ git checkout pull/1879`

`$ git checkout pull/1879` \
`$ git pull https://git.openjdk.org/jfx.git pull/1879/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1879`

`$ git pr show -t 1879`

</details>

<a href="https://git.openjdk.org/jfx/pull/1879.diff">https://git.openjdk.org/jfx/pull/1879.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1879#issuecomment-3217229894)
</details>

### DIFF
--- a/modules/javafx.graphics/src/test/java/test/javafx/scene/ParentTest.java
+++ b/modules/javafx.graphics/src/test/java/test/javafx/scene/ParentTest.java
@@ -25,6 +25,7 @@
 
 package test.javafx.scene;
 
+import com.sun.javafx.scene.LayoutFlags;
 import com.sun.javafx.scene.NodeHelper;
 import test.com.sun.javafx.pgstub.StubToolkit;
 import com.sun.javafx.sg.prism.NGGroup;
@@ -43,6 +44,8 @@ import javafx.scene.Node;
 import javafx.scene.Parent;
 import javafx.scene.ParentShim;
 import javafx.scene.Scene;
+import javafx.scene.layout.HBox;
+import javafx.scene.layout.VBox;
 import javafx.scene.shape.Rectangle;
 import javafx.stage.Stage;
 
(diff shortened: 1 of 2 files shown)


kevinrushforth:
The fix looks correct. Can you provide an automated test?

kevinrushforth:
I recommend removing `(public api)` from the title of the JBS bug and this PR -- we don't typically do that and it seems superfluous here.

hjohn:
@kevinrushforth I think I managed a suitable unit test.  It fails without the fix, and passes with the fix.

modules/javafx.graphics/src/test/java/test/javafx/scene/ParentTest.java
```diff
@@ -556,7 +559,91 @@ public void assertAndClear(boolean b) {
...
+
+        root.layout();
 
+        // Assert that after another layout pass all are clean again:
+        // Note: we still modify the sibling, but since its layoutX is unchanged now, no further pass is triggered
+        assertEquals(LayoutFlags.CLEAN, ParentShim.getLayoutFlag(root));
+        assertEquals(LayoutFlags.CLEAN, ParentShim.getLayoutFlag(level1));
+        assertEquals(LayoutFlags.CLEAN, ParentShim.getLayoutFlag(level2));
+        assertEquals(LayoutFlags.CLEAN, ParentShim.getLayoutFlag(leaf));
+        assertEquals(LayoutFlags.CLEAN, ParentShim.getLayoutFlag(sibling));
```
andy-goryachev-oracle, line 646: Excellent test, thank you!

Question:
When a situation like this happens in reality, we might see a momentary flicker due to the layout spanning several pulses, correct?
The platform itself can't really do anything, except for the application code to call the layout() explicitly to avoid flicker, right?

kevinrushforth: This is already the case irrespective of this PR. The point of this PR is to ensure that the flags are consistently set as to which part(s) of the scene graph are dirty and need to be reevaluated for the next layout pass.

modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
```diff
@@ -1066,7 +1066,15 @@ void requestParentLayout(boolean forceParentLayout) {
...
-                p.requestLayout();
+
+                /*
+                 * The forceParentLayout flag must be propagated to mark all ancestors
+                 * as needing layout. Failure to do so while performingLayout is true
+                 * would stop the propagation mid-tree. This leaves some nodes as needing
+                 * layout, while its ancestors are clean, which is an inconsistent state.
+                 */
+
+                p.requestLayout(forceParentLayout);
```
andy-goryachev-oracle, line 1077: very, very minor suggestion to makes it more compact (it's ok to leave things as is):

```

                // The forceParentLayout flag must be propagated to mark all ancestors
                // as needing layout. Failure to do so while performingLayout is true
                // would stop the propagation mid-tree. This leaves some nodes as needing
                // layout, while its ancestors are clean, which is an inconsistent state.
                p.requestLayout(forceParentLayout);
```

modules/javafx.graphics/src/test/java/test/javafx/scene/ParentTest.java
mstr2: We should probably resurrect #433 at some point.

andy-goryachev-oracle: Or maybe even make it more generic: detect when (a small number of) additional layout passes are needed and do them right away instead of waiting for another pulse?

mstr2: That's exactly what this old PR does.

hjohn: > When a situation like this happens in reality, we might see a momentary flicker due to the layout spanning several pulses, correct?

Well, if layout runs (due to a significant change), you'll see components be re-arranged, and if a 2nd pulse is needed, you would probably not notice as things were moved around a lot anyway.  But on top of that, there is a good chance the 2nd pass is just recalculating everything without actually changing anything because it is often triggered without a good reason.  For example:

Take an HBox for example.  During layoutChildren, it will position its children.  If this involves moving a child to the right or left because an earlier child has shrunk/grown then layout X is changed and triggers the logic in Node (without good reason I might add).  This is in part because HBox does not update or set the "current layout child" at all (none of the containers do this, except Parent's layoutChildren which is almost always overridden).  So this code below in Node will trigger and force another layout pass:

                    if (p != null && !p.isCurrentLayoutChild(Node.this)) {
                        if (isManaged()) {
                            // Force its parent to fix the layout since it is a managed child.
                            p.requestLayout(true);
                        } else {
                            // Parent size changed, parent's parent might need to re-layout
                            p.clearSizeCache();
                            p.requestParentLayout();
                        }
                    }
                    
All the ingredients match:
- LayoutX was modified on a child
- Its parent is not `null` (it's the HBox)
- It is not the current layout child (HBox doesn't update this before modifying a position on a child)
- The child in question is managed (all children in an HBox generally are)

End result: schedule another layout pass

Now, if in the 2nd pass none of the X (or Y) positions change again, then this 2nd pass will end up doing nothing.  Quite wasteful.

> The platform itself can't really do anything, except for the application code to call the layout() explicitly to avoid flicker, right?

Calling `layout` won't do much when the flags are bad, as layout basically operates based on those flags.  So if you call layout at the root, and the flags say that it is `CLEAN` then it stops right there.  

However there are many things that will "fix" the situation; this can be a nearby Node that changes size for some reason, or say the user resizing the Window.  Generally, you will almost never notice as often something triggers layout again, fixing the flags, but if say a custom control is careful about when to redo calculations (because they're even more expensive than usual) and uses tricks like `setNeedsLayout(true)` then a part of the UI may not get updated as you'd expect as the flags got into a bad state.  

Also see the long note on top of this PR.... the way the flags are managed, and the way that Parent allows `requestLayout` to be overridden by children (really bad design there), it is still possible to mess up the flags.  All this PR does is remove one of the ways (I'd remove the other way as well, but it would mean making `requestLayout` final).

hjohn:
Thanks for the quick reviews @andy-goryachev-oracle @kevinrushforth

andy-goryachev-oracle:
sorry for the long delay...

modules/javafx.graphics/src/test/java/test/javafx/scene/ParentTest.java
hjohn: > Or maybe even make it more generic: detect when (a small number of) additional layout passes are needed and do them right away instead of waiting for another pulse?

But also avoid passes that are not needed... I mean HBox almost always triggering a redundant 2nd pass makes the logic in `Node` suspect IMHO.

I get why the logic is **really** there; a user may modify layoutX randomly on some managed control outside a layout pass; in that case you do want the Parent that is managing those nodes to take control and "fix" what the user broke, but it is triggering much more often than needed.

andy-goryachev-oracle: that was a good, informative discussion, thank you all!

PR #1881, opened 2025-09-02
8366568: Allow replacing built in application menu in system menu bar

beldenfox:
This PR provides an API for replacing the application menu that JavaFX places in the system menu bar on macOS. The application menu is the one next to the Apple menu that is titled with the application's name and contains the Quit menu item. The JavaFX version of this menu has many limitations so this PR allows a developer to simply replace it outright.

Technically there’s no reason to restrict this to only one menu so the API allows the developer to specify an observable list of Menus known as the common system menus. Like the current application menu the common menus are present for all JavaFX windows and remain present even if all windows lose focus. Any per-window menus provided by a MenuBar with useSystemMenuBar set will appear alongside the common system menus.

In addition the PR provides some calls that allow a developer to implement the Hide <app name>, Hide Others, and Show All menu items that are expected in the application menu.

---------

---------
### Progress
- \[ \] Change requires CSR request [JDK-8369047](https://bugs.openjdk.org/browse/JDK-8369047) to be approved

### Issues
 * [JDK-8366568](https://bugs.openjdk.org/browse/JDK-8366568): Allow replacing built in application menu in system menu bar (**Enhancement** - P4)
 * [JDK-8369047](https://bugs.openjdk.org/browse/JDK-8369047): Allow replacing built in application menu in system menu bar (**CSR**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1881/head:pull/1881` \
`$ git checkout pull/1881`

`$ git checkout pull/1881` \
`$ git pull https://git.openjdk.org/jfx.git pull/1881/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1881`

`$ git pr show -t 1881`

</details>

<a href="https://git.openjdk.org/jfx/pull/1881.diff">https://git.openjdk.org/jfx/pull/1881.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1881#issuecomment-3245835331)
</details>


kevinrushforth:
The idea seems interesting. It will need discussion on the mailing list, which I see you've started.

/reviewers 2 reviewers
/csr needed

kevinrushforth:

andy-goryachev-oracle:
@beldenfox have you decided not to proceed with this PR?

beldenfox:
@andy-goryachev-oracle There were no replies in this PR discussion or on the mailing list.

The system menu bar on Mac has plenty of issues which have been around for a long time. We do not provide About or Preference menu items and they can’t be added ([JDK-8300180](https://bugs.openjdk.org/browse/JDK-8300180), [JDK-8091683](https://bugs.openjdk.org/browse/JDK-8091683)). The menu items can’t be localized ([JDK-8088585](https://bugs.openjdk.org/browse/JDK-8088585), [JDK-8343728](https://bugs.openjdk.org/browse/JDK-8343728)). The behavior of the Quit menu item is wrong in multiple ways and can’t be overridden and the item can't be removed ([JDK-8293700](https://bugs.openjdk.org/browse/JDK-8293700), [JDK-8087607](https://bugs.openjdk.org/browse/JDK-8087607), [JDK-8091776](https://bugs.openjdk.org/browse/JDK-8091776)).

I get the impression that developers who care about the Mac turned to third-party solutions a long time ago. Perhaps this PR is too little too late. I would be happy to re-open it if there's any interest.

andy-goryachev-oracle (commented):
I'll take a look at this.

andy-goryachev-oracle:
please sync up with the latest `master` branch

modules/javafx.controls/src/main/java/javafx/scene/control/MenuBar.java
```diff
@@ -172,6 +174,51 @@ public final boolean isUseSystemMenuBar() {
...
+    /**
+     * The common system menus in the system menu bar.
+     *
+     * A list of menus which are always present in the system menu bar (if the
+     * platform supports it). If a MenuBar uses the system menu bar its menus
+     * will appear alongside the common system menus. If this list is
+     * modified at runtime the system menu bar will update as expected.
+     *
+     * @return the common system menus list property
+     * @since 26
```
andy-goryachev-oracle: needs to be updated to 28 most likely

modules/javafx.controls/src/main/java/javafx/scene/control/MenuBar.java
```diff
@@ -172,6 +174,51 @@ public final boolean isUseSystemMenuBar() {
...
+                }
+
+            };
+        }
+        return commonSystemMenus;
+    }
+
+    private static SimpleObjectProperty<ObservableList<Menu>> commonSystemMenus;
+    public static final void setCommonSystemMenus(ObservableList<Menu> value) {
+        commonSystemMenusProperty().setValue(value);
```
andy-goryachev-oracle, line 217: what is the use case for this?

would it make more sense to change to `final commonSystemMenus` and disallow setting the list instance?  the application can just `setAll()`.

beldenfox:
I've re-opened this. I know there's at least one developer out there that's mixing AWT and JavaFX in an unsupported way just so they can add an "About" item to the application menu (using AWT's application menu with JavaFX windows).

Same code as before but I added a facility for testing the system menu code. In MacApplication.java there's a call to find menu items in NSApp.mainMenu so tests can verify that the system menu code is working. I know that this sounds like JavaFX policing itself but it's the best I could come up with; a simple, isolated bit of code that can be used to verify the behavior of a much more complicated chunk of code. I haven't tested it yet but we should also be able to use this to verify the interaction between AWT and JavaFX when both are using their system menu bars.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/MenuBarSkin.java
```diff
@@ -488,6 +492,40 @@ public static void setDefaultSystemMenuBar(final MenuBar menuBar) {
...
+            }
+        }
+        Toolkit.getToolkit().getSystemMenu().setCommonMenus(wrappedCommonSystemMenus);
+    }
+
+    /**
+     * Set the common system menus. Intended as a private API.
+     * @param menuList list of common menus
+     */
+    public static void setCommonSystemMenus(final ObservableList<Menu> menuList) {
```
andy-goryachev-oracle, line 511: This adds a public API (the skin is public).

Do we *want* to have this as public API?  If not, we should probably use a helper.

beldenfox:
Before delving into the code there should be some discussion on scope. This is the Big Comprehensive Fix. It allows a developer to do much more than manipulate or replace the application menu. They can also implement Mac-like behavior by installing a single application-wide system MenuBar that's used across all windows and that persists even when all windows are minimized.

We could implement something much smaller in scope. This PR proves that the application menu can be implemented as a standard JavaFX Menu control (currently it's built using lower-level parts that are private). We could just provide an API that allows a developer to retrieve the application menu as a standard JavaFX Menu control. Then they could customize the items inside it at will just like any other Menu. That would limit the API to just tweaking the application menu but also greatly reduce the scope of testing.

modules/javafx.graphics/src/main/java/javafx/application/ApplicationServices.java
```diff
@@ -0,0 +1,69 @@
...
+ * questions.
+ */
+
+package javafx.application;
+
+import com.sun.javafx.application.PlatformImpl;
+
+/**
+ * This class provides services for an Application. This includes
+ * methods to show and hide other applications.
```
andy-goryachev-oracle, line 32: what is the purpose of this?  how is it relevant to the application menu?

tests/system/src/test/java/test/javafx/scene/control/CommonSystemMenusTest.java
```diff
@@ -0,0 +1,323 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+import test.util.Util;
+
+public class CommonSystemMenusTest {
```
andy-goryachev-oracle, line 51: Does this PR impact swing/fx interop?  If yes, shouldn't we have more tests that verify that the system menus are correctly handled between swing and fx?

andy-goryachev-oracle (commented):
A few questions:
1. what is the status of the other JBS tickets in relation to this PR?  Are any of these getting fixed by this change?
2. what about swing/fx interop?  do we need more tests developed, or testing done?
3. is there any way to insert the standard platform menu such as Help (one with the search box)?
4. some added public APIs added seem unnecessary, what was the reason for their inclusion?
5. (probably unrelated) the first system menu is "java" - is there a way to customize that?
 
Also, it would be *awfully* nice to have a small demo app to showcase the feature and the related issues, possibly with a swing frames as well.  I tried to cannibalize the test code in

https://github.com/andy-goryachev-oracle/Test/blob/8366568.system.menu/src/goryachev/tests/SystemMenu_8366568.java

but it is somewhat lacking.

modules/javafx.controls/src/main/java/javafx/scene/control/MenuBar.java
beldenfox: At one point I distinguished between a `null` value and an empty list. The `null` value reverted to the old behavior (the built-in application menu) and an empty list removed the application menu entirely. I got rid of that distinction at some point; now an empty list and a `null` both cause the built-in application menu to appear. A call that returns an observable list would better match how other JavaFX API's work so I'm open to changing over to that style.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/MenuBarSkin.java
beldenfox: Good catch. No, this shouldn't be public. I was waiting some discussion of the approach before going through the tedium of hooking up the helper.

modules/javafx.graphics/src/main/java/javafx/application/ApplicationServices.java
beldenfox: If you want to replace the default application menu you need to provide three standard items, "Hide AppName", "Hide Others", and "Show All". These require calling into the platform code to implement. There was no obvious place to add these calls so I added ApplicationServices.

tests/system/src/test/java/test/javafx/scene/control/CommonSystemMenusTest.java
beldenfox: Yes, the Swing/JFX interop tests should be expanded. The interop code between Swing and JavaFX happens at a lower level so it should all work but, yes, it should be tested. That would be part of a re-write of the manual test cases that's already on my plate.

beldenfox:
@andy-goryachev-oracle I need to do some more research to answer some of your questions. In particular I need to do some more digging through the JDK issues to come up with a comprehensive list and do some research on the Help/search box issue. The one question I can answer is the title of the application menu ("java"). We do not control that. The OS consults the application bundle to determine that title and so for the java application it will always be "java". A JavaFX application can change that by providing a different bundle in their app package.

beldenfox:
I should add an additional bit of context (that I had forgotten). On the Mac it's expected that an application can remain active with a menu bar even when there are no windows open. Currently with a JavaFX app the system menu will empty out leaving only the application menu. Scene Builder uses a private API to install some default menus that are shown when all the windows go away. This PR provides a public API that could be used to cover that case as well.

modules/javafx.controls/src/main/java/javafx/scene/control/MenuBar.java
andy-goryachev-oracle: Is a no-menu case a valid one?  I mean, is this something we should be able to do for completeness sake, or it totally makes no sense?

andy-goryachev-oracle:
Speaking of the system menu and `ApplicationServices`, macos-only features, and the new API that are being brought in to the classes common to all the platforms: could there be a better way to express the desired functionality?

So on mac, we have a number of features that are only available there - the system menu, the notch, the key bar that I think they got rid of already.  One way to support these (used currently by javafx) is to add these API into common classes (`MenuBar.setUseSystemMenuBar()` for example) which would be no-op on other platforms.

The other possibility is to provide a proxy interface through which these platform-specific APIs will be available on the supported platform, and it will be null on all others.

Example:

```java
AppleSystemMenu m = Lookup.get(AppleSystemMenu.class);
if(m == null) {
  // not available, probably not a macos
} else {
  // macos system menu functionality is available, use it
  m.getSystemMenuBar();
  m.getHideAllAction();
  // etc.
}
```

These proxies can be multiplied as the platform evolves (notch? brain interface?), or even provided by a third party.

In other words, the common functionality is provided by the base classes, and the platform- and version- specific functionality - by the corresponding proxies.

What do you think?

beldenfox:
I cleaned up my manual test app and added it to this PR. Sorry it's so elaborate. Common menus have parentheses in their titles. There are no safeguards: you can shift the application menu around and even remove it.

modules/javafx.controls/src/main/java/javafx/scene/control/MenuBar.java
beldenfox: A Mac application should always have an application menu. The no-menu case would require a developer to add a copy of the application menu to every MenuBar in every window in their application. That doesn't sound useful to me but I could be wrong.

Currently if the number of common menus goes to zero for any reason I use the default application menu instead. It got complicated trying to make a distinction between null, an empty list, and a non-empty list where all the menus are hidden.

andy-goryachev-oracle: `<rant>`Yeah, I love it when my window is on monitor one, and the application menu is on monitor 3.  I guess it was a fine idea back then when the only CRT screen had 512 × 342 pixels, and by moving the mouse way up the application menu popped up.  All of this is gone now, but the paradigm remains, like that flower which used to be pollinated by beetles that went extinct 100 million years ago`</rant>`

Getting back to the issue - does it mean you are ok with removing the setter and only exposing the final `ObservableList`?  

What do other people think?

modules/javafx.graphics/src/main/java/javafx/application/ApplicationServices.java
andy-goryachev-oracle: This class does not feel right: it provides a public API to something that only exists on one platform.  Or do you have plans to offer similar functionality in all supported platforms?

I think this warrants a wider discussion.

modules/javafx.graphics/src/main/java/javafx/application/ApplicationServices.java
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @since 26
```
andy-goryachev-oracle, line 34: minor: `@since 28` here, and I don't think it needs to be mentioned in every method in this class.

tests/manual/controls/CommonSystemMenus.java
```diff
@@ -0,0 +1,305 @@
...
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
```
andy-goryachev-oracle, line 25: Thank you for providing a test application!

The place seems wrong - there is a bunch of existing system menu tests in `tests/manual/systemmenu`, so maybe there.

Also, it probably should not use the default package, but placed in its own.

beldenfox:
> Speaking of the system menu and `ApplicationServices`, macos-only features, and the new API that are being brought in to the classes common to all the platforms: could there be a better way to express the desired functionality?
> 
> So on mac, we have a number of features that are only available there - the system menu, the notch, the key bar that I think they got rid of already. One way to support these (used currently by javafx) is to add these API into common classes (`MenuBar.setUseSystemMenuBar()` for example) which would be no-op on other platforms.

The system menu bar is not Mac-centric by nature. The existing setUseSystemMenuBar API could potentially be implemented on Windows and Linux; the user would just see one platform-implemented menu bar per window instead of a shared one at the top of the screen. In fact it looks like this used to work but was removed? There are MenuBarDelegates for GTK and Windows still in the codebase and on Windows there's still code in Glass for building an HMENU. There's a middle layer in the Toolkit which assumes there's only one system menu bar but the machinery on either side of it could potentially allow one per window. Interesting.

But in general I agree with you. The system menu bar isn't inherently platform-specific but the application menu is. If we had a clean way of providing platform-specific calls I would consider closing out this PR and adding a Mac-specific call just for replacing the application menu. The only reason this PR allows for multiple common menus is that it seemed strange to provide a public API that was restricted to just one. But if we provided a Mac-specific API replacing just a single menu would make perfect sense.

andy-goryachev-oracle:
> The only reason this PR allows for multiple common menus

I guess my concern is not with the plurality of menus, but the fact that we are trying to add a large API surface for a single platform, knowing that it's very unlikely that other platforms decide to support these menus.

I want to know what other people think, but in my opinion (assuming we do want to provide platform-specific functionality like ApplicationServices) the decision should be somewhere between these two options:

1. implement it as a library outside of javafx
2. maybe consider a way to provide access to platform-specific services via proxy lookup described earlier

I don't think we will succeed by trying to find a viable intersection among the three major platforms.  (edit: not a challenge, just an opinion).

beldenfox:
I don't have a good answer for this. The application menu and its standard menu items (like "Hide Others") are very specific to the Mac so anything we do in this area will be platform-specific. AWT provides a set of functions for customizing aspects of the application menu which is designed to look platform agnostic but which only works on the Mac. JavaFX could engage in similar API pageantry but I don't see the point. Personally I would rather just turn these calls into no-ops on other platforms and let developers ignore them as appropriate.

In any case coming to a consensus on how to handle these calls requires more voices in the conversation.

> 1. implement it as a library outside of javafx

In that case I would just skip this PR entirely. There are already third-party libraries for handling Mac menus. They're not drop-in replacements for the existing MenuBar API but they cover a lot more ground.

PR #1883, opened 2025-09-03
8365402: Bump minimum JDK version for JavaFX to JDK 24

kevinrushforth:
This PR bumps the minimum version of the JDK needed to run JavaFX to JDK 24.

In order for JavaFX to be able to use more recent JDK features, we should increase the minimum version of the JDK that can run the latest JavaFX. Additionally, there is an ongoing cost to keeping JavaFX buildable and runnable on older versions of Java, and little reason to continue to do so.

This continues our current practice of setting the minimum JDK for JavaFX N to JDK N-2. JavaFX N is primarily intended for use with JDK N and we also build and test it against JDK N-1 (which is typically what we use as the boot JDK). Anything older than that, including the proposed minimum JDK N-2 (24 in this specific case), is untested.

This PR targeted to JavaFX 26, and must not be backported to any earlier version. This needs a CSR and a release note.

/reviewers 2
/csr

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -90,9 +90,9 @@ javadoc.top=<div style="padding: 6px; text-align: center; font-size: 80%;">This 
 jdk.docs.version=24
 jfx.build.jdk.version=24.0.1
 jfx.build.jdk.buildnum=9
-jfx.build.jdk.version.min=23
-jfx.build.jdk.buildnum.min=37
-jfx.jdk.target.version=23
+jfx.build.jdk.version.min=24
+jfx.build.jdk.buildnum.min=36
+jfx.jdk.target.version=24
 
 # The jfx.gradle.version property defines the version of gradle that is
 # used in the build. It must match the version number in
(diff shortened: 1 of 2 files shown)


kevinrushforth:
Reviewers: @arapte and either @johanvos or @tiainen

build.properties
```diff
@@ -90,17 +90,17 @@ javadoc.top=<div style="padding: 6px; text-align: center; font-size: 80%;">This
 jdk.docs.version=24
 jfx.build.jdk.version=24.0.1
 jfx.build.jdk.buildnum=9
-jfx.build.jdk.version.min=23
-jfx.build.jdk.buildnum.min=37
-jfx.jdk.target.version=23
+jfx.build.jdk.version.min=24
+jfx.build.jdk.buildnum.min=36
```
danielpeintner, line 94:
```suggestion
jfx.build.jdk.buildnum.min=38
```

Not 100% sure but I think the number should _increase_ and not _decrease_

kevinrushforth: No. Build numbers are only meaningful within a specific release. JDK 23 GA was build 37 (jdk-23+37). JDK 24 GA was build 36 (jdk-24+36).

kevinrushforth:
I plan to integrate this on Monday.

Maran23 (approved):
Nice step forward. Makes sense to also update the minimum required Gradle version.

PR #1885, opened 2025-09-07
Bulk change notifications for ObservableSet and ObservableMap

mstr2:
Implementation of bulk change notifications for ObservableSet and ObservableMap.

### DIFF
--- a/modules/javafx.base/src/main/java/com/sun/javafx/collections/ObservableMapWrapper.java
+++ b/modules/javafx.base/src/main/java/com/sun/javafx/collections/ObservableMapWrapper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -30,9 +30,12 @@ import javafx.collections.MapChangeListener;
 import javafx.collections.ObservableMap;
 
 import java.util.Collection;
+import java.util.ConcurrentModificationException;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
+import java.util.function.BiFunction;
 
 /**
  * A Map wrapper class that implements observability.
@@ -48,67 +51,6 @@ public class ObservableMapWrapper<K, V> implements ObservableMap<K, V>{
 
     public ObservableMapWrapper(Map<K, V> map) {
         this.backingMap = map;
-    }
-
-    private class SimpleChange extends MapChangeListener.Change<K,V> {
-
-        private final K key;
-        private final V old;
-        private final V added;
-        private final boolean wasAdded;
-        private final boolean wasRemoved;
-
-        public SimpleChange(K key, V old, V added, boolean wasAdded, boolean wasRemoved) {
-            super(ObservableMapWrapper.this);
-            assert(wasAdded || wasRemoved);
-            this.key = key;
-            this.old = old;
-            this.added = added;
-            this.wasAdded = wasAdded;
-            this.wasRemoved = wasRemoved;
-        }
-
-        @Override
-        public boolean wasAdded() {
-            return wasAdded;
-        }
-
-        @Override
-        public boolean wasRemoved() {
-            return wasRemoved;
-        }
-
-        @Override
-        public K getKey() {
-            return key;
-        }
-
-        @Override
-        public V getValueAdded() {
-            return added;
-        }
-
-        @Override
-        public V getValueRemoved() {
-            return old;
-        }
-
-        @Override
-        public String toString() {
-            StringBuilder builder = new StringBuilder();
-            if (wasAdded) {
-                if (wasRemoved) {
-                    builder.append(old).append(" replaced by ").append(added);
-                } else {
-                    builder.append(added).append(" added");
-                }
-            } else {
-                builder.append(old).append(" removed");
-            }
-            builder.append(" at key ").append(key);
-            return builder.toString();
-        }
-
     }
 
     protected void callObservers(MapChangeListener.Change<K,V> change) {
@@ -188,20 +130,154 @@ public class ObservableMapWrapper<K, V> implements ObservableMap<K, V>{
 
     @Override
     public void putAll(Map<? extends K, ? extends V> m) {
-        for (Map.Entry<? extends K, ? extends V> e : m.entrySet()) {
-            put(e.getKey(), e.getValue());
+        int size = m.size();
+
+        if (size == 1) {
+            var entry = m.entrySet().iterator().next();
+            put(entry.getKey(), entry.getValue());
+        } else if (size > 1) {
+            var change = new BulkChange.AddReplace<>(this, size);
+
+            for (Map.Entry<? extends K, ? extends V> e : m.entrySet()) {
+                K key = e.getKey();
+                V newValue = e.getValue();
+
+                if (!backingMap.containsKey(key)) {
+                    change.nextAdded(key, newValue);
+                } else {
+                    V oldValue = backingMap.get(key);
+
+                    if (!Objects.equals(oldValue, newValue)) {
+                        change.nextReplaced(key, oldValue, newValue);
+                    }
+                }
+            }
+
+            change.complete();
+            backingMap.putAll(m);
+            callObservers(change);
         }
     }
 
     @Override
-    public void clear() {
-        for (Iterator<Entry<K, V>> i = backingMap.entrySet().iterator(); i.hasNext(); ) {
-            Entry<K, V> e = i.next();
-            K key = e.getKey();
-            V val = e.getValue();
-            i.remove();
-            callObservers(new SimpleChange(key, val, null, false, true));
+    public void replaceAll(BiFunction<? super K, ? super V, ? extends V> function) {
+        Objects.requireNonNull(function);
+        MapChangeListener.Change<K, V> change = null;
+        int i = 0;
+
+        for (Map.Entry<K, V> entry : backingMap.entrySet()) {
+            K key;
+            V oldValue;
+
+            try {
+                key = entry.getKey();
+                oldValue = entry.getValue();
+            } catch (IllegalStateException ex) {
+                // This usually means the entry is no longer in the map.
+                throw new ConcurrentModificationException(ex);
+            }
+
+            // IllegalStateException thrown from function is not a ConcurrentModificationException.
+            V newValue = function.apply(key, oldValue);
+
+            try {
+                if (!Objects.equals(oldValue, newValue)) {
+                    entry.setValue(newValue);
+
+                    if (change instanceof SimpleChange) {
+                        int capacity = backingMap.size() - i + 1;
+                        var bulkChange = new BulkChange.AddReplace<>(ObservableMapWrapper.this, capacity);
+                        bulkChange.nextReplaced(change.getKey(), change.getValueRemoved(), change.getValueAdded());
+                        bulkChange.nextReplaced(key, oldValue, newValue);
+                        change = bulkChange;
+                    } else if (change instanceof BulkChange.AddReplace<K, V> bulkChange) {
+                        bulkChange.nextReplaced(key, oldValue, newValue);
+                    } else {
+                        change = new SimpleChange(key, oldValue, newValue, true, true);
+                    }
+                }
+            } catch (IllegalStateException ex) {
+                // This usually means the entry is no longer in the map.
+                throw new ConcurrentModificationException(ex);
+            }
+
+            ++i;
         }
+
+        if (change == null) {
+            return;
+        }
+
+        if (change instanceof BulkChange<K, V> bulkChange) {
+            bulkChange.complete();
+        }
+
+        callObservers(change);
+    }
+
+    @Override
+    public void clear() {
+        int size = backingMap.size();
+
+        if (size == 1) {
+            Iterator<Entry<K, V>> it = backingMap.entrySet().iterator();
+            Entry<K, V> entry = it.next();
+            K key = entry.getKey();
+            V val = entry.getValue();
+            it.remove();
+            callObservers(new SimpleChange(key, val, null, false, true));
+        } else if (size > 1) {
+            var change = new BulkChange.Remove<>(this, size);
+
+            for (Map.Entry<? extends K, ? extends V> e : backingMap.entrySet()) {
+                change.nextRemoved(e.getKey(), e.getValue());
+            }
+
+            change.complete();
+            backingMap.clear();
+            callObservers(change);
+        }
+    }
+
+    private boolean removeRetain(Collection<?> c, ContainsPredicate<K, V> p, boolean remove) {
+        MapChangeListener.Change<K, V> change = null;
+
+        for (Iterator<Entry<K, V>> it = backingMap.entrySet().iterator(); it.hasNext();) {
+            Entry<K, V> e = it.next();
+
+            if (remove == p.contains(c, e)) {
+                K key = e.getKey();
+                V value = e.getValue();
+
+                if (change instanceof SimpleChange) {
+                    int capacity = remove
+                        ? Math.min(backingMap.size() + 1, c.size())
+                        : backingMap.size() + 1;
+
+                    var bulkChange = new BulkChange.Remove<>(ObservableMapWrapper.this, capacity);
+                    bulkChange.nextRemoved(change.getKey(), change.getValueRemoved());
+                    bulkChange.nextRemoved(key, value);
+                    change = bulkChange;
+                } else if (change instanceof BulkChange.Remove<K, V> bulkChange) {
+                    bulkChange.nextRemoved(key, value);
+                } else {
+                    change = new SimpleChange(key, value, null, false, true);
+                }
+
+                it.remove();
+            }
+        }
+
+        if (change == null) {
+            return false;
+        }
+
+        if (change instanceof BulkChange<K, V> bulkChange) {
+            bulkChange.complete();
+        }
+
+        callObservers(change);
+        return true;
     }
 
     @Override
@@ -243,7 +319,7 @@ public class ObservableMapWrapper<K, V> implements ObservableMap<K, V>{
         return backingMap.hashCode();
     }
 
-    private class ObservableKeySet implements Set<K>{
+    private class ObservableKeySet implements Set<K>, ContainsPredicate<K, V> {
 
         @Override
         public int size() {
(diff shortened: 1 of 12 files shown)


modules/javafx.base/src/test/java/test/javafx/collections/ObservableMapTest.java
```diff
@@ -182,6 +180,21 @@ public void testPutAll(Callable<ObservableMap<String, String>> mapFactory) throw
...
+        setUp(mapFactory);
+
+        observableMap.replaceAll((key, value) -> switch (key) {
+            case "one" -> "2";
+            case "two" -> "3";
+            default -> value;
+        });
+
+        observer.assertMultipleCalls(call("one", "1", "2"), call("two", "2", "3"));
+    }
```
andy-goryachev-oracle, line 196: From the description:

```
The implementation is fully backwards-compatible for listeners that
are unaware of the new API. If the next() method is not called, then
all subsequent changes are delivered as usual by repeated listener
invocations.

If a listener only fetches some changes of a bulk operation (but stops
halfway through the operation), the remaining changes will also be
delivered with repeated listener invocations.
```

1. should this be included in javadoc?
2. could we add a test for this scenario?

andy-goryachev-oracle:
I'll try to finish the review this week.
/keepalive

andy-goryachev-oracle:
Before going into a full review, I'd like t ask this:
1. please enumerate all the bulk methods in `Map` and `Set` that support the new behavior in the description and possibly in the javadoc
2. do we have tests that cover all the bulk methods, exercising the following three scenarios:
- `next()` is not called, received all changes individually (probably so, as it is the current behavior)
- partial retrieval scenario where the remaining changes are received via individual events as described in javadoc and the description
- all changes received via the new methods

mstr2:
> Before going into a full review, I'd like t ask this:
> 
> 1. please enumerate all the bulk methods in `Map` and `Set` that support the new behavior in the description and possibly in the javadoc

`ObservableSetWrapper` and `ObservableMapWarpper` support bulk change notifications for all bulk operations. Since these classes aren't public API, the only place where we could document this for users is in `FXCollections.observableSet()` and `FXCollections.observableMap()`. I wonder if we want to document it explicitly, since it is supported in the entire framework by default. The only specification surface is then the new method on `SetChangeListener.Change` and `MapChangeListener.Change`.

> 2. do we have tests that cover all the bulk methods, exercising the following three scenarios:
> 
> * `next()` is not called, received all changes individually (probably so, as it is the current behavior)

The bulk operation tests are excercised both for bulk retrieval and individual retrieval. See `ObservableSetWrapperTest.TestObservableSetWrapper`.

> * partial retrieval scenario where the remaining changes are received via individual events as described in javadoc and the description

`ObservableSetWrapper.partialChangeIterationCausesSubsequentListenerInvocation`

> * all changes received via the new methods

Yes, see above.

modules/javafx.base/src/main/java/com/sun/javafx/collections/ObservableMapWrapper.java
```diff
@@ -188,22 +130,156 @@ public V remove(Object key) {
...
+            var change = new BulkChange.AddReplace<>(this, size);
+
+            for (Map.Entry<? extends K, ? extends V> e : m.entrySet()) {
+                K key = e.getKey();
+                V newValue = e.getValue();
+
+                if (!backingMap.containsKey(key)) {
+                    change.nextAdded(key, newValue);
+                } else {
+                    V oldValue = backingMap.get(key);
```
andy-goryachev-oracle, line 148: `containsKey()` followed by a `get()` - do you think it's possible to optimize this to avoid looking up twice?  maybe use `get() == null` ?

modules/javafx.base/src/main/java/com/sun/javafx/collections/ObservableMapWrapper.java
```diff
@@ -736,4 +776,220 @@ public int hashCode() {
...
+                builder.append(change.getValueAdded()).append(" added");
+            }
+        } else {
+            builder.append(change.getValueRemoved()).append(" removed");
+        }
+
+        return builder.append(" at key ").append(change.getKey()).toString();
+    }
+
+    private class SimpleChange extends MapChangeListener.Change<K,V> {
```
andy-goryachev-oracle, line 777: could this and `BulkChange` classes made static?

modules/javafx.base/src/main/java/com/sun/javafx/collections/ObservableSetWrapper.java
```diff
@@ -312,12 +413,36 @@ public boolean containsAll(Collection<?> c) {
...
+                    addedElement = element;
+                }
+            }
         }
-        return ret;
+
+        if (addedElement != null) {
+            callObservers(new SimpleAddChange(addedElement));
+            return true;
+        }
```
andy-goryachev-oracle, line 337: minor suggestion: 
```
if (addedElement != null) {
 ..
} else if (addedList != null) {
 ..
```

modules/javafx.base/src/main/java/com/sun/javafx/collections/ObservableSetWrapper.java
```diff
@@ -363,16 +488,38 @@ public boolean removeAll(Collection<?> c) {
...
             }
         }
-        return removed;
+
+        if (removedElement != null) {
+            callObservers(new SimpleRemoveChange(removedElement));
+            return true;
+        }
+
+        if (removedList != null) {
```
andy-goryachev-oracle, line 416: same suggestion with `else`

modules/javafx.base/src/main/java/com/sun/javafx/collections/ObservableSetWrapper.java
```diff
@@ -382,9 +529,16 @@ private boolean removeRetain(Collection<?> c, boolean remove) {
      */
     @Override
     public void clear() {
-        for (Iterator<E> i = backingSet.iterator(); i.hasNext(); ) {
-            E element = i.next();
-            i.remove();
+        if (backingSet.size() > 1) {
```
andy-goryachev-oracle: minor suggestion:
```
switch(backing.size()) {
case 0:
 break;
case 1:
 ..
default:
 ..
}
```
though the compiler is probably going to optimize the second call to `size()` out.

modules/javafx.base/src/main/java/com/sun/javafx/collections/SetListenerHelper.java
```diff
@@ -157,11 +157,13 @@ protected SetListenerHelper<E> removeListener(SetChangeListener<? super E> liste
...
-            } catch (Exception e) {
-                Thread.currentThread().getUncaughtExceptionHandler().uncaughtException(Thread.currentThread(), e);
-            }
+            do {
+                try {
+                    listener.onChanged(change);
+                } catch (Exception e) {
+                    Thread.currentThread().getUncaughtExceptionHandler().uncaughtException(Thread.currentThread(), e);
+                }
+            } while (change instanceof IterableSetChange<? extends E> c && c.nextChange());
```
andy-goryachev-oracle: this is also minor: we are doing instanceof in a loop, perhaps we just need to add a second code path:
```
if(change instanceof IterableSetChange c) {
 try {...}
} else {
 try {..}
```

modules/javafx.base/src/main/java/javafx/collections/MapChangeListener.java
```diff
@@ -99,6 +99,34 @@ public ObservableMap<K, V> getMap() {
          */
         public abstract V getValueRemoved();
 
+        /**
+         * Gets the next change in a series of changes.
+         * <p>
+         * Repeatedly calling this method allows a listener to fetch all subsequent changes of a bulk
+         * map modification that would otherwise be reported as repeated invocations of the listener.
+         * If the listener only fetches some of the pending changes, the rest of the changes will be
```
andy-goryachev-oracle, line 107: suggestion: 
If the listener only fetches some of the pending changes **via the XX() method**
?

modules/javafx.base/src/main/java/javafx/collections/MapChangeListener.java
```diff
@@ -99,6 +99,34 @@ public ObservableMap<K, V> getMap() {
...
+         * reported with subsequent listener invocations.
+         * <p>
+         * After this method has been called, the current {@code Change} instance is no longer valid and
```
andy-goryachev-oracle, line 110: The language might be a bit confusing (it is to me, at least, where you say it might be the same instance but it is no longer valid?  how is that even possible?).

Perhaps it could be rephrased to something like 'the Change instance returned by this method might be a different and must be used in subsequent operations' or something to that effect.

modules/javafx.base/src/test/java/test/com/sun/javafx/collections/ObservableMapWrapperTest.java
```diff
@@ -26,21 +26,118 @@
...
 import java.util.Set;
 
 import static org.junit.jupiter.api.Assertions.*;
 
 public class ObservableMapWrapperTest {
 
+    @Test
+    public void partialChangeIterationCausesSubsequentListenerInvocation() {
+        var trace = new ArrayList<String>();
+        var invocations = new int[1];
```
andy-goryachev-oracle, line 49: minor suggestion: an `AtomicInteger` is semantically better than an array

modules/javafx.base/src/test/java/test/com/sun/javafx/collections/ObservableMapWrapperTest.java
```diff
@@ -26,21 +26,118 @@
...
+
+        assertEquals(3, invocations[0]);
+        assertEquals(
+            List.of(
+                "a added at key k1",
+                "b added at key k2",
+                "c added at key k3",
+                "d added at key k4",
+                "e added at key k5",
+                "f added at key k6"),
```
andy-goryachev-oracle, line 78: this is confusing (to me): I don't see the boundary created by the next() call (i.e. between the bulk and individual changes)

or am I missing something?

modules/javafx.base/src/test/java/test/javafx/collections/ObservableMapTest.java
```diff
@@ -182,6 +180,21 @@ public void testPutAll(Callable<ObservableMap<String, String>> mapFactory) throw
         observer.assertMultipleCalls(call("oFoo", null, "OFoo"), call("pFoo", null, "PFoo"), call("foo", "bar", "foofoo"));
     }
 
+    @ParameterizedTest
+    @MethodSource("createParameters")
+    @SuppressWarnings("unchecked")
+    public void testReplaceAll(Callable<ObservableMap<String, String>> mapFactory) throws Exception {
```
andy-goryachev-oracle, line 186: should `ObservableSetTest` be modified as well?

modules/javafx.base/src/test/java/test/javafx/collections/ObservableMapTest.java
```diff
@@ -182,6 +180,21 @@ public void testPutAll(Callable<ObservableMap<String, String>> mapFactory) throw
...
+        setUp(mapFactory);
+
+        observableMap.replaceAll((key, value) -> switch (key) {
```
andy-goryachev-oracle, line 189: In addition to testing internal details in `ObservableXxxWrapperTests`, shouldn't we also test the bulk change functionality inside this test (i.e. testing the publicly accessible `Set`/`Map` variants)?

I mean, I would rather prioritize testing of the public APIs than some internal implementation.

modules/javafx.base/src/test/java/test/javafx/collections/ObservableMapTest.java
```diff
@@ -182,6 +180,21 @@ public void testPutAll(Callable<ObservableMap<String, String>> mapFactory) throw
...
+            case "one" -> "2";
+            case "two" -> "3";
+            default -> value;
+        });
+
+        observer.assertMultipleCalls(call("one", "1", "2"), call("two", "2", "3"));
```
andy-goryachev-oracle, line 195: Also, what I think might still be missing is the tests for all bulk methods (`clear`, `merge`, `putAll`, `replaceAll`) in 3 flavors: 
1. the legacy (no `next()`, individual events)
2. partial, one `next()` call followed by individual events
3. real bulk processing (while last)

same for `Set`s.

modules/javafx.graphics/src/main/java/com/sun/javafx/application/preferences/PlatformPreferences.java
```diff
@@ -76,8 +75,7 @@ public final class PlatformPreferences extends AbstractMap<String, Object> imple
     /** Contains the implementation of the property-based API. */
     private final PreferenceProperties properties = new PreferenceProperties(this);
 
-    private final List<InvalidationListener> invalidationListeners = new CopyOnWriteArrayList<>();
-    private final List<MapChangeListener<? super String, Object>> mapChangeListeners = new CopyOnWriteArrayList<>();
+    private MapListenerHelper<String, Object> helper;
```
andy-goryachev-oracle, line 78: I wonder if this change should be undertaken as a separate PR, it's a totally unrelated functionality.

andy-goryachev-oracle:
oh, could you please merge the latest master branch?

modules/javafx.base/src/main/java/com/sun/javafx/collections/ObservableMapWrapper.java
mstr2: This is possible if we know that the map doesn't contain null mappings. But since `ObservableMapWrapper` is a wrapper around an arbitray `Map`, we need to account for null mappings. As a consequence of that, we don't know if `map.get(key) == null` means that no mapping is present, or if a mapping is present but its value is `null`.

modules/javafx.base/src/main/java/com/sun/javafx/collections/ObservableMapWrapper.java
mstr2: Yes, I can do that for the existing `Change` implementations, but it's not directly related to bulk change notifications. Maybe another PR would be better.

modules/javafx.base/src/main/java/com/sun/javafx/collections/ObservableSetWrapper.java
mstr2: I've factored out the repeated call to `backingSet.size()`, and reordered the branches such that == 1 comes before > 1. This reads more cleanly now, and I think is also a bit better than a switch block.

modules/javafx.base/src/main/java/com/sun/javafx/collections/SetListenerHelper.java
mstr2: Yes, changed this here and also in `MapListenerHelper`.

modules/javafx.base/src/main/java/javafx/collections/MapChangeListener.java
mstr2: Hmm... there's no other method to call in order to fetch more changes. I could add "If the listener only fetches some of the pending changes **with the next() method**". But then again, just in the sentence before we explained that fetching more changes is done by "repeatedly calling this method" (i.e. the `next()` method).

modules/javafx.base/src/test/java/test/com/sun/javafx/collections/ObservableMapWrapperTest.java
mstr2: This test demonstrates that the sequence of changes is the same, no matter the number of change listener invocations. For this purpose, we assert that the number of changes is as expected (6), but the number of invocations is only 3. In this sense, not seeing a boundary between the invocations is the point.

modules/javafx.base/src/main/java/javafx/collections/MapChangeListener.java
mstr2: Yes, I see that this can be confusing. What I want to bring across is that:
1. You are not allowed to retain a reference to the "last" change instance and expect it to give you correct information after `next()` has been called.
2. You are not allowed to assume, just because the next change instance is the same object as the previous instance, that we're dealing with the same change.

How about this version:
```
 * Callers must not make any assumptions about the identity of the {@code Change} instance
 * returned by this method; even if the returned instance is the same as the current instance,
 * it must be treated as a distinct logical change. Callers must also not retain a reference
 * to a previous {@code Change} instance and expect it to be valid after this method has been
 * called. Calling any method on an earlier change may result in undefined behavior.
```

mstr2:
There's one behavioral change with this enhancement that I want to point out.

Previously, bulk operations were just repeated invocations of their single-element counterpart, i.e. `ObservableSetWrapper.addAll(Collection<E>)` would just repeatedly call `ObservableSetWrapper.add(E)` and thereby trigger repeated listener invocations.

Now consider this method:

```java
public interface SetChangeListener<E> {

    abstract class Change<E> {
        /**
         * An observable set that is associated with the change.
         * @return the source set
         */
        public ObservableSet<E> getSet() { ... }
    }
}
```

Previously, if a listener were to query the source set, it would always correspond to the _current_ state of the set.
With the bulk change enhancement, the source set corresponds to the _final_ state of the set (because all changes have already been applied in bulk).

modules/javafx.base/src/main/java/com/sun/javafx/collections/ObservableMapWrapper.java
andy-goryachev-oracle: since you are already here...  saves a pointer maybe

kevinrushforth: "since you are already here" is not usually sufficient justification for fixing an unrelated bug. So it probably makes sense to do for any new class added, but I'd leave the existing ones alone.

andy-goryachev-oracle: I thought it **is** a new class.  But sure, I am ok to keep the code as is.

modules/javafx.base/src/main/java/javafx/collections/MapChangeListener.java
andy-goryachev-oracle: this is still a bit confusing... "identity" where did that come from?

maybe just tell them what they need to do instead: discard the last instance and use the one returned by `next()` instead?  what do you think?

modules/javafx.base/src/test/java/test/com/sun/javafx/collections/ObservableMapWrapperTest.java
andy-goryachev-oracle: Ok, so if I were to write a test per the public API, two things should be tested: the changes and the bulk aspect.  So perhaps the toString() or the test's toString() can be modified to explicitly encode the bulk aspect.  In other words, the test might look like this:
```
expect([k3=c, k4=d],[k5=e],[k6=f])
```
if the test calls next on k3 and k4, then individual events come for k5, k6.
This way you test both API functions in one test.

andy-goryachev-oracle:
> There's one behavioral change with this enhancement that I want to point out.
we probably need to add this explanation to `SetChangeListener.Change.getSet()` javadoc.  I assume the same goes for `MapChangeListener.Change.getMap()` ?

andy-goryachev-oracle:
What about the ordering of bulk changes?  Should it be addressed in javadoc for bulk changes (in next())?

The ordering might impact tests.

modules/javafx.base/src/main/java/com/sun/javafx/collections/ObservableSetWrapper.java
```diff
@@ -529,17 +428,19 @@ private boolean removeRetain(Collection<?> c, boolean remove) {
...
             Iterator<E> it = backingSet.iterator();
             E element = it.next();
             it.remove();
             callObservers(new SimpleRemoveChange(element));
+        } else if (size > 1) {
+            @SuppressWarnings("unchecked")
+            E[] removed = (E[])new Object[size];
+            backingSet.toArray(removed);
+            backingSet.clear();
+            callObservers(new IterableSetChange.Remove<>(this, Arrays.asList(removed)));
```
andy-goryachev-oracle, line 443: suggestion:
```
            ArrayList<E> removed = new ArrayList<>(backingSet);
            backingSet.clear();
            callObservers(new IterableSetChange.Remove<>(this, removed));
```

PR #1897, opened 2025-09-11
8336332: Rework tests to avoid unrelated stderr output

andy-goryachev-oracle:
This PR removes unrelated `stderr` output in the headless test logs by redirecting it to an in-memory buffer.  Exceptions found in the buffer can be checked against the expected list.

In the case when any mismatch is detected, whether the type or the number of exceptions of particular type, the accumulated buffer gets dumped to `stderr` and the test fails.

## How To

To redirect stderr and later check the exceptions, surround your code with

`OutputRedirect.suppressStderr()` and either `OutputRedirect.checkStderr()` or `OutputRedirect.checkAndRestoreStderr()` (ideally, in the `finally` block).

To simply undo redirection, without checking, call  `OutputRedirect.restoreStderr()`.

To add the check to all the tests in the file, one can call the above mentioned methods inside  `@BeforeEach` and `@AfterEach`.

## Changes

- added `OutputRedirect` facility

## Miscellaneous

`ErrorLoggingUtiltity` name will be fixed in a followup https://bugs.openjdk.org/browse/JDK-8367995

---------
### Progress

### Issue
 * [JDK-8336332](https://bugs.openjdk.org/browse/JDK-8336332): Rework tests to avoid unrelated stderr output (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1897/head:pull/1897` \
`$ git checkout pull/1897`

`$ git checkout pull/1897` \
`$ git pull https://git.openjdk.org/jfx.git pull/1897/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1897`

`$ git pr show -t 1897`

</details>

<a href="https://git.openjdk.org/jfx/pull/1897.diff">https://git.openjdk.org/jfx/pull/1897.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1897#issuecomment-3285869212)
</details>

### DIFF
--- a/modules/javafx.base/src/test/java/test/com/sun/javafx/binding/SelectBindingTest.java
+++ b/modules/javafx.base/src/test/java/test/com/sun/javafx/binding/SelectBindingTest.java
@@ -107,7 +107,7 @@ public class SelectBindingTest {
     private ObservableList<?> dependencies;
 
     @BeforeAll
-    public static void setUpClass() throws Exception {
+    public static void setUpClass() {
         ErrorLoggingUtiltity.reset();
     }
 


kevinrushforth:
I see that you focused on javafx.base, which is where many of the problems are. Do you plan to update this PR for the other modules and for system tests (the latter also has several exceptions and lots of noise), or will you file a follow-up issue to take up after this one is done?

andy-goryachev-oracle:
good point.
no, this PR is only for the headful tests as they cause the most grief when analyzing the failures.

mstr2:
The `ErrorLoggingUtility` class manages static state, and requires users to follow a protocol to avoid messing up its internal state. Have you considered using a try-with-resources scope to make the API easier to use, for example:

```java
try (var util = ErrorLoggingUtility.suppressStdErr()) {
    // test something
    // ...
    //

    util.checkWarning(NullPointerException.class);
} // internal state is restored at the end of the block
```

andy-goryachev-oracle:
Good point.  It already has a static state (error log record), and on top of that try with resources will not work in the bulk use case of `@BeforeEach` and `@AfterEach`

Plus, it's a test utility.

I did try to explain the usage in javadoc - maybe that can be further clarified?

mstr2:
> Good point. It already has a static state (error log record), and on top of that try with resources will not work in the bulk use case of `@BeforeEach` and `@AfterEach`

Why wouldn't it work? If you _really_ need to, you can store the `AutoClosable` in `@BeforeEach` and then call the close method in `@AfterEach`. But I would also prefer to not scatter around code in before-each and after-each methods, but to have each test be as stand-alone as possible, especially when it comes to scoping.

Whether `ErrorLoggingUtility` has static state is also immaterial (that would only be important if tests were to be run in parallel), what matters is the API. Using an auto-closable scope has the following advantages:
1. Using the tool is easy, as you can only call methods that are applicable for the state of the test. If you want to check the log output, you need to have an active scope instance.
2. It can't be misused, as you can't call methods in the wrong order.
3. The static state will always be well-defined.

Yes, it's only a test utility, but the API is really suboptimal no matter how I look at it.

andy-goryachev-oracle:
These are all valid points, but keep in mind that `System.setErr()` is static to begin with.  All the system and headful tests are single threaded by design, and this is rather unlikely to change.

Whether to use BeforeEach/AfterEach or standalone test is up to the test developers (this PR contains both cases actually).  I would say re-designing the existing tests to be standalone (`SelectBindingTest` for example) is out of scope for this PR.

kevinrushforth:
Two general comments before I start looking at the PR:

1. Since you are only addressing the `javafx.base` tests in this PR, please update the JBS title (and PR title) to reflect that.

2. In the description you wrote:

> This PR removes unrelated stderr output in the headful test logs

These are "headless" tests, not headful.

modules/javafx.base/src/test/java/test/util/ErrorLoggingUtility.java
```diff
@@ -0,0 +1,232 @@
...
+        assertNotNull(errorLogRecord);
+        assertEquals(expectedLevel, errorLogRecord.getLevel());
+        assertTrue(expectedException.isAssignableFrom(errorLogRecord.getThrown().getClass()));
+        reset();
+        checked++;
+    }
+
+    private static PrintStream stderr;
+    private static AccumulatingPrintStream stderrCapture;
+    private static int checked;
```
kevinrushforth: This is unused (only ever written) and can be removed.

andy-goryachev-oracle:
> 1. Since you are only addressing the `javafx.base`

so that's where it gets complicated: turns out the utility gets used in more tests (graphics, fxml, see https://github.com/openjdk/jfx/pull/1905 ), so two things came up:

1. the utility needs to be moved to another package
2. the utility probably needs to be decoupled from Logging, as it creates unneeded dependency with `--add-exports javafx.base/com.sun.javafx.binding=ALL-UNNAMED` in #1905 
3. perhaps the two PRs need to be merged into one larger one

What do you think?

modules/javafx.base/src/test/java/test/util/ErrorLoggingUtility.java
```diff
@@ -0,0 +1,232 @@
...
+    ///
+    public static void checkStderr(Class<? extends Throwable> ... expected) {
+        if (stderrCapture != null) {
+            String text = stderrCapture.getAccumulatedOutput();
+            Map<String,Integer> errors = findErrors(text);
+            Map<String,Integer> exp = toMap(expected);
+            if(!errors.equals(exp)) {
+                stderr.println("Mismatch in thrown exceptions:\n  expected=" + exp + "\n  observed=" + errors);
+                stderr.println(text);
+            }
```
kevinrushforth: A mismatch should fail the test.

kevinrushforth:
> > 1. Since you are only addressing the `javafx.base`
> 
> so that's where it gets complicated: turns out the utility gets used in more tests (graphics, fxml, see #1905 ), so two things came up:
> 
> 1. the utility needs to be moved to another package
> 2. the utility probably needs to be decoupled from Logging, as it creates unneeded dependency with `--add-exports javafx.base/com.sun.javafx.binding=ALL-UNNAMED` in [8367567: Rework system tests to suppress unrelated stderr output #1905](https://github.com/openjdk/jfx/pull/1905)

Yes, I think this seems best. Rather than extend the existing javafx.base ErrorLoggingUtility, which is dependent on the (IMO rather hacky) Logging class in javafx.base itself, a separate test utility that just focuses on capturing and parsing stderr and has no dependencies on logging seems best.

> 3. perhaps the two PRs need to be merged into one larger one

Maybe that would be best, although I think it would be OK to keep them separate as long as you do a proof of concept that other modules can use it without needing changes to the exports, etc.

modules/javafx.base/src/test/java/test/util/ErrorLoggingUtility.java
```diff
@@ -0,0 +1,232 @@
...
+                }
+            });
+        return m;
+    }
+
+    private static final Pattern EXCEPTION_PATTERN = Pattern.compile(
+        "(?:" +
+            // catches lines starting with things like "Exception in thread "main" java.lang.RuntimeException:"
+            "^" +
+            "(?:" +
```
kevinrushforth: This might be easier to read and maintain with text blocks.

andy-goryachev-oracle: it's a complete gibberish as continuous text.  not an easy thing to parse mentally.

andy-goryachev-oracle:
thanks for suggestions!  I'll revert back to DRAFT since there will be more structural changes, then we'll see if we can separate the two PRs or combine them.

modules/javafx.base/src/test/java/test/com/sun/javafx/binding/ErrorLoggingUtility.java
```diff
@@ -25,15 +25,15 @@
...
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import com.sun.javafx.binding.Logging;
-import com.sun.javafx.binding.Logging.ErrorLogger.ErrorLogRecord;
 import com.sun.javafx.binding.Logging.ErrorLogger;
+import com.sun.javafx.binding.Logging.ErrorLogger.ErrorLogRecord;
+import com.sun.javafx.logging.PlatformLogger.Level;
 
-public class ErrorLoggingUtiltity {
+public class ErrorLoggingUtility {
```
kevinrushforth: I see you left in the name change to `ErrorLoggingUtility`. I was hoping you'd revert that along with the other changes to this file, since they are now unrelated and make the PR much larger to review. Yes, I know it's a mechanical change, but I find it hard to see what has "really" changed now.

Suggestion: revert the name change to this utility, since it is no longer relevant.

kevinrushforth: Further, there are at least some files in javafx.base that are unchanged except for this name change, a reordering of imports, and a copyright year update. See `modules/javafx.base/src/test/java/test/javafx/binding/BindingsArrayTest.java` which has a large number of changes, but I think the whole file would be reverted and untouched (meaning no need for 2025) if you revert the name change.

andy-goryachev-oracle: yikes.  to be be renamed in [JDK-8367995](https://bugs.openjdk.org/browse/JDK-8367995)

modules/javafx.base/src/test/java/test/com/sun/javafx/binding/SelectBindingTest.java
```diff
@@ -101,14 +105,11 @@ public void removePropertyChangeListener(String property, PropertyChangeListener
...
-    @BeforeAll
-    public static void setUpClass() {
-        System.err.println("SelectBindingTest : log messages are expected from these tests.");
-        ErrorLoggingUtiltity.reset();
-    }
-
     @BeforeEach
     public void setUp() throws Exception {
+        OutputRedirect.suppressStderr();
+        ErrorLoggingUtiltity.reset();
```
kevinrushforth: Moving the call to `reset` to a `@BeforeEach` method is an unrelated change. It seems safer to restore the `@BeforeAll` method with just the call to reset.

modules/javafx.base/src/test/java/test/javafx/binding/BindingsCreateBindingTest.java
```diff
@@ -67,11 +67,17 @@ private static interface Functions<S> {
     private  T value1;
     private  T defaultValue;
 
-    @BeforeAll
-    public static void setUpClass() {
+    @BeforeEach
+    public void beforeEach() {
+        OutputRedirect.suppressStderr();
         ErrorLoggingUtiltity.reset();
```
kevinrushforth: Same comment as in earlier class: it seems best to leave the reset in a `@BeforeAll` method, unless there is a compelling reason to change it.

modules/javafx.base/src/test/java/test/javafx/util/OutputRedirect.java
```diff
@@ -0,0 +1,234 @@
...
+    public static void restoreStderr() {
+        if (stderr != null) {
+            System.setErr(stderr);
+            stderr = null;
+            stderrCapture = null;
+        }
+    }
+
+    /// Returns the captured output, if any, or `null`.
+    /// @return the captured output string, or `null`
```
kevinrushforth: The empty string might be a better choice if there is no output.

modules/javafx.base/src/test/java/test/javafx/util/OutputRedirect.java
```diff
@@ -0,0 +1,234 @@
...
+    public static void checkStderr(Object ... expected) {
+        if (stderrCapture != null) {
+            boolean err = false;
+            String text = stderrCapture.getAccumulatedOutput();
+
+            // exceptions
+            Map<String, Integer> errors = findErrors(text);
+            Map<String, Integer> exp = toMap(expected);
+            if (!errors.equals(exp)) {
+                stderr.println("Mismatch in thrown exceptions:\n  expected=" + exp + "\n  observed=" + errors);
```
kevinrushforth, line 93: This should cause the test to fail. Set `err = true` here.

modules/javafx.base/src/test/java/test/javafx/util/OutputRedirect.java
```diff
@@ -0,0 +1,234 @@
...
+                if (Throwable.class.isAssignableFrom(c)) {
+                    String name = c.getName();
+                    Integer v = m.get(name);
+                    if (v == null) {
+                        m.put(name, Integer.valueOf(1));
+                    } else {
+                        m.put(name, Integer.valueOf(v + 1));
+                    }
+                } else {
+                    throw new IllegalArgumentException("must specify either Class<? extends Throwable>: " + c);
```
kevinrushforth: Typo: remove the "either" since there isn't another one in this case.

modules/javafx.base/src/test/java/test/javafx/util/OutputRedirect.java
```diff
@@ -0,0 +1,234 @@
...
+            name = m.group(1);
+            if (name == null) {
+                name = m.group(2);
+            }
+            return name;
+        }
+        return null;
+    }
+
+    // should I leave this test here?  to test the test?
```
kevinrushforth: Seems reasonable.

modules/javafx.base/src/test/java/test/javafx/util/OutputRedirect.java
```diff
@@ -0,0 +1,234 @@
...
+    public static void checkAndRestoreStderr(Object ... expected) {
+        try {
+            checkStderr(expected);
+        } finally {
+            restoreStderr();
+        }
+    }
+
+    private static Map<String, Integer> toMap(Object... expected) {
+        HashMap<String, Integer> m = new HashMap<>();
```
kevinrushforth, line 132: Suggestion: Given how they are used, you might consider using a `Set` (which could be created as either a `LinkedHashSet` or `TreeSet`) instead. This would simplify the logic a bit.

tests/system/src/test/java/test/com/sun/javafx/application/SingleExitCommon.java
```diff
@@ -137,13 +138,32 @@ private void doTestCommon(boolean implicitExit,
             boolean reEnableImplicitExit, boolean stageShown,
             boolean appShouldExit) {
 
-        doTestCommon(implicitExit, reEnableImplicitExit, stageShown,
-                ThrowableType.NONE, appShouldExit);
+        doTestCommon(implicitExit, reEnableImplicitExit, stageShown, ThrowableType.NONE, appShouldExit);
```
kevinrushforth: Minor: This is an unrelated formatting change in a method that is otherwise untouched.

tests/system/src/test/java/test/launchertest/ModuleLauncherTest.java
```diff
@@ -56,7 +57,16 @@ public class ModuleLauncherTest {
 
     private final int testExitCode = ERROR_NONE;
 
-    private void doTestLaunchModule(String appModulePath, String testAppName) throws Exception {
+    private void doTestLaunchModule(String appModulePath, String testAppName, Object ... expected) throws Exception {
```
kevinrushforth: No caller of this method passes anything for the newly added `Object ... expected` varargs parameter. I recommend reverting the addition of the parameter.

andy-goryachev-oracle:
regex: I can try commenting what each part means

modules/javafx.base/src/test/java/test/javafx/util/OutputRedirect.java
andy-goryachev-oracle: this method is not used anymore, will remove.

modules/javafx.base/src/test/java/test/javafx/util/OutputRedirect.java
andy-goryachev-oracle: No, I need the counts.
JDK has neither `MultiMap`s nor counting sets.

kevinrushforth: Oh, nm. I missed that the value is a count of that particular exception class.

modules/javafx.base/src/test/java/test/javafx/util/OutputRedirect.java
```diff
@@ -0,0 +1,232 @@
...
+                    Integer v = m.get(name);
+                    if (v == null) {
+                        m.put(name, Integer.valueOf(1));
+                    } else {
+                        m.put(name, Integer.valueOf(v + 1));
+                    }
+                } else {
+                    throw new IllegalArgumentException("must specify Class<? extends Throwable>: " + c);
+                }
+            } else if(x instanceof String) {
```
kevinrushforth: Minor: space after `if`

modules/javafx.base/src/test/java/test/com/sun/javafx/binding/SelectBindingTest.java
```diff
@@ -102,13 +107,14 @@ public void removePropertyChangeListener(String property, PropertyChangeListener
     private ObservableList<?> dependencies;
 
     @BeforeAll
-    public static void setUpClass() {
-        System.err.println("SelectBindingTest : log messages are expected from these tests.");
+    public static void setUpClass() throws Exception {
```
kevinrushforth: Minor: this method previously didn't have `throws Exception` and probably doesn't need it.

modules/javafx.base/src/test/java/test/javafx/util/OutputRedirect.java
kevinrushforth: Although, I would remove or reword the comment asking whether you should leave it if you've decided to do so.

modules/javafx.base/src/test/java/test/javafx/util/OutputRedirect.java
```diff
@@ -0,0 +1,232 @@
...
+        return m;
+    }
+
+    private static Map<String, Integer> findErrors(String text) {
+        HashMap<String, Integer> m = new HashMap<>();
+        text.
+            lines().
+            map((s) -> findException(s)).
+            forEach((c) -> {
+                if (c != null) {
```
kevinrushforth: Minor suggestion: filter for non-null Strings before the `forEach` rather than checking for non-null in the `forEach` ?

modules/javafx.base/src/test/java/test/javafx/util/OutputRedirect.java
```diff
@@ -0,0 +1,232 @@
...
+
+    /// This regex matches either of the two patterns which might appear in the output:
+    ///
+    /// `Exception in thread "main" java.lang.RuntimeException:`
+    ///
+    /// or
+    ///
+    /// `java.lang.NullPointerException: ...`
+    private static final Pattern EXCEPTION_PATTERN = Pattern.compile(
+        "(?:" +
```
kevinrushforth, line 179: Looks reasonable to me. :)

kevinrushforth (commented):
Last batch of comments. Initial testing looks good.

modules/javafx.base/src/test/java/test/javafx/util/OutputRedirect.java
andy-goryachev-oracle: my version is _more efficient_ :-)

kevinrushforth: I doubt it. Anyway, it doesn't really matter all that much.

andy-goryachev-oracle: off topic, but it is more efficient - check out

https://github.com/andy-goryachev-oracle/Test/blob/main/src/goryachev/research/MoreEfficient.java

tests/system/src/test/java/test/javafx/scene/Snapshot1Test.java
```diff
@@ -221,22 +222,27 @@ public void testBadSceneCallback1() {
...
+                        // message to be printed.
+                        return "";
+                    }
+                };
+
+                tmpScene.snapshot(cb, null);
+                Util.sleep(SLEEP_TIME);
+            } finally {
+                OutputRedirect.checkAndRestoreStderr(ClassCastException.class);
+            }
```
kevinrushforth: This fails for me, which is not surprising since the snapshot is asynchronous. You probably want to move everything (the suppressStderr and the try/finally with the checkAndRestoreStderr) out of the runAndWait, but even that might not be sufficient without a delay after `latch.getCount()`.

andy-goryachev-oracle: you are right, I missed the `runAndWait()`

tests/system/src/test/java/test/javafx/scene/Snapshot1Test.java
```diff
@@ -217,69 +218,76 @@ public void testSceneCallback() {
...
+            } catch (InterruptedException ex) {
+                fail(ex);
             }
-        } catch (InterruptedException ex) {
-            fail(ex);
-        }
 
-        assertEquals(0, latch.getCount());
+            assertEquals(0, latch.getCount());
+        } finally {
```
kevinrushforth: Since the ClassCast exception happens after the latch countdown, I recommend a sleep (of, say, 100 msec) after checking that the latch count is 0.

tests/system/src/test/java/test/javafx/scene/Snapshot1Test.java
```diff
@@ -339,69 +347,76 @@ public void testNodeCallback() {
...
```
kevinrushforth: Same here: a short sleep after the assertEquals.

kevinrushforth (commented):
Shapshot1Test still fails for me. This time the exceptions are being captured in the print stream, but the parsing of the output reports 2 exceptions instead of the expected 1:

```
    Mismatch in thrown exceptions:
      expected={java.lang.ClassCastException=1}
      observed={java.lang.ClassCastException=1, Exception=1}
```

andy-goryachev-oracle:
"Exception is snapshot callback" is coming from 
Node:2421 and Scene:1612

Why in the world do we have unsuppressible System.err.println() in there??

andy-goryachev-oracle:
I decided to remove Snapshot1Test from this PR: there is enough value added already, and there is more work needed to remove stderr output coming from CssParser anyway, so let's postpone it until the next test sprint.

follow up:
[JDK-8368023](https://bugs.openjdk.org/browse/JDK-8368023)
[TestBug] Rework tests to clean up stderr

modules/javafx.base/src/test/java/test/com/sun/javafx/binding/SelectBindingTest.java
kevinrushforth: Are you going to revert the "throws Exception" ?

modules/javafx.base/src/test/java/test/javafx/util/OutputRedirect.java
kevinrushforth: It might be interesting to see micro-benchmark results, but yeah, off topic.

kevinrushforth (approved):
Looks good. My testing shows that there are no more "failed" messages for a successful build and that the number of Exception messages are reduced. The remainder can be handled by the follow-on bug you filed.

kevinrushforth:
> "Exception is snapshot callback" is coming from Node:2421 and Scene:1612
> 
> Why in the world do we have unsuppressible System.err.println() in there??

We have many such places in the JavaFX runtime. We already have a task filed to examine these: [JDK-8320266](https://bugs.openjdk.org/browse/JDK-8320266).

modules/javafx.base/src/test/java/test/com/sun/javafx/binding/SelectBindingTest.java
andy-goryachev-oracle: done

kevinrushforth:

andy-goryachev-oracle:
thank you for reviewing!

/integrate

PR #1904, opened 2025-09-15
8359108: Mac - When Swing starts First, native application menu doesn't work for JavaFX

pabulaner:
This pull request fixes the system menu bar on MacOS when combining windows of Swing and JavaFX.

# Behavior before

If for some reason You needed to initialize AWT before JavaFX and You wanted to install the system menu bar from the JavaFX side, this wasn't possible. This issue is persistent even when You didn't open a Swing window. 

One scenario where this kind of issue happens is when You use install4j (see https://www.ej-technologies.com/install4j). In this case AWT is initialized by install4j and therefore You can't use the JavaFX system menu bar anymore.

# Behavior after

The fix allows JavaFX to install a system menu bar even if it is initialized after AWT. This is achieved by only changing code inside JavaFX. Each JavaFX window stores the previously installed menu bar when gaining focus and will restore this menu bar if the focus was lost. This only happens if the system menu bar installed by the JavaFX window is still unchanged.

# Tests

This PR introduces tests for the system menu bar in addition to verifying its own behavior / changes. The tests include single- and multi-window tests while interacting with Swing. The tests ensure that the menu bar stays the same for each window, no matter how You switch focus between them.

# Additional benifits 

This fix is not specifically for AWT, but allows JavaFX to interact much more compatibly with other frameworks that make use of the system menu bar.

# Review from AWT

In the previous PR related to this one, the comment was made that the folks from AWT should take a look at this fix. It would be great and much appreciated if someone could initiate it.

# Add disable flag?

We could also add a flag to prevent JavaFX from installing a system menu bar for users who have found other fixes for their projects / setups. This could be used to restore the previous behavior when AWT is initialized first.

Co-Author: @FlorianKirmaier

---------
### Progress

### Issue
 * [JDK-8359108](https://bugs.openjdk.org/browse/JDK-8359108): Mac - When Swing starts First, native application menu doesn't work for JavaFX (**Bug** - P3)

### Reviewers

### Contributors
 * Florian Kirmaier `<fkirmaier@openjdk.org>`

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1904/head:pull/1904` \
`$ git checkout pull/1904`

`$ git checkout pull/1904` \
`$ git pull https://git.openjdk.org/jfx.git pull/1904/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1904`

`$ git pr show -t 1904`

</details>

<a href="https://git.openjdk.org/jfx/pull/1904.diff">https://git.openjdk.org/jfx/pull/1904.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1904#issuecomment-3292992538)
</details>


kevinrushforth:
This approach seems much better than the previous in that it is activated only when a JavaFX Stage becomes the focused (key) window, and it saves / restores the menu such that whenever an different (e.g., AWT) window is focused, the AWT menu will be installed.

In addition to reviewing and checking that it only touches the menu when it should, testing will be critical.

Reviewers: @kevinrushforth @andy-goryachev-oracle @beldenfox
Cc: @prrace @honkar-jdk

/reviewers 3

kevinrushforth:
As mentioned above, testing will be key here. This needs to work reliably in the case of a multi-window app when the application is hidden and restored, and when switching between the windows of the app. The automated tests should already cover some of the cases. There may be others to consider.

Here are a few cases that I can think of to check:

* Pure FX app: 2 FX Stages (AWT toolkit not initialized at all) -- the fix should do nothing
* 2 FX Stages with FX initialized first (AWT is initialized last, but no AWT/Swing windows) -- the fix should do nothing
* 2 FX Stages with AWT initialized first (but no AWT/Swing windows) -- the fix should install the FX window when either Stage is focused (there will not be any AWT menu)
* a mix of AWT Windows and FX Stages with FX initialized first -- FX will install the menu and keep it there, since AWT won't install a menu when it doesn't own the NSApplication
* a mix of AWT Windows and FX Stages with FX initialized first -- AWT menu will be installed when an AWT window is active, FX menu will be installed when an FX Stage is active
* 2 AWT Stages one of which has a JFXPanel with an FX scene, initialize AWT first -- fix should be a noop
* 2 AWT Stages one of which has a JFXPanel with an FX scene, initialize FX first -- fix should be a noop

I presume that the automated tests will cover some of these, but we might want additional tests.

pabulaner:
Thanks for the test case ideas, my tests cover the AWT initialized first tests, but implementing the other tests You proposed won't be that difficult, as it is just copying the code and moving and deleting some lines.
I will get to add those tests probably until tomorrow.

pabulaner:
So I added most of the tests You proposed, except for the last three. 

For the last two I just need some more time to implement them.

For the other test with AWT being initialized last and still installing the menu is something not directly implemented in this fix. I think AWT won't install a system menu if it is running embedded and therefore this would need its own fix inside AWT itself (but I don't know exactly).

pabulaner:
I have now also added the JFXPanel test, but only for the first of the two tests involving the JFXPanel (so where AWT is initialized first). In the other case AWT won't install a menu bar, since JavaFX was initialized first, so nothing will happen.

beldenfox:
I’ve been reviewing how the platform code interacts with the rest of the system. For others who want to take a look at this PR here’s a quick run-down of how the current master branch works.

When JavaFX is not embedded the toolkit builds a single NSMenu to use as the system menu. It installs that NSMenu as NSApp.mainMenu when the first top-level window is shown but before it becomes the key window. This happens inside the _setMenubar call in GlassWindow.m. That menu stays in place as NSApp.mainMenu for the lifetime of the JavaFX application. Subsequent calls to _setMenubar and the logic that sets NSApp.mainMenu inside windowDidBecomeKey: are basically no-ops since it’s always the same NSMenu and it has already been installed.

It might look like the system is setting a separate menubar for each window but it’s not. There’s only one, it stays in place, and the toolkit rebuilds its content as windows gain and lose focus. And it always contains at least one menu, namely the default application menu.

The system also calls _setMenubar to hand the system NSMenu to each PopupWindow when it’s first shown. There doesn’t seem to be any point to this; the NSMenu was already installed by the popup’s owner and MenuBarSkin ignores popups when updating the system menu bar.

When JavaFX is embedded none of the above happens.

This PR doesn’t seem to change any behavior in the non-embedded case. In the embedded case everything changes. JavaFX now builds a system-wide NSMenu and hands it off to each top-level window but it only gets installed when that window becomes key. The menu is un-installed when the window resigns key (and the previous mainMenu is restored). That’s all new behavior but it seems to be working reliably.

Also new with this PR: if a popup is displayed in an FXPanel the system will call _setMenubar to set the system menu. The platform code is handling this correctly (basically ignoring it) but the checks it’s making shouldn’t be necessary. It would be far better if createTKPopupStage stopped calling _setMenubar.

This PR is turning on some code further up in the system that can lead to some minor bugs. For example, if you add a MenuBar to an FXPanel’s scene and then turn on useSystemMenuBar the entire menubar will become inaccessible. This is similar to another bug that’s already present in the master branch; if you add a MenuBar to a PopupWindow and then turn on useSystemMenuBar the entire menubar will disappear. These issues are all contrived and it’s easy for developers to avoid them but it would be good to tighten up the code a bit. The bugs I’ve found so far are minor enough not to block this PR and could be addressed in a separate issue. This should be easy to clean up though I can’t personally provide further assistance until November.

(The primary problem is in MenuBarSkin.setSystemMenu. It crawls up the stage ownership chain to decide which MenuBar to use to populate the system menu bar. It should ignore windows that aren’t eligible for the system menu bar like embedded and popup stages. One possible solution is to call getMenubar on each stage to see if the toolkit initialized it with the system menu.)

Still no show-stoppers, just some unnecessary complications from the way PopupWindows keep trying to force the system menu onto the platform code and a few very minor bugs that no developer is likely to notice.

pabulaner:
Thanks for the detailed feedback, much appreciated. So as far as I understood Your feedback it means that the PR from Your side would be good to go, but maybe in the future some minor bug / issue fixes would be nice to have?

modules/javafx.graphics/src/main/native-glass/mac/GlassWindow.h
```diff
@@ -64,6 +64,11 @@
     BOOL                isSizeAssigned;
     BOOL                isLocationAssigned;
 
+    // Stores the menu of the host application when running embedded.
+    // This is used to allow JFX to install its own system menu
+    // without interfering with the hosting application.
+    NSMenu* hostMenu;
```
kevinrushforth: Minor: align the field name to match the rest of the fields in this file

tests/system/src/test/java/test/javafx/stage/MacOSSystemMenuJFXPanelSwingFirstTest.java
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright (c) 2017, 2024, Oracle and/or its affiliates. All rights reserved.
```
kevinrushforth: Replace `2024` with `2025`. Also, unless this file was substantially copied from an existing file (which it doesn't look like it was), remove the `2017,`.

This applies to all of the new files.

tests/system/src/test/java/test/javafx/stage/MacOSSystemMenuJFXPanelSwingFirstTest.java
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package test.javafx.stage;
```
kevinrushforth: I recommend moving these tests to a new package, perhaps `test.com.sun.glass.ui.mac`, since this is testing macOS glass functionality.

Or maybe, `test.robot...`, since even though they aren't robot tests, they rely on Window focus working.

tests/system/src/test/java/test/javafx/stage/MacOSSystemMenuTestBase.java
```diff
@@ -0,0 +1,427 @@
+package test.javafx.stage;
```
kevinrushforth: This needs a copyright header block.

tests/system/src/test/java/test/javafx/stage/MacOSSystemMenuTestBase.java
```diff
@@ -0,0 +1,427 @@
...
+
+    protected final List<JFrame> swingWindows = new ArrayList<>();
+
+    private CountDownLatch latch = null;
+
+    /////////////////////////////////////////////////////////
+    ///
+    /// Helpers for creation and focusing of windows
+    ///
+    /////////////////////////////////////////////////////////
```
kevinrushforth: Using `///` is reserved for markdown style javadoc comments. Replace this with something else.

tests/system/src/test/java/test/javafx/stage/MacOSSystemMenuTestBase.java
```diff
@@ -0,0 +1,427 @@
...
+        try {
+            latch.await();
+        } catch (InterruptedException e) {}
+    }
+
+    private void releaseLock() {
+        latch.countDown();
+    }
+
+    protected void sleep(long millis) {
```
kevinrushforth: This is unused, and duplicates the functionality of `Util.sleep` anyway. I recommend to remove it.

tests/system/src/test/java/test/javafx/stage/MacOSSystemMenuTestBase.java
```diff
@@ -0,0 +1,427 @@
...
+
+            return result.toString();
+        }
+    }
+
+    /////////////////////////////////////////////////////////
+    ///
+    /// Helpers for synchronization
+    ///
+    /////////////////////////////////////////////////////////
```
kevinrushforth: Replace `///` with something else.

tests/system/src/test/java/test/javafx/stage/MacOSSystemMenuTestBase.java
```diff
@@ -0,0 +1,427 @@
...
+import java.util.concurrent.CountDownLatch;
+import java.util.stream.Collectors;
+
+import javax.swing.JFrame;
+import javax.swing.JMenu;
+import javax.swing.JMenuBar;
+import javax.swing.JMenuItem;
+import javax.swing.SwingUtilities;
+
+import org.junit.jupiter.api.Test;
```
kevinrushforth: This is an unused import and should be removed. Especially since if there ever were a method in this class annotated with `@Test` it wouldn't work correctly since the tests in this test group are written so that there can be no more than one `@Test` in a test class, including its superclasses.

kevinrushforth (commented):
This is a very preliminary review. I haven't run it yet, but I will soon. I left a few inline comments of things I happened to spot.

The fix itself looks good to me. 

The test framework and set of tests looks fine on first glance, but I haven't looked at them in detail. I will do so, along with running them

pabulaner:
Thanks for the improvements. I should have implemented all of them, please take a look again.

beldenfox:
> Thanks for the detailed feedback, much appreciated. So as far as I understood Your feedback it means that the PR from Your side would be good to go, but maybe in the future some minor bug / issue fixes would be nice to have?

I need to do a more detailed review of the platform code but for now I haven't seen any problems with it. To review the changes to, say, _setMenubar I needed to first understand when it is called and that's what I was focusing on. 

There is an unexpected testing burden. When a popup appears in an FXPanel (as a tooltip, combo box, or menu) the core code will try to set the system menu bar for no good reason and it's up to the platform code to make sure it doesn't succeed. I'm not sure it warrants a system test but some testing needs to be done for these cases. This core behavior really should be addressed in a follow up issue but that's not something I can do until November.

tests/manual/systemmenu/MacOSSystemMenuTestBase.java
```diff
@@ -0,0 +1,444 @@
...
+import javax.swing.JMenu;
+import javax.swing.JMenuBar;
+import javax.swing.JMenuItem;
+import javax.swing.SwingUtilities;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+
+public class MacOSSystemMenuTestBase {
```
kevinrushforth, line 59: Somewhere in this base class, add a static method annotated with `@BeforeAll` that does an `assumeTrue(PlatformUtil.isMac())` to ensure that these tests don't execute on the other platforms.

kevinrushforth (commented):
I ran the tests today on my Mac M1 system. I left one comment inline about needing to ensure that these tests only run on macOS.

The following two tests fail consistently for me:

```
MacOSSystemMenuMultiWindowFXOnlySwingLast
MacOSSystemMenuMultiWindowTest
```

The following fail intermittently:

```
MacOSSystemMenuSetWithSwingTestFirst
MacOSSystemMenuJFXPanelSwingFirstTest
```

Also, I noticed that running all 7 tests takes over 6 minutes on my system. I tracked down the reason to the large number of calls to fork the `osascript` process.

Here is the count by test:


Test name | Number of calls
--------- | ---------------
MacOSSystemMenuJFXPanelSwingFirstTest | 253 
MacOSSystemMenuMultiWindowFXOnlySwingFirst | 241 
MacOSSystemMenuMultiWindowFXOnlySwingLast | 276 
MacOSSystemMenuMultiWindowTest | 276 
MacOSSystemMenuMultiWindowWithSwingFirstTest | 407 
MacOSSystemMenuSetWithSwingTestFirst | 466 
MacOSSystemMenuSingleWindowWithSwingFirstTest | 165 

It isn't a deal-breaker, but it would be nice to cut this down a bit if without sacrificing test coverage (e.g,. I think it's good to test with at least one sub-menu and menus with different number of items)

kevinrushforth:
I note that your branch is very out of date. Please merge the latest master into your branch. There is a minor merge conflict in `modules/javafx.graphics/src/main/native-glass/mac/GlassWindow+Overrides.m` that you will need to resolve. Leave the commented out `NSLog` (which is present in the latest master and conflicts with your change) as the first line of the function.

pabulaner:
I have rebased to the master and ran the tests again on another MacBook, but for me everything seems to work. Could You maybe provide me with some more information what exactly doesn't work with the tests for You?

kevinrushforth:
> I have rebased to the master

To echo what the Skara said: next time, please _merge_ master instead of rebasing to master.. Rebasing makes reviewing new change since last review difficult at best.

> for me everything seems to work. Could You maybe provide me with some more information what exactly doesn't work with the tests for You?

Sure. I'll provide the logs.

kevinrushforth:
Here are the assertion failures for the two failing tests:

```
MacOSSystemMenuMultiWindowFXOnlySwingLast > test() FAILED
    org.opentest4j.AssertionFailedError: Menu size is different: 5 != 2 ==> expected: <true> but was: <false>
        at app//org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:214)
        at app//test.com.sun.glass.ui.mac.MacOSSystemMenuTestBase.compareMenus(MacOSSystemMenuTestBase.java:350)
        at app//test.com.sun.glass.ui.mac.MacOSSystemMenuMultiWindowFXOnlySwingLast.test(MacOSSystemMenuMultiWindowFXOnlySwingLast.java:70)
```

```
MacOSSystemMenuMultiWindowTest > test() FAILED
    org.opentest4j.AssertionFailedError: Menu size is different: 5 != 2 ==> expected: <true> but was: <false>
        at app//org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:214)
        at app//test.com.sun.glass.ui.mac.MacOSSystemMenuTestBase.compareMenus(MacOSSystemMenuTestBase.java:350)
        at app//test.com.sun.glass.ui.mac.MacOSSystemMenuMultiWindowTest.test(MacOSSystemMenuMultiWindowTest.java:69)
```

I instrumented the code and it is picking up the menus of the Terminal app from which I ran `gradle test`. If I click on another app (e.g., the Finder) after starting `gradle` but before the test actually starts then the test passes.

I fired off a headful test on our Jenkins system and none of the tests pass. They all timeout or fail with an error from `osascript`.

All of this leads me to think that it might be better to convert these tests to manual test. Instead of using `osascript`, you could have each window show a `TextArea` with a list of what the top level menu items should be when you click in the window. What are your thoughts on this?

modules/javafx.graphics/src/main/native-glass/mac/GlassWindow+Overrides.m
```diff
@@ -69,6 +69,10 @@ - (void)windowDidChangeScreen:(NSNotification *)notification
 
 - (void)windowDidBecomeKey:(NSNotification *)notification
 {
+    // store host menu if running embedded, otherwise we
+    // just store a default menu
+    self->hostMenu = [NSApp mainMenu];
+
     //NSLog(@"windowDidBecomeKey: %p", self);
```
kevinrushforth: Minor merge issue: This should be the first line of this method.

modules/javafx.graphics/src/main/native-glass/mac/GlassWindow+Overrides.m
```diff
@@ -92,6 +96,19 @@ - (void)windowDidBecomeKey:(NSNotification *)notification
...
+    }
+
+    // restore menu of host application if running embedded,
+    // otherwise we just restore a default menu
+    if ([NSApp mainMenu] == menu) {
+        [NSApp setMainMenu:self->hostMenu];
+        [[NSApp mainMenu] update];
+    }
+
     //NSLog(@"windowDidResignKey: %p", self);
```
kevinrushforth: Minor merge issue: This should be the first line of this method.

kevinrushforth (commented):
I left a couple comments about a (minor) merge issue.

pabulaner:
As long as manual tests are considered a good practice in such cases, I would be fine with it.
This would of course mean that the test cases won't automatically detect if something doesn't work anymore.
Otherwise I see no problem with manual tests.

pabulaner:
So I have fixed Your minor merge issues and have converted the tests to manual tests. Maybe You can take a look and give me some feedback on what might need to be changed, since I don't know exactly what requirements manual tests need to satisfy.

PS: Sorry for the rebasing, I am used to this workflow from my work, but I will use merge in the future.

kevinrushforth:
> As long as manual tests are considered a good practice in such cases, I would be fine with it. This would of course mean that the test cases won't automatically detect if something doesn't work anymore. Otherwise I see no problem with manual tests.

Exactly. Automated tests are preferred where feasible. This seems like a case where they aren't quite (which is too bad, since I liked the idea of your automated tests).

Manual tests seem fine in this case. I'll review and test them. Thanks.

pabulaner:
Hey, are there any news on this PR?

FlorianKirmaier:
If we can't run the Tests on the CI, i think we should at least keep them - because they are really valuable for every person working on this topic in the future.

Anything left for us to do?

kevinrushforth:
This is on my list to review this week. Sorry for the delay.

pabulaner:
So are there any updates?

kevinrushforth:
Not yet, but it's still on my list.

tests/manual/systemmenu/MacOSSystemMenuTestBase.java
```diff
@@ -442,6 +474,18 @@ private static String getMenuReaderProcess(InputStream in) throws IOException {
      *
      **************************************************/
 
+    protected void waitForUser() {
+        while (!nextStep) {
+            try {
+                Thread.sleep(250);
+            } catch (InterruptedException e) {
+                // empty
```
kevinrushforth: Even assuming we could keep this as part of our automated test suite (which we can't), ignoring an `InterruptedException` (or any other similar exception) in a test is an anti-pattern. Among other things, it will defeat the Junit timeout mechanism.

kevinrushforth (commented):
I see that you made the tests manual (good), but left them as part of our automated test suite (not good). These tests will hang our automated test jobs, so cannot stay where they are.

You will need to move them somewhere underneath `tests/manual`. I recommend creating a new directory, called SystemMenu or something similar. You can either put them in the default package or create a package underneath the SystemMenu directory as you prefer.

You will either need an ant or gradle script to compile and run them, or else remove (comment out) the uses of JUnit. The latter is probably easier but do whichever you prefer.

I ran the automated tests as is manually on my macOS 26 system, and they work as expected for me. It might be helpful to have a "Fail" button in addition to "Next", but in the interest of getting this bug fix in, that could be done as a follow up.

So I think this is close to being ready to go in.

kevinrushforth:
> These tests will hang our automated test jobs, so cannot stay where they are.

To clarify, everything under `tests/system` need to be able to run unattended on a system in a build & test farm that doesn't have a user sitting in front of the system.

pabulaner:
Thanks again for the feedback, I implemented it, but had to reset a previous commit. I moved the tests to the manual directory under systemmenu and am now throwing a runtime exception if an interrupt occurs.

kevinrushforth:
They are in the right place now, but without build script support, they can't easily be run. We have an open enhancement to wire up manual tests to the build, but it isn't implemented yet. So each test needs to do one of two things in order to be able to run it:

1. Remove all dependencies on other packages, including (especially) Junit, and provide a main method for each of the tests, so that a developer can run "javac *.java" and then "java MyTest".
2. Provide a simple script (gradle or ant) to build and run the tests

I'm not sure which is easier. I'm out of time to finish the review before the holidays, but will pick it up next year. I'd like to see this get into jfx26 if possible.

pabulaner:
Thanks for the fast reply!

I have removed all external dependencies and added a main method to each test. 
I also have added a README.md on how to compile and run the tests, since You need to add the javafx modules.

kevinrushforth:
@pabulaner  The latest version of the tests build and run fine for me. Thanks!

I'll do some additional testing and complete my review very soon. In the mean time, can you do two thing?

1. Enable GitHub Actions (GHA) Workflows in your personal fork. This will allow sanity tests to run when you push code changes to your branch of your personal fork, and allow the Skara bot to report the results of the tests in the PR. Navigate to the home page of your GitHub repo, click on the "Actions" tab, and then click on the green button that says something like: "I understand my workflows, go ahead and enable them".
2. Merge (not rebase) the latest upstream master into your branch and push it. That will trigger a GHA test run on your now-up-to-date branch.

kevinrushforth (commented):
The tests all work as expected. The only thing I notice is that they don't always exit when the last window is closed, but given the nature of the tests, that isn't surprising. Improving this could be done as a follow-up.

In parallel with your enabling a GHA run, I'll do one last CI headful test run. When both are successful, I'll approve it.

tests/manual/systemmenu/README.md
```diff
@@ -0,0 +1,13 @@
+# Compiling and Running
+
+To compile the tests, run the following command:
```
andy-goryachev-oracle: Actually, separate compile step is not necessary, the test apps run with the following command from within this directory (you do need to run the gradle build first):

```
java --module-path ../../../build/sdk/lib/ --add-modules javafx.controls,javafx.fxml,javafx.swing MacOSSystemMenuJFXPanelSwingFirstTest.java
```

tests/manual/systemmenu/MacOSSystemMenuJFXPanelSwingFirstTest.java
```diff
@@ -0,0 +1,96 @@
...
+import javax.swing.SwingUtilities;
+
+public class MacOSSystemMenuJFXPanelSwingFirstTest extends MacOSSystemMenuTestBase {
+
+    public static void main(String[] args) throws InterruptedException, IOException {
+        new MacOSSystemMenuJFXPanelSwingFirstTest().test();
+    }
+
+    public void test() throws InterruptedException, IOException {
+        initSwing(List.of(TEST_MENUS_0, TEST_MENUS_1));
```
andy-goryachev-oracle, line 61: It would be nice to have some instructions / acceptance criteria in each test.
Assume the tester knows nothing about the issue.

tests/manual/systemmenu/MacOSSystemMenuMultiWindowTest.java
```diff
@@ -0,0 +1,81 @@
...
+import javax.swing.JMenuBar;
+import javax.swing.JMenuItem;
+import javax.swing.SwingUtilities;
+
+public class MacOSSystemMenuMultiWindowTest extends MacOSSystemMenuTestBase {
+
+    public static void main(String[] args) throws InterruptedException, IOException {
+        new MacOSSystemMenuMultiWindowTest().test();
+    }
+
```
andy-goryachev-oracle, line 57: I've got this upon exiting via system menu "Quit java" once:

```
Exception in thread "InvokeLaterDispatcher" java.lang.RuntimeException: Main Java thread is detached.
	at javafx.graphics@26-internal/com.sun.glass.ui.mac.MacApplication._submitForLaterInvocation(Native Method)
	at javafx.graphics@26-internal/com.sun.glass.ui.mac.MacApplication.submitForLaterInvocation(MacApplication.java:375)
	at javafx.graphics@26-internal/com.sun.glass.ui.InvokeLaterDispatcher.run(InvokeLaterDispatcher.java:122)
```

Also, if I just launch it and then immediately select "Quit java" from the system menu, it hangs with a beach ball and requires ctrl-C in the terminal to kill it.

tests/manual/systemmenu/MacOSSystemMenuMultiWindowWithSwingFirstTest.java
```diff
@@ -0,0 +1,87 @@
...
+import javax.swing.JFrame;
+import javax.swing.JMenu;
+import javax.swing.JMenuBar;
+import javax.swing.JMenuItem;
+import javax.swing.SwingUtilities;
+
+public class MacOSSystemMenuMultiWindowWithSwingFirstTest extends MacOSSystemMenuTestBase {
+
+    public static void main(String[] args) throws InterruptedException, IOException {
+        new MacOSSystemMenuMultiWindowWithSwingFirstTest().test();
```
andy-goryachev-oracle, line 55: this test requires selecting SourceLauncher -> Quit java from the system menu twice to close - is this intentional?

tests/manual/systemmenu/MacOSSystemMenuMultiWindowWithSwingFirstTest.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public void test() throws InterruptedException, IOException {
+        initSwing(List.of(TEST_MENUS_0, TEST_MENUS_2));
+        initJavaFX(List.of(TEST_MENUS_1, TEST_MENUS_3));
+
+        focusJavaFX(0);
```
andy-goryachev-oracle, line 62: Minor: noticed that the "Next" button seem to work only once in this and other tests.  Is it intentional?

tests/manual/systemmenu/MacOSSystemMenuSingleWindowWithSwingFirstTest.java
```diff
@@ -0,0 +1,74 @@
...
+import javax.swing.JMenu;
+import javax.swing.JMenuBar;
+import javax.swing.JMenuItem;
+import javax.swing.SwingUtilities;
+
+
+public class MacOSSystemMenuSingleWindowWithSwingFirstTest extends MacOSSystemMenuTestBase {
+
+    public static void main(String[] args) throws InterruptedException, IOException {
+        new MacOSSystemMenuSingleWindowWithSwingFirstTest().test();
```
andy-goryachev-oracle, line 56: this test hangs when fx window is closed first via the close title bar icon, followed by closing the swing window with same method.

andy-goryachev-oracle (commented):
Some minor comments, one exception, and many tests hang on closing using the system menu, or the quit menu in the dock, or the close icon in the title bar.

beldenfox:
@kevinrushforth has indicated that he would like to get this into jfx26 if possible so we might want to move some of this cleanup to a follow-up PR.

At a minimum the JavaFX windows need to be made larger so the user can see the entire menu tree. And some instructions need to be added to the README.

As I understand it the user should launch the test, verify that the contents of the system menu bar match what's in the currently focused window, press the Next button, verify, press Next, verify, etc. The big issue is that the tests enter what are essentially busy loops waiting for the user to press Next. If you don't keep pressing Next all the way to the end things get very confusing. I think that needs to be re-thought but, again, we might want to push a re-write of that scope off to another PR.

A smaller issue is that the user doesn't know when to stop pressing Next. If there's an easy fix for that (like removing the Next button from the last window) you might consider doing that.

The one area not covered by these tests is what happens when *all* windows lose focus. The easiest way to make this happen is to iconify them all. Unfortunately this means there will be no windows around to provide guidance to the user. You might want to add a brief sentence to the README suggesting that after a test is done the user should iconify all the windows and verify that some system menu is still around. It's difficult to summarize what the expected results are. For strictly JavaFX windows they should see a single menu. For a mix of Swing and JavaFX windows they should see the set of menus associated with the last Swing window that had focus (assuming they can remember which one that was). Probably good enough to have the tester iconify all the windows and verify that the system doesn't explode.

tests/manual/systemmenu/README.md
beldenfox: Even easier:

`java @../../../build/run.args MacOSSystemMenuMultiWindowTest.java `

tests/manual/systemmenu/MacOSSystemMenuJFXPanelSwingFirstTest.java
pabulaner: I added it to the readme, since it is basically the same procedure for every test.

tests/manual/systemmenu/MacOSSystemMenuMultiWindowTest.java
pabulaner: The tests are not meant for exiting via the system menu. They are intended to be run, so the user can test it, and then terminate themselves at the end.

tests/manual/systemmenu/MacOSSystemMenuMultiWindowWithSwingFirstTest.java
pabulaner: For me this test works fine, with the next button doing exactly what it should.

tests/manual/systemmenu/MacOSSystemMenuSingleWindowWithSwingFirstTest.java
pabulaner: Again, the tests are intended only for validating the contents of the system menu bar and then pressing next. Everything else is not intended to work.

tests/manual/systemmenu/MacOSSystemMenuMultiWindowWithSwingFirstTest.java
pabulaner: Same as for the previous comment.

tests/manual/systemmenu/README.md
pabulaner: Thanks, very nice and clean solution!

pabulaner:
So I think I fixed all the major comments. I added a System.exit() to every test, so after pressing the next button, the windows will all close. I also edited the readme, so the user knows what is expected and how to run the tests.

pabulaner:
If I missed something important for the approval which needs to be addressed right now, please let me know.

kevinrushforth (approved):
All tests now exit for me after presing "Next" the last time. The quick headful CI test I mentioned passed and I see that the GHA run has passed.

I think any further improvements to the test can be done in a follow-up issue.

kevinrushforth:
@pabulaner I see this in your Description:

> Co-Author: @FlorianKirmaier

Consider crediting Florian with: `/contributor add @FlorianKirmaier`.

tests/manual/systemmenu/MacOSSystemMenuTestBase.java
```diff
@@ -275,6 +279,8 @@ protected void initSwingWindow(boolean fullscreen, int id, List<Element> menus)
         JMenuBar menuBar = new JMenuBar();
         JFrame window = new JFrame();
 
+        window.setSize(WINDOW_WIDTH, WINDOW_HEIGHT);
```
andy-goryachev-oracle, line 282: this line is likely unnecessary since the Scene is sized in L238, but it does not matter much.

andy-goryachev-oracle (approved):

beldenfox (approved):
I'll approve this as-is. I'd still like to see some cleanup and expansion of the tests (I had to test the "all windows iconified" cases separately) but that can be done later. I'll enter a bug for that sometime next week.

kevinrushforth:
@pabulaner This is now ready to integrate.

pabulaner:
Thanks, much appreciated! (Both for the final integration and the very helpful feedback)

crschnick:
I just noticed that the desktop methods of AWT like `Desktop.getDesktop().setPreferencesHandler()` do no longer work in newer releases of JavaFX. When you first start AWT then JavaFX, they worked fine previously. Now, no menu option is showing up anymore. I haven't fully determined yet which build exactly introduced this regression, but this PR looks to be the most obvious one. Would that be an intentional consequence of this PR?

beldenfox:
@crschnick No that would not be intentional. It seems unlikely that this PR is the cause unless your app deliberately mixes top level Swing and JavaFX windows. But I can't say for sure until I have a chance to look into this next week. Do you have standalone code that reproduces the bug?

crschnick:
You can check out https://github.com/crschnick/jfx-test/blob/master/src/main/java/io/xpipe/jfx_runner/MenuBar.java. It uses the new jfx-runner gradle scripts, so you can run and debug it from that repo by just running ./gradlew run

beldenfox:
@crschnick If I understand you correctly in the past calling `Desktop.getDesktop().setPreferencesHandler()` would ensure that a Preferences item showed up in the system application menu. I can't reproduce that behavior. I've taken your MenuBar.java sample code and run it with JDK 21 and JavaFX 17 and other combinations and can't get that to happen. No matter what I do I only see the JavaFX  application menu (4 items), never the AWT one (at least 6).

BTW I'm testing your MenuBar.java outside of the jfx-runner gradle system. I've already got my own shell scripts that allow me to switch between different JFX and Java versions. I did need to tweak MenuBar.java to make `main` public.

crschnick:
@beldenfox Do you mean that you don't see any menu bar entry in general, or only for this specific case where the order of execution is different?

I know that the menu bar entries were correctly shown in the past in earlier JDK, JavaFX, and macOS versions. I don't have an older macOS system than 26, so I can't test it going back.

I also noticed that the Desktop setAboutHandler() method seems to be broken in general for AWT for macOS 26, so maybe this is more related to the macOS version as that method worked for earlier versions as well.

beldenfox:
@crschnick I never see the Settings menu item regardless of order of execution. This is true even in older versions of JavaFX, long before this PR came along.

To see the Settings menu item in a JavaFX app you need to initialize it so AWT creates the application menu instead of JavaFX. I don't see how you can do that with test code you provided even in earlier versions of JavaFX. I would like to look into this but need a stand-alone program that reproduces the problem e.g. that shows the Settings menu in an older version of JavaFX but not in the current version.

crschnick:
@beldenfox What macOS version are you on? I see the settings entry in the reproducer I posted on macOS 26 with JavaFX master and JDK 25. The call to `GraphicsEnvironment.getLocalGraphicsEnvironment()` will initialize AWT in the reproducer

crschnick:
Or more specifically, I see the settings entry in the reproducer if I uncomment the respective line where the comment says. The default state of the reproducer is to show that the settings entry is not shown, but can be when changing the order by uncommenting

beldenfox:
I think I understand what's going on now (it helps that I'm no longer on vacation and no longer on jury duty). Again, this PR concerns what happens when you have a top-level JavaFX window focused inside an AWT application.

In the past the only application menu you would see would be provided by AWT. The AWT application menu always has an About item and optionally has a Settings item if the client has called `Desktop.setPreferencesHandler`. With this PR when a JavaFX window is focused you'll see the JavaFX version of the application menu, not the AWT one. That might be considered a regression though it's hard to say since I think this whole setup is officially unsupported (?)

@crschnick The MenuBar.java test you provided illustrates something entirely new. When a top-level JavaFX window is focused in an AWT application and you call `Desktop.setPreferencesHandler` AWT will manipulate the JavaFX platform menu and insert the Settings item into it. The result is a weird menu with Settings in the wrong location and a mix of NSMenuItems created by Glass and AWT. Again, technically a bug but maybe the resolution is just to tell developers to not do that.

The MenuBar.java test only reproduces the issue when run inside the jfx-test harness. When run as a standalone program the issue doesn't reproduce (Settings never appears in the application menu). This is true even after I tweaked `main` to ensure it ran. In the future please provide a stand-alone reproducer.

crschnick:
@beldenfox Thanks for the explaination.

Would it be possible for JavaFX to not touch anything for the menu bar when no system menu bar is set from the JavaFX side? That would make these problems go away in the smoothest fashion.

The test harness shouldn't do anything differently than just launching the test program with the latest jfx changes, so I am not sure why that works differently. So it should behave identical to a standalone reproducer. I will try to figure that one out.

beldenfox:
> Would it be possible for JavaFX to not touch anything for the menu bar when no system menu bar is set from the JavaFX side? That would make these problems go away in the smoothest fashion.

Yes, this should be possible. I'm working on it now.

PR #1913, opened 2025-09-19
8367991: Update RegionBackground tests to use ScreenCaptureTestWatcher utility

jayathirthrao:
We are seeing very rare intermittent failures in RegionBackgroundFillUITest & RegionBackgroundImageUITest tests in linux. We already have [JDK-8328217](https://bugs.openjdk.org/browse/JDK-8328217) to fix this issue.

To identify the root cause better, we want to capture the test content when it is failing using ScreenCaptureTestWatcher utility.

Since we wan to capture the test window before test stage is cleared. We are now clearing the test stage and next test setup. Initially tried updating this behaviour in `VisualTestBase` itself, but that causes tests like `test.robot.test3d.PointLightIlluminationTest` as it overrides the test setup functions.

This change is tested to make sure ScreenCaptureTestWatcher takes appropriate screen capture on failure and CI full headful test run is green.

### DIFF
--- a/tests/system/src/test/java/test/robot/javafx/scene/layout/RegionBackgroundFillUITest.java
+++ b/tests/system/src/test/java/test/robot/javafx/scene/layout/RegionBackgroundFillUITest.java
@@ -74,18 +74,7 @@ public class RegionBackgroundFillUITest extends RegionUITestBase {
         });
         runAndWait(() -> robot = new Robot());
         Util.parkCursor(robot);
-        screenCapture = null;
-        runAndWait(() -> {
-            stage = getStage();
-            region = new Region();
-            region.setPrefSize(REGION_WIDTH, REGION_HEIGHT);
-            region.relocate(REGION_LEFT, REGION_TOP);
-            scene = new Scene(root = new Group(region), WIDTH, HEIGHT);
-            scene.setFill(SCENE_FILL);
-            stage.setScene(scene);
-            stage.setAlwaysOnTop(true);
-            stage.show();
-        });
+        initialize();
     }
 
     /**
--- a/tests/system/src/test/java/test/robot/javafx/scene/layout/RegionBackgroundFillUITest.java
+++ b/tests/system/src/test/java/test/robot/javafx/scene/layout/RegionBackgroundFillUITest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -27,14 +27,25 @@ package test.robot.javafx.scene.layout;
 import static org.junit.jupiter.api.Assumptions.assumeTrue;
 import java.util.concurrent.TimeUnit;
 import javafx.scene.layout.BackgroundFill;
+import javafx.scene.Scene;
+import javafx.scene.Group;
+import javafx.scene.layout.Region;
 import javafx.scene.paint.Color;
+import javafx.scene.robot.Robot;
+import javafx.stage.Stage;
+import test.util.Util;
+import test.util.ScreenCaptureTestWatcher;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.Timeout;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.AfterEach;
 
 /**
  */
 @Timeout(value=20000, unit=TimeUnit.MILLISECONDS)
+@ExtendWith(ScreenCaptureTestWatcher.class)
 public class RegionBackgroundFillUITest extends RegionUITestBase {
 
     /**************************************************************************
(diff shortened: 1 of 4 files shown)


kevinrushforth (commented):
This approach requires each test to copy logic from the base class into the test class, so it will be harder to maintain and doesn't scale well.

My recommendation is to either:

1. Make the changes to move the hiding of the changes to the base class methods, change all subclasses that currently have a different `@BeforeEach` method _not_named `doSetup` to rename their `@BeforeEach` method to `doSetup()` and call `super.doSetup()` as the first statement of the overridden method
2. Create a protected utility method in VisualTestBase that a subclass can call.

I think 1 might be better, since it keeps all of the logic in the base class, but 2 would be OK as well, since the actual cleaning of the stages would remain in the base class, but it does require overriding the two methods.

tests/system/src/test/java/test/robot/javafx/scene/layout/RegionBackgroundImageUITest.java
```diff
@@ -39,8 +49,37 @@
  *                                                                        *
  *************************************************************************/
 @Timeout(value=20000, unit=TimeUnit.MILLISECONDS)
+@ExtendWith(ScreenCaptureTestWatcher.class)
 public class RegionBackgroundImageUITest extends RegionUITestBase {
 
+    @BeforeEach
+    public void doSetup() {
+        runAndWait(() -> {
```
andy-goryachev-oracle: suggestion:

we've got `Util::shutdown()` which `hide()`s all the windows then calls `Platform::exit`.

we could add `Util::hideAllWindows(Runnable)` that can be used in these tests, and remove stages from the `VisualTestBase class`:

```
    /**
     * This synchronous method first hides all the open {@code Window}s in the platform thread,
     * then invokes {@link Platform.exit()}.
     */
    public static void shutdown() {
        hideAllWindows(Platform::exit);
    }

    /// Hides all open `Window`s then invokes `after` (if non-null) in the FX Application Thread.
    public static void hideAllWindows(Runnable after) {
        runAndWait(() -> {
            List.
                copyOf(Window.getWindows()).
                forEach(Window::hide);
            if (after != null) {
                after.run();
            }
        });
    }
```

What do you think?

tests/system/src/test/java/test/robot/javafx/scene/layout/RegionBackgroundImageUITest.java
```diff
@@ -39,8 +49,37 @@
...
+
+    /**
+     * This method is overridden and doing nothing, because we want the stage
+     * to be active until screen capture is done by ScreenCaptureTestWatcher
+     * utility on test failure. Stale test stage will be cleared before next
+     * test stage is created during setup.
+     */
+    @AfterEach
+    public void doTeardown() {
+
```
andy-goryachev-oracle: please remove empty line

tests/system/src/test/java/test/robot/javafx/scene/layout/RegionBackgroundFillUITest.java
```diff
@@ -49,6 +60,34 @@ public class RegionBackgroundFillUITest extends RegionUITestBase {
...
+
+    /**
+     * This method is overridden and doing nothing, because we want the stage
+     * to be active until screen capture is done by ScreenCaptureTestWatcher
+     * utility on test failure. Stale test stage will be cleared before next
+     * test stage is created during setup.
+     */
+    @AfterEach
+    public void doTeardown() {
+
```
andy-goryachev-oracle: please remove empty line

tests/system/src/test/java/test/robot/javafx/scene/layout/RegionBackgroundImageUITest.java
kevinrushforth: Seems like a good idea, but I'd recommend doing it as a separate follow-on bug.

Btw, I would add a `Window... windows` parameter and then this would also be useful for test that only want to close a subset of their windows (some such tests have the main stage remain showing throughout the test and don't want that window closed).

andy-goryachev-oracle: last time I checked, no system test needs to specify which windows to hide.

kevinrushforth: Maybe not at `@AfterAll` time, but there are tests that close the active window in `@AfterEach` leaving the primary stage shown.

kevinrushforth: Anyway, this is another reason why this should be a follow-up issue: so we can discuss the requirements of a more general utility.

jayathirthrao:
> This approach requires each test to copy logic from the base class into the test class, so it will be harder to maintain and doesn't scale well.
> 
> My recommendation is to either:
> 
>     1. Make the changes to move the hiding of the changes to the base class methods, change all subclasses that currently have a different `@BeforeEach` method _not_named `doSetup` to rename their `@BeforeEach` method to `doSetup()` and call `super.doSetup()` as the first statement of the overridden method
> 
>     2. Create a protected utility method in VisualTestBase that a subclass can call.
> 
> 
> I think 1 might be better, since it keeps all of the logic in the base class, but 2 would be OK as well, since the actual cleaning of the stages would remain in the base class, but it does require overriding the two methods.

I have followed the 1st recommendation and updated to include main change in base class and now all sub-classes override the base class setup functions with same name. I have checked all classes which inherit from VisualTestBase and updated them.

Also PointLightIlluminationTest maintains first created stage and uses it for all consecutive tests, so added flag in base class to make sure we re-use the stage. CI headful test run with latest update is green.

tests/system/src/test/java/test/robot/javafx/scene/layout/RegionBackgroundImageUITest.java
jayathirthrao: I will leave this update to `test.util.Util` for follow up issue.

tests/system/src/test/java/test/robot/javafx/scene/layout/RegionBackgroundImageUITest.java
jayathirthrao: The function itself is removed.

tests/system/src/test/java/test/robot/javafx/scene/layout/RegionBackgroundFillUITest.java
jayathirthrao: The function itself is removed.

tests/system/src/test/java/test/robot/test3d/PointLightIlluminationTest.java
```diff
@@ -81,6 +82,8 @@ public void setupEach() {
             });
             // Ensure that the scene is really displayed, before the tests begin
             waitFirstFrame();
+            // The same scene can be used for all the illumination tests by this class
```
kevinrushforth: Minor: "scene" --> "stage" ?

tests/system/src/test/java/test/robot/testharness/VisualTestBase.java
```diff
@@ -55,11 +55,13 @@ public abstract class VisualTestBase {
     // Singleton Application instance
     private static MyApp myApp;
     // Scene instances used for testing
-    private List<Stage> stages = new ArrayList<>();
+    private static List<Stage> stages = new ArrayList<>();
 
     // Glass Robot instance
     Robot robot;
 
+    protected static boolean clearStage = true;
```
kevinrushforth: Minor: would `clearStages` (plural) be a better name?

kevinrushforth (approved):
LGTM. I left a couple minor comments inline and will reapprove if you make any changes.

tests/system/src/test/java/test/robot/javafx/scene/layout/RegionBackgroundImageUITest.java
jayathirthrao: Follow-up issue : [JDK-8368363](https://bugs.openjdk.org/browse/JDK-8368363)

PR #1915, opened 2025-09-20
8368166: Media query should accept multiple rules

mstr2:
Given a media query with more than one rule:

```css
@media (prefers-color-scheme: dark) {
    .foo1 {
        -fx-background-color: black;
    }
    .foo2 {
        -fx-background-color: white;
    }
}
```

The following CSS parser error is encountered: `Expected RBRACE`

The reason for this bug is that the CSS parser mistakenly expects that after the first rule was parsed, the media query should be terminated with a closing curly brace. This is obviously incorrect, the fix is relatively simple.

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/css/CssParser.java
+++ b/modules/javafx.graphics/src/main/java/javafx/css/CssParser.java
@@ -4331,15 +4331,8 @@ final public class CssParser {
 
             currentToken = nextToken(lexer);
 
-            while (expectedRBraces > 0) {
-                if (!consumeRBrace(lexer)) {
-                    return;
-                }
-
-                if (mediaRule != null) {
-                    mediaRule = mediaRule.getParent();
-                }
-
+            while (expectedRBraces > 0 && currentToken != null && currentToken.getType() == CssLexer.RBRACE) {
+                mediaRule = mediaRule.getParent();
                 currentToken = nextToken(lexer);
                 expectedRBraces--;
             }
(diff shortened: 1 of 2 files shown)


andy-goryachev-oracle:
The change looks good, verified with the monkey tester on macOS 15.6.1 M1.

Question:
When I supply a stylesheet with two missing close braces, the master branch complains about `Expected RBRACE at [3,0]` to stderr, the code in this PR does not.

Here is my CSS:
```
@media (prefers-reduced-motion: reduce) {
  .root { -fx-accent: yellow; }
  .foo { bar: bax; }
    @media (prefers-color-scheme: light) {
      .foo { bar: baxss; }
      .root { -fx-accent: red; }
    }
    @media (prefers-color-scheme: dark) {
      .foo { bar: baxss; }
      .root { -fx-accent: green; }
```

(the css is updated via `Scene::getStylesheets`)

mstr2:
> Question: When I supply a stylesheet with two missing close braces, the master branch complains about `Expected RBRACE at [3,0]` to stderr, the code in this PR does not.

I've added some code that emits a parser error when a closing curly brace is missing.

andy-goryachev-oracle:
Thank you for updating the error message!

What are the numbers in square brackets?  Is it sufficient for the user to determine where the problem is?

For example, when I enter a single '{' in the monkey tester CSS Playground -> Custom CSS field, I get the following output:

`WARNING: CSS Error parsing '{: Expected RBRACE at [-1,-1]`

what does `[-1,-1]` refer to?

mstr2:
> Than you for updating the error message!
> 
> What are the numbers in square brackets? Is it sufficient for the user to determine where the problem is?
> 
> For example, when I enter a single '{' in the monkey tester CSS Playground -> Custom CSS field, I get the following output:
> 
> `WARNING: CSS Error parsing '{: Expected RBRACE at [-1,-1]`
> 
> what does `[-1,-1]` refer to?

It's the line and column number. [-1,-1] means that the CSS parser doesn't know where the error is. One of the many deficiencies of the CSS implementation.

andy-goryachev-oracle:
Is this just a matter of initializing the values to [0,0] perhaps?

Regardless, I think the current changes are acceptable.  Thanks for clarifications!

mstr2:
> Is this just a matter of initializing the values to [0,0] perhaps?
> 
> Regardless, I think the current changes are acceptable. Thanks for clarifications!

This happens when `currentToken` is null, which probably (only?) happens at the end of the token stream. So the last column of the last line would be a better choice than [0,0].

andy-goryachev-oracle:
> So the last column of the last line would be a better choice than [0,0].

Right, the general idea is to show the correct information to the user.  It can either be addressed this as a part of this PR or a follow up.

mstr2:
> > So the last column of the last line would be a better choice than [0,0].
> 
> Right, the general idea is to show the correct information to the user. It can either be addressed this as a part of this PR or a follow up.

This would have to be another PR, because it's preexisting and unrelated to this bug.

kevinrushforth:

arapte (approved):
LGTM. Verified that with this change, nested @media queries with multiple rules can be provided. Also, the older way of providing the @media query separately for each rule also works fine.

PR #1916, opened 2025-09-22
8368219: Skip MenuDoubleShortcutTest on macOS & Linux

jayathirthrao:
MenuDoubleShortcutTest > nonMacMenuBarComesBeforeScene()/macSceneComesBeforeMenuBar() was failing intermittently and it was made more robust under [JDK-8364405](https://bugs.openjdk.org/browse/JDK-8364405) and it was verified with multiple runs and it was passing.

But looks like we still see this test failing in our CI. We need to skip this test and test fix will be worked upon under [JDK-8368074](https://bugs.openjdk.org/browse/JDK-8368074)


kevinrushforth (approved):
LGTM

kevinrushforth:
Since this is a simple addition to skip failing tests, no need to wait 24 hours.

PR #1917, opened 2025-09-23
8368375: Skip intermittently failing SWTCursorsTest

jayathirthrao:
This test is failing intermittently on macOS and Linux platforms and it will be fixed under [JDK-8367599](https://bugs.openjdk.org/browse/JDK-8367599).

We need to skip this until [JDK-8367599](https://bugs.openjdk.org/browse/JDK-8367599) is fixed.

javafx.base/com.sun.javafx.PlatformUtil helps to determine individual platforms on which test is running. But this module is exposed only to system tests and not to SWT tests. So i am replicating this test utility code in SWTCursorsTest.

---------
### Progress

### Issue
 * [JDK-8368375](https://bugs.openjdk.org/browse/JDK-8368375): Skip intermittently failing SWTCursorsTest (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1917/head:pull/1917` \
`$ git checkout pull/1917`

`$ git checkout pull/1917` \
`$ git pull https://git.openjdk.org/jfx.git pull/1917/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1917`

`$ git pr show -t 1917`

</details>

<a href="https://git.openjdk.org/jfx/pull/1917.diff">https://git.openjdk.org/jfx/pull/1917.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1917#issuecomment-3322853386)
</details>

### DIFF
--- a/modules/javafx.swt/src/test/java/test/javafx/embed/swt/SWTCursorsTest.java
+++ b/modules/javafx.swt/src/test/java/test/javafx/embed/swt/SWTCursorsTest.java
@@ -35,7 +35,7 @@ import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Shell;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.Timeout;
-import org.junit.jupiter.api.Assumptions;
+import org.junit.jupiter.api.Disabled;
 
 import java.util.concurrent.TimeUnit;
 
@@ -44,18 +44,11 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 public class SWTCursorsTest extends SWTTest {
 
     private FXCanvas canvas;
-    private static final String os = System.getProperty("os.name");
-    private static String javafxPlatform =
-        System.getProperty("javafx.platform");
-    private static final boolean ANDROID = "android".equals(javafxPlatform) ||
-        "Dalvik".equals(System.getProperty("java.vm.name"));
 
+    @Disabled("JDK-8367599")
     @Test
     @Timeout(value = 10000, unit = TimeUnit.MILLISECONDS)
     public void testImageCursor() throws Throwable {
-        // Skip this test on Linux & macOS because of JDK-8367599
-        Assumptions.assumeFalse(os.startsWith("Mac") ||
-            (os.startsWith("Linux") && !ANDROID));
         runOnSwtThread(() -> {
             Display display = Display.getCurrent();
             final Shell shell = new Shell(display);


modules/javafx.swt/src/test/java/test/javafx/embed/swt/SWTCursorsTest.java
```diff
@@ -43,10 +44,18 @@
 public class SWTCursorsTest extends SWTTest {
 
     private FXCanvas canvas;
+    private static final String os = System.getProperty("os.name");
+    private static String javafxPlatform =
+        System.getProperty("javafx.platform");
+    private static final boolean ANDROID = "android".equals(javafxPlatform) ||
+        "Dalvik".equals(System.getProperty("java.vm.name"));
```
kevinrushforth: Is there a reason not to use `PlatformUtil` to do the OS checks?

jayathirthrao:
> Have you run it on Windows enough to know that it is stable there? If so, then this seems fine, although I'm curious as to why you duplicated the platform check logic rather than just calling the PlatformUtil methods.

I was under the impression that after these tests were enabled in [JDK-8169285](https://bugs.openjdk.org/browse/JDK-8169285) we are running them on windows machine in our CI, which is wrong. So this test is not reliable run on windows machine at all. So its better if we just disable this test completely.

Regarding not using PlatformUtil methods : If i import PlatformUtil class in these tests i get `package com.sun.javafx is declared in module javafx.base, which does not export it to the unnamed module`. I tried adding `--add-exports=javafx.base/com.sun.javafx.PlatformUtil=ALL-UNNAMED` in compiler args of swt job, that didn't work out. Also i was not sure whether we need to export this javafx.base class or not. So i went ahead and added utility code in test.

Since we will be disabling this test completely, i will go ahead and remove all PlatformUtil related code.

kevinrushforth:
Since this is just skipping a failing test, no need to wait 24 hours.

jayathirthrao:
> Looks good.
> 
> Since we are disabling the test on all platforms, this PR's description should probably be updated to remove mentioning of `PlatformUtils`.

I think preserving the description as is maintains the order of commits and why particular decisions were taken during review.

kevinrushforth:
> Regarding not using PlatformUtil methods : If i import PlatformUtil class in these tests i get `package com.sun.javafx is declared in module javafx.base, which does not export it to the unnamed module`. I tried adding `--add-exports=javafx.base/com.sun.javafx.PlatformUtil=ALL-UNNAMED` in compiler args of swt job, that didn't work out. Also i was not sure whether we need to export this javafx.base class or not. So i went ahead and added utility code in test.
> 
> Since we will be disabling this test completely, i will go ahead and remove all PlatformUtil related code.

As you say, this is no longer needed for this PR. If something like this arises in the future, the way to add the needed export is to create a `modules/javafx.swt/src/test/addExports` file (the system tests and most of the modules already have one, but `javafx.swt` does not) and the add the following line:

```
--add-exports javafx.base/com.sun.javafx=ALL-UNNAMED
```

(note that exports are done at the package level not per class)

PR #1921, opened 2025-09-27
8368021: Window buttons of extended RTL stage are on the wrong side

mstr2:
When an extended stage is shown with RTL orientation (either using a RTL window + `Scene.nodeOrientation == INHERIT`, or using a LTR window and `Scene.nodeOrientation == RIGHT_TO_LEFT`), the default window buttons are placed on the wrong side of the window. This bug only manifests on Windows and Linux, both of which use `HeaderButtonOverlay` to render the default window buttons.

`HeaderButtonOverlay` is not a part of the scene graph, it is shown on top of the scene graph as an overlay (like the warning overlay that appears when entering full-screen mode). For [CSS-related reasons](https://github.com/openjdk/jfx/pull/1605#issuecomment-2967977276), the parent of an overlay is the scene root (but the scene root doesn't know that). This implementation detail can mess up the calculation of orientation flags and mirroring transforms in `Node`, as depending on the `NodeOrientation` of the root node, the code may mistakenly mirror (or not mirror) the orientation.

The solution I've come up with is as follows: the overlay node is marked with the `Node.INHERIT_ORIENTATION_FROM_SCENE` flag, which causes it to resolve its effective orientation against the scene only, and never against the root node. With this change, the effective orientation and mirroring transforms are computed correctly.

The easiest way to test this fix is with Monkey Tester -> Tools -> Stage Tester.

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/ViewSceneOverlay.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/ViewSceneOverlay.java
@@ -27,6 +27,7 @@ package com.sun.javafx.tk.quantum;
 
 import com.sun.javafx.scene.NodeHelper;
 import com.sun.javafx.scene.SceneHelper;
+import javafx.geometry.NodeOrientation;
 import javafx.scene.Node;
 import javafx.scene.Parent;
 import javafx.scene.Scene;
@@ -49,7 +50,9 @@ final class ViewSceneOverlay {
     ViewSceneOverlay(Scene scene, ViewPainter painter) {
         this.scene = scene;
         this.painter = painter;
-        this.subscription = scene.rootProperty().subscribe(this::onSceneRootChanged);
+        this.subscription = Subscription.combine(
+            scene.rootProperty().subscribe(this::onSceneRootChanged),
+            scene.effectiveNodeOrientationProperty().subscribe(this::onEffectiveNodeOrientationInvalidated));
     }
 
     public void dispose() {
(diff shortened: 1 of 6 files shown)


kevinrushforth:
This looks like a reasonable fix to me.

Reviewers: @andy-goryachev-oracle @Ziad-Mid 

/reviewers 2

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/ViewSceneOverlay.java
```diff
@@ -49,7 +50,9 @@ final class ViewSceneOverlay {
     ViewSceneOverlay(Scene scene, ViewPainter painter) {
         this.scene = scene;
         this.painter = painter;
-        this.subscription = scene.rootProperty().subscribe(this::onSceneRootChanged);
+        this.subscription = Subscription.combine(
```
andy-goryachev-oracle: should this var be called `subscriptions` then?

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/ViewSceneOverlay.java
```diff
@@ -49,7 +50,9 @@ final class ViewSceneOverlay {
...
+            scene.rootProperty().subscribe(this::onSceneRootChanged),
+            scene.effectiveNodeOrientationProperty().subscribe(this::onEffectiveNodeOrientationInvalidated));
```
andy-goryachev-oracle, line 55: side note to @hjohn : 
would it make sense to add a `Subscription.subscribe(Runnable)` method for cases like this?

Ziad-Mid: I do not have a Linux machine to test it. I am only able to test on mac / win which seems to work fine

kevinrushforth:
I can do a quick test of this on Linux (Ubuntu 24.04).

Ziad-Mid (approved):
LGTM

PR #1080, opened 2023-04-02
8305418: [Linux] Replace obsolete XIM as Input Method Editor

tsayao:
This replaces obsolete XIM and uses gtk api for IME.
Gtk uses [ibus](https://github.com/ibus/ibus)

Gtk3+ uses relative positioning (as Wayland does), so I've added a Relative positioning on `InputMethodRequest`.

[Screencast from 17-09-2023 21:59:04.webm](https://github.com/openjdk/jfx/assets/30704286/6c398e39-55a3-4420-86a2-beff07b549d3)

---------
### Progress
- [x] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue

### Issue
 * [JDK-8305418](https://bugs.openjdk.org/browse/JDK-8305418): [Linux] Replace obsolete XIM as Input Method Editor (**Enhancement** - P3)

### Reviewers
 * [Kevin Rushforth](https://openjdk.org/census#kcr) (@kevinrushforth - **Reviewer**)
 * [Martin Fox](https://openjdk.org/census#mfox) (@beldenfox - Committer) Review applies to [7d8e5bcc](https://git.openjdk.org/jfx/pull/1080/files/7d8e5bcc9ffa520962b936d11a7b0575de88c843)
 * [Lukasz Kostyra](https://openjdk.org/census#lkostyra) (@lukostyra - **Reviewer**) Review applies to [0c0c23bd](https://git.openjdk.org/jfx/pull/1080/files/0c0c23bd1127e5fcd9c55f97d508ab8f2d6788bd)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1080/head:pull/1080` \
`$ git checkout pull/1080`

Update a local copy of the PR: \
`$ git checkout pull/1080` \
`$ git pull https://git.openjdk.org/jfx.git pull/1080/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 1080`

View PR using the GUI difftool: \
`$ git pr show -t 1080`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jfx/pull/1080.diff">https://git.openjdk.org/jfx/pull/1080.diff</a>

</details>
[...]

PR #1922, opened 2025-09-28
8367045: [Linux] Dead keys not working

tsayao:
The bug report is on Linux Mint which does not use ibus (the gnome default).

With this change, it continues to work on ibus, but also works works without it (on mint) and with fcitx.

It seems it's correct - but it's weird that it needed those bits before. I would say it's related to  `IBUS_ENABLE_SYNC_MODE` (see [this comment](https://github.com/openjdk/jfx/pull/1080#issuecomment-2412202297) on #1080.), but it seems to be still set to 0 on Ubuntu 24.04.

@Glavo: Can you check if Chinese input is still correct?
@andzsinszan: Can you check for Japanese?

---------
### Progress

### Issue
 * [JDK-8367045](https://bugs.openjdk.org/browse/JDK-8367045): [Linux] Dead keys not working (**Bug** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1922/head:pull/1922` \
`$ git checkout pull/1922`

`$ git checkout pull/1922` \
`$ git pull https://git.openjdk.org/jfx.git pull/1922/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1922`

`$ git pr show -t 1922`

</details>

<a href="https://git.openjdk.org/jfx/pull/1922.diff">https://git.openjdk.org/jfx/pull/1922.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1922#issuecomment-3343484773)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/gtk/GlassApplication.cpp
+++ b/modules/javafx.graphics/src/main/native-glass/gtk/GlassApplication.cpp
@@ -196,6 +196,10 @@ JNIEXPORT void JNICALL Java_com_sun_glass_ui_gtk_GtkApplication__1init
     gdk_window_set_events(root, static_cast<GdkEventMask>(gdk_window_get_events(root) | GDK_PROPERTY_CHANGE_MASK));
 
     platformSupport = new PlatformSupport(env, obj);
+
+    // Set ibus to sync mode
+    // This seems to have been patched on Ubuntu, but left here for compatibility
+    setenv("IBUS_ENABLE_SYNC_MODE", "1", 1);
 }
 
 /*
--- a/modules/javafx.graphics/src/main/native-glass/gtk/GlassApplication.cpp
+++ b/modules/javafx.graphics/src/main/native-glass/gtk/GlassApplication.cpp
@@ -198,7 +198,6 @@ JNIEXPORT void JNICALL Java_com_sun_glass_ui_gtk_GtkApplication__1init
     platformSupport = new PlatformSupport(env, obj);
 
     // Set ibus to sync mode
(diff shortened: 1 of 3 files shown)


beldenfox:
With this PR all text input is delivered as committed text in an InputMethodEvent. JavaFX never receives  a KEY_TYPED KeyEvent. This simplifies things but will cause all manner of follow-on problems. We need a better solution.

I cannot easily run Linux Mint since it doesn't provide an ARM version. This bug does not reproduce for me in a stock Ubuntu 24.04.3 environment which defaults to IBus. Does Linux Mint default to a different input method framework such as fcitx?

tsayao:
> With this PR all text input is delivered as committed text in an InputMethodEvent. JavaFX never receives a KEY_TYPED KeyEvent. This simplifies things but will cause all manner of follow-on problems. We need a better solution.
> 
> I cannot easily run Linux Mint since it doesn't provide an ARM version. This bug does not reproduce for me in a stock Ubuntu 24.04.3 environment which defaults to IBus. Does Linux Mint default to a different input method framework such as fcitx?

Thanks for refreshing my memory - I had a feeling that I was missing something.

The equivalent on Ubuntu would be to launch `im-config`, select `xim` and restart the session with Xorg.

tsayao:
Typed ' followed by e on a Brazilian Portuguese layout to output é and then typed "ni hao"  with Chinese Pinyin.

master branch:
image

with this PR:
image

beldenfox:
I've updated the original bug report with more details. I'll submit a PR, the solution is similar to work I did on the Mac side.

tsayao:
closing in favor of #1924

PR #1926, opened 2025-10-03
DRAFT v2 - 8252373: [macOS] Stage with owner disappears when moved to another screen

kevinrushforth:
This is a preliminary draft PR to fix [JDK-8252373](https://bugs.openjdk.org/browse/JDK-8252373) where an owned stage will disappear on macOS if it is moved to another screen with the Mission Control setting of "Displays have separate spaces" enabled (which it is by default).

This implements a proof of concept where we manage owned windows ourself in glass rather than relying on NSWindow's childWindow feature which we currently do. This matches what AWT does.

### Status

* Window stacking is implemented and works correctly based on my testing -- also tested with alwaysOnTop and full screen
* Window closing still works as before where closing an owner window will close all owned (child) windows
* NOT YET IMPLEMENTED: iconifying an owner window should iconify all owned (child) windows

Note: this PR currently has some logging information, which will be commented out or removed when I put out the real PR.

---------
### Progress

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1926/head:pull/1926` \
`$ git checkout pull/1926`

`$ git checkout pull/1926` \
`$ git pull https://git.openjdk.org/jfx.git pull/1926/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1926`

`$ git pr show -t 1926`

</details>

<a href="https://git.openjdk.org/jfx/pull/1926.diff">https://git.openjdk.org/jfx/pull/1926.diff</a>

</details>

### DIFF
--- a/apps/toys/Hello/src/main/java/hello/HelloModality.java
+++ b/apps/toys/Hello/src/main/java/hello/HelloModality.java
@@ -35,10 +35,13 @@ import javafx.scene.Scene;
 import javafx.scene.control.Button;
 import javafx.scene.control.CheckBox;
 import javafx.scene.control.RadioButton;
+import javafx.scene.control.ToggleButton;
 import javafx.scene.control.ToggleGroup;
+import javafx.scene.control.Tooltip;
 import javafx.stage.FileChooser;
 import javafx.stage.Modality;
 import javafx.stage.Stage;
+import javafx.stage.Window;
 import javafx.stage.StageStyle;
 
 public class HelloModality extends Application {
@@ -65,6 +68,13 @@ public class HelloModality extends Application {
         checker.setLayoutX(25);
         checker.setLayoutY(40);
         root.getChildren().add(checker);
+
+        // Setup AlwaysOnTop checker
+        final CheckBox onTopChecker = new CheckBox("AlwaysOnTop");
+        onTopChecker.setSelected(false);
+        onTopChecker.setLayoutX(25);
+        onTopChecker.setLayoutY(70);
+        root.getChildren().add(onTopChecker);
 
         // Setup Modality Selection
         final ToggleGroup modalityGroup = new ToggleGroup();
@@ -112,6 +122,7 @@ public class HelloModality extends Application {
         root.getChildren().add(suButton);
 
         Button button = new Button("Create Dialog");
+        button.setTooltip(new Tooltip("Creates a new stage"));
         button.setLayoutX(100);
         button.setLayoutY(200);
         button.setOnAction(new EventHandler<ActionEvent>() {
@@ -120,6 +131,9 @@ public class HelloModality extends Application {
 
                 boolean owned = checker.isSelected();
                 dialog.initOwner(owned ? stage : null);
+
+                boolean alwaysOnTop = onTopChecker.isSelected();
+                dialog.setAlwaysOnTop(alwaysOnTop);
 
                 Modality modality;
                 if (applicationButton.isSelected()) {
@@ -153,7 +167,8 @@ public class HelloModality extends Application {
                 Group dialogRoot = (Group) dialogScene.getRoot();
 
                 Button dialogButton = new Button("Dismiss");
-                dialogButton.setLayoutX(275);
+                dialogButton.setTooltip(new Tooltip("Hides (closes) this stage"));
+                dialogButton.setLayoutX(375);
                 dialogButton.setLayoutY(200);
                 dialogButton.setOnAction(new EventHandler<ActionEvent>() {
                     @Override public void handle(ActionEvent e) {
@@ -170,6 +185,7 @@ public class HelloModality extends Application {
         root.getChildren().add(button);
 
         Button button2 = new Button("Click Me");
+        button2.setTooltip(new Tooltip("Prints a message"));
         button2.setLayoutX(200);
         button2.setLayoutY(200);
         button2.setOnAction(new EventHandler<ActionEvent>() {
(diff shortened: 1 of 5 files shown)


kevinrushforth:
@andy-goryachev-oracle If you have some time, could you test this PR on your setup using the HelloModality program from this patch as well as MonkeyTester? Let me know if anything doesn't work as you would expect.

All of the changes are equally relevant to single screen, but multi-screen testing is important, too.

andy-goryachev-oracle:
will do

andy-goryachev-oracle:
HelloModality: do you want to rename the button to "Create Stage", and should we add a Dialog to the mix?

andy-goryachev-oracle:
If I create two non-modal child windows with the same owner, both with 'always on top' set, I am able to place either one on top by clicking.  Is this right?
Screenshot 2025-10-03 at 10 08 55

andy-goryachev-oracle:
- created an owned, window-modal (or application-modal) child
The parent window flickers momentarily (goes on top) when I click on it.

andy-goryachev-oracle:
- created three consecutive window-modal windows
double click on the bottom parent's title bar moves it to the lower left corner of the screen (instead of maximizing it I guess?)

note that double clicking on the child windows does nothing (apart from some flicker)

Screenshot 2025-10-03 at 10 34 59

kevinrushforth:
> If I create two non-modal child windows with the same owner, both with 'always on top' set, I am able to place either one on top by clicking. Is this right?

Yes, this is expected. If both (or neither) are "always on top" they are at the same stacking level. That's true today for non-owned windows (and true on Windows as well for either owned or non-owned).

> * created an owned, window-modal (or application-modal) child
>   The parent window flickers momentarily (goes on top) when I click on it.

Yes, that's because the OS brings it on top and then we restack them. It isn't always atomic -- if there were a way to do this, that would be ideal, but I don't know of a way. Btw, you can see this today with non-owned application-modal windows.

kevinrushforth:
> created three consecutive window-modal windows
> double click on the bottom parent's title bar moves it to the lower left corner of the screen (instead of maximizing it I guess?)

Yes, I see it too. I also see this behavior today (current master) with non-owned application-modal windows.

> note that double clicking on the child windows does nothing (apart from some flicker)

clicking on which child window of the three? I would expect that from the first two; only the last of the three should be active.

andy-goryachev-oracle:
> I would expect that from the first two; only the last of the three should be active.

that's right.  my point is that this might be (an existing) bug - I would expect the main window to ignore double clicks.

kevinrushforth:
> that's right. my point is that this might be (an existing) bug - I would expect the main window to ignore double clicks.

Yes, it's preexisting. So one thing that might come out of this effort, in addition to a fix for the multi-screen problem, is one or more follow-up bugs.

apps/toys/Hello/src/main/java/hello/HelloModality.java
```diff
@@ -35,10 +35,13 @@
 import javafx.scene.control.Button;
 import javafx.scene.control.CheckBox;
 import javafx.scene.control.RadioButton;
+import javafx.scene.control.ToggleButton;
```
andy-goryachev-oracle, line 40: copyright year

andy-goryachev-oracle:
Can't find any more issues with the 'displays have separate spaces' setting OFF.

Next week will continue with the 'displays have separate spaces' setting ON (since changing it requires logout).

kevinrushforth:
> Can't find any more issues with the 'displays have separate spaces' setting OFF.
> 
> Next week will continue with the 'displays have separate spaces' setting ON (since changing it requires logout).

Thanks! Next week I plan implement the iconify part of the fix as well.

andy-goryachev-oracle:
Two things happened:
- a forced upgrade to macOS 15.7.1
- set 'display have separate spaces' to ON

The behavior of HelloModality app looks good.

However, my recollection of the separate spaces option was that one can't move from one space to another, yet I can easily drag a window from one monitor to another.  The system info tool in the monkey tester shows the following display configuration which is as expected (notice the secondary is above the primary):

```
  Primary Screen:
    dpi=151
    bounds={x=0, y=0, w=1800, h=1169}
    visual bounds={x=0, y=44, w=1800, h=1076}
    output.scale.x=2
    output.scale.y=2
  Screen #2:
    dpi=108
    bounds={x=-760, y=-1440, w=2560, h=1440}
    visual bounds={x=-760, y=-1415, w=2560, h=1415}
    output.scale.x=1
    output.scale.y=1
```

While I've tested the case when the new window is repositioned to another screen via Mission Control, once scenario cannot be tested - when the new window is actually placed either to another screen or somewhere outside of any screen area.  Perhaps add a couple of buttons to HelloModality?

kevinrushforth:
> HelloModality: do you want to rename the button to "Create Stage"

Yes, I'll do that.

> and should we add a Dialog to the mix?

The dialog implementation just uses a Stage, so it wouldn't add anything as far as owned / modal windows goes.

kevinrushforth:
> Two things happened:
> 
> * a forced upgrade to macOS 15.7.1
> * set 'display have separate spaces' to ON
> 
> The behavior of HelloModality app looks good.

Thanks for testing!

> However, my recollection of the separate spaces option was that one can't move from one space to another, yet I can easily drag a window from one monitor to another. The system info tool in the monkey tester shows the following display configuration which is as expected (notice the secondary is above the primary):

You must be misremembering. It's always been possible to drag windows between screens using the default separate spaces mode.

> While I've tested the case when the new window is repositioned to another screen via Mission Control, once scenario cannot be tested - when the new window is actually placed either to another screen or somewhere outside of any screen area. Perhaps add a couple of buttons to HelloModality?

I assume you mean the case where a window is initially shown on another screen (or outside any screen)? I'll take a look at this.

Btw, I've found a couple corner cases that I'm still looking at. I also have an initial implementation of iconify / deiconify. I'll push an update tomorrow.

andy-goryachev-oracle:
> I assume you mean the case where a window is initially shown on another screen (or outside any screen)?

yes, both cases.

kevinrushforth:
> > I assume you mean the case where a window is initially shown on another screen (or outside any screen)?
> 
> yes, both cases.

I still need to test this, but I am confident that it will work.

kevinrushforth:
@andy-goryachev-oracle I have pushed a working version of iconify. I also fixes a couple corner cases and fix the case where the primary window enters full-screen with owned, non-modal windows.

I think this is functionally complete now.

andy-goryachev-oracle:
release the hounds!  (make it RFR?)
also, update the description?

I'll be testing it tomorrow.

kevinrushforth:
I still need to do a little more cleanup. I'll then close this PR and create a fresh PR from a clean branch (one that doesn't have "WIP" in the name so that it will get GHA testing).

kevinrushforth:
Closing in favor of #1932

PR #1927, opened 2025-10-03
8368972: Create implementation of menu bar accessibility component

azuev-java:
Create the accessibility peer for menu bar component also fix the problem when with the voiceover is on the <Ctrl>-F10 is pressed the name of the selected menu is not announced.

---------
### Progress

### Issue
 * [JDK-8368972](https://bugs.openjdk.org/browse/JDK-8368972): Create implementation of menu bar accessibility component (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1927/head:pull/1927` \
`$ git checkout pull/1927`

`$ git checkout pull/1927` \
`$ git pull https://git.openjdk.org/jfx.git pull/1927/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1927`

`$ git pr show -t 1927`

</details>

<a href="https://git.openjdk.org/jfx/pull/1927.diff">https://git.openjdk.org/jfx/pull/1927.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1927#issuecomment-3367191629)
</details>

### DIFF
--- a/.github/workflows/submit.yml
+++ b/.github/workflows/submit.yml
@@ -150,7 +150,7 @@ jobs:
   macos_x64_build:
     name: macOS x64
     needs: validation
-    runs-on: "macos-13"
+    runs-on: "macos-15-intel"
 
     env:
       # FIXME: read this information from a property file
@@ -212,8 +212,8 @@ jobs:
           java -version
           which ant
           ant -version
-          # We want to use Xcode 15.4, but 15.2 is the highest that the GHA macOS 13 runner supports
-          sudo xcode-select --switch /Applications/Xcode_15.2.app/Contents/Developer
+          # Production builds use Xcode 15.4, but 16.0 is the lowest that the GHA macOS 15 runner supports
+          sudo xcode-select --switch /Applications/Xcode_16.0.app/Contents/Developer
           xcodebuild -version
 
       - name: Build JavaFX artifacts
--- a/buildSrc/win.gradle
+++ b/buildSrc/win.gradle
@@ -326,9 +326,9 @@ WIN.glass.rcFlags = [
 WIN.glass.ccFlags = [ccFlags].flatten()
 WIN.glass.linker = linker
 WIN.glass.linkFlags = (IS_STATIC_BUILD ? [linkFlags] : [linkFlags, "delayimp.lib", "gdi32.lib", "urlmon.lib", "Comdlg32.lib",
-        "winmm.lib", "imm32.lib", "shell32.lib", "Uiautomationcore.lib", "dwmapi.lib",
+        "winmm.lib", "imm32.lib", "shell32.lib", "Uiautomationcore.lib", "dwmapi.lib", "version.lib",
         "/DELAYLOAD:user32.dll", "/DELAYLOAD:urlmon.dll", "/DELAYLOAD:winmm.dll", "/DELAYLOAD:shell32.dll",
-        "/DELAYLOAD:Uiautomationcore.dll", "/DELAYLOAD:dwmapi.dll"]).flatten()
+        "/DELAYLOAD:Uiautomationcore.dll", "/DELAYLOAD:dwmapi.dll", "/DELAYLOAD:version.dll"]).flatten()
 WIN.glass.lib = "glass"
 
 WIN.decora = [:]
--- a/modules/javafx.graphics/.classpath
+++ b/modules/javafx.graphics/.classpath
@@ -3,9 +3,26 @@
 	<classpathentry kind="src" path="src/main/java"/>
 	<classpathentry kind="src" path="build/gensrc/jsl-prism"/>
 	<classpathentry kind="src" path="build/gensrc/jsl-decora"/>
-	<classpathentry kind="src" path="build/hlsl/Decora"/>
-	<classpathentry kind="src" path="build/hlsl/Prism"/>
-	<classpathentry kind="src" path="build/msl/"/>
+	<classpathentry kind="src" path="build/msl">
+		<attributes>
+			<attribute name="ignore_optional_problems" value="true"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry kind="src" path="build/hlsl/Decora">
+		<attributes>
+			<attribute name="ignore_optional_problems" value="true"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry kind="src" path="build/hlsl/Prism">
+		<attributes>
+			<attribute name="ignore_optional_problems" value="true"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry kind="src" path="src/main/resources">
+		<attributes>
+			<attribute name="optional" value="true"/>
+		</attributes>
+	</classpathentry>
 	<classpathentry kind="src" output="testbin" path="src/shims/java">
 		<attributes>
 			<attribute name="test" value="true"/>
@@ -14,11 +31,6 @@
 	<classpathentry kind="src" output="testbin" path="src/test/java">
 		<attributes>
 			<attribute name="test" value="true"/>
-			<attribute name="optional" value="true"/>
-		</attributes>
-	</classpathentry>
-	<classpathentry kind="src" path="src/main/resources">
-		<attributes>
 			<attribute name="optional" value="true"/>
 		</attributes>
 	</classpathentry>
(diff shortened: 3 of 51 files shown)


andy-goryachev-oracle:
/reviewers 2

andy-goryachev-oracle:
I see no difference in VoiceOver announcements between master and this PR - is it expected?

One thing I noticed is that when the pulldown menu is shown, and I hit LEFT or RIGHT arrow keys, the VoiceOver does not announce the current menu (either master or this PR, so it might be a pre-existing issue).  It does announce the name of the current menu occasionally, but I can't quite determine when.

Screenshot 2025-10-03 at 15 52 36

modules/javafx.graphics/src/main/native-glass/mac/a11y/AccessibleBase.m
```diff
@@ -62,7 +62,7 @@ + (void) initializeRolesMap {
     [rolesMap setObject:@"JFXMenuItemAccessibility" forKey:@"MENU"];
     [rolesMap setObject:@"JFXMenuItemAccessibility" forKey:@"MENU_ITEM"];
     [rolesMap setObject:@"JFXMenuItemAccessibility" forKey:@"RADIO_MENU_ITEM"];
-    [rolesMap setObject:@"JFXMenuItemAccessibility" forKey:@"CHECK_MENU_ITEM"];
```
arapte, line 65: Is the removal of CHECK_MENU_ITEM intentional ?

azuev-java: No, that was an accident.

andy-goryachev-oracle: ... and I totally missed that!
Question: how come none of the tests failed?  Do we have test coverage in this area?

azuev-java: I do not think so - and i do not think the non-manual test would be possible and even with the manual test it will pass because removal of binding just means that old code will be used and the old code still works - until Apple pulls the plug on old API.

azuev-java:
> I see no difference in VoiceOver announcements between master and this PR - is it expected?
Yes.
 
> One thing I noticed is that when the pulldown menu is shown, and I hit LEFT or RIGHT arrow keys, the VoiceOver does not announce the current menu (either master or this PR, so it might be a pre-existing issue). It does announce the name of the current menu occasionally, but I can't quite determine when.

When the pull down menu is shown the accessibility focus is on the popup itself (unless you select a menu item in it) and we still did not find any way to tell VoiceOver that the popup has some name - we added all the possible API methods - label, title, titleElement, value, name - VO just ignores all of them so unfortunately there's nothing we can do on our side. This API is not wired for the per-window menu.

andy-goryachev-oracle (approved):
thanks for the explanations!

modules/javafx.graphics/src/main/native-glass/mac/a11y/JFXMenuItemAccessibility.m
```diff
@@ -33,6 +33,16 @@ - (NSAccessibilityRole)accessibilityRole
...
+        if ([super accessibilityChildren] != NULL && [[super accessibilityChildren] count] != 0) {
+            id object = [super accessibilityChildren][0];
+            if ([object respondsToSelector:@selector(accessibilityValue)]) {
+                return [object accessibilityValue];
+            }
+        } else {
+            return @"";
+        }
+    }
     return [super accessibilityTitle];
```
arapte: A couple recommendations for this method: 
1. Use `nil` instead of `NULL`
2. Store `[super accessibilityChildren]` locally

```
- (NSString * _Nullable)accessibilityLabel {
    NSString *accTitle = [super accessibilityTitle];
    if (accTitle != nil) {
        return accTitle;
    }

    NSArray *accChildren = [super accessibilityChildren];
    if (accChildren != nil && accChildren.count != 0) {
        id child0 = accChildren[0];
        if ([child0 respondsToSelector:@selector(accessibilityValue)]) {
            return [child0 accessibilityValue];
        }
    } else {
        return @"";
    }
    return nil;
}
```

azuev-java: Makes sense. I updated the pull request.

modules/javafx.graphics/src/main/native-glass/mac/a11y/JFXMenuItemAccessibility.m
```diff
@@ -33,7 +33,17 @@ - (NSAccessibilityRole)accessibilityRole
...
+    if (title == nil) {
+        NSArray *children = [super accessibilityChildren];
+        if (children != nil && [children count] != 0) {
+            if ([children[0] respondsToSelector:@selector(accessibilityValue)]) {
+                return [children[0] accessibilityValue];
+            }
+        }
+        return @"";
+    }
+    return title;
```
arapte, line 46: Minor query: Is `return @""` statement required ? Cant' we rely on `return title;` (which would return nil in case) ?

azuev-java: > Minor query: Is `return @""` statement required ? Cant' we rely on `return title;` (which would return nil in case) ?

Yes, otherwise on menu item with no accessibility title it will report "Menu separator". Since right now VoiceOver is not able to put focus on real menu separator item even with the VO shortcuts (i think Apple broke it in macOS 11) it can only confuse user.

andy-goryachev-oracle:
I might have found a weird regression (using the monkey tester).

To reproduce:
- open the menu bar page
- turn on the Voice Over
- click on a menu in the inner MenuBar, down arrow until the last submenu
- press SPACE to open the submenu
- the VO announces "you are on blah blah, to select this menu press control-option-space"
- press control-option-space

What I see is that somehow both the main menu and the inner menu bar are activated.  LEFT/RIGHT arrow keys result in both menu bars selection changing, etc.  This does not happen in the master branch (it does not seem to mention the control-option-space):


Screenshot 2025-10-09 at 13 09 19

andy-goryachev-oracle:
Could you also sync up with the master branch please?

andy-goryachev-oracle:
> a weird regression

any comments?

azuev-java:
> > a weird regression
> 
> any comments?

I was testing it and so far i can see that regression but i do not fully understand the reason behind it. Also this test behaves strangely with the useSystemMenuBar property toggle - when the "No Custom Menus" selected and the "use system menu bar" options checked the menu "No Custom Menus" is in the system menu bar but if i select anything else the items appear in the test frame. I have pulled changes from the master though.

andy-goryachev-oracle:
> i can see that regression but i do not fully understand the reason behind it

Is there enough value in this PR to treat the regression as a separate issue, or should we fix it in this PR?  

I am not sure that this is a rare corner case, since it happens when selecting a menu with an accessibility shortcut.  Maybe the fix is for the skin to consume that specific key event?

azuev-java:
> > i can see that regression but i do not fully understand the reason behind it
> 
> Is there enough value in this PR to treat the regression as a separate issue, or should we fix it in this PR?
> 
> I am not sure that this is a rare corner case, since it happens when selecting a menu with an accessibility shortcut. Maybe the fix is for the skin to consume that specific key event?

Well i can trigger the same behavior without the new code with MonkeyTester - so it is not a regression. But i have to understand the reason of why it happens because i think - it is the same reason behind me not able to implement the editable text component accessibility - the fact that we are forwarding events to wrong or multiple nodes when accessibility is involved. So it deserves a separate bug and it deserves a lot of attention.

andy-goryachev-oracle:
> Well i can trigger the same behavior without the new code with MonkeyTester

Created a follow-up ticket: https://bugs.openjdk.org/browse/JDK-8369733
Could you add a comment with the steps to reproduce please?

azuev-java:
> > Well i can trigger the same behavior without the new code with MonkeyTester
> 
> Created a follow-up ticket: https://bugs.openjdk.org/browse/JDK-8369733 Could you add a comment with the steps to reproduce please?

Added step by step instruction on how to reproduce the same problem with the current nightly build of JavaFX.

PR #1929, opened 2025-10-04
8210547: [Linux] Uncontrolled framerate

tsayao:
As Michael Zucchi pointed out on the mailing list, the high framerate occurs because `glXSwapBuffers() `operates asynchronously. To ensure proper synchronization, you can call `glFinish() `afterward, which blocks until the buffer swap is fully completed. However, when using `glXSwapIntervalSGI`, the swap interval setting applies globally rather than per drawable. In contrast, `glXSwapIntervalEXT` provides per-drawable control, allowing finer-grained vsync behavior.

I don't know if there are scenarios when the unlimited frame rate is needed - if so we should provide a option.

See [https://wikis.khronos.org/opengl/Swap_Interval](https://wikis.khronos.org/opengl/Swap_Interval)

It also selects the correct visual for transparency which needs to be depth = 32 for X11.

---------
### Progress

### Issue
 * [JDK-8210547](https://bugs.openjdk.org/browse/JDK-8210547): [Linux] Uncontrolled framerate (**Bug** - P5)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1929/head:pull/1929` \
`$ git checkout pull/1929`

`$ git checkout pull/1929` \
`$ git pull https://git.openjdk.org/jfx.git pull/1929/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1929`

`$ git pr show -t 1929`

</details>

<a href="https://git.openjdk.org/jfx/pull/1929.diff">https://git.openjdk.org/jfx/pull/1929.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1929#issuecomment-3368250455)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/native-prism-es2/x11/X11GLContext.c
+++ b/modules/javafx.graphics/src/main/native-prism-es2/x11/X11GLContext.c
@@ -276,7 +276,6 @@ JNIEXPORT jlong JNICALL Java_com_sun_prism_es2_X11GLContext_nInitialize
         if (ctxInfo->glXSwapIntervalEXT == NULL) {
             ctxInfo->glXSwapIntervalEXT = (PFNGLXSWAPINTERVALEXTPROC)
                 glXGetProcAddress((const GLubyte *)"glXSwapIntervalEXT");
-            ctxInfo->glXSwapIntervalEXT(dInfo->display, dInfo->win, 0);
         }
     } else if (isExtensionSupported(ctxInfo->glxExtensionStr,
             "GLX_SGI_swap_control")) {
@@ -286,8 +285,13 @@ JNIEXPORT jlong JNICALL Java_com_sun_prism_es2_X11GLContext_nInitialize
         if (ctxInfo->glXSwapIntervalSGI == NULL) {
             ctxInfo->glXSwapIntervalSGI = (PFNGLXSWAPINTERVALSGIPROC)
                 glXGetProcAddress((const GLubyte *)"glXSwapIntervalSGI");
-            ctxInfo->glXSwapIntervalSGI(0);
         }
+    }
+
+    if (ctxInfo->glXSwapIntervalEXT != NULL) {
+        ctxInfo->glXSwapIntervalEXT(dInfo->display, dInfo->win, 0);
+    } else if (ctxInfo->glXSwapIntervalSGI != NULL) {
+        ctxInfo->glXSwapIntervalSGI(0);
     }
 
     ctxInfo->state.vSyncEnabled = JNI_FALSE;


kevinrushforth:

/reviewers 2

tsayao:
I'm not sure of the solution, but it's a starting point.

kevinrushforth:
@arapte If you could also take a look that would be helpful.

notzed:
Nice, I (Michael Z) can confirm it fixes the frame throttling here, although in a simple animation all rendered frames have a pulse-presentation latency of 33ms -- which is somewhat better than the 48ms previously.

I've been doing some testing of the pulse-presentation latency, taking timestamps of the start of the pulse, the end of the render (before glXSwapBuffers), and the GL_TIMESTAMP of when glXSwapBuffers() is processed (via glQueryCounter()).  Plus tracking the pulse number all the way through so they can be matched.  This is some of the output with the patch:

    event                               pulse-submit   present

    pulse 1050 17484397796
    present 1048 17467793037 17484194371  -   +16.697  +33.098
    pulse 1051 17501064359
    present 1049 17484490371 17500860741  -   +16.782  +33.152
    pulse 1052 17517726363
    present 1050 17501155852 17517524593  -   +16.758  +33.127
    pulse 1053 17534403585

(This is in logging order, not temporal)

As a comparison using the existing SwapIntervalSGI and calling glFinish() after all scenes have rendered results in a pulse-presentation latency of 16.7ms instead.  Adding the glFinish() to the SwapIntervalEXT version has no effect.

    pulse 1050 17478236976
    present 1048 17461465333 17461674222  -   +16.569  +16.778
    pulse 1051 17494927765
    present 1049 17478075556 17478273482  -   +16.443  +16.641
    pulse 1052 17511531058
    present 1050 17494771852 17494973482  -   +16.535  +16.737
    pulse 1053 17528240031

??? I don't know why they'd be different given they both have the same order of pulses, but the timing code is the same.

For the curious, [I blogged about the timings I did yesterday](https://www.zedzone.au/post/0199c6c0edb7), I haven't added any plots with this patch but they look the same as "Build 2" but with another frame of latency to the present times (around 33 vs 16-17ms).

> I don't know if there are scenarios when the unlimited frame rate is needed - if so we should provide a option.

Well if prism.vsync=false it wont call swapinterval and will just run at the pulse rate, plus there's javafx.animation.fullspeed?

notzed:
Sorry I was wrong and should've tested with more than 1 window.  The change to glXSwapIntervalEXT() doesn't work properly and behaves as if glFinish() is called for each window after glXSwapBuffer(), each causing a new wait for vblank.

lukostyra:
> > I don't know if there are scenarios when the unlimited frame rate is needed - if so we should provide a option.
> 
> Well if prism.vsync=false it wont call swapinterval and will just run at the pulse rate, plus there's javafx.animation.fullspeed?

`javafx.animation.fullspeed=true` has the same effect as `prism.vsync=false` and more. [Both disable vsync](https://github.com/openjdk/jfx/blob/master/modules/javafx.graphics/src/main/java/com/sun/prism/impl/PrismSettings.java#L118) in order to uncap the framerate from monitor's refresh rate, and in addition to that, `javafx.animation.fullspeed=true` also [sets the pulse timer to immediately trigger the next pulse](https://github.com/openjdk/jfx/blob/master/modules/javafx.graphics/src/main/java/com/sun/scenario/animation/AbstractPrimaryTimer.java#L263).

While not applicable during day-to-day JavaFX use, unlimited framerate can actually be quite useful. We currently actively use `javafx.animation.fullspeed=true` when working on Metal and D3D12 to do performance and stability testing. It would be good to keep it around and functional on Linux as well.

modules/javafx.graphics/src/main/native-prism-es2/x11/X11GLFactory.c
```diff
@@ -215,7 +218,26 @@ JNIEXPORT jlong JNICALL Java_com_sun_prism_es2_X11GLFactory_nInitialize
...
+            fbConfig = fbConfigList[i];
+            visualInfo = vi;
+            break;
+        }
+
+        XFree(vi);
+    }
+
+    if (visualInfo == NULL) {
+        fbConfig = fbConfigList[0];
```
lukostyra, line 237: If I understand this correctly, we're just fetching the first available option from the list which might not always have windows transparency. If that's the case - would it make sense to print some sort of warning on the terminal that window transparency might not work correctly? I think it could be a valuable information to the user, or at least when debugging in the future.

@kevinrushforth it's also a question to you, cause I'm not 100% sure what the JFX strategy would be in this case - if we would want this to be printed always as a warning, or only when we ex. activate `-Dprism.verbose=true`

lukostyra (commented):
I only have one code review remark, I still have to do my testing for this change.

lukostyra:
I had problems verifying this on my Ubuntu VM - the framerate did go a bit lower, but it also did not properly cap after this change (FPS was fluctuating between 40 and 100 with 60Hz refresh rate).  However, I think this must be a problem related specifically to the VM GPU driver.

I dug out an old laptop with Intel UHD Graphics card that runs Fedora natively and (after hours of installing updates) can confirm both the issue reproduces on latest JFX version and that this change does fix the issue.

I also noted that there is a third swap control extension, `MESA_swap_control`. I tried it hoping it would help on the Ubuntu VM but with the same result as the above. I think there's no reason to pursue the solution to this any further for the VM, it works on native systems properly and that matters the most. I'll be happy to approve this once we address the comment I left.

tsayao:
From what I understood, @notzed concluded that the change didn’t work as expected. Please let me know if I misunderstood.

notzed:
Yeah it doesn't work.  FWIW I've now got an OCP done - it took a couple of weeks so I moved onto other things and haven't revisited it.  I didn't get much traction discussing it on the list either.

I do have a couple of patches which fix the problem but I wanted to do some latency testing to confirm it's further utility - but should I just submit a pull request anyway?

lukostyra:
> I do have a couple of patches which fix the problem but I wanted to do some latency testing to confirm it's further utility - but should I just submit a pull request anyway?

@notzed so if I understand this right, this will be a patch that effectively replaces this one?

If whatever your changes do will replace this PR we should probably close this PR and let you submit your own PR. But depending on how complex it is we might have to spend quite a bit of time testing it.

I think for a start, once your changes are ready and your testing is complete, you could create a draft PR and link it here so we can take a closer look at it and decide what to do with it.

notzed:
I thought this one was done, so absence of any feedback at all I just opened https://github.com/openjdk/jfx/pull/1981

lukostyra:
As discussed in #1981 that approach is in fact better. I think we can close this PR in favor of the other one.

tsayao:
Closing in favor of #1981. I will send the improvement on the visual selection on another PR.

PR #1935, opened 2025-10-13
8089514: [TableView, TreeView, ListView, TreeTableView] Clicking outside of the edited cell, node, or entry should commit the value

Maran23:
This is an implementation for the long standing issue of allowing to commit a cell value when the focus is lost or the editing index (cell) changed. This also contains [JDK-8089311](https://bugs.openjdk.org/browse/JDK-8089311) (for better understanding the usecase with `TextField` cells, but we may split this later on).

TLDR
-
- New method `stopEdit()`
- Called when the editing operation was stopped OR the focus is lost. 
  - Both are situations where we can not guess what should happen. We do know though that cancelling the edit without further customization is wrong. Instead, developers can now decide what to do
  - Examples: 
    - Editing operation can be stopped by the cell container, because the developer called `edit()` with another index
    - The user scrolled until the cell is reused
    - Focus is lost because the user clicks on another focusable Node
  - The default implementation will call `cancelEdit()`
    -  To be backwards compatible
    - Because we do not know what value we should commit. A `Cell` does not know what it means to be in the edit state. However, subclasses will know that better. A `TextFieldTableCell` knows the value of the `TextField` and can directly commit (or convert before) the text entered from the User.
- The existing `TextFieldXXXCell`s will utilize and call `stopEdit()`, and they know exactly the value they can commit (the text)

Behavioral Change
-
- Developers can override `stopEdit()` and call `commitEdit()` with the desired value
- All `TextFieldXXXCell`s will now commit their value when the editing operation was stopped
  - They still cancel their edit when Escape is pressed
- Developers that subclassed `TextFieldXXXCell`s to implement their own commit-on-focus-lost handling may need to remove their custom logic
  - As outlined in the comments, this is probably a very rare case because you can not access the `TextField` when subclassing it
    - But if developers did indeed subclass it: All `Cell` operations, that change the editing state, will check `isEditing()` before. So there might be no problem at all and they will continue working, However, it is recommended to adapt to the new `stopEdit` flow
- Developers can check out the `TextFieldXXXCell`s if they wish to adapt the pattern to their own cell implementations

- From my personal experience, in every project I saw a custom implementation of a `TextFieldTableCell` (Note: Not using or  subclassing the one from JavaFX) that tried to implement focus lost on their own.
I guess everyone who tried to support this behavior found their way to do so (in a hacky way).
  - Those implementations can now adapt the `stopEdit` method and remove their custom logic if they wish

---

<details>
<summary>Considerations and Follow Ups</summary>

Considerations
-
- I tried to make the API minimal, and without breaking changes (other than the `TextField` cells committing their values, but we may split this up)
- Clicking the `ScrollBar` now commits/cancels the edit. I checked other applications and this is very common. But something I want to note. This can be fixed by utilizing `focusWithin` instead of relying on hacky focus requests.
- It might be hard for a developer to exactly know the cause why `stopEdit` is called (is it because I called `edit()`, or because the focus is lost, or because I scrolled and the cell is reused?
However, this does not seem like a problem, as the developer has full control and can easily implement checks or register listeners if needed. The default behavior is intuitive.

Possible Follow Ups
-
- The Cell Container focus behavior is, well, weird right now. That is why consuming the event is needed to better support this PR. One thing we may can consider is using the `focusWithin` property instead for all 4 Cell Containers. And not calling `requestFocus` for nearly every `MOUSE_PRESSED` event. This is a problem with other `Control`s as well, so needs more research.
- See, if `focusTraversable` can be improved, as it right now means not grabbing focus at all. (Existing 'problem', might not even be one).

</details>

---------

---------
### Progress
- \[ \] Change requires CSR request [JDK-8380576](https://bugs.openjdk.org/browse/JDK-8380576) to be approved

### Issues
 * [JDK-8089514](https://bugs.openjdk.org/browse/JDK-8089514): [TableView, TreeView, ListView, TreeTableView] Clicking outside of the edited cell, node, or entry should commit the value (**Enhancement** - P3)
 * [JDK-8089311](https://bugs.openjdk.org/browse/JDK-8089311): [TableCell] commit on focus lost not possible in every case (**Bug** - P4)
 * [JDK-8092017](https://bugs.openjdk.org/browse/JDK-8092017): TextFieldListCell forgets entered value on focus change (**Enhancement** - P3)
 * [JDK-8091001](https://bugs.openjdk.org/browse/JDK-8091001): Editable TableView commit on focus change doesn't always work (**Enhancement** - P4)
 * [JDK-8380740](https://bugs.openjdk.org/browse/JDK-8380740): [TableView, TreeTableView] clearing editable flag when editing (**Bug** - P4)
 * [JDK-8380576](https://bugs.openjdk.org/browse/JDK-8380576): [TableView, TreeView, ListView, TreeTableView] Clicking outside of the edited cell, node, or entry should commit the value (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1935/head:pull/1935` \
`$ git checkout pull/1935`

`$ git checkout pull/1935` \
`$ git pull https://git.openjdk.org/jfx.git pull/1935/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1935`

`$ git pr show -t 1935`

</details>

<a href="https://git.openjdk.org/jfx/pull/1935.diff">https://git.openjdk.org/jfx/pull/1935.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1935#issuecomment-3397322244)
</details>


kevinrushforth:
I think this needs more discussion before it is ready for formal review. Perhaps it should be moved to Draft until the discussion converges on an approach?

/reviewers 2 reviewers
/csr

Maran23:
> I think this needs more discussion before it is ready for formal review. Perhaps it should be moved to Draft until the discussion converges on an approach?

Jup! That was also the intention of this PR, to have a proof of concept to look at.

Still, this PR can already be tested by volunteers, as not the API but the approach itself needs feedback too!

Maran23:
@andy-goryachev-oracle I wonder if a new page in the monkey tester with a table and various controls to test out focus loss makes sense? Just an idea though :)

andy-goryachev-oracle:
> a new page in the monkey tester with a table and various controls to test out focus

In a sense, we already have that: the TableView page allows for editing the cells, and there is plenty of various controls in the options pane.
There is a full-blown UI panel in the Stage page (in a standalone repo), through you have to open it separately.

We could also cook up completely new Focus page with all of our controls and event handlers built-in.  Is that what you envision?

Maran23:
> In a sense, we already have that: the TableView page allows for editing the cells, and there is plenty of various controls in the options pane. There is a full-blown UI panel in the Stage page (in a standalone repo), through you have to open it separately.
> 
> We could also cook up completely new Focus page with all of our controls and event handlers built-in. Is that what you envision?

Yeah, something like that. I will do some testing as well soon, it would be very interesting how a table with the correct focus mechanism reacts to different controls, especially with validation, e.g. a table row need to have values so the button is not disabled - and what happens if you commit an invalid value by a focus loss click on a button.

andy-goryachev-oracle:
> it would be very interesting how a table with the correct focus mechanism reacts to different controls

Added an option to the monkey tester which allows to trace focus events and start/stop/cancelEdit calls:

Screenshot 2025-10-15 at 13 45 02

I suppose we can add a simple validator as well (if you do please let me know and I merge your changes in).

danielpeintner:
It's a pity that this suggestion has received relatively little attention. I think JavaFX is a very good framework. Nevertheless, it's very annoying that TableView behaves differently than most users want it (or expect it) to be. 
Not committing a value when focus is lost or clicking outside a cell is counterintuitive.

I think the proposals brings JavaFX a BIG step forward w.r.t out-of-box usability instead of having developers to work around it... with limited success.
Is there any possibility that the idea will be pursued further?

Maran23:
> It's a pity that this suggestion has received relatively little attention. I think JavaFX is a very good framework. Nevertheless, it's very annoying that TableView behaves differently than most users want it (or expect it) to be. Not committing a value when focus is lost or clicking outside a cell is counterintuitive.

I agree and I plan to come back to this. Currently, I don't have much freetime (I'm on the JavaLand right now). 

If you can, feel free to try it out and give feedback! More information is also in the Mailing list entry: https://mail.openjdk.org/archives/list/openjfx-dev@openjdk.org/message/6ZP7PDMPXFKFPQOW64TUZCWNNQCL7KTT/

The current state is stable and is fine for the usecase. It also follows the other commit-on-focus-loss Controls. There are some workflows that are a bit weird or may not work (e.g. non focus traversable button, more in the mailing list), but those are all preexisting issues that we do have with e.g. `ComboBox` already.

Maran23:
Copying my text from the mailing list:

```md
I did a lot of focus tests the last weeks and wrote a big sampler application as well.
I will provide the source code as Gist in the PR description at one point.
 
First of all, things look pretty good! There are some problematic cases, and at least one bug from what I can see.
Note: All the problematic cases below also affect all other Controls that commit their value on focus loss, like DatePicker, Spinner, ...

What works:
- By default, the focus loss commit works well with all Controls out of the box
- Mnemonics work. They will first request focus before they trigger the action
 
Problematic cases:
- A Tab selection change is fired before the focus is requested on the TabPane. While the focus lost commit still works, it is in my opinion too late. Think about disabling a Tab when something in a Table is invalid (which we know after the commit)
- List/Tree/Table/View: Selection and focus is changed before the actual cell container will receive the focus. Even more weird: For both Tables, the selection change is fired BEFORE the focus change. List- and TreeView have the correct order. Again, in case we want to disable this Control when another Table is invalid, which we know right at the commit, we will still first trigger a selection and focus change
 
Buggy cases:
- As John also mentioned, non focus traversable Controls are completely broken. A non focus traversable Button or CheckBox will not request any focus. Therefore, Cells, DatePicker, ... can not commit their value. Other Controls set to non focus traversable like TextField or the TextField inside the DatePicker will request focus, the DatePicker button will not. This seems like a bug to me. Could also be something we want to ignore, as we can say: In this case, all focus loss commits are broken, so it is up to you. Feedback welcome
 
Other cases:
- MenuBar, Menu, MenuItem will not trigger focus at all. This might be expected. I don't know what to think about that. Feedback welcome
```

The application for testing commit on focus lost:
image

https://gist.github.com/Maran23/7f1cef17b3ed638f3b9c65f911a8b390

Maran23:
/open

I just did some retesting, and everything works as expected. I think this is ready for review. All issues that I found do exist with other commit-on-focus-loss `Control`s as well (`ComboBox`, `DatePicker`, `Spinner`). So there really is no blocker for this feature.

Note: I will eventually search/create tickets for my focus findings (and if I have even more time, fix them). So those are on my list. But again, no blocker for this feature! :)

Maran23:
/issue add JDK-8091001

Note: We could also close the issues that are very closely related.
I decided to first add all issues here that I can confirm are fixed by this PR, so we do not have any stale issues later on and I feel like we can manage this better in one place (here). 

But fine with whatever we agree on (Either way, it's a bit clearer now what issues are affected).

andy-goryachev-oracle:
FYI: Filed a possibly related bug: https://bugs.openjdk.org/browse/JDK-8380740

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/CellBehaviorBase.java
```diff
@@ -173,7 +173,7 @@ public void mousePressed(MouseEvent e) {
         } else {
             latePress  = isSelected();
             if (!latePress) {
-                doSelect(e.getX(), e.getY(), e.getButton(), e.getClickCount(),
+                doSelect(e, e.getX(), e.getY(), e.getButton(), e.getClickCount(),
```
andy-goryachev-oracle: might as well pass only the event pointer

modules/javafx.controls/src/main/java/javafx/scene/control/Cell.java
```diff
@@ -353,26 +351,39 @@ public Cell() {
...
+     * The default behavior is to cancel the edit.
+     * This method is meant to be subclassed in case the default behavior is not enough.
+     * For example, subclasses decide to rather commit the edit operation instead of cancelling.
+     */
+    public void stopEdit() {
+        if (!isEditing()) {
+            return;
+        }
+
+        cancelEdit();
```
andy-goryachev-oracle, line 388: this might be clearer:
```
        if (isEditing()) {
            cancelEdit();
        }
```

modules/javafx.controls/src/main/java/javafx/scene/control/TableCell.java
```diff
@@ -569,39 +569,40 @@ private void updateFocus() {
...
     }
 
     /**
-     * Switches an editing cell into not editing without changing control's
-     * editing state.
+     * Stops the edit operation.
+     * If not overwritten, will cancel the edit without changing control'sediting state.
      */
-    private void doCancelEdit() {
+    private void doStopEdit() {
```
andy-goryachev-oracle, line 596: feels like a misnomer: typically `doSomething` denotes a terminal operation, but here it calls stopEdit()...  it's ok, just confusing.

modules/javafx.controls/src/main/java/javafx/scene/control/TableRow.java
```diff
@@ -322,10 +322,10 @@ private void updateEditing() {
...
         boolean rowMatch = editCell == null ? false : editCell.getRow() == getIndex();
 
-        if (! isEditing() && rowMatch) {
+        if (!isEditing() && rowMatch) {
             startEdit();
-        } else if (isEditing() && ! rowMatch) {
-            cancelEdit();
+        } else if (isEditing() && !rowMatch) {
+            stopEdit();
         }
```
andy-goryachev-oracle, line 329: I know it is in the original code, but we evaluate isEditing() twice.  
This might be slightly better:
```
if (isEditing()) {
    if(!rowMatch) {
        stopEdit();
    }
} else {
    if(rowMatch) {
        startEdit();
    }
}    
```

modules/javafx.controls/src/main/java/javafx/scene/control/TreeCell.java
```diff
@@ -592,34 +592,37 @@ private void updateEditing() {
...
+        } else if (!match && editing) {
+            doStopEdit();
         }
     }
 
-    private void doCancelEditing() {
-        // If my tree item is not the one being edited then I need to cancel
+    /**
+     * Stops the edit operation.
+     * If not overwritten, will cancel the edit without changing control'sediting state.
```
andy-goryachev-oracle: `control'sediting`
Also, what does it mean: "without changing control's editing state" exactly?

modules/javafx.controls/src/main/java/javafx/scene/control/TreeTableCell.java
```diff
@@ -564,44 +564,43 @@ private void updateFocus() {
...
+        } else if (!match && editing) {
+            doStopEdit();
         }
     }
 
     /**
-     * Switches an editing cell into not editing without changing control's
-     * editing state.
+     * Stops the edit operation.
+     * If not overwritten, will cancel the edit without changing control'sediting state.
```
andy-goryachev-oracle: `control'sediting`
Also, what does it mean: "without changing control's editing state" exactly?

andy-goryachev-oracle (commented):
Some early (and minor) code review comments.  This needs more testing and consideration.

Would it be possible to expand the PR description to illustrate a couple of examples wrt changes in behavior and also the kind of changes the application need to make to adapt to the new behavior?

Perhaps extract it into a JEP-like markdown page written for people who have minimal knowledge of the problem?

One question I must ask is whether it's possible / worth it to hide the new behavior behind a system property, for people who spent a lot of time making their apps work with the existing behavior and who are reluctant / unable to spend more time fixing it again?

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/CellBehaviorBase.java
Maran23: I tried to keep the diff minimal. But if you like, I can also change that, fine for me

modules/javafx.controls/src/main/java/javafx/scene/control/TableRow.java
Maran23: Yeah I agree. I would rather create a new ticket, as this is not the only thing that is weird in cell code. Will have a deeper look soon and create a ticket to improve that things.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/CellBehaviorBase.java
andy-goryachev-oracle: the diff is not structural, and since you added the event pointer it makes less sense to dereference and pass all the fields...

Maran23: Then I would also adjust `simpleSelect` as well - same problem there.

Maran23:
> Would it be possible to expand the PR description to illustrate a couple of examples wrt changes in behavior and also the kind of changes the application need to make to adapt to the new behavior?

Sure. But to be honest, I don't know what developers need to adjust. It's hard to say what existing apps have done so far to fix this issue - or whether they've simply taught their users to always press the Enter key. 😀

Since this change has no impact on existing cells, it's not a problem.
It only affects subclassed `TextFieldXXXCell`s. I don’t think developers have done that very often. Since you can't access the `TextField`, but you need it to implement the commit when focus is lost behavior.

That's exactly what I did. A custom cell implementation with a `TextField`, a bunch of listeners, and an attempt to guess when `cancelEdit` is called even though a commit should happen instead - and then triggering a `commitEdit` instead.

> One question I must ask is whether it's possible / worth it to hide the new behavior behind a system property, for people who spent a lot of time making their apps work with the existing behavior and who are reluctant / unable to spend more time fixing it again?

This would only help for the case described above - If developers subclasses `TextFieldXXXCell`s. We could do that here, sure, and I don't have a problem with that at all! But I don't think this was done often, if at all.

modules/javafx.controls/src/main/java/javafx/scene/control/Cell.java
```diff
@@ -353,26 +351,39 @@ public Cell() {
...
     }
 
+    /**
+     * Stops the edit operation of the cell.
+     * This method is called when another cell is edited or the focus changed.
+     * <p>
+     * The default behavior is to cancel the edit.
+     * This method is meant to be subclassed in case the default behavior is not enough.
+     * For example, subclasses decide to rather commit the edit operation instead of cancelling.
+     */
```
andy-goryachev-oracle, line 382: needs a `@since 27`

modules/javafx.controls/src/main/java/javafx/scene/control/Cell.java
```diff
@@ -353,26 +351,39 @@ public Cell() {
...
             }
         });
 
         // initialize default pseudo-class state
         pseudoClassStateChanged(PSEUDO_CLASS_EMPTY, true);
     }
 
+    /**
+     * Stops the edit operation of the cell.
+     * This method is called when another cell is edited or the focus changed.
```
andy-goryachev-oracle: called by whom?
is it meant to be called by the application?
can it be made `protected`?

andy-goryachev-oracle (commented):
Still a preliminary review - I am trying to understand the details at the high level.

So I see one new API (that should be listed in the **API** section), and a bunch of behavioral changes that are listed in the **API** and **Considerations** sections.

Maybe move the behavioral changes into its own **Behavioral Changes** section?  Is it possible/feasible to show the changes in a "Before/After" table, or at least provide some examples to start with?

Specifically,

> All TextField cells now commit their value (instead of cancel) on focus loss

what happens when there is a validation error coming from the value converter?

> TextField Escape handling was badly implemented (it was never really called, as the cell container handled Escape before)

what was the problem before exactly?  how does it work now?

modules/javafx.controls/src/main/java/javafx/scene/control/Cell.java
Maran23: Made it public as the other edit methods are as well. Also this method is inteded to be called from outside, e.g. behavior (even though that is not public, you could very well write your own in the skin. 

I adjusted the javadoc to make more clear, what it does and why. Probably needs some more fine tuning.

modules/javafx.controls/src/main/java/javafx/scene/control/TableRow.java
Maran23: I found some more oddities in the cell classes. Every cell has a slightly different implementation for almost everything (updateEditing, updateFocus, updateSelection, editing, ...). 

I want to create a Ticket and soon a PR to address this. Just want to check in if that is also something you would like to see?

andy-goryachev-oracle: Hard to say - I haven't looked at the cells specifically.  The biggest concern, I think, is the risk of regression, followed by the shortage of bandwidth - do we want to spend time on small cleanup vs. huge issues (like this PR).

Having said that, I think it might be a good idea, especially if there is a ton of bugs filed for a particular component or behavior.  At the end, we have to rely on our judgement to determine when it's worth doing or not doing something.

What are the major pain points you want to work on?

Maran23: I went into a rabbit hole yesterday. The 6 `IndexedCell` subclasses are so different. Every cell implemented `updateFocus`, `updateEditing`, `updateSelection` a bit different. E.g. when to clear it, when to return. Of course no tests.

It's crazy how that made it into JavaFX. The best way would be to unify the behavior and write parameterized tests for all 6 cells and test the same methods.

To be honest I don't know how to go from here. It would be nice to unify and have a stable reliable cell implementation for all cells. 
But as you said, it is a rather 'small' cleanup, although it will make our life easier in the future.

andy-goryachev-oracle: Exactly my sentiment!

We have ~3,500 bugs filed (some of them are probably duplicates, but I think there is probably as many not yet filed, since I find half a dozen each week), so we know it's a large and complex system.

Architecturally, there are things I find absolutely [REDACTED], like `Event.target` field which requires to spawn multiple copies of each event, or a lack of an input map.  There are things that should have done differently/better, like focus subsystem or the text layout.  It's overwhelming, but javafx is also an incredible toolkit that gives us ability to develop cross-platform applications in a sane manner.  

One thing we _could_ do is, as you said, develop tests.  We (sometimes) have simple API tests, but we also need to document the behavior (at least the default behavior) and codify that in form of tests.  Do we expect all cells handle the focus in a consistent manner?  A parameterized test might be a very good idea!  So thank you for your effort and participation, many thanks indeed!

Maran23: > Architecturally, there are things I find absolutely [REDACTED]

Same. I probably had the same word in my mind, but decided to rewrite my previous comment so it is not included.🙃

> Do we expect all cells handle the focus in a consistent manner

Yes, and they mostly do that. What they don't do is to revoke focus or editing or selection in the same manner, especially is the selection mode changes or the focus model is `null`ed

> So thank you for your effort and participation, many thanks indeed!

I really appreciate that.  It’s not easy to always allocate time for this, especially when you’re working on what seems like a small task that then turns into a much bigger problem and leads to a whole series of "maybe we should tackle this first" scenarios.

modules/javafx.controls/src/main/java/javafx/scene/control/cell/CellUtils.java
```diff
@@ -242,6 +242,11 @@ static <T> void cancelEdit(Cell<T> cell, final StringConverter<T> converter, Nod
 
     static <T> TextField createTextField(final Cell<T> cell, final StringConverter<T> converter) {
         final TextField textField = new TextField(getItemText(cell, converter));
+        textField.focusedProperty().addListener(_ -> {
+            if (!textField.isFocused() && cell.isEditing()) {
```
andy-goryachev-oracle: minor: check for isEditing() first, as it will fail fast in case when no action is needed (focused, but not editing).

modules/javafx.controls/src/main/java/javafx/scene/control/cell/CellUtils.java
```diff
@@ -255,7 +260,7 @@ static <T> TextField createTextField(final Cell<T> cell, final StringConverter<T
             cell.commitEdit(converter.fromString(textField.getText()));
             event.consume();
         });
-        textField.setOnKeyReleased(t -> {
+        textField.setOnKeyPressed(t -> {
```
andy-goryachev-oracle, line 263: this is interesting - could there be any compatibility issues after changing from `onKeyReleased` to `onKeyPressed`?  what was the reason for this change?

Also, this should probably be mentioned in the `Behavioral Changes` section.

modules/javafx.controls/src/main/java/javafx/scene/control/Cell.java
```diff
@@ -353,26 +351,44 @@ public Cell() {
...
-        super.focusedProperty().addListener(new InvalidationListener() {
-            @Override public void invalidated(Observable property) {
-                pseudoClassStateChanged(PSEUDO_CLASS_FOCUSED, isFocused()); // TODO is this necessary??
-
-                // The user has shifted focus, so we should cancel the editing on this cell
-                if (!isFocused() && isEditing()) {
-                    cancelEdit();
-                }
+        focusedProperty().addListener(_ -> {
+            pseudoClassStateChanged(PSEUDO_CLASS_FOCUSED, isFocused()); // TODO is this necessary??
```
andy-goryachev-oracle: is this comment _still_ necessary?

andy-goryachev-oracle (commented):
The code changes look good (with some minor comments).

It may take me some time to test all the controls involved and all the scenarios, as well as to review the prior discussion (for example, https://github.com/kleopatra/swingempire-fx/wiki/CellEditAPI )

in the mean time, you may want to keep this thing alive by merging the latest master and issuing  a `/keepalive` command?

modules/javafx.controls/src/main/java/javafx/scene/control/Cell.java
Maran23: I think `pseudoClassStateChanged` is not necessary here. Maybe I will file a bug and create a simple PR for that, good point.

andy-goryachev-oracle: Not sure I understand your response: the code that toggles PSEUDO_CLASS_FOCUSED is likely to be necessary (it's a private constant, and this is the only spot it's referenced from).
The `comment` does not seem to be needed.

Or am I completely off the mark here?

Maran23: This is already done in the `focused` property, so not needed manually at all:

https://github.com/openjdk/jfx/blob/e091e1cbbc446026f6dce9b796bfc8de73fe219d/modules/javafx.graphics/src/main/java/javafx/scene/Node.java#L10230

https://github.com/openjdk/jfx/blob/e091e1cbbc446026f6dce9b796bfc8de73fe219d/modules/javafx.graphics/src/main/java/javafx/scene/Node.java#L8400-L8404

modules/javafx.controls/src/main/java/javafx/scene/control/cell/CellUtils.java
Maran23: Can mention it, yes. But I can't think of any change.
Before, this was never called. Because the behavior of the cell handled it first. Because we were listening only on key release here. Now, we are first, and we can cancel the edit (as before) instead of any focus loss commit.

andy-goryachev-oracle:
Sort of related question, since I don't have much experience with editable virtualized controls.  One thought keeps nagging me while looking at these changes: why do we make decision to cancel or commit at the cell level?  Should it be rather done on the control level, since it's the control that has the full picture?

Maybe we should consider a sort of editing policy property which would allow the application to specify what to do in each particular case (focus lost, ESC, ...), or even based on the content of the cell if that's a part of the requirement?  So, instead of the application overriding the cells, it would simply provide a policy or use one of the standard policies?

Does it make sense?

andy-goryachev-oracle (approved):
Tested `TableView`, `TreeView`, `TreeTableView`, and `ListView` on macOS.  All work as expected - a major improvement in usability, I think.  Many thanks for your hard work!

This PR also fixes 
https://bugs.openjdk.org/browse/JDK-8380740
you can probably `/issue add` it.

Maran23:
> Sort of related question, since I don't have much experience with editable virtualized controls. One thought keeps nagging me while looking at these changes: why do we make decision to cancel or commit at the cell level? Should it be rather done on the control level, since it's the control that has the full picture?
> 
> Maybe we should consider a sort of editing policy property which would allow the application to specify what to do in each particular case (focus lost, ESC, ...), or even based on the content of the cell if that's a part of the requirement? So, instead of the application overriding the cells, it would simply provide a policy or use one of the standard policies?
> 
> Does it make sense?

Well it kind of works like that already. Just that the concerns are separated in:
- `TableView`, provides the items and columns
- `TableColumn` need to decide what to do on start, commit and cancel (`setOnEditCommit(..)`)
  -  Typically, you call a setter with your new value on commit
- `TableCell` need to know what it should do on edit start, commit and cancel (e.g. show a graphic like `TextField`, show a `ComboBox`, a Popup, ....)
  - It also need to decide when to cancel (e.g. Escape) or commit. If focus loss commit is not desired, it does not need to be implemented
    - Some Cells may do not even care about Escape or focus loss, like a `CheckBoxTableCell`, which can just be clicked (will also never be in the editing state really).
  - Inform the column then about the event, but might also need to know how to convert e.g. a String into type <T> 

The API makes sense and that is also why cell implementations like the `TextFieldTableCell` exist, so not much code is needed in order to create tables with everything that is usually needed. So most of the time, you don't need to override anything.

So any API on `TableView` level might not really work - too many different things are involved. And the cells can be vastly different. It is hard to find a common point here.
And most of the time there is not much configuration needed, and hopefully even less after this PR.

> Should it be rather done on the control level, since it's the control that has the full picture?

With that said, there are things that are very very hard to actually find out / implement. Yes, the control should have the full picture, but has not. For example, there is no easy to find out what cell is actually editable. The table knows if it is editable and if the columns are, but has no idea about the `TableRow` or `TableCell`. Very similar that it does not know what the `TableCell` does, beside edit, cancel and commit.

Also the table does not get a notification event that a commit happened, so it does not know the full picture in this regard either.
That is why at one point implemented something like that in my columns:

```
        Event.fireEvent(tableView, new CommitEvent<>(this, tableView, ExtendedTable.preCommitEvent(), item));
        commitValue(item, newValue);
        Event.fireEvent(tableView, new CommitEvent<>(this, tableView, ExtendedTable.commitEvent(), item));
```

That would be indeed helpful.

andy-goryachev-oracle:
looks like some tests need to be fixed

andy-goryachev-oracle:
> And most of the time there is not much configuration needed, and hopefully even less after this PR.

+1

> Yes, the control should have the full picture, but has not. For example, there is no easy to find out what cell is actually editable. The table knows if it is editable and if the columns are, but has no idea about the TableRow or TableCell.

Exactly.  In swing, it's delegated to the model if I recall correctly, but in FX we just have an observable list of items.  So the only place the application can implement a per-cell editability is to override TableView::edit().  Guess it is possible, just not elegant.



> Also the table does not get a notification event that a commit happened, so it does not know the full picture in this regard either.
> That is why at one point implemented something like that in my columns:

```
        Event.fireEvent(tableView, new CommitEvent<>(this, tableView, ExtendedTable.preCommitEvent(), item));
        commitValue(item, newValue);
        Event.fireEvent(tableView, new CommitEvent<>(this, tableView, ExtendedTable.commitEvent(), item));
```

This looks like a great idea!  We have event system, and we should use it for this purpose as it allows to
- inform interested parties about changes
- allow the interested parties to veto a particular action

So I guess this is also possible with the existing APIs, perhaps we should consider making it a built-in functionality?

Maran23:
Seems like https://github.com/openjdk/jfx/pull/2124 was already worth it - it did uncover that `testJDK8119995` actually created a new temporary stage before although we did not expect it (Which was of course not visible to me when writing the test). Therefore, the test could not really test what it was supposed to.

Maran23:
> This looks like a great idea! We have event system, and we should use it for this purpose as it allows to
> 
> * inform interested parties about changes
> * allow the interested parties to veto a particular action
> 
> So I guess this is also possible with the existing APIs, perhaps we should consider making it a built-in functionality?

This would get a +1 from me. This is very useful. I used that often for e.g. saving. So I have 20 columns with different `setOnEditCommit` handlers (usually calling a setter, but could be more complex) and one on the table with the `ExtendedTable.commitEvent()`, where I just get the item and save it.

modules/javafx.controls/src/test/java/test/javafx/scene/control/cell/TextFieldTableCellTest.java
```diff
@@ -410,7 +410,7 @@ public void afterEach() {
         assertEquals(text, textField.getSelectedText());
 
         MouseEventFirer mouse = new MouseEventFirer(textField);
-        mouse.fireMousePressed(MouseButton.SECONDARY);
+        mouse.fireMouseReleased(MouseButton.SECONDARY);
```
andy-goryachev-oracle: is this right?  RELEASED followed by another RELEASE?

Maran23: Yes, because if you do a right click on a `TextField`, it will trigger the `ContextMenu` only on release.

EDIT: And the second RELEASE is to retrigger the `ContextMenu`

andy-goryachev-oracle: I recall this behavior might be platform-specific: on macOS the context menu appears on the PRESS, on windows on RELEASE.
Have you tested the change on all three platforms (incl. linux)?

Maran23: This is a headless test, so the OS does not play any role here.
I don't know 100% what the original purpose of this test was - as it never fulfilled any role.
But this is close to the original test while we test a real table setup.

In any OS it behaves correctly: Opening the `ContextMenu` from within the `TextField` will work and not stop the edit process (tested in the monkeytester). We try to reproduce that somewhat.

EDIT: We could also e.g. send a `ContextMenu` event instead (like a real JavaFX app does). But that would change the test again compared to the OG test. But would be fine for me

andy-goryachev-oracle: That's not what I meant.  The test, for whatever reason, is testing a perfectly valid scenario - mouse press followed by a release.  The real question is why did you have to change it?

The test works fine with and without your changes on macOS.  There might be something in the fx code that handles platform differences between mac and windows, so if the test failed on windows, it means this PR broke something.

andy-goryachev-oracle: yep, the original test is broken on windows with this change.

we shouldn't be changing test by replacing a legal sequence of events to an illegal one just to silence it.

andy-goryachev-oracle (changes requested):

modules/javafx.controls/src/test/java/test/javafx/scene/control/cell/TextFieldTableCellTest.java
Maran23: I must be blind - I thought that the test previously pressed and pressed again. And was wondering what it tried to achieve - just realized that it was press -> release.

well, that is probably what happens if you name your test `test_rt_32869` and let me guess what it does 🙃

andy-goryachev-oracle: point taken.
Rindfleischetikettierungsüberwachungsaufgabenübertragungsgesetz
:-)

Maran23: Thanks for bringing this up (although I don't know if RT really is an abbreviation for **R**indfleische**t**ikettierungsüberwachungsaufgabenübertragungsgesetz)

Jokes aside. What I found out is that the `TextField` has no skin yet - so no behavior which can actually consume the send events).
image

I now completely renamed the test, added javadoc, added more assertions like `assertNotNull(textField.getScene());` and also wrote the same for `TreeTableView`. No more guessing and weird names!

andy-goryachev-oracle (approved):
looks good on macos and windows - needs to be tested on linux.

modules/javafx.controls/src/main/java/javafx/scene/control/Cell.java
Maran23: Created: https://github.com/openjdk/jfx/pull/2176

andy-goryachev-oracle (approved):

andy-goryachev-oracle:
sorry, please reopen this PR.
@lukostyra could you be the second reviewer once it's been reopened?

modules/javafx.controls/src/main/java/javafx/scene/control/Cell.java
```diff
@@ -353,24 +351,42 @@ public Cell() {
...
          * ListView defines zero or one cell as being the "focused" cell. This cell
          * would have focused set to true.
          */
-        super.focusedProperty().addListener(new InvalidationListener() {
-            @Override public void invalidated(Observable property) {
-                // The user has shifted focus, so we should cancel the editing on this cell
-                if (!isFocused() && isEditing()) {
-                    cancelEdit();
-                }
+        focusedProperty().addListener(_ -> {
```
beldenfox, line 359: I realize this is old behavior but using the focused property is questionable. The focused property will be set to false if the user activates a different window. In most toolkits clicking on a different window would not commit text entry. The text should just sit there waiting for the user to re-activate the original window.

The more correct way of doing this is to observe the scene's focusOwner since that doesn't change when the window loses focus. But it's possible you're picking up other behavior by monitoring the focused property that's important, I know the controls mess with it in various ways.

andy-goryachev-oracle: > The text should just sit there waiting for the user to re-activate the original window.

this behavior might depend on the application requirements, so perhaps it needs to be configurable?

Maran23: This just follows the same behavior as `DatePicker`, `ComboBox`, `Spinner`. 

However, I do think we could improve that handling in a follow up. 
I have this idea for more than a year already:
Instead of registering a focus listener, we should rather have a protected `focusChanged` method. 
Two advantages:
- Developers can finally completely override that behavior
- Less listeners

With the `focusOwner` idea, we may have a third advantage / usecase:
Since focus is a concept on the scene, we also could call this method on focus owner change instead.

And I think this can be a toggle on the `Scene` and is actually possible to implement with full backward-compatibility. I think. 

Something I would explore after this PR. It is not a blocker IMO.

PR #1945, opened 2025-10-23
8370498: Improve how Node detects whether a layout property change requires a new layout pass

hjohn:
This new check is much more accurate to detect whether a parent is currently laying out its children. The previous code almost never worked, resulting in additional unnecessary layouts.

---------
### Progress
- [ ] Change must be properly reviewed (3 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 2 [Authors](https://openjdk.org/bylaws#author))

### Issue
 * [JDK-8370498](https://bugs.openjdk.org/browse/JDK-8370498): Improve how Node detects whether a layout property change requires a new layout pass (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1945/head:pull/1945` \
`$ git checkout pull/1945`

`$ git checkout pull/1945` \
`$ git pull https://git.openjdk.org/jfx.git pull/1945/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1945`

`$ git pr show -t 1945`

</details>

<a href="https://git.openjdk.org/jfx/pull/1945.diff">https://git.openjdk.org/jfx/pull/1945.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1945#issuecomment-3437148254)
</details>


hjohn:
It looks like there is one failing test for ToolBarSkin.  Likely there is something relying on the superfluous 2nd layout.  I'm looking into it.

kevinrushforth:
Reviewers: @johanvos @kevinrushforth @arapte

We will need a unit test for this. What is the risk of regression? Are there additional tests that we could add to help detect any regressions?

/reviewers 3

hjohn:
`ToolBarSkinTest` failed due to a combination of using the StubToolkit (which doesn't run pulses) and the reuse of the `ToolBar` node in several different stages (which somehow is allowed without warning).  I've modified the test so that it still detects the original problem that it intended to fix (I reverted the fix for https://bugs.openjdk.org/browse/JDK-8364049 and checked that the rewritten test still fails without the fix).

The rewritten test works without problem with this fix, and also without this fix.

The reason why the test failed is because the size cache in `Parent` was re-used between tests. Before this PR, the test would do this:
- Creates a Stage with a reusable ToolBar
- Shows the stage (this triggers a layout, it's not because of a pulse as `StubToolKit` doesn't run those)
- Lots of layout occurs, including an attempt to start a 2nd layout pass (which `StubToolKit` won't run)
- The triggering of the 2nd layout pass however would clear the size cache in preparation for a next layout pass
- The stage is hidden, and the 2nd layout pass never runs
- A new Stage is created with the reused ToolBar, and luckily its size cache was cleared so it had to redo the calculations with the new render scale.

In the version with the fix in this PR applied, no 2nd layout pass is triggered, and thus the size cache was not cleared, and the reuse of the ToolBar node would then happily use size values belonging to the old render scale.

I also tested `ToolBar` on my own system (with a real program), and dragged the window from a monitor with 150% to 125% scale, and  the tool bar looked okay.

hjohn:
> Reviewers: @johanvos @kevinrushforth @arapte
> 
> We will need a unit test for this. 

I think I can add one that confirms that no 2nd layout pass occurs when it is not needed when detected changes are part of the current pass.

> What is the risk of regression?

I think it is certainly possible that some poorly constructed `layoutChildren` methods may inadvertently be relying on a 2nd layout pass occurring whenever there is a major UI change (when a control changes sufficiently that its parent must also change size).  If that 2nd pass however did anything more than "confirming" the first layout pass, then this likely would result in an infinite layout loop (whereas now it would stop after a single pass).  Assuming that controls that create infinite layout loops are fixed before they reach a wide audience (as you'd likely notice the drain on resources) it seems to me that it is unlikely such controls are in active use.

Note that it is really easy to get the "old" behavior; just return `false` in `Parent::inLayoutChildren` -- I've been using this to test differences.  If we're really worried this may cause problems, we could turn this into a system property so people may opt out of this fix.

The benefits of this fix could be a reduction in layout passes by half when doing things like resizing windows.  This may be quite noticeable on heavy controls like TableViews.

> Are there additional tests that we could add to help detect any regressions?

Code that may not work correctly with this fix is code where a container is unable to size and position its direct children in a stable fashion in one pass.  If the container is tracking state between layout passes, is not clearing caches at the right time, etc, then a 2nd layout pass (with all else being equal) may result in its direct children being positioned differently.  It's possible some code out there has such problems; they would currently require a 2nd (and perhaps a 3rd or 4th pass) before reaching a stable set of positions and sizes for its direct children.  Note that if they never stabilize, the code before this fix would just keep running layout passes indefinitely.  The fixed code will only do so once regardless.

As soon as anything other than the container and its direct children is modified (ie. a sibling container, a grand child, or a child of a sibling container), then this code won't block another layout pass, so any regressions will be limited to containers being unable to decide how to position its children in a single pass.  If the layout children code modifies some grand child, or some other node it happens to have a reference to, then this will still result in a new layout pass.

I could write a test that will work with the old code, and fail with the new code (simple counting layout passes, or having some state that requires multiple passes would work to "detect" this fix) but am unsure what purpose that would serve.  It's not a supported or documented use of the layout system.  Layouts that need time to settle should still have a strict opinion on the positioning of a single container's direct children, but are likely being jostled by other influences (sibling containers, grand children).  Such situations should still trigger further layout passes until everything converges (hopefully).

Note that "failure" here would just mean that it shows you the result of the first pass; FX would not crash or go into infinite loops.

hjohn:
> _Mailing list message from [Andy Goryachev](mailto:andy.goryachev@oracle.com) on [openjfx-dev](mailto:openjfx-dev@mail.openjdk.org):_
> 
> > We will need a unit test for this.
> 
> Perhaps we could even get it a bit further than that. What would you say could be a sufficiently comprehensive set of tests/scenarios to reduce the probability of regression?
> 
> Let's say, we limit the depth of the hierarchy to 3 (node-parent-parent), then the combinatorial complexity should still be manageable.
> 
> What do you think?

As much as I'd like to help out more, after analyzing this problem already for almost two full days, I don't have time to write a comprehensive test suite for one of the more complex systems in JavaFX.  I can of course extend existing tests to get behavioral coverage on this new fix.

So, I think we then should choose what we want.  We can revert https://bugs.openjdk.org/browse/JDK-8360940 and go back to a situation where the layout system can easily get into a bad state while using documented API's, or we can take a critical look at this new fix, which is fixing a long standing problem that was hidden by the bug that was fixed, and could be a major performance improvement for complex UI's.

I think manual testing with some large UI's and/or Monkey Tester should be more than sufficient to discover if this is likely to cause regressions, and I've already been doing so before I submitted this fix (I tend to just apply these fixes to my own code directly, to see how they work out).  

If we're still unsure, a system property can turn this on/off easily, at the cost of getting some redundant layout passes in the "off" mode (as it was before).

I already knew that the detection that `Node` was doing for triggering a 2nd layout pass was bogus, because `isCurrentLayoutChild` is simply not updated at the correct moments for this to work.  It works if you ask if a Node's **parent** is the current layout child of the **grand parent**, but it does not work for one level deeper: asking if a node is the current layout child of its parent.

That the logic looked correct at first glance may have have convinced the original authors that it does work, but it doesn't.  More importantly, it can't be fixed as a fix would require **each** modification of layout X/Y to wrap this change by setting the current layout child first -- that would require updating all layout containers, all 3rd party layout containers, and all 3rd party code that calls `relocate` or sets layout X/Y directly -- an impossibility IMHO.  That's also the reason I didn't fix that immediately, as at the time I didn't see a workable solution.

kevinrushforth:
@hjohn Thank you for providing the additional analysis. This will help those of us who will review this proposed fix evaluate it to see if we can poke any holes in it.

As you say, we have two choices before us: fix this bug or revert the fix for [JDK-8360940](https://bugs.openjdk.org/browse/JDK-8360940). If we can convince ourselves that this is the right fix, taking this fix would be preferable to re-introducing JDK-8360940.

As for testing, I think manual testing using various apps (e.g., Ensemble, MonkeyTester, SceneBuilder) is needed. And while we also need a new test for _this_ fix -- that is, a test that fails before and passes after the fix -- I would agree that adding a whole battery of functional tests for layout is out of scope and better done as a follow-on.

Finally, as you point out, we could provide a system property to go back to the way we did it before. If we do that (which I am not advocating at this point in the review), we might consider having that flag also revert the behavior of [JDK-8360940](https://bugs.openjdk.org/browse/JDK-8360940).

johanvos:
> would agree that adding a whole battery of functional tests for layout is out of scope and better done as a follow-on.

I'm not sure I agree with this. Personally, I consider the battery of functional tests to be more important than the fix, so without the tests, I would be very conservative.
However, we're still relatively early in the 26 release cycle, and we won't do a 26 LTS, so risks on real-world drama's are limited. We can still revert this PR (if it got merged) and the previous one later in the 26-development phase.

I'll look into a more deterministic test scenario -- the suggestion from Andy with 3 levels make sense (although reality is much more complex, with Platform.runLater() running in between 2 pulses, and *that* is non-deterministic as it depends on processor speed etc.

hjohn:
> > would agree that adding a whole battery of functional tests for layout is out of scope and better done as a follow-on.
> 
> I'm not sure I agree with this. Personally, I consider the battery of functional tests to be more important than the fix, so without the tests, I would be very conservative. However, we're still relatively early in the 26 release cycle, and we won't do a 26 LTS, so risks on real-world drama's are limited. We can still revert this PR (if it got merged) and the previous one later in the 26-development phase.
> 
> I'll look into a more deterministic test scenario -- the suggestion from Andy with 3 levels make sense (although reality is much more complex, with Platform.runLater() running in between 2 pulses, and _that_ is non-deterministic as it depends on processor speed etc.

It would actually be much easier to create deterministic tests **after** this change (unless we're dealing with a converging layout that can sometimes happen with biased containers).  In the before situation, you should check if another pulse is scheduled after the layout pass, and execute it as well to be sure that the layout is stable.  In the after situation, this will almost never be the case and the presence of a pulse request can be asserted to be false.

In any case, I'd still recommend either reverting https://bugs.openjdk.org/browse/JDK-8364049 or combining it with this fix.  The intermediate situation is still fine for most cases, but there can be nasty surprises if left in by itself (albeit only in somewhat self-contradicting layout code).

I think having both the fixes guarded behind a system property averts most risks.  I'd recommend setting it on by default though, to get real world feedback, as I think we do want to incorporate this fix permanently if it works out.  I think it definitely much closer matches the intent of the original implementation.

johanvos:
> I could write a test that will work with the old code, and fail with the new code (simple counting layout passes, or having some state that requires multiple passes would work to "detect" this fix) but am unsure what purpose that would serve.

The main purpose is testing for regression. We want to make sure a future PR does not re-introduces the old situation.

arapte:
The method `isCurrentLayoutChild()` was introduced with the fix for [JDK-8137252](https://bugs.openjdk.org/browse/JDK-8137252).
With this PR change, the issue [JDK-8137252](https://bugs.openjdk.org/browse/JDK-8137252) recurs.

hjohn:
> The method `isCurrentLayoutChild()` was introduced with the fix for [JDK-8137252](https://bugs.openjdk.org/browse/JDK-8137252). With this PR change, the issue [JDK-8137252](https://bugs.openjdk.org/browse/JDK-8137252) recurs.

Thanks, I will have a look.  If I can resolve the issue, I'll add a test case for this problem as well so we can detect a regression more easily.

hjohn:
The fix in https://bugs.openjdk.org/browse/JDK-8137252 is only "solving" the problem because now a relayout **always** occurs (until nobody modifies layout X/Y to different values).  The `isCurrentLayoutChild` check never returns `true` for a `StackPane` or any other layout container, because they don't update the current layout child before positioning said child.  The code in `Parent::layoutChildren` is a red herring and will not apply this logic magically to all other layout containers.

So, yes, https://bugs.openjdk.org/browse/JDK-8137252 fixes the problem, but it does so at the cost of an extra layout pass in all cases.  In the sample application, even if there was only a Label in that stack pane and no Ellipse, then the first time it is laid out, it will do a 2nd pass because **Label's** layout X/Y was modified...

I'm now looking further how this problem relates to `Shape` and its subtypes, as this fix is not a problem for node types based on `Parent`.

hjohn:
So I've been playing with the program in https://bugs.openjdk.org/browse/JDK-8137252 and here are some observations:

- It is binding width/height of a sibling to another (size altering) property of another sibling.  This is questionable practice, especially when both siblings are **managed** and part of a container that uses managed children (StackPane in this case).  You get away with this because `Shape`s are not resizable (by the standard width/height mechanism), and so the container can't force its preferred size onto a `Shape`.  If it was anything else (like a Region or Button) then after changing the size, relayout would run, and your change would be overridden, at least, until you set that control to unmanaged.

- Because containers do not resize non-resizable controls, but do set their positions, this is a very special half managed/half unmanaged territory.

- The whole thing works by doing a **2nd** layout pass.  This means that when you modify the width/height of the source sibling, that this will trigger a first layout pass.  During the layout pass, you modify the values of the destination sibling.  This then requires another layout pass.  Two layout passes aside, this will show up on your screen as the **layout "jumping"** (ie. it shows the destination sibling first in the wrong position, then in the 2nd pass it corrects this).  I don't see why anyone would want to do it this way, as this will make the layout jump.  In fact, if you go further, and bind more properties in these ways, you can get a 3rd layout pass, and a 4th etc.  This is not a good solution.

Currently, on each change (showing the stage, changing the source sibling's size) we see the layoutX of the destination sibling jumping:

          lx = 0.0
          Showing scene
          lx = 9.333332697550457
          lx = 17.99999936421712
          Triggering 2nd change
          lx = 17.333334604899086
          lx = 34.00000127156576
          
> What's the correct way then to draw a circle around a label that doesn't make your layout jump?

The wanted solution does not make it easy.  Because the label's width is used as radius, the desired circle actually is twice as wide as the label.  The stack pane however is unaware of this requirement as the circle is not resizable and does not have min/pref/max properties.  It has some similarities with solutions that must know the size of text (in the correct font after CSS has been applied) because other elements depend on it.

The solution is to add a listener to the label's `needsLayoutProperty`.  Every time something of consequence changes on the Label, one can now calculate how large the circle should be.  It is a bit more involved, but you get a **single** layout pass, and there is no UI jumping anymore:

```java
public class Test1 extends Application {

        launch(args);
    }

    @Override

        Label l1 = new Label("1 2");
        StackPane sp = new StackPane(l1, new EllipseWrapper(l1));

        sp.relocate(200, 100);

        Pane topPane = new Pane(sp);
        Scene scene = new Scene(topPane, 600, 400);

        sp.setStyle("-fx-border-color: RED;");

        stage.setScene(scene);
        System.out.println("Showing scene");

        stage.show();

        Thread.ofVirtual().start(() -> {
          for(;;) {
            try {
              Thread.sleep(1500);
            }
            catch(InterruptedException e) {
              break;
            }

            Platform.runLater(() -> {
              l1.setText("" + (int)(2000 * Math.random()));
            });
          }
        });
    }

    // wrapper because Ellipse has no layout properties for StackPane to use:
    class EllipseWrapper extends Region {
      private final Ellipse ellipse = new Ellipse();
      private final Label label;

      public EllipseWrapper(Label label) {
          this.label = label;
          ellipse.setOpacity(0.5);
          getChildren().add(ellipse);

          // using a change listener to ensure property is revalidated:
          label.needsLayoutProperty().subscribe(v -> { if(v) requestLayout(); });
      }

      @Override
      protected double computePrefWidth(double height) {
          return 2 * label.prefWidth(-1);  // proper calculation for layout!
      }

      @Override
      protected double computePrefHeight(double width) {
          return 2 * label.prefHeight(-1);
      }

      @Override
      protected void layoutChildren() {
          double w = getWidth();
          double h = getHeight();
          ellipse.setRadiusX(w / 2);
          ellipse.setRadiusY(h / 2);
          ellipse.setCenterX(w / 2);
          ellipse.setCenterY(h / 2);
      }
  }
}
```

hjohn:
IMHO there are now two ways forward:

1. We wish to support this rather odd program in https://bugs.openjdk.org/browse/JDK-8137252 despite it **requiring** two layout passes (with a visible UI jump).  This is possible by allowing only non-resizable containers to unconditionally trigger the relayout logic (instead of having everything trigger it because of the current disfunctional `isCurrentLayoutChild` logic):

                    if (!isResizable() || (p != null && !p.inLayoutChildren())) {

2. We don't want to support this, as IMHO, it is bad practice to bind to layout properties as this will inevitably lead to multiple layout passes and jumps (which makes JavaFX UI's appear flakey).

I've already written a couple of tests that nicely capture the advantage of this fix.  The tests clearly show that only one layout pass now occurs with this fix.

modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
```diff
@@ -1280,19 +1282,27 @@ public final void layout() {
                     break;
                 }
                 performingLayout = true;
-                layoutChildren();
+                inLayoutChildren = true;
+
+                try {
+                    layoutChildren();
+                }
+                finally {
```
andy-goryachev-oracle, line 1290: minor suggestion:
`} finally {`

andy-goryachev-oracle (commented):
First of all, I would like to thank you John for looking into the layout problems.  We've got these long standing issues that are very difficult to debug and fix.

I think this is valuable work as it definitely improves the platform, so Danke schön.

The reason I asked about tests and test scenarios is the possibility of regression.  Case in point - with this PR, on macOS with an external monitor at scale=1:

Image

I would second @johanvos in suggesting that the regression is what we should be guarding against, and perhaps expanding the tests.

hjohn:
> First of all, I would like to thank you John for looking into the layout problems. We've got these long standing issues that are very difficult to debug and fix.
> 
> I think this is valuable work as it definitely improves the platform, so Danke schön.
> 
> The reason I asked about tests and test scenarios is the possibility of regression. Case in point - with this PR, on macOS with an external monitor at scale=1:

It is starting to look like there may be more code relying on double layouts than I thought, even though most controls work absolutely fine.  I think that it is still worth pursuing eliminating the need for these (as you can temporarily see the "wrong" positions), but it may take longer as each of these will need an investigation.

I remain convinced though that the original fix in https://bugs.openjdk.org/browse/JDK-8137252 was applied too hastily, and only worked because it, unintentionally, simply always allowed double layouts.

I'll see if I can reproduce the menu issue and what is the culprit there.  I'm not on Mac though, so hoping that the problem is also present on other platforms.

johanvos:
> It is starting to look like there may be more code relying on double layouts than I thought, even though most controls work absolutely fine. I think that it is still worth pursuing eliminating the need for these (as you can temporarily see the "wrong" positions), but it may take longer as each of these will need an investigation.
> 
> I remain convinced though that the original fix in https://bugs.openjdk.org/browse/JDK-8137252 was applied too hastily, and only worked because it, unintentionally, simply always allowed double layouts.

That is very well possible, but it is what it is. I believe the main goal of the layout phase in a pulse is to make sure that "ultimately", all direct/indirect requests are handled. The secondary goal is that it should be done as efficient as possible, e.g. do not require 2 (or more) passes unless absolutely needed. 
The issue is, though, that there are a huge amount of possible scenarios, leading to situations that can not be dealt with by a single, optimized flow. I started documenting and analyzing scenarios, and even a very basic 2 node case poses issues that can go wrong.
Looking at it with our openjfx glasses, some of these scenarios are really bad. However, developers without internal knowledge often have no idea how and why it can go wrong. Adding bindings/listeners between siblings is a very common pattern, and when looking at my old code, I made tons of "mistakes" by having too much bindings that asked the layout phase to solve an almost impossible job. 
At least the layout system in JavaFX gives developers lots of freedom, and it promises to handle all edge cases. That fulfills the main goal (correct rendering, perhaps after a number of pulses, leading to flickering), but it makes the second goal (top-efficiency) really hard.

To make it harder (but very understandable), the layout phase spans a number of classes, where different concepts/choices are made (e.g. both Parent and Node have internal state that is used to determine whether to initiate a new pulse). That makes it really hard to come up with a system that would typically be used in these situations: use an algorithm that always works (although maybe less performant), and use optimizations in specific cases (e.g. no bindings in properties of children in a chontainer -> use this branch).

Something I've been thinking about every now and then is to introduce runtime warnings (ideally compiler *errors*, but that would require lots of upfront analysis), where the layout subsystem can warn that "a pretty complex situation occurred", e.g. when it is running into cyclic conditions that are not trivial to resolve without performance degradation. But that would be a major effort and conceptual change.

TLDR: I feel your pain and it can be very frustrating having to deal with non-optimal but valid code, deployed in the wild.

andy-goryachev-oracle:
> I'll see if I can reproduce the menu issue and what is the culprit there.

Thanks!  For what it's worth, the issue also appears when I move the window from the main retina screen (scale=2) to an external monitor (scale=1).  I would imagine it should be easy to reproduce on Windows, especially with a fractional scale.

andy-goryachev-oracle:
> That fulfills the main goal (correct rendering, perhaps after a number of pulses, leading to flickering), but it makes the second goal (top-efficiency) really hard.

I might be way off, but I wanted to ask you this:

How many pulses are needed to finish the layout?  If we ignore for a second some pathological cases when the layout process never ends causing continuous flicker, is there a safe upper limit?

What I am getting at is - what if we run more than one layout pass (`Scene::doLayoutPass`) per pulse?  In other words, if the layout is still dirty, we keep doing the layout until it's settled, without the associated re-rendering and flicker, and if it's still dirty after N cycles we print a warning (if said warning is enabled)?

What do you think?

hjohn:
> > That fulfills the main goal (correct rendering, perhaps after a number of pulses, leading to flickering), but it makes the second goal (top-efficiency) really hard.
> 
> I might be way off, but I wanted to ask you this:
> 
> How many pulses are needed to finish the layout? If we ignore for a second some pathological cases when the layout process never ends causing continuous flicker, is there a safe upper limit?
> 
> What I am getting at is - what if we run more than one layout pass (`Scene::doLayoutPass`) per pulse? In other words, if the layout is still dirty, we keep doing the layout until it's settled, without the associated re-rendering and flicker, and if it's still dirty after N cycles we print a warning (if said warning is enabled)?
> 
> What do you think?

It's not a bad idea to just run layout again if after running layout the root is still dirty.  That would kill the flicker, but would just reinforce bad patterns.  Also, the CPU cost is still there (ie. if you resize a Window with say a large complex tableview, then it will still have to do a whole lot of calculations twice, while the 2nd run basically changes nothing).

andy-goryachev-oracle:
> Also, the CPU cost is still there

that's true, but the cost will be there anyway - but now we are skipping the rendering and removing the flicker.  So it's a win-win, as long as the layout converges.

One example is when the layout must further change based on the current layout pass, such as when the scroll bar appears or disappears.

Also, doing these burst micro-layouts might be independent of any other work we are doing in terms of removing "bad patters" (the scroll bar scenario above is not really a bad pattern on itself, just a fact of life, I think).

johanvos:
> How many pulses are needed to finish the layout? If we ignore for a second some pathological cases when the layout process never ends causing continuous flicker, is there a safe upper limit?

Realistically, I'd say 10 iterations is pretty common, and I wouldn't be surprised if it goes much higher in some common applications. With 10 iterations, I mean the amount of "layout phases" that are chained by a `requestNextPulse` during the previous layout phase. Once there is a layout phase without a `requestNextPulse`, I consider the rendering "stable".

This will have a few other major impacts that are hard to predict:
1. what about CSS passes? repeat those too? (probably yes -> really expensive)
2. the flow of apps will be completely different. As long as a pulse is running, no Runnables scheduled via Platform.runLater() can be executed. If you do 10 layout passes inside a pulse, the time between the Runnables being executed becomes an order of magnitude more than before, and that will have a major impact for some applications.

I fear that approach is going to be even more disruptive for existing applications.

johanvos:
> It's not a bad idea to just run layout again if after running layout the root is still dirty.

There must be historical reasons why this is not the case. @kevinrushforth might have more background info. I would guess that the original design goal was to keep a pulse as short as possible, and the `requestNextPulse` inside a pulse would be useless with this approach. I think the `requestNextPulse` was created for this reason: don't stop the world for too long, render what we have, and try to render again as soon as possible (in the next pulse). This prevents the pulse runnable from lock the JavaFX Application Thread for too long -- as said above, there are other users for this Thread.

andy-goryachev-oracle:
Interesting, thanks!  I was thinking more of 2-3 cycles, actually.

kevinrushforth:
> > It's not a bad idea to just run layout again if after running layout the root is still dirty.
> 
> There must be historical reasons why this is not the case. @kevinrushforth might have more background info. I would guess that the original design goal was to keep a pulse as short as possible, and the `requestNextPulse` inside a pulse would be useless with this approach. I think the `requestNextPulse` was created for this reason: don't stop the world for too long, render what we have, and try to render again as soon as possible (in the next pulse). This prevents the pulse runnable from lock the JavaFX Application Thread for too long -- as said above, there are other users for this Thread.

From what I can remember, that was indeed the main reason.  Most of the layout stabilizes pretty quickly (1-3 passes), but as Johan points out, there can be cases where ~ 10 are needed. It is better for responsiveness to do run event handlers, app Runnables (via runLater), and allow animation to proceed after each layout + CSS pass than iterate until stable.

hjohn:
> > How many pulses are needed to finish the layout? If we ignore for a second some pathological cases when the layout process never ends causing continuous flicker, is there a safe upper limit?
> 
> Realistically, I'd say 10 iterations is pretty common, and I wouldn't be surprised if it goes much higher in some common applications. With 10 iterations, I mean the amount of "layout phases" that are chained by a `requestNextPulse` during the previous layout phase. Once there is a layout phase without a `requestNextPulse`, I consider the rendering "stable".

In what software do you get 10 iterations as "common"?  That's close to 200 ms @ 60 fps.  The software would look and perform worse than decades old software.  I'm running quite complex software with FX, and never does a layout take more than 1 pass (not counting the superfluous pass made by FX currently) as that would be absolutely unacceptable for my work that must look absolutely smooth and polished.
 
> This will have a few other major impacts that are hard to predict:
> 
> 1. what about CSS passes? repeat those too? (probably yes -> really expensive)

Ehr, no definitely not.  Just like you shouldn't be modifying layout positions during layout, you should not be doing things (during layout) that modify CSS styles because the CSS pass has already completed.  Modifying CSS during layout (which changes anything size or position related) is a guaranteed jump of your UI as a next pass is required.  This looks flakey and unprofessional.

For example, you also should not be modifying the scene graph during layout, because if you say add a new child (like a list cell or something) that cell will be rendered **without** styles, resulting in things like white flashes on your black background because that's the default background.  This kind of stuff must be done in a `Scene::addPreLayoutListener` if you want to ensure your application looks smooth.

> 2. the flow of apps will be completely different. As long as a pulse is running, no Runnables scheduled via Platform.runLater() can be executed. If you do 10 layout passes inside a pulse, the time between the Runnables being executed becomes an order of magnitude more than before, and that will have a major impact for some applications.

Nobody is suggesting running layout passes until the UI settles.

Anyway, the only practical use would be to cover up self created layout problems, and encourage more bad behavior, so perhaps it is better to not even consider this.

You mentioned that FX "promises to handle all edge cases".  Do you care to show me where it does so?  Because FX would be the first system with complex layouts that would be doing so.

hjohn:
> > Also, the CPU cost is still there
> 
> that's true, but the cost will be there anyway - but now we are skipping the rendering and removing the flicker. So it's a win-win, as long as the layout converges.

If nothing else happens in the scene, then it would also save a rendering pass yes.

> One example is when the layout must further change based on the current layout pass, such as when the scroll bar appears or disappears.

That should not be the case because of how the layout pass is split into the `compute` calls and the `layoutChildren` calls.

The scene or any layout root needs to know what size things will be, and so will use the `prefWidth` (etc) calls (that delegate to the `compute` calls) before doing any `layoutChildren` call.  These calls cascade down as deep as necessary through-out most of the scene graph (even parts not needing layout), and this would be expensive if they weren't also aggressively cached.

The `compute` calls don't need to be a `one call = one result` deal.  Often additional calls are done for more complex layouts (like for biased controls where the opposite `compute` call is done first so the related size can be passed to the wanted `compute` axis).  And this can also happen for `ScrollPane` which can look at the result of a `compute` call then decide it needs a scroll bar, redo the computation of how much space is available for the content (which may cause the other scroll bar to also be needed).  It can then position its children correctly in one pass.

See `ScrollPaneSkin#layoutChildren` code where you can see it do several checks in a row to determine the correct size when scrollbars must appear, all in one pass :)

Also of note is that it always creates and adds the scroll bars, and just keeps them invisible until needed, as creating them on demand would be too late and cause flicker (unless you do this in a prelayout listener).

johanvos:
> You mentioned that FX "promises to handle all edge cases". Do you care to show me where it does so? Because FX would be the first system with complex layouts that would be doing so.

That is a bit taken out of its context. My text said:
```
At least the layout system in JavaFX gives developers lots of freedom, and it promises to handle all edge cases. That fulfills the main goal (correct rendering, perhaps after a number of pulses, leading to flickering), but it makes the second goal (top-efficiency) really hard.
```
What I wanted to say with this is that the *current* handling gives priority to making sure things got rendered (e.g. adding a second pulse in cases where it's not needed), and not to performance. If you want to change that, all good. I simply tried to give my reading of historical context, while trying to understand why things are the way they are.
Again, I'm not saying this should not be modified (rather the contrary). I had the impression you wondered about why it was like it was, and I tried to answer that. Sorry if that message came over a bit hyperbolic.

hjohn:
> What I wanted to say with this is that the _current_ handling gives priority to making sure things got rendered (e.g. adding a second pulse in cases where it's not needed), and not to performance. If you want to change that, all good. I simply tried to give my reading of historical context, while trying to understand why things are the way they are. Again, I'm not saying this should not be modified (rather the contrary). I had the impression you wondered about why it was like it was, and I tried to answer that. Sorry if that message came over a bit hyperbolic.

Sorry, I definitely misinterpreted that and got a bit wound up about it.  

I'm however unsure how to proceed.  I don't think I can solve all the problems we've been seeing:
- Problem 1: Layout flags not being updated correctly, leading to layouts stop being updated (JDK-8360940) -- fixed by #1879 
- Problem 2: The problem Johan Vos posted in the mailing list where layout X/Y **always** gets modified twice, and although it converges to the same value each time, the layout X/Y logic will now always trigger another layout due to #1879 (before it would just get into a state where some things needed layout, but it wasn't propagated)
- Problem 3: The test program that binds on width/height properties in JDK-8137252 -- I'm dubious if that's worth supporting.  The fix was to always do a 2nd layout (as the `isCurrentLayoutChild` precondition doesn't actually work)
- Problem 4: The menu bar not working correctly (but that may be fixable)

Without any fixes, 2 + 3 + 4 works.

With just the layout flags fix 1 + 3 + 4 works.

With this PR 1 + 2 will work, and probably can get 4 fixed as well.

With this PR **and** perhaps always triggering another pass if a non-resizable control was modified, we may be able to get all of them working.  I however can't really see the logic why non-resizable controls should be an exception here, and if we can't also run into the same problems with a resizable control if I modified the example in JDK-8137252.

:-)

Maran23:
> With this PR 1 + 2 will work, and probably can get 4 fixed as well.

I would prefer this option. 
In my opinion, it is worth pursuing this option because it improves consistency, performance and simply makes sense to me.
The previous behavior felt like an oversight that just worked accidently.

I wonder if we need to combine what Johan said: Having warnings (if easily possible) for things you should rather not do.
For example, reading what you wrote above:

>  Modifying CSS during layout (which changes anything size or position related) is a guaranteed jump of your UI as a next pass is required
> For example, you also should not be modifying the scene graph during layout, because if you say add a new child (like a list cell or something) that cell will be rendered without styles
 
This is good information I even did not know in full detail (although it looks like I made this correct for my own components, because I did what JavaFX is doing). 
Having warnings or information for that would be nice. Since we now can detect correctly when we layout children, we may can detect is something triggered another CSS pass although it should better not do that now, and what to do instead. This way, even is something now breaks, we could get information why this unusual combination does not work as expected.

hjohn:
> I wonder if we need to combine what Johan said: Having warnings (if easily possible) for things you should rather not do.

I'll give this some thought as well.  The trouble is that a lot of things are allowed during layout, and how to distinguish the good from the bad. I still haven't looked at the menu problem that Andy mentioned, but if I can think of a good way to do some  warnings, it might help to figure out quickly what it might be doing wrong.

> This is good information I even did not know in full detail (although it looks like I made this correct for my own components, because I did what JavaFX is doing).
> Having warnings or information for that would be nice. Since we now can detect correctly when we layout children, we may can detect is something triggered another CSS pass although it should better not do that now, and what to do instead. This way, even is something now breaks, we could get information why this unusual combination does not work as expected.

Yeah, I discovered this one in a dark-mode application, where I was adding new cells (on demand) during layout, and I noticed a brief white flash.  Solved it with `Scene::addPreLayoutListener`.  No idea what the virtual flows do :)

Maran23:
I did tests on 3 different applications and could not spot any problems!

What I did not test:
- `ToolBar`
- `BubbleChart`
- `AreaChart`
- `Pagination`
- Most of the 3D `Shape` classes

hjohn:
> I did tests on 3 different applications and could not spot any problems!
> 
> What I did not test:
> 
> * `ToolBar`
> * `BubbleChart`
> * `AreaChart`
> * `Pagination`
> * Most of the 3D `Shape` classes

Thanks very much :)  

I'm hoping I will get around to looking at the menu bar issue this weekend, and seeing what causes the problems there.

hjohn:
> > I'll see if I can reproduce the menu issue and what is the culprit there.
> 
> Thanks! For what it's worth, the issue also appears when I move the window from the main retina screen (scale=2) to an external monitor (scale=1). I would imagine it should be easy to reproduce on Windows, especially with a fractional scale.

@andy-goryachev-oracle I ran the Monkey Tester with the changes, and also created a menu in my own test program, but I'm not seeing any issues on Windows.  I have a 3 monitor setup, 2x 150%, 1x 100%

I'll take a look at the `MenuBar` code directly and see if I can spot problems, and perhaps add some debug logs in this branch.

mstr2:
> The reason I asked about tests and test scenarios is the possibility of regression. Case in point - with this PR, on macOS with an external monitor at scale=1:
> 
> Image
> I would second @johanvos in suggesting that the regression is what we should be guarding against, and perhaps expanding the tests.

I can reproduce this issue on macOS and two displays with different DPI scale. Dragging the window to the other display has the menus appear with ellipses. Moving the mouse pointer over the menu items does not change that, but clicking on the menus causes a relayout, which fixes the issue.

hjohn:
> First of all, I would like to thank you John for looking into the layout problems. We've got these long standing issues that are very difficult to debug and fix.
> 
> I think this is valuable work as it definitely improves the platform, so Danke schön.
> 
> The reason I asked about tests and test scenarios is the possibility of regression. Case in point - with this PR, on macOS with an external monitor at scale=1:
> 
> Image
> I would second @johanvos in suggesting that the regression is what we should be guarding against, and perhaps expanding the tests.

@andy-goryachev-oracle 

I've had access to a Mac for a bit, and I attached my 4K screen to it and set it to full resolution (which makes it scale=1).

What I found is that MonkeyTester indeed shows the bug with this change.  However, I am now convinced that it is more of a coincidence, as I can reproduce the bug also without this change.  In other words, there is a bug, and this one change just happens to make it more obvious in the MonkeyTester case.

What it looks like is that the render scale is not set correctly for a short time, causing calculations (and roundings) to be done with the wrong render scale. I created a small sample with an HBox and 3 labels (as I've already narrowed it down to labels; it is not the MenuBar, it is not the MenuButtons, it is not Buttons, it's Labels).  What happens is the following timeline:

- RenderScaleX = 1.0 (default value I guess)
- RenderScaleX becomes 2.0 (perhaps that's the built-in screen render scale)
- Three labels get asked their horizontal sizes; they round these results to multiples of 0.5 ->  (19.0, 43.5, 117.5)
- These sizes are used by HBox to set its width (it becomes 231.0 pixels label widths + padding 3x 17.0)
- The layout pass finishes and the three labels are laid out (they get the horizontal sizes 19.0, 43.5, 117.5) **correct**
- RenderScaleX becomes 1.0
- The labels get asked their sizes; this time they round these to multiples of 1.0; padding was also rounded differently, it is now 18.0 instead of 17.0
- The HBox is **not** adjusted
- The labels are laid out again, but they get sizes that are slightly too small: 17.0, 43.0, 117.0 **incorrect**
  - The problem got a lot worse because we lost 3 pixels to extra padding, and some to the label content itself being rounded differently...

The conclusion here is that this is entirely expected as HBox is not being size adjusted the 2nd time, and so has to render 3 labels with slightly too little space. The result is that all 3 labels get an ellipsis.

This same thing happens for the MenuBar (it is basically a HBox with Buttons in it which are primarily labels).

So questions to ask:

- Why is render scale being adjusted twice?
- Why are layout calculations being done when render scale is not yet final?
- Why isn't all layout invalidated after a render scale change so HBox's size gets recalculated?

Things that are likely unrelated:
- The target monitor render scale is likely not relevant, it can occur with any other render scale, if there is a difference between primary screen and the target monitor; it just happens to show up much easier when going from scale=2 to scale=1 due to larger calculations differences that are not "absorbed" by snapping of the values.
- The MenuBar, Labels, Buttons, HBox; they all seem to be doing what they should be doing
  - I did notice that LabeledSkinBase is not snapping values of its text size calculations and is mixing snapped and unsnapped values (similar to https://bugs.openjdk.org/browse/JDK-8370652 ) but that's not the primary cause of the problem; I was worried that on Mac the text size calculations may have been "unstable" (until they're rendered on screen) but they return the exact same values both before and after the render scale changes.

The answers are likely to be found in how `sizeToScene` does things...

Alternatively, it could be a caching problem of widths/heights (see `Parent`) that is not being invalidated on a render scale change...

hjohn:
Update on previous reply:

I did some more investigation into how Scenes are sized.  In Window there is a huge invalidation listener when its showing state is invalidated.  It sets up the bounds and render scale of its peer, does an initial sizing of the scene, etcetera.

Now, the order of operations here is causing a slight issue.  Basically what happen is:
- Window is created but not visible
- Default location is somewhere on a scale=2 screen (even though the X value is already set on the Stage to be on a different screen)
- Window is shown, triggering the invalidation listener
- The render scale (still 2.0) is set on the peer
- For new windows, a "sizeToScene" type code is called on the Scene
  - This does layout calculations with render scale=2
- The invalidation listener then proceeds to adjust the size and position of the Window peer
  - This calls `applyBounds` which discovers the peer is on an entirely different screen, triggering a render scale change to 1.0
  - The render scale change doesn't trigger much of anything
- The invalidation listener (still the same callback) continues and calls `applyBounds` again
- It exits without doing anything further (scene is not resized again)

So to summarize:
- Scene is sized **before** it is moved to the correct location (which may be a different screen)
- Only after it is sized, it is discovered a different render scale should be in effect
- No action is taken to adjust the scene size, scene stays sized incorrectly

Quick fix I found:
- Call `applyBounds` before setting the renderscale on the peer or sizing the scene
  - The problem is now gone **AND** for an outside observer on RenderScale property it never changes from 1.0 -> 2.0 -> 1.0

andy-goryachev-oracle:
> * Call `applyBounds` before setting the renderscale on the peer or sizing the scene

fixes the issue with the menu, but creates a different failure in layout, see https://github.com/openjdk/jfx/pull/2007#issuecomment-3657112663

andy-goryachev-oracle:
To add to an earlier discussion regarding the number of iterations required for the layout to converge: there are situations where the layout goes into infinite update mode - I've seen these back in java 8 fx days, and it might happen even today.

Below is an example I captured with the monkey tester on 2025/12/03 using the current the master branch of that day:


https://github.com/user-attachments/assets/1a4cd2d5-4b71-4f25-b43b-287f22f10f82

The video could not capture the flicker exactly probably due to capture limitations, but the flicker continued until the user clicks on something or resizes the window.

hjohn:
> To add to an earlier discussion regarding the number of iterations required for the layout to converge: there are situations where the layout goes into infinite update mode - I've seen these back in java 8 fx days, and it might happen even today.
> 
> Below is an example I captured with the monkey tester on 2025/12/03 using the current the master branch of that day:
> 
>  video1256508024.mp4 
> The video could not capture the flicker exactly probably due to capture limitations, but the flicker continued until the user clicks on something or resizes the window.

It's nothing that can be resolved directly by some outside intervention, it just a bug in (what looks like) `ListView`.  Only debugging what is triggering a new layout pass each time will help find the cause.  These kind of bugs are unavoidable when controls or skins are not producing stable results from their layout calculations.  Often calculations eventually converge, but that's really only the best case scenario, and certainly not the only possible outcome.  You seem to have found a case where the calculations are not converging.

This is why IMHO any layout that relies on convergence is already broken; it just happens to broken in a way that eventually leads to a stable result...

mstr2:
> This is why IMHO any layout that relies on convergence is already broken; it just happens to broken in a way that eventually leads to a stable result...

A single top-down layout pass only works when the layout constraints are acyclic. However, JavaFX expressly supports cyclic constraints, so the layout system can't chicken out. As soon as there's a cyclic dependency (parent layout is constrained by child, child layout is constrained by parent), you've got a fixed-point problem and need to solve it in some form, for example with multiple layout phases or iteration.

Cyclic dependencies can appear in several situations:
* CSS/layout coupling: styling influences sizing, sizing influences styling
* Parent size depends on child size
* Child wraps based on parent width
* Baseline alignment
* ...

Looking at how other frameworks do this, we see that WPF and Android use two layout phases: a constraint-carrying measurement pass (given some available size constraint, how much space would each node like to have?), and a second arrange pass that authoritatively tells each node how much space it is given.

Such a system can break some cycles, especially the kind where parent and child sizes mutually constrain each other. It's not a catch-all solution; there can be higher-level cycles, for example with a scroll bar that, upon appearing, causes the layout to reflow, which then again influences the scroll bar. Android detects relayout requests during the layout phase, and will allow for one additional round of measure+layout in the same frame; if a second relayout request is encountered, it is deferred to the next frame.

In JavaFX, we don't have two phases of layout to break the easier cycles, and we don't have multiple rounds of layout. I think it is obvious that we need to do something about that, and pretending that cyclic constraints don't exist is not that.

What I believe can work for JavaFX is a layout system that basically does this for each frame:
```java
while (!layoutClean) {
    doCssPass();
    doLayoutPass();
}
```
If this doesn't settle in a very small number of iterations (something like 3), we can do the remainder of the work in subsequent frames to prevent unresponsive applications. Maybe we can even detect oscillations (e.g. scroll bar flip-flop) and suspend layout in these cases.

andy-goryachev-oracle:
> for example with multiple layout phases or iteration.

That's exactly right.

To illustrate, I can point to `VFlow:1545` and `VFlow:1559` where a change in either scroll bar visibility triggers `layoutCells()`.  It is the responsibility of the developer of such a component to ensure that the layout converges, i.e. that the logic does not cause infinite recursion.  In the `VFlow` case, it does not - once the vertical scrollbar is determined to have shown up, it cannot disappear; same goes for  the horizontal scrollbar afterwards.

Speaking of the layout system snippet, I might suggest limiting the fast loop to a few iterations only. to guard against locking up:

```
if(!layoutClean) {
  int fastSteps = 2; // or 3, no more
  for(int i=0; i<fastSteps; i++) {
    doCssPass();
    doLayoutPass();
    if(layoutClean) {
      break;
    }
  }
}
```

hjohn:
> > This is why IMHO any layout that relies on convergence is already broken; it just happens to broken in a way that eventually leads to a stable result...
> 
> A single top-down layout pass only works when the layout constraints are acyclic. However, JavaFX expressly supports cyclic constraints, so the layout system can't chicken out. As soon as there's a cyclic dependency (parent layout is constrained by child, child layout is constrained by parent), you've got a fixed-point problem and need to solve it in some form, for example with multiple layout phases or iteration.
> 
> Cyclic dependencies can appear in several situations:
> 
> * CSS/layout coupling: styling influences sizing, sizing influences styling
> * Parent size depends on child size
> * Child wraps based on parent width
> * Baseline alignment
> * ...

Are you referring to media queries for the window size when you say that sizing can influence styling?  The others can be resolved similar to how `ScrollPane` does this (see below).

> Looking at how other frameworks do this, we see that WPF and Android use two layout phases: a constraint-carrying measurement pass (given some available size constraint, how much space would each node like to have?), and a second arrange pass that authoritatively tells each node how much space it is given.

I would say that there's also two passes in FX, although less obvious.  The first pass is basically the calling of the compute methods.  These cascade down all the way down to the leaf level as each compute will call child compute methods (but can use cached results) before any layout methods are called (apart from the root one).  I'm not sure what that pass should be called, but it does sound a lot like what WPF is doing.  Only after this does the layout pass cascade down again via the layout calls.

This is why some layout problems can sometimes be solved by overriding a compute method and doing some unorthodox work there (for which the correct mechanism is to install a pre-layout listener).
 
> Such a system can break some cycles, especially the kind where parent and child sizes mutually constrain each other. It's not a catch-all solution; there can be higher-level cycles, for example with a scroll bar that, upon appearing, causes the layout to reflow, which then again influences the scroll bar.

`ScrollPane` handles that in one pass.  Basically it computes the size of the children, and if it determines there is need of a scroll bar, it does those calculations again (which can result in the appearance of the 2nd scrollbar, and the calculations are repeated one more time; this is quite visible in `ScrollPaneSkin`).  In effect, it solves this problem internally without waiting for another layout pass, as compute methods can be called as often as desired.

> Android detects relayout requests during the layout phase, and will allow for one additional round of measure+layout in the same frame; if a second relayout request is encountered, it is deferred to the next frame.
> 
> In JavaFX, we don't have two phases of layout to break the easier cycles, and we don't have multiple rounds of layout. I think it is obvious that we need to do something about that, and pretending that cyclic constraints don't exist is not that.
> 
> What I believe can work for JavaFX is a layout system that basically does this for each frame:
> 
> ```java
> while (!layoutClean) {
>     doCssPass();
>     doLayoutPass();
> }
> ```
> 
> If this doesn't settle in a very small number of iterations (something like 3), we can do the remainder of the work in subsequent frames to prevent unresponsive applications. Maybe we can even detect oscillations (e.g. scroll bar flip-flop) and suspend layout in these cases.

I think doing a loop of layout passes up to a limit is fine, although I'm not sure if CSS needs to take part in that.  One should not be changing styles during layout, that's a pre-layout concern (which is before CSS runs) and not something to do in the `compute` chain of methods nor the `layoutChildren` methods.

Of course, JavaFX being very unrestrictive, one can easily fiddle with CSS during layout calls, and create bad layouts, which hopefully converge, but at some point there is **nothing** that can be done.  As these problems tend to be very visible, they're usually fixed quickly, unlike converging layouts which escape notice.  This is why I think we should warn when certain actions are taken during layout.  A layout that converges now, may stop converging when say `modena.css` is updated or replaced.

mstr2:
> Are you referring to media queries for the window size when you say that sizing can influence styling? The others can be resolved similar to how `ScrollPane` does this (see below).

Yes, but also in a more general sense. Applications might also manually add or remove style classes depending on the size of controls.

> I would say that there's also two passes in FX, although less obvious. The first pass is basically the calling of the compute methods. These cascade down all the way down to the leaf level as each compute will call child compute methods (but can use cached results) before any layout methods are called (apart from the root one). I'm not sure what that pass should be called, but it does sound a lot like what WPF is doing. Only after this does the layout pass cascade down again via the layout calls.

That's true, but it's also less powerful than a dedicated two-pass approach. The main difference is that WPF/Android communicate down constraints to children during this first pass, asking: "given my available size, how large would you like to be?" The compute methods of JavaFX don't communicate any information from parents to their children, which is why children can't change their answer depending on the constraints imposed by their parents.

> `ScrollPane` handles that in one pass. Basically it computes the size of the children, and if it determines there is need of a scroll bar, it does those calculations again (which can result in the appearance of the 2nd scrollbar, and the calculations are repeated one more time; this is quite visible in `ScrollPaneSkin`). In effect, it solves this problem internally without waiting for another layout pass, as compute methods can be called as often as desired.

> I think doing a loop of layout passes up to a limit is fine, although I'm not sure if CSS needs to take part in that. One should not be changing styles during layout, that's a pre-layout concern (which is before CSS runs) and not something to do in the `compute` chain of methods nor the `layoutChildren` methods.

Let's assume a control where the number of children depend on its size. Maybe it arranges a set of nodes that it was provided with in some way, and fills the remaining "unused" slots in its layout with placeholder nodes. This implies that the control only knows how many placeholder nodes it needs after it was laid out. Adding those nodes after the layout pass is done means that they now don't have CSS styling. I think a case like this is already accounted for in the existing layout system, but solving this correctly in an iterative system probably means that CSS needs to be part of the loop.

hjohn:
> > Are you referring to media queries for the window size when you say that sizing can influence styling? The others can be resolved similar to how `ScrollPane` does this (see below).
> 
> Yes, but also in a more general sense. Applications might also manually add or remove style classes depending on the size of controls.

True, and that's fine, but I think we need to warn if they do that during layout.  That's just not the right time, as layouts are very restricted in what you should be doing.  Most users never write a `layoutChildren` method and most aren't crazy enough to do these things in `compute` methods.  They just need to be aware of things like pre/post layout listeners, and we need to ensure that user code is not called during layout (see below for cells...)

> That's true, but it's also less powerful than a dedicated two-pass approach. The main difference is that WPF/Android communicate down constraints to children during this first pass, asking: "given my available size, how large would you like to be?" The compute methods of JavaFX don't communicate any information from parents to their children, which is why children can't change their answer depending on the constraints imposed by their parents.

Yeah, FX has a limited variant of this with Content Bias.  It can ask "If I give you 400 pixels of width, how much height do you want?"

> > `ScrollPane` handles that in one pass. Basically it computes the size of the children, and if it determines there is need of a scroll bar, it does those calculations again (which can result in the appearance of the 2nd scrollbar, and the calculations are repeated one more time; this is quite visible in `ScrollPaneSkin`). In effect, it solves this problem internally without waiting for another layout pass, as compute methods can be called as often as desired.
> 
> > I think doing a loop of layout passes up to a limit is fine, although I'm not sure if CSS needs to take part in that. One should not be changing styles during layout, that's a pre-layout concern (which is before CSS runs) and not something to do in the `compute` chain of methods nor the `layoutChildren` methods.
> 
> Let's assume a control where the number of children depend on its size. Maybe it arranges a set of nodes that it was provided with in some way, and fills the remaining "unused" slots in its layout with placeholder nodes. This implies that the control only knows how many placeholder nodes it needs after it was laid out. Adding those nodes after the layout pass is done means that they now don't have CSS styling. I think a case like this is already accounted for in the existing layout system, but solving this correctly in an iterative system probably means that CSS needs to be part of the loop.

Although I think I see why CSS needs to be added to the loop (and more), I still think that you should refrain from adding/removing children during layout, especially if they are cells.  This still best handled in a pre-layout listener because if there is no follow-up loop, you will see CSS flicker (white background).  Even in a looped system this can then occur when loops are exhausted.

The View classes are good examples of dynamic controls that need to have more or less children depending on their size. They should be initializing these children (including calling `Cell#updateItem`) as part of a pre-layout listener.  This allows:

- Flicker free adding of children (CSS runs still, no need to rely on a next loop)
- User code doing whatever they want in `updateItem` which is not the case if that method is called during the layout phase (there are bugs currently where sometimes `updateItem` is called during layout.. very bad, as users are unaware that this is the case)

If it turns out there weren't enough (invisible) cells to fill the entire view port, then the View class can request another layout pass, which can be looped immediately (and should include calling the pre-layout listeners again).  So I think you are right here  that we should include CSS in the loop so newly added children in a 2nd loop also get the correct styles (hopefully it's cheap if not much changed).  So I think it perhaps could work like this:

1. Pulse starts
2. Pre-layout listeners called
   - View controls add/remove children, call `updateItem` based on their current size
      - `updateItem` can do whatever it wants, including modifying styles, adding/removing children, user code need not worry about layout semantics
3. CSS pass

LAYOUT STARTS -- refrain from doing "more" work besides positioning/sizing

4. Sizes get computed (by `resizeRootToPreferredSize`) down to leafs
5. Layout methods are called (according to dirty status) down to leafs

LAYOUT ENDS

6. Either during layout or in a post-layout listener, the View class discovers it did not have enough cells, it asks for a new layout
7. After post layout listeners have run, check if another layout is needed immediately, go to step 2 (unless limit exceeded)
8. Pulse ends

In this way we can prevent unnecessary layouts, and also prevent doing dangerous things during layout.

Maran23:
> 1. Pulse starts
> 2. Pre-layout listeners called
>    
>    * View controls add/remove children, call `updateItem` based on their current size
>      
>      * `updateItem` can do whatever it wants, including modifying styles, adding/removing children, user code need not worry about layout semantics
> 3. CSS pass
> 
> LAYOUT STARTS -- refrain from doing "more" work besides positioning/sizing
> 
> 4. Sizes get computed (by `resizeRootToPreferredSize`) down to leafs
> 5. Layout methods are called (according to dirty status) down to leafs
> 
> LAYOUT ENDS
> 
> 6. Either during layout or in a post-layout listener, the View class discovers it did not have enough cells, it asks for a new layout
> 7. After post layout listeners have run, check if another layout is needed immediately, go to step 2 (unless limit exceeded)
> 8. Pulse ends
> 
> In this way we can prevent unnecessary layouts, and also prevent doing dangerous things during layout.

Could we streamline this process with e.g. a new protected method (initially empty / delegating to `skinBase`) that is called (instead of setting pre-layout listeners everywhere). Something like `prelayout()`. We then can start shifting update logic there for `Controls`, together as an Umbrella Task? Probably need to discuss this first in the mailing list, but I agree this would be a nice idea. When doing, we should also document that, maybe in `docs` folder like the jdk has (was already discussed at one point in the mailing list at one point, so we may should just start).

hjohn:
> > 1. Pulse starts
> > 2. Pre-layout listeners called
> >    
> >    * View controls add/remove children, call `updateItem` based on their current size
> >      
> >      * `updateItem` can do whatever it wants, including modifying styles, adding/removing children, user code need not worry about layout semantics
> > 3. CSS pass
> > 
> > LAYOUT STARTS -- refrain from doing "more" work besides positioning/sizing
> > 
> > 4. Sizes get computed (by `resizeRootToPreferredSize`) down to leafs
> > 5. Layout methods are called (according to dirty status) down to leafs
> > 
> > LAYOUT ENDS
> > 
> > 6. Either during layout or in a post-layout listener, the View class discovers it did not have enough cells, it asks for a new layout
> > 7. After post layout listeners have run, check if another layout is needed immediately, go to step 2 (unless limit exceeded)
> > 8. Pulse ends
> > 
> > In this way we can prevent unnecessary layouts, and also prevent doing dangerous things during layout.
> 
> Could we streamline this process with e.g. a new protected method (initially empty / delegating to `skinBase`) that is called (instead of setting pre-layout listeners everywhere). Something like `prelayout()`. We then can start shifting update logic there for `Controls`, together as an Umbrella Task? 

I'm not sure having a `prelayout` method would be better. If we'd have to call it on the whole hierarchy (cascading down similar to `layout`) that could be a lot of calls.  If it would be limited to only `NEEDS_LAYOUT` nodes, then `prelayout` isn't quite equivalent (but perhaps sufficient).  I only expect the more complicated controls to need a pre-layout listener (the View classes).

Listeners are reasonably efficient (and targetted), they only suffer from one problem: removal can be slow if the amount of listeners on a single target is large (but IMHO that's a solvable problem).

modules/javafx.graphics/src/test/java/test/javafx/scene/NodeTest.java
```diff
@@ -2135,4 +2230,13 @@ public void treeVisibleFlagUpdatesChildrenTreeVisibleFlag() {
         assertTrue(NodeHelper.isTreeVisible(n1));
         assertTrue(NodeHelper.isTreeVisible(n2));
     }
+
+    private static void assertLayoutFlags(Parent leaf, LayoutFlags... flags) {
+        List<Parent> subtree = Stream.iterate(leaf, Objects::nonNull, Parent::getParent).toList().reversed();
+
+        for (int i = 0; i < subtree.size(); i++) {
+            Parent p = subtree.get(i);
+            assertEquals(flags[i], ParentShim.getLayoutFlag(p), "" + p);
```
Maran23, line 2239: Minor: Description can be improved here

Maran23 (approved):
Looks good to me. 

Did a bunch of tests with it on several application. Everything fine.
Also runs within a JPackaged Application for over 2 months now without problems.

Minor nitpick to improve the `assertEquals` message. Since we have some ideas to improve this at one point, I can imagine this test failing / need to be reworked at one point, so a helpful message would be nice!

Also the minor nitpick that Andy mentioned (try-finally).
Note that the Copyright Header need to be adjusted to 2026.

hjohn:
Not sure if there is a way forward with this one.

andy-goryachev-oracle:
Do we still have this issue:
https://github.com/openjdk/jfx/pull/1945#issuecomment-3657134694

Or did I miss a new commit since then?

hjohn:
> Do we still have this issue: [#1945 (comment)](https://github.com/openjdk/jfx/pull/1945#issuecomment-3657134694)
> 
> Or did I miss a new commit since then?

You said that was on the current master branch, or is this an issue with this PR specifically?

andy-goryachev-oracle:
ok, I'll take another look.

btw, at least one minor comment was not reacted to: https://github.com/openjdk/jfx/pull/1945/changes#r2466840973

also, whatever happened to the idea to re-layout 2 or 3 times immediately in sequence?

hjohn:
> ok, I'll take another look.

Thanks.

> btw, at least one minor comment was not reacted to: https://github.com/openjdk/jfx/pull/1945/changes#r2466840973

Yes, will do so if this is deemed acceptable at all.

> also, whatever happened to the idea to re-layout 2 or 3 times immediately in sequence?

If this doesn't introduce any problems, we could either incorporate this or do this in a separate PR. Although non-converging layouts are not preferred, it may not be always possible to avoid these, and doing perhaps 2 or 3 layouts before rendering a frame would at least make those situations a bit more bearable as less bad frames are rendered where the UI is still converging.

It would require a bit more work then this simple adjustment, as you'd probably want to have a parameter to play with (how many layouts max) and perhaps another parameter to turn on warnings for extra layout passes to aid debugging so we may eventually perhaps resolve these issues at least in FX controls (or discover there are really situations that are unresolvable with 1 pass). There is also I believe a related PR for this, so we should coordinate there.

If we're really still unsure about this, we could put it behind a parameter easily enough so at least it can be investigated more easily on various configurations.

I wouldn't have touched this if it weren't for the fact that the current check doesn't work at all; there was definitely an intention behind the check to only do an extra layout pass if needed, but it currently just always keeps doing layout passes until nothing changed at all.

andy-goryachev-oracle:
> If this doesn't introduce any problems, we could either incorporate this or do this in a separate PR.

I agree.

modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
```diff
@@ -1280,19 +1282,27 @@ public final void layout() {
...
+                    inLayoutChildren = false;
+                }
+
                 // Intended fall-through
             case DIRTY_BRANCH:
                 for (int i = 0, max = children.size(); i < max; i++) {
                     final Node child = children.get(i);
-                    currentLayoutChild = child;
+
```
andy-goryachev-oracle, line 1298: very very minor: since you are touching the file, perhaps you _could_ make the equivalent change:
```
                    if (child instanceof Parent cp) {
                        cp.layout();
                    } else if (child instanceof SubScene cs) {
                        cs.layoutPass();
                    }
```

andy-goryachev-oracle (changes requested):
So this regression has not been addressed:

Image

this happens when the MT opens on an external monitor at scale=1 (??).  move the window to retina scale=2, and it looks normal.  move the window back to external - ellipses appear again.

hjohn:
> So this regression has not been addressed:
> 
> Image
> this happens when the MT opens on an external monitor at scale=1 (??). move the window to retina scale=2, and it looks normal. move the window back to external - ellipses appear again.

Just checking, was this change included? https://github.com/openjdk/jfx/pull/2007 ?

If it was included, then this is apparently another problem being exposed by this change.

andy-goryachev-oracle:
> was this change included? #2007 ?

I merge the latest master when reviewing any PR that's been out for too long, so yes.

My concern is a high risk of regression (even if the regression is cosmetic): even if we fix this particular failure, other apps might fail in a similar fashion somewhere else.

Maran23:
> > was this change included? #2007 ?
> 
> I merge the latest master when reviewing any PR that's been out for too long, so yes.
> 
> My concern is a high risk of regression (even if the regression is cosmetic): even if we fix this particular failure, other apps might fail in a similar fashion somewhere else.

I tested this change in a number of applications, including ControlsFX and custom components and there really is no problem. So at least I'm very confident about this change, especially since the previous check tried to do the very same thing.

My guess is that `ToolBar` behaves bad for whatever reason and there is probably a bad implementation in the `Skin`.

hjohn:
> > was this change included? #2007 ?
> 
> I merge the latest master when reviewing any PR that's been out for too long, so yes.

Thanks, the problem looked similar.  I will see if it is reproducable on Windows, or if this is again a Mac specific issue.

> My concern is a high risk of regression (even if the regression is cosmetic): even if we fix this particular failure, other apps might fail in a similar fashion somewhere else.

We could introduce a property to switch it back off, so we can get this tested in a wider audience, as I do agree that it will be hard to predict the exact effects given that the old behavior which did a final unnecessary layout each time has part of FX for a long time.

andy-goryachev-oracle:
@hjohn I don't understand why you want to commit to a known regression.

Let's find out why the regression happened and then we can try with a system property, that sounds like a good idea.  For reference, there is nothing special in the monkey tester: the stock `MenuBar` is inside of a `BorderPane` which is the root of a `Scene`.

Maran23:
Is the `MenuBar` issue reproducable only on MacOS? I want to have a look myself but I never had any issue and I'm using this patch for about 3 months. But only on Windows.

I tried with up to 225% on Windows and switching the Monkey Tester between two windows, but it always works.

Maran23:
The `MenuBarSkin` basically just delegates to the `HBox`, making this even more weird. 
I wonder if the whole product here need to be snapped, not individually with `snappedLeftInset() + snappedRightInset()`.
image

But since it works completely fine on Windows, it sounds like a problem somewhere else. I remember that there were some differences when the application is dragged onto another screen - Windows made one more relayout than MacOS or something? I think @beldenfox found something out there, but can't remember where it was.

But I can only guess right now - once again need a Mac. :(

beldenfox:
> But since it works completely fine on Windows, it sounds like a problem somewhere else. I remember that there were some differences when the application is dragged onto another screen - Windows made one more relayout than MacOS or something? I think @beldenfox found something out there, but can't remember where it was.

On Mac the OS tracks the window size is in logical units and it remains the same even when it's moved to a screen with a different scaling factor. So on macOS you won't get View.notifyResize or Window.notifyResize when a window moves, you'll only get Window.notifyScaleChanged.

The Windows OS works in pixels so the dimensions of a window do change when it moves to a screen with a different scaling factor. I haven't tested this but I'm almost certain you'll get both View.notifyResize and Window.notifyResize in addition to Window.notifyScaleChanged. I would look at how ViewEvent.RESIZE and WindowEvent.RESIZE are being handled. Whatever's going on in those blocks of code will trigger on Windows but not on macOS.

Maran23:
> The Windows OS works in pixels so the dimensions of a window do change when it moves to a screen with a different scaling factor. I haven't tested this but I'm almost certain you'll get both View.notifyResize and Window.notifyResize in addition to Window.notifyScaleChanged. I would look at how ViewEvent.RESIZE and WindowEvent.RESIZE are being handled. Whatever's going on in those blocks of code will trigger on Windows but not on macOS.

Thanks. When I have access to a mac, I will try it out. Really would like to see this PR merged, as this improves the layout calls quite a bit.

PR #1958, opened 2025-11-02
Remove dependency on java.desktop from javafx.base and javafx.graphics

Maran23:
This PR makes the `java.desktop` requirement static for `javafx.base` and `javafx.graphics`.
With this changes, a JavaFX app without `Swing` and the `WebView` will be much smaller (results below).

Consequences:
- `java.desktop` need to be loaded (required) when using anything from the `javafx.beans.property.adapter` package (the property classes from `java.beans.XXX` are used here)
- `java.desktop` need to be loaded (required) when using printing, that is e.g. `PrinterJob.createPrinterJob().showPrintDialog(..)`

Results:
- Removing `java.desktop` gives a huge size boost! I benchmarked the following values when using `jlink` with: [`zip-6`, `--no-man-pages`, `--no-header-files`, `--strip-debug`, `--strip-java-debug-attributes`] to build an own runtime

Tried on a real application (here called `myapp`) with JDK-25 on Windows 11 that has dependencies to: [`javafx-base`,  `javafx-graphics`, `javafx-fxml`, `javafx-controls`]

---

`master`

`70.009.288 Bytes`

```
myapp
java.base@25
java.datatransfer@25
java.desktop@25
java.prefs@25
java.scripting@25
java.xml@25
javafx.base@26-internal
javafx.controls@26-internal
javafx.fxml@26-internal
javafx.graphics@26-internal
jdk.localedata@25
jdk.unsupported@25
```

-------------------------------------

`This PR`

`57.266.538 Bytes`

```
myapp
java.base@25
java.scripting@25
java.xml@25
javafx.base@26-internal
javafx.controls@26-internal
javafx.fxml@26-internal
javafx.graphics@26-internal
jdk.localedata@25
jdk.unsupported@25
```

---

`This PR` + https://github.com/openjdk/jfx/pull/1957

`57.249.459 Bytes`

```
myapp
java.base@25
java.scripting@25
java.xml@25
javafx.base@26-internal
javafx.controls@26-internal
javafx.fxml@26-internal
javafx.graphics@26-internal
jdk.localedata@25
```

---

So we save around 13MB. If building a native package, e.g. an `.msi` installer, this will be even higher.

This change however probably need some discussion, if we want to go that direction + add a note in the release notes, that if one of the two things mentioned above are needed, an requirement to `java.desktop` is needed (in case of a modular project).

---------
### Progress
- [ ] Change requires a CSR request matching fixVersion jfx27 to be approved (needs to be created)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1958/head:pull/1958` \
`$ git checkout pull/1958`

`$ git checkout pull/1958` \
`$ git pull https://git.openjdk.org/jfx.git pull/1958/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1958`

`$ git pr show -t 1958`

</details>

<a href="https://git.openjdk.org/jfx/pull/1958.diff">https://git.openjdk.org/jfx/pull/1958.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1958#issuecomment-3478222689)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/module-info.java
+++ b/modules/javafx.graphics/src/main/java/module-info.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -40,9 +40,10 @@
  * @since 9
  */
 module javafx.graphics {
-    requires java.desktop;
     requires java.xml;
     requires jdk.unsupported;
+
+    requires static java.desktop;
 
     requires transitive javafx.base;
 
--- a/modules/javafx.base/src/main/java/module-info.java
+++ b/modules/javafx.base/src/main/java/module-info.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2015, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -31,7 +31,7 @@
  * @since 9
  */
 module javafx.base {
-    requires java.desktop;
+    requires static java.desktop;
     requires static jdk.jfr;
 
     exports javafx.beans;


kevinrushforth:
Three quick comments before I review.

1. This will need a CSR
2. Removing the direct dependency on `java.desktop` from `javafx.base`, making it a `requires static`, seems reasonable, as I wrote in the description of the enhancement request. This will need testing.
3. Removing the direct dependency on `java.desktop` from `javafx.graphics`, making it a `requires static` will almost certainly not be accepted as long as the implementation depends on it. I'll add more details when I review.

/reviewers 2 reviewers
/csr

Maran23:
>     3. Removing the direct dependency on `java.desktop` from `javafx.graphics`, making it a `requires static` will almost certainly not be accepted as long as the implementation depends on it. I'll add more details when I review.

I agree, this change has the biggest 'risk'. Still, I would like to bring this up for discussion, as there is a huge benefit (much higher than I thought).
I have already considered discussing this on the mailing list, but I would like to know first whether this change might be out of question, as we would not need to discuss it in that case.

kevinrushforth:
> > 3. Removing the direct dependency on `java.desktop` from `javafx.graphics`, making it a `requires static` will almost certainly not be accepted as long as the implementation depends on it. I'll add more details when I review.
> 
> I agree, this change has the biggest 'risk'. Still, I would like to bring this up for discussion, as there is a huge benefit (much higher than I thought). I have already considered discussing this on the mailing list, but I would like to know first whether this change might be out of question, as we would not need to discuss it in that case.

While I might not quite go as far as to say "out of the question", it is unlikely that such a proposal would be accepted. This would be an incompatible change, which we very rarely do, especially for features that are still useful (and even then, not without notice, such as deprecating API for removal, followed by later removal). Applications that use JavaFX Printing APIs would stop working in some cases, unpredictably so, since it would work as long as any other module required java.desktop, but would fail at runtime if not (with a confusing error message). The recently-added ImageIO support added to images would also behave differently. Whereas today loading a ".tiff" file will just work (by falling back to the ImageIO TIFF loader shipped with java.desktop), it would fail to load (gracefully, but it would fail nonetheless).

The difference between the `javafx.graphics` usage and the `javafx.base` usage, is that the `javafx.graphics` usage of `java.desktop` for printing is incidental, and is an implementation detail as far as the app is concerns: There is nothing in the API docs that suggests the use of the Java2D printing API. So we would effectively have to tell application developers -- via API docs, a CSR, and a release note (at least) -- "if you want to use these APIs, your application now has to add a dependency on `java.desktop` because we no longer do". That will be a hard sell.

So, feel free to raise it on the mailing list. In the mean time, either move this to Draft or revert the change in javafx.graphics.

modules/javafx.graphics/src/main/java/module-info.java
```diff
@@ -40,10 +40,11 @@
  * @since 9
  */
 module javafx.graphics {
-    requires java.desktop;
```
kevinrushforth, line 43: As commented on in general comments, this is an incompatible change. A general discussion on the mailing list is needed to determine whether we want to consider this.

kevinrushforth (changes requested):

nlisker:
There was a discussion on this on the mailing list: https://mail.openjdk.org/pipermail/openjfx-dev/2023-November/043731.html.

Maran23:
> There was a discussion on this on the mailing list: https://mail.openjdk.org/pipermail/openjfx-dev/2023-November/043731.html.

Thanks. Is there a way to answer to it, even if I don't have the original email anymore?

nlisker:
> > There was a discussion on this on the mailing list: https://mail.openjdk.org/pipermail/openjfx-dev/2023-November/043731.html.
> 
> Thanks. Is there a way to answer to it, even if I don't have the original email anymore?

Not sure. I sent an email from that thread now so you can latch onto it.

modules/javafx.graphics/src/main/java/module-info.java
Maran23: Is there a way to 'warn' if someone does not explicitly `requires java.desktop`, so we could issue a warning when using the Print API first, and then after a while change it to `requires static`?

Maran23: I researched what possibilities we might have and to answer my own question, I don't think there is anything we can do 😕.
I guess the best 'option' we can do is to issue a warning at application startup or when printing/image loading is used with:

`WARNING: Future version X will require applications using the Print-API to declare the module java.desktop explicitly.`

Which is probably not 'helpful enough' as we eventually still break existing applications?

koppor:
We are working on an open source desktop application used by more than 10.000 people world wide and want to shrink down our distribution size. We are trying hard to replace "WebView" by a [HTML-to-RichTextArea renderer](https://github.com/JabRef/html-to-node) and aim to [reduce image size](https://github.com/JabRef/jabref/issues/16217).

As a mid-term vision, we would also work on removing `java.desktop` on our side.

Therefore, this PR is highly welcome! 🤩

---

Links:

- [RichTextArea](https://openjfx.io/javadoc/26/jfx.incubator.richtext/jfx/incubator/scene/control/richtext/RichTextArea.html)

PR #1969, opened 2025-11-10
8390915: Provide DrawingContext API similar to GraphicsContext API for WritableImage

hjohn:
This PR adds a `getDrawingContext` method to `WritableImage`, which works similar to `Canvas::getGraphicsContext` and shares the same signatures. Key features include:

- **Shape rendering**: `strokeRect`, `fillRect`, `strokeOval`, `fillOval`, `strokeArc`, `fillArc`, `strokePolyline`, `fillPolygon`, etc.
- **Stroke and fill attributes**: `lineWidth`, `lineCap`, `lineJoin`, `miterLimit`, `fillRule`, `stroke` and `fill` paints.
- **Global graphics settings**: `globalAlpha` and `globalBlendMode`.
- **Image drawing**: draw other `Image` instances with scaling and source/destination rectangles.
- **Text rendering**: render text
- **Save/Restore**: store current stroke, font, dashes, etc and restore them later
- **Paths**: begin a path, with lines, curves, etc, then stroke or fill it
- **Clips**: support rectangular clips in the SW renderer

This feature enables direct software rendering to `WritableImage` without requiring a `Canvas` + snapshot.

**Additional notes**:

- The implementation leverages the software stack consisting of the Marlin rasterizer and Pisces compositor/painter.

**Example usage**:

```java
WritableImage img = new WritableImage(400, 400);
DrawingContext ctx = img.getDrawingContext();
ctx.setFill(Color.RED);
ctx.fillRect(50, 50, 100, 100);
```

See the sample program `RandomShapesDemo` to see a `WritableImage` and `Canvas` side by side performing the same operations:

image

Newer version:

image

---------

---------
### Progress

### Issue
 * [JDK-8390915](https://bugs.openjdk.org/browse/JDK-8390915): Provide DrawingContext API similar to GraphicsContext API for WritableImage (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1969/head:pull/1969` \
`$ git checkout pull/1969`

`$ git checkout pull/1969` \
`$ git pull https://git.openjdk.org/jfx.git pull/1969/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1969`

`$ git pr show -t 1969`

</details>

<a href="https://git.openjdk.org/jfx/pull/1969.diff">https://git.openjdk.org/jfx/pull/1969.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1969#issuecomment-5462862109)
</details>


modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
```diff
@@ -1506,8 +1506,8 @@ public void factoryReleased() {
 
     @Override
     public PlatformImage createPlatformImage(int w, int h) {
-        ByteBuffer bytebuf = ByteBuffer.allocate(w * h * 4);
-        return com.sun.prism.Image.fromByteBgraPreData(bytebuf, w, h);
+        IntBuffer buf = IntBuffer.allocate(w * h);
+        return com.sun.prism.Image.fromIntArgbPreData(buf, w, h);
```
hjohn, line 1507: This seems to be no problem to change (an image heavy application still runs absolutely fine), but it is kind of a global change.  If there are issues with this, we could make a specific method for writable images to use so we always get an `int[]` buffer that the software renderer expects.

As it is now, the renderer is writing directly into the underlying image storage, without any copies being made (which is nice and efficient).

andy-goryachev-oracle:
This is interesting, similar to `BufferedImage.createGraphics()` in AWT.
 
Questions:
- so this will **aways** be slower than Canvas?
- are the results going to be exactly the same, or there will be platform-specific differences in anti-aliasing etc?

hjohn:
> This is interesting, similar to `BufferedImage.createGraphics()` in AWT.
> 
> Questions:
> 
> * so this will **aways** be slower than Canvas?

See my mailinglist answer, but in short, there are cases where `WritableImage` (which is not a Node) can be faster, especially when you want to access the results with say a `PixelReader`.

> * are the results going to be exactly the same, or there will be platform-specific differences in anti-aliasing etc?

I think I missed a nuance here in my mailinglist reply.  I think the software renderer will always give the same results, regardless of platform (interesting for certain use cases I suppose, and perhaps for testing as well).  It will almost certainly differ however from GPU renderings of the same operations, but for supported operations it may be hard to distinguish what was GPU and what was CPU rendered.

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
andy-goryachev-oracle: I am curious why the byte buffer was chosen initially.
Will `IntBuffer` be better on every platform?

Also, `w * h` might be negative if the product is greater than ~2B, though it will result in an `IllegalArgumentException` with a cryptic "capacity expected to be negative" message.

andy-goryachev-oracle:
> See my mailinglist answer

https://mail.openjdk.org/pipermail/openjfx-dev/2025-November/057463.html

(for reference, it's a good answer)

hjohn:
Adding a note here for myself:

```java
    private final AbstractNotifyListener platformImageChangeListener =
            new AbstractNotifyListener() {
        @Override
        public void invalidated(Observable valueModel) {
            invalidateWidthHeight();
            NodeHelper.markDirty(ImageView.this, DirtyBits.NODE_CONTENTS);
            NodeHelper.geomChanged(ImageView.this);
        }
    };
```

The above code in `ImageView` responds to **pixel** changes in the platform image, however, such a change can never influence the width/height (it is triggered by `pixelsDirty` in `WritableImage`).  Because it also invalidates width/height, and triggers a `geomChanged` a new layout is triggered.  If you happen to be drawing something in your writable image as part of `layoutChildren` via `getDrawingContext` or the pixel writer, this may then trigger infinite layouts.

I get the impression that the code is unaware that even though it is called "platform image **CHANGE** listener" there is no actual image reference change; only some pixels were dirty in the same image...

I think the `invalidateWidthHeight` and `NodeHelper.geomChanged` lines need to be removed as part of this PR.

andy-goryachev-oracle:
> I think the `invalidateWidthHeight` and `NodeHelper.geomChanged` lines need to be removed as part of this PR.

This explains why my Canvas-based table cell implementation resulted in a continuous layout.

This looks like a bug to me, and I think we should fix it separately.

andy-goryachev-oracle:
Also, perhaps we need to add a condition to suppress `invalidateWidthHeight` and `NodeHelper.geomChanged` only when it's the same image - we do need to handle the actual image change.

andy-goryachev-oracle:
Created https://bugs.openjdk.org/browse/JDK-8372007 to investigate the re-layout case.

hjohn:
> > I think the `invalidateWidthHeight` and `NodeHelper.geomChanged` lines need to be removed as part of this PR.
> 
> This explains why my Canvas-based table cell implementation resulted in a continuous layout.

Maybe, but I doubt it, I don't think `Canvas` interacts with `Image` / `WritableImage`.  You may have hit on a similar bug.

> This looks like a bug to me, and I think we should fix it separately.

Yeah, this is a bug no doubt, and can be fixed separately also.

> Also, perhaps we need to add a condition to suppress invalidateWidthHeight and NodeHelper.geomChanged only when it's the same image - we do need to handle the actual image change.

There should not be an image change, this is some internal platform image, and it changes, but not in the way we normally think of it for Change listeners (the reference does not change, but the contents do).  It seems that in `Image` this was added:

    void pixelsDirty() {  // called by WritableImage / PixelWriter to signal pixels were modified
        platformImagePropertyImpl().fireValueChangedEvent();
    }

Basically it just fires a change event (directly) but without there actually being a change in the value of that property (it's a bit dirty, but luckily internal code).  So I'm pretty sure there is never a geometry change via this route (I mean, `WritableImage` is basically constructed with a fixed size, so how could there be?)

Maran23:
Just realized this is closed. I still think this is a nice enhancement.

Unfortunately I can't test on any of my existing applications, as I never really needed `Canvas` and custom drawing. So I hope other can weight in if that solves their usecase. Code wise, this is pretty straight forward as the change is very isolated.

hjohn:
I've not given up on it, just lack a bit of time to work on it (and my use case has changed enough that this is not an immediate need for me currently). It should still be quite useful for cases where you must read back the result of drawing operations (for say undo/redo or drawing a custom cursor) as having to make a snapshot of the entire area could be much more expensive.

It can also be useful for reproducible results, as the CPU renderer will work the same in all cases, and could be used to synthesize images/icons instead of storing those as an image. It could be also be useful for say an SVG implementation that works directly with WritableImage (instead of via BufferedImage/AWT).

/open

hjohn:
Discovered another reason why I think we should integrate this.

A large `Canvas` (say 4096x4096 pixels) gets scaled up to 8192x8192 pixels when there is a display attached to your system that has more than 1.0 renderscale. That represents a 256 MB texture that is being allocated that lives on the GPU.

In contrast, a WritableImage does no such thing. It remains at the indicated render scale, using only about 64 MB of VRAM. Also ImageView + WritableImage uses texture tiling, which means it can exceed the max allowed texture size (ie. you can make a 64k x 1k writable image, something that Canvas can't do at all).

andy-goryachev-oracle:
The size argument is a good one.

One last question: would it make more sense to update the `Canvas` instead to remove the size limitation, instead of creating some parallel way of doing the same thing, possibly introducing subtle and not so subtle differences?

Another aspect: let's say we decide to go with two parallel APIs - should we then make sure they mirror each other exactly?  If so, would it create another maintenance burden?

hjohn:
> The size argument is a good one.
> 
> One last question: would it make more sense to update the `Canvas` instead to remove the size limitation, instead of creating some parallel way of doing the same thing, possibly introducing subtle and not so subtle differences?

I thought about that, but this is a lot harder than it is to do this for writable image. Writable image does tiling only **after** drawing operations have completed (it splits the normal RAM buffer into textures that overlap by 1 pixel, all pre-existing code).

In contrast, Canvas let's the GPU do the drawing, and drawing across multiple tiles is a lot harder than drawing in a single buffer, then tiling later. For example, a thick brush line, or some effect like blur, is really hard to get right when it crosses textures. The software renderer doesn't have this problem as it still has a full normal RAM buffer.

So, although we may be a bit smarter about what Canvas does, it doesn't look like a trivial quick win. I think it is better to see `Canvas` as a gateway to a single GPU texture, with all limitations that apply there (size restrictions).  The automatic scale up is an unfortunate implementation decision IMHO that perhaps we can have an option for to disable, so the user can handle this themselves (with `setScaleX/Y` and a `Group` if need be, so they get an exact 1:1 pixel backed texture).

Also, I may do a short investigation to see if we can have `Canvas` gracefully degrade (and perhaps `ImageView` as well) when the allocation of a texture fails due to size restrictions or insufficient VRAM.  Currently, it starts throwing NPE's and the UI is borked.  I mean it would be infinitely better to just show an empty square or something, instead of having the whole layout break.

> Another aspect: let's say we decide to go with two parallel APIs - should we then make sure they mirror each other exactly? If so, would it create another maintenance burden?

So far I've mirrored the API of `GraphicsContext` in the `DrawingContext` interface. My intention is to have `GraphicsContext` implement `DrawingContext`, and document it similar to how the collections API is documented with methods throwing `UnsupportedOperationException` when a specific drawing context can't handle a specific operation.

So for example, the `DrawingContext` would document the `setGlobalBlendMode` like this:

```
  /**
   * Sets the global blend mode.
   * ...
   * @param op the {@code BlendMode} that will be set or null.
   * @throws UnsupportedOperationException if the given blend mode is not supported by this implementation
   */
  void setGlobalBlendMode(BlendMode op);
```

The software renderer will then throw `UnsupportedOperationException` for blend modes it doesn't support.

The main differences between the GPU renderer and the software renderer are a tiny list:
- Software renderer doesn't support more than 1 blend mode
- Software renderer can't do arbitrary clips (ie. a path clip)
- Software renderer can't do transform rotations/shears + a rectangular clip at the same time

For clipping therefore I've documented it like this:

```
  /**
   * Sets the current transform. Only 2D transforms are supported. The only
   * ...
   * @param xform The affine to be copied and used as the current transform.
   * @throws UnsupportedOperationException if a transform containing rotation
   *         or shear is set while a clip is active and the implementation does
   *         not support such clips
   */
  void setTransform(Affine xform);
```

In the end, it would mean a `DrawingContext` is a full fledged API, implemented for both the GPU and SW pipelines, with differences that are minor enough that most users will be able to switch between the two without any adjustments. Only when using arbitrary clips and/or special blend modes do they need to make adjustments.

hjohn:
> If so, would it create another maintenance burden?

If we can move most of the docs to `DrawingContext` from `GraphicsContext` then the `GraphicsContext` can drop most of its docs (and just have `@Override`) with a few tiny pieces where it mentions its capabilities.

So I don't think so, and we'd gain a very nice new API that probably has uses we haven't thought of yet :)

I was thinking you could create a memory backed writable image that is say 100kx1k in size, and have the automatic tiling "scroll in" parts that are invisible, and you could make a classic side scroller game with it... using the software drawing operations to fill it with the world you want :)

hjohn:
So I've completed this now, as far as I think is reasonable. `WritableImage`'s `DrawingContext` has now got almost full parity with `GraphicsContext` from `Canvas`, including fonts, dashed lines, paths, clips (rectangular only), transforms and save/restore functionality.

I added a new demo image showing they render almost pixel perfect the same (this is not too surprising as the software renderer is the core of the JavaFX software pipeline).

Please review :)

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
hjohn: I don't see a specific reason for the choice -- possibly because Java2D did it that way. All pipelines can deal with either format without any issues that I could see, and should do so without performance penalties:

- D3D -- `INT_ARGB_PRE` is actually native for D3D
- Metal -- directly supports `INT_ARGB_PRE` no conversion needed
- ES2 (OpenGL) -- format can be specified, and both `INT_ARGB_PRE` and `BYTE_BGRA_PRE` uses the same path, no CPU conversion needed
- Software -- uses `INT_ARGB_PRE` natively, so bonus there

From a user perspective, you can observe this change via `getPixelFormat` -- but as we offer no direct access to pixels, there is no need to know. The user can also request pixels to be copied, but they have to specify their preferred destination format so this will always end up the same regardless.

Overall, this has no downsides for existing code and pipeline, but a huge upside for the software renderer.

> Also, w * h might be negative if the product is greater than ~2B, though it will result in an IllegalArgumentException with a cryptic "capacity expected to be negative" message.

That's pre-existing, and with this change it will take a 4x larger image before that happens.

modules/javafx.graphics/src/main/java/javafx/scene/canvas/GraphicsContext.java
```diff
@@ -1912,225 +1403,22 @@ public void bezierCurveTo(double xc1, double yc1, double xc2, double yc2, double
...
-            coords[0] = path.getCurrentX();
-            coords[1] = path.getCurrentY();
+        else {
             try {
-                curState.transform.inverseTransform(coords, 0, coords, 0, 1);
-            } catch (NoninvertibleTransformException e) {
-                return false;
+                path.arcTo(curState.transform, (float) x1, (float) y1, (float) x2, (float) y2, (float) radius);
+            }
+            catch (IllegalPathStateException | NoninvertibleTransformException e) {
```
andy-goryachev-oracle, line 1417: aren't these braces incompatible with the coding style?

modules/javafx.graphics/src/main/java/javafx/scene/canvas/GraphicsContext.java
```diff
@@ -2339,143 +1498,58 @@ public void clip() {
...
 
-    /**
-     * Returns true if the the given x,y point is inside the path.
-     *
-     * @param x the X coordinate to use for the check.
-     * @param y the Y coordinate to use for the check.
-     * @return true if the point given is inside the path, false
-     * otherwise.
-     */
+    @Override
```
andy-goryachev-oracle: is this a new method in the `GraphicsContext`?  if so, it needs `@since`.

modules/javafx.graphics/src/main/java/javafx/scene/image/WritableImage.java
```diff
@@ -157,6 +160,28 @@ public WritableImage(@NamedArg("reader") PixelReader reader,
...
     }
 
+    private WeakReference<SWDrawingContext> drawingContextRef;
+
+    /**
+     * Returns the {@link DrawingContext} associated with this image.
+     *
+     * @return the {@link DrawingContext} associated with this image, never {@code null}
+     */
+    public DrawingContext getDrawingContext() {
```
andy-goryachev-oracle: should this method be `final`?

modules/javafx.graphics/src/main/java/javafx/scene/image/WritableImage.java
```diff
@@ -157,6 +160,28 @@ public WritableImage(@NamedArg("reader") PixelReader reader,
...
+        SWDrawingContext context = drawingContextRef == null ? null : drawingContextRef.get();
+
+        if (context == null) {
+            if (!(getWritablePlatformImage() instanceof com.sun.prism.Image img)) {
+                throw new IllegalStateException("platformImage must be a prism image");
```
andy-goryachev-oracle: 1. this exception must be documented
2. what is a "prism image" from the user perspective?  could the message be rephrased to give the user a better understanding?

modules/javafx.graphics/src/test/java/test/com/sun/javafx/geom/Path2DTest.java
```diff
@@ -206,17 +206,6 @@ void testArcTo() {
                 }
             }
         }
-        RectBounds rectBounds = new RectBounds(10, 20, 20, 30);
```
andy-goryachev-oracle, line 209: seems unrelated, but I would rather clean up this unused code.

modules/javafx.graphics/src/test/java/test/com/sun/prism/sw/SWDrawingContextTest.java
```diff
@@ -0,0 +1,1445 @@
...
+    private static final Field SCREENS_FIELD;
+
+    static {
+        try {
+            SCREENS_FIELD = Screen.class.getDeclaredField("screens");
+
+            SCREENS_FIELD.setAccessible(true);
+        }
+        catch (Exception e) {
+            throw new IllegalStateException(e);
```
andy-goryachev-oracle, line 92: illegal state?  is this the right exception for the occasion?

andy-goryachev-oracle (commented):
- The implementation leverages the software Pisces renderer.
- This lays the groundwork for future support of text rendering and path operations.

aren't these already implemented in this version?  or something is still missing?

modules/javafx.graphics/src/main/java/javafx/scene/canvas/GraphicsContext.java
hjohn: Ah, yes, it is; as the SW rendered can't do arbitrary clips, but rectangular clips are still useful, I added this method as a compromise.

modules/javafx.graphics/src/test/java/test/com/sun/javafx/geom/Path2DTest.java
hjohn: you're right, it just annoys me when files have warnings, as I can't see as easily if the files I'm committing are warning free then :)

modules/javafx.graphics/src/test/java/test/com/sun/prism/sw/SWDrawingContextTest.java
hjohn: It hardly matters, this is test code. I use `IllegalStateException` for things that can't happen (which is the case here).

hjohn:
> * The implementation leverages the software Pisces renderer.
> * This lays the groundwork for future support of text rendering and path operations.
> 
> aren't these already implemented in this version? or something is still missing?

Right, I'll remove that -- that was before I finished the PoC.

bourgesl:
> > * The implementation leverages the software Pisces renderer.
> > * This lays the groundwork for future support of text rendering and path operations.
> > 
> > aren't these already implemented in this version? or something is still missing?
> 
> Right, I'll remove that -- that was before I finished the PoC.

Since 2016 the Marlin renderer replaced the Pisces renderer to improve quality & performance... (software ie cpu-bound rendering).

Please clarify if needed: is it using only software rendering (including blending) or it uses the 'usual' pipeline ie marlin+gpu blending ?

hjohn:
> > > * The implementation leverages the software Pisces renderer.
> > > * This lays the groundwork for future support of text rendering and path operations.
> > > 
> > > aren't these already implemented in this version? or something is still missing?
> > 
> > 
> > Right, I'll remove that -- that was before I finished the PoC.
> 
> Since 2016 the Marlin renderer replaced the Pisces renderer to improve quality & performance... (software ie cpu-bound rendering).

@bourgesl Marlin replaced the Pisces rasterizer, but didn't fully replace the whole stack (so parts of Pisces were removed, but not the compositor/painter).

So the stack used by `WritableImage` is Marlin + Pisces (sw_prism.dll), which is a full software stack, regardless of what FX itself uses or what you configured (FX will still be fully accelerated).

I should have stated this more accurately.

> Please clarify if needed: is it using only software rendering (including blending) or it uses the 'usual' pipeline ie marlin+gpu blending ?

For `WritableImage` it will always use Marlin + Pisces painter (so full CPU stack), while other parts of FX can (simultaneously) use the Marlin + D3D/ES2/Metal (GPU) stack.

So the demo that is included in this PR is showing you Marlin+CPU (`WritableImage`) and Marlin+GPU (`Canvas`) side by side

bourgesl:
Good answer, that's obvious now, blending is software too.

modules/javafx.graphics/src/main/java/javafx/scene/canvas/GraphicsContext.java
```diff
@@ -1912,225 +1403,22 @@ public void bezierCurveTo(double xc1, double yc1, double xc2, double yc2, double
...
-            y0 = (float) (path.getCurrentY() - curState.transform.getMyt());
-        } else {
-            coords[0] = path.getCurrentX();
-            coords[1] = path.getCurrentY();
+        else {
             try {
-                curState.transform.inverseTransform(coords, 0, coords, 0, 1);
-            } catch (NoninvertibleTransformException e) {
-                return false;
+                path.arcTo(curState.transform, (float) x1, (float) y1, (float) x2, (float) y2, (float) radius);
```
andy-goryachev-oracle, line 1414: does it need `markPathDirty()` here?

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
```diff
@@ -0,0 +1,1296 @@
...
+     *
+     * @param img a prism image, cannot be {@code null}
+     * @param pixelsDirty a consumer of dirty rectangles, cannot be {@code null}
+     * @throws NullPointerException if any argument is {@code null}
+     * @throws IllegalStateException if the image is not backed by a writable pixel buffer
+     *     in the correct format.
+     */
+    public SWDrawingContext(com.sun.prism.Image img, Consumer<Rectangle> pixelsDirty) {
+        int[] data = switch (img.getPixelBuffer()) {
+            case IntBuffer ib -> ib.array();
```
andy-goryachev-oracle: is it possible to get a sliced array at this point?
IntBuffer::array() ignores the arrayOffset so the subsequent operations might overwrite wrong elements.

modules/javafx.graphics/src/main/java/javafx/scene/image/WritableImage.java
```diff
@@ -157,6 +160,28 @@ public WritableImage(@NamedArg("reader") PixelReader reader,
         getPixelWriter().setPixels(0, 0, width, height, reader, x, y);
     }
 
+    private WeakReference<SWDrawingContext> drawingContextRef;
```
andy-goryachev-oracle: is it right to use the weak reference here?
what happens if the gc collects the current context, will the next call to getDrawingContext() return one with default paint, transform, etc.?

modules/javafx.graphics/src/main/java/javafx/scene/image/WritableImage.java
```diff
@@ -157,6 +160,28 @@ public WritableImage(@NamedArg("reader") PixelReader reader,
...
+     */
+    public DrawingContext getDrawingContext() {
+        SWDrawingContext context = drawingContextRef == null ? null : drawingContextRef.get();
+
+        if (context == null) {
+            if (!(getWritablePlatformImage() instanceof com.sun.prism.Image img)) {
+                throw new IllegalStateException("platformImage must be a prism image");
+            }
+
+            context = new SWDrawingContext(img, rect -> bufferDirty(rect));
```
andy-goryachev-oracle: a single `PixelBuffer` may source multiple `WriteableImage`s, this callback invalidates only that owned by the context, leaving the other siblings showing stale pixels.

modules/javafx.graphics/src/main/java/javafx/scene/canvas/GraphicsContext.java
```diff
@@ -2339,143 +1498,58 @@ public void clip() {
...
-     * Returns true if the the given x,y point is inside the path.
-     *
-     * @param x the X coordinate to use for the check.
-     * @param y the Y coordinate to use for the check.
-     * @return true if the point given is inside the path, false
-     * otherwise.
-     */
+    @Override
+    public void clipRect(double x, double y, double w, double h) {
+        beginPath();
```
andy-goryachev-oracle: will this clobber the current path?

scenario: build a path, call clipRect(), call fill() -> wrong path will be filled.

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
```diff
@@ -0,0 +1,1296 @@
...
+    @Override
+    public void clipRect(double x, double y, double w, double h) {
+        if (transform.getMxy() != 0 || transform.getMyx() != 0) {
+            throw new UnsupportedOperationException("clipping is not supported under a transform with rotation or shear");
+        }
+
+        double x1 = Math.floor(Math.min(x, x + w));
+        double y1 = Math.floor(Math.min(y, y + h));
+        double x2 = Math.ceil(Math.max(x, x + w));
+        double y2 = Math.ceil(Math.max(y, y + h));
```
andy-goryachev-oracle: the clip coordinates are rounded before transformation (L565), possibly repeatedly.  later transformation will change the existing clip.  should it be done at the last possible moment?

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
```diff
@@ -0,0 +1,1296 @@
...
+
+    @Override
+    public void stroke() {
+        if (path.getNumCommands() == 0) {
+            return;
+        }
+
+        applyStrokeParameters();
+
+        graphics.setTransform(BaseTransform.IDENTITY_TRANSFORM);
```
andy-goryachev-oracle, line 1010: will applying the `IDENTITY_TRANSFORM` prevent scaling of stroke attributes like line width and dash lengths?

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
```diff
@@ -0,0 +1,1296 @@
...
+                continue;
+            }
+
+            Point2D pt = run.getLocation();
+            RectBounds lineBounds = run.getLineBounds();
+            float runX = pt.x + layoutX;
+            float runY = pt.y + layoutY - lineBounds.getMinY();
+
+            if (stroke) {
+                graphics.draw(strike.getOutline(run, BaseTransform.getTranslateInstance(runX, runY)));
```
andy-goryachev-oracle, line 1176: a wide stroke will go outside the min/max dirty bounds used later in L1167, possibly corrupting the pixels.

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
```diff
@@ -0,0 +1,1296 @@
...
+    }
+
+    @Override
+    public double getGlobalAlpha() {
+        return globalAlpha;
+    }
+
+    @Override
+    public void setGlobalAlpha(double alpha) {
+        this.globalAlpha = Math.clamp(alpha, 0.0, 1.0);
```
andy-goryachev-oracle: the clamping here violates the contract: "any valid double [value] can be set".
the clamping should be done upon use.

modules/javafx.graphics/src/main/java/javafx/scene/canvas/GraphicsContext.java
```diff
@@ -1912,225 +1403,22 @@ public void bezierCurveTo(double xc1, double yc1, double xc2, double yc2, double
...
-     * @param x1 the X coordinate of the first point of the arc.
-     * @param y1 the Y coordinate of the first point of the arc.
-     * @param x2 the X coordinate of the second point of the arc.
-     * @param y2 the Y coordinate of the second point of the arc.
-     * @param radius the radius of the arc in the range {0.0-positive infinity}.
-     */
+    @Override
     public void arcTo(double x1, double y1, double x2, double y2, double radius) {
         if (path.getNumCommands() == 0) {
             moveTo(x1, y1);
```
andy-goryachev-oracle, line 1409: this needs lineTo(): this code will not render the expected dot using a round cap.

andy-goryachev-oracle (changes requested):
quite a few things.  I would suggest adding unit tests for all the discovered problematic scenarios, if possible.

modules/javafx.graphics/src/test/java/test/com/sun/prism/sw/SWDrawingContextTest.java
andy-goryachev-oracle: good point, but maybe `throw new AssertionError(e);` instead (as in BorderTest:588)

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
hjohn: I've adjusted this to allow for both slices and direct buffers.

modules/javafx.graphics/src/main/java/javafx/scene/image/WritableImage.java
hjohn: Yeah, this was ill conceived -- it's been a while since I wrote that, and I think I wanted to ensure that we're not holding on to too much resources after you're no longer using the drawing context.

Reevaluating this though, I think the resources used pale in comparison to a the likely size of the underlying pixel data of a `WritableImage`. Still good to do it lazy though when you use images that you manipulate directly.

modules/javafx.graphics/src/main/java/javafx/scene/canvas/GraphicsContext.java
hjohn: Yes, it would have. Fixed now.

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
hjohn: I've changed this to immediately transform the clip to device space. This has the benefit that there is no repeated rounding anymore, and also that one can now set a clip, and still do a rotation transform after setting the clip (the other way around is still not allowed).

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
hjohn: Yes, it would have. Everything is already in (transformed) device space, with the exception of the stroke width and dash lengths.  I've adjusted this now to transform these if scale != 1.0

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
hjohn: Fixed

modules/javafx.graphics/src/main/java/javafx/scene/image/WritableImage.java
hjohn: This exception can't happen, we only have prism images (`PlatformImage` is an interface with 1 implementation) -- so I've removed it and just did a hard cast.

modules/javafx.graphics/src/main/java/javafx/scene/image/WritableImage.java
hjohn: Yes, it is the safer choice

modules/javafx.graphics/src/main/java/javafx/scene/canvas/GraphicsContext.java
hjohn: Added since tag

modules/javafx.graphics/src/main/native-prism-sw/JJavaSurface.c
```diff
@@ -149,7 +150,17 @@ surface_acquire(AbstractSurface* surface, JNIEnv* env, jobject surfaceHandle) {
...
     }
+
+    /*
+     * The pixel data of this surface may start at a non-zero offset into the backing array (for example when
+     * the surface wraps a sliced IntBuffer); shift the base pointer so that all reads and writes are relative
+     * to the surface origin.
+     */
+
+    surface->super.data = (void *)((jint *)surface->super.data +
+        (*env)->GetIntField(env, surfaceHandle, fieldIds[SURFACE_DATA_OFFSET]));
```
mstr2: This might violate the contract of [GetPrimitiveArrayCritical](https://docs.oracle.com/en/java/javase/26/docs/specs/jni/functions.html#getprimitivearraycritical-releaseprimitivearraycritical) in two ways:
1. "Inside a critical region, native code must not call other JNI functions", but you're calling `GetIntField`.
2. The pointer passed to `ReleasePrimitiveArrayCritical` must be the exact pointer as returned by `GetPrimitiveArrayCritical`, you can't do pointer arithmetic with it.

A solution could be to read the data offset before entering the critical region, retain the original base pointer separately, use base + offset only as the renderer pointer, and then pass the base pointer to `ReleasePrimitiveArrayCritical`.

modules/javafx.graphics/src/main/java/javafx/scene/image/WritableImage.java
```diff
@@ -157,6 +160,26 @@ public WritableImage(@NamedArg("reader") PixelReader reader,
...
+     * created from a {@link PixelBuffer} must use a pixel format of that type.
+     *
+     * @return the {@link DrawingContext} associated with this image, never {@code null}
+     * @throws IllegalStateException if the pixel storage of this image is not in {@code INT_ARGB_PRE}
+     *     format (for example, when created from a {@code BYTE_BGRA_PRE} {@code PixelBuffer})
+     */
+    public final DrawingContext getDrawingContext() {
+        if (drawingContext == null) {
+            // note: there is only one implementation of PlatformImage (by QuantumToolkit) so the hard cast here is safe
+            drawingContext = new SWDrawingContext((com.sun.prism.Image) getWritablePlatformImage(), rect -> bufferDirty(rect));
```
mstr2: Multiple `WritableImage` instance can be backed by the same `PixelBuffer`, and `PixelBuffer.updateBuffer()` accounts for that fact by notifying all associated writable images.

However, in this method, you're only invalidating the `WritableImage` from which the `DrawingContext` was obtained. This leaves all other writable images in a stale state.

Maybe this could be solved by using the existing `PixelBuffer.updateBuffer()` logic if the image is backed by a `PixelBuffer`.

modules/javafx.graphics/src/main/native-prism-sw/JJavaSurface.c
hjohn: Thanks, I've fixed this, my JNI/C is a bit rusty. I did some investigation, and it seems at least Hot Spot doesn't care what pointer you pass it -- I was a bit surprised it worked at all (I do have a Java based test that covers this).

I also fixed the potential div by zero there now (for some reason surfaces allow width/height of 0) and cleaned it up a bit since there is an almost exact copy in JDirectBufferSurface.c

modules/javafx.graphics/src/main/java/javafx/scene/image/WritableImage.java
hjohn: I made `bufferDirty` in `PixelBuffer` package private, and call that now from `WritableImage`.  Added a test as well (expanded the stubbing infra a bit for that).

modules/javafx.graphics/src/main/java/javafx/scene/image/WritableImage.java
hjohn: I added a fix for this (via `bufferDirty` in `PixelBuffer`).  I also create the `DrawingContext` now via `Toolkit` which allows me to avoid the exception here (even though it will never fire) as well as allowing me to test the drawing context shared buffer scenario via stub toolkit.

modules/javafx.graphics/src/main/java/javafx/scene/canvas/GraphicsContext.java
hjohn: Yes, it did, fixed. I also noticed that in the SVG append code there was a pre-existing bug that forgot to mark the path dirty -- fixed that as well.

modules/javafx.graphics/src/test/java/test/com/sun/prism/sw/SWDrawingContextTest.java
hjohn: I've removed this... it turns that by calling `Screen.getMainScreen()` it tries to get a screen, which would fail in tests since there were no screens... but we don't use the screen at all in the software renderer... so just by calling `Screen.getMainScreen()` we created this unnecessary dependency.

`SWDrawingContext` now just leaves the screen `null` as it doesn't need it, and I added a bit of documentation on `SWResourceFactory` to explain that `getScreen` might be `null` instead.

modules/javafx.graphics/src/main/java/javafx/scene/image/WritableImage.java
hjohn: This has been modified a bit; Toolkit now provides the DrawingContext. No risk of failure as the toolkit is responsible for providing images as well so it can match types. This also allows the StubToolkit to provide stub images and matching drawing context for tests.

modules/javafx.graphics/src/main/java/javafx/scene/canvas/GraphicsContext.java
hjohn: ah, I was too eager here, fixed now

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
hjohn: Although I think I actually fixed the original concern already with a previous commit (accounting for the stroke), I've refactored this to be simpler now.

I now use the actual rendered pixels (via `SWGraphics`) to determine the dirty rect. A deeper investigation turns out that the system basically already knows this, and it is more accurate than the estimation code that I had before (ie. expand by stroke/miter/etc). It only required relatively minor changes, and removes a lot of "guess" code from `SWDrawingContext`.

I've confirmed this also by supporting italic fonts in the font stubs (a simple guess there won't take overhang of glyphs into account). The test failed before and now passes.

hjohn:
@andy-goryachev-oracle @mstr2  I think this is ready for a re-review; I did do some substantial changes (like adding sliced/direct buffer support) to cover the gaps rather than rejecting these, so this PR has become a bit bigger, but I think most of it is still in scope of this PR and makes the feature more polished.

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
bourgesl: Yes I think marlin renderer determined the appropriate shape's bounds in pixels and subpixels...
See https://github.com/openjdk/jfx/blob/6670f897b0614dfc6ae9f2b08162a56006870870/modules/javafx.graphics/src/main/java/com/sun/marlin/Renderer.java#L1426

It is trivial to share the info through jfx api...

bourgesl: See https://github.com/openjdk/jfx/blob/6670f897b0614dfc6ae9f2b08162a56006870870/modules/javafx.graphics/src/main/java/com/sun/prism/impl/shape/DMarlinRasterizer.java#L92

hjohn: @bourgesl Yeah, I didn't even have to reach that deep, `SWGraphics` has the information already, I only needed to return it: see https://github.com/openjdk/jfx/pull/1969/changes/5e7b57d36ea439aed20afe51c12a420a6ac74215#diff-db59d0f42d49901076557d479771d2668c651374b2dd7aa17f014cd572ca2605

bourgesl: Perfect!
Moreover, it could be adapted if the final shape is a full - rectangle like cropped disk => rect.
😉
I did the trick in marlin-2D

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/QuantumToolkit.java
mstr2: I agree that there's no performance penalty with regards to the native texture formats. The only potential performance impact that I can see concerns existing uses of `PixelReader` and `PixelWriter`. However, this goes both ways, and is not in itself an argument either in favor or against any particular pixel format.

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
```diff
@@ -0,0 +1,1294 @@
...
+ * This class provides a familiar JavaFX-style drawing API for Prism images, enabling modification of the image contents
+ * using lines, rectangles, ovals, rounded rectangles, arcs, polygons, and images without the need for a Canvas or
+ * {@link javafx.scene.image.PixelWriter}.
+ * <p>
+ * It uses Prism's {@link Graphics} and {@link BasicStroke} for rendering, supporting strokes, fills, alpha
+ * transparency, limited blend modes, and automatic dirty-region tracking to efficiently mark affected pixels.
+ * <p>
+ * Features include:
+ * <ul>
+ *   <li>Stroke and fill management (line width, line caps, joins, miter limits, dashes).</li>
```
mstr2: I'd strongly prefer if you used 4 spaces of indentation like everywhere else in the codebase, and that you didn't include the closing `</li>`. I think we should consider adding the 4-spaces rule to our very limited style guide, as I don't think that repeated discussions or a free-for-all "everyone invents their own style" is useful.

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
```diff
@@ -0,0 +1,1294 @@
...
+            graphics.setClipRect(null);
+        }
+        else {
+            graphics.setClipRect(clip);
+        }
+    }
+
+    private Rectangle transformRect(double x1, double y1, double x2, double y2) {
+        double[] src = {x1, y1, x2, y1, x1, y2, x2, y2};
+        double[] dst = new double[8];
```
mstr2, line 593: You could use scratch arrays stored in the class to prevent unnecessary allocations here (maybe reuse `coords`).

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
```diff
@@ -0,0 +1,1294 @@
...
+    }
+
+    @Override
+    public void clip() {
+        throw new UnsupportedOperationException("path based clipping is not supported");
+    }
+
+    @Override
+    public boolean isPointInPath(double x, double y) {
+        // TODO: HTML5 considers points on the path to be inside, but we
```
mstr2: What is left to do here?

modules/javafx.graphics/src/main/java/javafx/scene/image/DrawingContext.java
```diff
@@ -0,0 +1,1670 @@
...
+ * and {@link javafx.scene.canvas.Canvas#getGraphicsContext2D()}.
+ * <p>
+ * The provider of this interface may be associated with a {@link Node} which may
+ * be attached to a {@link Scene}. The operations provided here must be called from the
+ * JavaFX Application Thread. A provider that is not associated with any scene may permit
+ * the operations to be called from any single other thread.
+ * <p>
+ * The {@code DrawingContext} maintains the following rendering attributes
+ * which affect various subsets of the rendering methods:
+ * <table class="overviewSummary" style="width:80%; margin-left:auto; margin-right:auto">
```
mstr2, line 55: Please indent the HTML elements of this table, it is unreadable in source code.

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
bourgesl: Always avoid allocation in hot path...

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
hjohn: Nothing I think, this comment comes from the `GraphicsContext` implementation, and would apply equally to this one. A bit late to change it now when it is already public API.

mstr2: Then it's probably best to remove the TODO.

hjohn: I've modified it into a comment explaining the situation and added a test. I didn't add it to the docs, so we can change our minds later still...

modules/javafx.graphics/src/main/java/javafx/scene/image/DrawingContext.java
hjohn: Yeah, this "junk" style is copied mostly verbatim from `GraphicsContext` when it is was still a proof-of-concept; I've adjusted it now.

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
hjohn: I will leave this until there is proof it is a performance hazard (this is only called by `clipRect`); given that it is likely that the draw operation itself is far more expensive, I think we should keep it as plain Java until we have reason to change it.

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
hjohn: Sorry, adjusted, that doc was a bit of a mess. I totally agree indents should be 4 spaces, and that we can even make that part of the limited style guide. My main criteria is always readable, not strictly conformant to anyone's style-of-the-day.

hjohn:
Regarding performance:

I removed the TODO about dirty rect handling in `SWDrawingContext` now. I've added what I think is a sufficient optimization to `Image.Serial` that may also benefit other cases where multiple rects get dirtied in between render operations.

The way it works is that the dirty rect is now accumulated until something actually has read it. This means that doing multiple drawing operations in a single frame now just expand the dirty rect instead of invalidating the previous one immediately (which then requires a full image upload). Before just doing two draw ops would require a full upload; now you can do dozens of ops and it will only upload the changed area.

I noticed that `PixelWriter` is very poor with this (manipulating a single pixel requires a full upload), but that is a bit out of scope of this PR to fix.

modules/javafx.graphics/src/main/java/javafx/scene/image/WritableImage.java
```diff
@@ -157,6 +159,34 @@ public WritableImage(@NamedArg("reader") PixelReader reader,
...
+     * which is the case for images created with the {@code (width, height)} constructor. An image
+     * created from a {@link PixelBuffer} must use a pixel format of that type.
+     *
+     * @return the {@link DrawingContext} associated with this image, never {@code null}
+     * @throws IllegalStateException if the pixel storage of this image is not in {@code INT_ARGB_PRE}
+     *     format (for example, when created from a {@code BYTE_BGRA_PRE} {@code PixelBuffer})
+     */
+    public final DrawingContext getDrawingContext() {
+        if (drawingContext == null) {
+            drawingContext = Toolkit.getToolkit().createDrawingContext(getWritablePlatformImage(), this::notifyDrawingContextDirty);
```
andy-goryachev-oracle, line 175: `setPlatformImage()` L1086 can replace the platform image (as a result of `snapshot`, for instance).  this will disassociate the cached `drawingContext` from the platform image.

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
```diff
@@ -0,0 +1,1300 @@
...
+
+    @Override
+    public void fill() {
+        if (path.getNumCommands() == 0) {
+            return;
+        }
+
+        path.setWindingRule(fillRule == FillRule.EVEN_ODD ? Path2D.WIND_EVEN_ODD : Path2D.WIND_NON_ZERO);
+
+        graphics.setTransform(BaseTransform.IDENTITY_TRANSFORM);
```
andy-goryachev-oracle, line 982: will that work with non-uniform transform scaling or shear?
also, what would happen if an absolute gradient or image patterns are used?

can we add these cases to the test?

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
```diff
@@ -0,0 +1,1300 @@
...
+
+        applyStrokeParameters();
+
+        /*
+         * The path coordinates were transformed into device space as they
+         * were added, and must be drawn with an identity transform. The stroke
+         * widths and dash lengths must still be adjusted if the scale != 1.0
+         */
+
+        double scale = Math.sqrt(transform.getMxx() * transform.getMxx() + transform.getMyx() * transform.getMyx());
```
andy-goryachev-oracle, line 1005: will it work correctly with a non-uniform scale?

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
```diff
@@ -0,0 +1,1300 @@
...
+
+        if (buffer.hasArray()) {
+            return new SWRTTexture(factory, width, height, buffer.array(), buffer.arrayOffset());
+        }
+
+        if (buffer.isDirect()) {
+            return new SWRTTexture(factory, width, height, buffer);
+        }
+
+        throw new IllegalStateException("image buffer is neither heap array-backed nor direct");
```
andy-goryachev-oracle, line 238: Will `ByteBuffer.allocate().asIntBuffer()` be rejected?

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
```diff
@@ -0,0 +1,1300 @@
...
+        return fontSmoothingType;
+    }
+
+    @Override
+    public void fillText(String text, double x, double y) {
+        drawText(text, x, y, 0, false);
+    }
+
+    @Override
+    public void fillText(String text, double x, double y, double maxWidth) {
```
andy-goryachev-oracle, line 1097: what's the behavior of `fillText()` when `maxWidth` <= 0?

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
```diff
@@ -0,0 +1,1300 @@
...
+        }
+    }
+
+    @Override
+    public void fill() {
+        if (path.getNumCommands() == 0) {
+            return;
+        }
+
+        path.setWindingRule(fillRule == FillRule.EVEN_ODD ? Path2D.WIND_EVEN_ODD : Path2D.WIND_NON_ZERO);
```
andy-goryachev-oracle, line 980: this overwrites the current winding rule - it it compatible with the API spec in `DrawingContext` ?

modules/javafx.graphics/src/main/java/javafx/scene/image/DrawingContext.java
```diff
@@ -0,0 +1,1694 @@
...
+ *             in various path filling, stroking, or clipping operations.
+ *         </td>
+ *     </tr>
+ *     <tr class="altColor">
+ *         <th scope="row" class="colLast" style="width:15%">{@link #setFillRule(javafx.scene.shape.FillRule) Fill Rule}</th>
+ *         <td class="colLast" style="width:10%; text-align:center; color:#0c0">Yes</td>
+ *         <td class="colLast" style="width:10%; text-align:center">{@link FillRule#NON_ZERO NON_ZERO}</td>
+ *         <td class="colLast">
+ *             The method used to determine the interior of paths for a path fill or
+ *            clip operation.
```
andy-goryachev-oracle, line 226: minor: bad indent

modules/javafx.graphics/src/main/java/com/sun/prism/sw/SWDrawingContext.java
```diff
@@ -0,0 +1,1300 @@
...
+    @Override
+    public void drawImage(Image img, double sx, double sy, double sw, double sh, double dx, double dy, double dw, double dh) {
+        if (img == null || img.getProgress() < 1.0) {
+            return;
+        }
+
+        Object platformImage = Toolkit.getImageAccessor().getPlatformImage(img);
+
+        // Ensure it's a Prism image
+        if (!(platformImage instanceof com.sun.prism.Image prismImage)) {
```
andy-goryachev-oracle, line 1206: `GraphicsContext::drawImage()` silently ignores a `null` image, should this method simply return instead of throwing an exception?

andy-goryachev-oracle (commented):
Here is the next batch of comments...

PR #1970, opened 2025-11-11
8333146: Move maven publication logic from build.gradle

Maran23:
Note for reviewers: Some testing was done in a follow-up PR, so worth checking as well: https://github.com/openjdk/jfx/pull/2144

Follow up documentation draft PR: https://github.com/openjdk/jfx/pull/2208

---

This PR splits the maven publishing logic into an own file, `maven-publish.gradle`, that is next to the root `build.gradle`.

The `build.gradle` will apply the `maven-publish.gradle`. The `maven-publish.gradle` will then configure the Maven related properties and register all modules for publication.

This way, we decoupled the logic that much, the only things we need to do:
- apply from `maven-publish.gradle`
- Call `configureMavenPublication` later in the build chain.

To better understand the context, this is the commit where the whole Maven Publishing logic was introduced. I moved all of that logic out of the main `build.gradle`: 5a18677f

This will reduce the size by ~170 lines for the `build.gradle`.

Tested with:
- `./gradlew -PMAVEN_PUBLISH=true -PMAVEN_VERSION=custom publishToMavenLocal`
- `./gradlew -PMAVEN_PUBLISH=true publishToMavenLocal`
- `./gradlew -PMAVEN_PUBLISH=true -PMILESTONE_FCS=true publishToMavenLocal`

Everything still works:
-> Example: javafx.base from the local .m2 repository
image

I think this is a good step and an easy way to split out functionality without blowing things up. We might want to do that for other parts as well.

Note: I also fixed the deprecated `buildDir`, the deprecated `project.task` method and 2 warnings where it seems like he might not be able to infer the type (changing `def` to the actual type). 
-> The file is completely green, no warnings or deprecations.
Note2: I did a small improvement to the `addMavenPublication` method. It is now more 'typesafe', e.g. projects must be passed in directly, not as String. If a project does not exist, the build will fail with an exception.

---------

---------
### Progress

### Issue
 * [JDK-8333146](https://bugs.openjdk.org/browse/JDK-8333146): Move maven publication logic from build.gradle (**Enhancement** - P4)

### Reviewers
 * [Joeri Sykora](https://openjdk.org/census#sykora) (@tiainen - Author)
 * [Johan Vos](https://openjdk.org/census#jvos) (@johanvos - **Reviewer**) Review applies to [ea0b427c](https://git.openjdk.org/jfx/pull/1970/files/ea0b427c23001f5aa7c1b077aee4ea66da294bd7)
 * [Michael Strauß](https://openjdk.org/census#mstrauss) (@mstr2 - **Reviewer**) Review applies to [c30ec300](https://git.openjdk.org/jfx/pull/1970/files/c30ec3001b175f0ba73b3a0fe049d70b7b0deb8d)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1970/head:pull/1970` \
`$ git checkout pull/1970`

`$ git checkout pull/1970` \
`$ git pull https://git.openjdk.org/jfx.git pull/1970/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1970`

`$ git pr show -t 1970`

</details>

<a href="https://git.openjdk.org/jfx/pull/1970.diff">https://git.openjdk.org/jfx/pull/1970.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1970#issuecomment-3515470753)
</details>


kevinrushforth:
Reviewers: @johanvos @tiainen and one of @kevinrushforth or @arapte 

/reviewers 2 reviewers

nlisker:
Note that there's a relevant stale PR in https://github.com/openjdk/jfx/pull/1232 by @jjohannes. Maybe he'd like to take a look.

tiainen (approved):

Maran23:
> Note that there's a relevant stale PR in #1232 by @jjohannes. Maybe he'd like to take a look.

Thanks. It looks like my work will not block/interfere the implementation of that PR, which is good. 
It should be easy to pick the changes from there over in the new `maven-publish.gradle`. 
Should also help for reviewing, since it is more isolated. :-)

jjohannes:
Thanks for the ping. I have to rework my PR anyway. There is no need to pay special attention to it. Any change that makes the publishing setup clearer is good IMO. I think this PR is a good step forward.  

(I didn't get back to the topic yet on the my PR, because it is more work as I initially expected. What is complicating things me there is how conceptually get the setup right with the Jars for different platform built on different machines. I have to set aside some mote time for it.)

maven-publish.gradle
```diff
@@ -0,0 +1,128 @@
+/**
```
kevinrushforth: This file should nave a copyright notice, copied from `build.gradle`.

Maran23: Completely forgot about it. Added!

Maran23:
I found a way to further decouple the maven publishing logic! 

Now we only need to: `apply from: 'maven-publish.gradle'` after all projects were defined. Later in the build chain, we need to call the `configureMavenPublication` Closure from 'maven-publish.gradle', and we are done.

It will work like before, so there is no logical change. But still need a re-review. I retested all configurations.

kevinrushforth:
> It will work like before, so there is no logical change. But still need a re-review. I retested all configurations.

Correction: It isn't just a re-review that is pending. This PR requires review from @johanvos and @arapte (or me, but I'll defer to Ambarish).

johanvos:
@tiainen reviewed it and is ok with it. I had a quick look, and it doesn't cause regression. Also note that we're making progress with the build-openjfx-using-openjdk project, and we hope to create our builds with that approach in the near future.

johanvos (approved):

Maran23:
> Correction: It isn't just a re-review that is pending. 

Indeed! My bad, I didn't express myself clearly. What I meant was that everyone who has already checked this needs to check it again, as some things have changed (not just minor things like the copyright).

Maran23:
> @tiainen reviewed it and is ok with it. I had a quick look, and it doesn't cause regression. Also note that we're making progress with the build-openjfx-using-openjdk project, and we hope to create our builds with that approach in the near future.

That sounds very promising. Good to hear!

arapte:
Could you please verify this observations: 
1. With this change the jar files are generated without class files i.e. sdk is not built. Use command :
`gradle -PMAVEN_PUBLISH=true publishToMavenLocal`. 
Note that default task `sdk` is not specified in the command.
The command completes in just a few seconds (~20 seconds), whereas the `sdk` task can take a few minutes
Attaching the screenshot, please observe that size of jar files are just a few 100 bytes. They should be in KBs.

Screenshot 2025-11-18 at 14 34 01

<br><br>
2. If `sdk` task is added to the above command as : `gradle -PMAVEN_PUBLISH=true publishToMavenLocal sdk`
then the build fails with below error:
```

> Task :base:buildModuleMac FAILED

[Incubating] Problems report is available at: build/reports/problems/problems-report.html

FAILURE: Build failed with an exception.

* What went wrong:
A problem was found with the configuration of task ':base:buildModuleMac' (type 'Copy').
  - Gradle detected a problem with the following location: 'modules/javafx.base/build/module-classes'.
    
    Reason: Task ':base:modularPublicationJarMac' uses this output of task ':base:buildModuleMac' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed.
    
    Possible solutions:
      1. Declare task ':base:buildModuleMac' as an input of ':base:modularPublicationJarMac'.
      2. Declare an explicit dependency on ':base:buildModuleMac' from ':base:modularPublicationJarMac' using Task#dependsOn.
      3. Declare an explicit dependency on ':base:buildModuleMac' from ':base:modularPublicationJarMac' using Task#mustRunAfter.
    
    For more information, please refer to https://docs.gradle.org/9.2.0/userguide/validation_problems.html#implicit_dependency in the Gradle documentation.

* Try:
> Declare task ':base:buildModuleMac' as an input of ':base:modularPublicationJarMac'
> Declare an explicit dependency on ':base:buildModuleMac' from ':base:modularPublicationJarMac' using Task#dependsOn
> Declare an explicit dependency on ':base:buildModuleMac' from ':base:modularPublicationJarMac' using Task#mustRunAfter
> Run with --scan to generate a Build Scan (powered by Develocity).

Deprecated Gradle features were used in this build, making it incompatible with Gradle 10.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

For more on this, please refer to https://docs.gradle.org/9.2.0/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.

BUILD FAILED in 1m 15s
```

Maran23:
> Could you please verify this observations:
> 
> 1. With this change the jar files are generated without class files i.e. sdk is not built. Use command :
>    `gradle -PMAVEN_PUBLISH=true publishToMavenLocal`.
>    Note that default task `sdk` is not specified in the command.
>    The command completes in just a few seconds (~20 seconds), whereas the `sdk` task can take a few minutes
>    Attaching the screenshot, please observe that size of jar files are just a few 100 bytes. They should be in KBs.

Thanks. I can reproduce this! I always built the sdk first and then run the publishing. But you are right, this should work together as well. My guess is that the maven task need to depend on the build task, that should fix both problems. Will do some testing and report back.

Maran23:
Both issues should be fixed now. It really was just a matter of depending the maven publication task on the `buildModulesTask`, which makes the most sense to me.

arapte:
I ran the same command `gradle -PMAVEN_PUBLISH=true publishToMavenLocal` with the latest changes in the PR and without.

With this PR, for a module a few additional tasks get executed compared to without PR.
For example for graphics module, following additional tasks are executed.
```
> Task :graphics:modularJarStandaloneMac
> Task :graphics:copyNativeFilesStandaloneMac NO-SOURCE
> Task :graphics:copyLibFilesStandaloneMac
> Task :graphics:copyLegalStandaloneMac
> Task :graphics:cleanOpenLegalStandaloneMac
> Task :graphics:copyClassFilesMac
> Task :graphics:copyBinFilesMac NO-SOURCE
> Task :graphics:copyLibFilesMac
> Task :graphics:copySourceFilesMac
> Task :graphics:copyDocFilesMac
> Task :graphics:copyBuildPropertiesMac
> Task :graphics:copyClosedLegalStandaloneMac
> Task :graphics:copyLegalMac
```

So, the content of build directories differ. With this change there are 2 additional directories created under build dir: `modular-sdk` and `sdk`

Screenshot 2025-11-19 at 10 37 10


I am not sure which is more correct behavior or if it can cause any issue.
Though seems safe, as ideal way would be to do a full build and then publish.
I would like to request @kevinrushforth and @johanvos to please weigh in.

Maran23:
> So, the content of build directories differ. With this change there are 2 additional directories created under build dir: `modular-sdk` and `sdk`
> I am not sure which is more correct behavior or if it can cause any issue. Though seems safe, as ideal way would be to do a full build and then publish. I would like to request @kevinrushforth and @johanvos to please weigh in.

Thanks a lot for testing as well!
I saw that too and thought the same, it should be more safe and makes more sense. Before, it was depending on each individual module build task like so:
https://github.com/openjdk/jfx/blob/f87448ec156608527d77a4204e98e08052ffecd1/build.gradle#L5850
Which I think is much more error prone and also risks that we forget something or it is executed too early in the future.

kevinrushforth:
@Maran23 I agree that the change @arapte noted above is fine, and your explanation makes sense.

I did find one other difference in artifacts created.

With the current master, running `gradle sdk` creates `build/publications` (because the sdk task indirectly depends on the tasks that create the artifacts for publication). With this patch, it doesn't. That might be an OK change.

However, I also note that with this patch, even `gradle all` doesn't create `build/publications`. Consider adding a dependency to either `gradle sdk` or, more likely, `gradle all` so that still creates `build/publications`.

Maran23:
> However, I also note that with this patch, even `gradle all` doesn't create `build/publications`. Consider adding a dependency to either `gradle sdk` or, more likely, `gradle all` so that still creates `build/publications`.

Thanks for testing!
Just so we have a common understanding: Depending on the `all` task seems to be a better choice than `sdk`? I guess because then we are sure that everything did run before, right?

kevinrushforth:
> > However, I also note that with this patch, even `gradle all` doesn't create `build/publications`. Consider adding a dependency to either `gradle sdk` or, more likely, `gradle all` so that still creates `build/publications`.
> 
> Thanks for testing! Just so we have a common understanding: Depending on the `all` task seems to be a better choice than `sdk`? I guess because then we are sure that everything did run before, right?

Either is fine, but having the `all` task depend on it seems a better choice (the fact that the `sdk` task currently depends on it seems somewhat arbitrary).

Maran23:
> Either is fine, but having the `all` task depend on it seems a better choice (the fact that the `sdk` task currently depends on it seems somewhat arbitrary).

I had a look and gave it a shot. 

If we depend on the `all` task, we will now also build all `apps`, which I think is not a good thing. Since this is unexpected if we just want to (build and) publish the Maven Libraries. In my opinion, the `sdk` task makes more sense (or a task that will do everything except the apps).

kevinrushforth:
> > Either is fine, but having the `all` task depend on it seems a better choice (the fact that the `sdk` task currently depends on it seems somewhat arbitrary).
> 
> I had a look and gave it a shot.
> 
> If we depend on the `all` task, we will now also build all `apps`, which I think is not a good thing. Since this is unexpected if we just want to (build and) publish the Maven Libraries. In my opinion, the `sdk` task makes more sense (or a task that will do everything except the apps).

I think you might have it backwards: `all` should depend on the task to build the publications, not the other way around. The goal is that `gradle all` should produce `build/publications`.

Maran23:
> I think you might have it backwards: `all` should depend on the task to build the publications, not the other way around. The goal is that `gradle all` should produce `build/publications`.

Ah okay, I got you now! I will have a look how this can be achieved.

Maran23:
> I think you might have it backwards: `all` should depend on the task to build the publications, not the other way around. The goal is that `gradle all` should produce `build/publications`.

Did you have a specific idea in mind when writing the message? 
I ran a few tests, but nothing worked without explicitly writing `publishToMavenLocal` within the command.
And I don't want to worsen the separation between the main and publish gradle files.
If there's no idea at the moment, then I'll continue looking for a good solution!

EDIT: Also, `-PMAVEN_PUBLISH=true` need to be specified always as of now. Again, this is because it is completely separated now (and don't run if it is set to false, which will save some seconds for the normal case where this is not needed). I don't have a good idea for this right now.

Maran23:
> I think you might have it backwards: `all` should depend on the task to build the publications, not the other way around. The goal is that `gradle all` should produce `build/publications`.

Here is an attempt to do that without coupling the build with maven-publish:
https://github.com/Maran23/jfx/commit/d325d9931b37c64641fb9064e5deadbfa5766fe8

This probably needs to be done with `compileTargets` (maybe?) then. But the idea is the same.

Maran23:
I think I found a good solution for the: `:all` task should also build publications problem.

Maran23:
@jjohannes, @koppor told me you are the Gradle Expert to ask for feedback😄.
If you have a bit time, could you review this PR? :)

- I was mostly wondering if this is the right location for the new `maven-publish.gradle`, or should/can we move it to `buildSrc`?
- And if my idea with `isAllTaskRequested` seems fine, or if there is a better way?

I saw that in https://github.com/openjdk/jfx/pull/1232 you made some publishing improvements and used `apply plugin: 'java-library'` instead, but this PR is only about extracting the existing publishing logic for now.

mstr2:
This PR seems to be stale, even though it has two approving reviews by @johanvos and @tiainen.
Is there any reason why this doesn't move forward?

mstr2 (approved):
This seems to be a straightforward refactor.

kevinrushforth:
Before this goes in, I want to review and test it on our CI to be sure it doesn't break anything.

Maran23:
I think this might be a good candidate after RDP1.

jjohannes:
@Maran23 as you asked for feedback: I think this is a good first step.

----

My general recommendation for  improving the build setup in the _openjdk_ project is to move things into `buildSrc` as so-called _pre-compiled script plugins_. This gives more structure and type-safety. In this approach, every "aspect" of the build becomes its own plugin and you then compose the separate aspects to one build configuration.

The difference would be that you would place the `.gradle` files here, by which they automatically become plugins:

```
buildSrc/src/main/groovy/javafx.gradle-build.maven-publish.gradle
```

Then I would eventually replace the big `build.gradle` in the root with individual `build.gradle` files in each module. The files have a plugins block to identify and their individual dependencies blocks. 

```
plugins {
  id 'javafx.gradle-build.maven-publish'
  id 'javafx.gradle-build.other-build-featue'
}

dependencies {
  api project(':base') 
}
```

Further reading Resources:
- An example project that shows this setup with more explanation in the readme:
  https://github.com/jjohannes/gradle-project-setup-howto
- Why you want to separate build logic/config from dependency definitions:
  https://javarca.de

With such a setup, one can also eventually move out `buildSrc` into a different repository and build and publish it independently. This repo would then only contains the dependency definitions and not all the complex building details. (Which, IIRC, was a topic discussed in the PR where I tried – and failed so far – to add Gradle Metadata publishing).

Another interesting option for a Java-centric project as this one: Once you move all configuration to `buildSrc`, you can write it in `.java` files (instead of `.gradle` or `.groovy`). Here is an example of build configuration in a separate repo written in Java: https://github.com/gradlex-org/internal-build-conventions

Maran23:
> @Maran23 as you asked for feedback: I think this is a good first step.

Thank you for your feedback and the explanation, really appreciate it.

Since the OpenJDK build system is currently being evaluated, I don't think we need to refactor the current `build.gradle` (at least until we know more). 

The reason why I did the 'Maven publishing' extraction regardless is mostly because of https://github.com/openjdk/jfx/pull/2144 (and well, learning Gradle😀), which makes testing locally much easier.
It seems that people like @FlorianKirmaier and @eduardsdv also use this way of testing. 
For me, this makes it very easy to try a version for many different applications - I just change the JavaFX version in the `pom.xml` or `build.gradle` and can start testing.

Additionally, whatever the build system will be in the future, I would still like to support the 'Maven Publishing' process, if possible.
So if we can remove most of Gradle, I hope that this refactoring makes it easier to keep the relevant parts we might still need (or, if we don't need Gradle anymore, still enable us developers to use it with our locally installed Gradle on our own).

In any case, moving the publishing file into `buildSrc` so it will become a plugin sounds like a good next step. This is something I would like to experiment with, after this PR is merged. (@kevinrushforth, this might be a good discussion for the mailing list?).
Not just to follow the convention (and again, learning), but PRs like https://github.com/openjdk/jfx/pull/2156 will again make the `build.gradle` grow and I would rather like to see those as new plugins. And hopefully the insights I gained when doing that will help doing so in the future.

Maran23:
I created a draft PR here: https://github.com/openjdk/jfx/pull/2208 to add some explaination how developers can use this feature (`publishToMavenLocal`) to easily test their applications with a branch or local changes. 

Actually, I wish I'd known that from the start. 
Because before I knew about the Maven publishing, I tested the changes by adding the libraries manually (using an IntelliJ feature), which takes much longer to do right.

So to document that is IMO a good step to enable (potential) contributors to easily test their changes or other PRs with their existing applications and code.

Maran23:
@johanvos / @tiainen could you rereview this PR?
Really would like to get this finalized - especially as this will make testing much easier for many people (including myself, using this for > 1 year already).

buildSrc/src/main/groovy/com/sun/javafx/gradle/CompileTarget.java
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright (c) 2025, Oracle and/or its affiliates. All rights reserved.
```
kevinrushforth: 2026

maven-publish.gradle
```diff
@@ -0,0 +1,223 @@
...
+import com.sun.javafx.gradle.CompileTarget
+
+boolean isAllTaskRequested() {
+    return gradle.startParameter.taskNames.any { String taskName ->
+        taskName == "all" || taskName == ":all"
+    }
+}
+
+defineProperty("MAVEN_PUBLISH", "false")
+ext.IS_MAVEN_PUBLISH = Boolean.parseBoolean(MAVEN_PUBLISH) || isAllTaskRequested()
```
kevinrushforth: This flag was formerly `false` by default even when running `gradle all`. Are there other side effect to now setting this flag to `true`? Did you consider a separate flag that qualifies the creation of the `publications` dir? It's probably OK as-is, since I didn't see any issues with the build.

kevinrushforth (approved):
Sorry for the delay. I ran a build on our CI system and all looks good. I left one comment about a copyright year and will reapprove if you change it.

I also ran a pair of builds -- one from `master` and one from this PR branch (with master merged in) -- using `SOURCE_DATE_EPOCH` and they are identical, aside from a known issue that the generated `.bss` files in the `javafx.controls` module are not reproducible.

Please wait for either @johanvos or @tiainen

maven-publish.gradle
Maran23: There is no side effect, it should behave as before.

Another idea I just got: We could similar as before always call `configureMavenPublication` but it will only execute the tasks during `gradle all`, as before anyway. This will make the code a bit better.
(My initial approach was to combine the Maven publishing and the `publications` folder. But based on your comments, that are two separate things really).
Pushed the changes.

On a side note, I was wondering if we even need `IS_MAVEN_PUBLISH`. The artifacts are only published when running Gradle with `publishToMavenLocal` anyway.

So when `MAVEN_PUBLISH` is false, `publishToMavenLocal` is not found and does not work since we don't apply the plugin.
Setting `MAVEN_PUBLISH` to true without `publishToMavenLocal` will apply the plugin but do nothing.

kevinrushforth (approved):
Yes, this looks better. Thanks.

tiainen (approved):
Looking good to me.

kevinrushforth:
This has had enough eyes on it, and an earlier version was approved earlier by two other "R"eviewers, so I am lowering the review requirement to 2 total (1 reviewer + 1 author).

OK to integrate

/reviewers 2

PR #1981, opened 2025-11-23
8210547: [Linux] Uncontrolled framerate

notzed:
As discussed on the mailing list.

OpenGL requires an explicit glFinish() to synchronise with the video refresh.  The current usage is incorrect and against the specification.  This patch calls glFinish() after all windows have been drawn following a pulse.

---------
### Progress

### Issue
 * [JDK-8210547](https://bugs.openjdk.org/browse/JDK-8210547): [Linux] Uncontrolled framerate (**Bug** - P5)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1981/head:pull/1981` \
`$ git checkout pull/1981`

`$ git checkout pull/1981` \
`$ git pull https://git.openjdk.org/jfx.git pull/1981/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1981`

`$ git pr show -t 1981`

</details>

<a href="https://git.openjdk.org/jfx/pull/1981.diff">https://git.openjdk.org/jfx/pull/1981.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1981#issuecomment-3567904683)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/PaintCollector.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/PaintCollector.java
@@ -434,10 +434,14 @@ final class PaintCollector implements CompletionListener {
                 if (!needsHint) {
                     needsHint = gs.isSynchronous();
                 }
+            }
+
+            for (final GlassScene gs : dirtyScenes) {
                 // On platforms with a window manager, we always set doPresent = true, because
                 // we always need to rerender the scene  if it's in the dirty list and we do a
                 // swap on a per-window basis
                 gs.setDoPresent(true);
+                gs.setDoVSync(needsHint && dirtyScenes.getLast() == gs);
                 try {
                     gs.repaint();
                 } catch (Throwable t) {
--- a/modules/javafx.graphics/src/main/java/com/sun/prism/es2/ES2SwapChain.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/prism/es2/ES2SwapChain.java
@@ -183,8 +183,16 @@ class ES2SwapChain implements ES2RenderTarget, Presentable, GraphicsResource {
 
     @Override
     public boolean present() {
-        boolean presented = drawable.swapBuffers(context.getGLContext());
-        context.makeCurrent(null);
+        return drawable.swapBuffers(context.getGLContext());
+    }
+
+    @Override
+    public boolean present(boolean vsync) {
+        boolean presented = present();
+
+        if (vsync)
+            context.getGLContext().finish();
+
         return presented;
     }
 
@@ -308,6 +316,7 @@ class ES2SwapChain implements ES2RenderTarget, Presentable, GraphicsResource {
         }
 
         if (drawable != null) {
+            context.makeCurrent(null);
             drawable.dispose();
             drawable = null;
         }
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassScene.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassScene.java
@@ -65,6 +65,7 @@ abstract class GlassScene implements TKScene {
     private volatile boolean entireSceneDirty = true;
 
     private boolean doPresent = true;
+    private boolean doVSync = false;
     private final AtomicBoolean painting = new AtomicBoolean(false);
 
     private final boolean depthBuffer;
@@ -305,6 +306,14 @@ abstract class GlassScene implements TKScene {
 
     public final synchronized boolean getDoPresent() {
         return doPresent;
+    }
+
+    public final synchronized void setDoVSync(boolean value) {
+        doVSync = value;
+    }
+
+    public final synchronized boolean getDoVsync() {
+        return doVSync;
     }
 
     protected Color getClearColor() {
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/PresentingPainter.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/PresentingPainter.java
@@ -104,7 +104,7 @@ final class PresentingPainter extends ViewPainter {
 
                 /* present for vsync buffer swap */
                 if (vs.getDoPresent()) {
-                    if (!presentable.present()) {
+                    if (!presentable.present(vs.getDoVsync())) {
                         disposePresentable();
                         sceneState.getScene().entireSceneNeedsRepaint();
                     }
--- a/modules/javafx.graphics/src/main/java/com/sun/prism/Presentable.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/prism/Presentable.java
@@ -57,6 +57,10 @@ public interface Presentable extends RenderTarget {
      */
     public boolean present();
 
+    public default boolean present(boolean vsync) {
+        return present();
+    }
+
     public float getPixelScaleFactorX();
     public float getPixelScaleFactorY();
 }
--- a/modules/javafx.graphics/src/main/native-prism-es2/PrismES2Defs.h
+++ b/modules/javafx.graphics/src/main/native-prism-es2/PrismES2Defs.h
@@ -263,6 +263,7 @@ struct ContextInfoRec {
 #ifdef UNIX /* LINUX || SOLARIS */
     char *glxExtensionStr;
     PFNGLXSWAPINTERVALSGIPROC glXSwapIntervalSGI;
+    PFNGLXSWAPINTERVALEXTPROC glXSwapIntervalEXT;
 #endif /* LINUX || SOLARIS */
 
     /* gl function pointers */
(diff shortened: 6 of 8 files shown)


kevinrushforth:
This will need careful review.

Reviewers: At least two of: @lukostyra @arapte @kevinrushforth 

/reviewers 2

kevinrushforth:
@notzed Please enable GHA testing for your repo. Go you your "Actions" tab in your repo and enable them with the green "enable workflows" button. Then push an empty commit to trigger a test run.

kevinrushforth:
@notzed As with the fix @arapte did for PR #1978, you will need to qualify this with `PrismSettings.isVsyncEnabled`.

My quick take on your PR is that the changes are pretty intrusive and not limited to the ES2 pipeline. I recommend exploring whether the changes can be more targeted.

notzed:
prism.vsync=false behaves visually the same as it currently does because double buffering isn't enabled via glXSwapIntervalSGI().  i.e it runs at roughly 62.5fps by default.  One caveat is that glFinish() will pause the CPU to wait for the GPU and that may have throughput impacts, or it may be desirable to avoid over-buffering.

glFinish() is the only way to sync the cpu with the video card[1] and honour the expectations of prism (i.e. that presentation with 'vsync supported' is throttled by the frame-rate) in the same way directx is.  And the only place it makes sense is immediately after glSwapBuffers().  But it can't be done per-window otherwise all windows are displayed round-robin with the frame rate.

The problematic point is the QuantumTooklit.vsyncHint() call which itself causes unthrottled pulses - they are only throttled in practice by Presentable.present().  vsyncHint() also seems to not be used anywhere to actually indicate vsync, it just fires another pulse after rendering is finish which just upsets the pulse regularity.

I tried a number of variations over a couple of weeks and this was by far the simplest I could come up with but i'm open to suggestions.

Note [1] also suggests a 'high precision timer' should be used for frame timing instead of vsync, and gdk_timeout used by the x11 backend is definitely not suitable for this purpose, even a simple ScheduledExecutorService (as used by monocle and headless) is far more accurate.  But discovering accurate frame-rates is problematic (they should be rational but the gtk api's don't support that, and gtk4 and wayland  has it's whole own idea of animation timing) and they can change at runtime or even across screens.  I do have another patch to enable nanosecond (or it could be microsecond) frame times and a floating point animation rate ... but that is far far more intrusive.

[1] https://wikis.khronos.org/opengl/Swap_Interval#GPU_vs_CPU_synchronization

lukostyra:
I tested your changes but they do not fix the original problem which #1929 addresses. I just tested this on my Intel UHD machine and the framerate is not tied to VSync (FPS meter from the original issue is around 1000fps instead of 60fps).

I think it's getting pretty confusing regarding both this change and Thiago's PR, so I would like to figure some things out before we progress. I looked back into the mailing list discussion but I'm still not fully clear on everything.

Is this change actually supposed to address the problem from [JDK-8210547](https://bugs.openjdk.org/browse/JDK-8210547)? If so, is this change in place of #1929 or does it depend on it?

notzed:
This is for JDK-8210547.

1929 is separate, it doesn't work for me when you have more than 1 window open.  It's effectively just the same as calling glFinish() on every glxSwapBuffers() which just tanks the framerate (and for me whether it uses *SGI or *EXT makes no difference).

I suppose I could see if using glXSwapIntervalEXT() makes a difference here.

FWIW this is what i'm using to test, it intentionally advances on animation pulses rather than using the the timestamp to visualise spurious pulses.  With javafx-25 it goes mental but with the given patch it runs at the right rate but with spurious pulses from the vsyncHint stuff.

BEWARE!  It uses Robot to take over the mouse, ESC/Q or alt-f4 quits.

[Latency.java](https://github.com/user-attachments/files/23870834/Latency.java)

lukostyra:
> I suppose I could see if using glXSwapIntervalEXT() makes a difference here.

It might be better to switch to `glXSwapIntervalEXT()` then. It seems to be more consistent across all platforms (ie. works on Intel where the others don't for some reason). I'll be waiting for your updates, and I think we can close #1929 then. Thanks for attaching the test program too, it'll come in handy.

Even though these changes are simple, they do go quite deep so we'll need to spend some extra time reviewing and testing them.

A couple of formalities as well:
- PR title must begin with issue number without `JDK-` and followed by a colon
- Text following the issue number must be the same as the JBS issue counterpart

So in this case, PR should be named exactly like #1929 for Skara to properly connect it.

Last but not least, when you update this branch please don't use `git rebase` and force pushes, as it breaks the review tool and makes it harder for us to iteratively review your changes. If you ever need to catch up to master or to resolve any conflicts, use `git merge` instead.

notzed:
Sorry I made a mistake with the push, Kevin directed me to send an "empty commit" to trigger automation and I couldn't work out how to do that otherwise at the time - until just after I'd done it and it was too late then.

I'll look at merging in bits of 1929, sans the visual stuff which seems unrelated.

modules/javafx.graphics/src/main/java/com/sun/prism/es2/ES2SwapChain.java
```diff
@@ -183,8 +183,16 @@ private void drawTexture(ES2Graphics g, RTTexture src,
...
-        boolean presented = drawable.swapBuffers(context.getGLContext());
-        context.makeCurrent(null);
+        return drawable.swapBuffers(context.getGLContext());
+    }
+
+    @Override
+    public boolean present(boolean vsync) {
+        boolean presented = present();
+
+        if (vsync)
```
lukostyra, line 193: I would add a comment here describing the reason for this addition so we have it for future reference - something along the lines of glFinish needing to be called only on the last presented Stage in order to properly synchronize them all to VSync.

lukostyra (approved):
Change looks good to me and it now works as expected on my old Intel laptop (framerate is properly capped without `-Dquantum.multithreaded=false`). I have one more comment suggestion before approving this while I run some additional tests in the meantime.

modules/javafx.graphics/src/main/java/com/sun/prism/es2/ES2SwapChain.java
```diff
@@ -183,8 +183,16 @@ private void drawTexture(ES2Graphics g, RTTexture src,
...
+            context.getGLContext().finish();
```
kevinrushforth, line 194: In addition to a comment, you need to add curly braces. We do not allow a single statement body for an "if" or "else" if it in another line.

```suggestion
        // ADD A COMMENT HERE
        if (vsync) {
            context.getGLContext().finish();
        }
```

lukostyra (changes requested):
I also just noticed I gave my approval earlier, while I meant to simply comment on this request.

notzed:
No worries.  It's been so long I wasn't sure it was even still active, 
but  I'll have a look next week at where it's all at.

  Z

tsayao:
I noticed that `Application.java` has a  `staticScreen_getVideoRefreshPeriod` - I have not tested it, but I wonder if this would affect the issue.

lukostyra:
> I noticed that `Application.java` has a `staticScreen_getVideoRefreshPeriod` - I have not tested it, but I wonder if this would affect the issue.

It seems to provide the same 0.0 refresh period across all platforms, which then `QuantumToolkit` picks up. If refresh period is 0.0 the Toolkit should use platform native vsync. So I think this is correct, it's the problem with specifically Intel and their drivers not catching that vsync should be enabled.

lukostyra:
@notzed when you come back to this issue remember to call Skara command `/template append` - the PR description changed a bit and needs updating (+ accepting a checkbox that appears in the process).

tsx84:
Hi,
I came across your changes while investigating an unrelated problem - a somewhat random SIGSEGV crash inside the native graphics driver, preceded by a BadAlloc error from glXMakeCurrent. I applied your patch to the 17u branch (which is what we currently work with) to see if by any chance it would help :)

With regards to my crash it did help, but in my application I noticed visible artifacts around menu items from a menu bar when hovering over those items. The items themselves were readable, but around the popup a different part of the main screen was shown, rotated and mirrored. So the graphics content between the main window and the popup was not completely separated. Sadly, I cannot share an image of the artifacts.

What helped in my case: https://github.com/openjdk/jfx/compare/master...tsx84:jfx:GLX_MakeCurrent_Bug

I figured that it might be necessary to invalidate the Java-side tracking of the current drawable after presenting, so that glContext.makeCurrent and glContext.bindFBO(0) are called reliably on the next createGraphics() call — while still avoiding the extra glXMakeCurrent call to the dummy drawable.

I did not test my fix combined with your other changes, just wanted to share my observation :)

Greetings

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/PaintCollector.java
```diff
@@ -434,10 +434,14 @@ final void renderAll() {
...
                     needsHint = gs.isSynchronous();
                 }
+            }
+
+            for (final GlassScene gs : dirtyScenes) {
                 // On platforms with a window manager, we always set doPresent = true, because
                 // we always need to rerender the scene  if it's in the dirty list and we do a
                 // swap on a per-window basis
                 gs.setDoPresent(true);
+                gs.setDoVSync(needsHint && dirtyScenes.getLast() == gs);
```
tsx84, line 444: a minor note: maybe the compiler is clever enough to pull dirtyScenes.getLast() out of the loop, but I would to it explicitly

PR #1982, opened 2025-11-24
8372415: Stage size should match visual window bounds

mstr2:
On Windows, the `Stage.width` and `Stage.height` correspond to the window size as returned by `GetWindowRect`.

Up until Windows 10, the size of a window was identical to its visual borders. However, since Windows 10 has introduced thin visual window borders, the window manager adds an invisible border of a few pixels around the window to make it easier to resize the window. Since `GetWindowRect` returns the window size _including_ these invisible borders, the location and size of a `Stage` isn't exactly what we'd expect.

For example, if we place a `Stage` at `setX(0)` and `setY(0)`, the window appears with a small distance from the screen edge, and the window size extends a few pixels beyond its visual borders (in the following images, the screenshot size corresponds to the window size; note the invisible padding around the edges):
window-size-1

What we actually want is to have the visual borders line up with the edges of the screen, and have the window size correspond to the visual borders:
window-size-2

The implementation is quite simple: instead of `GetWindowRect`, we use `DwmGetWindowAttribute(DWMA_EXTENDED_FRAME_BOUNDS)`. This gives us the bounds of the visual window borders. If this function fails, we fall back to `GetWindowRect` (now, I don't know why `DwmGetWindowAttribute(DWMA_EXTENDED_FRAME_BOUNDS)` would ever fail... maybe an old Windows version in a remote desktop scenario?).

---------
### Progress

### Issues
 * [JDK-8372415](https://bugs.openjdk.org/browse/JDK-8372415): Stage size should match visual window bounds (**Bug** - P4)
 * [JDK-8285983](https://bugs.openjdk.org/browse/JDK-8285983): JavaFX reports maximized window size larger than the physical screen (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1982/head:pull/1982` \
`$ git checkout pull/1982`

`$ git checkout pull/1982` \
`$ git pull https://git.openjdk.org/jfx.git pull/1982/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1982`

`$ git pr show -t 1982`

</details>

<a href="https://git.openjdk.org/jfx/pull/1982.diff">https://git.openjdk.org/jfx/pull/1982.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1982#issuecomment-3571106775)
</details>


beldenfox:
I haven't uncovered any problems so far but it took me a while to get my bearings. You're working with two coordinate systems, the one used by JavaFX which doesn't include the invisible border and the one used by Windows which does. The naming convention is confusing since the "extended" rectangle is actually the smaller of the two. It would also be good to see some comments clarifying that `m_insets`, `m_maxSize`, and `m_minSize` are all specified in the JavaFX system.

mstr2:
> I haven't uncovered any problems so far but it took me a while to get my bearings. You're working with two coordinate systems, the one used by JavaFX which doesn't include the invisible border and the one used by Windows which does. The naming convention is confusing since the "extended" rectangle is actually the smaller of the two. It would also be good to see some comments clarifying that `m_insets`, `m_maxSize`, and `m_minSize` are all specified in the JavaFX system.

I don’t think it’s quite "two coordinate systems" (JavaFX vs Windows). `Stage.width/height` are intended to include the normal system decorations (title bar + borders), which historically also matches what people mean by "window size": it includes the non-client area, but not purely visual effects like drop shadows. The weirdness on Windows 10+ is that GetWindowRect started including an extra invisible resize border that’s not part of the visible frame, so the window ends up being reported a few pixels larger than it looks. Using `DwmGetWindowAttribute(DWMA_EXTENDED_FRAME_BOUNDS)` is basically a way to get the visually perceived bounds back.

beldenfox:
I was thinking that JavaFX and Windows have different ideas of where the window origin (0, 0) lies; Windows thinks it's at the top left of drop shadow area and JavaFX thinks it's at the top left of the title bar (in this PR). But you're right, what's relevant is that there are two rectangles of different sizes. I still think it could be clearer that, say, `m_insets` applies to the smaller of the two (which is the extended one).

mstr2:
I've added a comment to clarify that `m_minSize`, `m_maxSize`, and `m_insets` are specified with respect to the extended frame bounds.

mstr2:
With this fix applied, the problem described in [JDK-8285983](https://bugs.openjdk.org/browse/JDK-8285983) disappears and the reported position and size of a full-screen window is as a sane person would expect it to be (position = 0,0 and size = visual bounds of the screen).

I suggest we re-open the bug (which was closed as "not an issue") and add it to this PR.

beldenfox (changes requested):
The existing changes look fine. GitHub doesn't allow me to comment on unchanged code so I'll just have to write up my comments here.

In GlassView.cpp the `_getX` and `_getY` routines return the location of the scene relative to the upper-left corner of the window. I believe these offsets should be relative to the extended frame bounds instead of GetWindowRect.

In GlassWindow.cpp there's a call to the Java routine `notifyMoving` (which lives in WinWindow.java). At that call site the x and y coordinates are based on GetWindowRect and the anchor is relative to GetWindowRect but the insets are relative to the extended frame bounds. `notifyMoving` is also called from the `setBounds` routine in WinWindow.java. In that case I believe it's being called with the coordinates of the extended frame bounds, an anchor relative to GetWindowRect, and insets relative to the extended frame bounds. I think this all needs further review.

mstr2:
I've looked at every call to `GetWindowRect` and, where appropriate, replaced it with `GetExtendedFrameBounds`.
In addition to that, I also tested the case when JavaFX is running in a DPI-unaware process. I've added code that detects this situation and applies the correct mapping from physical coordinates into virtualized coordinates.

modules/javafx.graphics/src/main/native-glass/win/Utils.cpp
```diff
@@ -107,6 +108,176 @@ jint GetModifiers()
...
+        impl_t() {
+            HMODULE hModule = GetModuleHandleW(L"user32.dll"); // user32 is already loaded
+
+            pGetThreadDpiAwarenessContext = reinterpret_cast<FnGetThreadDpiAwarenessContext*>(
+                GetProcAddress(hModule, "GetThreadDpiAwarenessContext"));
+
+            pSetThreadDpiAwarenessContext = reinterpret_cast<FnSetThreadDpiAwarenessContext*>(
+                GetProcAddress(hModule, "SetThreadDpiAwarenessContext"));
+
+            // Only load GetProcessDpiAwareness if GetThreadDpiAwarenessContext is not available (pre-Win10).
```
beldenfox, line 145: What versions of Windows does JavaFX support? I just spent a half hour searching and couldn't find any statement on this for either JavaFX or the JDK.

As of October of this year the original version of Windows 10 left Microsoft's Long Term Servicing Channel. As of now the earliest version of Windows supported by Microsoft is Win10 2016 (1607), the version where the thread DPI awareness calls were first added. Is JavaFX supporting and testing earlier versions of Windows? If not you should be able to just make direct calls to the thread DPI awareness API's.

This is probably a topic for the mailing list. I'm still surprised I couldn't find this information anywhere.

mstr2: Even if we didn't _officially_ support Windows 10, I think it's a little too early to remove functionality given that this OS still has a ~40% share of Windows installations.

kevinrushforth: Definitely. We don't need to worry about Windows 8 or earlier, but it's premature to break Windows 10.

beldenfox: I don't want to stop supporting Windows 10.  I'm referring specifically to Windows 10 RTM, the very first version of 10, which has fallen entirely out of Microsoft's support system. If we only need to support versions of Windows that Microsoft is supporting that would be Windows 10 2016 (1607) in which case there's no need to mess with all the DLL and function pointer retrieval in this code. You can just issue direct calls to SetThreadDpiAwarenessContext and friends.

I will add that though we might say in this thread that we don't need to worry about Windows 8 or earlier I really and truly could not find any public statement to that effect anywhere.

mstr2: If we removed that path, then for Windows 10 < 1607, we wouldn't be able to skip coordinate mapping for the usual (99.9%) case when the Java process is running with per-monitor DPI awareness. This wouldn't cause JavaFX to stop working, we'd just be back at the status quo where we don't account for the visible window frame.

We still need to dynamically load the function though, as it definitely won't be available on Win10 < 1607.

beldenfox: > We still need to dynamically load the function though, as it definitely won't be available on Win10 < 1607.

Officially you're doing the right thing by writing code that works on all versions of Windows 10. But this assumes there are users out there with the original release of Windows 10 (no updates past 2015) who will want to install a JavaFX 26-based app. Are there? Can we even test this code? I don't see how this is sustainable going forward.

This is not really a question for you. I should take it to the mailing list.

mstr2: I can't test this in a systematic way, as I can't be bothered to get an installation of an ancient version of Windows. However, in this particular case, I _did_ test what would happen if the new functions can't be loaded by simply commenting out the code that loads the new functions.

beldenfox:
Sorry it's been so long since my last review. I did a few tests just to confirm that the Windows platform does behave very differently from macOS and Linux. When attempting to place windows side-by-side or aligned with the edges of the screen there's a much bigger gap on Windows than the other platforms.

I feel that the code could be easier to understand and review. For example, when imposing min and max bounds it would be clearer if the code converted from window rect to extended rect, applied the bounds, and then converted back. I am willing to review the code as it stands but would appreciate it if you could take a look and see if the conversions could be more centralized for clarity.

mstr2:
> I feel that the code could be easier to understand and review. For example, when imposing min and max bounds it would be clearer if the code converted from window rect to extended rect, applied the bounds, and then converted back. I am willing to review the code as it stands but would appreciate it if you could take a look and see if the conversions could be more centralized for clarity.

I don't understand what you're suggesting. Are you talking about `WM_GETMINMAXINFO`?

beldenfox:
> I don't understand what you're suggesting. Are you talking about `WM_GETMINMAXINFO`?

Yes. The code retrieves the shadow and extended rects and computes the difference between them in _setBounds, WM_GETMINMAXINFO, and HandleWindowPosChanging and there are a couple places where those differences are used to convert from one rect to another using similar blocks of code. In _setBounds the min and max are applied to the extended rect which is obviously correct but in WM_GETMINMAXINFO I actually have to think about what you're doing which isn't complicated but, well, I'm not a patient person (mea culpa). It's possible that these similarities are superficial and it's not worth gathering them into a helper object or helper routines but at first blush it looks like I'm reviewing the same code multiple times and figuring out which direction it's going each time.

mstr2:
I've put some of the repeated calculations into a `FrameBounds` helper object.

modules/javafx.graphics/src/main/native-glass/win/Utils.cpp
```diff
@@ -107,6 +108,198 @@ jint GetModifiers()
...
+/*
+ * Returns the visible window bounds (excluding the DWM shadow / invisible resize borders) in the same
+ * coordinate space that the rest of Glass expects: logical coordinates in the caller's DPI-awareness space.
+ *
+ * While DwmGetWindowAttribute(DWMWA_EXTENDED_FRAME_BOUNDS) reports the visible frame bounds in physical
+ * coordinates (it is not DPI-virtualized for the caller), GetWindowRect() returns the window rectangle
+ * in the caller's DPI-awareness space. If the caller is DPI-unaware or system-DPI-aware, Windows may
+ * DPI-virtualize the returned coordinates.
+ *
+ * It is possible (though uncommon) for the JVM to run in a DPI-unaware or system-DPI-aware context.
```
beldenfox, line 120: Do you have a recommended way of testing JavaFX as DPI unaware? This is hard to do these days presumably due to the manifest you mention.

modules/javafx.graphics/src/main/native-glass/win/Utils.cpp
```diff
@@ -107,6 +108,198 @@ jint GetModifiers()
...
+         * We can only safely skip mapping when we know the current process is per-monitor DPI aware.
+         * In that case, caller-space coordinates are already physical pixels, matching DWMWA_EXTENDED_FRAME_BOUNDS.
+         * If we can't prove that we're PM-aware, we don't skip mapping.
+         */
+        BOOL canSkipMapping() const {
+            // Supported on Windows 10+
+            if (pGetThreadDpiAwarenessContext) {
+                DPI_AWARENESS_CONTEXT currentAwareness = pGetThreadDpiAwarenessContext();
+
+                return currentAwareness == DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE ||
```
beldenfox: On my system canSkipMapping always returns false. I think to compare DPI_AWARENESS_CONTEXT values you have to use AreDpiAwarenessContextsEqual. The documentation for DPI_AWARENESS_CONTEXT sure makes it look like an enum or some other type of simple scalar but apparently it's not.

modules/javafx.graphics/src/main/native-glass/win/Utils.cpp
mstr2: I've only managed to do this by changing compatibility settings for the Java launcher (right-click on java.exe, go to properties->compatibility).

beldenfox: Have you tried using PhysicalToLogicalPointForPerMonitorDPI? I did some quick tests and that call seems to be obeying the DPI awareness settings and performing the correct conversions. Since it was introduced in Windows 8.1 it can be called unconditionally. To access it you'll need to bump the Windows API version up since it's currently set at 7. For my testing I bumped it up to 10 in common.h:

 ```
 #define _GLASS_COMMON_

 #ifndef _WIN32_WINNT
-    #define _WIN32_WINNT 0x0601
+    #define _WIN32_WINNT 0x0A00
 #endif
 #ifndef _WIN32_IE
     #define _WIN32_IE 0x0500
@@ -37,6 +37,10 @@
     #define _WIN32_WINNT_ _WIN32_WINNT
 #endif

+#ifndef NTDDI_VERSION
+    #define NTDDI_VERSION NTDDI_WIN10
+#endif
+
 #pragma warning(disable : 4675)
```

Maran23:
I wonder if this fixes the issue I brought up with the border in https://github.com/openjdk/jfx/pull/2188

Maran23:
Any plan to reopen this?

PR #1986, opened 2025-11-25
8372530: Easier placement of windows with positioning anchor

mstr2:
This enhancement allows `Stage` to be placed on the screen similar to a popup window, where a user-specified positioning anchors defines a point on the stage that should coincide with a given location on the screen. For this purpose, the following new methods are added to `Stage`:

```java
public class Stage {
    public void relocate(AnchorPoint screenAnchor, AnchorPoint stageAnchor);
    public void relocate(AnchorPoint screenAnchor, Insets screenPadding, AnchorPoint stageAnchor, AnchorPolicy);
    public void relocate(Screen, AnchorPoint screenAnchor, Insets screenPadding, AnchorPoint stageAnchor, AnchorPolicy);
}
```

## AnchorPoint
`AnchorPoint` is a point that is either specified in absolute coordinates, or relative to the screen or stage:
```java
var anchor1 = AnchorPoint.proportional(0.5, 0.5); // center of the screen/stage
var anchor2 = AnchorPoint.absolute(100, 100); // absolute coordinates within screen/stage
```

For example, a stage that sits flush with the bottom-right corner of the screen can be shown as follows:
```java
var screenAnchor = AnchorPoint.proportional(1, 1); // or use the AnchorPoint.BOTTOM_RIGHT constant
var stageAnchor = AnchorPoint.proportional(1, 1);
stage.relocate(screenAnchor, stageAnchor);
stage.show();
```

## AnchorPolicy
`AnchorPolicy` controls how the anchor may be adjusted when the preferred placement doesn't fit within the screen bounds:

```java
public enum AnchorPolicy {
    FIXED,
    FLIP_HORIZONTAL,
    FLIP_VERTICAL,
    AUTO
}
```

* `FIXED`: always use the provided anchor; only adjust the resulting position to fit within the screen.
* `FLIP_HORIZONTAL`: if the preferred placement violates horizontal constraints, try a horizontally flipped anchor (e.g. top-left to top-right) before falling back to the original anchor.
* `FLIP_VERTICAL`: likewise for vertical constraints.
* `AUTO`: automatically choose the most suitable flip:
   if only horizontal constraints are violated, acts like `FLIP_HORIZONTAL`;
   if only vertical constraints are violated, acts like `FLIP_VERTICAL`;
   if both are violated, try a diagonally flipped anchor (both axes) and pick the placement that requires the least adjustment.

This is useful for popup-like behavior where you have a preferred "opening direction", but want the window to flip to the opposite side of the reference point when there isn’t enough space (e.g. "prefer below, but open above if below doesn’t fit").

### PopupWindow support
The new `PopupWindow.anchorPolicy` property adds the same "flip the anchor when it would go off-screen" capability for popups. This complements the existing `PopupWindow.anchorLocation` / `PopupWindow.autoFix` behavior: applications can keep using `PopupWindow.anchorLocation` to express the preferred anchor, while `PopupWindow.anchorPolicy` controls how the popup may flip that anchor as a fallback when the preferred placement wouldn't fit within screen bounds.

## Edge constraints
`screenPadding` specifies per-edge padding relative to the screen, expressed as an `Insets`:

* An inset value >= 0 enables a constraint for that edge and specifies the minimum distance in pixels that the stage must keep from that screen edge.
* An inset value < 0 disables the constraint for that edge.

Enabled constraints effectively shrink the usable screen area by the given insets.

---------
### Progress
- [ ] Change requires a CSR request matching fixVersion jfx27 to be approved (needs to be created)

### Issue
 * [JDK-8372530](https://bugs.openjdk.org/browse/JDK-8372530): Easier placement of windows with positioning anchor (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/1986/head:pull/1986` \
`$ git checkout pull/1986`

`$ git checkout pull/1986` \
`$ git pull https://git.openjdk.org/jfx.git pull/1986/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 1986`

`$ git pr show -t 1986`

</details>

<a href="https://git.openjdk.org/jfx/pull/1986.diff">https://git.openjdk.org/jfx/pull/1986.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/1986#issuecomment-3634799264)
</details>


modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
```diff
@@ -1315,4 +1380,160 @@ private void setPrefHeaderButtonHeight(double height) {
...
+     *       {@code (0,0)} is the top-left corner and {@code (1,1)} is the bottom-right corner.
+     *       Example: {@code (0.5, 0.5)} anchors the center of the window.
+     *   <li><b>Absolute</b>: {@code x} and {@code y} are pixel offsets from the window's top-left corner.
+     *       Example: {@code (0, 0)} anchors the top-left corner; {@code (10, 10)} anchors a point 10px
+     *       right and 10px down from the top-left.
+     * </ul>
+     *
+     * @since 26
+     */
+    public static final class Anchor {
```
Maran23: This could be a record too

modules/javafx.graphics/src/main/java/javafx/stage/Window.java
```diff
@@ -1117,6 +1117,9 @@ public final ObjectProperty<EventHandler<WindowEvent>> onHiddenProperty() {
                                                            0, 0);
                     }
 
+                    // Give subclasses a chance to adjust the window bounds
```
Maran23: Unrelated to this Enhancement, but I wonder if the `centerOnScreen` line before could also be rewritten using this system (at one point). Now, it looks like he might be centering the `Window`, and then fix the bounds to something else, wasting some time.

modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
mstr2: Yes it could, but I don't like the boolean parameter that it would expose. Boolean parameters are usually not a good idea, as they lead to "boolean blindness" at the call site. I've moved this entire class to `javafx.geometry`, as it might be useful for other code in the future, and I don't want to proliferate two nested versions of an anchor point (one in `Stage`, the other in `PopupWindow`).

modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
```diff
@@ -297,6 +302,76 @@ public Stage(@NamedArg(value="style", defaultValue="DECORATED") StageStyle style
...
+     * Calling this method is equivalent to calling
+     * {@code show(anchorX, anchorY, anchor, ClampPolicy.BOTH, Insets.EMPTY)}.
+     *
+     * @param anchorX the requested horizontal location of the anchor point on the screen
+     * @param anchorY the requested vertical location of the anchor point on the screen
+     * @param anchor the point on the stage that should coincide with {@code (anchorX, anchorY)} on the screen
+     * @throws NullPointerException if {@code anchor} is {@code null}
+     * @since 26
+     */
+    public final void show(double anchorX, double anchorY, AnchorPoint anchor) {
```
andy-goryachev-oracle: would it make more sense to move these new methods to `Window`, so they can also work for the `Popup` and its hierarchy ?

mstr2: `PopupWindow` works differently, it has a `anchorLocation` property which is a bit limited. We also couldn't move `PopupWindow.AnchorLocation` anywhere else (as this would be a breaking change), which would make the signature of the show() method very awkward: On `Stage.show(x, y, PopupWindow.AnchorLocation)` you'd have a parameter that is declared on an unrelated class.

andy-goryachev-oracle: Yeah, this is unfortunate.
Can we do anything to make the positioning of Popups and its hierarchy easier?  Maybe add another `show()` with a `ClampPolicy` argument?

Use case: I want to open a PopupWindow below the owner node, but: if the popup cannot be placed there because the whole thing is too close to the bottom of the screen, I want the popup to be **above** the node, and not on top of it as it works currently.

modules/javafx.graphics/src/main/java/javafx/stage/ClampPolicy.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Specifies how a {@link Stage} should be clamped to the screen bounds.
+ * <p>
+ * Clamping adjusts the computed window position so that the window is shown within the screen bounds.
+ * Clamping can be applied independently on the horizontal axis, the vertical axis, both axes, or not at all.
+ *
+ * @since 26
+ */
+public enum ClampPolicy {
```
andy-goryachev-oracle: maybe call it `RelocationPolicy`?

Maran23:
I will test and review this. Especially test with the scenarios where I used the workaround described in the mailing list. 
And give feedback accordingly (API and Functionality-wise).
Will probably take a bit, as there are some other things (like Direct3D12) on my list.

modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
mstr2: I've redesigned the feature a bit, and found that we can indeed make this compatible with `PopupWindow`. You now have the option to specify `PopupWindow.anchorPolicy`, which defaults to the current behavior, but can also be specified to work as you describe (using `AnchorPolicy.AUTO`).

modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
```diff
@@ -1213,6 +1217,88 @@ public void toBack() {
...
+        if (isShowing()) {
+            request.accept(this);
+        } else {
+            this.relocationRequest = request;
+        }
+    }
+
+    @Override
+    public void centerOnScreen() {
+        relocationRequest = null; // cancel previous relocation request
```
Maran23: As far as I understand, with the new anchor system, a center on screen can easily be achieved as well? I wonder if we could also deprecate this method instead. But not sure if I understand it correctly.

mstr2: No, the anchor is always a point _on the window being shown_, not on the screen. All anchor APIs require you to specify an explicit screen location, and the window is then moved so that the anchor coincides with that screen location. So to emulate `centerOnScreen()`, you'd have to manually get the screen bounds, and then calculate the center.

andy-goryachev-oracle: Is there something that prevents us to invent a new anchor type?

mstr2: Maybe we can specify two anchors:
```java
void Stage.relocate(AnchorPoint screenAnchor, AnchorPoint stageAnchor);
```

Then, if you want to center the window on the screen, you can do:
```java
myStage.relocate(AnchorPoint.proportional(0.5, 0.5), AnchorPoint.proportional(0.5, 0.5));
// or myStage.relocate(AnchorPoint.CENTER, AnchorPoint.CENTER);
```

andy-goryachev-oracle: I like it, because we can easily cover almost all the possible cases.  

Should we also allow to specify the `Screen`?  In a third parameter to this extended method?  And if the `Screen` is `null`, the logic would use the default screen or the owner's one?  What do you think?

mstr2: I've added another overload that accepts a `Screen`. The overloads _without_ the screen parameter will always refer to the _current_ screen. This allows users to relocate a stage on its current screen as if it was the primary screen (i.e. with `(0, 0)` being the top-left corner).

modules/javafx.graphics/src/main/java/javafx/stage/Window.java
mstr2: I've re-implemented `centerOnScreen` in terms of `WindowLocationAlgorithm`, so it now uses the same code paths as the `relocate()` methods.

modules/javafx.graphics/src/main/java/com/sun/javafx/stage/WindowRelocator.java
```diff
@@ -0,0 +1,327 @@
...
+                windowWidth, windowHeight,
+                stageAnchor, anchorPolicy,
+                screenBounds, screenPadding);
+
+            if (stageAnchor.isProportional()) {
+                gravityX = stageAnchor.getX();
+                gravityY = stageAnchor.getY();
+            } else {
+                gravityX = stageAnchor.getX() / windowWidth;
+                gravityY = stageAnchor.getY() / windowHeight;
```
andy-goryachev-oracle, line 105: It might be a good idea to explain what "gravity" is, maybe in `ComputedLocation` or `WindowLocationAlgorithm`.

Here, for example, L101 tells us the gravity is measured in pixels, while L104 indicates the gravity is dimensionless.

modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
```diff
@@ -1213,6 +1218,127 @@ public void toBack() {
...
+     * @param stageAnchor   An anchor point in absolute or proportional stage coordinates.
+     * @param anchorPolicy  Controls whether an alternative stage anchor may be used when the preferred anchor would
+     *                      place the stage outside the usable screen area. Depending on the policy, the preferred
+     *                      anchor location may be mirrored across the vertical/horizontal center line of the stage,
+     *                      or an anchor might be selected automatically. If no alternative anchor yields a better
+     *                      placement, the specified {@code stageAnchor} is used.
+     * @throws NullPointerException if any of the parameters is {@code null}
+     * @since 26
+     */
+    public final void relocate(Screen screen, AnchorPoint screenAnchor, Insets screenPadding,
```
andy-goryachev-oracle, line 1322: would it be conceptually simpler to offer just one new method with all the possible parameters (screen anchors, insets) instead of several?  I presume these new methods will be rarely used.
or, if there is one frequent case, provide an additional utility method for that case?

andy-goryachev-oracle:
some preliminary comments, I'll test it more in the next couple of days.

modules/javafx.graphics/src/main/java/javafx/stage/AnchorPolicy.java
```diff
@@ -0,0 +1,91 @@
...
+public enum AnchorPolicy {
+
+    /**
+     * Always use the preferred anchor and never select an alternative anchor.
+     * <p>
+     * If the preferred anchor places the window outside the usable screen area, the window position is
+     * adjusted for the window to fall within the usable screen area. If this is not possible, the window
+     * is biased towards the edge that is closer to the anchor.
+     */
+    FIXED,
```
andy-goryachev-oracle, line 52: should we mention that this one corresponds to the default behavior?

modules/javafx.graphics/src/main/java/javafx/stage/AnchorPolicy.java
```diff
@@ -0,0 +1,91 @@
...
+     *   <li>If only horizontal constraints are violated, it behaves like {@link #FLIP_HORIZONTAL}.
+     *   <li>If only vertical constraints are violated, it behaves like {@link #FLIP_VERTICAL}.
+     *   <li>If both horizontal and vertical constraints are violated, it attempts a diagonal flip,
+     *       then a horizontal flip, and finally a vertical flip.
+     * </ul>
+     * If no alternative anchor yields a better placement, the original anchor is used and the final
+     * position is adjusted for the window to fall within the usable screen area.
+     * If this is not possible, the window is biased towards the edge that is closer to the anchor.
+     */
+    AUTO
```
andy-goryachev-oracle, line 90: Is there a need for something like NONE, where the dev wants to force a particular behavior regardless of the constraints, even if the popup spans the screens?
(I don't suppose we do, but wanted to ask)

modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
```diff
@@ -631,6 +636,32 @@ public final ReadOnlyDoubleProperty anchorYProperty() {
         return anchorY.getReadOnlyProperty();
     }
 
+    /**
+     * Controls whether an alternative anchor location may be used when the preferred
+     * {@link #anchorLocationProperty() anchorLocation} would place the popup window outside the screen bounds.
+     * Depending on the policy, the preferred anchor location may be mirrored to the other side of the window
+     * horizontally or vertically, or an anchor location may be selected automatically.
+     * <p>
+     * If no alternative anchor location yields a better placement, the specified {@code anchorLocation} is used.
```
andy-goryachev-oracle, line 645: what happens when the policy is set to `null`?

andy-goryachev-oracle (commented):
So far looks very good.  I'll continue testing tomorrow.

FYI:
I've added new options to the Popup/Stage pages in the standalone Monkey Tester on a separate branch:

https://github.com/andy-goryachev-oracle/MonkeyTest/tree/window.relocation

modules/javafx.graphics/src/main/java/javafx/geometry/AnchorPoint.java
```diff
@@ -0,0 +1,192 @@
...
+     * <p>
+     * This constant is equivalent to {@code AnchorPoint.proportional(1, 1)}.
+     */
+    public static final AnchorPoint BOTTOM_RIGHT = new AnchorPoint(1, 1, true);
+
+    private final double x;
+    private final double y;
+    private final boolean proportional;
+
+    private AnchorPoint(double x, double y, boolean proportional) {
```
andy-goryachev-oracle, line 112: Do we want to specify what happens when numerical values are NaN?
I suppose the infinities are ok, since they will be clamped accordingly.

modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
```diff
@@ -631,6 +636,32 @@ public final ReadOnlyDoubleProperty anchorYProperty() {
...
+     *
+     * @defaultValue {@link AnchorPolicy#FIXED}
+     * @since 26
+     */
+    private final ObjectProperty<AnchorPolicy> anchorPolicy =
+            new SimpleObjectProperty<>(this, "anchorPolicy", AnchorPolicy.FIXED);
```
andy-goryachev-oracle, line 651: perhaps this property should be lazily created?

(I wish Java would have added the `lazy` keyword:
```
private final lazy ObProp anchorPolicy = new SimObProp(...);
```
)

andy-goryachev-oracle (approved):
Looks good!

I wonder if it would make sense to create an exhaustive test for all possible combinations of parameters in relocation policy.  For example, the test cases can be generated programmatically, the combination and the outcome can be captured to a CSV file to be used as a source file for the actual unit test.
Just a suggestion, I think the current tests are probably sufficient.

andy-goryachev-oracle:
oh no, we let it expire...  I still think it's a useful change.

andy-goryachev-oracle:
@mstr2 do you want to reopen this PR?  I think it'll be very useful.

Maran23:
I agree, this is probably a good candidate for JavaFX 28. Was also on my list to do more testing and replace some calculations I implemented on my own to mimic this feature.

PR #2006, opened 2025-12-12
8373452: DataFormat threading and API issues

andy-goryachev-oracle:
### Summary

This PR makes deprecates the `DataFormat` constructor (for removal):
```
public DataFormat(@NamedArg("ids") String... ids)
```

and replaces it with
```
public static DataFormat of(String ... ids)
```

### Problem

There seems to be several issues with DataFormat API and implementation discovered during a Clipboard-related code review:

1. `static DataFormat::lookupMimeType(String)` is not thread safe: while iterating over previously registered entries in the `DATA_FORMAT_LIST` another thread might create a new instance (DataFormat L227)

2. `public DataFormat(String...)` constructor might throw an `IllegalArgumentException` if one of the given mime types is already assigned to another `DataFormat`. The origin of this requirement is unclear, but one possible issue I can see is if the application has two libraries that both attempt to create a `DataFormat` for let's say `"text/css"`. Then, depending on the timing or the exact code path, an exception will be thrown for which the library(-ies) might not be prepared. The constructor is also not thread safe.

3. To avoid a situation mentioned in bullet 2, a developer would is typically call `lookupMimeType()` to obtain an already registered instance, followed by a constructor call if such an instance has not been found. An example of such code can be seen in webkit/UIClientImpl:299 - but even then, despite that two-step process being synchronized, the code might still fail if *some other* library or the application attempts to create a new instance of DataFormat, since the constructor itself is not synchronized.

4. `DataFormat(new String[] { null })` is allowed but makes no sense!

Why do we need to have the registry of previously created instances? Unclear. My theory is that the DataFormat allows to have multiple mime-types (ids) - example being `DataFormat.FILES = new DataFormat("application/x-java-file-list", "java.file-list");` - and the registry was added to prevent creation of a `DataFormat` with just one id for some reason.

What should be done?
- find out why we need this registry in the first place i.e. what could happen if we have multiple DataFormat instances with overlapping ids.
- if the registry is needed add a new factory method, something like `DataFormat::of(String ...)` which is properly synchronized. This method will be called by the constructor to retain the backward compatibility.
- deprecate (possibly for removal) `DataFormat::lookupMimeType(String)`, or keep it but have it properly synchronized

### Dangers

1. adding synchronization might lead to deadlocks if the application or library has existing code synchronized around some other object and not `DataFormat.class`.

### Alternatives

We could possibly prevent the application code creating `DataFormat`s with multiple ids by 
- creating a `public DataFormat(String)` constructor
- allowing multiple instances with the same id by removing the registry or using the registry only for `DataFormat`s with multiple ids

/reviewers 2
/csr

---------

---------
### Progress
- \[ \] Change requires a CSR request matching fixVersion jfx27 to be approved (needs to be created)

### Issue
 * [JDK-8373452](https://bugs.openjdk.org/browse/JDK-8373452): DataFormat threading and API issues (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2006/head:pull/2006` \
`$ git checkout pull/2006`

`$ git checkout pull/2006` \
`$ git pull https://git.openjdk.org/jfx.git pull/2006/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2006`

`$ git pr show -t 2006`

</details>

<a href="https://git.openjdk.org/jfx/pull/2006.diff">https://git.openjdk.org/jfx/pull/2006.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2006#issuecomment-3774745580)
</details>


modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
```diff
@@ -124,27 +120,29 @@ public class DataFormat {
...
-                if (lookupMimeType(id) != null) {
-                    throw new IllegalArgumentException("DataFormat '" + id +
-                            "' already exists.");
-                }
-            }
-            this.identifier = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(ids)));
-        } else {
-            this.identifier = Collections.<String>emptySet();
+    @Deprecated(since = "27")
+    private DataFormat(@NamedArg("ids") String... ids) {
```
kevinrushforth: This is an incompatible change. You will need to go through the usual process for such changes where it is deprecated for removal in one release and removed in a subsequent release.

kevinrushforth (commented):
Removing the constructor is an incompatible change.

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
andy-goryachev-oracle: should we also print a warning to `stderr`?

andy-goryachev-oracle: resolved offline: no, since stderr will likely be noticed by the user rather than the developer.

andy-goryachev-oracle:
@nlisker or @mstr2 could you take a look at this please?

/keepalive

crschnick:
Does this PR also fix https://mail.openjdk.org/archives/list/openjfx-dev@openjdk.org/thread/53OR5TWXLJKMUMYDYXQV4TJNFLNMYCR5/ ? If not, you could add this problem to the list of issues as well

andy-goryachev-oracle:
@crschnick : the stack trace you referred does not look related.  Does it have a JBS ticket?  If not, could you come up with a reproducer or steps to reproduce, and file a bug please?

crschnick:
I haven't filed one yet because I'm not sure yet which part of the issue is the actual bug. I also can't reproduce it yet. If you look at the mailing list reply of @mstr2, there is a reentrant issue when calling mimesFromSystem as it calls gtk_main_iteration(). This means that when mimesFromSystem is called in an event handler, the gtk_main_iteration can send other events while we are still in the first event handler. In some cases this leads to an exception, but will happen much more often without exceptions (although this is also probably wrong).

mstr2:
I've been staring at this code, scratching my head and wondering what is going on.

First of all, the global `DataFormat` registry does indeed expose some real problems. However, replacing the constructor with an interned `DataFormat.of(String...)` does not address the underlying design problem. It seems to me that the main question here is not whether `DataFormat` should be created with a constructor or a factory method, the main question is what `DataFormat` is supposed to mean.

The proposed `DataFormat.of(...)` still requires a global ownership model for identifiers. That model is ambiguous for overlapping formats:
```java
DataFormat a = DataFormat.of("text/foo");
DataFormat b = DataFormat.of("text/foo", "text/bar");
```

There is no universally correct interned result here:
* returning `a` loses the "text/bar" information
* returning `b` changes what earlier callers meant by "text/foo"
* throwing keeps the current failure mode
* mutating `a` to include "text/bar" would break immutability and hash-based collections

So the factory does not actually solve the semantic problem. It also does not remove the need to support existing public API, as both `DataFormat(String...)` and `DataFormat.lookupMimeType(String)` are already public.

The current implementation relies too much on identity and registry lookup. For example, `QuantumClipboard` maps native MIME strings back through `DataFormat.lookupMimeType(...)`, which makes clipboard behavior depend on unrelated application initialization order.

I think we should treat `DataFormat` primarily as an immutable value descriptor: `new DataFormat("text/foo", "text/bar")` means that this format can be represented by any of the specified identifiers.

Then we define another operation, let's call it `boolean DataFormat.matches(String)`. This method returns `true` if any of the identifiers is equal to the given string. This is the question that clipboard code _actually_ needs to answer. (Note that this operation is not strictly required, as the identifiers are already exposed with `DataFormat.getIdentifiers()`, so anyone can answer the question that a `matches` method would answer.)

`Clipboard.getContentTypes()` should report what the clipboard _actually_ advertises, not an arbitrary alias group that an application may have registered globally. If the native clipboard reports only `text/foo` then `getContentTypes()` can reasonably return `Set.of(new DataFormat("text/foo"))`.

But this must still work:
```java
clipboard.hasContent(new DataFormat("text/foo", "text/bar")) // true
clipboard.getContent(new DataFormat("text/foo", "text/bar")) // reads text/foo
```
These methods must be implemented in terms of `matches`. The clipboard implementation should be changed as follows:
* `getContentTypes()` discovers the actual identifiers on the clipboard (for example "text/foo")
* `hasContent(format)` and `getContent(format)` match a user-provided `DataFormat`, which may be an alias group for one of the identifiers on the clipboard (for example ["text/foo", "text/bar"])
* `QuantumClipboard` should stop using `lookupMimeType()` to decide whether content exists
* `QuantumClipboard` should stop depending on identity comparisons for built-in formats like image, URL, files, RTF, etc.

This would fix the synchronization issue without relying on global interning.

nlisker:
My thoughts are similar to @mstr2's.

First, I'm not confident the implementation of `DataFormat` is compatible with its use. The class is used in conjunction with `Clipboard`/`Dragboard` as a marker for the type of data it contains.
Because `DataFormat` is just a thin wrapper of a `Set<String>` (where the strings are identifiers/mime types), and the content is populated through a `Map<DataFormat, Object>`, it essentially serves as an intermediary between the ids and the value. That is, the clipboard is essentially `ids -> value`.
Likewise, the content is retrieved by `DataFormat`s, not by the ids they contain. For example,
```java
var df = new DataFormat("text/a", "text/b");
var cc = new ClipboardContent();
cc.put(df, "value");
clipboard.setContent(cc);
// ... and then
clipboard.getContent(df); // "value"
```
There is no direct way to ask the clipboard specifically for a `"text/a"` despite having matching content:
```java
var df2 = new DataFormat("text/a");
clipboard.get(df2); // null
```
This does not strike me as a correct behavior to begin with. The receiving application is interested in identifiers, not aggregates of identifiers: `getContent(String id)`. If the receiving application is interested in `"text/plain"`, but the publishing application wanted to provide more information about the text and used `new DataFormat("text/plain", "text/rtf")`, the content will not be found. As Michael said, a manual lookup for an id in all data formats is possible.

What a publishing application can do in this case is use 1 `DataFormat` per id: put both `new DataFormat("text/plain")` and `new DataFormat("text/rtf")` in the clipboard content with the same value. Indeed, the docs of `Clipboard` say
> The `Clipboard` operates on the concept of having a single conceptual item on the clipboard at any one time -- though it may be placed on the clipboard in different formats.

and give the example
```java
content.putString("Some text");
content.putHtml("<b>Some</b> text");
```
so that there are 2 `DataFormat`s, not a single one aggregating the plain and html ids. But then what purpose does `DataFormat` serve?

As for the multiplicity problem - the clipboard could contain more than one value per identifier:
```java
var df1 = new DataFormat("text/a", "text/b");
var df2 = new DataFormat("text/a", "text/c");
var cc = new ClipboardContent();
cc.put(df1, "value1");
cc.put(df2, "value2");
clipboard.setContent(cc);
```
and what should an application interested in `"text/a"` find, `"value1"` or `"value2"`? Perhaps this is why there can't exist more than 1 association of an identifier with a `DataFormat` and the global registry is needed (although `WeakReferenceQueue` seems like a poor choice for it).

My conclusion is that the content should actually be represented as `Map<String, Object>` where the string is the identifier. This eliminates the problem of multiplicity by virtue of the `Map` contract: the user can then decide to solve multiplicity either by "last entry wins" via standard `put` or by "first entry wins" via `putIfAbsent`.
The threading issue can be eliminated by using a `ConcurrentMap`, but it shouldn't be about concurrently registering an identifier with a `DataFormat`, but rather about populating the clipboard concurrently. The synchronization should be done at a different level, in my opinion.

andy-goryachev-oracle:
I agree with your sentiment - the current design is the issue.  This PR does not address the design issue but resolves an exception, while keeping things backward-compatible mostly.

If we set to fix the design issue, I agree with @mstr2 that the semantics should be around `matches`, possibly choosing a better match.

For example, if the application defines `DataFormat APP_FILES = new DataFormat("application/x-java-file-list")` it should prefer that over `DataFormat.FILES` in the clipboard operations when a file list is present in the clipboard.  At the same time, asking for `DataFormat.FILES` should also work since it "matches" the content.

This way, instead of new API we'll have a new behavior which makes more sense. 

I am going to take it back to draft.  Thank you @mstr2 @nlisker for a good discussion!

nlisker:
Looking at the discussion again, I'm still not sure why `DataFormat` is needed at all (except for backwards compatibility). The behaviors that @mstr2 suggests are just `Map` methods. `DataFormat` ends up obscuring this through aggregation and then the new behavior unpacks this aggregation, so why aggregate to begin with?

If the clipboard's content is `Map<String, Object>` then

> * `getContentTypes()` discovers the actual identifiers on the clipboard (for example "text/foo")

is `Map#keySet()` and

> * `hasContent(format)` and `getContent(format)` match a user-provided `DataFormat`, which may be an alias group for one of the identifiers on the clipboard (for example ["text/foo", "text/bar"])

are `Map#containsKey(String)` and `Map#get(String)`.

Then we also gain the various put/compute/merge methods from `Map`.

Is `DataFormat` an OS-level concept?

andy-goryachev-oracle:
> I'm still not sure why `DataFormat` is needed at all (except for backwards compatibility).

I think there might have been two reasons: 
1. it's a similar concept to AWT `DataFlavor`
2. they wanted to create an alias between two mime types for `DataFormat.FILES`

On the platform level, it's ultimately tied to a mime type.  Which means, as @mstr2 correctly pointed out, that in case of the alias, the semantics should be either-or in respect to mime type.  In other words, if the clipboard contains "application/x-java-file-list" data, it should work as if it contains `DataFormat.FILES` data.

So we could fix the problem at the implementation level, without introducing new APIs.

I am going to close this PR and submit an alternative solution in the next few days.

PR #2015, opened 2025-12-19
8375588: Enhanced property metadata

mstr2:
Implementation of [enhanced property metadata](https://gist.github.com/mstr2/2fec0303fc440b8eaeb126befc76eb5c).

### New API
This PR includes the following API additions:

1. `ReadOnlyProperty.getDeclaringClass()` and its default implementation.
2. The `javafx.beans.property.AttachedProperty` interface.
3. New constructors for all `Simple<*>Property` and `ReadOnly<*>Wrapper` classes, accepting the declaring class of the property.

The declaring class is stored in a new field in the `Simple<*>Property` classes. If a legacy constructor is used that doesn't specify the declaring class, the `ReadOnlyProperty.getDeclaringClass()` default implementation is called the first time the `Simple<*>Property.getDeclaringClass()` method is called, and its result is stored for future retrieval.

### Testing
For testing, this PR also includes the `test.util.property.PropertyMetadataVerifier` tool. It systematically tests all public and protected properties of a class, and ensures conformance to the following rules:
* `ReadOnlyProperty.getBean()` returns the object instance of the enclosing class, or the target object instance if the property is an attached property.
*  `ReadOnlyProperty.getName()` returns the name of the property, which must correspond to the name of the property getter (excluding the word "Property").
* `ReadOnlyProperty.getDeclaringClass()` returns the enclosing class of the property getter.
* The declaring class of a `Simple<*>Property` or `ReadOnly<*>Wrapper` must be specified in the constructor, not resolved at runtime.
* `getBean()`, `getName()`, and `getDeclaringClass()` must not be overridden in subclasses of `Simple<*>Property` or `ReadOnly<*>Wrapper`.
* An instance property does not implement `AttachedProperty`.
* An instance property has a parameterless property getter.
* An attached property implements `AttachedProperty`.
* An attached property has a static single-argument property getter that accepts the target object.
* `AttachedProperty.getTargetClass()` returns the class of the single parameter of the static property getter.
* A property getter does not return an instance of `ReadOnly<*>Wrapper`, it returns the result of calling `ReadOnly<*>Wrapper.getReadOnlyProperty()`.

Many properties in existing JavaFX classes violate the `PropertyMetadataVerifier` rules in some way or shape. This PR won't address these issues, this will be done in a future cleanup PR.

---------

---------
### Progress

### Issue
 * [JDK-8375588](https://bugs.openjdk.org/browse/JDK-8375588): Enhanced property metadata (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2015/head:pull/2015` \
`$ git checkout pull/2015`

`$ git checkout pull/2015` \
`$ git pull https://git.openjdk.org/jfx.git pull/2015/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2015`

`$ git pr show -t 2015`

</details>

<a href="https://git.openjdk.org/jfx/pull/2015.diff">https://git.openjdk.org/jfx/pull/2015.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2015#issuecomment-3766268620)
</details>


hjohn:
Let me say that I'd really like a mechanism such as this to be part of FX, but I do have some questions:

I think the `getDeclaringClass` on `ReadOnlyProperty` is okayish; however, I'd probably never take the time to override this method with a proper value, mostly because this requires creating a subclass to be efficient.  I know that within JavaFX, creating a subclass for each and every property (resulting in 1000's of subclasses) is "the standard", but for my own use, I never do this as it is just too much boilerplate.

I do however have a bit of problem with `isAttached`; what does this have to do with properties in general?  The description doesn't clarify what it means really, or what it even means outside the very specific case where HBox can "donate" properties to its children.

So to me, the attached part feels out of place for a low-level property API, unless there are compelling reasons and other use cases we'd envision outside of having CSS stylable child properties.  If not, I don't see why we shouldn't make this a method on `Stylable` or something more Node/CSS specific like in this PR: #1714

mstr2:
> I think the `getDeclaringClass` on `ReadOnlyProperty` is okayish; however, I'd probably never take the time to override this method with a proper value, mostly because this requires creating a subclass to be efficient. I know that within JavaFX, creating a subclass for each and every property (resulting in 1000's of subclasses) is "the standard", but for my own use, I never do this as it is just too much boilerplate.

You can also specify it in the constructor of the `Simple*`-properties, along with the bean and name.

> I do however have a bit of problem with `isAttached`; what does this have to do with properties in general? The description doesn't clarify what it means really, or what it even means outside the very specific case where HBox can "donate" properties to its children.
> 
> So to me, the attached part feels out of place for a low-level property API, unless there are compelling reasons and other use cases we'd envision outside of having CSS stylable child properties. If not, I don't see why we shouldn't make this a method on `Stylable` or something more Node/CSS specific like in this PR: #1714

I think attached-ness is a fundamental aspect of properties, and making it visible in the type system is a logical choice. Compare this to WPF or Avalonia, where attached properties are also explicitly modelled. It's a piece of information that you can't reliably get from a property's metadata otherwise.

hjohn:
> > I do however have a bit of problem with `isAttached`; what does this have to do with properties in general? The description doesn't clarify what it means really, or what it even means outside the very specific case where HBox can "donate" properties to its children.
> > So to me, the attached part feels out of place for a low-level property API, unless there are compelling reasons and other use cases we'd envision outside of having CSS stylable child properties. If not, I don't see why we shouldn't make this a method on `Stylable` or something more Node/CSS specific like in this PR: #1714
> 
> I think attached-ness is a fundamental aspect of properties, and making it visible in the type system is a logical choice. Compare this to WPF or Avalonia, where attached properties are also explicitly modelled. It's a piece of information that you can't reliably get from a property's metadata otherwise.

I never used either framework, so "attached" is completely meaningless to me, and the description didn't enlighten me.  What would this mean when I have just a bunch of properties (like a model)?  It seems to me it is giving some extra information related to some kind of parent/child relationship that's outside the definition of the properties themselves.

I'm only looking for a use case where this would be useful if I only have imported `javafx.base` to better understand it, and how it may benefit properties in general.

I did some checking, and this was suggested by AI:
- A `Task` is a simple model with a permanent property (name).
- A `Workflow` defines an attached property `activeInWorkflow`, which exists only while the task is in the workflow.
- `isAttached()` signals that this property is context-dependent, not intrinsic to `Task`.
- Listeners or code can use `isAttached()` to safely determine whether the property is meaningful.

mstr2:
An attached property is a property whose semantics are defined by one class, but whose value is associated with instances of another class. In other words, the declaring class "owns" the meaning, but the associated object owns the value.

For example, `GridPane.columnIndex` doesn't describe anything about the `GridPane`; instead, it describes something about _another object_ as it relates to the `GridPane`. Another example is `HeaderBar.prefButtonHeight`, which is a property of `Stage`, but that property is only meaningful as it relates to `HeaderBar`.

This is not a parent/child relationship and it is not primarily about CSS, it's about meaning and context. For example, you could decorate arbitrary model objects with context-specific state: an entity could be "valid" as it relates to a specific validator or rule, it could be "selected" in a specific context, or it could carry metadata (constraints, tags, flags) that is only meaningful for a particular framework.

JavaFX already has this concept today, but we're not using the `Property` API to model it. Layout constraints like `GridPane.columnIndex` or `HBox.margin` are implemented as static getter/setter pairs. Since they are not first-class properties, you don't get any of the nice property features (bindings, listeners, styling, etc).

If we expose attached properties with the `Property` API and you want to introspect a property, "is this intrinsic to its bean, or is it contextual/constraint-like?" is a fundamental question. The `AttachedProperty` marker interface answers that question.

It's important to note that the attached property model does not come with a scope model. If I set `GridPane.columnIndex` on some child of `GridPane`, it doesn't lose that property just because it's removed from the `GridPane`. That would be a major complication that's also unexpected.

mstr2:
I removed the `ReadOnlyProperty.isAttached()` method, and instead added the `AttachedProperty` marker interface. It can not only be used to test whether a property is an attached property, but also answers the question to which kind of object it can be attached.

andy-goryachev-oracle:
> JavaFX already has this concept today, but we're not using the `Property` API to model it. Layout constraints like `GridPane.columnIndex` or `HBox.margin` are implemented as static getter/setter pairs. Since they are not first-class properties, you don't get any of the nice property features (bindings, listeners, styling, etc).

I always found this design to be weird and unnecessary - these should be the properties/constraints of the container or the layout, not something stored in the child nodes.  What happens when the said child is moved from one container to another??

It might be just me, but I fail to see the value of this proposal.

mstr2:
> I always found this design to be weird and unnecessary - these should be the properties/constraints of the container or the layout, not something stored in the child nodes. What happens when the said child is moved from one container to another??
> 
> It might be just me, but I fail to see the value of this proposal.

This is not a good argument against this proposal, as JavaFX has always had attached properties and they're not going away. This just makes the existing concept more useful by using the `Property` API instead of simple getters and setters.

andy-goryachev-oracle:
Not going away - possibly.  But it does not change the fact that these properties are stored in the wrong place.  Adding some sort of band-aid code to help with some use cases seems a bit wrong to me.  It does not mean that we don't have the actual problem.

Perhaps you could describe a couple of use cases where the new API would shine, use cases where it is impossible to achieve the goal using the existing APIs.

mstr2:
> Not going away - possibly. But it does not change the fact that these properties are stored in the wrong place. Adding some sort of band-aid code to help with some use cases seems a bit wrong to me. It does not mean that we don't have the actual problem.

I would say that they are stored exactly where they belong, in the object which which they are associated. From an implementation standpoint, I agree with you that they are stored in the wrong place, namely in the `Node.getProperties()` map, which is user-accessible. An attached property should not disappear because user code calls `Node.getProperties().clear()`, the storage for these properties should be opaque. This is a defect that I plan to address in a future enhancement.

> 
> Perhaps you could describe a couple of use cases where the new API would shine,

I already did:

>> For example, you could decorate arbitrary model objects with context-specific state: an entity could be "valid" as it relates to a specific validator or rule, it could be "selected" in a specific context, or it could carry metadata (constraints, tags, flags) that is only meaningful for a particular framework.

As far as layout constraints go, they are _not_ associated with the container. It doesn't make sense so say that the row index is a property of the grid; it's a property of an object within the grid. Similarly, it doesn't make sense to say that `HeaderBar.prefButtonHeight` is a property of the header bar; it's a property of the stage that is only relevant in relation to the header bar.

> use cases where it is impossible to achieve the goal using the existing APIs.

Demonstrating impossibility is an unreasonable demand.

In general, formalizing attached properties adds nothing conceptually, it merely straightens out rough edges of what's already there and gives a name to convention.

andy-goryachev-oracle:
Thank you for clarifications!  

I also suspect I did not get my point across about storing the layout properties in the node vs. the layout container.  What happens when the code moves the node with these properties to another similar container?  Whose responsibility it is to clear the unrelated properties and why is this needed in the first place?  If the property (grid index, growth priority) makes no sense outside of the particular container, why is it stored in the node as opposed to let's say a hashtable within the layout container?  This makes no sense to me, but okay, it's grandfathered like any other weird thing in FX.  But why add more weird code on top of weird code?

So I am not convinced: the use case you mentioned is just too abstract, and nothing tells me that it can't be implemented using existing APIs.  What is the problem you are trying to solve, apart from some unnamed "tools"?  These tools can simply query the `Node.getProperties()` for these properties, given the keys are Strings, can't they?

hjohn:
> > JavaFX already has this concept today, but we're not using the `Property` API to model it. Layout constraints like `GridPane.columnIndex` or `HBox.margin` are implemented as static getter/setter pairs. Since they are not first-class properties, you don't get any of the nice property features (bindings, listeners, styling, etc).
> 
> I always found this design to be weird and unnecessary - these should be the properties/constraints of the container or the layout, not something stored in the child nodes. What happens when the said child is moved from one container to another??
> 
> It might be just me, but I fail to see the value of this proposal.

This is basically the JavaFX way to attach layout information to nodes, used by `GridPane`, `HBox`, `StackPane`, etc.  In Swing this is handled differently by making layout constraints a 2nd parameter when adding a node (like `GridBagConstraints`), in FX this is done via the general `Properties` map which is much more flexible.  Extending this system so they become full, observable properties, and even styleable properties is a very logical step.  Making them styleable for example will allow you to move them to CSS:

```css
.button {
    -fx-hbox-margin: 10 10 10 10;
    -fx-hbox-hgrow: ALWAYS;
}
```
This is currently one of the biggest missing features when it comes to styling controls, having to go to programmatic configuration for these kinds of styling properties.

andy-goryachev-oracle:
Thank you @hjohn this example makes sense.

So, if I understand it correctly, the proposal is to add two-tiered styleable properties?  Are there limitations as to which attached property is legal for which node, or any attached property (including custom) can be attached to any node?  I understand that CSS part is not there yet, but perhaps this PR should instead go via `HasAttachedProperties` route?

hjohn:
> Thank you for clarifications!
> 
> I also suspect I did not get my point across about storing the layout properties in the node vs. the layout container. What happens when the code moves the node with these properties to another similar container?

Nothing happens, just like for the past 15 years.  The properties can live on there, until they're needed again when the node becomes part of a layout container.

> Whose responsibility it is to clear the unrelated properties and why is this needed in the first place? 

This falls on the user, as they're the ones setting these properties, if they care (they don't stack up, there is a fixed number of them) and leaving them alone doesn't break anything.  So, there is no need to clean them at all.  It's exceedingly rare that a Node is re-used in such a way that this would matter (ie. moving a Node from a HBox to a GridPane, instead of what usually happens, which is just to recreate the Node).

> If the property (grid index, growth priority) makes no sense outside of the particular container, why is it stored in the node as opposed to let's say a hashtable within the layout container? This makes no sense to me, but okay, it's grandfathered like any other weird thing in FX. But why add more weird code on top of weird code?

Using a hash table actually makes this worse. Reparenting or temporarily removing a node would mean you also have to move or reconstruct its constraints, or lose them entirely. With attached properties, the metadata naturally travels with the node. In practice this case is rare anyway, so it shouldn’t carry much weight in the design discussion.

Swing’s choices predate modern UI frameworks, and its designers were involved in creating JavaFX, so they clearly understood the trade-offs. Since then, almost all modern UI frameworks have converged on the attached-property style, including modern web layout via CSS.

FX already follows this model, which is great as it was conceived over 15 years ago. What’s weird is not that layout metadata lives on the node, but that it’s exposed only through static setters instead of real, styleable, observable properties. Continuing this design would remove some of the weirdness, not add to it.  It would also allow these properties to integrate cleanly with the CSS engine, opening up an entirely new way of controlling layout, where there exists a huge gap now.

mstr2:
> Thank you @hjohn this example makes sense.
> 
> So, if I understand it correctly, the proposal is to add two-tiered styleable properties? Are there limitations as to which attached property is legal for which node, or any attached property (including custom) can be attached to any node?

There can be limitations. For example, `HeaderBar.prefButtonHeight` can only be attached to `Stage`, while layout constraints can only be attached to `Node`. The restriction is indicated by the target parameter of the static accessor, for example:
```java
class HBox {
    public static void setMargin(Node node, Insets value);
}
```

This proposal formalizes a target-type restriction with the `AttachedProperty` interface:
```java
interface AttachedProperty {
    Class<?> getTargetClass();
}
```

If you're given a property instance, this allows you to then query the class of objects for which it can be set.

hjohn:
> So, if I understand it correctly, the proposal is to add two-tiered styleable properties? Are there limitations as to which attached property is legal for which node, or any attached property (including custom) can be attached to any node? I understand that CSS part is not there yet, but perhaps this PR should instead go via `HasAttachedProperties` route?

It's not really two-tiered, or I am unsure what you mean by that.  The properties are just coming from the container as a source, but they are still first class properties in all other respects.

There aren't any limitations what a container could have as an attached property; the properties are container type specific, and other container types cannot accidentally get confused by them -- although we could perhaps share a few common property types among FX's containers; margin might be a good example (other frameworks also do this kind of sharing).  So if you moved a node from GridPane to HBox, you wouldn't need to set the margin again, but may have to adjust container specific properties.

modules/javafx.base/src/test/java/test/util/property/PropertyMetadataVerifier.java
```diff
@@ -0,0 +1,332 @@
...
+                propertyClass.getDeclaredMethod("getDeclaringClass");
+
+                if (isSimplePropertyClass(propertyClass) || isReadOnlyWrapperClass(propertyClass)) {
+                    if (ReflectionUtils.getFieldValue(property, "declaringClass") == null) {
+                        fail("Declaring class of %s must be specified in the constructor of %s".formatted(
+                            propertyInfo.displayName(),
+                            getSimplePropertyOrReadOnlyWrapperClass(originalPropertyClass).getName()));
+                    }
+                } else if (isSimplePropertyOrDerivedClass(propertyClass)) {
+                    fail(("%s#getDeclaringClass() must not be overridden in %s, pass %s "
```
hjohn, line 172: nit: unnecessary parentheses here

modules/javafx.base/src/test/java/test/util/property/PropertyMetadataVerifier.java
```diff
@@ -0,0 +1,332 @@
...
+        }
+
+        if (isSimplePropertyOrDerivedClass(propertyClass)) {
+            do {
+                try {
+                    propertyClass.getDeclaredMethod(methodName);
+
+                    if (isSimplePropertyClass(propertyClass)) {
+                        return;
+                    } else {
```
hjohn: nit: no need for `else`

modules/javafx.base/src/main/java/javafx/beans/property/ReadOnlyBooleanWrapper.java
```diff
@@ -80,6 +80,31 @@ public ReadOnlyBooleanWrapper(Object bean, String name,
         super(bean, name, initialValue);
     }
 
+    /**
+     * The constructor of {@code ReadOnlyBooleanWrapper}
```
hjohn: nit: Here and in a lot of other places, missing punctuation at end of sentence of main javadoc comment.

modules/javafx.base/src/main/java/javafx/beans/property/AttachedProperty.java
```diff
@@ -0,0 +1,49 @@
...
+ * a dynamic test via {@code instanceof AttachedProperty} might be required at runtime to detect its presence.
+ *
+ * @since 27
+ */
+public interface AttachedProperty {
+
+    /**
+     * Returns the class of objects to which this property can be attached.
+     *
+     * @return the target class
```
hjohn, line 66: If this can ever return `null` (pretty sure it can), then I think that should be documented so there is no need to guess.

modules/javafx.base/src/main/java/javafx/beans/property/ReadOnlyProperty.java
```diff
@@ -54,4 +56,53 @@ public interface ReadOnlyProperty<T> extends ObservableValue<T> {
      */
     String getName();
 
+    /**
```
hjohn, line 62: Should we add docs to `getBean()` as well?

If I understand correctly, the declaring class can be say `HBox` for an attached property, and the bean instance would then be a child of the hbox.  The documentation for `getBean` then seems outdated, or could be clarified in light of this?

As it stands, one might think that `getBean().getClass() == getDeclaringClass()` (and it often is).

modules/javafx.base/src/main/java/javafx/beans/property/ReadOnlyBooleanWrapper.java
```diff
@@ -120,5 +147,18 @@ public Object getBean() {
...
+        }
+    }
+
+    private class AttachedReadOnlyPropertyImpl extends ReadOnlyPropertyImpl implements AttachedProperty {
+
+        @Override
+        public Class<?> getTargetClass() {
+            return ((AttachedProperty)ReadOnlyBooleanWrapper.this).getTargetClass();
+        }
     }
```
hjohn, line 157: This is clumsier than I had hoped.

modules/javafx.base/src/main/java/javafx/beans/property/SimpleBooleanProperty.java
```diff
@@ -103,4 +118,33 @@ public SimpleBooleanProperty(Object bean, String name, boolean initialValue) {
...
+     * @param initialValue the initial value
+     * @since 27
+     */
+    public SimpleBooleanProperty(Object bean, Class<?> declaringClass, String name, boolean initialValue) {
+        super(initialValue);
+        this.bean = bean;
+        this.declaringClass = declaringClass;
+        this.name = (name == null) ? DEFAULT_NAME : name;
+    }
 }
```
hjohn, line 145: nit: I see this was already the case for existing, but just saying, I don't like the code duplication in the constructors.

modules/javafx.graphics/src/main/java/javafx/scene/layout/HeaderBar.java
```diff
@@ -230,7 +231,18 @@ public static ObjectProperty<HeaderDragType> dragTypeProperty(Node child) {
             return (ObjectProperty<HeaderDragType>)property;
         }
 
-        var property = new SimpleObjectProperty<HeaderDragType>(child, "dragType");
+        class PropertyImpl extends SimpleObjectProperty<HeaderDragType> implements AttachedProperty {
```
hjohn: This seems to be primarily an infrastructure PR, we should consider to leave these specific changes out.  I think there are many classes that could be adjusted, for which we could create a few grouped PR's.

I also noticed there is basically no users yet of the new getTargetClass method, I assume that was done deliberately as well.

hjohn (commented):
Hm, perhaps now that the idea here has had some time to sink in, perhaps the initial version with attachedness directly specified on the `ReadOnlyProperty` interface was better (also in light of the hoops we now need to jump through in the wrappers).  Unless of course you think this *is* the better solution.

It seems to me that since we're already adding some weight to all the `Simple` property classes, and now additional mini-classes just to implement `AttachedProperty`, a solution where there is one or two default methods on `ReadOnlyProperty` is lighter, even if they may seem out of place (like I said, the idea needed some time to sink in, and there may be a good enough case to have this on simply **all** properties, UI related or not).

Best however to see what other reviewers think.

All in all, I think this is a good infrastructural change to build upon for making all static properties attached (I'm especially excited for `HBox`, `VBox` and `GridPane`), and I assume there will be follow-up PR's (where I could perhaps help) to modify these classes, as well and supporting these to be styleable by CSS.

The included `HeaderBar` changes may be better done separately.  Also are you sure this will work for styleable properties as well?

andy-goryachev-oracle:
two-tiered: I meant one set of normally declared properties, the other set of properties that have a meaning only in the context of some layout container (and they must be cleared when the node is moved to another container).

andy-goryachev-oracle:
What do you think of
```
Set<AttachedProperty> props = AttachedProperty.from(node);
```
without the Node having to implement `HasAttachedProperties` interface?

hjohn:
> two-tiered: I meant one set of normally declared properties, the other set of properties that have a meaning only in the context of some layout container (and they must be cleared when the node is moved to another container).

I think the need for clearing is a bit overblown; currently when you do `HBox.setMargin(node, margin)` nobody is clearing this either when the node is moved, so I don't see this as an urgent requirement now.  Clearing is possible, but it is a user task:  `HBox.setMargin(node, null)`.

andy-goryachev-oracle:
> the need for clearing is a bit overblown;

I agree; my point is that the property belongs in the container, not the child node.  I know - grandfathered, nothing we can do at this point.

hjohn:
> > the need for clearing is a bit overblown;
> 
> I agree; my point is that the property belongs in the container, not the child node. I know - grandfathered, nothing we can do at this point.

The current state of affairs in FX matches much closer with most frameworks created after Swing, having the attached properties be part of the child, not the container.  In other words, FX is the more modern and more flexible design here, and I'm sure the designers also drew upon lessons learned in this regard from Swing.

hjohn:
Perhaps it could help to think of attached properties as simple configuration on the node, not as temporary state owned by a container. You may not know which container a node will end up in, but you can still set things like a margin, or tell it to grow if it’s placed in a VBox. You might do the same for other container types it could live in. When the node is actually added to a parent, that parent just reads the properties it understands and ignores the rest.

From that point of view, it’s hard to justify why a container should remove properties it didn’t set. Auto-removal makes moving a node between parents lossy and surprising, because you lose configuration that was intentionally added. A simpler rule is: keep attached properties on the node, let containers use what applies to them, and only remove a property if the code that added it explicitly does so.

modules/javafx.base/src/test/java/test/util/property/PropertyMetadataVerifier.java
mstr2: They are not unnecessary, they group the string concatenation on which the `formatted()` method is called.

modules/javafx.base/src/main/java/javafx/beans/property/AttachedProperty.java
mstr2: It never returns `null`, as an attached property must have a target class. I mean it can return `null`, but that would be an implementation bug.

modules/javafx.base/src/main/java/javafx/beans/property/ReadOnlyProperty.java
mstr2: `getBean().getClass() == getDeclaringClass()` is only true if the property is declared in the most derived class. For most JavaFX properties, this would not be the case. In any case, I've updated the documentation a bit.

modules/javafx.base/src/main/java/javafx/beans/property/ReadOnlyBooleanWrapper.java
mstr2: I think it's okay as it's a very localized implementation detail. You'll probably forget that it even exists very soon and move on.

modules/javafx.base/src/main/java/javafx/beans/property/SimpleBooleanProperty.java
mstr2: I've changed to telescoping constructor invocation where possible.

modules/javafx.graphics/src/main/java/javafx/scene/layout/HeaderBar.java
mstr2: I included this as a small demonstration for an attached property implementation.
I don't think that there will be users of `getTargetClass()` in the JavaFX framework, at least at the moment (it didn't exist before, so why would there be users).

mstr2:
> Hm, perhaps now that the idea here has had some time to sink in, perhaps the initial version with attachedness directly specified on the `ReadOnlyProperty` interface was better (also in light of the hoops we now need to jump through in the wrappers). Unless of course you think this _is_ the better solution.

I think it is the better solution, as it also provides a place for the `getTargetClass()` API.

> It seems to me that since we're already adding some weight to all the `Simple` property classes, and now additional mini-classes just to implement `AttachedProperty`, a solution where there is one or two default methods on `ReadOnlyProperty` is lighter, even if they may seem out of place (like I said, the idea needed some time to sink in, and there may be a good enough case to have this on simply **all** properties, UI related or not).

We're not adding weight to _all_ simple property classes. With some sensible assumptions (memory alignment is not guaranteed, the VM is free to do anything it wants), the following classes will grow from 40 to 48 bytes:
`SimpleFloatProperty`
`SimpleIntegerProperty`
`SimpleObjectProperty`
`SimpleStringProperty`

The following classes have enough unused space in their layout to accomodate the new field:
`SimpleBooleanProperty` (40)
`SimpleDoubleProperty` (48)
`SimpleLongProperty` (48)
`SimpleListProperty` (56)
`SimpleMapProperty` (56)
`SimpleSetProperty` (56)

> The included `HeaderBar` changes may be better done separately. Also are you sure this will work for styleable properties as well?

I don't see why it wouldn't. This PR doesn't add much functionality in any case. Making this work for styleable properties will require a fair bit of additional work, something along the lines of what you've proposed a few months ago.

mstr2:
> What do you think of
> 
> ```
> Set<AttachedProperty> props = AttachedProperty.from(node);
> ```
> 
> without the Node having to implement `HasAttachedProperties` interface?

Attached properties are not exclusive to `Node`, we already have an attached property for `Stage` in JFX 26 (`HeaderBar.prefButtonHeight`). In any case, having something along the lines of `AttachedProperty.from(node)` would entangle the implementation of `Node` (and other classes) with `AttachedProperty`.

mstr2:
> > the need for clearing is a bit overblown;
> 
> I agree; my point is that the property belongs in the container, not the child node. I know - grandfathered, nothing we can do at this point.

I understand that you don't like the idea of attached properties, and that you'd rather have narrow-scoped information colocated with layout containers. However, I'm glad that the JavaFX designers chose attached properties instead, as I think that this is a superior formalism. We're just finishing what they started.

andy-goryachev-oracle:
> Attached properties are not exclusive to `Node`, we already have an attached property for `Stage` in JFX 26 (`HeaderBar.prefButtonHeight`).

you could also have `.of(Stage)`.

> In any case, having something along the lines of `AttachedProperty.from(node)` would entangle the implementation of `Node` (and other classes) with `AttachedProperty`.

How so?  In fact, it completely separates the feature from whatever carrier is used.

I really dislike this PR as it brings, in my opinion, unnecessary weight and code to the objects used everywhere for what seems to be a small side feature.  I am ok with adding support to it, just not the way this PR does.

I would say `AttachedProperty.of()` or something like that way is probably the least intrusive solution that I would support.

mstr2:
> > Attached properties are not exclusive to `Node`, we already have an attached property for `Stage` in JFX 26 (`HeaderBar.prefButtonHeight`).
> 
> you could also have `.of(Stage)`.

Yes, and that brings in `Stage` into the API of `AttachedProperty`, where it doesn't belong.

But I think I'm not even clear on what you are proposing here. This PR does two things:
1. It adds the `ReadOnlyProperty.getDeclaringClass()` method.
2. It adds the `AttachedProperty` interface, which property classes can implement.

It seems like you don't object to the first addition, but only to the second addition. But what exactly does your proposed code do, then:
```java
Set<AttachedProperty> props = AttachedProperty.from(node);
```

What are these `AttachedProperty` objects that I can get from a node? They can't be the actual `Property` instances, as that's what you're objecting to. What can I do with these objects?

andy-goryachev-oracle:
@mstr2 : I don't understand the use case(s) then.  

@hjohn provided one: setting things like `-fx-hbox-hgrow` via CSS, so I thought that's the kind of property one can stash in the Node.properties.  These properties might provide whatever extra information you want from the attached properties, if I understand all this correctly.

I do not understand why you need to add unrelated and unused fields and methods to the regular property classes that neither need nor use it.  Again, if I understand it correctly.

Does your JEP explain the problem in sufficient detail so even an old timer like me who is perpetually stuck in Swing can understand?

mstr2:
> @mstr2 : I don't understand the use case(s) then.
> 
> @hjohn provided one: setting things like `-fx-hbox-hgrow` via CSS, so I thought that's the kind of property one can stash in the Node.properties. These properties might provide whatever extra information you want from the attached properties, if I understand all this correctly.
> 
> I do not understand why you need to add unrelated and unused fields and methods to the regular property classes that neither need nor use it. Again, if I understand it correctly.
> 
> Does your JEP explain the problem in sufficient detail so even an old timer like me who is perpetually stuck in Swing can understand?

Some enhancements are motivated by missing features, some are motivated by rough edges in a taxonomy. Consider the recent addition of pattern matching for primitives in Java: that's not a feature people were missing sorely, it's just what you get when you smooth out the rough edges.

This PR falls into the second category: it smooths out rough edges in the JavaFX property system. Properties carry metadata to tell you about themselves: their name and the object to which they belong. The declaring class is a missing piece of information (and it's _certainly_ not unrelated, as you say). Whatever criticism you have about why property metadata exists, your criticism should then also be levied against the existing "bean" and "name" metadata. The argument is the same.

mstr2:
I've run into another scenario where having first-class styleable attached properties would make life _much_ easier. I've placed a `TabPane` such that its tabs overlap a `HeaderBar`. In order for the header bar to work correctly, I need to set the `HeaderBar.dragType` property on the tab pane and on its tabs (which are produced by a skin and not easily accessible from JavaFX code).

Actually getting hold of the tabs of a `TabPane` is complicated by the fact that they only exist after the skin has been inflated by CSS; they don't exist after I've typed `var tabPane = new TabPane()` in code. So I need to manually inflate the skin, or defer configuration to the next pulse. After that, I need to use `Node.lookupAll(".tab")` to get the nodes that represent the tabs, or I need to traverse the children of the tab pane until I've found the tabs. Only then can I set the `HeaderBar.dragType` property on those nodes.

If we had styleable attached properties, I could just provide a stylesheet for my tab pane:
```css
.tab-pane {
    -fx-headerbar-dragtype: transparent-subtree;
}

.tab-pane .tab {
    -fx-headerbar-dragtype: none;
}
```

mstr2:
I'd like to conclude the discussion of this enhancement, if possible, for the JFX 27 release. The HeaderBar API, which is on course to be finalized in JFX 27, uses attached properties in the way this enhancement formalizes them (see `dragType`, `buttonType`, `prefButtonType`, `leftSystemInset`, `rightSystemInset`, `minSystemHeight` in `HeaderBar`).

If we _don't_ go forward with this formalism, then the API in HeaderBar should be removed; otherwise it will forever be the odd one out compared to other attached properties (of the non-observable kind). Note that we don't discuss the _existence_ of attached properties here, they've always been an integral part of JavaFX. This enhancement only extends the `Property` API to also be applicable to attached properties.

modules/javafx.base/src/main/java/javafx/beans/property/ReadOnlyProperty.java
```diff
@@ -54,4 +59,54 @@ public interface ReadOnlyProperty<T> extends ObservableValue<T> {
...
+     *           with a signature compatible with {@code ReadOnlyProperty<?> <name>Property()}, where
+     *           {@code <name>} is the name of the property as returned by the {@link #getName()} method.
+     *           The return type must be a subtype of {@code ReadOnlyProperty<?>}. The class that declares
+     *           such a method is the declaring class of the property.
+     *           However, implementations are advised to override this method and return the declaring class
+     *           directly, instead of relying on the reflective auto-discovery mechanism. Attached properties
+     *           must always override this method.
+     * @since 28
+     */
+    default Class<?> getDeclaringClass() {
```
Maran23, line 81: I'm a little worried that this might accidentally reduce performance. 
If developers don't know/forget to set the declaring class and rely on this reflective fallback without knowing it.

But I don't have an idea to solve this right now. Just thinking out loud.

mstr2: All properties in JavaFX will override this method and directly return their declaring class. So this will only impact third-party properties for which 1) the declaring class is queried, and 2) the `getDeclaringClass()` method is not overridden, and the declaring class is not specified in the property constructor.

Also, keep in mind that the declaring class is only resolved once, and then cached for repeated queries. I can't imagine that this will be a performance problem.

Maran23: But only for the `SimpleXXXProperty` classes right now. Not cached for other classes like e.g. `StyleableBooleanProperty`.

Am I understanding you right that another PR will override this method in JavaFX nodes?

mstr2: Yes, a separate PR will override `getDeclaringClass()` for every property implementation in JavaFX.

Aside from that, we might want to add a constructor overload that accepts the declaring class to the SimpleStyleable* property classes as well.

Maran23: Okay that makes sense then, thanks.

mstr2: I've added constructor overloads for all `SimpleStyleable*` classes as well.

modules/javafx.base/src/main/java/javafx/beans/property/ReadOnlyProperty.java
```diff
@@ -54,4 +59,54 @@ public interface ReadOnlyProperty<T> extends ObservableValue<T> {
...
+        String propertyName = name + "Property";
+
+        do {
+            try {
+                Method method = beanClass.getDeclaredMethod(propertyName);
+
+                if ((method.getModifiers() & Modifier.STATIC) == 0
+                        && ReadOnlyProperty.class.isAssignableFrom(method.getReturnType())) {
+                    return beanClass;
+                }
```
Maran23: I was still wondering if it might make sense to log a `warning` or just an `info` that this is really is a fallback implementation and developers should rather override this / set the declaring `Class` instead.

All opinions welcome.

Maran23 (commented):
beside one thing where I'm curious about other opinions, I think this a good way forward.

modules/javafx.base/src/main/java/javafx/beans/property/ReadOnlyProperty.java
Maran23: On a similar note, when the reflection failed, it will always be retried, right? Since we return `null` and so the subclass will call this again the next time

modules/javafx.base/src/main/java/javafx/beans/property/SimpleStringProperty.java
```diff
@@ -58,49 +59,83 @@ public String getName() {
     }
 
     /**
-     * The constructor of {@code StringProperty}
+     * {@inheritDoc}
+     *
+     * @since 28
+     */
+    @Override
+    public Class<?> getDeclaringClass() {
```
Maran23, line 69: Minor question, but since we now add the new API, we should probably think about it.
Should we rather make all `getDeclaringClass()` methods final for all `Simple*Property` classes?

modules/javafx.base/src/main/java/javafx/beans/property/ReadOnlyProperty.java
```diff
@@ -39,10 +41,13 @@
 public interface ReadOnlyProperty<T> extends ObservableValue<T> {
 
     /**
-     * Returns the {@code Object} that contains this property. If this property
-     * is not contained in an {@code Object}, {@code null} is returned.
+     * Returns the {@code Object} that is associated with this property.
+     * <p>
+     * For instance properties, this is the object that contains the property.
+     * For {@linkplain AttachedProperty attached properties}, it is the object to which the property is attached.
+     * Is this property is not associated with an object, {@code null} is returned.
```
Maran23: Typo
```suggestion
     * If this property is not associated with an object, {@code null} is returned.
```

Maran23 (commented):
Two comments regarding API and doc.

modules/javafx.base/src/main/java/javafx/beans/property/ReadOnlyProperty.java
mstr2: I've added code that ensures the reflective discovery is only called once, and a negative result is also cached.
I don't think that a log output is necessary, because a `null` result is a legitimate result when the property isn't declared in a class.

modules/javafx.base/src/main/java/javafx/beans/property/SimpleStringProperty.java
mstr2: Good question. Developers can override `getBean()` and `getName()` even for the Simple* properties (and we probably can't change that). I think we should then also allow that for `getDeclaringClass()` to be consistent.

modules/javafx.base/src/main/java/javafx/beans/property/ReadOnlyProperty.java
```diff
@@ -54,4 +59,54 @@ public interface ReadOnlyProperty<T> extends ObservableValue<T> {
...
+                    return beanClass;
+                }
+            } catch (NoSuchMethodException ignored) {
+                // fall through
+            }
+
+            beanClass = beanClass.getSuperclass();
+        } while (beanClass != null);
+
+        return null;
```
hjohn: Is it perhaps an idea here to use a cache for this? Many properties over the lifetime of an FX application will share the same bean class and name; something like this:

```
private static final ClassValue<ConcurrentHashMap<String, Class<?>>> CACHE = new ClassValue<>() {
    @Override
    protected ConcurrentHashMap<String, Class<?>> computeValue(Class<?> beanClass) {
        return new ConcurrentHashMap<>();
    }
};

static Class<?> lookup(Class<?> beanClass, String propertyName) {
    Class<?> result = CACHE.get(beanClass)
        .computeIfAbsent(propertyName, name -> {
            Class<?> found = findDeclaringClass(beanClass, name);
            return found == null ? NOT_FOUND : found;
        });

    return result != NOT_FOUND ? result : null;
}
```

The `ClassValue` is specifically intended for this kind of use.

modules/javafx.base/src/main/java/javafx/beans/property/SimpleObjectProperty.java
```diff
@@ -61,49 +64,85 @@ public String getName() {
...
+     * The constructor of {@code SimpleObjectProperty}.
+     *
+     * @param bean the bean of this property
+     * @param declaringClass the class in which this property is declared
+     * @param name the name of this property
+     * @since 28
+     */
+    public SimpleObjectProperty(Object bean, Class<?> declaringClass, String name) {
+        this(bean, name);
+        this.declaringClass = declaringClass;
```
hjohn: One could pass in `declaringClass` here as `null`

- Should we allow this?
- If so, what does it mean?
  - `null` means look-up just in time
  - `null` means don't look this up, it is intended to not be specified

modules/javafx.base/src/test/java/test/util/property/PropertyMetadataVerifier.java
```diff
@@ -0,0 +1,332 @@
...
+                || !ReadOnlyProperty.class.isAssignableFrom(method.getReturnType())
+                || method.getParameterCount() >= 2) {
+            return null;
+        }
+
+        Class<?> declaringClass = method.getDeclaringClass();
+        String propertyName = methodName.substring(0, methodName.length() - "Property".length());
+        String displayName = declaringClass.getName() + "." + propertyName;
+
+        return method.getParameterCount() == 1
```
hjohn: nit (test code): This code assumes that a parameter count of 1 means means it is the static accessor, but an instance one could also have a parameter:

ie:

    public Property<String> findKeyedProperty(String key) { ... }

would be found; later it is called with `method.invoke(null, bean)`

hjohn (commented):
I think this looks good, some minor findings

modules/javafx.base/src/main/java/javafx/beans/property/ReadOnlyProperty.java
Maran23: Thats a good question. I understood this code as 'fallback, should usually not run' and then I think we don't need a cache, but Michael probably knows better.

modules/javafx.base/src/main/java/javafx/beans/property/SimpleObjectProperty.java
mstr2: Good question! I see no reason to second-guess the user of this API. If they explicitly specify `null` as the declaring class, then so be it. This can't happen accidentally for unsuspecting code, as these are new constructor overloads.

modules/javafx.base/src/main/java/javafx/beans/property/ReadOnlyProperty.java
mstr2: This is an easy optimization, so I added a class-associated cache.

hjohn (approved):
LGTM

modules/javafx.base/src/main/java/javafx/beans/property/ReadOnlyProperty.java
```diff
@@ -89,24 +89,26 @@ default Class<?> getDeclaringClass() {
             return null;
         }
 
-        Class<?> beanClass = bean.getClass();
-        String propertyName = name + "Property";
+        return DeclaringClassCache.computeIfAbsent(this, name, property -> {
+            Class<?> declaringClass = property.getBean().getClass();
```
Maran23: Minor: Big lambda, I think a method reference would improve the readability quite a bit. Tried locally, something like that works and looks better IMO:

`return DeclaringClassCache.computeIfAbsent(this, name, ReadOnlyProperty::findClassReflectively);`

modules/javafx.base/src/test/java/test/util/property/PropertyMetadataVerifier.java
```diff
@@ -0,0 +1,332 @@
...
+    private static Class<?> getSimplePropertyOrReadOnlyWrapperClass(Class<?> propertyClass) {
+        do {
+            if (isSimplePropertyClass(propertyClass) || isReadOnlyWrapperClass(propertyClass)) {
+                return propertyClass;
+            }
+
+            propertyClass = propertyClass.getSuperclass();
+        } while (propertyClass != null);
+
+        throw new AssertionError();
```
Maran23, line 300: Minor: Should we add a message here?

Maran23 (approved):
I like the idea of the sentinel value. 
It makes my objection/question "Should we log a warning?" much less of an issue. 
So LGTM.

hjohn (approved):

modules/javafx.base/src/test/java/test/util/property/PropertyMetadataVerifier.java
mstr2: I wouldn't overthink this, since this code will never be reached. It's not something you can do wrong when you use this testing tool.

Maran23: Got it.

PR #2041, opened 2026-01-15
8328167: ScrollTo in ListView still doesn't work since VirtualFlow was modified

winrid:
Should fix: https://bugs.openjdk.org/browse/JDK-8328167?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel

I have an application with a TableView that adds items and scrolls to them, and it's very important the last added item is visible to the user. When the first item is added that triggers the scroll, and you do a `platform.runLater -> scrollTo(lastNewItem)` the scrollTo does not work.

I think this is because of something to do with rendering phases, as two nested runLaters do work it seems, but that's pretty naughty.

I tried adding a test for this but I think the way the javafx tests runs differ a bit from an actual app? I am happy to add a test maybe with a recommendation on the approach. I have a reproducer with documentation here: https://github.com/winrid/jfx-jdc-8328167-scrollto-reproducer

---------
### Progress

### Errors

### Issue
 * [JDK-8328167](https://bugs.openjdk.org/browse/JDK-8328167): ScrollTo in ListView still doesn't work since VirtualFlow was modified (**Bug** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2041/head:pull/2041` \
`$ git checkout pull/2041`

`$ git checkout pull/2041` \
`$ git pull https://git.openjdk.org/jfx.git pull/2041/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2041`

`$ git pr show -t 2041`

</details>

<a href="https://git.openjdk.org/jfx/pull/2041.diff">https://git.openjdk.org/jfx/pull/2041.diff</a>

</details>

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
@@ -299,6 +299,9 @@ public class VirtualFlow<T extends IndexedCell> extends Region {
     private boolean sizeChanged = false;
     private final BitSet dirtyCells = new BitSet();
 
+    // Tracks index that needs to be made fully visible after layout
+    private int pendingScrollToIndex = -1;
+
     Timeline sbTouchTimeline;
     KeyFrame sbTouchKF1;
     KeyFrame sbTouchKF2;
@@ -1361,6 +1364,22 @@ public class VirtualFlow<T extends IndexedCell> extends Region {
         lastPosition = getPosition();
         recalculateEstimatedSize();
         cleanPile();
+
+        // Ensure pending scroll target is fully visible after layout.
+        // Only adjust if the cell is cut off at the bottom of the viewport.
+        if (pendingScrollToIndex >= 0) {
+            T cell = getVisibleCell(pendingScrollToIndex);
+            if (cell != null) {
+                double cellPosition = getCellPosition(cell);
+                double cellLength = getCellLength(pendingScrollToIndex);
+                double viewportLength = getViewportLength();
+                // Only adjust if cell extends beyond viewport and is smaller than viewport
+                if (cellPosition + cellLength > viewportLength && cellLength <= viewportLength) {
+                    scrollTo(cell);
+                }
+            }
+            pendingScrollToIndex = -1;
+        }
     }
 
     /**
@@ -1623,6 +1642,9 @@ public class VirtualFlow<T extends IndexedCell> extends Region {
         if (! posSet) {
             adjustPositionToIndex(index);
         }
+
+        // Mark this index to be checked for full visibility after layout
+        pendingScrollToIndex = index;
 
         requestLayout();
     }
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
@@ -300,7 +300,7 @@ public class VirtualFlow<T extends IndexedCell> extends Region {
     private final BitSet dirtyCells = new BitSet();
 
     // Tracks index that needs to be made fully visible after layout
-    private int pendingScrollToIndex = -1;
+    int pendingScrollToIndex = -1;
 
     Timeline sbTouchTimeline;
     KeyFrame sbTouchKF1;
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
@@ -940,6 +940,10 @@ public class VirtualFlow<T extends IndexedCell> extends Region {
 
         @Override protected void invalidated() {
             super.invalidated();
+            // Any change to position supersedes a pending scrollToTop/scrollTo
+            // target — scrollbar drags, wheel, scrollTo(cell), external
+            // setPosition, or property bindings all land here.
+            pendingScrollToIndex = -1;
             adjustAbsoluteOffset();
             requestLayout();
         }
@@ -1590,6 +1594,11 @@ public class VirtualFlow<T extends IndexedCell> extends Region {
 
             adjustPositionToIndex(index);
             addAllToPile();
+
+            // Same offset-estimation caveat as scrollToTop(int): re-verify
+            // full visibility after layout, once real cell sizes are known.
+            pendingScrollToIndex = index;
+
             requestLayout();
         }
     }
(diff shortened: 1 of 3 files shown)


winrid:
I would like to note that I signed the OCA, but I followed the guide in the repo before creating the PR, and that guide does not say to `Please fill in your GitHub username in the "Username" field of the application.`, so I didn't. I guess that means it won't automatically work.

johanvos:
Hi @winrid 
I didn't look into the details yet, but in any case it is really important to have a regression test for this.
A large number of issues in VirtualFlow are caused by changes that fix other issues. They often cause undetected regression, because we don't have enough tests. The VirtualFlow area is used in many different circumstances, where there are often implicit assumptions about how it should behave that are contradicting other assumptions in other circumstances.
Hence, in order to prevent going back and forth, we need regression tests for every change we do, so that future PR's can't undo what is done in this PR.

winrid:
@johanvos For sure, would simply asserting via a stub that `scrollTo` is called when expected be enough? That should catch most regressions. I tried adding a test that actually checks if the element is visible, but it only worked outside of the test harness.

winrid:
@johanvos can you please confirm my testing strategy would work before I do the work? The double platform runlater is still not reliable... so need this fix

johanvos:
> @johanvos For sure, would simply asserting via a stub that `scrollTo` is called when expected be enough? That should catch most regressions. I tried adding a test that actually checks if the element is visible, but it only worked outside of the test harness.

That is a good test indeed. It doesn't test everything though, as the behaviour depends on threading. As a consequence, scrollTo might be invoked twice, but you'll probably assert on the first callback only.

But that is something in general that makes testing VF pretty hard. I believe this needs to be fixed at the root level, so that we can have tests that rely on deterministic behavior, instead of tests that wait "some time" to "make sure all possible side-effects are done".

winrid:
@johanvos added unit test and test that mocks scrollTo and ensures it's called, and red/greened both.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java
```diff
@@ -1361,6 +1364,22 @@ void adjustPosition() {
         lastPosition = getPosition();
         recalculateEstimatedSize();
         cleanPile();
+
+        // Ensure pending scroll target is fully visible after layout.
+        // Only adjust if the cell is cut off at the bottom of the viewport.
+        if (pendingScrollToIndex >= 0) {
```
Maran23, line 1374: One thing I wonder:
1. Is this 100% needed like this, as we may miss information or accuracy due to the cell sizes when we call `scrollToTop(index)`
2. Or is `scrollToTop(index)` broken, concrete: the method `adjustPositionToIndex` in it. So does not calculate the offset correctly?

winrid: @Maran23 it's a race condition. You can see my reproducer above that does 100 tests with and without the fix. Sometimes it *does* work.

Maran23: It would be interesting to know why this is a race. And does it work even without `Platform.runLater` after an item is inserted?

I guess this is indeed a weird case, where the `VirtualFlow` did not yet process the item change (a layout is requested, not done yet) but we already queue in a `scrollToTop` while we can not yet meaningful know the size. 

So this could be a fix, but I'm afraid if we call `scrollToTop`, save the index and then call `scrollTo` with another cell, we basically confuse the `VirtualFlow` to scroll to the index while we already did scroll to somewhere else in the meantime. We probably need to check the other `scrollTo*` methods as well.

winrid: @Maran23 It requires two nested runLaters to work. One is not sufficient. Even then it only works most of the time which is bad.

I'll check the other scroll to methods.

Maran23: Yes. What I mean is that with your fix, we never need a runLater, right?

winrid: @Maran23 Right. My reproducer documents and illustrates this by overriding the javafx classes: https://github.com/winrid/jfx-jdc-8328167-scrollto-reproducer/tree/main/src/main/java/javafx/scene/control/skin

winrid: @Maran23 

>Or is scrollToTop(index) broken, concrete: the method adjustPositionToIndex in it. So does not calculate the offset correctly?

Correct. It uses an estimate, fixing this there would require measuring every preceding cell, which defeats virtualization, right?

This is why the fix runs at the end of layoutChildren(), sizes are known by then.

There is a bug with other scroll actions as you mentioned, fixing.

winrid: @Maran23 pushed

winrid: @Maran23 are you still interested in reviewing this

Maran23: I think overall this might be good. I wish we can do it without this - because it adds more undocumented behavior and unclear specifications. But this is a preexisting problem. That `scrollTo` MAY request a layout is already weird.

Note that you need to sign the OCA still for further review.

winrid: @Maran23 I signed the OCA. Sorry, I don't want to do it again.

Maran23: Can you retype /signed again? In a new comment, maybe that will help

winrid: Done

winrid: @Maran23 also note https://github.com/openjdk/jfx/pull/2041#issuecomment-3839524410

Maran23:
> I would like to note that I signed the OCA, but I followed the guide in the repo before creating the PR...

Checked the FAQ, you need to write or ping the project maintainer here to get this sorted.

winrid:
I will never contribute to openjfx again.

PR #2048, opened 2026-01-21
8375578: Enable translucent window backdrops

beldenfox:
This PR enables translucent window backdrops for JavaFX stages on macOS and Windows 11. Since we’re reliant on the operating system for these effects (they typically require real-time blurring of the desktop) I needed to flesh out a fairly complete prototype to sort out the API. I will start a discussion about the API on the mailing list.

There’s a crude manual test for trying out the different backdrop materials.

	java @build/run.args -Djavafx.enablePreview=true tests/manual/stage/BackdropTest.java

You’ll want to drag the windows around to avoid having them overlap each other since they’re all created in the center of the screen. For windows without title bars you can click anywhere on the background to drag the window except for TRANSPARENT stages on Windows which are a bit tricker to get a hold of; try to click on a text label.

If you create an UNDECORATED stage on Windows the backdrop won’t be translucent initially. This can be corrected by changing the stage’s color scheme. This is an OS bug that I haven’t found a workaround for.

The changes on Windows 11 are minimal since we’re just invoking an OS feature by calling DwmSetWindowAttribute. I did need to make two small changes to the D3D9 Prism code to ensure that the swap chain and back buffer support an alpha channel so JavaFX can composite its content on top of the backdrop. This is the same way the old UNIFIED stage style worked before it became unreliable (see [JDK-8154847](https://bugs.openjdk.org/browse/JDK-8154847)).

On macOS I moved the GlassHostView so it’s now a permanent part of the NSWindow. For some time the host view has been a remnant left over from an older approach to implementing fullscreen. Now it serves as a common parent for the NSVisualEffectView that provides the backdrop and the GlassView3D that contains the JavaFX content. Making it the permanent contentView of the NSWindow simplifies some code.

To validate the API I did prototype this for Windows 10 (thanks @mstr2!). Well, I prototyped this using DirectComposition so it should work on Win10 but I can't test Win10 myself. Using DirectComposition is much more involved so I shelved that implementation for now but it does inform the API. It’s the reason the backdrop needs to be specified before the Java window is shown and the platform window created.

---------

---------
### Progress
- \[ \] Change requires CSR request [JDK-8390690](https://bugs.openjdk.org/browse/JDK-8390690) to be approved

### Issues
 * [JDK-8375578](https://bugs.openjdk.org/browse/JDK-8375578): Enable translucent window backdrops (**Enhancement** - P4)
 * [JDK-8390690](https://bugs.openjdk.org/browse/JDK-8390690): Enable translucent window backdrops (**CSR**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2048/head:pull/2048` \
`$ git checkout pull/2048`

`$ git checkout pull/2048` \
`$ git pull https://git.openjdk.org/jfx.git pull/2048/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2048`

`$ git pr show -t 2048`

</details>

<a href="https://git.openjdk.org/jfx/pull/2048.diff">https://git.openjdk.org/jfx/pull/2048.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2048#issuecomment-3781001757)
</details>


modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
```diff
@@ -567,6 +567,39 @@ public final Window getOwner() {
...
+     *
+     * @param backdrop the backdrop for this stage.
+     *
+     * @throws IllegalStateException if this property is set after the stage
+     * has ever been made visible.
+     *
+     * @defaultValue StageBackdrop.DEFAULT
+     */
+    @SuppressWarnings("deprecation")
+    public final void initBackdrop(StageBackdrop backdrop) {
```
mstr2, line 580: This API has the potential to age poorly, as enums are tricky to evolve. For example, what if a future backdrop type is configurable with a parameter? This could be improved by having the enum implement a sealed interface, so we can later add other backdrop types that can't neatly fit into an enum.

I also wonder how graceful degradation works with this API. It might be useful to be able to model a request like: give me a glass backdrop with a set of parameters, and if the platform doesn't support that, give me a `TRANSIENT` backdrop, and if the platform doesn't support that, I just want an opaque backdrop.

crschnick: Could perhaps the scene preferences be used for things like this? E.g. similar to having access to the platform styling keys, a user could manually override some of them to achieve the same result as with enums. That would solve any potential parameters and the degradation problem better

modules/javafx.graphics/src/main/native-glass/mac/GlassHostView.m
```diff
@@ -26,6 +26,30 @@
...
+    jfxView = view;
+    if (jfxView) {
+        [self addSubview: jfxView positioned: NSWindowAbove relativeTo: backdropView];
+    }
+}
+
+// Called when the window is first created.
+-(void)setBackdrop:(NSVisualEffectMaterial)material
+{
+    NSVisualEffectView* effect = [[NSVisualEffectView alloc] initWithFrame: self.bounds];
```
mstr2, line 51: Would this also work with `NSGlassEffectView` to get the Liquid Glass backdrop effect?

modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
mstr2: I'm not sure about scene preferences, as the supported backdrop materials are not exactly a _preference_. It's not something users can change about the system.

How about something like this:
```java
public sealed interface StageBackdrop {

    enum Generic implements StageBackdrop {
        DEFAULT,
        WINDOW,
        TABBED,
        TRANSIENT
    }

    record Specialized(String name, Map<String, Object> parameters) implements StageBackdrop {
        public Specialized(String name) {
            this(name, Map.of());
        }
    }
}
```

```java
public class Stage {

    public final void initBackdrop(StageBackdrop... backdrop) {
        ...
    }
}
```

The `initBackdrop` method would take a list of options in order of preference. Maybe I could then configure my backdrop like so:
```java
myStage.initBackdrop(
    // first choice
    new StageBackdrop.Specialized(
        "NSGlassEffect",
        Map.of("style", "NSGlassEffectViewStyleClear")
    ),

    // second choice
    StageBackdrop.Generic.TRANSIENT
);
```

modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
```diff
@@ -567,6 +567,39 @@ public final Window getOwner() {
         return owner;
     }
 
+    private StageBackdrop backdrop = StageBackdrop.DEFAULT;
+
+    /**
+     * Specifies the backdrop for this stage. This must be done prior to
+     * making the stage visible.
```
crschnick, line 570: Is there is a reason on why this can't be changed later on? At least in my use cases, changing the backdrop type dynamically later on works without issues

modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
hjohn: Should we have a (new kind of) fallback system, or should we do this with the existing `Platform.isSupported(ConditionalFeature)` and have the user decide based on the result of that function?

beldenfox: I agree that an enum is inadequate. The two API's we have to work with (macOS and Win11) are based on enums; you specify the type of window or content (e.g. tooltip, menu, sidebar, window) and let the OS decide what material to use. macOS pro-actively enumerates a lot of different window types though it looks like many of them map to the same material under the hood. I'm not entirely sure the glass effect recently added to macOS is appropriate for JavaFX but it is a prime example of a specialized platform-specific backdrop that can't easily be squeezed into an enum.

So, yes, we should expand StageBackdrop so we can provide more information. How to do that might be beyond my pay grade (I'm not a Java expert). I don't think the existing `Platform.isSupported()` API would be appropriate since it takes a separate enum type.

mstr2: > Should we have a (new kind of) fallback system, or should we do this with the existing `Platform.isSupported(ConditionalFeature)` and have the user decide based on the result of that function?

Expanding on that question: if we have platform-specific backdrops that may have a big impact on styling and readability, stylesheets may want to query the actual backdrop used via media queries.

modules/javafx.graphics/src/main/native-glass/mac/GlassHostView.m
beldenfox: I'm experimenting with that now. I can get it to work but it only provides a nice effect for transparent stages. And it is an entirely different beast than the other backdrops. For example, it doesn't respond to the dark/light color scheme setting and rounds the corners even on transparent stages.

modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
beldenfox: I prototyped a version of this for Win10. On that platform we have to roll our own backdrops using DirectComposition or Windows.UI.Composition and that requires building the HWND a little differently, firing up a different painter, and re-plumbing the way pixels are delivered from Prism. I'm not entirely sure we can set all that up after the HWND is created. It's just far far easier to have the backdrop in hand from the start. (Well, all we really need to know when we create the platform window is that we'll be using a backdrop. The actual backdrop choice could be specified dynamically later but I wasn't sure what purpose that would serve.)

I'm not sure how this would work on Linux. The Wayland folks recently added an extension for blurring window backgrounds and it looks like it can be dynamically added and removed but I can't test that right now and have no idea which compositors have implemented it.

modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
beldenfox: The platforms seem to assume you'll draw content on the backdrop using the PlatformPreference colors (there's no way of asking the platforms for backdrop-specific color information). But it still might be good for the stylesheets to know which backdrop is in effect.

A more interesting question is whether a stylesheet can ask for a backdrop to be applied. The CSS machinery isn't set up for this; it's designed to apply styles to nodes, not windows (AFIK). And you probably don't want to apply a backdrop to, say, a menu or tooltip since the backdrop would fill the entire stage forcing it to be rectangular (no rounded corners). If the CSS called for a drop shadow the backdrop would be behind the shadow. I considered ways of fixing this but didn't come up with a workable solution.

I'll be traveling for a few days so it might be a while before I re-enter the conversation. Thanks for all the feedback!

modules/javafx.graphics/src/main/native-glass/win/GlassWindow.cpp
```diff
@@ -338,7 +339,7 @@ LRESULT GlassWindow::WindowProc(UINT msg, WPARAM wParam, LPARAM lParam)
             }
             break;
         case WM_DWMCOMPOSITIONCHANGED:
-            if (m_isUnified && (IS_WINVISTA)) {
+            if (m_isUnified || m_hasBackdrop) {
```
crschnick, line 343: Should a backdrop automatically set the window margin akin to UNIFIED (and maybe EXTENDED)? In my opinion, this should be left to the stage style for consistency.

Like if I use the normal stage style with a backdrop, I might not want the full backdrop to be visible like that. Maybe I only want the backdrop the be visible in the title bar.

beldenfox: Good question. I had to do this to make the backdrops work. I tweaked Prism to ensure the swap chain supported an alpha channel but apparently that's not enough. To get D3D9 to alpha composite onto the backdrop we have to make this call to invoke the old Aero glass logic (the effect UNIFIED was originally designed for). I'll keep looking into this but for now when there's a backdrop DECORATED and UNIFIED look the same.

I'll also take a look at DWMSBT_AUTO which states that it only draws the backdrop in the title bar but doesn't give us any control over the effect.

Most Microsoft apps that employ these backdrops extend the backdrop into the title bar though there's a lot of subtle variation (and then there's Weather which looks like it was written by a different company).

kevinrushforth:
This will need a CSR.

/reviewers 2
/csr

andy-goryachev-oracle:
This probably needs 3 reviewers instead of 2.

crschnick:
Depending on how long this takes, would it be possible to extract the fix for the unified stage BackBufferFormat (if it actually fixes all cases) as a separate PR? That would make the review easier to split a feature addition and bugfix into two issues. I can definitely do the testing for the fix

beldenfox:
> Depending on how long this takes, would it be possible to extract the fix for the unified stage BackBufferFormat (if it actually fixes all cases) as a separate PR?

It would probably be faster for you to test and submit the PR yourself. The fix is just the two lines in native-prism-d3d where the BackBufferFormat is set to D3DFMT_A8R8G8B8 instead of being left unspecified. Based on the write up for bug [JDK-8154847](https://bugs.openjdk.org/browse/JDK-8154847) if you don't set the back buffer format what you get varies depending on the graphic card. It should be OK to enable alpha unconditionally, in the end it should only affect UNIFIED. Most stage styles wipe out the alpha channel by drawing an opaque background. The exceptions are TRANSPARENT and UNIFIED and only UNIFIED uses the swap chain.

If you'd prefer I can enter the PR but it would probably have to wait until next week.

crschnick:
Can I somehow attribute the PR authorship to you in skara?

kevinrushforth:
> Can I somehow attribute the PR authorship to you in skara?

Yes: `/contributor add @beldenfox` will do it.

crschnick:
@beldenfox you can now rebase on master and remove the BackBufferFormat adjustments in this PR. That should make this PR more straightforward

beldenfox:
This PR has been on hold while we figure out how it would work with the D3D12. For legacy technologies like D3D9 the OS performs some magic; invoking the old Aero glass effect by calling ::DwmExtendFrameIntoClientArea automatically allows the D3D9 swap chain to composite over the backdrop. For D3D12 it looks like we need to implement this directly by building a compositing swap chain. We haven't sorted out exactly how this will work.

I would like to enable dynamically changing the backdrop after the window is shown (moving the call from Stage.initBackdrop to Stage.setBackdrop or Scene.setBackdrop). But it remains to be seen whether we can get that to work with D3D12 or not.

beldenfox:
I've updated this PR to treat StageBackdrops more like fonts. A backdrop is now an object that's built by passing a material to a factory method. Materials are specified using strings and the list of materials can vary by platform. All platforms that support backdrops will provide two default materials ("Window" and "Partial") and some number of materials that are platform-specific. For platform-specific materials the naming convention is the same as platform preferences; macOS prefixes the material name with "macOS." and Windows prefixes it with "Windows.".

In the future we may want to provide a factory method that accepts material-specific options but at the moment there are none.

I added "macOS.Glass" and "macOS.ClearGlass" to the list of available materials on macOS 26 but had to remove them since they require the macOS 26 SDK and the build machines are still on macOS 15.

RealCrystalNight:
For Linux/Wayland support, the ext-background-effect-v1 protocol (which is now implemented in KWin, with KDE Plasma 6.7 shipping this support on June 16, 2026, [wlroots](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/include/wlr/types/wlr_ext_background_effect_v1.h) and [Niri](https://github.com/niri-wm/niri/pull/3483) with planned support coming to other DE's such as [Cosmic](https://github.com/pop-os/cosmic-comp/pull/2179) and [Hyprland](https://github.com/hyprwm/Hyprland/pull/13211) as well as others) could be a viable path forward. Unlike some of the platform-specific constraints discussed above, this protocol supports dynamically adding/removing background effects, which aligns with the goal of supporting setBackdrop() after the window is shown. 

Reference: https://wayland.app/protocols/ext-background-effect-v1

Real-world implementations in Ghostty (https://github.com/ghostty-org/ghostty/pull/10727) and Quickshell (https://github.com/quickshell-mirror/quickshell/pull/566) demonstrate how this works in practice.
This would only cover Wayland, of course X11 would likely need a separate mechanism or remain unsupported for backdrops. Most major desktop environments now default to Wayland, so this would cover the majority of Linux users. 

Additional notes:
KDE 6.7, which includes [ext-background-effect-v1](https://www.phoronix.com/news/KDE-Brings-Back-Air-Theme), is planned for release on [Tuesday 16th June 2026](https://9to5linux.com/kdes-new-css-based-style-engine-union-is-coming-to-kde-plasma-6-7).

For testing purposes you can test this on KDE Neon Unstable edition as a Live ISO or in a VM:
https://neon.kde.org/download

tsayao:
@RealCrystalNight At the moment, JavaFX runs on Wayland through XWayland, which prevents it from using Wayland-specific extensions. Native Wayland support would be required for that, but implementing it is a significant effort.

beldenfox:
@RealCrystalNight Thank you for the details, I'll take a look at the PR's you linked to. Last time I looked it seemed that the Wayland protocols only handled blurring the content behind the window. On macOS and Win11 the OS starts there and then layers other effects on top of the blur to improve text legibility, handle light and dark mode, and provide feedback on the window's active state. If that's still the case it will take some effort to get Wayland in line with macOS and Win11.

My plan is to keep the current Stage.installBackdrop API rather than switch to a dynamic Stage.setBackdrop call. This won't be an issue if we ever want to enable CSS support for specifying the backdrop. For popups like menus changes to the stylesheets are processed before the platform window is created so the desired backdrop is known at window creation time anyway. I don't currently have a use case for Stage.setBackdrop and it does simplify matters to know the backdrop at window creation time and only handle setting it up once.

And, yes, I'm working on the CSR.

mstr2:
The proposed API feels very stringly-typed, and I'm not sure if that's a good model. Specifically, I'm looking at future enhancements that expose configuration options which are available on some [platforms](https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/ui/apply-mica-win32) and which we might want to support.

I've played around with your code for a bit and came up with a slightly different idea of how we could design an API. Feel free to criticize or dismiss:

```java
// Public API:
public sealed interface StageBackdrop permits DefaultStageBackdrop, PlatformStageBackdrop {

    // Default backdrops are available as static constants
    StageBackdrop WINDOW = DefaultStageBackdrop.WINDOW;
    StageBackdrop PARTIAL = DefaultStageBackdrop.PARTIAL;

    /**
     * {@return the name of the backdrop}
     */
    String name();

    // Potential future enhancement:
    /**
     * {@return a backdrop with the specified option}
     *
     * @throws NullPointerException if {@code key} or {@code value} is {@code null}
     * @throws IllegalArgumentException if the option is not supported by this backdrop
     */
    StageBackdrop withOption(String key, Object value);

    /**
     * {@return all platform backdrops supported on this system}
     *
     * For systems where backdrops are not supported this will be an empty list.
     */
    static List<StageBackdrop> getPlatformBackdrops() {
        return PlatformStageBackdrop.BACKDROPS;
    }
}

// Platform-independent implementations:
public enum com.sun.javafx.stage.DefaultStageBackdrop implements StageBackdrop {
    WINDOW,
    PARTIAL
}

// Named platform implementations:
public record com.sun.javafx.stage.PlatformStageBackdrop(String name) implements StageBackdrop {
    public static final List<StageBackdrop> BACKDROPS =
        Toolkit.getToolkit().getBackdropMaterials().stream()
            .map(name -> (StageBackdrop)new PlatformStageBackdrop(name))
            .toList();
}
```

`StageBackdrop.getPlatformBackdrops()` would only return the platform-prefixed named backdrops, not the two platform-independent variants. Users would only interact with `StageBackdrop` objects, not with strings. This also gets rid of the "material" terminology, which I find a bit confusing (I ask the backdrop for a list of materials, and then turn a material back into a backdrop?).

The only way to obtain a `StageBackdrop` is to use a static default instance, choose one of the platform-provided backdrops, or potentially in the future start with a platform-provided backdrop, and derive a new variant with options from it. Backdrop objects are always immutable.

Here's how the API could be used:
```java
// Default backdrop
stage.initBackdrop(StageBackdrop.WINDOW);

// Platform-specific backdrop with default fallback
stage.initBackdrop(StageBackdrop.getPlatformBackdrops().stream()
    .filter(b -> b.name().equals("Windows.Transient"))
    .findFirst()
    .orElse(StageBackdrop.WINDOW));

// Configurable backdrop with default fallback
stage.initBackdrop(StageBackdrop.getPlatformBackdrops().stream()
    .filter(b -> b.name().equals("Windows.Mica"))
    .map(b -> b.withOption("TintColor", Color.RED))
    .findFirst()
    .orElse(StageBackdrop.WINDOW));
```

beldenfox:
@mstr2 Thank you for looking into this in such detail. I really appreciate it.

I've been going back and forth on your proposal over the last few days. I'm now in favor of changing the API to more or less match what you've written up. There's a lot to be said for firmly separating the standard backdrops from the platform-dependent ones and providing different means of accessing them. I'll start migrating the API later this week.

mstr2:
One thing we should probably do is document the platform backdrops that might be available, similar to how we document the platform preferences that are potentially available.

Short of letting users choose from a list of backdrops, there is very little developers can do with the backdrop names at runtime if they don't know the potentially available names in advance.

crschnick:
Could that approach potentially support customization? For example liquid glass allows a tint color, etc. Also, would stuff like the window border radius be considered a part of the backdrop? Because that is also available to be changed via native APIs

Edit: I see the customization method `withOption`, so you can disregard the first question

beldenfox:
> Also, would stuff like the window border radius be considered a part of the backdrop? Because that is also available to be changed via native APIs

For the clear glass effect on macOS you can change the corner radius and, yes, that would be one of the backdrop options. I wouldn't provide that for any of the other backdrop effects. You can change the corner radius of an NSVisualEffectView's layer but not the corner radius of the NSWindow so the edging visuals don't match. In theory you could use a transparent stage to get rid of the NSWindow visuals and try to draw the edging lines using JavaFX but I'm dubious that you would get good results.

In the next round I'll be removing the non-clear glass effect on macOS since there's some graphic artifacts showing up in the corners. I don't think it's designed for use by top-level windows.

crschnick:
How would the API handle cases where a `withOption` call is not supported for a certain system? For example when a certain platform implementation does not support setting corner radius values or other options. And how would the user find this out? Via some `isSupported()` method?

beldenfox:
In the API I'm working on you can query for supported options. The return is a Map<String, Class<?>> containing the name of each option and the class of object it accepts.

Personally I favor ignoring options that aren't supported. After all we ignore backdrops and StageStyles if they're not supported. Throwing exceptions introduces a bunch of testing issues and seems excessive.

tsayao:
What if the backdrop effect is technically supported by the platform, but the user has disabled transparency effects or enabled a “reduced transparency” accessibility setting?

beldenfox:
> What if the backdrop effect is technically supported by the platform, but the user has disabled transparency effects or enabled a “reduced transparency” accessibility setting?

For the standard backdrops (WINDOW and PARTIAL) reduced transparency doesn't matter much. In normal operation the window backdrop will track the window's color scheme (LIGHT or DARK). The translucency effect might go away but the overall background color will be the same.

The real complication comes from the other accessibility settings; on Windows the backdrop will also respond to the high contrast themes. The most appropriate way to draw on a standard backdrop is to make no assumptions about what color it's going to display and consult Platform Preferences for appropriate foreground colors.

For the platform-specific backdrops anything goes. Most of them will follow the window's color scheme, some will not. The clear glass effect on macOS is always light colored when "reduced transparency" is off and ignores the color scheme. But if "reduced transparency" is on it starts tracking the color scheme. Most of the other effects on macOS track the window's color scheme and platform accessibility settings.

modules/javafx.graphics/src/main/java/javafx/stage/StageBackdrop.java
```diff
@@ -0,0 +1,124 @@
...
+        return List.of(WINDOW, PARTIAL);
+    }
+
+    /**
+     * Gets the names of platform backdrops supported on this system. The list
+     * may be empty.
+     *
+     * @return An unmodifiable list of names of the supported platform backdrops.
+     */
+    public static List<String> getPlatformBackdropNames() {
```
mstr2: What's the reason for returning a list of names instead of a list of backdrops? If you returned the latter, there would be no need for a factory method to create a backdrop instance, and no way to pass in a wrong string.

beldenfox:
I've pushed a new version. StageBackdrops now come in two flavors, the standard backdrops which are referenced using constants (WINDOW and PARTIAL) and the platform-dependent backdrops which are referenced by name (string).

The selection of platform backdrops is not final, the current crop are just for demo purposes. I had to remove the macOS liquid glass backdrops since they require a more recent SDK than the build machines support. If you want to try them out you'll need to edit MacWindow.java (look for ClearGlass) and GlassHostView.m (look for setGlassBackdrop).

There's now a call to query a backdrop for its supported options. The call for setting an option has been moved to Stage since you can change the value at any time. That call is currently a no-op since the only backdrop that supports options is macOS.ClearGlass and it's not available. Hooking this up isn't rocket science but is a bit tedious so I'm putting it off.

modules/javafx.graphics/src/main/java/javafx/stage/StageBackdrop.java
beldenfox: When prototyping the CSS code and even the sample app the names were all that mattered. The utility functions I wrote to map strings to backdrops always had to deal with invalid names anyway. It also seems that a factory function aligns better with existing JavaFX API's (see Font) and forcing clients to iterate does not. And I realize that it probably doesn't matter here but iteration doesn't scale well. Overall I felt that a factory function was a better fit.

Put another way, we could return a list of backdrops instead of names but I would still want to provide a static function for finding a backdrop by name. But maybe I'm focusing too much on the specific use cases I encountered.

mstr2:
> There's now a call to query a backdrop for its supported options. The call for setting an option has been moved to Stage since you can change the value at any time. That call is currently a no-op since the only backdrop that supports options is macOS.ClearGlass and it's not available. Hooking this up isn't rocket science but is a bit tedious so I'm putting it off.

Having `setBackdropOption` be a method on `Stage` seems to be a very loose-fitting API; for me, a backdrop option is a property of a backdrop (as the name implies), not a property of a stage. The canonical pattern to define properties on an object that are only meaningful in the context of another object is an [attached property](https://github.com/openjdk/jfx/pull/2015). However, I don't think that this is a good fit when the natural carrier for this API is `StageBackdrop`.

Maybe you moved backdrop options to `Stage` because that allows `StageBackdrop` to essentially be an immutable key that identifies a backdrop. But it also clearly suggests to me that if the configuration of a backdrop is dynamically modifiable, it can't be an immutable object at the same time.

I don't see a huge problem of having `StageBackdrop` be a mutable object. The only sharp edge is the two standard backdrops, which are static constants and should therefore be intrinsically immutable.

beldenfox:
> Maybe you moved backdrop options to `Stage` because that allows `StageBackdrop` to essentially be an immutable key that identifies a backdrop.

No, it's because the same `StageBackdrop` can be installed on multiple stages.

The problem is partly one of naming. A StageBackdrop object doesn't represent a backdrop, it encapsulates the information needed for a stage to build a backdrop. That creation info is sharable across stages but each actual backdrop is bound to a specific stage. Once a stage's backdrop is built there needs to be a separate API for accessing it and changing its dynamic options. I don't think it's a good idea to try to make one object serve both purposes.

Let me think about the naming here. Every alternative I can think of seems cumbersome and overly pedantic.

mstr2:
> > Maybe you moved backdrop options to `Stage` because that allows `StageBackdrop` to essentially be an immutable key that identifies a backdrop.
> 
> No, it's because the same `StageBackdrop` can be installed on multiple stages.
> 
> The problem is partly one of naming. A StageBackdrop object doesn't represent a backdrop, it encapsulates the information needed for a stage to build a backdrop.

Leaving naming discussions aside, I'm questioning the semantics here. I don't really understand what you mean with this sentence. What information is encapsuled by a `StageBackdrop`? As currently proposed, it has no configurable payload and it is used like an immutable identifier; aside from its name, there doesn't seem to be any information. Your implementation confirms this interpretation, as you're just deriving an int-based ID from the backdrop name, which is just another form of the name itself.

> That creation info is sharable across stages but each actual backdrop is bound to a specific stage. Once a stage's backdrop is built there needs to be a separate API for accessing it and changing its dynamic options. I don't think it's a good idea to try to make one object serve both purposes.

In the class metaphor of OOP, the class is the blueprint of a thing, and the instance is the thing itself with concrete values. The methods for accessing the values are purposefully colocated with the class that defines their meaning. You're proposing to have the configuration API be located on a different class, which is very unusual.

beldenfox:
Imagine that sometime in the future we decide to add a TintColor option to the WINDOW backdrop style (stranger things have happened). With the current proposed API we could write code that sets different tint colors on each stage:
```
stage1.initBackdrop(StageBackdrop.WINDOW);
stage2.initBackdrop(StageBackdrop.WINDOW);
stage1.setBackdropOption("TintColor", Color.YELLOW);
stage2.setBackdropOption("TintColor", Color.GREEN);
```
The first two lines convey the style of backdrop we want associated with each stage. The last two lines access the stage-specific backdrops to update their tint colors. A more accurate API would be:
```
stage1.initBackdropStyle(StageBackdropStyle.WINDOW);
stage2.initBackdropStyle(StageBackdropStyle.WINDOW);
stage1.getBackdrop().setOption("TintColor", Color.YELLOW);
stage2.getBackdrop().setOption("TintColor", Color.GREEN);
```
where Stage.getBackdrop() returns an object representing the backdrop associated with that stage.

With this design clients specify the style/type/class of backdrop in the init call and let the stage create the backdrop instance. Then they use a separate call to retrieve that instance to manipulate it. That's the distinction I was trying to make in my earlier message.

beldenfox:
I've updated the API to distinguish between the style of backdrop associated with a stage and the backdrop itself.

Keep in mind that everything related to setting backdrop options (including the StageBackdrop class itself) will probably be removed. There's only one backdrop which supports options and that's the ClearGlass effect on macOS. That effect can't be supported until JavaFX updates to the macOS 26 SDK. For now the code is present for prototyping and testing but commented out.

beldenfox:
On Mac there was overlap between the standard and platform backdrop styles (WINDOW was the same as macOS.Window, PARTIAL was the same as macOS.Sidebar). On Windows there was no overlap. I removed the overlap on macOS so the two platforms match. I also added macOS.HUD largely because the visual effect is similar to Windows.Transient.

I'm not sure what criteria to use to determine which platform backdrop styles to include. On Windows I added Windows.Transient for completeness. macOS provides a different material for every type of window you might want to create but many of them seem to map to the same visual effect. It's not clear to me that any of these are worth adding at this point. Eventually I would like to support macOS.ClearGlass so it's worth keeping the platform backdrop style API around even if we decide to drop the ones in this PR.

I want to proceed with the CSR but am having difficulty creating the specification diff files. Perhaps I'm overlooking something obvious or my Google-Fu is weak today but I can't find clear instructions on how to create API diffs. Any pointers would be appreciated.

mstr2:
I think the main point of tension in the API is that we have two seemingly different forms of backdrops:
1. Constant backdrops: `WINDOW` and `PARTIAL`, which are exposed as static fields.
2. Configurable backdrops, which include dynamic parameters that can change at runtime.

Your solution is to split off the "class" of the backdrop from the configuration object: `StageBackdropStyle` is the "class", and `StageBackdrop` is an object that can hold dynamic parameters. However, the way you've designed this -- call `initBackdropStyle()` and then a configuration object is created for you that you can access with another API -- that's not a pattern that is used anywhere else in JavaFX (that I know of).

I think there shouldn't be two API elements to configure a single aspect about the `Stage`, it should be a single API element. Instead, move the complexity from `Stage` to `StageBackdrop`.

Here are two other ideas how you could achieve that:
1. Have a separate `ConfigurableStageBackdrop` interface that enables you to set dynamic options. This is not implemented by the `WINDOW` and `PARTIAL` constants, which is why they can be constants in the first place:
    ```java
    public sealed interface StageBackdrop permits StandardStageBackdrop, ConfigurableStageBackdrop {
        StageBackdrop WINDOW = StandardStageBackdrop.WINDOW;
        StageBackdrop PARTIAL = StandardStageBackdrop.PARTIAL;
        
        static StageBackdrop of(String name);
        static ConfigurableStageBackdrop ofConfigurable(String name);
    }

    public interface ConfigurableStageBackdrop extends StageBackdrop {
        void setOption(String key, Object value);
    }
    ```

2. If you expect that we might add options to the predefined constants, then that's a strong argument that these shouldn't be constants in the first place, and that _all_ backdrops should be able to carry options:
    ```java
    public final class StageBackdrop { // maybe implement Map<String, Object>
        private StageBackdrop();

        public static StageBackdrop newWindowStyle();
        public static StageBackdrop newPartialStyle();
        public static StageBackdrop newPlatformStyle(String styleName, Supplier<StageBackdrop> fallback);

        public StageBackdrop setOption(String key, Object value);
    }
    ```
    In this example, the API forces callers to provide a fallback backdrop in case the named backdrop is not available. This ensures that developers don't need to worry about `null`:
    ```java
    var backdrop = StageBackdrop.newPlatformStyle("macOS.ClearGlass", StageBackdrop::newWindowStyle)
                                .setOption("TintColor", Color.RED);
    // backdrop is not null
    ```


Looking forward, we'll want to also support translucent backdrops on Linux at some point in the future. With GTK4/Wayland, we could use [ext-background-effect-v1](https://wayland.app/protocols/ext-background-effect-v1#ext_background_effect_surface_v1:request:set_blur_region) to achieve that, which says: "The blur algorithm is subject to compositor policies."
I'm bringing this up because platform implementations might not only allow developers to use _more_ types of backdrops, but also come with restrictions that allow no choice at all: In this case, there would be no difference between `WINDOW` and `PARTIAL` on Linux. Maybe this means our API should only have a single default backdrop, and for all other backdrops, you'd use platform-specific names? In this case, a potential API could be reduced to just:
```java
public final class StageBackdrop { // maybe implement Map<String, Object>
    private StageBackdrop();

    public static StageBackdrop newDefaultStyle();
    public static StageBackdrop newPlatformStyle(String styleName); // if not available, returns default style

    public StageBackdrop setOption(String key, Object value);
}
```


Another aspect we should consider is the interaction with `Scene.Preferences.reducedTransparency`: This preference can be toggled at runtime, potentially by users of the application. Should it have an effect on a potential translucent window background?


> I want to proceed with the CSR but am having difficulty creating the specification diff files. Perhaps I'm overlooking something obvious or my Google-Fu is weak today but I can't find clear instructions on how to create API diffs. Any pointers would be appreciated.

The diffs are [here](https://github.com/openjdk/jfx/pull/2048#issuecomment-3781001757).

beldenfox:
> I think the main point of tension in the API is that we have two seemingly different forms of backdrops:
> 
> 1. Constant backdrops: `WINDOW` and `PARTIAL`, which are exposed as static fields.
> 2. Configurable backdrops, which include dynamic parameters that can change at runtime.

This PR assumes that any backdrop can have dynamic parameters, including WINDOW and PARTIAL. The API for setting options is the same for all styles so we can easily add dynamic parameters to the standard backdrop styles in the future.

> I think there shouldn't be two API elements to configure a single aspect about the `Stage`, it should be a single API element. Instead, move the complexity from `Stage` to `StageBackdrop`.

This raises the possibility that a client could take the option-setting object (the StageBackdrop) and assign it to multiple stages. We could:

1. Allow it; when an option changes all of the stages that reference that StageBackdrop get updated. I don’t think it’s intuitive or useful to share options across stages and it would complicate the implementation and testing.

2. Prevent it; force the user to create a new StageBackdrop instance for each stage. We would be forcing clients to create multiple objects to support a feature (per-stage option setting) most of them won't even use.

With the API in this PR the vast majority of calls will be either `initBackdropStyle(StageBackdropStyle.WINDOW)` or `initBackdropStyle(StageBackdropStyle.PARTIAL)` which mirror the existing API calls for setting the StageStyle. I'm hoping most clients won't use the platform-specific styles and I doubt they'll have much use for the options API. The only style which takes options is a novelty (macOS.ClearGlass) and I don't see that changing anytime soon.

>    In this example, the API forces callers to provide a fallback backdrop in case the named backdrop is not available. This ensures that developers don't need to worry about `null`:

I think clients would rather develop their own fallback strategies.

> Looking forward, we'll want to also support translucent backdrops on Linux at some point in the future. With GTK4/Wayland, we could use [ext-background-effect-v1](https://wayland.app/protocols/ext-background-effect-v1#ext_background_effect_surface_v1:request:set_blur_region) to achieve that, which says: "The blur algorithm is subject to compositor policies." I'm bringing this up because platform implementations might not only allow developers to use _more_ types of backdrops, but also come with restrictions that allow no choice at all: In this case, there would be no difference between `WINDOW` and `PARTIAL` on Linux. Maybe this means our API should only have a single default backdrop, and for all other backdrops, you'd use platform-specific names?

Mac and Windows take the same approach for specifying a backdrop: you describe the type of content you’re displaying and the OS picks an appropriate visual. At the API level there’s not one style per visual effect, there’s one style per content type. It’s possible that some content types share the same visual effect but that may change in a future OS release. I’m inclined to take the same approach for the JavaFX API so it doesn’t matter if WINDOW and PARTIAL aren’t visually distinct on all platforms.

> Another aspect we should consider is the interaction with `Scene.Preferences.reducedTransparency`: This preference can be toggled at runtime, potentially by users of the application. Should it have an effect on a potential translucent window background?

With the exception of macOS.ClearGlass all of the existing backdrops interact with the window’s color scheme, activation state, and multiple platform preferences. We don't have control over any of this. Reduced transparency is the easiest; it removes translucency but doesn’t have much effect on the background color. The accessibility contrast modes do change the background color. In the end the assumption is that backdrops draw using a background color similar to `PlatformPreferences.backgroundColor` and so all text should be drawn using `PlatformPreferences.foregroundColor`. This is about all we can guarantee about the standard backdrops.

This will complicate the implementation on Linux. I suspect the Wayland extension will only handle blurring and it will be up to us to track the platform preferences to get WINDOW and PARTIAL to behave the same way they do on Mac and Windows.

mstr2:
> This raises the possibility that a client could take the option-setting object (the StageBackdrop) and assign it to multiple stages. We could:
> 
> 1. Allow it; when an option changes all of the stages that reference that StageBackdrop get updated. I don’t think it’s intuitive or useful to share options across stages and it would complicate the implementation and testing.
> 2. Prevent it; force the user to create a new StageBackdrop instance for each stage. We would be forcing clients to create multiple objects to support a feature (per-stage option setting) most of them won't even use.

The closest thing to `StageBackdrop` is `Paint`: it's a configurable object that describes some visual appearance. Most paints are immutable, but `ImagePattern` is mutable when it represents an animated image. Paints are reusable within a scene and across multiple scenes, so when you say sharing configuration objects is neither intuitive nor useful, you're arguing against prior art. I'm also not sure what to make of your presented arguments: implementation and testing difficulty shouldn't be the main driving force behind API design (they _can_ be major considerations, but grounding an API shape in those considerations should require a high bar of justification). Similarly, "forcing clients to create multiple objects" seems like a pretty neutral statement to me; by itself it's not a great argument in favor or against any particular API shape.

Because `StageBackdrop` and `Paint` are so similar, I've experimented at one point to have a paint type like `BackdropBlur extends Paint` that, when applied to the scene background, would enable the backdrop blur effect for the window (and by extension, when applied to a region within the JavaFX scene graph, would enable a similar effect within the scene graph).

In the end, I've come to the conclusion that `StageBackdrop` is sufficiently dissimilar from the other paint types that it shouldn't be an actual `Paint`. However, I still think that the similarity that undoubtedly exists should result in similar patterns of use. This is compounded by my observation that your proposed API (_you_ say which type of thing you want to have, and _JavaFX_ creates another type of configuration object for you internally) doesn't seem to fit in with existing JavaFX patterns.

I'm not ready to "disagree and commit" to your proposal, but I hope that other people will join the discussion to move it forward.

> > In this example, the API forces callers to provide a fallback backdrop in case the named backdrop is not available. This ensures that developers don't need to worry about `null`:
> 
> I think clients would rather develop their own fallback strategies.

Sounds reasonable. However, you're exposing `null` to callers in an indirect way. It's one thing to have a pair of `initBackdrop(StageBackdrop)` and `getBackdrop() -> StageBackdrop`, where the getter returns whatever you pass into the setter; and if you pass `null` into the setter, you'll also receive `null` from the getter. Another thing entirely is what you are proposing: you call `initBackdropStyle(StageBackdropStyle)`, and the return value of `getBackdrop() -> StageBackdrop` is now either `null`  or an object you didn't create.

> Mac and Windows take the same approach for specifying a backdrop: you describe the type of content you’re displaying and the OS picks an appropriate visual. At the API level there’s not one style per visual effect, there’s one style per content type. It’s possible that some content types share the same visual effect but that may change in a future OS release. I’m inclined to take the same approach for the JavaFX API so it doesn’t matter if WINDOW and PARTIAL aren’t visually distinct on all platforms.

Sounds reasonable.

> With the exception of macOS.ClearGlass all of the existing backdrops interact with the window’s color scheme, activation state, and multiple platform preferences. We don't have control over any of this. Reduced transparency is the easiest; it removes translucency but doesn’t have much effect on the background color. The accessibility contrast modes do change the background color. In the end the assumption is that backdrops draw using a background color similar to `PlatformPreferences.backgroundColor` and so all text should be drawn using `PlatformPreferences.foregroundColor`. This is about all we can guarantee about the standard backdrops.
> 
> This will complicate the implementation on Linux. I suspect the Wayland extension will only handle blurring and it will be up to us to track the platform preferences to get WINDOW and PARTIAL to behave the same way they do on Mac and Windows.

I understand that when we use a platform-provided backdrop, the "reduce transparency" OS setting will be automatically applied. In my scenario, the OS-level setting is not enabled, but `Scene.Preferences.reducedTransparency` is. This is a completely supported scenario in JavaFX, with the implication that well-behaved applications will respect the setting and reduce their use of transparency. My question is what we should do about the transparent background in this case. It won't be made non-transparent by the OS automatically, because the OS-level "reduced transparency" setting is not enabled. Still I think we should, where possible, respect the JavaFX-level setting and remove the transparent backdrop material.

beldenfox:
> The closest thing to `StageBackdrop` is `Paint`: it's a configurable object that describes some visual appearance. Most paints are immutable, but `ImagePattern` is mutable when it represents an animated image. Paints are reusable within a scene and across multiple scenes, so when you say sharing configuration objects is neither intuitive nor useful, you're arguing against prior art.

I don't see anything mutable in an `ImagePattern` that references an animated `Image`.

> Paints are reusable within a scene and across multiple scenes, so when you say sharing configuration objects is neither intuitive nor useful, you're arguing against prior art.

I was making a cost/benefit case and arguing that the benefit was zero. You haven't said anything to change my mind on the benefit = zero part. And I don't see prior art here; JavaFX avoids sharing mutable objects whenever possible.

I can imagine changing the API so the client creates the StageBackdrop instead of the Stage. Assigning the same backdrop to multiple stages would still be prohibited as opposed to the current API where it's impossible.

> I'm not ready to "disagree and commit" to your proposal, but I hope that other people will join the discussion to move it forward.

I hope so, too.
 
> Still I think we should, where possible, respect the JavaFX-level setting and remove the transparent backdrop material.

If an application wants to create a UI with reduced transparency wouldn't they just avoid using backdrops in the first place? I'm not that familiar with the preferences API but I thought it was the app developer's job to respond to these settings.

This preference can change on the fly but I'm not sure we can respond by making the OS backdrop come and go (that's why it's an init call). We can do this on Mac and Win11 using D3D9 but I'm doubtful about the D3D12 backend and Linux is a big unknown. If a client wants reduced transparency outside the OS setting they could just not use backdrops or cover the backdrop with an opaque fill. I don't think the JavaFX core code can do much better than that.

mstr2:
> I don't see anything mutable in an `ImagePattern` that references an animated `Image`.

It's not mutable from an API perspective, but it is mutable in its implementation. The change is triggered by a timer, not by user code. However, I think the question of mutability is a bit tangential, maybe I shouldn't even have brought it up. It's also not my main point. My main point is that `StageBackdrop` is quite closely related to `Paint`, which are shareable configuration objects created by users, and should therefore have a similar API.

> I was making a cost/benefit case and arguing that the benefit was zero. You haven't said anything to change my mind on the benefit = zero part. And I don't see prior art here; JavaFX avoids sharing mutable objects whenever possible.

You're again trying to ground the API shape in terms of implementation concerns: that by allowing backdrops to be shared across stages we'll increase the cost (which cost?) without adding benefit. But a backdrop is essentially an observable map, where is the cost in allowing it to be shared? It shouldn't matter too much whether we have a single listener or multiple listeners.

So you take the cost argument, and use it to explain why a backdrop needs to have a separate API element next to the backdrop style, and why it shouldn't be created by users of the API. I hope that I've made it clear that I consider two entangled API elements for what could be a single API element to be highly unusual, which comes with an appropriately high bar of justification.

> If an application wants to create a UI with reduced transparency wouldn't they just avoid using backdrops in the first place? I'm not that familiar with the preferences API but I thought it was the app developer's job to respond to these settings.
> 
> This preference can change on the fly but I'm not sure we can respond by making the OS backdrop come and go (that's why it's an init call). We can do this on Mac and Win11 using D3D9 but I'm doubtful about the D3D12 backend and Linux is a big unknown. If a client wants reduced transparency outside the OS setting they could just not use backdrops or cover the backdrop with an opaque fill. I don't think the JavaFX core code can do much better than that.

Preferences are the user's choice (compared to the developer's choice when building the app), and should ideally be respected out of the box, without requiring developers to "opt in" by branching over preferences (setting a backdrop when reducedTransparency=false, and not setting a backdrop otherwise).

I don't think that we should be automagically and permanently "downgrade" a backdrop to no backdrop when reducedTransparency is enabled, as if by ignoring the argument of the init call, and setting it to null. Because if we did, then what would happen if reducedTransparency is later disabled? We'd expect the backdrop to reappear, but since we've permanently downgraded to "no backdrop", nothing happens.

That leaves us with two other options: make it a best-effort operation, or don't respect the reducedTransparency preference at all. In the first case, we could dynamically remove the backdrop on platforms where this is possible, and don't do anything on platforms where this isn't possible.

beldenfox:
> > I don't see anything mutable in an `ImagePattern` that references an animated `Image`.
> 
> My main point is that `StageBackdrop` is quite closely related to `Paint`, which are shareable configuration objects created by users, and should therefore have a similar API.

And Paints are immutable by design.

> You're again trying to ground the API shape in terms of implementation concerns: that by allowing backdrops to be shared across stages we'll increase the cost (which cost?) without adding benefit. But a backdrop is essentially an observable map, where is the cost in allowing it to be shared?

And you're again arguing based on your notion of API consistency and not utility. With your proposed API this call:
```
stage.getBackdrop().setOption("TintColor", Color.RED);
```
could change the tint color for multiple stages. Now that, that is unusual in the JavaFX API. And I think unexpected. Will developers consider this useful or behavior that they have to work around?

> But a backdrop is essentially an observable map, where is the cost in allowing it to be shared?

I will update this PR or create a new PR with your proposed API. Currently the backdrop maintains a Map, not an ObservableMap, and that will need to change since we'll need to engineer this object to be shared. We'll be adding a constraint to the implementation.

In theory there's a testing burden with this but since I have no idea how to write automated tests for this feature that point may be moot.

> Preferences are the user's choice (compared to the developer's choice when building the app), and should ideally be respected out of the box, without requiring developers to "opt in" by branching over preferences (setting a backdrop when reducedTransparency=false, and not setting a backdrop otherwise).

Again, I'm not that familiar with this API. The OS level setting for "reducedTransparency" is clearly the user's choice. I'm less clear on what it means to override the OS setting at the Scene level.

The OS backdrop selects a background color based on the window's state and various preferences and draws that beneath the Scene's fill color. This is the behavior that developers will have to design for and test against. For example, clients of the backdrop API will automatically be exposed to the to high and low contrast mode settings.

To honor a reduced transparency override we'll need to disable the OS backdrop which will also throw away its drawing of a background color. That will need to be replaced.

I've given this some thought. I was considering making backdrops "cross-platform" by emulating the background color drawing behavior on platforms that don't support backdrops natively. Because, again, the translucency is really beside the point in terms of development and testing. Win11 barely has any translucency.

So it's not just a question of reduced transparency. What does it mean for a client to override the system's background color in the Scene settings? In theory this should also mean disabling the OS backdrop since it can't honor that setting. It's not a question of how to choose a background color, it's a question of when we want to disable the OS backdrop. Whenever the client has overriding reducedTransparency? Whenever the client has overridden the Scene's background color? Whenever the window's color scheme doesn't match the platform's color scheme (at which point the OS will draw a background color that doesn't match the global background color preference)? Under these conditions either JavaFX needs to figure out the background color or we need to provide guidance to developers on what to do.

crschnick:
About the reduced transparency setting, I don't see a realistic way of supporting an explicit override from the JavaFX side. Most backdrop effects are controlled by the OS and can not be modified freely. I see a lot of possible work + issues but don't see a lot of use for manually handling the reduced transparency setting in case it is manually overriden in a JavaFX app. In 99% of cases, the OS setting is what the user wants and gets.

In general, backdrops can't really be modified a lot. On Windows, the Mica effect forces you into a fixed color scheme dictacted by the OS theme, you can't even set the color. Apart from Mica, which has no transparency, there is also the older Acrylic effect of Windows 10, but that isn't really used anymore. So on Windows, transparency isn't really a thing in backdrops any more.

On macOS 27, the transparency of e.g. the liquid glass effect is controlled by a separate liquid glass opacity slider in the settings, This is a change from macOS 26. There is also a separate general reduced transparency setting in the macOS settings, so essentially you will have two OS settings affecting the end result. I also don't think an app can override those settings.

In general, you don't have to overthink this feature. Just giving users access to the standard OS backdrops will be enough for most. The point why people request this feature is to allow having a more native-looking app, not to build crazy customized custom window backdrops.

mstr2:
> > My main point is that `StageBackdrop` is quite closely related to `Paint`, which are shareable configuration objects created by users, and should therefore have a similar API.
> 
> And Paints are immutable by design.

True, it's not a perfect analogy.

> And you're again arguing based on your notion of API consistency and not utility. With your proposed API this call:
> 
> ```
> stage.getBackdrop().setOption("TintColor", Color.RED);
> ```
> 
> could change the tint color for multiple stages. Now that, that is unusual in the JavaFX API. And I think unexpected. Will developers consider this useful or behavior that they have to work around?

You think this is unexpected; I think this is exactly what should happen (and what I'd intuitively think _would_ happen based on how Java generally works for identity objects). If I share an object instance, then I expect changes to the object state be visible across all use sites. For example, if I had a global backdrop object, and I'd set it on every stage I create, then I'd also expect all stages to react when I change its configuration. I think I'd consider this a feature, not a bug.

> > Preferences are the user's choice (compared to the developer's choice when building the app), and should ideally be respected out of the box, without requiring developers to "opt in" by branching over preferences (setting a backdrop when reducedTransparency=false, and not setting a backdrop otherwise).
> 
> Again, I'm not that familiar with this API. The OS level setting for "reducedTransparency" is clearly the user's choice. I'm less clear on what it means to override the OS setting at the Scene level.

Overriding preferences at the Scene level is meant to give applications the capability to give users choice, for example by choosing a color scheme that's different from the OS. Applications could want to do this for a number of reasons: their UI could simply not be optimized for dark mode, or they might use legacy controls (maybe Swing controls) without a good dark mode integration. Other applications may simply want to allow users to choose the color scheme for just this application, which is generally a good idea because some users prefer to have individual applications with a color scheme different from the OS.

In this example, an application won't magically adapt by just changing the color scheme at the Scene level. This requires themes that actually offer distinct light/dark modes. Neither Caspian nor Modena do this, but future themes will (we've recently added CSS media queries just for this purpose).

So what does it mean to override OS preferences? It means that the theme and/or the application consult these preferences for their styling decisions, such that the preference is meaningfully reflected in the user experience. In another example, we recently had control-level animations consult the reducedMotion preference before playing. In yet another example, we've had the system-decorated title bar (the one you'll get with a `StageStyle.DECORATED` window) change its appearance based on the Scene color scheme. All of this works out of the box (as will a future theme with dark mode support).

The last example, where the system-decorated title bar adapts to the Scene color scheme, is quite comparable to a window backdrop. We tell the OS to change something under its control based on Scene-local user preference. So in the ideal case, I'd also want the same thing to happen for backdrops: when our window is in light mode, then we get the light backdrop. If it has reduced transparency, the backdrop won't be translucent. The question is how much of that is realistically achievable.

Let's just for a moment think about what happens if the window backdrop is completely decoupled from Scene preferences. If a theme supports both light and dark mode, and an application allows users to switch color schemes, this would mean that a user trying to have a light application color scheme on a dark OS scheme would most likely have a bad experience when the window uses a backdrop (because that implies dark text on a dark background).

> The OS backdrop selects a background color based on the window's state and various preferences and draws that beneath the Scene's fill color. This is the behavior that developers will have to design for and test against. For example, clients of the backdrop API will automatically be exposed to the to high and low contrast mode settings.
> 
> To honor a reduced transparency override we'll need to disable the OS backdrop which will also throw away its drawing of a background color. That will need to be replaced.
> 
> I've given this some thought. I was considering making backdrops "cross-platform" by emulating the background color drawing behavior on platforms that don't support backdrops natively. Because, again, the translucency is really beside the point in terms of development and testing. Win11 barely has any translucency.
> 
> So it's not just a question of reduced transparency. What does it mean for a client to override the system's background color in the Scene settings? In theory this should also mean disabling the OS backdrop since it can't honor that setting. It's not a question of how to choose a background color, it's a question of when we want to disable the OS backdrop. Whenever the client has overriding reducedTransparency? Whenever the client has overridden the Scene's background color? Whenever the window's color scheme doesn't match the platform's color scheme (at which point the OS will draw a background color that doesn't match the global background color preference)? Under these conditions either JavaFX needs to figure out the background color or we need to provide guidance to developers on what to do.

Well, it comes down to that, doesn't it? Either we come up with something, or we provide guidance to avoid the worst pitfalls. That being said, I'd really like it if the backdrop would at least be shown in the Scene's color scheme. But that doesn't seem to be possible on Windows with the `DWMWA_SYSTEMBACKDROP_TYPE` approach, it seems it would require using `MicaController` from the Windows App SDK. I'm not sure if we can depend on that.

mstr2:
> About the reduced transparency setting, I don't see a realistic way of supporting an explicit override from the JavaFX side. Most backdrop effects are controlled by the OS and can not be modified freely. I see a lot of possible work + issues but don't see a lot of use for manually handling the reduced transparency setting in case it is manually overriden in a JavaFX app. In 99% of cases, the OS setting is what the user wants and gets.
> 
> In general, backdrops can't really be modified a lot. On Windows, the Mica effect forces you into a fixed color scheme dictacted by the OS theme, you can't even set the color. Apart from Mica, which has no transparency, there is also the older Acrylic effect of Windows 10, but that isn't really used anymore. So on Windows, transparency isn't really a thing in backdrops any more.

If you use [MicaController](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.composition.systembackdrops.micacontroller?view=windows-app-sdk-2.0) from the Windows App SDK, you can pass in a [SystemBackdropConfiguration](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.composition.systembackdrops.systembackdropconfiguration?view=windows-app-sdk-2.0) object to toggle light/dark mode independently from the OS, or alternatively manually set TintColor, TintOpacity, LuminosityOpacity, and FallbackColor of the Mica material.

> On macOS 27, the transparency of e.g. the liquid glass effect is controlled by a separate liquid glass opacity slider in the settings, This is a change from macOS 26. There is also a separate general reduced transparency setting in the macOS settings, so essentially you will have two OS settings affecting the end result. I also don't think an app can override those settings.
> 
> In general, you don't have to overthink this feature. Just giving users access to the standard OS backdrops will be enough for most. The point why people request this feature is to allow having a more native-looking app, not to build crazy customized custom window backdrops.

Most software developers have no special accessibility needs, which is reflected in the software they create and its lack of consideration for people who do. We don't need to overthink this, but we need to _think_ before we act. There are obvious interactions of this proposed new feature with existing features; it may be that in the end, we'll just acknowledge those friction points and move on, but we can't skip the thinking.

crschnick:
> If you use [MicaController](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.composition.systembackdrops.micacontroller?view=windows-app-sdk-2.0) from the Windows App SDK, you can pass in a [SystemBackdropConfiguration](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.composition.systembackdrops.systembackdropconfiguration?view=windows-app-sdk-2.0) object to toggle light/dark mode independently from the OS, or alternatively manually set TintColor, TintOpacity, LuminosityOpacity, and FallbackColor of the Mica material.

Ah wow, I have never seen an app that uses that. If that is also available from the win32 API, then I retract my statement about backdrops not being configurable on Windows.

> Most software developers have no special accessibility needs, which is reflected in the software they create and its lack of consideration for people who do. We don't need to overthink this, but we need to think before we act. There are obvious interactions of this proposed new feature with existing features; it may be that in the end, we'll just acknowledge those friction points and move on, but we can't skip the thinking.

That is true. As long as everything at least honors the OS settings though wrt accessibility, it should be good enough

beldenfox:
I’ve updated this PR to replace `Stage.initBackdropStyle` with `Stage.initBackdrop`. Backdrops can be shared. Setting an option on a StageBackdrop should affect all the stages that reference that backdrop.

On Windows 11 when high contrast is on the OS stops drawing backdrops. With a dark high contrast theme Win11 will fill the entire window with a bright color like yellow or light blue. This color extends into the non-client area to form a band on three sides of the window. In the client area we need to cover this color with an appropriate background color which we normally avoid doing since we don’t want to cover the backdrop effect.

I’ve updated this PR to draw a background color in the client area when high contrast is on. The color is applied as the clear color so it’s drawn beneath the Scene’s fill. I refer to this as an emulated backdrop; it reproduces the basic color logic of a platform backdrop without the translucent effect.

I’ve updated the code so a developer can also use other translucent Paints as a scene’s fill. In previous versions of the code this only worked if the Paint was a Color. Now it also works if the Paint is, say, an ImagePattern. There was a bug inherited from the UNIFIED stage code which would cause non-Color Paints to draw incorrectly.

beldenfox:
I'm working on the CSR for this. Any comments on the documentation would be welcome.

Still unsure which platform-specific backdrops to include and whether to document them. On Windows 11 there are only three backdrop types. Two are covered by WINDOW and PARTIAL. The third is known as Transient and it's quite translucent (l wouldn't try to draw text on it). I haven't seen any place in the Win11 UI where it's used. The closest equivalent visually on macOS is HUD (Head's Up Display). Both work well when a semi-transparent tint is layered on top which is the only reason to I included them.

I would like to expose macOS.ClearGlass after we switch to the macOS 26 SDK. Beyond that I'm thinking of removing three macOS backdrops (Menu, ToolTip, and Popover) since they apply to PopupWindows and developers can't apply backdrops to those stages anyway. The only reason to include macOS.HUD is to match Windows.Transient.

mstr2:
> I'm working on the CSR for this. Any comments on the documentation would be welcome.

I recommend to create the CSR at the end of the review process, when all the details have settled.
This PR is on my radar, I'll get to it soon.

mstr2:
I find this new version of the API to be significantly cleaner and more intuitive.

However, I still don't understand the stringly-typed roundtrip when you call `StageBackdropStyle.getPlatformStyleNames()`, from which you then choose a name, and pass it back into `StageBackdropStyle.style(String)`. This seems like an unnecessary detour, because `StageBackdropStyle` instances are just immutable identifiers. You could directly return a list of all supported instances, from which users can choose one and then directly pass it into `new StageBackdrop(StageBackdropStyle)`.

In most cases, however, applications probably won't enumerate the available styles and pick one, but instead directly specify the one they want to use. For this purpose, I recommend to have `StageBackdropStyle.style(String)` not fail silently by returning `null`. Instead, returning an `Optional<StageBackdropStyle>` would make the API easier to use:

```java
var style = StageBackdropStyle.style("Windows.Transient")
    .or(() -> StageBackdropStyle.style("macOS.ClearGlass"))
    .orElse(StageBackdropStyle.WINDOW);

stage.initBackdrop(style);
```

beldenfox:
> However, I still don't understand the stringly-typed roundtrip when you call `StageBackdropStyle.getPlatformStyleNames()`, from which you then choose a name, and pass it back into `StageBackdropStyle.style(String)`.

As you said, most clients will specify the one they want by name and when searching for a backdrop they'll be comparing names no matter what API we provide. If a client wants a list of platform backdrops they can build that list easily enough. However it's an easy function to write and mirrors getStandardBackdrops so I'll add it. Are you proposing we also remove getPlatformStyleNames and just provide getPlatformBackdrops?

> For this purpose, I recommend to have `StageBackdropStyle.style(String)` not fail silently by returning `null`. Instead, returning an `Optional<StageBackdropStyle>` would make the API easier to use:

Thanks for the suggestion. Will do.

PR #2052, opened 2026-01-26
8325564: ComboBox popup does not correctly resize after update when on display

Ziad-Mid:
Fixed popup size not changing to show more items or less when changing the number of items in `ComboBox` by adding a `popupResize` variable which triggers a request layout when there is change in items list.
Tested with the test app in bug and MonkeyTester.

---------

---------
### Progress

### Issue
 * [JDK-8325564](https://bugs.openjdk.org/browse/JDK-8325564): ComboBox popup does not correctly resize after update when on display (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2052/head:pull/2052` \
`$ git checkout pull/2052`

`$ git checkout pull/2052` \
`$ git pull https://git.openjdk.org/jfx.git pull/2052/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2052`

`$ git pr show -t 2052`

</details>

<a href="https://git.openjdk.org/jfx/pull/2052.diff">https://git.openjdk.org/jfx/pull/2052.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2052#issuecomment-3801901673)
</details>


modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxPopupControl.java
```diff
@@ -556,6 +557,11 @@ void reconfigurePopup() {
         final Point2D p = getPrefPopupPosition();
 
         final Node popupContent = getPopupContent();
+        if (popupResize && popupContent instanceof Region) {
```
andy-goryachev-oracle: minor stylistic suggestion:
```
        if (popupResize && popupContent instanceof Region r) {
            r.setMinSize(Region.USE_COMPUTED_SIZE, Region.USE_COMPUTED_SIZE);
            r.setPrefSize(Region.USE_COMPUTED_SIZE, Region.USE_COMPUTED_SIZE);
        }
```

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxPopupControl.java
```diff
@@ -579,6 +587,15 @@ void reconfigurePopup() {
                 ((Region)popupContent).setPrefSize(newWidth, newHeight);
             }
         }
+
+        popupResize = false;
+    }
+
+    final void requestPopupLayout() {
```
andy-goryachev-oracle: misnomer: the name suggests it's a request, but in fact it does much more by calling
```
reconfigurePopup();
sizePopup();
```

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxPopupControl.java
andy-goryachev-oracle: also, look in `sizePopup()`, there is `if (popupContent instanceof Region) {` with the `else` clause which handles other cases.  

Should any possible fix be in `sizePopup()` instead?

andy-goryachev-oracle (changes requested):
The fix does not look right: I don't think the new flag is needed, but instead the popup should be properly sized.

On top of that, the fixed code does not work: the first update shows a tiny popup where most of the area is covered with scrollbars, and the second update (after 2 seconds) still displays scrollbars.  The scrollbars should not be there at all:

Image

Image

(on macOS retina, scale=2)

there is more: on an external monitor at scale=1, if I press the [Update] button then quickly click on the combo box, the popup is shown but very small, and it does not change:

Image

andy-goryachev-oracle (changes requested):
The last change is a step in the right direction.  Three issues:

1. Please merge the latest master, there was a change in sizePopup() which makes your code fail to compile.  In general, it's a good idea to merge the latest master if your PR is long going or there was a change in a related area.
2. The reproducer shows the popup in the body of the first `KeyFrame` via `comboBox.show();`.  Once the popup is shown, the second background update does not resize the popup and you see the scrollbars.
3. If you remove the `comboBox.show();` line, the expectation is that the popup will not be shown, but it is, resulting in this:

Image



Also, I've noticed the size of the popup control in the step 2 changes (also present in the master branch, so I guess it's expected)

andy-goryachev-oracle:
@crschnick would you be able to review this PR?

andy-goryachev-oracle:
Looks much better!

Two things:
1. please change the copyright year to 2026
2. noticed runLater() in `recomputePopupLayout()`, so I decided to apply CSS to the popup.  You can use
https://github.com/andy-goryachev-oracle/Test/blob/main/src/goryachev/bugs/ComboBox_PopupSize_8325564.java

On my primary retina display I get the scrollbars first time around.  The scrollbars disappear on second showing:

Screenshot 2026-02-02 at 15 05 29

```
.combo-box-popup > .list-view {
    -fx-padding: 100px
}
```

crschnick:
> @crschnick would you be able to review this PR?

Sorry, I didn't see this comment until now. I can take a look

Also what happened to bridgekeeper? That inactivity timeout looks wrong

crschnick:
@andy-goryachev-oracle Can I still review this?

andy-goryachev-oracle:
Let me check with @Ziad-Mid ...

kevinrushforth:
> Also what happened to bridgekeeper? That inactivity timeout looks wrong

Yes, this looks like a hiccup of some sort. The initial "inactive for more than 4 weeks and will be automatically closed if another 4 weeks passes..." was correct. The "hey it's been 4 more weeks" auto-close one hour later was not. I'm reopening it.

andy-goryachev-oracle:
... and I think @Ziad-Mid is still working on it, https://github.com/openjdk/jfx/pull/2052#issuecomment-3837743037 has not been addressed.

Ziad-Mid:
> ... and I think @Ziad-Mid is still working on it, [#2052 (comment)](https://github.com/openjdk/jfx/pull/2052#issuecomment-3837743037) has not been addressed.

Yes, I still do

andy-goryachev-oracle:
So I think the problem identified in https://github.com/openjdk/jfx/pull/2052#issuecomment-3837743037 is still there.

I've updated the test not only to add items but also remove them (making the list smaller, just in case) - see the latest version
https://github.com/andy-goryachev-oracle/Test/blob/main/src/goryachev/bugs/ComboBox_PopupSize_8325564.java

to reproduce,
- click the combobox to show the popup,
- click elsewhere to dismiss the popup
- click update button
- open the popup again

Expecting to see no scrollbar, but:

Screenshot 2026-04-22 at 12 12 08

andy-goryachev-oracle (changes requested):
still see the scrollbar.

crschnick:
In general I would say that if you have to rely on using Platform.runLater for some kind of layouting and size changes within a Control or Skin, the approach is usually not the right one. I would try to replace the runLater call with another implementation that fixes the issue as well somehow.

Ziad-Mid:
Tested with latest test from Andy and it works.

crschnick:
While fixing https://bugs.openjdk.org/browse/JDK-8338145, I found the cause to be the fact that the popup did not have the right size computed before the first show. That is why the fix calls `sizePopup()` twice. Maybe this issue is related and somehow eliminates the need to call sizePopup() twice?

andy-goryachev-oracle:
> While fixing https://bugs.openjdk.org/browse/JDK-8338145, I found the cause to be the fact that the popup did not have the right size computed before the first show. That is why the fix calls `sizePopup()` twice. Maybe this issue is related and somehow eliminates the need to call sizePopup() twice?

A good question.  There seems to be something wrong in the overall logic (probably pre-existing) in regards to CSS.  Can we even applyCss() before the popup is shown?  Wouldn't it impact the styles if there exists more complex dependency on the parent window?

Another scenario is when the stylesheet changes mid-flight, in one of the animation frames, see https://github.com/andy-goryachev-oracle/Test/blob/main/src/goryachev/bugs/ComboBox_PopupSize_8325564.java L40.

The popup padding is not updated if the popup is already shown:

Screenshot 2026-06-01 at 14 56 07

but if the user hides the popup and then re-opens it, the padding is correct:
Screenshot 2026-06-01 at 14 56 38

Again, this might be a sign of some deeper problem (or possibly a rare corner case).  The fix looks weird, I admit, but at least it works correctly, seemingly without any flicker.
I am inclined to call it "good enough".

andy-goryachev-oracle (approved):

andy-goryachev-oracle:
@Ziad-Mid you might want to sync up with the master branch, this PR has been out for far too long.

andy-goryachev-oracle:
@crschnick or @arapte could one of you be the second reviewer?

crschnick:
Yes, I can do that. Give me a few days

arapte:
The fix looks good overall. Tested with a few scenarios with a sample program.
Could you please include a system headful test. 
suggestion for test steps:
1. Create a Stage with size such a ComboBox popup of 10 items does not exceed Stage.
2. Add a ComboBox with 10 items
3. Pick the color of stage background
4. Open ComboBox popup and pick color at item 5 (Do not move mouse)
5. Set only 2 items to ComboBox
6. Pick color again
=> Color in step 6 should be = stage background color and != color picked in step 4

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxPopupControl.java
```diff
@@ -25,6 +25,7 @@
 
 package javafx.scene.control.skin;
 
+import javafx.application.Platform;
```
crschnick, line 28: Unused import

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxPopupControl.java
```diff
@@ -467,6 +472,8 @@ private void sizePopup(boolean first) {
         if (popupContent instanceof Region) {
             // snap to pixel
             final Region r = (Region) popupContent;
+            r.setMinSize(Region.USE_COMPUTED_SIZE, Region.USE_COMPUTED_SIZE);
```
crschnick, line 475: What if someone has bound the size of the popup content from the outside? Then this will fail

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxPopupControl.java
```diff
@@ -467,6 +472,8 @@ private void sizePopup(boolean first) {
...
+            r.setPrefSize(Region.USE_COMPUTED_SIZE, Region.USE_COMPUTED_SIZE);
```
crschnick, line 476: Setting those sizes looks to be the wrong approach. I can see that without it, this PR does no longer work, but there should a more straightforward way of getting the right height in sizePopup() than to change the sizes like that. For example, why is the prefHeight not properly updating in this method without those calls?

crschnick (changes requested):
So while the fix works, I still think that it is not the right solution. Manually changing the size of the popup content should not be necessary imo when done right

FlorianKirmaier:
As mentioned in the mailing - the PR fixes a rendering bug I've encountered with the latest JavaFX Versions.
Regarding the PR, i really miss a Unit-Test. This would really make sure we fully understand the issue - and a regression won't happen anymore. It would also make it much easier to investigate other solutions.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxPopupControl.java
FlorianKirmaier: I agree. Usually the code shouldn't mutate these variables at all.
This can easily run into conflict with custom css.

PR #2068, opened 2026-02-08
8377426: Adjust scene background to color scheme

mstr2:
When a non-opaque scene fill color is used with a stage style other than `StageStyle.TRANSPARENT`, the actual fill color is always white. This doesn't work well when the scene uses a dark color scheme. A practical solution is to allow non-opaque scene fill colors, and blend them on top of a white or black background (depending on color scheme) to derive an opaque color that adapts intuitively to the color scheme.

To test this, simply create a scene that uses a non-opaque fill color and observe the scene background when the color scheme is changed.

This PR includes a system test, run it with:
```
./gradlew -PFULL_TEST=true -PUSE_ROBOT=true :systemTests:test --tests test.robot.javafx.scene.SceneFillTest.testSceneFill
```

/reviewers 2

---------
### Progress

### Error

### Warnings
&nbsp;⚠️ Patch contains a binary file (modules/javafx.graphics/src/test/resources/test/com/sun/javafx/image/alpha-hole-green-frame-64x64.png)
&nbsp;⚠️ Patch contains a binary file (modules/javafx.graphics/src/test/resources/test/com/sun/javafx/image/blue-bg-soft-red-circles-64x64.png)
&nbsp;⚠️ Patch contains a binary file (modules/javafx.graphics/src/test/resources/test/com/sun/javafx/image/layered-landscape-foliage-dominant-96x64.png)
&nbsp;⚠️ Patch contains a binary file (modules/javafx.graphics/src/test/resources/test/com/sun/javafx/image/near-greens-majority-skew-64x64.png)
&nbsp;⚠️ Patch contains a binary file (modules/javafx.graphics/src/test/resources/test/com/sun/javafx/image/olive-majority-confetti-noise-1024x1024.png)
&nbsp;⚠️ Patch contains a binary file (modules/javafx.graphics/src/test/resources/test/com/sun/javafx/image/orange-field-purple-dots-64x64.png)
&nbsp;⚠️ Patch contains a binary file (modules/javafx.graphics/src/test/resources/test/com/sun/javafx/image/teal-field-black-speckle-64x64.png)
&nbsp;⚠️ Patch contains a binary file (modules/javafx.graphics/src/test/resources/test/com/sun/javafx/image/transparent-amber-overlay-cutouts-64x64.png)
&nbsp;⚠️ Patch contains a binary file (modules/javafx.graphics/src/test/resources/test/com/sun/javafx/image/transparent-teal-overlay-slits-64x64.png)
&nbsp;⚠️ Patch contains a binary file (modules/javafx.graphics/src/test/resources/test/com/sun/javafx/image/transparent-violet-overlay-windows-64x64.png)
&nbsp;⚠️ Patch contains a binary file (modules/javafx.graphics/src/test/resources/test/com/sun/javafx/image/warm-cool-diagonal-gradient-96x96.png)
&nbsp;⚠️ Patch contains a binary file (modules/javafx.graphics/src/test/resources/test/com/sun/javafx/image/white-bg-red-circle-64x64.png)

### Issue
 * [JDK-8377426](https://bugs.openjdk.org/browse/JDK-8377426): Adjust scene background to color scheme (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2068/head:pull/2068` \
`$ git checkout pull/2068`

`$ git checkout pull/2068` \
`$ git pull https://git.openjdk.org/jfx.git pull/2068/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2068`

`$ git pr show -t 2068`

</details>

<a href="https://git.openjdk.org/jfx/pull/2068.diff">https://git.openjdk.org/jfx/pull/2068.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2068#issuecomment-3866056960)
</details>


modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
```diff
@@ -376,6 +376,8 @@ private Scene(Parent root, double width, double height, Paint fill,
 
         addEventFilter(MouseEvent.MOUSE_PRESSED, pressedHandler);
         addEventFilter(TouchEvent.TOUCH_PRESSED, pressedHandler);
+
+        context.colorSchemeProperty().addListener((_, _, newValue) -> markDirty(DirtyBits.DARK_SCHEME_DIRTY));
```
andy-goryachev-oracle: `newValue` is not used, should it be an invalidation listener instead?

andy-goryachev-oracle:
hmm, something is wrong:
```
> Task :systemTests:test

SceneFillTest > testSceneFill(String, String, String) > [3] rgba(0, 0, 0, 0.75), rgb(63, 63, 63), BLACK FAILED
    org.opentest4j.AssertionFailedError: expected: <0x3f3f3fff> but was: <0x404040ff>
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
        at app//org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1145)
        at app//test.robot.javafx.scene.SceneFillTest.lambda$testSceneFill$1(SceneFillTest.java:96)

SceneFillTest > testSceneFill(String, String, String) > [4] rgba(127, 127, 127, 0.5), rgb(191, 191, 191), rgb(63, 63, 63) FAILED
    org.opentest4j.AssertionFailedError: expected: <0x3f3f3fff> but was: <0x404040ff>
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
        at app//org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1145)
        at app//test.robot.javafx.scene.SceneFillTest.lambda$testSceneFill$2(SceneFillTest.java:106)

SceneFillTest > testSceneFill(String, String, String) > [6] rgba(0, 100, 0, 0.8), rgb(50, 131, 50), rgb(0, 80, 0) FAILED
    org.opentest4j.AssertionFailedError: expected: <0x328332ff> but was: <0x328333ff>
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
        at app//org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1145)
        at app//test.robot.javafx.scene.SceneFillTest.lambda$testSceneFill$1(SceneFillTest.java:96)

6 tests completed, 3 failed

> Task :systemTests:test FAILED

[Incubating] Problems report is available at: file:///Users/angorya/Projects/jfx3/jfx/rt/build/reports/problems/problems-report.html

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':systemTests:test'.
> There were failing tests. See the report at: file:///Users/angorya/Projects/jfx3/jfx/rt/tests/system/build/reports/tests/test/index.html

* Try:
> Run with --scan to get full insights from a Build Scan (powered by Develocity).

Deprecated Gradle features were used in this build, making it incompatible with Gradle 10.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

For more on this, please refer to https://docs.gradle.org/9.3.0/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.

BUILD FAILED in 1m 14s
256 actionable tasks: 125 executed, 131 up-to-date
dhcp-10-89-197-66:rt angorya$ git status
On branch pull/2068
```

mstr2:
It seems like the sampled color is off by a tiny bit. Are you running the test on macOS? I wonder why the rendered colors would _almost_ match, but be ever-so-slightly different compared to Windows.

andy-goryachev-oracle:
yes, macOS 26.2
the small difference is probably due to the color space, I think we had similar issues before.

also, on mac (edit: in light mode), I don't see the white background.  what I do see is the background flickering momentarily to black before going to the set color, here is the test code (I might be doing something wrong though):

https://github.com/andy-goryachev-oracle/Test/blob/main/src/goryachev/bugs/Stage_Background_8377426.java

andy-goryachev-oracle:
In the dark mode, I do see the screen flickering to white before going to the set color, using the same test with lighter color:
```
        sc.setFill(Color.rgb(255, 255, 255, 0.2));
```

(the master branch simply goes full white w/o flicker in the dark mode)

beldenfox:
When writing a value to the screen and reading it back the components can shift a bit. There's a utility for comparing two colors (assertColorEquals) that allows for a tolerance when comparing the components. You can see sample code for this in SRGBTest.java.

JavaFX specifies colors in the sRGB space which employs a gamma (power) curve. For any given color component Prism presumably converts it to a linear space, performs blending operations, converts back to sRGB, then rounds the result to a byte to hand it off to the OS. Most Macs displays don't use sRGB natively so the byte then gets converted to the display's color space using more math and float <=> byte conversions. Then when the Robot reads the color back a bunch of this process is reversed.

This PR is blending two colors in the sRGB space instead of a linear space. That's a little odd but it seems that other parts of the system do this (I see the Color.interpolate method works in sRGB) so maybe we'll just let it pass.

beldenfox:
> In the dark mode, I do see the screen flickering to white before going to the set color, using the same test with lighter color:
> 
> ```
>         sc.setFill(Color.rgb(255, 255, 255, 0.2));
> ```

Good catch. When the stage is first shown JavaFX takes the scene's fill color and passes it off to the OS to become the window's background color. In the process it throws away the alpha component so here it's setting the background to white. That's what you'll see between the time the window is first made visible and when the rendering thread delivers pixels to the window. You'll also see flashes of the background color while resizing the window. See the calls to platformWindow.setBackground in WindowStage.java.

The Windows code throws this information away. macOS and Linux forward the color on to the platform. Fun fact: you can use this feature to tint the title bar on macOS.

mstr2:
I've changed the implementation to get rid of the "white flash" that can sometimes be observed. For that purpose, I'm trying to find a solid color that best matches the actual paint that JavaFX will later render. This solid color is then passed to the OS to be the window's background color. After the first frame is rendered, this background color is no longer visible because JavaFX draws the entire background.

Finding a color that is representative of the actual JavaFX paint is done as follows:
1. `Color`: Blend the color on top of a white or black background, depending on the scene's color scheme.
2. `LinearGradient`, `RadialGradient`: Blend the gradient on top of a white or black background, and sample it at various locations to compute the average color.
3. `ImagePattern`: Compute the dominant color of the image by approximating its color palette with a clustering algorithm, and then select the color cluster with the most pixels in it. For large images, the runtime cost of this algorithm can be quite high, so its fidelity is throttled depending on the input image to not exceed 32000 samples.

Blending is done in sRGB space, as that's what the JavaFX rendering pipeline will do.

With this change, you shouldn't observe a wrong background before the first frame is rendered in most cases.

andy-goryachev-oracle:
The last change brings, in my opinion, an unreasonable amount of fairly complicated code for something that can only be observed momentarily.  Overkill?

The main issue, if I understand it correctly, is that the background flashes with a wrong color for the given color scheme, i.e. white in the dark mode.  Could we fix just that?  I think it might be ok to flash black/white for dark/light mode correspondingly.  Or use the panel background color from the theme, if it is available at this point?

What do you think?

mstr2:
> The last change brings, in my opinion, an unreasonable amount of fairly complicated code for something that can only be observed momentarily. Overkill?

Depends on your aesthetic sensitivities, I guess. It's true that it is a fair bit of code to solve this, but on the other hand, that brief flash does look a bit unpolished and cheap. And depending on the styling of the application, it's almost impossible to work around it. Personally, I often find that that last bit of polish separates a good application from a great one.

> The main issue, if I understand it correctly, is that the background flashes with a wrong color for the given color scheme, i.e. white in the dark mode. Could we fix just that? I think it might be ok to flash black/white for dark/light mode correspondingly. Or use the panel background color from the theme, if it is available at this point?

Setting the background color to white or black works for applications whose backgrounds are close to white or black. The flash gets noticeably worse the more the application background deviates from white or black, and especially so if colors are involved.

andy-goryachev-oracle:
Thank you for clarifying.  I accept the aesthetic reasoning, though still feel the amount of code you've added specifically for the ~200 ms of startup might be just a bit too much.  Can it be useful in other places?

Would it be possible to alter https://github.com/andy-goryachev-oracle/Test/blob/main/src/goryachev/bugs/Stage_Background_8377426.java perhaps to allow for executing the different use cases?  To help with the review process?

mstr2:
> Thank you for clarifying. I accept the aesthetic reasoning, though still feel the amount of code you've added specifically for the ~200 ms of startup might be just a bit too much. Can it be useful in other places?

`PaintSampler` might be useful as it allows you to analyze gradients without actually rendering them, but I have no concrete plan to use it for another feature. In general, I'd like JavaFX support different color spaces in the future (color transitions tend to look much better in Oklab or linear sRGB), for which we'd probably reuse many of the color space transformations that currently live in `ImageUtils`.

> Would it be possible to alter https://github.com/andy-goryachev-oracle/Test/blob/main/src/goryachev/bugs/Stage_Background_8377426.java perhaps to allow for executing the different use cases? To help with the review process?

Here's a program with some examples:

```java
public class SceneFillTest extends Application {

    @Override
        var scene = new Scene(new VBox(), 800, 600);

        // Transparent black/white
        scene.setFill(Color.rgb(0, 0, 0, 0.5));
        //scene.setFill(Color.rgb(255, 255, 255, 0.5));

        // Stark linear gradient: transparent to black
        // scene.setFill(new LinearGradient(
        //     0, 0, 0, 1, true, CycleMethod.NO_CYCLE,
        //     new Stop(0, Color.TRANSPARENT), new Stop(1, Color.BLACK)));

        // Soft linear gradient with colors
        // scene.setFill(new LinearGradient(
        //     0, 0, 0, 1, true, CycleMethod.NO_CYCLE,
        //     new Stop(0, Color.RED), new Stop(1, Color.ORANGE)));

        // Image pattern
        // scene.setFill(new ImagePattern(new Image("path/to/bg1.png")));

        stage.setScene(scene);
        stage.show();
    }
}
```

And here are some background images, one with transparency and one without, which you can use to test the image pattern fill: [bg1.png](https://github.com/user-attachments/assets/748d2ed8-f9a1-42fd-a75f-8d4282c7bc22), [bg2.png](https://github.com/user-attachments/assets/412d3015-c55f-4052-8496-5dc1d4864d8f)

You should observe that with solid backgrounds, there should be no visible flash. Stark gradients have no really good single-color approximation, the algorithm will pick the average color of the gradient for the initial background. Soft gradients tend to work well, also notice that there is no jarring change from black/white to color.

With images, the initial color is the dominant color of the image, composited on the window background. Also make sure to toggle dark/light mode while the program is running to see the (partially) transparent scene fills adjust to the color scheme.

andy-goryachev-oracle:
Thank you for providing examples, that made my life easier.

I've updated https://github.com/andy-goryachev-oracle/Test/blob/main/src/goryachev/bugs/Stage_Background_8377426.java with your examples, below are the results:

**Master Branch**

Light Mode
0 flashes black
1 flashes red
2 flashes black
3 flashes red
4 flashes black
5 flashes black

Dark Mode
0 flashes black
1 flashes red
2 flashes black
3 flashes red
4 flashes black
5 flashes black


**PR #2068**

Light Mode
0 nf
1 nf
2 flashes black ??
3 flashes red
4 nf
5 nf

Dark Mode
0 nf
1 nf
2 nf
3 nf
4 nf
5 nf
6 flashes white?

nf - no flashing, or flashes too fast to figure out the color


Well, it does look better (if for a split second).  I still not sure if it's worth all the additional code, but at least it's java and not native.

Do you know why it has to flash in the first place?  Why can't it render the content and _then_ show it?

mstr2:
> I've updated https://github.com/andy-goryachev-oracle/Test/blob/main/src/goryachev/bugs/Stage_Background_8377426.java with your examples, below are the results:

The flashes you're still observing are probably not black or white, but a mid-grey (which is the average color of the linear gradient). With gradients where one end is really bright and the other end is really dark, there's no good single-color approximation.

> Well, it does look better (if for a split second). I still not sure if it's worth all the additional code, but at least it's java and not native.
> 
> Do you know why it has to flash in the first place? Why can't it render the content and _then_ show it?

The reason is that we ask the OS to open up a window, and after the window is shown, JavaFX begins to render the first image. So there's a natural delay between the window opening up and the arrival of the first pixels. I'm not aware of a way to reliably delay the window until the first frame is ready, because we need a window buffer to render the image (a D3D swap chain, or a Metal drawable), and that buffer is usually not available until the window is shown.

Another variant of the same problem is live resizing of a window, where you can sometimes observe that the rendered content lags behind the size of the window. On Windows, this effect is visible in almost all applications.

modules/javafx.graphics/src/main/java/com/sun/javafx/image/ImageUtils.java
```diff
@@ -0,0 +1,443 @@
...
+        int w = (int)Math.floor(image.getWidth());
+        int h = (int)Math.floor(image.getHeight());
+        long numPixels = (long)w * (long)h;
+
+        // Choose sampleStep so that (w/step)*(h/step) <= maxSamples
+        int sampleStep;
+        if (numPixels <= (long) maxSamples) {
+            sampleStep = 1;
+        } else {
+            sampleStep = (int)Math.ceil(Math.sqrt((double)numPixels / (double)maxSamples));
```
andy-goryachev-oracle: `(double)numPixels`
typecast is not needed

andy-goryachev-oracle (commented):
This works well on macOS 26.2 (did not test windows/linux).

My biggest gripe is the amount of extra code and effort brought in to slightly improve the flickering color.

In fact, I suspect the added computation might actually exacerbate the problem in the case of image, since it has to compute the background color using non-trivial and time-consuming algorithms.

I would think it might work just as well simply by using the black/white background corresponding to the current theme.  I am curious what other reviewers would say.

mstr2:
> This works well on macOS 26.2 (did not test windows/linux).
> 
> My biggest gripe is the amount of extra code and effort brought in to slightly improve the flickering color.
> 
> In fact, I suspect the added computation might actually exacerbate the problem in the case of image, since it has to compute the background color using non-trivial and time-consuming algorithms.
> 
> I would think it might work just as well simply by using the black/white background corresponding to the current theme. I am curious what other reviewers would say.

It's a valid question. I will point out, however, that the runtime cost shouldn't be too high, as the sampling fidelity is reduced for large images (small images are sampled more often to capture more detail, large images are sampled less often because their details are usually spread over a larger area).

There are many examples where just black or white works well enough. But playing around with it, I can just as easily come up with examples where it doesn't work so well, especially with a colorful background.

beldenfox:
The window background color is only sent to the OS once when the window is first shown. When the color scheme changes the OS window background isn't updated so you can get resize flashing again at least on macOS.

When a window is expanded additional space is made along the right and bottom edges. You might get a better color match by sampling the fill along one of these edges or only in the lower right corner of the window.

mstr2:
> The window background color is only sent to the OS once when the window is first shown. When the color scheme changes the OS window background isn't updated so you can get resize flashing again at least on macOS.

Right, I've fixed that.

> When a window is expanded additional space is made along the right and bottom edges. You might get a better color match by sampling the fill along one of these edges or only in the lower right corner of the window.

Maybe, but this could also be not as good a match for the initial background color, which can be observed over the entire area of the window.

andy-goryachev-oracle:
> There are many examples where just black or white works well enough.

that's my point.  

The big issue (a white flash is the dark mode) can be easily solved with a simple and clean code, a definite win and a no-brainer.  Pursuing rare corner cases at the cost of introducing a large chunk of code ... not so.

beldenfox:
That PR might be missing some common cases. For example, a developer may want the scene's fill color to track the background color in PlatformPreferences. When that pref changes and becomes the new fill we should make sure the color (which is opaque) gets sent on to the OS.

It also bothers me that a developer has to rely solely on our built-in heuristics when the scene's fill is an image. What if a developer wants to specify a better color directly? Would it make sense to add a field to ImagePattern (perhaps using the heuristic in this PR as a fallback)? Or maybe provide a public API for setting the window's background color independently of the scene's fill?

I like that this PR tries to make JavaFX apps look good by default but we should at least think about providing API's so a motivated developer can solve the problem directly and not rely on our heuristics.

mstr2:
> That PR might be missing some common cases. For example, a developer may want the scene's fill color to track the background color in PlatformPreferences. When that pref changes and becomes the new fill we should make sure the color (which is opaque) gets sent on to the OS.

What's missing? When either `Scene.fill` or `Preferences.colorScheme` changes, the window background color is updated and sent to the OS (see `WindowStage` lines 109, 117).

> It also bothers me that a developer has to rely solely on our built-in heuristics when the scene's fill is an image. What if a developer wants to specify a better color directly? Would it make sense to add a field to ImagePattern (perhaps using the heuristic in this PR as a fallback)? Or maybe provide a public API for setting the window's background color independently of the scene's fill?
> 
> I like that this PR tries to make JavaFX apps look good by default but we should at least think about providing API's so a motivated developer can solve the problem directly and not rely on our heuristics.

If a developer wants to specify their own color, they can surely just set the scene fill to exactly that color, and then place a layout container with an `ImagePattern` background at the root of the scene graph. Nothing here seems so complicated that it would warrant new API.

beldenfox:
> What's missing? When either `Scene.fill` or `Preferences.colorScheme` changes, the window background color is updated and sent to the OS (see `WindowStage` lines 109, 117).

Sorry about that. I was playing with a test app and misinterpreted what I was seeing. I seem to have encountered a bug. I'll add a review comment to the code.

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/WindowStage.java
```diff
@@ -86,15 +91,31 @@ public class WindowStage extends GlassStage {
...
+                    if (platformWindow != null) {
+                        platformWindow.setDarkFrame(colorScheme == ColorScheme.DARK);
+                        updatePlatformWindowBackground();
+                    }
+                });
+
+            fxStage.sceneProperty()
+                .flatMap(Scene::fillProperty)
+                .subscribe(() -> {
+                    QuantumToolkit.getToolkit().checkFxUserThread();
```
beldenfox, line 116: This isn't working reliably for me. I have a test app that sets the scene's fill to different Colors. The first time I set the fill this line is triggered but after that it doesn't fire even though the scene's fill property is being invalidated.

If I change this to `fxStage.getScene().fillProperty().subscribe(() -> {});` it works consistently.

modules/javafx.graphics/src/main/java/com/sun/javafx/util/Utils.java
```diff
@@ -217,6 +221,69 @@ public static boolean contains(String src, String s) {
...
+                PaintSampler sampler = GradientUtils.newLinearGradientSampler(g, bounds);
+                yield calculateAverageSRGB(sampler, 32, background, bounds);
+            }
+
+            case RadialGradient g -> {
+                PaintSampler sampler = GradientUtils.newRadialGradientSampler(g, bounds);
+                yield calculateAverageSRGB(sampler, 32, background, bounds);
+            }
+
+            case ImagePattern pattern -> ImageUtils.computeDominantColor(pattern.getImage(), background);
```
beldenfox, line 244: I know JavaFX has the notion of background image loading. Is that relevant here? This code is changing the timing; it will access the image's pixels before the stage is even shown instead of waiting for a render pass.

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/WindowStage.java
mstr2: Good catch, I've fixed this.

modules/javafx.graphics/src/main/java/com/sun/javafx/util/Utils.java
mstr2: `ImagePattern` doesn't accept background-loading images, so the problem doesn't exist.

modules/javafx.graphics/src/main/java/com/sun/javafx/image/ImageUtils.java
```diff
@@ -0,0 +1,443 @@
...
+
+        // Guard against too many clusters for too few points to reduce empty clusters.
+        int upper = Math.max(1, nEst / 300);
+        clusters = Math.min(clusters, upper);
+        clusters = Math.max(1, Math.min(clusters, Math.min(maxClusters, nEst)));
+
+        return computeDominantColor(image, background, clusters, sampleStep, 0);
+    }
+
+    /**
```
beldenfox: I'm onboard with the rest of this PR but this big chunk of code is above my pay grade. Is the conversion from sRGB to CIELAB correct? Is it a correct implementation of k-means clustering? Is it even valid to perform clustering for CIELAB values? Rather than try to answer these questions I generally go get a snack and put off the review for another week.

mstr2: Good questions.

> Is the conversion from sRGB to CIELAB correct?

Well, let's take [CSS Color 4](https://www.w3.org/TR/css-color-4/#predefined-to-lab-oklab), which specifies the following conversion sequence: decode sRGB -> linear RGB -> XYZ(D65) -> adapt to D50 -> Lab(D50).

This implementation is a bit simpler: decode sRGB -> linear RGB -> XYZ(D65) -> Lab(D65). It skips the D65->D50 adaptation step, and computes Lab directly using D65 reference-white constants. That's not entirely correct from a standards perspective, but it is internally consistent: forward and inverse conversions both use D65. I'd say that's generally good enough for our purpose.

> Is it a correct implementation of k-means clustering?

Yes, it first initializes the centers with k-means++, and the optimization loop is Lloyd's algorithm: assign each point to the nearest center, replace each center by the mean of its assigned points, repeat until convergence.

> Is it even valid to perform clustering for CIELAB values?

It's valid and a very good idea™. [MathWorks](https://www.mathworks.com/content/dam/mathworks/tag-team/Objects/c/88360_93001v00_Color-Based_Seg_K-Means_Clustering_2016.pdf) has a clustering example where the first step is "Converting the Image from the RGB Color Space to the L*a*b* Color Space".

The reason why it's a good idea is that Euclidean distance in CIELAB is roughly perceptual, much more so than RGB. And since our optimization algorithm minimizes Euclidean distance, it's a good fit for it.


That being said, and looking at this code again, the cluster-selection heuristic might be a bit weak:
```java
int upper = Math.max(1, nEst / 300);
clusters = Math.min(clusters, upper);
```
This collapses to a single cluster for nEst < 600, which can probably happen in the real world for a very small background image that has two distinct regions (maybe very bright and very dark). If we only select a single cluster, we end up with muddy gray. This might not be what we want (or it might be what we want... after all, there's no _correct_ way to do this).

andy-goryachev-oracle: Is there any way to allow the application to control that somehow?  This is complicated code that I think is not necessary in 90% of common cases.

beldenfox: My main point is that it's a big commitment to review this code. Even if you're familiar with the color conversions and clustering algorithms involved (I'm not) it's just a lot of code to go over.

It's helpful to have vetted sample code to compare to so thanks for the pointer to the w3 page on color conversions. Were you working with a similar reference algorithm for the k-means clustering code?

> > Is it even valid to perform clustering for CIELAB values?
> 
> It's valid and a very good idea™. [MathWorks](https://www.mathworks.com/content/dam/mathworks/tag-team/Objects/c/88360_93001v00_Color-Based_Seg_K-Means_Clustering_2016.pdf) has a clustering example where the first step is "Converting the Image from the RGB Color Space to the L_a_b* Color Space".

The MathWorks example you pointed to seems to be clustering in 2D using only the chroma components. When I searched for this most sample code was doing the same thing and handling luminance separately. However according to Wikipedia (which is, of course, infallible) there's precedent for performing clustering with luminance included.

mstr2:
I've refactored the dominant-color algorithm to hopefully make it easier to review:
1. Color conversion is factored out into the `ColorConversion` class, including independent unit tests.
2. K-means is factored out into the `KMeans3` class, including independent unit tests. I've also included a large testing dataset that I've created with a third-party tool to verify that the algorithm works.
3. The cluster-parameter selection heuristic is completely removed. Instead, we now always use four color clusters.
4. I've added a variety of sample images to test the algorithm, and it seems to do a good job on those images.

These changes make the actual dominant color algorithm quite simple, and easier to reason about in isolation. Note that we don't specify how the algorithm works, and we make no guarantees that it'll always find the best background color.

> Is there any way to allow the application to control that somehow? This is complicated code that I think is not necessary in 90% of common cases.

Yes, just set your own solid background color. That always works, and is the best choice for developers who want absolute predictability.

PR #2091, opened 2026-03-01
8378895: Reduce object allocations in Renderer.getPeerInstance()

mstr2:
The current implementation of `Renderer.getPeerInstance()` looks up mappings by concatenating strings to form a combined key:
```java
peer = peerCache.get(name + "_" + unrollCount);
```

This can be improved by not using strings to store the combined key, but using a simple `PeerCacheKey` class instead.
The `Renderer` class then uses a single reusable `PeerCacheKey` object to look up mappings, making the lookup allocation-free.

/reviewers 2

---------

---------
### Progress

### Issue
 * [JDK-8378895](https://bugs.openjdk.org/browse/JDK-8378895): Reduce object allocations in Renderer.getPeerInstance() (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2091/head:pull/2091` \
`$ git checkout pull/2091`

`$ git checkout pull/2091` \
`$ git pull https://git.openjdk.org/jfx.git pull/2091/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2091`

`$ git pr show -t 2091`

</details>

<a href="https://git.openjdk.org/jfx/pull/2091.diff">https://git.openjdk.org/jfx/pull/2091.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2091#issuecomment-3979006672)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
@@ -83,7 +83,7 @@ public abstract class Renderer {
     }
 
     public static final String rootPkg = "com.sun.scenario.effect";
-    private static final Map<FilterContext, Renderer> rendererMap = new HashMap<>(1);
+    private static final Map<FilterContext, Renderer> rendererMap = HashMap.newHashMap(1);
 
     private static final class PeerCacheKey {
         String name;
--- a/modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -84,10 +84,37 @@ public abstract class Renderer {
     }
 
     public static final String rootPkg = "com.sun.scenario.effect";
-    private static final Map<FilterContext, Renderer> rendererMap =
-        new HashMap<>(1);
-    private Map<String, EffectPeer> peerCache =
-        Collections.synchronizedMap(new HashMap<String, EffectPeer>(5));
+    private static final Map<FilterContext, Renderer> rendererMap = new HashMap<>(1);
+
+    private static final class PeerCacheKey {
+        String name;
+        int unrollCount;
+
+        PeerCacheKey(String name, int unrollCount) {
+            this.name = name;
+            this.unrollCount = unrollCount;
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            return o instanceof PeerCacheKey other
+                && unrollCount == other.unrollCount
+                && name.equals(other.name);
+        }
+
+        @Override
+        public int hashCode() {
+            return 31 * name.hashCode() + unrollCount;
+        }
+
+        @Override
+        public String toString() {
+            return name + unrollCount;
+        }
+    }
+
+    private final Map<PeerCacheKey, EffectPeer> peerCache = Collections.synchronizedMap(new HashMap<>(5));
+    private final PeerCacheKey peerCacheKey = new PeerCacheKey(null, -1); // reusable key to look up mappings
     private final ImagePool imagePool;
 
     protected static final boolean verbose = Boolean.getBoolean("decora.verbose");
@@ -234,36 +261,23 @@ public abstract class Renderer {
      * @param unrollCount
      * @return cached peer for this name and unroll count
      */
-    public final synchronized EffectPeer
-        getPeerInstance(FilterContext fctx, String name, int unrollCount)
-    {
-        // first look for a previously cached peer using only the base name
-        // (e.g. GaussianBlur); software peers do not (currently) have
-        // unrolled loops, so this step should locate those...
-        EffectPeer peer = peerCache.get(name);
+    public final synchronized EffectPeer getPeerInstance(FilterContext fctx, String name, int unrollCount) {
+        // The lookup key is only used to look up mappings, it is never inserted into the map.
+        peerCacheKey.name = name;
+        peerCacheKey.unrollCount = unrollCount;
+        EffectPeer<?> peer = peerCache.get(peerCacheKey);
         if (peer != null) {
             return peer;
-        }
-        // failing that, if there is a positive unrollCount, we attempt
-        // to find a previously cached hardware peer for that unrollCount
-        if (unrollCount > 0) {
-            peer = peerCache.get(name + "_" + unrollCount);
-            if (peer != null) {
-                return peer;
-            }
         }
 
         peer = createPeer(fctx, name, unrollCount);
         if (peer == null) {
-            throw new RuntimeException("Could not create peer  " + name +
-                                       " for renderer " + this);
+            throw new RuntimeException("Could not create peer  " + name + " for renderer " + this);
         }
-        // use the peer's unique name as the hashmap key
-        peerCache.put(peer.getUniqueName(), peer);
 
+        peerCache.put(new PeerCacheKey(name, unrollCount), peer);
         return peer;
     }
-
 
     /**
      * Returns this renderer's current state.
--- a/modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
@@ -26,7 +26,6 @@
 package com.sun.scenario.effect.impl;
 
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
@@ -113,7 +112,7 @@ public abstract class Renderer {
         }
     }
 
-    private final Map<PeerCacheKey, EffectPeer> peerCache = Collections.synchronizedMap(new HashMap<>(5));
+    private final Map<PeerCacheKey, EffectPeer<?>> peerCache = HashMap.newHashMap(5);
     private final PeerCacheKey peerCacheKey = new PeerCacheKey(null, -1); // reusable key to look up mappings
     private final ImagePool imagePool;
 
@@ -258,25 +257,27 @@ public abstract class Renderer {
      *
      * @param fctx filter context - same as this renderer's context
      * @param name not-unrolled name of the peer
-     * @param unrollCount
+     * @param unrollCount the unroll count of the peer
      * @return cached peer for this name and unroll count
      */
-    public final synchronized EffectPeer getPeerInstance(FilterContext fctx, String name, int unrollCount) {
-        // The lookup key is only used to look up mappings, it is never inserted into the map.
-        peerCacheKey.name = name;
-        peerCacheKey.unrollCount = unrollCount;
-        EffectPeer<?> peer = peerCache.get(peerCacheKey);
-        if (peer != null) {
+    public final EffectPeer getPeerInstance(FilterContext fctx, String name, int unrollCount) {
+        synchronized (peerCache) {
+            // The lookup key is only used to look up mappings, it is never inserted into the map.
+            peerCacheKey.name = name;
+            peerCacheKey.unrollCount = unrollCount;
+            EffectPeer<?> peer = peerCache.get(peerCacheKey);
+            if (peer != null) {
+                return peer;
+            }
+
+            peer = createPeer(fctx, name, unrollCount);
+            if (peer == null) {
+                throw new RuntimeException("Could not create peer  " + name + " for renderer " + this);
+            }
+
+            peerCache.put(new PeerCacheKey(name, unrollCount), peer);
             return peer;
         }
-
-        peer = createPeer(fctx, name, unrollCount);
-        if (peer == null) {
-            throw new RuntimeException("Could not create peer  " + name + " for renderer " + this);
-        }
-
-        peerCache.put(new PeerCacheKey(name, unrollCount), peer);
-        return peer;
     }
 
     /**
@@ -299,10 +300,19 @@ public abstract class Renderer {
                                              String name, int unrollCount);
 
     /**
-     * Returns current cache of peers.
+     * Clears and disposes all effect peers of this renderer.
      */
-    protected Collection<EffectPeer> getPeers() {
-        return peerCache.values();
+    protected final void clearPeers() {
+        EffectPeer<?>[] peers;
+
+        synchronized (peerCache) {
+            peers = peerCache.values().toArray(EffectPeer[]::new);
+            peerCache.clear();
+        }
+
+        for (EffectPeer<?> peer : peers) {
+            peer.dispose();
+        }
     }
 
     /**
--- a/modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
@@ -96,7 +96,7 @@ public abstract class Renderer {
 
         @Override
         public boolean equals(Object o) {
-            return o instanceof PeerCacheKey other
+            return o instanceof PeerCacheKey other // 'o' is never 'this'
                 && unrollCount == other.unrollCount
                 && name.equals(other.name);
         }
--- a/modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/prism/ps/PPSRenderer.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/prism/ps/PPSRenderer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2009, 2025, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2009, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -169,15 +169,11 @@ public class PPSRenderer extends PrRenderer {
      * May be called multiple times.
      */
     protected void dispose() {
-        // even if new peers are added from another thread while we're executing
-        // this on the rendering thread, they won't have any native resources
-        // since we're on the rendering thread, so no need to synchronize
-        for (EffectPeer peer : getPeers()) {
-            peer.dispose();
-        }
         synchronized (this) {
             state = DISPOSED;
         }
+
+        clearPeers();
         rf.removeFactoryListener(listener);
         rf = null;
         screen = null;
--- a/modules/javafx.graphics/src/test/addExports
+++ b/modules/javafx.graphics/src/test/addExports
@@ -59,6 +59,7 @@
 --add-exports javafx.graphics/com.sun.scenario=ALL-UNNAMED
 --add-opens javafx.graphics/com.sun.javafx.css.parser=ALL-UNNAMED
 --add-opens javafx.graphics/com.sun.scenario.animation=ALL-UNNAMED
+--add-opens javafx.graphics/com.sun.scenario.effect.impl=ALL-UNNAMED
 --add-opens javafx.graphics/javafx.css=ALL-UNNAMED
 --add-opens javafx.graphics/javafx.scene=ALL-UNNAMED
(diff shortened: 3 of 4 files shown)


modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
```diff
@@ -84,10 +84,37 @@ public static enum RendererState {
     }
 
     public static final String rootPkg = "com.sun.scenario.effect";
-    private static final Map<FilterContext, Renderer> rendererMap =
-        new HashMap<>(1);
-    private Map<String, EffectPeer> peerCache =
-        Collections.synchronizedMap(new HashMap<String, EffectPeer>(5));
+    private static final Map<FilterContext, Renderer> rendererMap = new HashMap<>(1);
```
nlisker: If this is meant to initialize the map to a size of 1 then `HashMap.newHashMap` should be used.

modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
```diff
@@ -84,10 +84,37 @@ public static enum RendererState {
...
+            return 31 * name.hashCode() + unrollCount;
+        }
+
+        @Override
+        public String toString() {
+            return name + unrollCount;
+        }
+    }
+
+    private final Map<PeerCacheKey, EffectPeer> peerCache = Collections.synchronizedMap(new HashMap<>(5));
```
nlisker: Isn't a `ConcurrentHashMap` a better alternative?

mstr2: No, I don't think so, and I think `Collections.synchronizedMap` is also pointless here. If this is supposed to allow concurrent access to the `peerCache` map, the original implementation is defective: `Collections.synchronizedMap` clearly states that iteration must be synchronized on the map instance, which this code fails to do.

I've changed the implementation to use a normal `HashMap`, and synchronize manually where it is accessed or modified. For this purpose, I've removed the `Renderer.getPeers()` method, which is only called in `PPSRenderer.dispose()` to iterate over the peers and dispose each one of them. Instead, the operation is now encapsulated in `Renderer.clearPeers()`.

modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
```diff
@@ -84,10 +83,37 @@ public static enum RendererState {
...
+        int unrollCount;
+
+        PeerCacheKey(String name, int unrollCount) {
+            this.name = name;
+            this.unrollCount = unrollCount;
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            return o instanceof PeerCacheKey other
```
andy-goryachev-oracle: typically
```
if(o == this) {
  return true;
}
```

modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
```diff
@@ -231,40 +257,29 @@ public void updateFloatTexture(LockableResource texture, FloatMap map) {
...
-            return peer;
-        }
-        // failing that, if there is a positive unrollCount, we attempt
-        // to find a previously cached hardware peer for that unrollCount
-        if (unrollCount > 0) {
-            peer = peerCache.get(name + "_" + unrollCount);
+    public final EffectPeer getPeerInstance(FilterContext fctx, String name, int unrollCount) {
+        synchronized (peerCache) {
+            // The lookup key is only used to look up mappings, it is never inserted into the map.
+            peerCacheKey.name = name;
```
andy-goryachev-oracle, line 266: this code is not re-entrant.  can you guarantee it's safe?

modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
```diff
@@ -285,10 +300,19 @@ protected abstract EffectPeer createPeer(FilterContext fctx,
...
     /**
-     * Returns current cache of peers.
+     * Clears and disposes all effect peers of this renderer.
      */
-    protected Collection<EffectPeer> getPeers() {
-        return peerCache.values();
+    protected final void clearPeers() {
+        EffectPeer<?>[] peers;
+
+        synchronized (peerCache) {
```
andy-goryachev-oracle, line 308: Having two monitors involved (peerCache here and peer's one elsewhere) might lead to a deadlock.  Just by looking at the code, I see one potential with createPeer() in Renderer:273 .  Could you take a look please?

andy-goryachev-oracle (commented):
I see several structural changes here (synchronization, changing the order of events around DISPOSED) that go beyond a simple optimization.

I am all for replacing string concatenation with a compound key like object, my main question though is how big of the problem is this?  Do we really incur significant overhead creating those keys?

modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
mstr2: Yes, but since the lookup key is never inserted into the map, `o` is never `this`. That would be an optimization that'll never be hit.

modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
mstr2: To be fair, I don't know why the code here is synchronized at all. I only ever see this method be called on the renderer thread (as it should be). However, I don't think that I've changed anything substantial here: the method was synchronized before, and is synchronized now.

modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
mstr2: I'm not calling any dangerous method while holding a lock here. The code first copies the values of the map into a local array, then releases the lock, and only then calls `EffectPeer.dispose()`.

modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
andy-goryachev-oracle: makes sense.  maybe add a comment then?

mstr2:
> I am all for replacing string concatenation with a compound key like object, my main question though is how big of the problem is this? Do we really incur significant overhead creating those keys?

As always with these kinds of optimizations: it depends on the application. Generally, you shouldn't expect this particular issue to be noticeable. But when you have 1000 of these "not generally noticeable" small issues, quantity can become a quality of its own.

modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
andy-goryachev-oracle: Hm, I meant something else: you are changing a peerCacheKey instance, are you sure it will never re-enter the same method (and clobber the same instance?)

You've also added synchronization on a different object here, which might create a deadlock.  And you do bring another good point - if this code is always called from the same thread, why synchronize at all?

modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
andy-goryachev-oracle: Right, my original comment should have been placed at L237, where inside a block synchronized on `peerCache`, you are calling `createPeer()`.  It looks like the latter is not synchronized on anything (can't see past reflection though), but if inside of `createPeer()` there is another code synchronized around a different object, there `might be` deadlock.

It looks like there isn't, but my question remains: why do you want to introduce these non-trivial changes in the first place, how big of a problem is this allocation is?

modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
mstr2: The reusable `peerCacheKey` instance is only used in these three lines:
```java
peerCacheKey.name = name;
peerCacheKey.unrollCount = unrollCount;
EffectPeer<?> peer = peerCache.get(peerCacheKey);
```

Since the `Map.get()` method won't call out to other code, I don't think there is any chance that something can go wrong in this particular segment of code.

> You've also added synchronization on a different object here, which might create a deadlock.

I don't see any potential problem that wasn't there before. Previously, this method synchronized on `this`, now it synchronizes on `peerCache`. That's not a material difference in itself, considering that there's only one other place where we synchronize on `peerCache`, and I've made sure that it won't call out to other methods.

modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
mstr2: I want to make these changes to improve the code. I don't really see the problem here...?

modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
```diff
@@ -84,10 +83,37 @@ public static enum RendererState {
...
     public static final String rootPkg = "com.sun.scenario.effect";
-    private static final Map<FilterContext, Renderer> rendererMap =
-        new HashMap<>(1);
-    private Map<String, EffectPeer> peerCache =
-        Collections.synchronizedMap(new HashMap<String, EffectPeer>(5));
+    private static final Map<FilterContext, Renderer> rendererMap = HashMap.newHashMap(1);
+
+    private static final class PeerCacheKey {
+        String name;
+        int unrollCount;
```
hjohn, line 90: Could this simply be `record PeerCacheKey(String name, int unrollCount) {}` ?

modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
Maran23: I'm also confused why this is synchronized. Looking at the subclasses, I can find patterns like that:
```
synchronized (this) {
    state = NOTREADY;
}
```

and comments like `// Must be called on the renderer thread`.

Nothing else is synchronized or uses synchronized collections. So if most of the code in the `Renderer`s is called with multiple threads, there will be many races and incoinsistent data. So might be worth to check in a follow-up.

modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
```diff
@@ -285,10 +300,19 @@ protected abstract EffectPeer createPeer(FilterContext fctx,
...
+     * Clears and disposes all effect peers of this renderer.
      */
-    protected Collection<EffectPeer> getPeers() {
-        return peerCache.values();
+    protected final void clearPeers() {
+        EffectPeer<?>[] peers;
+
+        synchronized (peerCache) {
+            peers = peerCache.values().toArray(EffectPeer[]::new);
+            peerCache.clear();
```
Maran23, line 310: So the cache was not cleared before, right? Is there any side effect / negative effect we need to be aware of?

Also, can't we just iteratoe over `peerCache.values()` and call dispose on every peer, and then clear the `Map`? So we can save the temp `Array`.

Maran23:
Hm it looks like my review did not refresh the inactive counter. Which might be expected, but still surprising.

modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
mstr2: The `EffectPeer` instances were disposed, but not removed from the map. I think synchronization is probably completely unnecessary here, because we only ever have a single renderer thread. I've still retained the synchronization in a safe way to ensure no regressions. The purpose of this enhancement is not to reevaluate whether we need synchronization at all, it's just to reduce object allocations on the rendering path.

Maran23: > The EffectPeer instances were disposed, but not removed from the map

Yes, that is what I mean. But is there any side effect we may have now that we remove those entries from the map?

If I understand you right, synchronization is the reason you chose to not do?:

```
 for (EffectPeer<?> peer : peerCache.values()) {
    peer.dispose();
}
peerCache.clear();
```

Because this would save us the temp array and is a bit easier too.

mstr2: > > The EffectPeer instances were disposed, but not removed from the map
> 
> Yes, that is what I mean. But is there any side effect we may have now that we remove those entries from the map?

No, I don't think so.

> If I understand you right, synchronization is the reason you chose to not do?:
> 
> ```
>  for (EffectPeer<?> peer : peerCache.values()) {
>     peer.dispose();
> }
> peerCache.clear();
> ```
> 
> Because this would save us the temp array and is a bit easier too.

The temporary array is nothing to worry about, this doesn't happen repeatedly on a hot path. The reason I'm doing it like this is to be able to call the dispose method without holding a lock, so as to not introduce any risk of deadlocks. Removing synchronization entirely seems reasonable, but is out of scope for this PR.

Maran23 (approved):
LGTM.

Might makes sense to reevaluate the synchronization at one point, but that involves more code.

hjohn (approved):
LGTM

modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/Renderer.java
Maran23: Initially I had the same thought, but the variables need to be modified (set), so not possible with a record unfortunately.

mstr2: The key needs to be mutable, using a records defeats the point of eliding object allocations.

andy-goryachev-oracle:
It feels like my concerns about synchronization were not properly addressed.  

Also, the use of mutable key (which might be ok in this particular case, hope there is no chance for re-entry), but this design is - let's say - unorthodox, might have created a booby trap.

mstr2:
> It feels like my concerns about synchronization were not properly addressed.

I did step through the effect peer creation path, and didn't see anything suspicious. However, I want to point out that this code was always called while holding a lock: in the old version, it was using the `this` monitor, in the new version it's using a more narrowly-scoped monitor that isn't accessible to outside code.

In any case, I'll create a [ticket](https://bugs.openjdk.org/browse/JDK-8384870) to look at removing the lock entirely, as I think it's probably not needed because rendering is single-threaded.

> Also, the use of mutable key (which might be ok in this particular case, hope there is no chance for re-entry), but this design is - let's say - unorthodox, might have created a booby trap.

Yes, it's not something you'd usually do, but in this specific case it saves us from lots of allocations on a hot path. Performance optimization is always a trade-off, but I think the risk of misuse is quite small here (it's localized and not reusable outside of the class).

andy-goryachev-oracle:
> In any case, I'll create a [ticket](https://bugs.openjdk.org/browse/JDK-8384870) to look at removing the lock entirely

Thanks!

> it's localized and not reusable outside of the class

Can you guarantee that this code will never be re-entered?  If so, maybe add a comment pointing to this requirement?  edit: too late for this Pr obviously, maybe in JDK-8384870 then?

(re: reentrancy, see https://github.com/openjdk/jfx/pull/1691 )

mstr2:
> Can you guarantee that this code will never be re-entered? If so, maybe add a comment pointing to this requirement? edit: too late for this Pr obviously, maybe in JDK-8384870 then?
> 
> (re: reentrancy, see #1691 )

I'm quite confident that this didn't structurally introduce a _new_ race condition. If there is one, it was already there before.

andy-goryachev-oracle:
> race condition.

not the race condition, re-entrance - when `getPeerInstance()` is called from within another call to `getPeerInstance()`.

For example, `createPeer()` is abstract, so a subclass can implement `createPeer()` which in turn calls `getPeerInstance()` for some reason.

mstr2:
> not the race condition, re-entrance - when `getPeerInstance()` is called from within another call to `getPeerInstance()`.
> 
> For example, `createPeer()` is abstract, so a subclass can implement `createPeer()` which in turn calls `getPeerInstance()` for some reason.

The same reentrant call could also have happened with the old implementation, as I didn't change which method is called from where. I don't see anything that has materially changed: neither the call stack, nor a potential race condition and an associated deadlock.

andy-goryachev-oracle:
> The same reentrant call could also have happened with the old implementation

The old code used `String` for a key, which is perfectly re-entrant.

But, after a closer look, your peerCacheKey is only used to lookup once, more importantly, no other code can intervene between the key is mutated and the cached value is returned.  _Unless someone changes it later_, this should be fine.

I am a bit concerned though that you integrated this PR without reaching agreement on all the comments first.

mstr2:
> I am a bit concerned though that you integrated this PR without reaching agreement on all the comments first.

Your technical questions were about synchronization and a potentially reentrant call stack. My explanation was that there are no material changes to the call stack and no material changes with regards to synchronization. You also asked why we should do this in the first place, which I answered by pointing out that lots of small inefficiencies on a hot path will add up.

After that, you didn't respond for two and a half months, so I considered these questions settled. If you didn't want this PR to move forward, I think you should have said so (or at least said clearly that you didn't feel your questions were answered satisfactorily). I usually consider review participation to be non-blocking if there is no explicit approval or disapproval, and all questions have been answered.

andy-goryachev-oracle:
> and all questions have been answered.

they weren't, I was waiting for your response (not the first time, I might add).

But you do bring a good point - we should indicate more clearly that the discussion is not over.  My understanding is that all who substantially participated in the review (or raised substantial concern) need to mark a PR as Approved, or explicitly say that a change is requested.  Also, if there is no response for some time, we usually add a comment.

`<rant>`
Gitlab has explicit [Resolve Thread] button, but in github this button seem to work backwards: the originator of the comment _cannot_ resolve the discussion.  I've started to use emoji to indicate the acquiescence, but it is a low key indicator.  Collapsing a discussion is also not very user-friendly, since often other participants have input too.
Oh well, time to go through the open PRs and check for unresolved discussions.
`</rant>`

Maran23:
> they weren't, I was waiting for your response (not the first time, I might add).

I'd like to weigh in here: I had the same impression at: https://github.com/openjdk/jfx/pull/1975#discussion_r2541102625

My questions weren't fully answered there either, and I didn't get the sense that you were willing to consider my arguments or even try them out. Especially regarding reconsidering the decision about `managed` or `needsLayoutProperty`. I came away with the impression that you probably wanted to take the 'easy' route without further refactoring or trying out.
And ultimately, this was merged without taking my arguments into account. And I IMO `raised substantial concern`/participated in the review.

Therefore, I would like to emphasize that this should apply to everyone.

andy-goryachev-oracle:
> I would like to emphasize that this should apply to everyone.

I agree!  In that particular case I did respond to your concerns, didn't I?  And the response was - this was _designed to work that way_, and I even tried to explain why.  Sometimes there is more than one way to peel the fruit, right?

crschnick:
Is this full agreement idea formalized somewhere? I was under the impression that if you get the needed amount of approvals, the PR can be merged unless the project lead blocks this. Getting full agreement on everything when more than the required number of reviewers are involved is in some cases not feasible imo.

kevinrushforth:
The guidelines in question are in the [Before you integrate or sponsor a PR](https://github.com/openjdk/jfx/blob/master/README-code-reviews.md#before-you-integrate-or-sponsor-a-pr) section of the code review policies.

Quoting a relevant piece:

> Before you integrate or sponsor the PR, ensure that the following have been done:
> 
> * All substantive feedback has been addressed, especially any objections from one with a Reviewer role
>     * If you have pushed any changes in response to a Reviewer's substantive comments, wait for them to re-review the latest version of your PR with those changes (to ensure they are satisfied with the way you addressed them)

As always, it's a judgment call. I reread the comments, and I don't think it was sufficiently clear that Andy's question was still outstanding, so I can easily see why Michael integrated this with the two approving reviews.

Perhaps the new issue Michael filed could be expanded to include evaluating any potential reentrancy issues (preexisting or otherwise)?

mstr2:
> > and all questions have been answered.
> 
> they weren't, I was waiting for your response (not the first time, I might add).

I still don't know which question I allegedly didn't answer.

> But you do bring a good point - we should indicate more clearly that the discussion is not over. My understanding is that all who substantially participated in the review (or raised substantial concern) need to mark a PR as Approved, or explicitly say that a change is requested. Also, if there is no response for some time, we usually add a comment.

I don't think that such a policy would make life easier in practice. For example, I sometimes comment on an isolated aspect of a PR, but without the intention of reviewing all of it. Having a rule that basically blocks integration of a PR because someone substantially participated in a discussion without the intention of explicitly approving or disapproving makes a slow process even slower. It's probably enough to just clearly state when you think the discussion is not finished.

PR #2092, opened 2026-03-02
8341852: Fix potential threading issue in WebView's RTImage

lukostyra:
This change fixes a potential threading issue in RTImage.draw() method in WebView. The issue has been theorized back in [JDK-8333374](https://bugs.openjdk.org/browse/JDK-8333374), I have not managed to figure out a reliable reproducer for it, but the circumstances for it to happen are there.

Issue can happen in `RTimage.draw()` method. In 99% of cases `RTImage.draw()` will be called only by QuantumRenderer thread (that is during regular drawing to a Stage/Scene), however there is also a possibility we will call `RTImage.draw()` from a different thread, notably with a `PrinterGraphics` object (or in other words, when printing). Then there is a chance we will run some of the initial checks in parallel to QuantumRenderer thread. I found two places that can be affected - the initial checks at the beginning of the method and the `getTexture()` call inside the code block responsible for printing.

The first place was fixed by running the checks on the render thread. That way we can ensure the render thread will not overwrite the `tex` reference from the render thread while the printing thread reads it. To make this happen I added `PrismInvoker.callOnRenderThread()` which functions like `PrismInvoker.runOnRenderThread()` but with a `Callable<>` instead of a `Runnable` object.

Second place can happen if, for some reason, in between previous checks and the `getTexture()` call QuantumRenderer will modify or free RTImage's tex reference (ex. while pruning the Vram pool) - then the `getTexture()` call will return `null` (introduced as NPE prevention from [JDK-8333374](https://bugs.openjdk.org/browse/JDK-8333374)) which can cause an NPE on QuantumRenderer thread. If that happens, we skip the `readPixels()` call which will print an empty RTImage.

I verified these changes on Windows and Linux with our test suite and with manual testing via `HelloWebView` and using an old `SimplePrint.java` reproducer from [JDK-8118415](https://bugs.openjdk.org/browse/JDK-8118415); both regular drawing and printing work as they did. I found simply loading `google.com` via both apps works to trigger this draw call multiple times (`SimplePrint.java` needed to be adjusted slightly to print out the web page instead of predefined HTML code). Test suite completes successfully (there were 3 web.test failures on my end but they also happened on master and refer to other, unrelated parts of javafx.web) and printing works the same way it used to.

---------

---------
### Progress

### Issue
 * [JDK-8341852](https://bugs.openjdk.org/browse/JDK-8341852): Fix potential threading issue in WebView's RTImage (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2092/head:pull/2092` \
`$ git checkout pull/2092`

`$ git checkout pull/2092` \
`$ git pull https://git.openjdk.org/jfx.git pull/2092/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2092`

`$ git pr show -t 2092`

</details>

<a href="https://git.openjdk.org/jfx/pull/2092.diff">https://git.openjdk.org/jfx/pull/2092.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2092#issuecomment-3984797499)
</details>

### DIFF
--- a/modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/PrismInvoker.java
+++ b/modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/PrismInvoker.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -25,6 +25,7 @@
 
 package com.sun.javafx.webkit.prism;
 
+import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.FutureTask;
 import java.util.concurrent.locks.ReentrantLock;
@@ -78,21 +79,46 @@ public final class PrismInvoker extends Invoker {
         PlatformImpl.runLater(r);
     }
 
+    private static boolean isOnRenderThread() {
+        return Thread.currentThread().getName().startsWith("QuantumRenderer");
+    }
+
     static void invokeOnRenderThread(final Runnable r) {
         Toolkit.getToolkit().addRenderJob(new RenderJob(r));
     }
 
     static void runOnRenderThread(final Runnable r) {
-        if (Thread.currentThread().getName().startsWith("QuantumRenderer")) {
+        if (isOnRenderThread()) {
             r.run();
         } else {
-            FutureTask<Void> f = new FutureTask<>(r, null);
+            FutureTask<Void> f = new FutureTask<Void>(r, null);
             Toolkit.getToolkit().addRenderJob(new RenderJob(f));
             try {
                 // block until job is complete
                 f.get();
             } catch (ExecutionException | InterruptedException ex) {
                 log.severe("RenderJob error", ex);
+            }
+        }
+    }
+
+    static <T> T callOnRenderThread(final Callable<T> c) {
+        if (isOnRenderThread()) {
+            try {
+                return c.call();
+            } catch (Exception ex) {
+                log.severe("RenderJob error", ex);
+                return null;
+            }
+        } else {
+            FutureTask<T> f = new FutureTask<T>(c);
+            Toolkit.getToolkit().addRenderJob(new RenderJob(f));
+            try {
+                // block until job is complete
+                return f.get();
+            } catch (ExecutionException | InterruptedException ex) {
+                log.severe("RenderJob error", ex);
+                return null;
             }
         }
     }
--- a/modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/PrismInvoker.java
+++ b/modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/PrismInvoker.java
@@ -91,7 +91,7 @@ public final class PrismInvoker extends Invoker {
         if (isOnRenderThread()) {
             r.run();
         } else {
-            FutureTask<Void> f = new FutureTask<Void>(r, null);
+            FutureTask<Void> f = new FutureTask<>(r, null);
             Toolkit.getToolkit().addRenderJob(new RenderJob(f));
             try {
                 // block until job is complete
@@ -111,7 +111,7 @@ public final class PrismInvoker extends Invoker {
                 return null;
             }
         } else {
-            FutureTask<T> f = new FutureTask<T>(c);
+            FutureTask<T> f = new FutureTask<>(c);
             Toolkit.getToolkit().addRenderJob(new RenderJob(f));
             try {
                 // block until job is complete
--- a/modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/RTImage.java
+++ b/modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/RTImage.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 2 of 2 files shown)


lukostyra:
Re: unsuccessful GHA check on Linux: this seems to be an intermittent failure in one test in graphics. We already have this on the backlog as [JDK-8357459](https://bugs.openjdk.org/browse/JDK-8357459). I re-ran the GHA Linux job - https://github.com/lukostyra/jfx/actions/runs/22580397605 - and it passed.

Not sure if GitHub will eventually realize this re-run happened, but leaving out this note for reviewers.

/reviewers 2

modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/PrismInvoker.java
```diff
@@ -78,15 +79,19 @@ private static boolean isEventThreadPrivate() {
         PlatformImpl.runLater(r);
     }
 
+    private static boolean isOnRenderThread() {
+        return Thread.currentThread().getName().startsWith("QuantumRenderer");
```
andy-goryachev-oracle, line 83: There are two more files with a similar check (`Disposer`, `NGCanvas`) - do you think  there should be a utility method for it?

Also, is there a better, faster way other than `startsWith()`?

lukostyra: Yes I think there probably should be, and there also probably is a better way. I'm also not a big fan of just checking thread's name, especially if we need to change it for any reason. When working on this bug I stumbled upon an ancient enhancement about this - [JDK-8090484](https://bugs.openjdk.org/browse/JDK-8090484) - so it seems there were at least some theorized attempts to improve this part.

I would consider it out of scope of this change though and improve this separately (probably as a part of the above issue), it almost certainly would be a larger improvement reaching over to graphics module.

**EDIT:** Also, regarding `Disposer` - I suspect you meant the one residing in graphics under `com/sun/prism/impl/Disposer.java`. We seem to have quite a bit of Disposers in the code base...

andy-goryachev-oracle: definitely out of scope for this PR.

since it's a real (non-virtual) thread, we could create a thread of specific type and just check `if(Thread.currentThread().getClass == QuantumRenderer.class)` ?

modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/PrismInvoker.java
```diff
@@ -78,15 +79,19 @@ private static boolean isEventThreadPrivate() {
...
         Toolkit.getToolkit().addRenderJob(new RenderJob(r));
     }
 
     static void runOnRenderThread(final Runnable r) {
-        if (Thread.currentThread().getName().startsWith("QuantumRenderer")) {
+        if (isOnRenderThread()) {
             r.run();
         } else {
-            FutureTask<Void> f = new FutureTask<>(r, null);
+            FutureTask<Void> f = new FutureTask<Void>(r, null);
```
andy-goryachev-oracle: I think this change is unnecessary:
`FutureTask<Void> f = new FutureTask<>(r, null);`

modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/PrismInvoker.java
```diff
@@ -96,4 +101,25 @@ static void runOnRenderThread(final Runnable r) {
...
+    static <T> T callOnRenderThread(final Callable<T> c) {
+        if (isOnRenderThread()) {
+            try {
+                return c.call();
+            } catch (Exception ex) {
+                log.severe("RenderJob error", ex);
+                return null;
+            }
+        } else {
+            FutureTask<T> f = new FutureTask<T>(c);
```
andy-goryachev-oracle: same here
`FutureTask<T> f = new FutureTask<>(c);`

andy-goryachev-oracle (commented):
Looks good from the code review perspective, only minor suggestions.

I'll do more testing tomorrow.

andy-goryachev-oracle:
I am getting assertion messages while printing:

```
PrinterJob: JavaFX PrinterJob Printer Print to PDF
 Collation = UNCOLLATED
 Copies = 1
 Sides = ONE_SIDED
 JobName = JavaFX Print Job
 Output file = 
 Page ranges = null
 Print color = COLOR
 Print quality = NORMAL
 Print resolution = Feed res=300dpi. Cross Feed res=300dpi.
 Paper source = Paper source : Automatic
 Page layout = Paper=Paper: A4 size=210.0x297.0 MM Orient=PORTRAIT leftMargin=54.0 rightMargin=54.0 topMargin=54.0 bottomMargin=54.0
Job Status = NOT_STARTED

*** unexpected PG access
java.lang.Exception: Stack trace
	at java.base/java.lang.Thread.dumpStack(Thread.java:2018)
	at javafx.graphics/javafx.scene.Node.getPeer(Node.java:2700)
	at javafx.graphics/javafx.scene.Node$1.getPeer(Node.java:539)
	at javafx.graphics/com.sun.javafx.scene.NodeHelper.getPeer(NodeHelper.java:213)
	at javafx.graphics/com.sun.prism.j2d.print.J2DPrinterJob$J2DPageable.printNode(J2DPrinterJob.java:1145)
	at javafx.graphics/com.sun.prism.j2d.print.J2DPrinterJob$J2DPageable.print(J2DPrinterJob.java:1138)
	at java.desktop/sun.lwawt.macosx.CPrinterJob$4.run(CPrinterJob.java:878)
	at java.desktop/sun.lwawt.macosx.CPrinterJob.printAndGetPageFormatArea(CPrinterJob.java:897)
	at java.desktop/sun.lwawt.macosx.CPrinterJob.printLoop(Native Method)
	at java.desktop/sun.lwawt.macosx.CPrinterJob.print(CPrinterJob.java:377)
	at javafx.graphics/com.sun.prism.j2d.print.J2DPrinterJob$PrintJobRunnable.run(J2DPrinterJob.java:863)
	at java.base/java.lang.Thread.run(Thread.java:1474)
```

the print job succeeds though.

This looks like https://bugs.openjdk.org/browse/JDK-8089053 exactly.

andy-goryachev-oracle (approved):
I was able to print simple HTML, simple HTML with an embedded image (data url) and google.com.
The assertions messages are pre-existing.

macOS 26.3 with installed virtual PDF printer from https://apps.apple.com/us/app/print-to-pdf-printer-app/id1639234272?mt=12

For ease of reviewing, I've updated the monkey tester to include File -> Print menu and `[P]rint` button on the WebView page.

modules/javafx.web/src/main/java/com/sun/javafx/webkit/prism/PrismInvoker.java
kevinrushforth: When fixing the follow-up, I'd rather see a `QuantumRenderer::isRenderThread` method added to encapsulate the details.

kevinrushforth (commented):

I left a (minor) comment on the follow-up issue.

lukostyra:
Bump, this still needs a second reviewer. @kevinrushforth or @arapte if you have the time?

kevinrushforth (approved):
Fix looks good. I did a quick sanity check and it works as expected.

PR #2093, opened 2026-03-02
8378970: Don't use exceptions for flow control in CssParser.colorValueOfString()

mstr2:
`CssParser.colorValueOfString()` parses a string into a `Color`. If this fails by throwing an exception, `null` is returned from `colorValueOfString()`, signalling to the caller that the color string might be a lookup instead.

Since color lookups can appear in many places in a CSS file, it is preferable to use return values for flow control instead of exceptions. For this purpose, we need non-throwing versions of methods that parse numbers and colors.

These are the changes in this PR:
1. Move color parsing from the `Color` class to `CssColorParser`. This is done so that we can access the new `tryParseColor(String)` method from `CssParser`, and also to separate concerns between color representation and parsing.
2. Add a non-throwing `CssNumberParser.tryParseDouble()`, which returns `NaN` to indicate a parsing failure. We can use `NaN` because it is not a valid return value of the parser.
3. Since color parsing also uses `Integer.parseInt()`, we need a non-throwing version of this method: `CssNumberParser.tryParseInt()`. Since we can't use any particular `int` as a return value that indicates failure, I've decided to return a `long` value instead, where a value less than `Integer.MIN_VALUE` indicates a parsing failure.

/reviewers 2

---------
### Progress

### Issue
 * [JDK-8378970](https://bugs.openjdk.org/browse/JDK-8378970): Don't use exceptions for flow control in CssParser.colorValueOfString() (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2093/head:pull/2093` \
`$ git checkout pull/2093`

`$ git checkout pull/2093` \
`$ git pull https://git.openjdk.org/jfx.git pull/2093/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2093`

`$ git pr show -t 2093`

</details>

<a href="https://git.openjdk.org/jfx/pull/2093.diff">https://git.openjdk.org/jfx/pull/2093.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2093#issuecomment-3985352289)
</details>


modules/javafx.graphics/src/main/java/com/sun/javafx/css/parser/CssNumberParser.java
```diff
@@ -35,6 +35,44 @@ private CssNumberParser() {}
     // Saturation: beyond this, doubles overflow/underflow anyway
     private static final int EXP_LIMIT = 10000;
 
+    /**
+     * Parses a number as an integer. If successful, returns the integer as a long value; otherwise
+     * returns a long value less than {@link Integer#MIN_VALUE} to indicate failure.
+     */
+    public static long tryParseInt(String s, int start, int end, int radix) {
```
mstr2, line 42: This code is taken from `Integer.parseInt(String, int, int, int)` in the JDK, and modified to return `Long.MIN_VALUE` instead of throwing `NumberFormatException`.

modules/javafx.graphics/src/test/java/test/com/sun/javafx/css/parser/CssNumberParserTest.java
```diff
@@ -31,530 +31,593 @@
...
-    @Test
-    public void parseIntegerAndSignedInteger() {
-        assertSameDouble(0.0, "0");
-        assertSameDouble(255.0, "255");
-        assertSameDouble(2.0, "+2");
-        assertSameDouble(-2.0, "-2");
-        assertSameDouble(42.0, "00042");
-    }
+    @Nested
+    class ParseIntegerTest {
```
mstr2, line 42: These tests are taken from `Integer/ParsingTest` in the JDK.

mstr2 (commented):
Note to reviewers: many changes in this PR are just whitespace changes. You can hide those by going to the "Files changed" tab in GitHub, clicking on the gear icon, and selecting "Hide whitespace".

modules/javafx.graphics/src/main/java/com/sun/javafx/css/parser/CssNumberParser.java
Maran23: I wonder if we really should do that. Whenever something in the JDK changes, we may need to port it to our Integer parsing as well. 
Can't we just do the same as we do with `Double`:

```
try {
    return Double.parseDouble(s.substring(start, end));
} catch (NumberFormatException _) {
    return Double.NaN;
}
```
But with `Integer`?

mstr2: The entire premise of this PR is to not use exceptions for flow control. In addition to that, what could possibly change with a very short algorithm that parses an int?

andy-goryachev-oracle: I'd say this might be a good decision to use `long` because throwing an exception is more expensive in any case.  And, the int32 logic is unlikely to change ;-)

Maran23: > The entire premise of this PR is to not use exceptions for flow control. In addition to that, what could possibly change with a very short algorithm that parses an int?

I'm all in for improving the control flow and not throwing exceptions just because we used named colors (lookup).
And +1 for the resulting performance improvement.

Just wondering about the particular copy&paste. I also don't think this function will change often in the future. But maintaining the same function again always comes with a bit more maintenance cost still. Thats why I'm asking. 

I'm fine with whatever we all agree on though!

hjohn:
If it helps, an exception for flow control can be made a lot lighter by re-using a static exception, with its stack trace disabled. Make it checked as well to ensure it never escapes the parsing logic where the user might see an exception without a stack trace.

I'm not really in favor of using `long` to signal problems with an `int`; I'd use either `Integer` or `OptionalInt`.

hjohn:
Perhaps return an object right away instead of the int?  Is there any reason why we can't return a `Color` and get it ASAP to FX's domain objects?

mstr2:
> Perhaps return an object right away instead of the int? Is there any reason why we can't return a `Color` and get it ASAP to FX's domain objects?

Return a `Color` from where, instead of doing what?

hjohn:
> > Perhaps return an object right away instead of the int? Is there any reason why we can't return a `Color` and get it ASAP to FX's domain objects?
> 
> Return a `Color` from where, instead of doing what?

I made the wrong assumption, because you said you can't use any `int` value as "error", so I assumed you need the entire int range and that would mean 32-bits which can hold RGBA = Color.

However, looking at it, I don't see any real place where you need the full int range. It seems to be parsing two (hex) digits at most, or something that gets put in a double. One of those is degrees, which could be anything I suppose as they get modded. But why would that cut off at 2147483647 (or -2147483648) and not something else arbitrary?  I see no reason why we can't use a special int value here (as the supported range is already arbitrary), or just parse it as `double` right away.

mstr2:
I don't think that returning a `long` from a method named `tryParseInt()` is fundamentally flawed. In fact, I think it's better than the alternatives, and you can even combine it with pattern matching in the future:
```java
if (tryParseInt("123") instanceof int value) {
   // parsing successful
}

**Comment by andy-goryachev-oracle** [comment]
`instanceof int` wow, what does bytecode look like?  do you know if any boxing/unboxing is involved?

**Comment by hjohn** [comment]
> `instanceof int` wow, what does bytecode look like? do you know if any boxing/unboxing is involved?

`instanceof` is pretty cheap in general already; the bytecode for doing this with a primitive type is not going to do boxing but will likely be intrisified with a mask check or something.

**Comment by mstr2** [comment]
> `instanceof int` wow, what does bytecode look like? do you know if any boxing/unboxing is involved?

The `instanceof` bytecode is not used for primitive pattern matching. It's just a method call to `java.lang.runtime.ExactConversionsSupport.isLongToIntExact(long)`, and that method just consists of one line:
```java
return n == (long)(int)n;
```

andy-goryachev-oracle:
we can't use it right now:

```
 error: primitive patterns are a preview feature and are disabled by default.
		if(lv() instanceof int v) {
		     ^
  (use --enable-preview to enable primitive patterns)
```

andy-goryachev-oracle:
```
void f() {
		if(lv() instanceof int v) {
```

bytecode:

```
  void f();
    Code:
         0: aload_0
         1: invokevirtual #7                  // Method lv:()J
         4: lstore_2
         5: lload_2
         6: lstore        4
         8: lload         4
        10: invokestatic  #13                 // Method java/lang/runtime/ExactConversionsSupport.isLongToIntExact:(J)Z
        13: ifeq          26
        16: lload_2
        17: l2i
        18: istore_1
```

hjohn:
I'm guessing this is part of larger effort to perhaps not need binary CSS format :)

hjohn:
I'm wondering one more thing, why don't we skip color parsing completely as we can easily tell if something is a variable? I believe CSS requires that they start with `--<letter>`.

mstr2:
> I'm wondering one more thing, why don't we skip color parsing completely as we can easily tell if something is a variable? I believe CSS requires that they start with `--<letter>`.

In JavaFX the dashes are not required. The `var()` function is also neither required nor supported.

mstr2:
> I'm guessing this is part of larger effort to perhaps not need binary CSS format :)

This particular change is just what I thought was a low-hanging fruit. What I'd _like_ to see is the entire CSS parser be replaced with something that's actually good, but I'm not sure we're ever going to get there.

Maran23:
> > I'm wondering one more thing, why don't we skip color parsing completely as we can easily tell if something is a variable? I believe CSS requires that they start with `--<letter>`.
> 
> In JavaFX the dashes are not required. The `var()` function is also neither required nor supported.

I wondered the same. Maybe we could document a recommendation and do a simple `startsWith("--")` check, so we can save some parsing cycles.

hjohn (approved):
In light that vars are not easy to recognize JavaFX's variant of CSS, I guess it is acceptable to first try parse them as colors.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/parser/CssNumberParser.java
Maran23: Just had a look and indeed it did not really changed often in the JDK. What was changed more often is the part with the exceptions. So this is probably fine. Still would like to see more opinions.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/parser/CssNumberParser.java
```diff
@@ -35,6 +35,44 @@ private CssNumberParser() {}
     // Saturation: beyond this, doubles overflow/underflow anyway
     private static final int EXP_LIMIT = 10000;
 
+    /**
+     * Parses a number as an integer. If successful, returns the integer as a long value; otherwise
+     * returns a long value less than {@link Integer#MIN_VALUE} to indicate failure.
```
andy-goryachev-oracle, line 40: minor:
can it return anything other than `Long.MIN_VALUE` on failure?
should we explicitly mention `Long.MIN_VALUE` then?

modules/javafx.graphics/src/test/java/test/com/sun/javafx/css/parser/CssNumberParserTest.java
andy-goryachev-oracle: minor: maybe add a comment stating that?

andy-goryachev-oracle (approved):
Looks good!

(will re-approve if you decide to address the comments)

modules/javafx.graphics/src/main/java/com/sun/javafx/css/parser/CssNumberParser.java
mstr2: It doesn't return anything other than `MIN_VALUE` on failure, but keep in mind that this really is only a simple, internal method that's only used in a single place (it even says so in the name of the class). So for this reason, I'm going to take the win (thanks for the review!) and defer further polish to the next time we'll talk about this method, which is probably never.

Maran23:
I think this might also close: https://bugs.openjdk.org/browse/JDK-8201135

andy-goryachev-oracle:
> I think this might also close: https://bugs.openjdk.org/browse/JDK-8201135

you are right. I'll close that ticket as duplicate.

PR #2102, opened 2026-03-09
8372398: [macOS] Shown ContextMenu does not close when macOS system menu is clicked

jperedadnr:
This PR adds a fix to close the popup windows on macOS when clicking over the system menu bar. 

According to the macOS standard behaviour for native applications, when a popup window is showing, and the user clicks on the system menu bar, the click event is consumed and the popup is closed. A second click is then required to open the system menu bar.

This is done by the popup windows directly, as they are NSMenu objects that enable a modal event tracking loop, capturing all events including those from the system menu bar, in order to dismiss the popup when the click is outside the window.

However, JavaFX just implements regular NSWindows, and there is no such event loop. Therefore, this PR adds a notification to the system menu instead. When the menu is about to open, the popup window gets a notification, which is processed to cancel the menu animation, preventing it from showing up, and also closing the popup. And then, with the popup closed, a new click from the user will open the system menus as usual.

This applies to all JavaFX menus from the system menu bar, that is: the application menu (the "java" menu when the application is not packaged yet) and other menus created by the application, if any. 

It doesn't apply to the Apple system menu, which is not handle by JavaFX, though. This causes a difference with native applications, that treat such menu in the same way.

---------

---------
### Progress

### Issue
 * [JDK-8372398](https://bugs.openjdk.org/browse/JDK-8372398): [macOS] Shown ContextMenu does not close when macOS system menu is clicked (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2102/head:pull/2102` \
`$ git checkout pull/2102`

`$ git checkout pull/2102` \
`$ git pull https://git.openjdk.org/jfx.git pull/2102/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2102`

`$ git pr show -t 2102`

</details>

<a href="https://git.openjdk.org/jfx/pull/2102.diff">https://git.openjdk.org/jfx/pull/2102.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2102#issuecomment-4026308249)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassMenu.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassMenu.m
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -178,8 +178,13 @@ static jfieldID  jPixelsScaleYField = 0;
 
 // JDK-8096139: do not use menuNeedsUpdate here, even though Cocoa prohibits
 // changing the menu structure during menuWillOpen...
-- (void)menuWillOpen: (NSMenu *)menu
+- (void)menuWillOpen: (NSMenu *)_menu
 {
+    // Post notification before menu opens so popup windows can be closed
+    [[NSNotificationCenter defaultCenter] postNotificationName:GlassMenuBarWillOpenNotification
+                                                        object:_menu
+                                                      userInfo:nil];
+
     GET_MAIN_JENV;
     if (env != NULL)
     {
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassMenu.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassMenu.m
@@ -32,6 +32,7 @@
 #import "GlassMenu.h"
 #import "GlassHelper.h"
 #import "GlassKey.h"
+#import "GlassWindow+Java.h"
 
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
 
@@ -180,10 +181,16 @@ static jfieldID  jPixelsScaleYField = 0;
 // changing the menu structure during menuWillOpen...
 - (void)menuWillOpen: (NSMenu *)_menu
 {
-    // Post notification before menu opens so popup windows can be closed
-    [[NSNotificationCenter defaultCenter] postNotificationName:GlassMenuBarWillOpenNotification
-                                                        object:_menu
-                                                      userInfo:nil];
+    if ([GlassWindow _hasGrab]) {
+        // If an auto-hide popup window is showing, it has an active grab.
+        // In this case, we close the popup instead of opening the menu:
+        [_menu cancelTrackingWithoutAnimation];
+
+        // _resetGrab fires FOCUS_UNGRAB on the popup's owner window, which
+        // will trigger doAutoHide() -> hide() on the popup.
+        [GlassWindow _resetGrab];
+        return;
+    }
 
     GET_MAIN_JENV;
     if (env != NULL)
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassWindow+Overrides.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassWindow+Overrides.m
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2025, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -37,6 +37,7 @@
 #import "GlassViewDelegate.h"
 #import "GlassApplication.h"
 #import "GlassScreen.h"
+#import "GlassMenu.h"
 
 #import <AppKit/NSGraphics.h> // NSBeep();
 
@@ -123,6 +124,14 @@
 - (void)windowWillClose:(NSNotification *)notification
 {
     //NSLog(@"windowWillClose");
+
+    NSWindow *window = notification.object;
+    BOOL isPopup = ([window styleMask] & NSWindowStyleMaskNonactivatingPanel) != 0;
+    if (isPopup) {
+        // If this window is a popup, remove the menu observer
+        [[NSNotificationCenter defaultCenter] removeObserver:self name:GlassMenuBarWillOpenNotification object:nil];
+    }
+
     // Remove self from list of owner's child windows
     if (self->owner != nil) {
         [self->owner removeChildWindow:self];
@@ -306,6 +315,21 @@
 
     // it's up to app to decide if the window should be closed
     return FALSE;
+}
+
+- (void)menuWillOpenHandler:(NSNotification *)notification {
+    // This handler is only called for popup windows as the observer
+    // is only registered for popups in _createWindowCommonDo
+
+    // When there is a mouse click over the system menu bar, and a popup window is showing,
+    // we consume the event by immediately closing the popup without allowing the menu to appear.
+    NSMenu *menu = notification.object;
+    if (menu != nil) {
+        [menu cancelTrackingWithoutAnimation];
+    }
+
+    // Close the popup window
+    [self->nsWindow close];
 }
 
 #pragma mark --- Title Icon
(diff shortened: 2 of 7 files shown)


beldenfox:
I ran into some issues with this PR. You can reproduce using the manual PopupControlTest or any app that has a standard (non-system) menu bar. Click on "Menu" to show the menu's popup, click on the system menu bar to dismiss it, and then click on "Menu" again. The popup won't appear.

I debugged this a bit. To track this sort of dismissal the Menu control relies on its popup window going through the auto-hide logic (I think auto-hide sets a transient property that Menu monitors to detect when the popup goes away). The auto-hide logic is normally triggered by a focus ungrab event. It looks like there's a similar case in the glass code already; when the user clicks on a window's title bar glass calls _ungrabFocus to ensure auto-hiding popups are closed (see sendEvent: in GlassWindow.m).

And do we have a policy on Copilot? In this case I think it got something half-right but also got several things wrong. I don't like leaving misinformation lying around in a PR but also don't want to spend time correcting an AI.

jperedadnr:
Thanks, @beldenfox, I can reproduce the issue with `PopupControlTest`. I'll have a look.

jperedadnr:
I've just pushed a change, that should fix the issue in PopupControlTest.

modules/javafx.graphics/src/main/native-glass/mac/GlassWindow+Overrides.m
```diff
@@ -308,6 +317,34 @@ - (BOOL)windowShouldClose:(NSNotification *)notification
...
+        [menu cancelTrackingWithoutAnimation];
+    }
+
+    // For auto-hide popups, _resetGrab fires FOCUS_UNGRAB on the popup's owner window, which
+    // will trigger doAutoHide() -> hide() on the popup.
+    [GlassWindow _resetGrab];
+
+    // For non-auto-hide popups, jWindowNotifyClose fires WINDOW_CLOSE_REQUEST,
+    // and that calls hide() on the popup.
+    if ([self->nsWindow isVisible]) {
```
beldenfox: Should non-auto-hide popups be hidden? The spec doesn't talk about them much and I'm not sure what they're used for.

jperedadnr: That's a very good question, indeed.

I'm testing a native Xcode project, and a `NSPanel` by default is non-auto-hide. For instance, `Finder -> Get Info` opens an auxiliary window, and it won't close when opening the system menu bar or clicking elsewhere. 

However, `Finder -> right click on a file-> Context Menu` is auto-hide (hides when clicking the system menu bar or elsewhere). 

The main difference is that the native ContextMenu window is actually a `NSMenu`, while in JavaFX we use `NSPanel` for all popups, and we install event handlers to treat it as auto-hide or non-auto-hide.

With this PR as is, the non-auto-hide popups should be closed when clicking over the system menu bar, but now I see that `PopupWindowPeerListener::closing` overrides `WindowPeerListener::closing` with an empty implementation, and `WINDOW_CLOSE_REQUEST` is not fired at all... 

So my call to `jWindowNotifyClose` in this case is wrong (it doesn't do anything). 

And while the non-auto-hide popup won't close (okay), the click on the system menu bar event is consumed and there is no way to show it while the popup is showing (wrong).

I'll rework this.

jperedadnr: @beldenfox It should be fixed now.

modules/javafx.graphics/src/main/native-glass/mac/GlassMenu.m
```diff
@@ -178,8 +178,13 @@ - (void)action:(id)sender
...
 // JDK-8096139: do not use menuNeedsUpdate here, even though Cocoa prohibits
 // changing the menu structure during menuWillOpen...
-- (void)menuWillOpen: (NSMenu *)menu
+- (void)menuWillOpen: (NSMenu *)_menu
 {
+    // Post notification before menu opens so popup windows can be closed
+    [[NSNotificationCenter defaultCenter] postNotificationName:GlassMenuBarWillOpenNotification
+                                                        object:_menu
+                                                      userInfo:nil];
+
```
beldenfox, line 194: If the system menu is canceled this code will still send out a jMenuOpeningMethod call for it. The matching jMenuClosedMethod will never be sent.

I don't think you need the notification (?). In menuWillOpen: you can just check to see if there's a grab window. If there's a grab window cancel tracking on the menu, clear the grab, and return. But I sort of like the notification, it helps verify that there really is a popup open and visible to the user.

modules/javafx.graphics/src/main/native-glass/mac/GlassWindow+Overrides.m
```diff
@@ -308,6 +317,30 @@ - (BOOL)windowShouldClose:(NSNotification *)notification
...
 }
 
+- (void)menuWillOpenHandler:(NSNotification *)notification {
+    // This handler is only called for popup windows as the observer
+    // is only registered for popups in _createWindowCommonDo
+
+    // If this popup is not visible or is a non-auto-hide popup, return early, without cancelling the menu event
+    // or hiding the popup.
+    // Note that only auto-hide popups have an active focus grab.
+    if (![self->nsWindow isVisible] || ![GlassWindow _hasGrab]) {
```
beldenfox: I'm not sure you need the isVisible check though there's no harm in leaving it in. As far as I know there should never be an invisible popup; when JavaFX hides the popup the NSWindow is destroyed, not hidden.

modules/javafx.graphics/src/main/native-glass/mac/GlassMenu.m
jperedadnr: Not sure what you mean about the system menu being canceled? The menu delegate only has `menuWillOpen:` and `menuDidClose:`, and it would be hard to catch a "cancel event" that prevented the menu from opening or when it was already opened. 
Worse case, if for whatever reason the system menu gets canceled, the popups get closed.
As for the `jmenuClosedMenu` not being sent, that would have been already the case before this PR, wouldn't it?

About notification vs grab: You are right! I could simply use the grab check instead of using the more complex notification pattern. 

This PR initially didn't have the static `hasGrab:` check, and it was simply using the notification. After your feedback, I added it, and now, as you say, we could benefit from having this quick access :

```
- (void)menuWillOpen: (NSMenu *)_menu
{
-    [[NSNotificationCenter defaultCenter] postNotificationName:GlassMenuBarWillOpenNotification ... ] 
+   if ([GlassWindow _hasGrab]) {
+       [_menu cancelTrackingWithoutAnimation];
+        [GlassWindow _resetGrab];
+        return;
+    }

...
```

This will prevent managing the observer lifecycle, broadcasting to multiple popup windows (just to check which one has the grab and then close it), and it implies less code changes. 

On the other hand, as you say, the notification pattern seems more clear, not only for separation of concerns, and maybe for other future cases where the system menu opening event is needed. But for now seems a little bit overkill.
So what do you think? Should I go ahead and change it removing the notification in favour of the grab?

modules/javafx.graphics/src/main/native-glass/mac/GlassWindow+Overrides.m
jperedadnr: If we remove the notification (as discussed in the other comment), this code would be removed.

If we keep it, the visible check is indeed not needed, since there can be only one auto-hide popup with an active grab (and it will be visible), which is the one that should cancel tracking and reset the grab.

Note that I had only the visible check before I added the grab one.

modules/javafx.graphics/src/main/native-glass/mac/GlassMenu.m
beldenfox: > Not sure what you mean about the system menu being canceled?

Sorry, I should have been clearer. The call to `[_menu cancelTrackingWithoutAnimation]` will cause the menu to not actually open but this PR will still send the jMenuOpeningMethod event. Since the menu never opened it will never close so there won't be a matching jMenuClosedMethod event.

> Should I go ahead and change it removing the notification in favour of the grab?

Yes, I think you should remove it. There might be some strange case where s_grabWindow isn't set to nil when it should be and we'll cancel the opening of a system menu unexpectedly. But the chance of that is very small and the problem would correct itself in a single click.

jperedadnr: Done!

beldenfox (approved):
LGTM and my manual testing all passed.

arapte (changes requested):
Sanity testing looks good to me. Tested several scenarios with ChoiceBox, ComboBox, DatePicker, ColorPicker
Could you please add a automated headful system test.

jperedadnr:
@arapte I've added a system test, with a couple of tests: one with auto-hide popups, that fails before this PR and passes after it, and one with non-auto-hide popups that passes before and after this PR (so behaviour doesn't change).

beldenfox (approved):
New test fails in master branch and succeeds in this PR. LGTM.

arapte (approved):
lgtm

PR #2106, opened 2026-03-12
8379942: Some non focus-traversable Controls do not request focus when clicking on it

Maran23:
As also discussed back then in the mailing list, there are weird issues around gaining focus when a `Control` is not focus traversable but got a click event.

- Some Controls do not call `requestFocus()` when they are not focus traversable and receive a mouse click
- It is very inconsistent which Controls do it and which do not. Sometimes, just a part of a `Control` will request focus, while another one will not
- Manually calling `requestFocus()` always works

It seems like there is a misconception between beeing not focus traversable and not requesting focus. The focus traversable property should only affect keyboard navigation really. A mouse click should always request a focus.

Check the Ticket for a reproducer with all `Control`s and a short list which Controls do not behave (and which do).

This PR removes the pattern that was wrongly used in some `Control`s.
From:
```
    if (getNode().isFocusTraversable()) {
        getNode().requestFocus();
    }
```
To:
```
    getNode().requestFocus();
```

---------
### Progress

### Issue
 * [JDK-8379942](https://bugs.openjdk.org/browse/JDK-8379942): Some non focus-traversable Controls do not request focus when clicking on it (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2106/head:pull/2106` \
`$ git checkout pull/2106`

`$ git checkout pull/2106` \
`$ git pull https://git.openjdk.org/jfx.git pull/2106/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2106`

`$ git pr show -t 2106`

</details>

<a href="https://git.openjdk.org/jfx/pull/2106.diff">https://git.openjdk.org/jfx/pull/2106.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2106#issuecomment-4050575186)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -8550,6 +8550,7 @@ public abstract sealed class Node
                         if (get()) {
                             _scene.initializeInternalEventDispatcher();
                         }
+                        focusSetDirty(_scene);
                     }
                 }
 
@@ -8583,8 +8584,9 @@ public abstract sealed class Node
      * determination.
      */
     private void focusSetDirty(Scene s) {
-        if (s != null && this == s.getFocusOwner()) {
-            s.setFocusDirty(true);
+        if (s != null &&
+            (this == s.getFocusOwner() || isFocusTraversable())) {
+                s.setFocusDirty(true);
         }
     }
 
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Node.java
@@ -8550,7 +8550,6 @@ public abstract sealed class Node
                         if (get()) {
                             _scene.initializeInternalEventDispatcher();
                         }
-                        focusSetDirty(_scene);
                     }
                 }
 
@@ -8584,9 +8583,8 @@ public abstract sealed class Node
      * determination.
      */
     private void focusSetDirty(Scene s) {
-        if (s != null &&
-            (this == s.getFocusOwner() || isFocusTraversable())) {
-                s.setFocusDirty(true);
+        if (s != null && this == s.getFocusOwner()) {
+            s.setFocusDirty(true);
         }
     }
 
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
@@ -2408,21 +2408,6 @@ public class Scene implements EventTarget {
     }
 
     /**
-     * Returns true if this scene is quiescent, i.e. it has no activity
-     * pending on it such as CSS processing or layout requests.
-     *
-     * Intended to be used for tests only
-     *
-     * @return boolean indicating whether the scene is quiescent
-     */
-    boolean isQuiescent() {
-        final Parent r = getRoot();
-        return !isFocusDirty()
-               && (r == null || (r.cssFlag == CssFlags.CLEAN &&
-                r.layoutFlag == LayoutFlags.CLEAN));
-    }
-
-    /**
      * A listener for pulses, used for testing. If non-null, this is called at
      * the very end of ScenePulseListener.pulse().
      *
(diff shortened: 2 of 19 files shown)


modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
```diff
@@ -2407,21 +2407,6 @@ void enableInputMethodEvents(boolean enable) {
...
 
-    /**
-     * Returns true if this scene is quiescent, i.e. it has no activity
-     * pending on it such as CSS processing or layout requests.
-     *
-     * Intended to be used for tests only
-     *
-     * @return boolean indicating whether the scene is quiescent
-     */
-    boolean isQuiescent() {
```
Maran23, line 2418: This method was unused.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -8584,9 +8583,8 @@ public String getName() {
      * determination.
      */
     private void focusSetDirty(Scene s) {
-        if (s != null &&
-            (this == s.getFocusOwner() || isFocusTraversable())) {
-                s.setFocusDirty(true);
+        if (s != null && this == s.getFocusOwner()) {
```
Maran23: Thinking about this twice, I will revert this change here.

The focus traversable is not checked in the focus dirty code: https://github.com/openjdk/jfx/blob/11d65f19372dc643a985fbeffab094b76578406c/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java#L2637-L2651

But the `Scene.this.focusInitial();` might change the initial focused node when the focus traversable changed from `false` to `true`.

So the scenario could be:
1. Nothing is the focus owner
2. A `Node` is set to focus traversable `true` (from `false`)
3. Now this code runs and `focusInitial()` is called to traverse to the now traversable `Node`

Maran23:
This exposed a problem inside `VirtualFlow`. This will very likely need a localized fix there. In any case, the concept of this PR can still be reviewed already.

mstr2:
> It seems like there is a misconception between beeing not focus traversable and not requesting focus. The focus traversable property should only affect keyboard navigation really. A mouse click should always request a focus.

Yes, the semantics of `focusTraversable` seem to be a bit unclear. There are three ways a control can receive focus: with keyboard navigation, with a mouse click, or programmatically by invoking `Node.requestFocus()`. (There's also the option of directly calling `setFocused(boolean)`, which I consider to be a defect and we should seriously consider deprecating this method.)

I don't see a whole lot of value of having a control accept focus with a mouse click, but not via keyboard navigation. What's the use case here? I would imagine that if a control is focusable with a mouse click, it should also be focusable with keyboard navigation.

However, I see quite a bit of value of having a control not accept focus _at all_, but still be fully functional otherwise. For example, consider a "Copy" or "Paste" button that act on the currently focused control. It's clear that such a button can never accept focus, as that would break the intended function.

With your proposed change, a button will always accept focus on click. This subtly breaks the cut/copy/paste buttons of `HTMLEditor`: they will now steal the focus, which removes the text cursor from the editor.

I see two ways forward here:
1. We could redefine `focusTraversable` to mean _focusable_, that is, whether it can accept focus at all.
2. If we want `focusTraversable` to only apply to keyboard navigation, then we need another `focusable` property that can be used to prevent controls from stealing focus.

I don't think we can do what you propose and force a control to always accept focus when clicked, but not accept it when keyboard-traversed. That would be a functional regression for applications that rely on the non-stealing focus behavior of buttons.

In any case, we need to discuss this on the mailing list since it affects core semantics of JavaFX.

Maran23:
> However, I see quite a bit of value of having a control not accept focus _at all_, but still be fully functional otherwise. For example, consider a "Copy" or "Paste" button that act on the currently focused control. It's clear that such a button can never accept focus, as that would break the intended function.
> 
> With your proposed change, a button will always accept focus on click. This subtly breaks the cut/copy/paste buttons of `HTMLEditor`: they will now steal the focus, which removes the text cursor from the editor.

This is a very interesting usecase.

> There are three ways a control can receive focus

Yes, other frameworks usually differ between keyboard and mouse focus.

> I see two ways forward here:

> 1. We could redefine focusTraversable to mean focusable, that is, whether it can accept focus at all.
> 2. If we want focusTraversable to only apply to keyboard navigation, then we need another focusable property that can be used to prevent controls from stealing focus.

This seems like a good way. The last time I wrote about those findings, I got no answer (Me and John thought this was a bug).

Maran23:
> There's also the option of directly calling setFocused(boolean), which I consider to be a defect and we should seriously consider deprecating this method

+1, agree.

Maran23:
closing because this is something we probably need to do after we settled with the focus changes.

PR #2124, opened 2026-03-21
8380574: [TestBug] MouseEventFirer should not create a temporary Stage

Maran23:
Very similar to https://github.com/openjdk/jfx/pull/1829, this PR removes the questionable behavior that the `MouseEventFirer` may create a temporary `Stage` for your `Node`.

Take the following test code:
```
Button button = new Button("Button");
MouseEventFirer mouse = new MouseEventFirer(button);
mouse.fireMousePressAndRelease();
mouse.fireMousePressAndRelease();
mouse.dispose()
```
What it does is to create a `Stage` in the first method. The second method does not. This is not immediately clear.
That is also the reason why the dispose method exists. To MAY clean it up. Or forget to call it.

This does not test a realistic scenario and the chance is quite high that developers used that methods without even knowing that it contains such logic.

So the idea is to remove the StageLoader code from MouseEventFirer and rather use it in the Test code before calling the Util methods.

For the example above, this would result in:
```
Button button = new Button("Button");
stageLoader = new StageLoader(button );
MouseEventFirer mouse = new MouseEventFirer(button);
mouse.fireMousePressAndRelease();
mouse.fireMousePressAndRelease();
```

There were only two real tests that did not have a `Stage` yet. So most of the tests already had a good setup and the questionable code did not run mostly.

---------

---------
### Progress

### Issues
 * [JDK-8380574](https://bugs.openjdk.org/browse/JDK-8380574): [TestBug] MouseEventFirer should not create a temporary Stage (**Enhancement** - P4)
 * [JDK-8293356](https://bugs.openjdk.org/browse/JDK-8293356): Testbug: broken usage of MouseEventFirer/VirtualFlowTestUtils (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2124/head:pull/2124` \
`$ git checkout pull/2124`

`$ git checkout pull/2124` \
`$ git pull https://git.openjdk.org/jfx.git pull/2124/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2124`

`$ git pr show -t 2124`

</details>

<a href="https://git.openjdk.org/jfx/pull/2124.diff">https://git.openjdk.org/jfx/pull/2124.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2124#issuecomment-4104824015)
</details>

### DIFF
--- a/modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/infrastructure/MouseEventFirer.java
+++ b/modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/infrastructure/MouseEventFirer.java
@@ -28,14 +28,10 @@ package test.com.sun.javafx.scene.control.infrastructure;
 import javafx.event.Event;
 import javafx.event.EventTarget;
 import javafx.event.EventType;
-import javafx.geometry.BoundingBox;
-import javafx.geometry.Bounds;
 import javafx.geometry.Point2D;
 import javafx.scene.Node;
-import javafx.scene.Scene;
 import javafx.scene.input.MouseButton;
 import javafx.scene.input.MouseEvent;
-import javafx.stage.Window;
 
 import java.util.Arrays;
 import java.util.List;
@@ -48,37 +44,11 @@ import java.util.List;
  * The default local coordinates are at the center of the target.
  */
 public final class MouseEventFirer {
-    private final EventTarget target;
 
-    private final Bounds targetBounds;
-    private StageLoader sl;
+    private final EventTarget target;
 
     public MouseEventFirer(EventTarget target) {
         this.target = target;
-
-        // Force the target node onto a stage so that it is accessible
-        if (target instanceof Node n) {
-            Scene s = n.getScene();
-            Window w = s == null ? null : s.getWindow();
-
-            if (w == null || w.getScene() == null) {
-                sl = new StageLoader(n);
-                targetBounds = n.getLayoutBounds();
-            } else {
-                targetBounds = n.getLayoutBounds();
-            }
-        } else if (target instanceof Scene scene) {
-            sl = new StageLoader(scene);
-            targetBounds = new BoundingBox(0, 0, scene.getWidth(), scene.getHeight());
-        } else {
-            throw new RuntimeException("EventTarget of invalid type (" + target + ")");
-        }
-    }
-
-    public void dispose() {
-        if (sl != null) {
-            sl.dispose();
-        }
     }
 
     public void fireMousePressAndRelease(KeyModifier... modifiers) {
@@ -160,15 +130,22 @@ public final class MouseEventFirer {
     }
 
     private void fireMouseEvent(EventType<MouseEvent> evtType, MouseButton button, int clickCount, double deltaX, double deltaY, KeyModifier... modifiers) {
+        if (!(target instanceof Node node)) {
+            throw new IllegalArgumentException("EventTarget: " + target + " must be a Node or needs special handling.");
+        }
+
+        if (node.getScene() == null || node.getScene().getWindow() == null) {
+            throw new IllegalArgumentException("Node: " + node + " must be in a Scene and Window " +
+                    "so that a correct MouseEvent can be fired.");
+        }
+
         // width / height of target node
-        final double w = targetBounds.getWidth();
-        final double h = targetBounds.getHeight();
+        final double w = node.getLayoutBounds().getWidth();
+        final double h = node.getLayoutBounds().getHeight();
 
         // x / y click position is centered
         final double x = w / 2.0 + deltaX;
         final double y = h / 2.0 + deltaY;
-
-        Node node = (Node) target;
 
         Point2D localP = new Point2D(x, y);
         Point2D sceneP = node.localToScene(localP);
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/cell/TextFieldTableCellTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/cell/TextFieldTableCellTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2013, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -30,19 +30,15 @@ import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
-import javafx.beans.property.ReadOnlyStringWrapper;
 import javafx.collections.FXCollections;
 import javafx.scene.control.TableCell;
 import javafx.scene.control.TableColumn;
 import javafx.scene.control.TableView;
-import javafx.scene.control.TextField;
 import javafx.scene.control.cell.TextFieldTableCell;
-import javafx.scene.input.MouseButton;
 import javafx.util.Callback;
 import javafx.util.StringConverter;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import test.com.sun.javafx.scene.control.infrastructure.MouseEventFirer;
 
 public class TextFieldTableCellTest {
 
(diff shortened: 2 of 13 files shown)


modules/javafx.controls/src/test/java/test/javafx/scene/control/cell/TextFieldTableCellTest.java
```diff
@@ -370,44 +366,4 @@ public void setup() {
...
-        cell.updateIndex(0);
-        cell.updateTableColumn(tc);
-        cell.setEditable(true);
-
-        tableView.edit(0, tc);
-        assertTrue(cell.isEditing());
-        assertNotNull(cell.getGraphic());
-
-        TextField textField = (TextField) cell.getGraphic();
-        MouseEventFirer mouse = new MouseEventFirer(textField);
```
Maran23, line 398: This is what happened before.
We created the following structure:
```
Group
└ TextField
```
And then testing the `TextField` in isolation without the table or cell (since both are not inside the scene graph).

Guess what, I also reverted the change, this testcase was supposed to test, and still green. It does not make any sense. So removed it. (And now, we avoid that this kind of test can be written, as you need to take care of the scene graph by e.g. just using the `stageLoader`).

andy-goryachev-oracle:
Most of the changes are trivial and in the tests, but I think the second pair of eyes wouldn't hurt.

/reviewers 2

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/infrastructure/MouseEventFirer.java
```diff
@@ -160,16 +130,23 @@ public void fireMouseEvent(EventType<MouseEvent> evtType, MouseButton button, do
     }
 
     private void fireMouseEvent(EventType<MouseEvent> evtType, MouseButton button, int clickCount, double deltaX, double deltaY, KeyModifier... modifiers) {
+        if (!(target instanceof Node node)) {
+            throw new IllegalArgumentException("EventTarget: " + target + " must be a Node or needs special handling.");
+        }
+
+        if (node.getScene() == null || node.getScene().getWindow() == null) {
+            throw new IllegalArgumentException("Node: " + node + " must be in a Scene and Window " +
```
andy-goryachev-oracle, line 138: maybe instruct to use StageLoader instead, addin an example in the comments?
something like this:
```
// The target node must be in a Scene and Window, example:
// private StageLoader sl;
// ..
// sl = new StageLoader(node);
// ...
// sl.dispose()
if (node.getScene() == null || node.getScene().getWindow() == null) {
            throw new IllegalArgumentException("Target node must be in a Scene and Window (use StageLoader?)");
```

modules/javafx.controls/src/test/java/test/javafx/scene/control/cell/TextFieldTableCellTest.java
andy-goryachev-oracle: I am a bit confused - you reverted the fix for https://bugs.openjdk.org/browse/JDK-8119995 ?

I am a bit reluctant to remove the test, can we keep it?

Maran23: Yes, I reverted the fix and this test is still green.
Because as I said, it does not test the table at all. The table (and therefore `VirtualFlow`, which was adjusted) is never in the scene graph.

We just spin up:
```
Group
└ TextField
```
And do 2 right click in the `TextField`. No table attached, nothing that is really tested here. To put it mildly, this test does absolutely nothing. 

If you want me to keep it, I will write:
`stageLoader = new StageLoader(textField);`
and we have the same behavior and a green test that tests nothing.

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/infrastructure/MouseEventFirer.java
Maran23: I thought about this as well, but my future plan (sneakpeak :)) is to rather have an abstract class that provides all needed methods, including the `stageLoader` cleanup and exception handler setup. And all controls test extend it. Saves us a lot of duplicate code, and if we need to change something there, we can do it just once.

But the biggest benefit why I want to push this is to make it easier to write tests for new contributors, as there is essentially only one way to spin up a stage, fire mouse events, all in one class and used in every test. No ambiguity.

And in this case, we would not really need this comment (I want to document everything on the abstract class then). If that seems good for you too, I would like to move on with that one time and rather do not clutter this class (or other test classes, effectively spreading all information around different test classes).

Maran23: (For reference, I followed the same message as in: https://github.com/openjdk/jfx/pull/1829)

andy-goryachev-oracle: simplifying the task of writing test cases is a noble goal.

modules/javafx.controls/src/test/java/test/javafx/scene/control/cell/TextFieldTableCellTest.java
andy-goryachev-oracle: unless you want to fix the test, "let's minimize the changes".
there is a non-zero probability of someone breaking the code in some other way and this test catching it, right?

Maran23: > there is a non-zero probability of someone breaking the code in some other way and this test catching it, right?

I guess right clicking into a `TextField` keeps the selected text. (`TextFieldTest` would be a better place then). What would your preference be here?

andy-goryachev-oracle: could we keep the test?

Maran23: Keep in this file, or change to a `TextField` test in `TextFieldTest` class, so same test logic but without the unused Table?

andy-goryachev-oracle: I don't understand - this test is about `TextFieldTableCell` (using `TableView` and everything) why do you want to move it to `TextFieldTest`?

Maran23: Because we only test the `TextField` right now. Current state of the test. Thats why I deleted it. Keeping it the same way will still only do that.

andy-goryachev-oracle: I could be mistaken, but it tests the TextField as a part of a larger construct.  The test _might_ be faulty, but that's another topic.

I don't feel comfortable just removing it.

Maran23: > I could be mistaken, but it tests the TextField as a part of a larger construct. The test _might_ be faulty, but that's another topic.

Yeah it tried, but failed. As I said.
Lets break it down. This is the very questionable fix:
https://github.com/openjdk/jfx/blob/0500668190ca16772f89b9a07987bc9df4c56fdc/modules/javafx.controls/src/main/java/javafx/scene/control/skin/VirtualFlow.java#L475-L498

Conditions:
- `VirtualFlow` must be in a scene graph -> not fulfilled
- Cells must be a part of the `VirtualFlow` -> not fulfilled (created with new)
- `TextField` must be a part of the Cell -> not fulfilled, it will be alone in the `Stage`, without its parent cell

I will now have a look if it can be easily rewritten so those conidtions are true. But I will not waste to much time, as the fix is also very questionable as well. Otherwise, I will restore what was there before, no changes.

Maran23: This is the test coverage:
image
image

Maran23: test improved, works the same, but with a real stage setup. All three Conditions above are now fulfilled.

modules/javafx.controls/src/test/java/test/javafx/scene/control/cell/TextFieldTableCellTest.java
```diff
@@ -370,44 +383,37 @@ public void setup() {
...
         mouse.fireMousePressed(MouseButton.SECONDARY);
-        assertEquals("Dummy Text", textField.getSelectedText());
+        assertEquals(text, textField.getSelectedText());
         mouse.fireMouseReleased(MouseButton.SECONDARY);
-        assertEquals("Dummy Text", textField.getSelectedText());
-
-        mouse.dispose();
+        assertEquals(text, textField.getSelectedText());
     }
+
```
andy-goryachev-oracle: extra newline

andy-goryachev-oracle (approved):
thank you for updating the test!

andy-goryachev-oracle:
@arapte or @lukostyra could one of you be the second reviewer?

andy-goryachev-oracle (approved):

Maran23:
/issue add JDK-8293356

Found this one today which, 4 years ago, already mentioned the broken usage of `MouseEventFirer` with the implicit stage creation.

lukostyra (approved):
Looks good, tests work correctly on Windows as well.

PR #2127, opened 2026-03-25
8380926: HeavyweightDialog does not check for a valid stage

crschnick:
This fixes issues with Alerts when the owner stage is not a valid stage with a scene and currently showing

---------

---------
### Progress

### Issue
 * [JDK-8380926](https://bugs.openjdk.org/browse/JDK-8380926): HeavyweightDialog does not check for a valid stage (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2127/head:pull/2127` \
`$ git checkout pull/2127`

`$ git checkout pull/2127` \
`$ git pull https://git.openjdk.org/jfx.git pull/2127/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2127`

`$ git pr show -t 2127`

</details>

<a href="https://git.openjdk.org/jfx/pull/2127.diff">https://git.openjdk.org/jfx/pull/2127.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2127#issuecomment-4127787679)
</details>

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/HeavyweightDialog.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/HeavyweightDialog.java
@@ -315,7 +315,7 @@ class HeavyweightDialog extends FXDialog {
             stage.renderScaleYProperty().unbind();
 
             Scene oldScene = oldStage.getScene();
-            if (oldScene != null && scene != null && dialogScene != null) {
+            if (oldScene != null && dialogScene != null) {
                 Bindings.unbindContent(dialogScene.getStylesheets(), oldScene.getStylesheets());
             }
         }
@@ -327,7 +327,7 @@ class HeavyweightDialog extends FXDialog {
             stage.renderScaleYProperty().bind(newStage.renderScaleYProperty());
 
             Scene newScene = newStage.getScene();
-            if (newScene != null && scene != null && dialogScene != null) {
+            if (newScene != null && dialogScene != null) {
                 Bindings.bindContent(dialogScene.getStylesheets(), newScene.getStylesheets());
             }
         }
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/HeavyweightDialog.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/HeavyweightDialog.java
@@ -52,7 +52,7 @@ class HeavyweightDialog extends FXDialog {
     final Stage stage = new Stage() {
         @Override public void centerOnScreen() {
             Window owner = HeavyweightDialog.this.getOwner();
-            if (owner != null) {
+            if (owner != null && owner.getScene() != null && owner.isShowing() && (!(owner instanceof Stage ownerStage) || !ownerStage.isIconified())) {
                 positionStage();
             } else {
                 if (getWidth() > 0 && getHeight() > 0) {
@@ -309,27 +309,25 @@ class HeavyweightDialog extends FXDialog {
     private void updateStageBindings(Window oldOwner, Window newOwner) {
         final Scene dialogScene = stage.getScene();
 
-        if (oldOwner != null && oldOwner instanceof Stage) {
-            Stage oldStage = (Stage) oldOwner;
+        if (oldOwner instanceof Stage oldStage) {
             Bindings.unbindContent(stage.getIcons(), oldStage.getIcons());
             stage.renderScaleXProperty().unbind();
             stage.renderScaleYProperty().unbind();
 
             Scene oldScene = oldStage.getScene();
-            if (scene != null && dialogScene != null) {
+            if (oldScene != null && scene != null && dialogScene != null) {
                 Bindings.unbindContent(dialogScene.getStylesheets(), oldScene.getStylesheets());
             }
         }
 
         // put the icons and stylesheets of the owner window into the dialog
-        if (newOwner instanceof Stage) {
-            Stage newStage = (Stage) newOwner;
+        if (newOwner instanceof Stage newStage) {
             Bindings.bindContent(stage.getIcons(), newStage.getIcons());
             stage.renderScaleXProperty().bind(newStage.renderScaleXProperty());
             stage.renderScaleYProperty().bind(newStage.renderScaleYProperty());
 
             Scene newScene = newStage.getScene();
-            if (scene != null && dialogScene != null) {
+            if (newScene != null && scene != null && dialogScene != null) {
                 Bindings.bindContent(dialogScene.getStylesheets(), newScene.getStylesheets());
             }
         }
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/AlertTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/AlertTest.java
@@ -30,11 +30,13 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 import java.util.Locale;
 import javafx.application.Platform;
+import javafx.scene.Scene;
 import javafx.scene.control.Alert;
 import javafx.scene.control.ButtonType;
 import javafx.scene.control.Dialog;
 import javafx.scene.control.DialogShim;
 import javafx.scene.control.HeavyweightDialogShim;
+import javafx.scene.layout.Region;
 import javafx.stage.Stage;
 import javafx.stage.WindowEvent;
 import org.junit.jupiter.api.AfterEach;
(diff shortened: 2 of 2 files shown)


modules/javafx.controls/src/main/java/javafx/scene/control/HeavyweightDialog.java
```diff
@@ -309,27 +309,25 @@ private void positionStage() {
...
-        if (oldOwner != null && oldOwner instanceof Stage) {
-            Stage oldStage = (Stage) oldOwner;
+        if (oldOwner instanceof Stage oldStage) {
             Bindings.unbindContent(stage.getIcons(), oldStage.getIcons());
             stage.renderScaleXProperty().unbind();
             stage.renderScaleYProperty().unbind();
 
             Scene oldScene = oldStage.getScene();
-            if (scene != null && dialogScene != null) {
+            if (oldScene != null && scene != null && dialogScene != null) {
```
Maran23: A bit confused by the code. So, do we need to check `scene != null` here?

modules/javafx.controls/src/main/java/javafx/scene/control/HeavyweightDialog.java
```diff
@@ -309,27 +309,25 @@ private void positionStage() {
...
-        if (newOwner instanceof Stage) {
-            Stage newStage = (Stage) newOwner;
+        if (newOwner instanceof Stage newStage) {
             Bindings.bindContent(stage.getIcons(), newStage.getIcons());
             stage.renderScaleXProperty().bind(newStage.renderScaleXProperty());
             stage.renderScaleYProperty().bind(newStage.renderScaleYProperty());
 
             Scene newScene = newStage.getScene();
-            if (scene != null && dialogScene != null) {
+            if (newScene != null && scene != null && dialogScene != null) {
```
Maran23: Do we need to check `scene != null` here?

modules/javafx.controls/src/main/java/javafx/scene/control/HeavyweightDialog.java
```diff
@@ -52,7 +52,7 @@ class HeavyweightDialog extends FXDialog {
     final Stage stage = new Stage() {
         @Override public void centerOnScreen() {
             Window owner = HeavyweightDialog.this.getOwner();
-            if (owner != null) {
+            if (owner != null && owner.getScene() != null && owner.isShowing() && (!(owner instanceof Stage ownerStage) || !ownerStage.isIconified())) {
```
Maran23, line 55: A test for the new condition: `owner.isShowing() && (!(owner instanceof Stage ownerStage) || !ownerStage.isIconified())` would also be a good idea

modules/javafx.controls/src/main/java/javafx/scene/control/HeavyweightDialog.java
crschnick: I removed the check, the older parts of the code often don't make a lot of sense

modules/javafx.controls/src/main/java/javafx/scene/control/HeavyweightDialog.java
crschnick: This would have to be manually tested to verify that the window is properly centered. I think automating that is quite hard

Maran23: Maybe we can just check if `x` and `y` are set or something like that?

crschnick: That would depend on the screen configuration which x and y values are right or wrong in this case. It also depends on the OS, the old (wrong) centering code resulted in different levels of offsets from the center on different systems for me

Maran23: As far as I can see, it uses the owner `Scene` and `DialogPane` sizes, so it might be possible to set all those values and get predictable X and Y values.

And since the tests are headless with the stub toolkit, it might be indeed predictable (same code should run for all OS). Worth a try at least

crschnick: Alright, I added tests for this. It should work if the screen sizes for the tests are the same everywhere

Maran23: I think it should!
See here:

https://github.com/openjdk/jfx/blob/a9703fa6e3c9b70e35207d646440849f3eee29e4/modules/javafx.graphics/src/test/java/test/com/sun/javafx/pgstub/StubToolkit.java#L120-L124

which is used for all headless tests. So this seems good!

andy-goryachev-oracle:
For the second part where the parent `Stage` has a `Scene` but is not visible, where do you expect the `Alert` to be positioned?  On macOS, it shows up at the top left corner which is what I would expect.

crschnick:
Hmm that should always be centered in that case. I will try this out on my mac when I have time.

At least on Windows and in the headless tests the alert is centered

andy-goryachev-oracle:
We'll need to test the fix on all three platforms (windows/macOS/linux)

/reviewers 2

andy-goryachev-oracle:
So the NPE is gone, very good.

What is not yet clear to me is how it chooses where to show the alert when the parent Stage is not shown.  Arguably, this is not a good state, so I'd expect the alert to show maybe at the center of the screen (which screen, btw, if we have two?)

So, using this reproducer
https://github.com/andy-goryachev-oracle/Test/blob/main/src/goryachev/bugs/Alert_NPE_8380926.java

where the NPE button focuses on a different stage before showing up the alert, I get the following picture (top menu / bottom dock is trimmed).  Before clicking on the NPE button, I moved both of the windows from their original positions: 

Screenshot 2026-04-07 at 09 27 30

What I see is the alert showing up not at the center of the screen but somewhere closer to the top side of the screen.
Similarly, if I move both stages to a different screen, the alert still appears on the primary screen.

Why is that?

tsayao:
@andy-goryachev-oracle 

I think it is the default behaviour. Y is placed on 1/3 of the screen.

From `Window.java`:

```
    private static final float CENTER_ON_SCREEN_X_FRACTION = 1.0f / 2;
    private static final float CENTER_ON_SCREEN_Y_FRACTION = 1.0f / 3;

    /**
     * Sets x and y properties on this Window so that it is centered on the
     * current screen.
     * The current screen is determined from the intersection of current window bounds and
     * visual bounds of all screens.
     */
    public void centerOnScreen() {
        xExplicit = false;
        yExplicit = false;
        if (peer != null) {
            Rectangle2D bounds = getWindowScreen().getVisualBounds();
            double centerX =
                    bounds.getMinX() + (bounds.getWidth() - getWidth())
                                           * CENTER_ON_SCREEN_X_FRACTION;
            double centerY =
                    bounds.getMinY() + (bounds.getHeight() - getHeight())
                                           * CENTER_ON_SCREEN_Y_FRACTION;

            x.set(centerX);
            y.set(centerY);
            peerBoundsConfigurator.setLocation(centerX, centerY,
                                               CENTER_ON_SCREEN_X_FRACTION,
                                               CENTER_ON_SCREEN_Y_FRACTION);
            applyBounds();
        }
    }
```

I remember seeing some change about not moving parent window with the modal on macos.

andy-goryachev-oracle:
Thank you @tsayao , good to know.

Looks like this is a perfectly acceptable behavior, given that an alert with a non-visible Stage is a bad scenario to begin with.

modules/javafx.controls/src/main/java/javafx/scene/control/HeavyweightDialog.java
```diff
@@ -309,27 +309,25 @@ private void positionStage() {
...
-        if (oldOwner != null && oldOwner instanceof Stage) {
-            Stage oldStage = (Stage) oldOwner;
+        if (oldOwner instanceof Stage oldStage) {
             Bindings.unbindContent(stage.getIcons(), oldStage.getIcons());
             stage.renderScaleXProperty().unbind();
             stage.renderScaleYProperty().unbind();
 
             Scene oldScene = oldStage.getScene();
-            if (scene != null && dialogScene != null) {
+            if (oldScene != null && dialogScene != null) {
```
andy-goryachev-oracle, line 318: is this the actual fix for the NPE?

modules/javafx.controls/src/test/java/test/javafx/scene/control/AlertTest.java
```diff
@@ -327,4 +329,46 @@ private void assertCloseRequestAccepted(Dialog<?> dialog, boolean normalClose) {
...
+
+    @Test public void alert_owner_noScene() {
+        dialog = new Alert(Alert.AlertType.NONE, "Hello World!");
+        dialog.getDialogPane().getButtonTypes().addAll(ButtonType.APPLY, ButtonType.CANCEL);
+        dialog.initOwner(new Stage());
+        assertResultValue(ButtonType.CANCEL, dialog, true);
+        assertCloseRequestAccepted(dialog, true);
+    }
+
+    @Test public void alert_owner_iconifiedCentering() {
```
andy-goryachev-oracle, line 341: minor: we probably should place `@Test` annotations on its own line in the new code.

modules/javafx.controls/src/main/java/javafx/scene/control/HeavyweightDialog.java
andy-goryachev-oracle: is it?

andy-goryachev-oracle (approved):
lgtm, some minor comments.

modules/javafx.controls/src/test/java/test/javafx/scene/control/AlertTest.java
Maran23: and maybe use camelCase completely? Since it is already used inside `iconifiedCentering`

Maran23:
Also note that the AI policy line is missing in your PR description. You can add it manually or with the command (there was a mail about it as well for more background).

andy-goryachev-oracle:
also needs `/template append` and the checkbox

modules/javafx.controls/src/test/java/test/javafx/scene/control/AlertTest.java
crschnick: I just settled on keeping it in line with the rest of the test file. The naming for tests is all over the place anyways, so I don't think this matters

modules/javafx.controls/src/main/java/javafx/scene/control/HeavyweightDialog.java
crschnick: This is part of the fix, there were multiple NPEs. The other part was done in L55

crschnick:
@andy-goryachev-oracle About the reproducer you linked, I tested it on macOS both with and without this fix.

Without this fix, the alert is shown in the top left corner for me. WIth the fix, it is properly centered. I only have one screen, so I can't comment on more complex setups. But the alert showing centered on the main screen without an owner with this fix is correct.

Maran23 (approved):
Looks good, fixes the NPE and the new condition makes sense as well.

PR #2128, opened 2026-03-25
8380935: TextFieldSkin caret width calculation is broken

crschnick:
This fixes a regression caused by the fix for https://bugs.openjdk.org/browse/JDK-8282290. The old if else block was broken by the other PR, which caused the calculation for the caret width to be never run.

---------

---------
### Progress

### Issue
 * [JDK-8380935](https://bugs.openjdk.org/browse/JDK-8380935): TextFieldSkin caret width calculation is broken (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2128/head:pull/2128` \
`$ git checkout pull/2128`

`$ git checkout pull/2128` \
`$ git pull https://git.openjdk.org/jfx.git pull/2128/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2128`

`$ git pr show -t 2128`

</details>

<a href="https://git.openjdk.org/jfx/pull/2128.diff">https://git.openjdk.org/jfx/pull/2128.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2128#issuecomment-4128632039)
</details>

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
@@ -252,11 +252,14 @@ public class TextFieldSkin extends TextInputControlSkin<TextField> {
                  * caret position is not updated.
                  * The caret pos is invalid in this case,
                  * hence it should be updated when caret path size is not 4 */
+
+                if (caretPath.getElements().size() != 0) {
+                    caretWidth = Math.round(caretPath.getLayoutBounds().getWidth());
+                }
+
                 updateTextNodeCaretPos(control.getCaretPosition());
             } else if (caretPath.getElements().size() == 4) {
                 // The caret is split. Ignore and keep the previous width value.
-            } else {
-                caretWidth = Math.round(caretPath.getLayoutBounds().getWidth());
             }
         });
 
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
@@ -247,19 +247,21 @@ public class TextFieldSkin extends TextInputControlSkin<TextField> {
         caretPath.layoutXProperty().bind(textTranslateX);
         textNode.caretShapeProperty().addListener(observable -> {
             caretPath.getElements().setAll(textNode.caretShapeProperty().get());
-            if (caretPath.getElements().size() != 4) {
-                /* On replacing same text using keyboard shortcut,
-                 * caret position is not updated.
-                 * The caret pos is invalid in this case,
-                 * hence it should be updated when caret path size is not 4 */
 
-                if (caretPath.getElements().size() != 0) {
+            /* On replacing same text using keyboard shortcut,
+             * caret position is not updated.
+             * The caret pos is invalid in this case,
+             * hence it should be updated when caret path size is not 4 */
+            boolean hasSplitCaret = caretPath.getElements().size() == 4;
+
+            // If the caret is split, ignore and keep the previous width value.
+            if (!hasSplitCaret) {
+                boolean hasCaret = caretPath.getElements().size() > 0;
+                if (hasCaret) {
                     caretWidth = Math.round(caretPath.getLayoutBounds().getWidth());
                 }
 
                 updateTextNodeCaretPos(control.getCaretPosition());
-            } else if (caretPath.getElements().size() == 4) {
-                // The caret is split. Ignore and keep the previous width value.
             }
         });
 
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/TextFieldTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/TextFieldTest.java
@@ -623,11 +623,11 @@ public class TextFieldTest {
         stage.show();
 
         assertTrue(txtField.getWidth() > TextInputSkinShim.getTextNode(txtField).getLayoutBounds().getWidth());
-        assertEquals(TextInputSkinShim.getTextTranslateX(txtField), 1, 0.0);
+        assertEquals(1, TextInputSkinShim.getTextTranslateX(txtField), 0.0);
 
         txtField.setText("This is a long text. this is  long text.");
         assertTrue(txtField.getWidth() < TextInputSkinShim.getTextNode(txtField).getLayoutBounds().getWidth());
-        assertEquals(0, TextInputSkinShim.getTextTranslateX(txtField), 0.0);
+        assertEquals(1, TextInputSkinShim.getTextTranslateX(txtField), 0.0);
     }
 
     @Test
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/TextFieldTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/TextFieldTest.java
@@ -623,7 +623,7 @@ public class TextFieldTest {
         stage.show();
 
         assertTrue(txtField.getWidth() > TextInputSkinShim.getTextNode(txtField).getLayoutBounds().getWidth());
-        assertEquals(TextInputSkinShim.getTextTranslateX(txtField), 0, 0.0);
+        assertEquals(TextInputSkinShim.getTextTranslateX(txtField), 1, 0.0);
 
         txtField.setText("This is a long text. this is  long text.");
         assertTrue(txtField.getWidth() < TextInputSkinShim.getTextNode(txtField).getLayoutBounds().getWidth());
@@ -646,7 +646,7 @@ public class TextFieldTest {
 
         txtField.setText("This is a long text. this is  long text.");
         assertTrue(txtField.getWidth() < TextInputSkinShim.getTextNode(txtField).getLayoutBounds().getWidth());
-        assertEquals(0, TextInputSkinShim.getTextTranslateX(txtField), 0.0);
+        assertEquals(1, TextInputSkinShim.getTextTranslateX(txtField), 0.0);
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


andy-goryachev-oracle:
Looks like a number of tests is currently failing.  Also, could you add a unit test specifically for this fix?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
```diff
@@ -252,11 +252,14 @@ public TextFieldSkin(final TextField control) {
...
+
+                if (caretPath.getElements().size() != 0) {
+                    caretWidth = Math.round(caretPath.getLayoutBounds().getWidth());
+                }
+
                 updateTextNodeCaretPos(control.getCaretPosition());
             } else if (caretPath.getElements().size() == 4) {
                 // The caret is split. Ignore and keep the previous width value.
-            } else {
-                caretWidth = Math.round(caretPath.getLayoutBounds().getWidth());
```
Maran23, line 259: I'm confused, isn't this effectively dead code?

`if (!4) { .. } elseif (4) { .. } else { when do we come here? }`

crschnick:
I think the tests are wrong. I looked at the logic for the caret at https://github.com/openjdk/jfx/blob/d695b3162faa310f955fb4f17aa11637b1979178/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java#L868 and if I understand it correctly, if the caret is fully to the left, the text should be shifted by caretWidth / 2. The tests assume that the text is not shifted in this case, which I don't think is accurate. If that would be the case, then the caret would be invisible

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
crschnick: Yes, it was. But it shouldn't have been

Maran23: Thanks for confirming. Yeah agree, that looks wrong.

crschnick: Looking at the code review for the commit that caused this, that is a perfect example of how diffs can mislead. The fact that it introduced dead code was not visible from the diffs, only if you looked at the whole context.

Maran23: And I even made a (IMO good) suggestion to improve the if condition 🙂
-> I made this comment because I felt like the if-else condition is confusing. So turns out, it is.

I would therefore recommend to improve it now and remove the else block. It is empty anyway.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
```diff
@@ -252,11 +252,14 @@ public TextFieldSkin(final TextField control) {
                  * caret position is not updated.
                  * The caret pos is invalid in this case,
                  * hence it should be updated when caret path size is not 4 */
+
+                if (caretPath.getElements().size() != 0) {
```
Maran23: `!caretPath.getElements().isEmpty()`

andy-goryachev-oracle:

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
```diff
@@ -252,11 +252,14 @@ public TextFieldSkin(final TextField control) {
...
+                    caretWidth = Math.round(caretPath.getLayoutBounds().getWidth());
```
andy-goryachev-oracle, line 261: I wonder if this should be `Region.snapSizeX()` instead of `Math.round()`:
not only Math.round() may round downward, but also produce mid-pixel results on systems with fractional scale.

Can you double check it on Windows at fractional scales (125%, 150%, 175%, 225%)?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
andy-goryachev-oracle: I think it's out of scope for this PR, for several reasons.

This code relies on current, undocumented behavior of the caret shape being either a single line or two lines encoded as `MoveTo`-`LineTo` pairs.  There are plans to change the caret shape for certain BIDI/RTL scenarios to indicate the direction, so the old assumptions might not hold anymore.

Also, in jfx25 we've added `LayoutInfo`/`CaretInfo` APIs in the `Text` and `TextFlow`, which should provide the caret information without relying on implementation details.

related umbrellas:
https://bugs.openjdk.org/browse/JDK-8343557
https://bugs.openjdk.org/browse/JDK-8300569

andy-goryachev-oracle (commented):
Looks like the problem is fixed on macOS, but it should be tested on Windows at fractional scales (I suspect the fix is not exactly right for fractional scales).

Also, it would help to provide a bit more information about the fix in the PR description (explain the change).

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
Maran23: > There are plans to change the caret shape for certain BIDI/RTL scenarios to indicate the direction, so the old assumptions might not hold anymore.

When that happens, this code of course need to be adjusted. But they will probably require more code changes. 
In any case, the empty `else if` block does not make sense

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
Maran23: Since `caretWidth` is not directly used here, but rather in other code to calculate the translation or related, it is very likely fine here, but the target values in the other calculations may need to be snapped.

crschnick: The caretWidth value is included in various calculations, and the results of those are snapped. So I think this works correctly. This PR just restores the old behaviour before it was broken.

I will check on Windows for various scales

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
andy-goryachev-oracle: This is an area of constant disagreement: I think it's ok to have minor cleanup like this one here, while @kevinrushforth much prefers to keep the PR focused on the fix and create a separate follow-up tickets when required, the rationale being backporting and merge conflicts.  Piling up unrelated changes also delays the fix.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
crschnick: Ok so Andy was right, the caret is sometimes still hidden on the right on different display scales. But even snapping the caret width does not always fix it. It requires a bigger fix for the skin in general.

However, I would argue that this is a different issue unrelated to this. I could open a follow-up to make the TextFieldSkin handle the display scale properly

Maran23: +1 for a follow up. I think not all final values that are later set in the translate etc. are snapped as of now. So we probably need to catch all locations at once.

andy-goryachev-oracle: Do I understand you correctly - the issue is still seen with the fix at fractional scales?  If so, then it would make sense to fix it in this PR.

crschnick: Yes, the issue of the missing caret still exists at some fractional scales. However, I would say that this is a separate issue, more general to the whole skin implementation.

So there are two issues here, which cause the same problem in some scenarios, but the causes and potential fixes don't really have a lot to do with each other. This one just repairs a small part that was broken during and earlier PR. To fix the caret issue on fractional scales, that would require a more investigation.

The JBS issue focuses on the broken caret width calculation, not the missing caret behavior so I think splitting this up into a separate issue makes more sense

andy-goryachev-oracle:
How would you propose to test this, from the user perspective?
_Yes, it works, but only partially?_
What is the goal here?

crschnick:
The caret shows on non-fractional display scales, which is already better than before. The user can test this.

In the other cases where it still doesn't show, that is not due to this fix being wrong, but due to a completely separate issue of how the skin is implemented. That can be implemented in another, more large-scale issue. There are probably more issues caused by the skin not properly snapping values than just the caret not showing if you go looking for them at fractional scales.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
crschnick: I refactored it to make it easier to understand what the condition is checking now

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TextFieldSkin.java
Maran23: I think the newest version looks very good and should still be easy to e.g. backport. And is now more clear, with variables that express the purpose (which makes comments obsolete, good!).

andy-goryachev-oracle:
The fix seems to work on windows, even with fractional scale (tested at 125%, 150%, 175%, 225%).

I also seems to work ok with the split caret, though I can't meaningfully test due to https://bugs.openjdk.org/browse/JDK-8242616 .

andy-goryachev-oracle (approved):
thank you for fixing the issue!
sorry for a long delay.

andy-goryachev-oracle:
Please issue `/template append` command and resolve the ai checkbox.

crschnick:
I honestly don't know what I tested back then, because now it works properly with all the scaling factors. Probably messed up the local jfx branches

andy-goryachev-oracle:
> Probably messed up the local jfx branches

Happened to me as well.  That's why I always merge with the latest master and rebuild completely for each review.

Maran23 (approved):
Looks good to me too and I like the new code more, as it is easier to understand just by reading the variable names.

If you like, you can adjust the copyright header (year), will reapprove if you do.

PR #2129, opened 2026-03-27
8222454: TreeTableView NullPointerException at doubleclick on empty "row"

plumstone:
This fixes the issue adressed here https://bugs.openjdk.org/browse/JDK-8222454

The added TreeTableViewTests were failing without the main code changes. I added the other tests to ensure the consistent functionality of related cell based components.
The method getItemCount() needed to be moved to CellBehaviorBase to do the necessary index check in doSelect() below. All subclasses now implement this method based on the cell container.

There is still some refactoring potential here, to reduce code duplications and ensure consistent behavior of all cell based components. This could be done in a follow-up PR.

---------
### Progress

### Issue
 * [JDK-8222454](https://bugs.openjdk.org/browse/JDK-8222454): TreeTableView NullPointerException at doubleclick on empty "row" (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2129/head:pull/2129` \
`$ git checkout pull/2129`

`$ git checkout pull/2129` \
`$ git pull https://git.openjdk.org/jfx.git pull/2129/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2129`

`$ git pr show -t 2129`

</details>

<a href="https://git.openjdk.org/jfx/pull/2129.diff">https://git.openjdk.org/jfx/pull/2129.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2129#issuecomment-4145602176)
</details>


modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/TableCellBehaviorBase.java
```diff
@@ -101,16 +100,16 @@ protected void doSelect(final double x, final double y, final MouseButton button
         // for out of bounds indexes. So, need to check
         final C tableCell = getNode();
 
+        int count = getItemCount();
```
plumstone, line 108: Moved up to check it earlier on

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/CellBehaviorBase.java
```diff
@@ -164,7 +164,7 @@ protected boolean isClickPositionValid(final double x, final double y) {
     }
 
     protected int getIndex() {
-        return getNode() instanceof IndexedCell ? ((IndexedCell<?>)getNode()).getIndex() : -1;
+        return getNode() != null ? getNode().getIndex() : -1;
```
Maran23: Can `getNode()` return null?

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/CellBehaviorBase.java
```diff
@@ -206,9 +206,12 @@ protected void doSelect(final double x, final double y, final MouseButton button
 
         final Control cellContainer = getCellContainer();
 
+        int count = getItemCount();
+        if (cell.getIndex() >= count) return;
```
Maran23: I think we should rather use the `if` with the braces.
```
if (cell.getIndex() >= count) {
    return;
}
```

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/TreeTableRowBehavior.java
```diff
@@ -52,6 +52,11 @@ public TreeTableRowBehavior(TreeTableRow<T> control) {
      *                                                                         *
      **************************************************************************/
 
+    /** {@inheritDoc} */
+    @Override protected int getItemCount() {
```
Maran23: `@Override` should be above the method

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/TableCellBehaviorBase.java
```diff
@@ -101,16 +100,16 @@ protected void doSelect(final double x, final double y, final MouseButton button
...
+        if (tableCell.getIndex() >= count) return;
+
         // If the mouse event is not contained within this tableCell, then
         // we don't want to react to it.
         if (! tableCell.contains(x, y)) return;
 
         final Control tableView = getCellContainer();
```
Maran23: I think this should be the very first check here, like other cell classes do. Otherwise `getItemCount` can produce a `NullPointer`.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/CellBehaviorBase.java
```diff
@@ -206,9 +206,12 @@ protected void doSelect(final double x, final double y, final MouseButton button
 
         final Control cellContainer = getCellContainer();
```
Maran23: Again, we should null check the `cellContainer` first. Like my other comment suggested. 
You should move `final T cell = getNode();` after the null check then.

Maran23 (changes requested):

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/TreeTableRowBehavior.java
plumstone: I agree, but this introduces a different code style for this class. I would postpone this to a ticket that cleans up formatting issues.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/CellBehaviorBase.java
plumstone: unsure, but the instanceof check also covered this.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/TreeTableRowBehavior.java
Maran23: Sure. But it is also fine if you 'break' it, as we did that several times in the past.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/CellBehaviorBase.java
Maran23: I would remove it, because `getNode()` is called often without any null check, so it does not makes sense that it is only done here.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/ListCellBehavior.java
```diff
@@ -50,6 +50,14 @@ public ListCellBehavior(ListCell<T> control) {
      *                                                                         *
      **************************************************************************/
 
+    /** {@inheritDoc} */
+    @Override protected int getItemCount() {
```
andy-goryachev-oracle, line 54: please place annotations on it's own line in the new code

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/TableRowBehavior.java
```diff
@@ -52,6 +52,14 @@ public TableRowBehavior(TableRow<T> control) {
      *                                                                         *
      **************************************************************************/
 
+    /** {@inheritDoc} */
+    @Override protected int getItemCount() {
```
andy-goryachev-oracle, line 56: same here

andy-goryachev-oracle (commented):

some preliminary comments, this PR also needs `/template append` and ai policy acceptance.

andy-goryachev-oracle:
please merge the latest master branch.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/CellBehaviorBase.java
```diff
@@ -201,14 +201,22 @@ public void mouseDragged(MouseEvent e) {
...
+        if (cellContainer == null) {
+            return;
+        }
+
         // we update the cell to point to the new tree node
         final T cell = getNode();
 
-        final Control cellContainer = getCellContainer();
+        int count = getItemCount();
+        if (cell.getIndex() >= count) {
```
andy-goryachev-oracle, line 213: is it possible to get a negative index at this point?
should we check for `index < 0` as well?

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/TableRowBehaviorBase.java
```diff
@@ -82,6 +82,11 @@ public TableRowBehaviorBase(T control) {
         final Control table = getCellContainer();
         if (table == null) return;
 
+        int count = getItemCount();
+        if (getNode().getIndex() >= count) {
```
andy-goryachev-oracle, line 86: same question about negative index

andy-goryachev-oracle (commented):
my testing on macOS testing looks good - it appears only `TreeTableView` exhibit this issue.

I think this fix must be tested on windows due to differences in the popup trigger.

@lukostyra could you be the second reviewer please?

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/CellBehaviorBase.java
Maran23: Should not be possible. Cells with `-1` index should never be visible, never even be rendered. 
`index > itemCount,` on the other hand, is possible, because if you have, let's say 5 items and enough space to show 10, then you have 5 empty cells where the index is higher than the item count.
And you can click them, e.g. if you register a mouse `EventHandler`, it will run.

With that said, a check for `-1` probably wouldn't hurt, but I don't think it's really necessary.

Maran23 (approved):
Looks good!

andy-goryachev-oracle:
@Maran23 what platform did you test?  windows?

Maran23:
> @Maran23 what platform did you test? windows?

Arch Linux. But also tried it some months ago already on Windows and it was fixed there too back then.

Maran23:
Tested on Windows 11 today, also fine.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/TableCellBehaviorBase.java
andy-goryachev-oracle: here we need to check whether the cell is empty, because getItemCount() does not check for null.

here is the reproducer, click on an empty row:

```java
package goryachev.bugs;

import javafx.beans.property.SimpleStringProperty;
import javafx.scene.Scene;
import javafx.scene.control.TableCell;
import javafx.stage.Stage;

// NPE scenario for PR2129
public class TableView_NPE_Test extends Application {

    private TableView<String> table;

    private final ObservableList<String> data = FXCollections.observableArrayList("A");

    @Override
        table = new TableView<>(data);
        table.setColumnResizePolicy(TableView.CONSTRAINED_RESIZE_POLICY_SUBSEQUENT_COLUMNS);

        TableColumn<String, String> col = new TableColumn<>("Column 1");
        col.setCellValueFactory((f) -> {
            return new SimpleStringProperty(f.getValue());
        });
        col.setCellFactory((tp) -> {
            TableCell<String,String> cell = new TableCell<>();
            cell.addEventFilter(MouseEvent.MOUSE_PRESSED, (ev) -> {
                table.setItems(null);
            });
            return cell;
        });
        table.getColumns().add(col);

        stage.setScene(new Scene(table, 500, 500));
        stage.show();
    }
}
```

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/CellBehaviorBase.java
```diff
@@ -201,14 +201,22 @@ public void mouseDragged(MouseEvent e) {
...
+            return;
+        }
 
         // If the mouse event is not contained within this TreeCell, then
         // we don't want to react to it.
-        if (cell.isEmpty() || ! cell.contains(x, y)) {
+        if (!cell.contains(x, y)) {
```
andy-goryachev-oracle, line 219: we need to keep `cell.isEmpty()` check here because the model size can change before an existing empty cell is updated (similarly to the cell filter reproducer I've provided earlier)

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/TableRowBehaviorBase.java
```diff
@@ -82,6 +82,11 @@ public TableRowBehaviorBase(T control) {
         final Control table = getCellContainer();
         if (table == null) return;
 
+        int count = getItemCount();
```
andy-goryachev-oracle, line 85: similarly to the `CellBehaviorBase` case, we'll need to check for `row.isEmpty()` before invoking `getItemCount()`.

andy-goryachev-oracle (changes requested):
some regression found, see inline.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/CellBehaviorBase.java
Maran23: I'm confused. Clicking an empty cell is allowed, I can't see why it should not.

Maran23: Instead, `getItemCount()` should not throw.

PR #2139, opened 2026-04-06
8354943: [Linux] Simplify and update glass gtk backend: window sizing, positioning, and state management issues

tsayao:
This is a continuation to [JDK-8236651](https://bugs.openjdk.org/browse/JDK-8236651) and it aims to stabilize the linux glass gtk backend.

It refactors the Glass GTK implementation with a primary focus on window sizing, positioning, and state management, addressing a number of long-standing issues.

Previously, three separate context classes existed, two of which were used for Java Web Start and Applets. These have been unified, as they are no longer required.

Additional tests have been introduced to improve coverage. Some tests produced different results depending on the StageStyle, so they have been converted to use `@ParameterizedTest` to exercise multiple styles.

Although the primary focus is XWayland, the changes have also been verified to work correctly on Xorg.

This replaces #1789. It removes the use of GdkWindow in favor of GtkWindow, reducing risk and simplifying the review process while preserving the same set of bug fixes. Additionally, #2025 requires a `GtkWindow` to be used when setting the parent of the file chooser dialog.

To show debug messages,  build with `-PCONF=DebugNative`  and run with `-Djdk.gtk.verbose=true`. Log categories can be passed with `-Dglass.gtk.logCategories=CATEGORY` 

`CATEGORY` can be one or more of:
- all
- size
- position
- focus
- state
- lifecycle
- input
- dialog

Multiple categories can be specified by separating them with commas (e.g. size,focus,input).

A manual test is provided:
`java @build/run.args tests/manual/stage/TestStage.java`

When a window property is set, it is reported immediately. However, once it reaches the native Glass layer, it may be adjusted or rejected, causing the property to be updated again. Introducing a delay helps ensure the final state has been applied before it is verified.

---------

---------
### Progress

### Issues
 * [JDK-8354943](https://bugs.openjdk.org/browse/JDK-8354943): [Linux] Simplify and update glass gtk backend: window sizing, positioning, and state management issues (**Enhancement** - P4)
 * [JDK-8316425](https://bugs.openjdk.org/browse/JDK-8316425): [Linux] Stage.setMaximized() before show() does not persist (**Bug** - P4)
 * [JDK-8316891](https://bugs.openjdk.org/browse/JDK-8316891): [Linux] Intermittent test failure in IconifyTest.canIconifyDecoratedStage (**Bug** - P4)
 * [JDK-8337400](https://bugs.openjdk.org/browse/JDK-8337400): [Linux] Initial window position is not centered on Ubuntu 24.04 / Xorg (**Bug** - P4)
 * [JDK-8353612](https://bugs.openjdk.org/browse/JDK-8353612): [Linux] Some of the SizeToSceneTest fail in Ubuntu 24.04 (**Bug** - P4)
 * [JDK-8355073](https://bugs.openjdk.org/browse/JDK-8355073): [Linux] View size glitch when resizing past max window size (**Bug** - P4)
 * [JDK-8353556](https://bugs.openjdk.org/browse/JDK-8353556): RestoreSceneSizeTest fails in Ubuntu 24.04 (**Bug** - P4)
 * [JDK-8321624](https://bugs.openjdk.org/browse/JDK-8321624): DualWindowTest fails intermittently on Linux (**Bug** - P4)
 * [JDK-8367893](https://bugs.openjdk.org/browse/JDK-8367893): StageFocusTest fails on some Linux systems (**Bug** - P4)
 * [JDK-8366009](https://bugs.openjdk.org/browse/JDK-8366009): Stage does not always enter full screen when shown with fullScreen set to true on Ubuntu 24.04 (**Bug** - P4)
 * [JDK-8365201](https://bugs.openjdk.org/browse/JDK-8365201): [Linux] Restoring the maximized state of a window makes it exceed the screen's bounds (**Bug** - P4)

### Reviewers
 * [Michael Strauß](https://openjdk.org/census#mstrauss) (@mstr2 - **Reviewer**) 🔄 Re-review required (review applies to [dfdc22af](https://git.openjdk.org/jfx/pull/2139/files/dfdc22af17c4b0ac595c0fa90b51c28738d50883))

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2139/head:pull/2139` \
`$ git checkout pull/2139`

`$ git checkout pull/2139` \
`$ git pull https://git.openjdk.org/jfx.git pull/2139/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2139`

`$ git pr show -t 2139`

</details>

<a href="https://git.openjdk.org/jfx/pull/2139.diff">https://git.openjdk.org/jfx/pull/2139.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2139#issuecomment-4194124835)
</details>


kevinrushforth:
We'll need to evaluate this approach vs that of #1789 and choose one or the other.

It is accurate to say that the main difference between the two approaches is that PR #1789 replaces GtkWindow with GdkWindow whereas this PR sticks with GtkWindow? What other differences will help guide the reviewers?

Does this PR address the same set of issues that PR #1789 does? That one lists 10 additional bugs as also being resolved by that PR.

/reviewers 2 reviewers

tsayao:
> We'll need to evaluate this approach vs that of #1789 and choose one or the other.
>
> It is accurate to say that the main difference between the two approaches is that PR #1789 replaces GtkWindow with GdkWindow whereas this PR sticks with GtkWindow? What other differences will help guide the reviewers?

With the introduction of #2025, a `GtkWindow` became required for the FileChooser. As a result, PR #1789 needed a relatively large change to work directly with the portal (with a fallback to the GTK FileChooser), enabling the use of a `GdkWindow` as the parent.

By reverting to `GtkWindow` instead of `GdkWindow`, this additional FileChooser-related change is no longer necessary.

Experience shows that such changes can introduce regressions. Since this PR preserves the use of `GtkWindow` and changes less, it is likely lower risk compared to #1789. A final conclusion depends on further testing of this PR.
 
> Does this PR address the same set of issues that PR #1789 does? That one lists 10 additional bugs as also being resolved by that PR.
> 

At this time, the answer is likely yes, as all tests are passing on Ubuntu 24.04 with XWayland. I will follow up with additional validation and explicitly list the confirmed fixes.

The decision between #1789 and this PR will ultimately depend on the outcome of these tests.

> /reviewers 2 reviewers

tsayao:
/issue add 8354943,8316425,8316891,8337400,8353612,8355073,8353556,8321624,8367893,8366009

palexdev:
@tsayao should you add [JDK-8365201](https://bugs.openjdk.org/browse/JDK-8365201?jql=summary%20~%20%22maximize*%22%20OR%20description%20~%20%22maximize*%22%20ORDER%20BY%20created%20DESC%2C%20lastViewed%20DESC) to the list too? 

Edit: oh sorry, I just read that this is an alternative to the other proposal. So the question is, does this fixes the above-mentioned issue too?

tsayao:
> @tsayao should you add [JDK-8365201](https://bugs.openjdk.org/browse/JDK-8365201?jql=summary%20~%20%22maximize*%22%20OR%20description%20~%20%22maximize*%22%20ORDER%20BY%20created%20DESC%2C%20lastViewed%20DESC) to the list too?
> 
> Edit: oh sorry, I just read that this is an alternative to the other proposal. So the question is, does this fixes the above-mentioned issue too?

I still need to set up a multi-monitor environment to properly test it, but it will probably fix this issue as well.

kevinrushforth:
Reviewers: @lukostyra and some combination of: @mstr2 @beldenfox @kevinrushforth

modules/javafx.graphics/src/main/native-glass/gtk/glass_general.h
```diff
@@ -250,7 +289,13 @@ struct jni_exception: public std::exception {
 #define ERROR2(msg, param1, param2) {fprintf(stderr, msg, param1, param2);fflush(stderr);}
 #define ERROR3(msg, param1, param2, param3) {fprintf(stderr, msg, param1, param2, param3);fflush(stderr);}
 #define ERROR4(msg, param1, param2, param3, param4) {fprintf(stderr, msg, param1, param2, param3, param4);fflush(stderr);}
+
 #else
+
+#define LOG(...) ((void)0)
```
lukostyra: Minor: while it probably won't matter much right now, a preferred syntax here is:

```
#define LOG(...) do {} while(0)
```

It compiles to the same thing (a noop) but it is the most language-conforming statement that can be used basically anywhere without triggering a sudden compiler error. Here there could be a problem when for example we would try to do:
```
if (cond)
    LOG("success");
else
    LOG("error");
```

As a side-note, this also should apply to other LOG macros in this file, but we already have [JDK-8356327](https://bugs.openjdk.org/browse/JDK-8356327) for this, so you can leave them and it will get adjusted there.

lukostyra (commented):
I have reviewed a part of this change but I noticed there's still some commits being added.

Let us know when the change is ready to review. In the meantime, I'm leaving off the comment I found.

tsayao:
> I have reviewed a part of this change but I noticed there's still some commits being added.
> 
> Let us know when the change is ready to review. In the meantime, I'm leaving off the comment I found.

Since it started using GtkWindow again, the behavior is slightly different in some situations, so I'm rechecking the possible scenarios—that’s why these commits.

I also added comments in some places because I no longer remembered why they were there (they're usually edge cases).

modules/javafx.graphics/src/main/native-glass/gtk/glass_general.h
tsayao: Done

tsayao:
I’m finishing up the final details and expect to submit the last commit soon, aside from review changes.

tsayao:
The pull request is almost ready. I'm just dealing with a few remaining failing tests.

lukostyra:
> The pull request is almost ready. I'm just dealing with a few remaining failing tests.

Not a problem.

If you want you can turn this PR into a Draft and then once you're done remove the Draft status, that way we'll know what's the situation with this PR and Skara should be a bit more relaxed when it comes to timing it out.

tsayao:
All tests pass on Xorg and XWayland (primarily on Ubuntu 24.04). The new tests were also validated on macOS Tahoe. Additionally, extensive manual testing was conducted.

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
```diff
@@ -580,312 +721,255 @@ bool WindowContextBase::set_view(jobject view) {
...
+                    sm_mouse_drag_window->get_gdk_window(), cursor, false);
+        } else if (sm_grab_window) {
             glass_gdk_mouse_devices_grab_with_cursor(
-                    WindowContextBase::sm_grab_window->get_gdk_window(), cursor, TRUE);
+                    sm_grab_window->get_gdk_window(), cursor, true);
         }
     }
 
+    if (gdk_cursor) {
+        g_object_unref(gdk_cursor);
```
mstr2: The `gdk_cursor` field is unref'd here and in the destructor of `WindowContext`, but `_setCustomCursor()` passes in a cursor owned by `com.sun.glass.ui.Cursor` on the Java side. This may potentially destroy a cursor that is still used on the Java side. If `WindowContext` needs to participate in refcounting, it should probably take its own `g_object_ref()` for incoming cursors.

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
```diff
@@ -899,135 +983,217 @@ bool WindowContextTop::get_frame_extents_property(int *top, int *left,
...
-    geometry.view_y = origin_y - root_y;
-    notify_window_move();
+     // See the comment on move_resize about only trusting it when mapped
+     if (mapped) {
+        view_size.set({event->width, event->height});
+        window_location.set({x, y});
+     }
 
-    glong to_screen = getScreenPtrForLocation(geometry.x, geometry.y);
+    glong to_screen = getScreenPtrForLocation(event->x, event->y);
```
mstr2, line 1185: `getScreenPtrForLocation()`, which calls `gdk_screen_get_monitor_at_point()`, expects coordinates "in the virtual screen". You pass in coordinates specified "relative to its parent", can you verify that this is intended?

modules/javafx.graphics/src/main/native-glass/gtk/glass_general.h
```diff
@@ -44,6 +44,10 @@
 #define GDK_FILTERED_EVENTS_MASK static_cast<GdkEventMask>(GDK_ALL_EVENTS_MASK \
                 & ~GDK_TOUCH_MASK)
 
+#define ALPHA_CHANNEL_ERROR_MSG \
+    "Can't create transparent stage, because your screen doesn't support alpha channel. " \
+    "You need to enable XComposite extension.\n"
```
mstr2: If you are touching this code, maybe you should consider improving the language. In particular, readers shouldn't be colloquially addressed with "you". Maybe something like:
"Cannot create a transparent stage because the screen does not support an alpha channel without enabling the XComposite extension."

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.h
```diff
@@ -25,20 +25,171 @@
...
+        if (value != newValue) {
+            value = newValue;
+            invalidate();
+        }
+    }
+
+    void invalidate() {
+        if (!onChange) return;
+        if (notifying) {
+            fprintf(stderr,"WARNING: Re-entrant call to Observable::invalidate() (possible cyclic dependency)\n");
```
mstr2, line 73: Should this use the `LOG` macro, or alternatively be hidden behind the `VERBOSE` flag?

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
```diff
@@ -1306,20 +1555,188 @@ bool WindowContextTop::on_top_inherited() {
...
+
+    if (!mapped) {
+        // If the window is not yet mapped, report the values back to Java so it can
+        // reapply/correct them after mapping. We cannot rely on them staying the same,
+        // as the window manager or GTK may modify them during configure event processing.
+        LOG(SIZE, log_id, "move_resize: view_size: %d,%d\n", boundsW, boundsH);
+        view_size.set({boundsW, boundsH});
+    }
+
+    if (mapped) {
```
mstr2, line 1673: This looks a bit suspicious: `move_resize()` uses `mapped` to decide whether to call `gtk_window_resize()` or `gtk_window_set_default_size()`. The old code used `gtk_widget_get_realized()` and resized any realized window. Since the `mapped` field is always false for a `POPUP` window, the new code might not resize an already-realized popup.

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
tsayao: Looking at the GTK code, `gdk_window_set_cursor()` already unreferences the currently assigned cursor, so I removed the explicit unref. The custom cursor unref was causing a crash.

modules/javafx.graphics/src/main/native-glass/gtk/glass_general.h
tsayao: I updated the error message (though it's very unlikely to ever be shown).

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
tsayao: The parent of a toplevel or a popup (override-redirect) window on X11 is the root window which covers the entire screen. But I've found some problems while testing it and fixed it.

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.h
tsayao: I think this is closer to an error than a log message. On the other hand, it's something that should be caught during development.

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
tsayao: Popups go through the same process of setting `mapped` to `true`. However, I'll do some testing to verify popup move and resize behavior.

mstr2: Only `WindowContext::process_map()` seems to set `mapped = true`, but not if `window_type == POPUP`.

tsayao: You're right, I'll fix it. That said, I'd still like to run some additional tests, as this slipped through unnoticed (at least on my end). There isn't any evident issue in the most common popup use cases, such as menus, tooltips, and dropdowns, but it may still affect less common scenarios.

tsayao: I've fixed it and done some testing.

tsayao:
A test build including this PR is available:

[Download](https://github.com/tsayao/jfx/releases/download/8354943/javafx_27_8354943_test.tar.gz)

Testing and feedback would be greatly appreciated.

palexdev:
> A test build including this PR is available:
> 
> [Download](https://github.com/tsayao/jfx/releases/download/8354943/javafx_27_8354943_test.tar.gz)
> 
> Testing and feedback would be greatly appreciated.
> 
> 

I'll test it asap. 
How much time do we have before this is merged? 

Because I'd like to test one more thing. Sometimes unresizable dialogs are shown with incorrect sizes, therefore the content is cut and our users cannot perform any action nor read the text in it's entirety. 
Unfortunately this is very random, hard to reproduce

kevinrushforth:
> I'll test it asap. How much time do we have before this is merged?
> 
> Because I'd like to test one more thing. Sometimes unresizable dialogs are shown with incorrect sizes, therefore the content is cut and our users cannot perform any action nor read the text in it's entirety. Unfortunately this is very random, hard to reproduce

Given how large a change this is, and that it is only now settled down to the point of being ready for serious review / testing, it will take a fair bit of time. This is the kind of change that is better to go in early in the release cycle, not close to rampdown, so it will almost certainly not be for JavaFX 27; maybe for 28.

palexdev:
> A test build including this PR is available:
> 
> [Download](https://github.com/tsayao/jfx/releases/download/8354943/javafx_27_8354943_test.tar.gz)
> 
> Testing and feedback would be greatly appreciated.

Hello @tsayao,
sorry for the late reply, I finally found some time to build our app with your test build and here are the results.
- [FIXED] Opening popups for example from menu buttons would misplace them at seemingly random positions
- [BUG] [JDK-8365201](https://bugs.openjdk.org/browse/JDK-8365201) is not fixed, we still don't see the native window decorations when restoring a window maximized
- [BUG] This is very hard to reproduce, sometimes dialogs are shown cut. Inspecting them with ScenicView reveals that the sizes are correct. Not sure if this is related to this PR, or it's an issue with the CSS system

Here are the console logs:
<details> 
  <summary>Logs </summary>
   torakiki@torakiki-lenovo:~/Scaricati/pdfsam-basic-exp/bin$ ./pdfsam.sh
WARNING: java.lang.System::load has been called by com.sun.glass.utils.NativeLibLoader in module javafx.graphics (jrt:/javafx.graphics)

Available monitors: 2
Work Area: x:0, y:260, w:4480, h:1026
convert monitor[0] -> glass Screen
[x: 0 y: 233 w: 1920 h: 1080]
Work Area: x:0, y:260, w:4480, h:1026
convert monitor[1] -> glass Screen
[x: 1920 y: 0 w: 2560 h: 1440]
INFO 10:28:39.667 org.pdfsam.gui.PdfsamApp.init [JavaFX-Launcher] Starting...
INFO 10:28:40.269 org.sejda.core.context.DefaultSejdaConfiguration.<init> [JavaFX Application Thread] Configuring Sejda 6.0.0.M6
INFO 10:28:42.736 org.pdfsam.gui.PdfsamApp.start [JavaFX Application Thread] Avviato in 3 seconds
Work Area: x:0, y:260, w:4480, h:1026
convert monitor[1] -> glass Screen
[x: 1920 y: 0 w: 2560 h: 1440]
INFO 10:28:42.936 org.pdfsam.service.update.UpdatesController.ates-checker-thread] È disponibile PDFsam 6.0.5
Setup filters
Filters: 1
description[0]: (*.pdf)
Patterns: 2
pattern[0]: *.pdf
pattern[1]: *.PDF
Filter ready
FileChooser selected files: 1
Add [/home/torakiki/Scaricati/***.pdf] into returned filenames
dump: length = 1
dump: s[0]: /home/torakiki/Scaricati/***.pdf
INFO 10:29:01.190 org.pdfsam.service.pdf.DefaultPdfLoadServicedf loaded
Work Area: x:0, y:260, w:4480, h:1026
convert monitor[1] -> glass Screen
[x: 1920 y: 0 w: 2560 h: 1440]
Setup filters
Filters: 1
description[0]: (*.pdf)
Patterns: 2
pattern[0]: *.pdf
pattern[1]: *.PDF
Filter ready
FileChooser selected files: 1
Add [/home/torakiki/Scaricati/***.pdf] into returned filenames
dump: length = 1
dump: s[0]: /home/torakiki/Scaricati/***.pdf
INFO 10:29:14.897 org.pdfsam.service.pdf.DefaultPdfLoadServicedf loaded
Work Area: x:0, y:260, w:4480, h:1026
convert monitor[1] -> glass Screen
[x: 1920 y: 0 w: 2560 h: 1440]
Work Area: x:0, y:260, w:4480, h:1026
convert monitor[1] -> glass Screen
[x: 1920 y: 0 w: 2560 h: 1440]
INFO 10:29:26.359 org.pdfsam.gui.PdfsamApp.stop [JavaFX Applic...
</details>

Edit: src code for our app if you need to take a look, [here](https://github.com/torakiki/pdfsam)

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
```diff
@@ -1037,266 +1196,349 @@ void WindowContextTop::process_configure(GdkEventConfigure* event) {
...
-void WindowContextTop::update_window_constraints() {
-    bool is_floating = !is_iconified && !is_fullscreen && !is_maximized;
+void WindowContext::remove_window_constraints() {
+    if (window_type == POPUP) return;
+    if (!mapped) return;
+
+    LOG(SIZE, log_id, "remove_window_constraints\n");
+
+    gtk_window_set_geometry_hints(GTK_WINDOW(gtk_widget), nullptr, nullptr,
+        (GdkWindowHints) (GDK_HINT_MIN_SIZE | GDK_HINT_MAX_SIZE));
```
mstr2, line 1206: What does this function do? The [set_geometry_hints](https://docs.gtk.org/gtk3/method.Window.set_geometry_hints.html) documentation say that `geom_mask` is a "bitmask indicating which struct fields should be paid attention to", but `geometry` is null.

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
```diff
@@ -1037,266 +1196,349 @@ void WindowContextTop::process_configure(GdkEventConfigure* event) {
...
         hints.min_height = h;
         hints.max_width = w;
         hints.max_height = h;
+
+        LOG(SIZE, log_id, "update_window_constraints: unresizable: %d,%d", w, h);
     }
 
-    gtk_window_set_geometry_hints(GTK_WINDOW(gtk_widget), NULL, &hints,
-                                  (GdkWindowHints)(GDK_HINT_MIN_SIZE | GDK_HINT_MAX_SIZE));
+     gtk_window_set_geometry_hints(GTK_WINDOW(gtk_widget), nullptr, &hints, (GdkWindowHints) flags);
```
mstr2, line 1261: After queueing `gtk_window_resize()`, `move_resize()` calls this method. At that moment `view_size` still contains the old configured size, so the code reinstalls min == max == old size before GTK processes the queued resize.

Maybe the temporary removal should use a zero mask:
```c
gtk_window_set_geometry_hints(
    GTK_WINDOW(gtk_widget),
    nullptr,
    nullptr,
    static_cast<GdkWindowHints>(0));
```

The fixed constraints should then be reapplied after the corresponding `GDK_CONFIGURE` updates `view_size`, not immediately after `gtk_window_resize()`. Maybe a "pending programmatic resize" flag would make that sequencing explicit.

tests/system/src/test/java/test/javafx/stage/StageSizingTest.java
```diff
@@ -0,0 +1,521 @@
...
+            s.setResizable(false);
+        });
+
+        sleep(GEOMETRY_DELAY);
+
+        assertFalse(getStage().isResizable(), "Stage should be unresizable");
+
+        runAndWait(() -> {
+            getStage().setWidth(NEW_WIDTH);
+            getStage().setHeight(NEW_HEIGHT);
```
mstr2, line 511: `Window.setWidth()` and `Window.setHeight()` update the Java properties immediately, before native confirmation. If GTK rejects the resize without emitting a configure event, the Java property assertions can still report the requested values while the native window remains at the old size.

modules/javafx.graphics/src/main/native-glass/gtk/GlassApplication.cpp
```diff
@@ -542,7 +545,8 @@ static void process_events(GdkEvent* event, gpointer data)
                     process_dnd_target(ctx, &event->dnd);
                     break;
                 case GDK_MAP:
-                    // fall-through
+                    ctx->process_map();
+                    break;
```
mstr2, line 549: Is this really intended? The previous implementation passed `GDK_MAP` to `gtk_main_do_event()`. Other GTK-managed events in the new implementation still do so, but map events are now intercepted completely. This suppresses the widget’s `map-event` signal and the default `GtkWindow` map handler.

Maybe we should dispatch to GTK:
```c
case GDK_MAP:
    gtk_main_do_event(event);
    ctx->process_map();
    break;
```

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
```diff
@@ -1037,266 +1196,349 @@ void WindowContextTop::process_configure(GdkEventConfigure* event) {
...
-void WindowContextTop::set_enabled(bool enabled) {
-    is_disabled = !enabled;
+void WindowContext::set_enabled(bool enabled) {
+    LOG(FOCUS, log_id, "set_enabled: %s\n", enabled ? "true" : "false");
+    is_enabled = enabled;
+
+    // When not enabled, disable minimize
+    if (frame_type == TITLED && (initial_wmf & GDK_FUNC_MINIMIZE)) {
+        if (!enabled) {
+            remove_wmf(GDK_FUNC_MINIMIZE);
```
mstr2, line 1472: If we are here, it means that `initial_wmf` contains `GDK_FUNC_MINIMIZE`. However, both `remove_wmf()` and `add_wmf()` check `initial_wmf`, which makes this entire operation a no-op:
```cpp
if (initial_wmf & wmf) return;
```

Maybe we need a different approach here, for example:
```cpp
if (!enabled) {
    current_wmf = static_cast<GdkWMFunction>(
        static_cast<int>(current_wmf) & ~static_cast<int>(GDK_FUNC_MINIMIZE));
} else {
    current_wmf = static_cast<GdkWMFunction>(
        static_cast<int>(current_wmf) | static_cast<int>(initial_wmf & GDK_FUNC_MINIMIZE));
}

gdk_window_set_functions(gdk_window, current_wmf);
```

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
```diff
@@ -580,312 +723,251 @@ bool WindowContextBase::set_view(jobject view) {
...
-                }
-
-                set_bounds(x, y, true, true, w, h, cw, ch, 0, 0);
-           }
-        }
+    // Gravity x, y are used in centerOnScreen(). Here it's used to adjust the position
+    // accounting decorations, so calculate the difference
+    if (xSet && gravity_x > 0 && dx != 0) {
+        x -= gravity_x * static_cast<float>(dx);
+        if (x < 0) x = 0;
```
mstr2, line 940: Why is `x` (and `y`) clamped to zero? Couldn't a monitor to the left of the primary monitor have negative coordinates?

Maran23:
This PR should merge in master again.
I tested it, especially if it improves issues I found on Linux and for now, it looks very good.

It actually fixes a very annoying dialog bug for me, where the dialog is not correcty layouted for the very first time.

See here, only reproducable on Linux, fixed with your PR:

[dialog.webm](https://github.com/user-attachments/assets/dcf8a907-e7fe-4d83-8c11-dc9658876d94)

palexdev:
> This PR should merge in master again.
> I tested it, especially if it improves issues I found on Linux and for now, it looks very good.
> 
> It actually fixes a very annoying dialog bug for me, where the dialog is not correcty layouted for the very first time.
> 
> See here, only reproducable on Linux, fixed with your PR:
> 
> [dialog.webm](https://github.com/user-attachments/assets/dcf8a907-e7fe-4d83-8c11-dc9658876d94)

We're having this issue with the dialogs too! Except that we set them to not be resizable, therefore the user has to close it through the X and re-open it. 
Usually this fixes the issue, some have reported to not be enough 🤷🏻

Maran23:
> therefore the user has to close it through the X and re-open it

Yep! Same for me. Works for me with this PR though. Is it the same for you?

palexdev:
> > therefore the user has to close it through the X and re-open it
> 
> Yep! Same for me. Works for me with this PR though. Is it the same for you?
> 

We're not 100% sure because it's hard to reproduce




@torakiki you might be interested in this

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
tsayao: It removes the minimum and maximum window size constraints, allowing the window to be resized programmatically even when those constraints would otherwise prevent it. This matches the behavior provided on other platforms.

tests/system/src/test/java/test/javafx/stage/StageSizingTest.java
tsayao: I took this into consideration. The native side will notify the Java side if the resize cannot be applied, so the Java properties will be updated accordingly. Since this notification is asynchronous, there's a delay before checking the values.

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
mstr2: I'm just reading the documentation and it says that the flags indicate "which struct fields should be paid attention to"; however, there is no struct since you're passing in `nullptr`. This might be a case of unclear documentation, if the function works in the way how you're using it, then all is good.

tests/system/src/test/java/test/javafx/stage/StageSizingTest.java
tsayao: You’re right. The test does pass despite the issue mentioned in the other comment about using the old min/max values, because in this case the Java side is not notified. I think the solution would be to have a Robot test take a screenshot and measure the actual window size.

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
tsayao: Fixed it, but will only work on older window managers such as xfwm4. Modern mutter (gnome) or kwin (kde) will ignore the request.

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
tsayao: Yes, fixed it.

modules/javafx.graphics/src/main/native-glass/gtk/GlassApplication.cpp
tsayao: I don’t remember the reasoning behind this change. I've put it back and will run systemTests to verify any possible issue.

tests/system/src/test/java/test/javafx/stage/StageSizingTest.java
tsayao: I fixed the issue where programmatically resizing a non-resizable window was not working correctly. Since the previous implementation was re-adding the constraints with the old values, size changes were not being propagated to Java, which caused the test to pass even though the behavior was incorrect.

It is possible for the test to pass in scenarios where Java is not notified that the resize was rejected. One way to make the test more robust would be to take a snapshot of the Scene and verify its actual size, but I’m not sure if that would be a good approach for a test.

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
tsayao: The problem with waiting for a pending event is that it may never arrive, especially since there are different compositors/window managers, and some of them may not send the event.

I changed update_window_constraints to accept a size parameter, passing the same size that was provided to the resize operation. Since both the resize and the geometry hints are involved, it is possible that one succeeds while the other fails, but I think that is unlikely, and I couldn’t find a better solution.

I did the zero mask change.

mstr2 (approved):

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
tsayao: It seems to work, but I changed it to pass `0` for the hints mask as suggested. This seems more consistent with the intended behavior.

tsayao:
@palexdev About [JDK-8365201](https://bugs.openjdk.org/browse/JDK-8365201) would you retest with the latest changes?

If it doesn’t work, perhaps it would be better to remove it from the list of fixed bugs and handle it separately, to avoid making changes to this PR.

palexdev:
> @palexdev About [JDK-8365201](https://bugs.openjdk.org/browse/JDK-8365201) would you retest with the latest changes?
> 
> If it doesn’t work, perhaps it would be better to remove it from the list of fixed bugs and handle it separately, to avoid making changes to this PR.

@tsayao I'd really like to retest it, but unfortunately both @torakiki and I are on vacation this week, and we don't have a second monitor available to try to reproduce it.

If you can wait until the end of the week, I'll happily test it again.

tests/manual/stage/TestStage.java
```diff
@@ -0,0 +1,883 @@
+/*
+ * Copyright (c) 2025 Oracle and/or its affiliates. All rights reserved.
```
Maran23: Here and the other new files:
Copyright should be 2026.

and isn't the line: `Copyright (c) 2026, Oracle ...`, so with a comma?

tests/system/src/test/java/test/util/Util.java
```diff
@@ -40,26 +40,63 @@
...
     /** Default startup timeout value in seconds */
     public static final int STARTUP_TIMEOUT = 15;
     /** Test timeout value in milliseconds */
     public static final int TIMEOUT = 10000;
 
+    /**
+     * Time in milliseconds to wait for window geometry changes (resize, move)
+     * to be processed. On Linux, these operations are asynchronous.
+     * The native side may adjust values after Glass has applied them on the Java side.
+     * <p>
```
Maran23: Here and below, the empty `<p>` tag is not needed.

tests/system/src/test/java/test/util/Util.java
```diff
@@ -40,26 +40,63 @@
...
 import javafx.stage.Window;
 import org.junit.jupiter.api.Assertions;
 import com.sun.javafx.PlatformUtil;
+import javafx.beans.value.ChangeListener;
 
 /**
  * Utility methods for life-cycle testing
  */
 public class Util {
+    public static final String PARAMETERIZED_TEST_DISPLAY = "{displayName} [{index}] {arguments}";
```
Maran23, line 61: Is this needed?

tests/system/src/test/java/test/robot/javafx/stage/StageLocationTest.java
```diff
@@ -0,0 +1,186 @@
...
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+import static test.util.Util.FOCUS_DELAY;
+import static test.util.Util.GEOMETRY_DELAY;
+import static test.util.Util.PARAMETERIZED_TEST_DISPLAY;
+import static test.util.Util.TIMEOUT;
+
+@Timeout(value = TIMEOUT, unit= TimeUnit.MILLISECONDS)
```
Maran23: minor: missing space right after unit
```suggestion
@Timeout(value = TIMEOUT, unit = TimeUnit.MILLISECONDS)
```

modules/javafx.graphics/src/main/native-glass/gtk/GlassView.cpp
```diff
@@ -269,6 +269,7 @@ JNIEXPORT void JNICALL Java_com_sun_glass_ui_gtk_GtkView__1uploadPixelsByteArray
     }
 }
 
+
```
Maran23: empty newline

modules/javafx.graphics/src/main/native-glass/gtk/GlassApplication.cpp
```diff
@@ -133,6 +133,10 @@ JNIEXPORT void JNICALL Java_com_sun_glass_ui_gtk_GtkApplication__1initGTK
     gdk_threads_enter();
     gtk_init(NULL, NULL);
 
+    const char* categories = env->GetStringUTFChars(logCategories, nullptr);
+    glass_gtk_log_init(categories);
+    env->ReleaseStringUTFChars(logCategories, categories);
```
Maran23: as `categories` could be null if I understand correctly, should the `ReleaseStringUTFChars` be called with `null`?

modules/javafx.graphics/src/main/java/com/sun/glass/ui/gtk/GtkWindow.java
```diff
@@ -130,15 +130,13 @@ protected boolean _setVisible(long ptr, boolean visible) {
     @Override
     protected boolean _minimize(long ptr, boolean minimize) {
         minimizeImpl(ptr, minimize);
-        notifyStateChanged(WindowEvent.MINIMIZE);
```
Maran23, line 133: `notifyStateChanged` is not called anymore, is it? Also saved as `jGtkWindowNotifyStateChanged` but not called anymore as far as I can see.

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
```diff
@@ -32,113 +32,297 @@
...
+            window_type(type),
+            initial_wmf(wmf),
+            current_wmf(wmf) {
+    jwindow = mainEnv->NewGlobalRef(_jwindow);
+
+    log_id = "";
+    LOG(LIFECYCLE, log_id, "WindowContext: created\n");
+
+    if (frame_type != TITLED) {
+        initial_wmf = GDK_FUNC_ALL;
```
Maran23: Should `current_wmf` be initialized here as well?

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
```diff
@@ -1306,20 +1549,188 @@ bool WindowContextTop::on_top_inherited() {
...
+    auto loc = window_location.get();
+
+    int newX = 0, newY = 0;
+
+    bool loc_set = false;
+    if ((xSet || loc.x.has_value()) && (ySet || loc.y.has_value())) {
+        newX = xSet ? x : loc.x.value();
+        newY = ySet ? y : loc.y.value();
+        loc_set = true;
+    }
```
Maran23, line 1605: What about the case when only `x` or only `y` changed?

Maran23 (commented):
some comments, will do some more testing in the coming days. But as I already commented, it fixes some problems where a dialog is not correctly sized (and not just a bit, completely unusable initially), so definitely a big improvement.

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
tsayao: It does handle that case. When only x changes, newX uses the new x value while newY uses the existing loc.y value (and vice versa for y).

modules/javafx.graphics/src/main/native-glass/gtk/glass_window.cpp
tsayao: Fixed

modules/javafx.graphics/src/main/java/com/sun/glass/ui/gtk/GtkWindow.java
tsayao: Yes, you're right. I removed both notifyStateChanged and jGtkWindowNotifyStateChanged since they are no longer called.

modules/javafx.graphics/src/main/native-glass/gtk/GlassApplication.cpp
tsayao: Added the null check

modules/javafx.graphics/src/main/native-glass/gtk/GlassView.cpp
tsayao: Fixed

tests/system/src/test/java/test/robot/javafx/stage/StageLocationTest.java
tsayao: Fixed

tests/system/src/test/java/test/util/Util.java
tsayao: Yes, it's the format to show on the test report.

tests/manual/stage/TestStage.java
tsayao: Fixed

tests/system/src/test/java/test/util/Util.java
tsayao: Removed the tag

palexdev:
https://github.com/user-attachments/assets/7dfa7bd8-b671-418d-9d3a-94bc8817a31a

@tsayao tested, I did not merge main into this, let me know if I was supposed to
Dialogs are fixed as @Maran23 was also reporting. We are still seeing issues with drop downs (from video) on a secondary display, maybe it's something related to the scaling?
And [JDK-8365201](https://bugs.openjdk.org/browse/JDK-8365201) is also still present

PR #2145, opened 2026-04-12
8301283: Util methods for computing text/width height giving up some performance

Maran23:
This one took me actually some hours to fully understand.
I recommend reading the ticket description made by Phil as it already explains most of it.

What is the problem (in `Utils`):
- `Utils.computeTextWidth(..)` does not initialize the `boundsType`. As a consequence, we do not cache the layout result
  - For the text width, it makes sense to set it, so the result is cached. 
  Note that the `boundsType` does not matter for the width calculation itself, but we really want it set so the result will be cached if possible
- At one point due to `Utils.computeTextHeight(..)` method call, the `boundsType` is initialized. Now calls to `Utils.computeTextWidth(..)` do cache the result
  - So in order to fix that, `Utils.computeTextWidth(..)` does now set the the property as well

Ok, but why only cache this specific usecase (Phil was also wondering about that) (in `PrismTextLayout`):
- First of all, the previous method name was confusing (me). 
What it actually does: 
  - decide if the layout result should be cached 
  - when we have a cache hit, if we can reuse it completely or ONLY the text runs
- So what `PrismTextLayout` does is to cache the layout result for simple cases with the default setting. 
  - This is biased towards modena, which specifies `-fx-bounds-type: logical_vertical_center;` for ever `Text` node
    - If one of those setting or the bounds type do not match, we will not cache the text layout or reuse it completely, but the text runs instead
- That means the logic is fine and there's a thought behind it: Only cache if the settings match those we are most likely to receive from the application
  - In case we have already a cache result but for example the `wrapWidth` or the `boundsType` is different, we can reuse the runs but not the whole cached layout result
  - So we should not cache results with `boundsType = 0`, as those will have another height than with `boundsType = BOUNDS_CENTER`
    - So the logic is unchanged there

As `StubTextLayout` extends `PrismTextLayout`, we can actually write meaningful tests:
- So I wrote a bunch that make clear how the cache works and when a `GlyphLayout` is created (due to a complete cache miss) 
- Note that the tests are green before and after this PR
- Because `PrismTextLayout.setContent(..)` will request the font strike (for `hashCode` calculation for the layout result cache), I needed to implement `hashCode` and `equals` for some of the stub test classes (the JavaFX font classes do that as well)
- I renamed the `StubFontContractTest` to a better name since we do test a bit more now
  - And back in 2024 when I wrote this test class, I forgot the copyright header 🙃. Added now

Can we improve more?
- The ticket suggests that we could cache more in `LabeledSkinBase`. There are also some TODOs in this class. This could be a potential follow-up.

---------

---------
### Progress

### Issue
 * [JDK-8301283](https://bugs.openjdk.org/browse/JDK-8301283): Util methods for computing text/width height giving up some performance (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2145/head:pull/2145` \
`$ git checkout pull/2145`

`$ git checkout pull/2145` \
`$ git pull https://git.openjdk.org/jfx.git pull/2145/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2145`

`$ git pr show -t 2145`

</details>

<a href="https://git.openjdk.org/jfx/pull/2145.diff">https://git.openjdk.org/jfx/pull/2145.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2145#issuecomment-4232220095)
</details>


kevinrushforth:

@prrace Would you also be able to review this? Or comment on the points raised in the PR description?

/reviewers 2

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TextLayoutUtilsContractTest.java
```diff
@@ -0,0 +1,366 @@
...
+/**
+ * This is a test for fonts, font loading, and their computed sizes.
+ * Note that while we only test the stub font loading here, we still want to verify some basic rules that apply
+ * for all headless tests. Some rules are somewhat derived from the real font loading in JavaFX.
+ * <p>
+ * See the Javadoc for every method to get more details.
+ * @see test.com.sun.javafx.pgstub.StubFontLoader
+ * @see test.com.sun.javafx.pgstub.StubTextLayout
+ */
+class TextLayoutUtilsContractTest {
```
andy-goryachev-oracle, line 54: Renaming this test was a good idea.
I might suggest adding a comment explaining that since it's a headless test, we are dealing with the StubToolkit and its simplified text layout.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/skin/Utils.java
```diff
@@ -167,6 +168,8 @@ public static double computeTextWidth(Font font, String text, double wrappingWid
         try {
             layout.setContent(text != null ? text : "", FontHelper.getNativeFont(font));
             layout.setWrapWidth((float)wrappingWidth);
+            layout.setLineSpacing(0);
+            layout.setBoundsType(TextLayout.BOUNDS_CENTER);
```
andy-goryachev-oracle, line 172: this change makes sense, but perhaps we should delegate to 
`computeTextHeight(Font font, String text, double wrappingWidth, double lineSpacing, TextBoundsType boundsType)` to eliminate parallel paths?

just like we do in `computeTextHeight(Font font, String text, double wrappingWidth, TextBoundsType boundsType)` just below it?

that is,

```
public static double computeTextWidth(Font font, String text, double wrappingWidth) {
  return computeTextHeight(font, text, wrappingWidth, 0, TextLayout.BOUNDS_CENTER);
}
```

andy-goryachev-oracle (commented):
No ill effects as far as i can see with the Monkey Tester on macOS 26.3.1

I do see build errors in Eclipse:
```
GlyphLayout cannot be resolved to a type	TextLayoutUtilsContractTest.java	Java Problem	/controls/src/test/java/test/javafx/scene/control/skin	line 245
GlyphLayout cannot be resolved to a type	TextLayoutUtilsContractTest.java	Java Problem	/controls/src/test/java/test/javafx/scene/control/skin	line 277
GlyphLayout cannot be resolved to a type	TextLayoutUtilsContractTest.java	Java Problem	/controls/src/test/java/test/javafx/scene/control/skin	line 310
GlyphLayout cannot be resolved to a type	TextLayoutUtilsContractTest.java	Java Problem	/controls/src/test/java/test/javafx/scene/control/skin	line 342
The type com.sun.javafx.text.GlyphLayout is not accessible	TextLayoutUtilsContractTest.java	Java Problem	/controls/src/test/java/test/javafx/scene/control/skin	line 30
```

(I've merged the latest master branch before testing)

andy-goryachev-oracle:
Can you add this change to fix the Eclipse warnings:

https://github.com/andy-goryachev-oracle/jfx/blob/ag.2145/modules/javafx.controls/.classpath

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/skin/Utils.java
Maran23: But then we would return the height? Which is wrong, as it is not the width?

andy-goryachev-oracle: oh right.  we can't really use the same code path, we'd need to release the layout after we get the requested value from it.  We could pass a lambda, but it's not worth it.  Please disregard.

Maran23: No problem, I was just confused if I miss something. 😁

andy-goryachev-oracle: there is another way to ensure that there is only one place where we configure the layout, something like this, but it's probably not for this PR:

```
    public static double getLineHeight(Font font, TextBoundsType boundsType) {
        // do it everywhere
        TextLayout layout = configureLayout("", font, boundsType, 0, 0);
        try {
            // JDK-8093957: Use the line bounds specifically, to include font leading.
            return layout.getLines()[0].getBounds().getHeight();
        } finally {
            release(layout);
        }
    }

    private static TextLayout configureLayout(String content, Font font, TextBoundsType boundsType, wrapWidth, linespacing) {
        TextLayout layout = layout();
        layout.setContent(content, FontHelper.getNativeFont(font));
        layout.setWrapWidth(wrapWidth);
        layout.setLineSpacing(linespacing);
        if (boundsType == TextBoundsType.LOGICAL_VERTICAL_CENTER) {
                layout.setBoundsType(TextLayout.BOUNDS_CENTER);
        } else {
                layout.setBoundsType(0);
        }
        return layout;
    }
```

Maran23:
> Can you add this change to fix the Eclipse warnings:
> 
> https://github.com/andy-goryachev-oracle/jfx/blob/ag.2145/modules/javafx.controls/.classpath

I don't feel to confident doing something like that, since I can not test it. Will just copy your code for now, but can not do any reasoning about it or fixes if someone git blame me in the future

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TextLayoutUtilsContractTest.java
Maran23: done

andy-goryachev-oracle:
Thanks!
All it does is adding `<attribute name="add-exports" value="javafx.graphics/com.sun.javafx.text=javafx.controls"/>` on L14

andy-goryachev-oracle (approved):
The changes look good!

The only product change is adding missing `TextLayout` configuration items in `Utils`.

andy-goryachev-oracle:

mstr2 (approved):

Maran23:
I will leave this open for another 2 days in case Phil @prrace want to have a look.

Ziad-Mid (approved):
LGTM

kevinrushforth:
@Maran23 the newly added tests fail when using JDK 26 or later to run the tests. I filed [JDK-8385737](https://bugs.openjdk.org/browse/JDK-8385737) to track the regression, which will need to be fixed in jfx27.

PR #2146, opened 2026-04-12
8382049: [TestBug] Multiple StubToolkit checks in JavaFX production code

Maran23:
This PR removes all `StubToolkit` checks from JavaFX production code.
There are only a few.

Two notes:

Note 1: `TableRowSkinBase`s `DO_ANIMATIONS` is now `true` for tests. It was not before due to a leak (as far as I could find when digging [JDK-8120657](https://bugs.openjdk.org/browse/JDK-8120657)). But I could not find any problem, even running all tests with `-Xmx512m`
Checking tests that trigger the affected code path and creating a heapdump after, I could not see anything weird:
> Heapdump after `TreeTableViewResizeTest`:
image
image

Note 2: Now when the tests run in CI, `JavaFX: using test.com.sun.javafx.pgstub.StubToolkit` is now printed 12 times (probably the amount of Unit Test Threads):
image
If this is a problem, we can decide if we want to change the print code in `Toolkit` (in a follow-up). 
-> Changed to `System.out` in this PR.

---------

---------
### Progress

### Issue
 * [JDK-8382049](https://bugs.openjdk.org/browse/JDK-8382049): [TestBug] Multiple StubToolkit checks in JavaFX production code (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2146/head:pull/2146` \
`$ git checkout pull/2146`

`$ git checkout pull/2146` \
`$ git pull https://git.openjdk.org/jfx.git pull/2146/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2146`

`$ git pr show -t 2146`

</details>

<a href="https://git.openjdk.org/jfx/pull/2146.diff">https://git.openjdk.org/jfx/pull/2146.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2146#issuecomment-4232657281)
</details>

### DIFF
--- a/modules/javafx.base/src/main/java/com/sun/javafx/PlatformUtil.java
+++ b/modules/javafx.base/src/main/java/com/sun/javafx/PlatformUtil.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2025, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -287,5 +287,15 @@ public class PlatformUtil {
 
     public static boolean isAndroid() {
        return ANDROID;
+    }
+
+    /**
+     * Returns true if animations should be done.
+     *
+     * @return true if animations should be done, false otherwise
+     * @implNote To save CPU cycles, animations are not done for embedded
+     */
+    public static boolean isDoAnimations() {
+        return !isEmbedded();
     }
 }
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/Toolkit.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/Toolkit.java
@@ -249,7 +249,7 @@ public abstract class Toolkit {
             TOOLKIT = (Toolkit)clz.getDeclaredConstructor().newInstance();
             if (TOOLKIT.init()) {
                 if (printToolkit) {
-                    System.err.println("JavaFX: using " + forcedToolkit);
+                    System.out.println("JavaFX: using " + forcedToolkit);
                 }
                 return TOOLKIT;
             }
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/Toolkit.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/Toolkit.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2025, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -225,8 +225,7 @@ public abstract class Toolkit {
             forcedToolkit = lookupToolkitClass(forcedToolkit);
         }
 
-        boolean printToolkit = verbose
-                || (userSpecifiedToolkit && !forcedToolkit.endsWith("StubToolkit"));
+        boolean printToolkit = verbose || userSpecifiedToolkit;
 
         try {
             Class clz = null;
--- a/modules/javafx.graphics/src/test/java/test/com/sun/javafx/pgstub/StubToolkit.java
+++ b/modules/javafx.graphics/src/test/java/test/com/sun/javafx/pgstub/StubToolkit.java
@@ -781,7 +781,7 @@ public class StubToolkit extends Toolkit {
         return false;
     }
 
-    private KeyCode platformShortcutKey = null;
+    private KeyCode platformShortcutKey;
 
     public void setPlatformShortcutKey(final KeyCode platformShortcutKey) {
         this.platformShortcutKey = platformShortcutKey;
--- a/modules/javafx.graphics/src/test/java/test/com/sun/javafx/pgstub/StubToolkit.java
+++ b/modules/javafx.graphics/src/test/java/test/com/sun/javafx/pgstub/StubToolkit.java
@@ -35,6 +35,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.Future;
+
+import com.sun.javafx.util.Utils;
 import javafx.application.ConditionalFeature;
 import javafx.geometry.Dimension2D;
 import javafx.scene.image.Image;
@@ -779,7 +781,7 @@ public class StubToolkit extends Toolkit {
         return false;
     }
 
-    private KeyCode platformShortcutKey = KeyCode.SHORTCUT;
+    private KeyCode platformShortcutKey = null;
 
     public void setPlatformShortcutKey(final KeyCode platformShortcutKey) {
         this.platformShortcutKey = platformShortcutKey;
@@ -787,6 +789,10 @@ public class StubToolkit extends Toolkit {
 
     @Override
     public KeyCode getPlatformShortcutKey() {
+        if (platformShortcutKey == null) {
+            return Utils.isMac() ? KeyCode.META : KeyCode.CONTROL;
+        }
+
         return platformShortcutKey;
     }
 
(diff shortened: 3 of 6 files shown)


modules/javafx.graphics/src/test/java/test/com/sun/javafx/pgstub/StubToolkit.java
```diff
@@ -779,14 +781,18 @@ public boolean isNestedLoopRunning() {
         return false;
     }
 
-    private KeyCode platformShortcutKey = KeyCode.SHORTCUT;
+    private KeyCode platformShortcutKey = null;
```
Maran23: This can be overridden by tests (`KeyCombinationTest`), otherwise the default is used that was previously already done in `KeyBinding` and `KeyModifier`.

andy-goryachev-oracle: very minor: `= null` is unnecessary.

andy-goryachev-oracle:
Re: TableRowSkinBases and DO_ANIMATIONS - there were quite a few memory leak fixes implemented recently (er, 2024), so enabling animations is likely to be ok (especially if it does not fail any headful CI runs).

Tickets in question:
https://bugs.openjdk.org/browse/JDK-8341687
https://bugs.openjdk.org/browse/JDK-8320232
https://bugs.openjdk.org/browse/JDK-8323543
https://bugs.openjdk.org/browse/JDK-8307538

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/Toolkit.java
```diff
@@ -225,8 +225,7 @@ public static synchronized Toolkit getToolkit() {
             forcedToolkit = lookupToolkitClass(forcedToolkit);
         }
 
-        boolean printToolkit = verbose
-                || (userSpecifiedToolkit && !forcedToolkit.endsWith("StubToolkit"));
+        boolean printToolkit = verbose || userSpecifiedToolkit;
```
andy-goryachev-oracle, line 228: I might suggest to change L252 to use stdout instead:

`System.out.println("JavaFX: using " + forcedToolkit);`

(I don't know why this is needed in the first place, but ok)

Maran23:
> Re: TableRowSkinBases and DO_ANIMATIONS - there were quite a few memory leak fixes implemented recently (er, 2024), so enabling animations is likely to be ok (especially if it does not fail any headful CI runs).

I had the same thought. 
And if there is a problem only in the headless tests, we can still decide to disable the animations for headless tests with e.g. a System Property or similar. 
But since other Controls also use Animations (`Pagination`, `TitledPane`, `TabPane` (Tab close)) and no issue there, I think this is fine.

andy-goryachev-oracle (commented):
I am running CI builds, will update on the result.  the code changes look good.

modules/javafx.graphics/src/test/java/test/com/sun/javafx/pgstub/StubToolkit.java
Maran23: Ah yes, I changed this to make clear that it can be overridden, but indeed unnecessary (it was not that clear before). Will remove.

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/Toolkit.java
Maran23: I don't know either.

But is this something we should change in this PR?
I mean for me, this is fine!
I just don't know if there could be any consequences/If it needs some kind of discussion

andy-goryachev-oracle: It makes the test log cleaner by removing the word "ERROR":

```
HeaderButtonOverlayTest > maximizeButtonIsDisabledWhenStageIsNotResizable() STANDARD_ERROR
```

We still see the output.  And it **is** a related change ;-)

andy-goryachev-oracle:
Out CI runs are all green.

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/Toolkit.java
Maran23: Ok, convinced :)

Maran23: done

andy-goryachev-oracle (approved):
thank you! lgtm

andy-goryachev-oracle:
sorry for letting this PR auto-expire.  please re-open.

andy-goryachev-oracle:
@lukostyra could you be the second reviewer please?

modules/javafx.base/src/main/java/com/sun/javafx/PlatformUtil.java
```diff
@@ -288,4 +288,14 @@ private static void loadProperties() {
...
        return ANDROID;
     }
+
+    /**
+     * Returns true if animations should be done.
+     *
+     * @return true if animations should be done, false otherwise
+     * @implNote To save CPU cycles, animations are not done for embedded
+     */
+    public static boolean isDoAnimations() {
```
mstr2, line 298: I think we shouldn't hard-code behavior for different platforms, especially since the assumption that **all** embedded platforms are totally unable to show animations is flawed at best. There are powerful embedded systems.

We've had `Scene.Preferences.reducedMotion` for quite some time now, so we should make animations depend on that preference instead. For backwards compatibility, we could make `Platform.Preferences.reducedMotion` default to `true` on embedded platforms.

andy-goryachev-oracle: I think it should be ok for this PR (corresponds to the status quo), but you do bring a good point: `isEmbedded()` is extremely vague - is it referring to a 8085 kind of embedded or embedded linux on 2 GHz arm chip?  

And what does it mean, exactly?  There are a few places that use it currently, like sub-pixel (LCD) font rendering, some caching in `TitledPaneSkin`, texture mapping in Phong material, etc.

Perhaps instead we should consider splitting it into individual features like `low power`, `resource constrained`, `reduced graphic quality` and the like.

What do you think?

Maran23: Agree. I initially wanted to add a System Property instead - but decided against because I want to keep the behavior for now.

But I'm definitely in favor of a better method.

Regarding `Platform.Preferences.reducedMotion`, this seems like a good candidate here and also `Pagination` prev/next Animation - maybe the `TitledPane` `animated` property default and some more. 
+1 from me.

I think this would not solve all cases though, as Andy mentioned there are some other locations where `reducedMotion` makes not that much sense probably, but instead something like `reduced graphic quality` or similar.

How about I create a ticket, and the two of you add any details I haven't covered - so we have a well-documented ticket? Then I can send an email and see what kind of feedback we get.

mstr2: In the past, Web CSS had media types that tried to define platforms like `tv`, `projection`, `handheld`, etc. with the same idea: some platforms should be treated differently than other platforms. Those are now deprecated, because the idea has proven to be incoherent. If I have a laptop computer, is it `screen`? Does it become `handheld` when I pick it up? What if I connect it to a video projector, now it's suddenly `projection`? Why should my application now look and feel different?

When we tie behaviors to specific platforms, we go down the same incoherent and wrong path. An embedded platform can be extremely resource-constrained, but it can also be more powerful than a desktop platform. We introduced `reducedMotion`, `reducedTransparency`, and `reducedData` to allow _applications_ instead of _platforms_ to be in charge of defining behavior characteristics, and those settings directly describe what they control; there's no need to guess what it should mean for an application.

The API is there, we just need to start using it. I would also prefer that we start by using what we already have, and not try to come up with even more settings without having used any of the existing settings. I think this can be done as part of this PR because it already touches the code; if you don't want to do that, then with a follow-up PR.

Maran23: Since I labeled this as a `testbug`, I would like to keep it that way. However, I created https://bugs.openjdk.org/browse/JDK-8385459, ready to grab.

This includes the first iteration of changes. I would love to see the `animated` property default consider `reducedMotion` - so that `TitledPane` and all Charts will have their default `animated` property set to  `reducedMotion`.

This way, we will by default honor the user settings and we do not need to bind it on our own on every instance. But that should be the second part.

mstr2 (approved):

lukostyra (approved):
LGTM. Ran both local and CI testing and everything looks good.

PR #2147, opened 2026-04-13
8381443: SystemMenuBarHelpMenuTest fails on macOS

dmarkov20:
The root cause of the test failure is that the robot does not successfully click on the "About" menu item. The test relies on hardcoded cursor coordinates and the Y coordinate value (80) is insufficient on some macOS systems with higher resolutions.

Replace the mouse click on the "About" menu item with keyboard interactions. This removes the dependency on hardcoded cursor positions and makes the test more robust across systems with different resolutions and font sizes.

---------

---------
### Progress

### Issue
 * [JDK-8381443](https://bugs.openjdk.org/browse/JDK-8381443): SystemMenuBarHelpMenuTest fails on macOS (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2147/head:pull/2147` \
`$ git checkout pull/2147`

`$ git checkout pull/2147` \
`$ git pull https://git.openjdk.org/jfx.git pull/2147/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2147`

`$ git pr show -t 2147`

</details>

<a href="https://git.openjdk.org/jfx/pull/2147.diff">https://git.openjdk.org/jfx/pull/2147.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2147#issuecomment-4237146030)
</details>

### DIFF
--- a/tests/system/src/test/java/test/robot/javafx/scene/SystemMenuBarHelpMenuTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/scene/SystemMenuBarHelpMenuTest.java
@@ -75,10 +75,6 @@ public class SystemMenuBarHelpMenuTest {
     private static final int MENU_BAR_X = 120;
     private static final int MENU_BAR_Y = 10;
 
-    // Estimated y offset for the Help -> About menu item, after the Spotlight search field that macOS inserts to
-    // this menu.
-    private static final int MENU_ABOUT_Y = 80;
-
     private static Robot robot;
     private static final AtomicBoolean aboutItemFired = new AtomicBoolean(false);
     private static final CountDownLatch aboutItemLatch = new CountDownLatch(1);
--- a/tests/system/src/test/java/test/robot/javafx/scene/SystemMenuBarHelpMenuTest.java
+++ b/tests/system/src/test/java/test/robot/javafx/scene/SystemMenuBarHelpMenuTest.java
@@ -48,7 +48,6 @@ import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Assumptions;
 import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import test.util.Util;
 
@@ -98,7 +97,6 @@ public class SystemMenuBarHelpMenuTest {
      * Verifies that after opening, the Help menu includes the Spotlight search field, which is
      * included seamlessly by the system.
      */
-    @Disabled("JDK-8381443")
     @Test
     void testHelpMenuHasSpotlight() {
         Assumptions.assumeTrue(PlatformUtil.isMac(), "System menu bar tests only apply to macOS");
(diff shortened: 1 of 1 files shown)


andy-goryachev-oracle:
sorry, it's a test fix.  one reviewer is enough
/reviewers 1

tests/system/src/test/java/test/robot/javafx/scene/SystemMenuBarHelpMenuTest.java
```diff
@@ -116,8 +110,8 @@ void testHelpMenuHasSpotlight() {
 
         // Step 2: click on About
         Util.runAndWait(() -> {
-            robot.mouseMove(MENU_BAR_X, MENU_ABOUT_Y);
-            robot.mouseClick(MouseButton.PRIMARY);
+            robot.keyType(KeyCode.DOWN);
+            robot.keyType(KeyCode.ENTER);
```
andy-goryachev-oracle, line 114: This is reasonable.
I suppose the only scenario when the system menu is not focused is when the test must fail, which is exactly what we need.

andy-goryachev-oracle (approved):
tested on macOS 26.3.1 with two monitors.

jperedadnr:
There was a reason for the mouse approach: the Spotlight search menu is not a node, it is a pure decoration that macOS adds to the Help menu. 

By adding a key event, we are always sure that the first JavaFX menuItem is going to be selected (key events won't be affected by decorations, whether they are there or not), and the test will never fail. 

However, the initial intention of this test was precisely to assert that there was a spotlight search decoration added precisely to the "Help" menu. I agree that using a hardcoded Y offset was probably a bad idea (as it didn't take into account multiple resolutions). 

Could we find a better way?

kevinrushforth:
I wonder if taking the screen size into account would help? Either way, the test is likely to be fragile unless we can find a reliable way to calculate the Y value of the About menu.

andy-goryachev-oracle:
maybe we can populate the menu with █ symbols and then detect these blocks in screen capture...

dmarkov20:
It looks like there is no reliable way to calculate the Y position of the About menu item. I would therefore suggest keeping the proposed change since it improves the stability of the test even though it does not cover one of the original scenarios the test was meant to verify. We could cover the Spotlight case in the macOS system menu with a separate manual test instead. 
@jperedadnr What do you think?

jperedadnr:
Technically, we could use an osascript like in the manual test MacOSSystemenuTestBase (see `getMenuReaderProcess`) to find the About position. But then the whole test, which was added to verify that the spotlight field was inserted when a "Help" menu was added to the system menu bar, could be solved just with this script.

So this test should probably be indeed moved to a manual test, with the sole purpose of visually confirming that the Spotlight is added under a "Help" menu, and optionally, use the osascript to verify it. Keeping it as a system test with the changes from this PR will always pass, so I don't see that much value for the test.

dmarkov20:
> Technically, we could use an osascript like in the manual test MacOSSystemenuTestBase (see `getMenuReaderProcess`) to find the About position. But then the whole test, which was added to verify that the spotlight field was inserted when a "Help" menu was added to the system menu bar, could be solved just with this script.
> 
> So this test should probably be indeed moved to a manual test, with the sole purpose of visually confirming that the Spotlight is added under a "Help" menu, and optionally, use the osascript to verify it. Keeping it as a system test with the changes from this PR will always pass, so I don't see that much value for the test.

Well, the test file contains two tests that verify different aspects of the system menu bar behavior on macOS. Overall it still looks useful and I would prefer to keep it with the proposed changes unless there are objections.

As for verifying that the Spotlight field is added under the "Help" menu, I will file a new bug to introduce a separate manual test for that purpose.

dmarkov20:
I have filed a follow-up issue, https://bugs.openjdk.org/browse/JDK-8382531 to address the Spotlight-related case.

As far as I can see there have been no objections to my proposal to keep SystemMenuBarHelpMenuTest with the proposed modifications. Could you please take a moment to review the changes?

PR #2151, opened 2026-04-16
8237018: Unstable test RenderLock1Test always fails on Windows

lukostyra:
`RenderLock1Test`'s instability comes from Windows focus stealing mechanism getting in the way.

The test displays an alert and relies on whether the alert's Button receives focus while other window is rendering in the background. To successfully receive focus on the test Button the application itself must be granted focus first, which [in some cases can be rejected by the OS](https://bugs.openjdk.org/browse/JDK-8351357).

On macOS I noticed no issues - the test ran 100 times in a loop and succeeded every time, no matter if Gradle ran with or without daemon. Similarly I ran the test on our CI systems on repeat 100 times and it also succeeded there.

On Linux the test does not work because when the alert-displaying Stage is closed the "lost focus" event does not fire. This is a Linux specific issue and it [already is documented in-code with an assumption](https://github.com/openjdk/jfx/blob/master/tests/system/src/test/java/test/renderlock/RenderLockCommon.java#L111).

On Windows the focus problems come from aforementioned focus stealing prevention. The solution to this is similar to other tests that showed this problem, which is to run tests using `--no-daemon` flag in Gradle. I similarily tested this by running the test using `--no-daemon` flag on repeat 100 times and the test always passed.

I think we can mark this test as stable and include it into the testing pool. 

---------

---------
### Progress

### Issue
 * [JDK-8237018](https://bugs.openjdk.org/browse/JDK-8237018): Unstable test RenderLock1Test always fails on Windows (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2151/head:pull/2151` \
`$ git checkout pull/2151`

`$ git checkout pull/2151` \
`$ git pull https://git.openjdk.org/jfx.git pull/2151/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2151`

`$ git pr show -t 2151`

</details>

<a href="https://git.openjdk.org/jfx/pull/2151.diff">https://git.openjdk.org/jfx/pull/2151.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2151#issuecomment-4261283805)
</details>

### DIFF
--- a/tests/system/src/test/java/test/renderlock/RenderLock1Test.java
+++ b/tests/system/src/test/java/test/renderlock/RenderLock1Test.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2014, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -25,13 +25,11 @@
 
 package test.renderlock;
 
-import static org.junit.jupiter.api.Assumptions.assumeTrue;
 import org.junit.jupiter.api.Test;
 
 public class RenderLock1Test extends RenderLockCommon {
     @Test
     public void windowCloseTest() throws Exception {
-        assumeTrue(Boolean.getBoolean("unstable.test"));
         doWindowCloseTest();
     }
 }


Maran23:
Makes sense. Should we maybe add Javadoc above the class to mention that? Maybe something like `If this test fails on Windows, makes sure to try re-running the tests with '--no-daemon' flag in Gradle.`

lukostyra:
> Makes sense. Should we maybe add Javadoc above the class to mention that? Maybe something like `If this test fails on Windows, makes sure to try re-running the tests with '--no-daemon' flag in Gradle.`

I have created a test for that some time ago called `StageFocusTest`, it tries to detect if both a newly created Stage and a Stage provided by `Application.start()` actually receive focus when being shown and leaves a message about `--no-daemon` flag. I would expect it to fail when this test fails as well.

kevinrushforth (approved):
LGTM

I ran the test 100 times in a loop and it passed every time. Since this is a simple test fix to enable a test, a single reviewer is sufficient.

/reviewer 1

kevinrushforth:
> /reviewer 1

Let's try that again, spelling it correctly this time.

/reviewers 1

Maran23:
> I have created a test for that some time ago called `StageFocusTest`, it tries to detect if both a newly created Stage and a Stage provided by `Application.start()` actually receive focus when being shown and leaves a message about `--no-daemon` flag. I would expect it to fail when this test fails as well.

Yeah I saw that and borrowed the message from there. But makes sense, most of the people usually run all tests and then get the helpful message from there.

Maran23 (approved):

PR #2154, opened 2026-04-17
8237868: java.lang.IndexOutOfBoundsException in FilteredList

Maran23:
This fixes an IOOBE and even improves the performance / change aggregation.
I suggest reading Jeanette's comment in the ticket, that helped a lot to understand it.

There was a code path that broke change events. 
Jeanette wondered as well in the comment why this exists at all, if this was meant to be some kind of optimization?

What I found out is: Because of this code path, we actually got 3 change events and the order was wrong. 
What we got with the test from Jeanette: `subChanges: { [B] removed at 1, [D] removed at 2, [C, E, F] removed at 1 }`

Removing the code path and just calling `insertRemoved(..)` directly will correctly find an alreadx existing sub change with `findSubChange` and aggregate the changes together. So now, we only get one event: `subChanges: { [B, C, D, E, F] removed at 1 }`. No error and we use the smart aggregation from `insertRemoved`.

My speculation why this code existed is that there may was a former bug, similar like [JDK-8208758](https://bugs.openjdk.org/browse/JDK-8208758) (#2117) where I also did a fix in the `insertRemoved(..)` method.

---------

---------
### Progress

### Issue
 * [JDK-8237868](https://bugs.openjdk.org/browse/JDK-8237868): java.lang.IndexOutOfBoundsException in FilteredList (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2154/head:pull/2154` \
`$ git checkout pull/2154`

`$ git checkout pull/2154` \
`$ git pull https://git.openjdk.org/jfx.git pull/2154/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2154`

`$ git pr show -t 2154`

</details>

<a href="https://git.openjdk.org/jfx/pull/2154.diff">https://git.openjdk.org/jfx/pull/2154.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2154#issuecomment-4271428449)
</details>

### DIFF
--- a/modules/javafx.base/src/test/java/test/javafx/collections/ListChangeBuilderTest.java
+++ b/modules/javafx.base/src/test/java/test/javafx/collections/ListChangeBuilderTest.java
@@ -720,6 +720,30 @@ public class ListChangeBuilderTest {
         list.doEndChange();
     }
 
+    @Test
+    public void testRemoveDisjointFirstThenGaps() {
+        var list = new ExposedObservableList<>(new ArrayList<>(List.of("A", "B", "C", "D", "E", "F")));
+
+        list.addListener((ListChangeListener.Change<? extends String> change) -> {
+            assertEquals(1, change.getList().size());
+            change.next();
+
+            assertEquals(List.of("B", "C", "D", "E", "F"), change.getRemoved());
+
+            assertFalse(change.next());
+        });
+
+        list.doBeginChange();
+
+        list.remove("D");
+        list.remove("F");
+        list.remove("B");
+        list.remove("E");
+        list.remove("C");
+
+        list.doEndChange();
+    }
+
     private static class ExposedObservableList<E> extends ObservableListWrapper<E> {
         ExposedObservableList(List<E> list) {
             super(list);
--- a/modules/javafx.base/src/main/java/javafx/collections/ListChangeBuilder.java
+++ b/modules/javafx.base/src/main/java/javafx/collections/ListChangeBuilder.java
@@ -207,10 +207,6 @@ final class ListChangeBuilder<E> {
 
         if (last != null && last.to == idx) {
             last.removed.add(removed);
-        } else if (last != null && last.from == idx + 1) {
-            last.from--;
-            last.to--;
-            last.removed.add(0, removed);
         } else {
             insertRemoved(idx, removed);
         }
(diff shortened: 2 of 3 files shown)


modules/javafx.base/src/test/java/test/javafx/collections/ListChangeBuilderTest.java
```diff
@@ -720,6 +720,30 @@ public void testIndirectAddRemoveOnSameIndexWithAdded() {
...
+
+        list.doBeginChange();
+
+        list.remove("D");
+        list.remove("F");
+        list.remove("B");
+        list.remove("E");
+        list.remove("C");
+
+        list.doEndChange();
```
andy-goryachev-oracle, line 744: do you think it would make sense to add scenarios with .filtered() and .sorted() described in the reproducer for https://bugs.openjdk.org/browse/JDK-8237868 as tests?

https://bugs.openjdk.org/secure/attachment/86526/SampleApp.java

Maran23: While it is not a problem inside `FilteredList` and `SortedList`, more tests don't hurt. So I will have a look. Especially since we don't have many tests for the `FilteredList` and `SortedList`.

andy-goryachev-oracle: Thank you so much!

Maran23: I just pushed a `FilteredListTest` test based on the Reproducer

andy-goryachev-oracle (approved):
I agree, the original "optimization" looked sus.

Verified that the added tests fail in master and pass with the fix.

[JDK-8195614](https://bugs.openjdk.org/browse/JDK-8195614) and [JDK-8195750](https://bugs.openjdk.org/browse/JDK-8195750) are still broken (as expected).

It's as if we need to write a comprehensive set of tests of `FilteredList` and maybe `SortedList`...

Maran23:
> It's as if we need to write a comprehensive set of tests of `FilteredList` and maybe `SortedList`...

Had the same thought when I checked the referenced issues. 

After this is merged, I will have a look in the other 2 issues. 
There is already a proposed solution in there, but I haven't checked yet to see if it's correct.

mstr2 (approved):
Looks good.

PR #2156, opened 2026-04-20
8319555: [TestBug] Utility for creating instruction window for manual tests

andy-goryachev-oracle:
Delivers the following enhancements:
- wires the manual tests to the main gradle build https://bugs.openjdk.org/browse/JDK-8296441
- removes the limitation of single source file for manual tests
- provides the `ManualTestWindow` base class for manual tests with pass/fail buttons and instruction text area
- allows for skipping tests on certain platforms
- modified EmojiTest to showcase the new `ManualTestWindow`
- provides the manual test runner `TestRunnerApp` which allows for running manual tests in sequence or any order

At the moment, the new code is placed in the `tests/manual2` directory (we might decide to move it back to tests/manual once all the tests are migrated and placed in their own packages).

The `gradle manualApps` builds the classes in the `tests/manual2/build/classes` dir and generates a jar file containing all the tests, `ManualTestWindow`, `TestRunnerApp`, and the related test infrastructure (utilities etc.) in the `tests/manual2/dist/ManualTests.jar`.

## Gradle Build

`gradle clean manualApps`

## ManualTestWindow

This class simplifies writing manual tests.  Example:

```
public class ManualTestExample extends ManualTestWindow {
    public ManualTestExample() {
        super(
            "Manual Test Example",
            """
            Instructions:
            1. you will see a button named "Test"
            2. press the button
            3. verify that the button can be pressed""",
            400, 250
        );
     }

     @Override
     protected Node createContent() {
         return new Button("Test");
     }
}
```

Screenshot 2025-03-28 at 11 50 16

## TestRunnerApp

Screenshot 2026-04-20 at 13 24 41

Features:

- loads default test plan
- allows to load a test plan from file

Possible improvements:

- execute single test within tests/manual2 without writing a test plan
- generating report

## Migrating Manual Tests

Migrating existing tests is simple:
- the tests must not use the default package
- `TestRunnerApp` allows to drop the `main()` method.
- FQCN of the test added to the default test plan
- the test can optionally extend the `ManualTestWindow`

## Questions to Reviewers

1. should the `apps` target depend on `manualApps` so it builds the tests automatically?
2. what other functionality should be implemented in the `ManualTestWindow`?
3. what other functionality should be implemented in the `TestRunnerApp`?
4. should we add an ability to launch all the manual tests (or a single one) from gradle, or running the `TestRunnerApp` with the default test plan is sufficient?

---------

---------
### Progress

### Issues
 * [JDK-8319555](https://bugs.openjdk.org/browse/JDK-8319555): [TestBug] Utility for creating instruction window for manual tests (**Enhancement** - P4)
 * [JDK-8296441](https://bugs.openjdk.org/browse/JDK-8296441): Build and run manual tests from gradle (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2156/head:pull/2156` \
`$ git checkout pull/2156`

`$ git checkout pull/2156` \
`$ git pull https://git.openjdk.org/jfx.git pull/2156/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2156`

`$ git pr show -t 2156`

</details>

<a href="https://git.openjdk.org/jfx/pull/2156.diff">https://git.openjdk.org/jfx/pull/2156.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2156#issuecomment-4284499164)
</details>


tests/manual2/src/main/java/com/oracle/test/manual/util/TestRunner.java
```diff
@@ -0,0 +1,133 @@
...
+
+        String[] cmd = {
+            javaExecutablePath,
+            "-ea",
+            "-Djavafx.enablePreview=true",
+            "--enable-native-access=javafx.graphics",
+            "-Dfile.encoding=UTF-8",
+            "-Dstdout.encoding=UTF-8",
+            "-Dstderr.encoding=UTF-8",
+            "-p", "../../build/sdk/lib",
```
arapte, line 67: With this, the individual tests will be launched using the sdk built in the repo and not using the javafx sdk used to launch the manualTests.jar
May be, using same sdk is expectation in perspective of user ?

tests/manual2/src/main/java/com/oracle/test/manual/util/TestRunner.java
```diff
@@ -0,0 +1,133 @@
...
+            this.future = f;
+            this.client = client;
+        }
+
+        @Override
+        public void run() {
+            try {
+                Process p = future.get();
+                int result = p.exitValue();
+                setResult(result, null);
```
arapte, line 97: If the test app exits without clicking any button, it is considered as PASS. may be it should make it mandatory to click either button ?

tests/manual2/build.xml
```diff
@@ -0,0 +1,101 @@
...
+			destdir="build/classes"
+			debug="true"
+			encoding="utf-8"
+			fork="true"
+			nowarn="true"
+			optimize="false"
+			source="25"
+			target="25"
+			includeantruntime="false"
+		>
```
arapte, line 50: minor:move to previous line eol.

tests/manual2/README.md
```diff
@@ -0,0 +1,30 @@
...
+gradle manualApps
+```
+
+
+## Run
+
+To launch, specify the path to the JavaFX SDK lib/ folder on the command line, example:
+
+```
+java -p "../../build/sdk/lib" --enable-native-access=javafx.graphics --add-modules ALL-MODULE-PATH -jar dist/ManualTests.jar
```
arapte, line 28: It is necessary to do `cd tests/manual2`, else loading `test-plan.txt` fails.

arapte (changes requested):
Providing a few comments, overall lgtm.

tests/manual2/src/main/java/com/oracle/test/manual/util/TestRunner.java
andy-goryachev-oracle: By design: the point is to test the current jfx code.  Though I'd welcome any suggestions.

tests/manual2/src/main/java/com/oracle/test/manual/util/TestRunner.java
andy-goryachev-oracle: added "Cancelled" state

tests/manual2/build.xml
andy-goryachev-oracle: this is by design

build.gradle
```diff
@@ -4970,6 +4979,39 @@ project(":apps") {
...
+        List<String> params = []
+        params << "-DtargetBld=$t.name"
+        params << "-Dcompile.patch=@${rootProject.buildDir}/${COMPILEARGSFILE}"
+        params << "-Drun.patch=@${rootProject.buildDir}/${RUNARGSFILE}"
+
+        def manualAppsJar = project.task("manualAppsJar${t.capital}") {
+            dependsOn(sdk)
+
+            doLast() {
+                ant(t.name,
```
Maran23, line 4995: This is (all things considered) a broader question, but I would really recommend not adding any more Ant tasks (and thus increasing our dependency on Ant).
The goal can also be accomplished without Ant (and without `build.xml`), which simplifies the configuration, and will make us only use Gradle and not another tool like Ant.

cc @nlisker also interested in your opinion.

nlisker: Yes, this will be going backwards. The external executions that Gradle can't follow mess up the caching. I recently added a [Gradle build file](https://github.com/nlisker/jfx/blob/9eaf738cc83a0d4ae845346efd8c5e8f3b510ec0/tests/performance/3DLighting/build.gradle) for the 3DLighting test application, and that one is external to the main build. Here we're wiring a new project to the main build, thus increasing the technical debt.

andy-goryachev-oracle: I don't quite get this aversion to `ant`.  This build file worked without changes since probably 2014, yet with `gradle` we had to go through upgrades, additional build steps...

mstr2: Gradle tries really hard to avoid doing work, which significantly speeds up repeated builds. But that only works if Gradle can precisely track the inputs and outputs of tasks, which it can't do if we're invoking another build system.

andy-goryachev-oracle: it also adds significant overhead [0] and a maintenance burden.

[0] https://mill-build.org/blog/1-java-compile.html

nlisker: > I don't quite get this aversion to `ant`.

`ant` is not aware of the build around it. As a result, it doesn't reuse code so it's very verbose and it's much slower since it does no caching or parallelizing. It just executes the same thing every time. Try to write a build for JavaFX with `ant` and with Gradle and you'll quickly see how Gradle does almost all of the work for you.

> This build file worked without changes since probably 2014

Which one? The one in this PR is completely new.

> yet with gradle we had to go through upgrades, additional build steps...

Because JavaFX proper changed and these manual tests didn't. These tests were not modularized unlike the JavaFX modules, JavaFX got new modules and these tests didn't, and we chose to update the Gradle versions but not the `ant` version. If you keep the Gradle version at 4, it will also work with no changes, why update it? When was the last time `ant` was upgraded in JavaFX? According to my file system, not since 2018 at least (version 1.10.5).

Besides, the Gradle build of JavaFX runs an ant-like script to build everything, Gradle isn't exactly invoked, it's more of a shell. It was written incorrectly and eventually it will need to be fixed, which is what I'm trying to do.

The complaints you're voicing are akin to the "Java is slow" mantra coming from people with a badly written Java program written in 2004, and then saying "I don't get the aversion to Fortran, I never needed to change it".

nlisker: > it also adds significant overhead [0] and a maintenance burden.
> 
> [0] https://mill-build.org/blog/1-java-compile.html

It's a marketing page from a competitor from 3 years ago that compares a clean build, whereas the advantage of caching build systems is in repetitive builds. I'm sure it is faster than Gradle in some scenarios. Those examples don't even say if the dependencies were cached in the Gradle cache or downloaded as part of the build time (looks like the latter from the benchmarks).
The Eclipse compiler is also faster to work with than javac because it's incremental, but a clean compile might show javac being faster.

Still, if you want to use Mill I'm fine with it. I'm sure it's good. What does it have to do with `ant`?

andy-goryachev-oracle: All my builds are clean builds, I had my share of problems stemming from stale artifacts.

Do you want to migrate all the build scripts in the project to use gradle?

kevinrushforth: To Nir's point, this isn't an existing build script. There are no build scripts for manual tests and all of our other tests use gradle. Right now, only "apps" use ant, and that's because they are legacy. I wouldn't use that as an argument for expanding the use of ant beyond the apps dir.

nlisker: > All my builds are clean builds, I had my share of problems stemming from stale artifacts.

Yes, because the Gradle build file is incorrect. I filed [JDK-8391666](https://bugs.openjdk.org/browse/JDK-8391666) because the stale artifacts aren't detected correctly, which results in compromised builds. [JDK-8389921](https://bugs.openjdk.org/browse/JDK-8389921) fixed an example of this with native sources. With all of that, it's still rare that I need to clean every build.

> Do you want to migrate all the build scripts in the project to use gradle?

Without fixing the main build file it won't help much because these apps/tests are downstream of the broken build. They'll run a bit better than ant. I have https://github.com/openjdk/jfx/pull/2294 sitting in the review queue for a couple of weeks (not complaining), and https://github.com/openjdk/jfx/pull/2305 ready to go in directly after. Then a couple more in the chain that I'm waiting with until the first one moves.

mstr2: > it also adds significant overhead [0] and a maintenance burden.
> 
> [0] https://mill-build.org/blog/1-java-compile.html

That's because it's an actual build system, and not an artificial benchmark that shows anything and nothing. Once you solve all the problems of a complex build, you _will_ have a complex build script and the artificial advantage goes away. In fact, ant is _much_ slower than Gradle on very complex builds once you account for repeated edit-and-build scenarios.

andy-goryachev-oracle:
Okay, I have no expertise in `gradle`.  @nlisker would you like to help writing a `gradle` build script?

nlisker:
I can have a look, but I need to know the requirements. For the 3DLighting app, I was [asked to not wire it to the main build](https://github.com/openjdk/jfx/pull/1387#discussion_r1542126594). Here, you wired the manual tests to the main build, so did the requirements change?
What tasks do you need to have working? I see 2 `Application` classes, does Gradle need to run both?

andy-goryachev-oracle:
thank you!

`gradle clean manualApps` should clean and rebuild the jar.

The value of the jar is that it packages all the manual tests in a single jar that can be invoked either via `TestRunnerApp` (for convenience), or via command line.

Ideally, there should be a way to launch an individual test from gradle just like with other tests.

nlisker:
So should it be wired to the main build or external?

andy-goryachev-oracle:
wired to the main build:
- `gradle clean`
- `gradle manualApps`

Maran23:
> So should it be wired to the main build or external?

Is it easy to do the build externally? Also for the already existing `apps` at one point? And make them still run on `gradle all`?

nlisker:
> > So should it be wired to the main build or external?
> 
> Is it easy to do the build externally? Also for the already existing `apps` at one point? And make them still run on `gradle all`?

Currently it's easier because you don't have to integrate with the main build although it's semantically wrong since they are consumers of the build's intermediate results. I did [this](https://github.com/nlisker/jfx/blob/08d1335ad6f14932aa3b9237a809c39f96eb9e14/tests/performance/3DLighting/build.gradle) with the 3DLighting app as a compromise. You can have their build depend on `all`, but it will be very wasteful. I think that the best design here is to have a composite build (`includeBuild`) for the apps.

nlisker:
Alright, I have something working, but there's a naming confusion. The dir is `manual2`, the Eclipse project name is `ManualTests2`, and the Gradle project name is `manualApps`. I suggest using 1 name for all of these. Do you mind picking just one?

andy-goryachev-oracle:
> manual2

The end result will drop the 2, once the existing manual tests migrate in https://bugs.openjdk.org/browse/JDK-8392960 .  Before and during the migration, we'd need to keep two different directories.

The other alternative is to use `manualApps/` dir instead of `manual2` and then `rm manual/`.  @kevinrushforth what do you think?

PR #758, opened 2022-03-17
8283346: Optimize observable ArrayList creation in FXCollections

Maran23:
This simple PR optimizes the observable `ArrayList` creation by using the ArrayList constructor/array size so that the underlying array will be initialized at the correct size which will speed up the creation as the array does not need to grow as a result of the `addAll` call.

I also added tests which will succeed before and after to verify that nothing got broken by this change.
Also I made a benchmark test. Results:

| Benchmark | Mode| Cnt | Score | Error | Units
| ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| ListBenchmark OLD  | thrpt | 25 | 722,842 | ± 26,93 | ops/s
| ListBenchmark NEW | thrpt  | 25 | 29262,274 | ± 2088,712 | ops/s

Edit: I also made a synthetic benchmark by measuring the same code below 100 times with `System.nanoTime`.
ListBenchmark OLD (avg): 21-23ms
ListBenchmark NEW (avg): 2 ms

<details><summary>Benchmark code</summary>

```
import javafx.collections.FXCollections;
import javafx.collections.ObservableList;
import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.Setup;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.annotations.TearDown;

import java.util.ArrayList;
import java.util.List;

@State(Scope.Benchmark)
public class ListBenchmark {

    List<String> strings;

    @Setup
    public void setup() {
        strings = new ArrayList<>();
        for(int i = 0; i< 100000;i++) {
            strings.add("abc: " + i);
        }
    }

    @TearDown
    public void tearDown() {
        strings = null;
    }

    @Benchmark
    public ObservableList<String> init() {
        return FXCollections.observableArrayList(strings);
    }
}
```

</details>

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issue
 * [JDK-8283346](https://bugs.openjdk.org/browse/JDK-8283346): Optimize observable ArrayList creation in FXCollections

### Reviewers
 * [Michael Strauß](https://openjdk.org/census#mstrauss) (@mstr2 - Author)
[...]

PR #2163, opened 2026-05-03
8195614: FilteredList throws ArrayIndexOutOfBoundsException if created with 1 element

Maran23:
This PR fixes the (rare) issue that an `ArrayIndexOutOfBoundsException` is thrown in `FilteredList`.

Some details of the problem:
- Can only be reproduced when there are multiple changes, as we need an intermediate state where we have more elements than we will have at the end after all change events
- The issue is somewhat rare: Normally, we have reserved a big enough array (`int[] filtered`) to compensate a temporary element increase. With a small size like 1, we can reproduce it easily by making two change events (In the reproducer and test, a 'move element to first index' change operation)

Details for the fix:
- `ensureSize(getSource().size());` is normally correct as our `FilteredList` will always be <= source list. 
But we can have the situation where we have temporarily more elements than the source list as we are processing all change events (e.g. add 1, remove 1).
  - To account for that, we will call `ensureSize` again when we could run into the situation where we have more elements than we reserved for
  - So that we do not over allocate the `filtered` array, I did not use the proposed change in the ticket but this one instead. We will really only (may) increase the size when we really need it
    - As we reserve more capacity in `filtered` than we may need, this does not happen often, especially not with more elements
    - Since we have a smart change event aggregation algorithm (especially since the fix in https://github.com/openjdk/jfx/pull/2154 and https://github.com/openjdk/jfx/pull/2117), we often have more compact change event where a big increase of elements temporary, which are removed in the next change are very unlikely (there will be only one change event with the elements that 'survived' the add and remove). 
      - Therefore, I do not see any performance concerns here, as `ensureSize(size + 1);` is very likely a noop, but in cases like this e will correctly reserve more capacity

/issue add JDK-8195750

---------

---------
### Progress

### Issues
 * [JDK-8195614](https://bugs.openjdk.org/browse/JDK-8195614): FilteredList throws ArrayIndexOutOfBoundsException if created with 1 element (**Bug** - P3)
 * [JDK-8195750](https://bugs.openjdk.org/browse/JDK-8195750): FilteredList throws ArrayIndexOutOfBoundsException on ListChangeEvent with multiple change (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2163/head:pull/2163` \
`$ git checkout pull/2163`

`$ git checkout pull/2163` \
`$ git pull https://git.openjdk.org/jfx.git pull/2163/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2163`

`$ git pr show -t 2163`

</details>

<a href="https://git.openjdk.org/jfx/pull/2163.diff">https://git.openjdk.org/jfx/pull/2163.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2163#issuecomment-4366964241)
</details>

### DIFF
--- a/modules/javafx.base/src/main/java/javafx/collections/transformation/FilteredList.java
+++ b/modules/javafx.base/src/main/java/javafx/collections/transformation/FilteredList.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -135,6 +135,8 @@ public final class FilteredList<E> extends TransformationList<E, E>{
 
     @Override
     protected void sourceChanged(Change<? extends E> c) {
+        ensureSize(getSource().size());
+
         beginChange();
         while (c.next()) {
             if (c.wasPermutated()) {
@@ -236,7 +238,6 @@ public final class FilteredList<E> extends TransformationList<E, E>{
 
     private void addRemove(Change<? extends E> c) {
         Predicate<? super E> pred = getPredicateImpl();
-        ensureSize(getSource().size());
         final int from = findPosition(c.getFrom());
         final int to = findPosition(c.getFrom() + c.getRemovedSize());
 
@@ -269,6 +270,7 @@ public final class FilteredList<E> extends TransformationList<E, E>{
             // Add the remaining elements
             while (it.nextIndex() < c.getTo()) {
                 if (pred.test(it.next())) {
+                    ensureSize(size + 1);
                     System.arraycopy(filtered, fpos, filtered, fpos + 1, size - fpos);
                     filtered[fpos] = it.previousIndex();
                     nextAdd(fpos, fpos + 1);
@@ -282,7 +284,6 @@ public final class FilteredList<E> extends TransformationList<E, E>{
 
     private void update(Change<? extends E> c) {
         Predicate<? super E> pred = getPredicateImpl();
-        ensureSize(getSource().size());
         int sourceFrom = c.getFrom();
         int sourceTo = c.getTo();
         int filterFrom = findPosition(sourceFrom);
--- a/modules/javafx.base/src/test/java/test/javafx/collections/FilteredListTest.java
+++ b/modules/javafx.base/src/test/java/test/javafx/collections/FilteredListTest.java
@@ -27,6 +27,7 @@ package test.javafx.collections;
 
 import com.sun.javafx.collections.ObservableListWrapper;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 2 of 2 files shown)


modules/javafx.base/src/main/java/javafx/collections/transformation/FilteredList.java
```diff
@@ -135,6 +135,8 @@ private Predicate<? super E> getPredicateImpl() {
 
     @Override
     protected void sourceChanged(Change<? extends E> c) {
+        ensureSize(getSource().size());
```
andy-goryachev-oracle, line 138: are you sure this is needed?

the tests in the two JBS tickets pass without it, `addRemove()` invokes this method in ~L239 (then calls it again with size+1 in L273).  `update()` calls it in ~L286, and `permutate()` does not need it.

perhaps `addRemove()` can be further optimized so as not to call `ensureSize()` twice?

andy-goryachev-oracle (commented):
minor optimization might be possible.

modules/javafx.base/src/main/java/javafx/collections/transformation/FilteredList.java
Maran23: `update()` and `permutate()` do not call it anymore. I relocated them into one single location: Here.
So functionality wise, it did not change.

We could probably remove it, as we will catch any grow on size in `addRemove()`. But it might hurt performance.
If we have a `FilteredList` with a size 1 and add 10_000 elements, we will call `ensureSize` a lot of times, which will lead to a degraded performance. 

But since we call it once here, we will immediately grow to the worst case size - if nothing need to be filtered and MAY only grow it in `addRemove()` when we temporarily have more elements due to processing the changes one by one.

andy-goryachev-oracle: I might be misunderstanding something.  

1. Is `ensureSize()` L138 needed for `permutate()` ?
2. `addRemove()` calls `ensureSize()` twice (L138 and L237).  can it call it only once?
3. `update()` does not change the filtered size, right?  why is it calling `ensureSize()` in L138?

Maran23: 1. No - but when we are only sorting (permutation) the list, the `ensureSize` should be a noop, since the size had to be increased at some point in the past to what we currently have and ensured the size
2. No - because a sub change temporarily adds an element which is later removed on another sub change. We unfortunately can not predict that/count that in beforehand in this `ensureSize` call. With that said, the second call in `addRemove` should happen very rarely - probably only with small sizes
3. It can. When we change an item, lets say A -> B and A is filtered out but B is not, we grow in size by 1 (or the other way around). But same as in the `permutation`,- the `ensureSize` should be a noop, since the size had to be increased at some point in the past to what we currently have and ensured the size

andy-goryachev-oracle: So, while I agree that it's likely to be a no-op, my original comment did pointed out a possible optimization (not that it will give us megabytes or free space or 10% less CPU consumption).

So in cases 1) and 2) we could call `ensureSize()` only when needed.  It's up to you whether you want to pursue this further or leave as is.

andy-goryachev-oracle (approved):

modules/javafx.base/src/main/java/javafx/collections/transformation/FilteredList.java
Maran23: Yeah I thought about it as well - but this is very hard to balance.
We reserve quite much space here. If we have 1000 elements, we reserve space for 1501 elements. That's huge. And we are in `FilteredList` - so the normal usecase is probably with quite a few elements filtered out.

But on the other hand growing in size just a few times is much better performance-wise vs. growing in size many more times. As I found out here already: https://github.com/openjdk/jfx/pull/758

mstr2 (approved):

PR #2165, opened 2026-05-06
8384006: ComboBox text does not update on String converter update

crschnick:
This PR adds a missing call to update the display value to the converter property listener

---------

---------
### Progress

### Issue
 * [JDK-8384006](https://bugs.openjdk.org/browse/JDK-8384006): ComboBox text does not update on String converter update (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2165/head:pull/2165` \
`$ git checkout pull/2165`

`$ git checkout pull/2165` \
`$ git pull https://git.openjdk.org/jfx.git pull/2165/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2165`

`$ git pr show -t 2165`

</details>

<a href="https://git.openjdk.org/jfx/pull/2165.diff">https://git.openjdk.org/jfx/pull/2165.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2165#issuecomment-4390685873)
</details>

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/ComboBox.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/ComboBox.java
@@ -316,9 +316,15 @@ public class ComboBox<T> extends ComboBoxBase<T> {
 
     // --- string converter
     /**
-     * Converts the user-typed input (when the ComboBox is
-     * {@link #editableProperty() editable}) to an object of type T, such that
-     * the input may be retrieved via the  {@link #valueProperty() value} property.
+     * Converts items of type T to a string representation and the other way around.
+     *
+     * The converter is used to obtain the string representation that should be displayed
+     * for the {@link #valueProperty() value} and any {@link #getItems() item}.
+     *
+     * When the ComboBox is {@link #editableProperty() editable}), the converter is used
+     * to convert the input to an object of type T, such that the input may be retrieved
+     * via the {@link #valueProperty() value} property.
+     *
      * @return the converter property
      */
     public final ObjectProperty<StringConverter<T>> converterProperty() { return converter; }
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
@@ -179,7 +179,7 @@ public class ComboBoxListViewSkin<T> extends ComboBoxPopupControl<T> {
             listView.requestLayout();
         });
         lh.addChangeListener(control.converterProperty(), e -> {
-            updateCellFactory();
+            listView.refresh();
             updateDisplayNode();
         });
         lh.addChangeListener(control.buttonCellProperty(), e -> {
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
@@ -178,7 +178,10 @@ public class ComboBoxListViewSkin<T> extends ComboBoxPopupControl<T> {
             if (listView == null) return;
             listView.requestLayout();
         });
-        lh.addChangeListener(control.converterProperty(), e -> updateListViewItems());
+        lh.addChangeListener(control.converterProperty(), e -> {
+            updateListViewItems();
+            updateDisplayNode();
+        });
         lh.addChangeListener(control.buttonCellProperty(), e -> {
             updateButtonCell();
             updateDisplayArea();
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
@@ -179,7 +179,7 @@ public class ComboBoxListViewSkin<T> extends ComboBoxPopupControl<T> {
             listView.requestLayout();
         });
         lh.addChangeListener(control.converterProperty(), e -> {
-            updateListViewItems();
+            updateCellFactory();
             updateDisplayNode();
         });
         lh.addChangeListener(control.buttonCellProperty(), e -> {
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
@@ -67,6 +67,7 @@ import javafx.scene.control.TabPane;
 import javafx.scene.control.TextField;
 import javafx.scene.control.Tooltip;
 import javafx.scene.control.skin.ComboBoxListViewSkin;
+import javafx.scene.control.skin.ListViewSkin;
 import javafx.scene.control.skin.VirtualFlow;
 import javafx.scene.input.KeyCode;
 import javafx.scene.layout.BorderPane;
@@ -93,6 +94,7 @@ import test.com.sun.javafx.scene.control.infrastructure.KeyEventFirer;
 import test.com.sun.javafx.scene.control.infrastructure.KeyModifier;
 import test.com.sun.javafx.scene.control.infrastructure.MouseEventFirer;
 import test.com.sun.javafx.scene.control.infrastructure.StageLoader;
+import test.com.sun.javafx.scene.control.infrastructure.VirtualFlowTestUtils;
 
 public class ComboBoxTest {
     private ComboBox<String> comboBox;
@@ -385,6 +387,33 @@ public class ComboBoxTest {
         });
 
         assertEquals("item1", field.getText());
+    }
+
+    @Test
+    public void testListUpdateOnStringConverterChange() {
+        ObservableList<String> items = FXCollections.observableArrayList("ITEM1", "ITEM2");
+        comboBox.setEditable(false);
+        comboBox.setItems(items);
+        comboBox.setValue("ITEM1");
+        comboBox.setConverter(new StringConverter<>() {
+            @Override
+            public String toString(String object) {
+                return object.toLowerCase();
+            }
+
+            @Override
+            public String fromString(String string) {
+                return "?";
+            }
+        });
+        comboBox.getItems().add("ITEM3");
+
+        ListView<String> list = (ListView<String>) ((ComboBoxListViewSkin<String>) comboBox.getSkin())
+                .getPopupContent();
+        list.setSkin(new ListViewSkin<>(list));
+        VirtualFlowTestUtils.assertCellTextEquals(list, 0, "item1");
+        VirtualFlowTestUtils.assertCellTextEquals(list, 1, "item2");
+        VirtualFlowTestUtils.assertCellTextEquals(list, 2, "item3");
     }
 
     @Test public void testNullSelectionModelDoesNotThrowNPEOnValueChange() {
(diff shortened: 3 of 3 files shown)


modules/javafx.controls/src/main/java/javafx/scene/control/ComboBox.java
```diff
@@ -316,9 +316,15 @@ public ComboBox(ObservableList<T> items) {
...
-     * Converts the user-typed input (when the ComboBox is
-     * {@link #editableProperty() editable}) to an object of type T, such that
-     * the input may be retrieved via the  {@link #valueProperty() value} property.
+     * Converts items of type T to a string representation and the other way around.
+     *
+     * The converter is used to obtain the string representation that should be displayed
+     * for the {@link #valueProperty() value} and any {@link #getItems() item}.
+     *
+     * When the ComboBox is {@link #editableProperty() editable}), the converter is used
+     * to convert the input to an object of type T, such that the input may be retrieved
```
andy-goryachev-oracle, line 325: I like the new javadoc very much, thank you!

the only nitpick is "the converter is used to convert", can it be re-phrased?

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
```diff
@@ -335,6 +335,56 @@ protected void updateItem(String item, boolean empty) {
         comboBox.setEditable(false);
     }
 
+    @Test public void testCellUpdateOnStringConverterChange() {
```
andy-goryachev-oracle: could you move the `@Test` annotation to its own line here and L363 please?

andy-goryachev-oracle (commented):
I will need to test it, here are some some preliminary comments -

modules/javafx.controls/src/main/java/javafx/scene/control/ComboBox.java
crschnick: I thought about, but something like mapped or transformed might not be the best either

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
crschnick: Fixed

andy-goryachev-oracle:
While the text field does get updated, the list does not.

To reproduce, change the converter from Number to Quoted in the monkey tester and show the pulldown.  The list gets only partially updated:

Screenshot 2026-05-07 at 14 39 58

crschnick:
This should be fixed now, it seems like the old implementation was not calling the right update methods.

I tested it with the monkey tester now as well.

andy-goryachev-oracle (approved):
Looks good now, thank you.

I wonder if it would make sense to add a test that checks the pulldown as well (the problem that the last commit fixed).  That test should fail without the last commit.

What do you think?

crschnick:
I tried to add a test for this, but I did not find an easy way to obtain that information from the skin

Maran23:
> I tried to add a test for this, but I did not find an easy way to obtain that information from the skin

Maybe call `lookup` to get the `ListView` and then just use `VirtualFlowTestUtils`?

crschnick:
@Maran23 Thanks for the tip, that worked. This should be good to go now

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
```diff
@@ -335,6 +337,85 @@ protected void updateItem(String item, boolean empty) {
...
+            }
+
+            @Override
+            public String fromString(String string) {
+                return "?";
+            }
+        });
+        comboBox.getItems().add("ITEM3");
+
+        ListView<String> list = (ListView<String>) ((ComboBoxListViewSkin<String>) comboBox.getSkin())
```
Maran23: can't you call `getListView()`, which is defined here in this class?

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
```diff
@@ -335,6 +337,85 @@ protected void updateItem(String item, boolean empty) {
...
+    }
+
+    @Test
+    public void testTextFieldUpdateOnStringConverterChange() {
+        ObservableList<String> items = FXCollections.observableArrayList("ITEM1", "ITEM2");
+        comboBox.setEditable(true);
+        comboBox.setItems(items);
+        comboBox.setValue("ITEM1");
+
+        TextField field = (TextField) ((ComboBoxListViewSkin<String>) comboBox.getSkin())
```
Maran23: Same here, call `getDisplayNode()` which is defined in this class and cast it to `TextField`

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
```diff
@@ -335,6 +337,85 @@ protected void updateItem(String item, boolean empty) {
...
+            @Override
+            public String fromString(String string) {
+                return "?";
+            }
+        });
+        comboBox.getItems().add("ITEM3");
+
+        ListView<String> list = (ListView<String>) ((ComboBoxListViewSkin<String>) comboBox.getSkin())
+                .getPopupContent();
+        list.setSkin(new ListViewSkin<>(list));
```
Maran23: This is needed because the `ComboBox` is not inside a `Stage`. 
Not needed when you call:
`sl =new StageLoader(comboBox);`

Just FYI, is probably fine as is. But most of the time you rather want to simulate a real scenario.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
```diff
@@ -335,6 +337,85 @@ protected void updateItem(String item, boolean empty) {
         comboBox.setEditable(false);
     }
 
+    @Test
+    public void testCellUpdateOnStringConverterChange() {
```
Maran23: Minor: Maybe name it `testDisplayNodeUpdateOnStringConverterChange`

Maran23 (commented):
Looks good to me. I also like the new javadoc. Some minor test comments.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
crschnick: Done

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
crschnick: Done

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
crschnick: I updated the name to testButtonCell... to indicate that we are testing the non-editable case here. Display node could mean both button cell or text field

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
```diff
@@ -178,7 +178,10 @@ public ComboBoxListViewSkin(final ComboBox<T> control) {
             if (listView == null) return;
             listView.requestLayout();
         });
-        lh.addChangeListener(control.converterProperty(), e -> updateListViewItems());
+        lh.addChangeListener(control.converterProperty(), e -> {
+            updateCellFactory();
```
Maran23: Here, I wonder if we just can call `listView.refresh()`. Because we effectively want all cells to 'recalculate' their value.

And there is one problem now with the current approach: If we have custom `cellFactory` that uses the `converter`, it won't get updated. 

Because in `updateCellFactory()` the `cellFactory` is the same as before, so will be a noop and the `ListView` items will not change.

<details>
<summary>Reproducer</summary>

```
import javafx.scene.Scene;
import javafx.scene.control.ComboBox;
import javafx.scene.control.ListCell;
import javafx.stage.Stage;
import javafx.util.StringConverter;

public class ComboBoxCustomListCell extends Application {

    private final ObservableList<String> strings = FXCollections.observableArrayList("Option 1", "Option 2",
            "Option 3");

        launch(args);
    }

    @Override
        final ComboBox<String> comboBox = new ComboBox<>();
        comboBox.setConverter(new StringConverter<>() {
            @Override
            public String toString(String object) {
                return "CONVERTER: " + object;
            }

            @Override
            public String fromString(String string) {
                return null;
            }
        });
        comboBox.setItems(FXCollections.observableArrayList(strings));
        comboBox.setCellFactory(_ -> new ListCell<>() {
            @Override
            public void updateItem(String item, boolean empty) {
                super.updateItem(item, empty);

                if (empty || item == null) {
                    setText(null);
                } else {
                    setText(comboBox.getConverter().toString(item));
                }
            }
        });

        BorderPane root = new BorderPane(comboBox);
        Button value = new Button("Change Converter");
        value.setOnAction(event -> {
            comboBox.setConverter(new StringConverter<>() {
                @Override
                public String toString(String object) {
                    return "NEW: " + object;
                }

                @Override
                public String fromString(String string) {
                    return null;
                }
            });
        });
        root.setBottom(value);
        Scene scene = new Scene(root, 600, 400);

        stage.setScene(scene);
        stage.show();
    }
}
```

</details>

Maran23 (changes requested):
There is one usecase from my testing that is still broken. But I also proposed a fix.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
crschnick: That is a good point, I updated the PR.

Maran23 (approved):

andy-goryachev-oracle (approved):
Looks good, and thank you for adding the tests!

Somewhat related, I tried to set a null converter, and that broke the monkey tester.  There is nothing in the API spec that talks about nullability of this property, but I would expect that setting it to `null` should be equivalent to choosing the default converter 
( `ComboBox.<T>defaultStringConverter()` , ComboBox:332 )

What do you guys think?

If you think we should allow a `null` converter, I can file a new RFE and add it under 
https://bugs.openjdk.org/browse/JDK-8334883

crschnick:
That would make sense, although I would see that as a low priority

Maran23:
> Somewhat related, I tried to set a null converter, and that broke the monkey tester. There is nothing in the API spec that talks about nullability of this property, but I would expect that setting it to `null` should be equivalent to choosing the default converter ( `ComboBox.<T>defaultStringConverter()` , ComboBox:332 )

This is an interesting usecase. From a quick look, other Controls will also not handle this. Example: `CheckBoxTreeTableCell`.
`TextFieldListCell` will throw an `IllegalStateException` - can be seen in `CellUtils#createTextField`.

`ChoiceBox` handles this case and just use `toString()`. `Spinner` as well with a factory without a converter.

I would als recommend that we do the same. Just call `toString()` as it is already done pretty much everywhere (including cells, when you don't set a `setCellValueFactory`.

So something like in `ChoiceBoxSkin`:

```
    private String getDisplayText(T value) {
        if (getSkinnable().getConverter() != null) {
            return getSkinnable().getConverter().toString(value);
        }
        return value == null ? "" : value.toString();
    }
```

andy-goryachev-oracle:
Filed https://bugs.openjdk.org/browse/JDK-8384534 for the null converter issue.

PR #2166, opened 2026-05-12

beldenfox:
This PR alters the way ComboBox and Spinner deliver KeyEvents to their TextField editors. When a ComboBox or Spinner is the focus owner it is the target of all key events. Currently the skin installs a filter to catch key events and re-fire most of them at the TextEdit. The skin copies the event, fires the copy at the TextField, and then consumes the original event. This confuses the system menu bar logic on macOS; only the original event can trigger a menu item and that event is always being consumed.

In this PR only the original key event makes its way up and down the event dispatch chain. To drive the TextField the skin delivers the event copy directly to the TextField's event dispatcher and only consumes the original event if the TextField consumes the copy.

---------

---------
### Progress

### Issue
 * [JDK-8354539](https://bugs.openjdk.org/browse/JDK-8354539): [macOS] ComboBox and Spinner disable system menu bar shortcuts (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2166/head:pull/2166` \
`$ git checkout pull/2166`

`$ git checkout pull/2166` \
`$ git pull https://git.openjdk.org/jfx.git pull/2166/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2166`

`$ git pr show -t 2166`

</details>

<a href="https://git.openjdk.org/jfx/pull/2166.diff">https://git.openjdk.org/jfx/pull/2166.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2166#issuecomment-4433074265)
</details>

### DIFF
--- a/modules/javafx.base/src/main/java/com/sun/javafx/event/EventUtil.java
+++ b/modules/javafx.base/src/main/java/com/sun/javafx/event/EventUtil.java
@@ -29,7 +29,6 @@ import java.util.concurrent.atomic.AtomicBoolean;
 
 import javafx.event.Event;
 import javafx.event.EventDispatchChain;
-import javafx.event.EventDispatcher;
 import javafx.event.EventTarget;
 
 public final class EventUtil {
@@ -73,14 +72,5 @@ public final class EventUtil {
         final EventDispatchChain targetDispatchChain =
                 eventTarget.buildEventDispatchChain(eventDispatchChain);
         return targetDispatchChain.dispatchEvent(event);
-    }
-
-    public static Event dispatch(Event event, EventDispatcher dispatcher) {
-        if (dispatcher == null || event == null) {
-            return null;
-        }
-        var chain = new EventDispatchChainImpl();
-        chain.append(dispatcher);
-        return chain.dispatchEvent(event);
     }
 }
--- a/modules/javafx.base/src/main/java/com/sun/javafx/event/EventUtil.java
+++ b/modules/javafx.base/src/main/java/com/sun/javafx/event/EventUtil.java
@@ -29,6 +29,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 
 import javafx.event.Event;
 import javafx.event.EventDispatchChain;
+import javafx.event.EventDispatcher;
 import javafx.event.EventTarget;
 
 public final class EventUtil {
@@ -72,5 +73,14 @@ public final class EventUtil {
         final EventDispatchChain targetDispatchChain =
                 eventTarget.buildEventDispatchChain(eventDispatchChain);
         return targetDispatchChain.dispatchEvent(event);
+    }
+
+    public static Event dispatch(Event event, EventDispatcher dispatcher) {
+        if (dispatcher == null || event == null) {
+            return null;
+        }
+        var chain = new EventDispatchChainImpl();
+        chain.append(dispatcher);
+        return chain.dispatchEvent(event);
     }
 }
--- a/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/skin/Utils.java
+++ b/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/skin/Utils.java
@@ -49,10 +49,12 @@ import javafx.beans.InvalidationListener;
 import javafx.beans.Observable;
 import javafx.beans.value.ObservableValue;
 import javafx.collections.ObservableList;
+import javafx.event.Event;
 import javafx.geometry.Bounds;
 import javafx.geometry.HPos;
 import javafx.geometry.Point2D;
 import javafx.geometry.VPos;
+import javafx.scene.Node;
 import javafx.scene.Scene;
 import javafx.scene.control.ContextMenu;
 import javafx.scene.control.MenuItem;
@@ -66,6 +68,7 @@ import javafx.scene.text.TextBoundsType;
 import com.sun.javafx.scene.NodeHelper;
 import com.sun.javafx.scene.control.ContextMenuContent;
 import com.sun.javafx.scene.control.behavior.MnemonicInfo;
+import com.sun.javafx.event.EventDispatchChainImpl;
 import com.sun.javafx.scene.text.FontHelper;
 import com.sun.javafx.scene.text.TextLayout;
 import com.sun.javafx.tk.Toolkit;
@@ -988,5 +991,15 @@ public class Utils {
 
     public static URL getResource(String str) {
         return Utils.class.getResource(str);
+    }
+
+    // Dispatches the event to the Node's dispatcher and returns
+    // true if the event was consumed.
+    public static boolean dispatchToNode(Event event, Node node) {
+        var dispatcher = node.getEventDispatcher();
+        if (dispatcher == null) return false;
+        var chain = new EventDispatchChainImpl();
+        chain.append(dispatcher);
+        return (chain.dispatchEvent(event.copyFor(node, node)) == null);
     }
 }
--- a/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/skin/Utils.java
+++ b/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/skin/Utils.java
@@ -993,8 +993,8 @@ public class Utils {
         return Utils.class.getResource(str);
     }
 
-    // Dispatches the event to the Node's dispatcher and returns
-    // true if the event was consumed.
+    // Dispatches the event only to the Node's dispatcher, not its entire
+    // dispatch chain. Returns true if the event was consumed.
     public static boolean dispatchToNode(Event event, Node node) {
         var dispatcher = node.getEventDispatcher();
         if (dispatcher == null) return false;
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxPopupControl.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxPopupControl.java
@@ -53,6 +53,7 @@ import com.sun.javafx.scene.control.FakeFocusTextField;
 import com.sun.javafx.scene.control.ListenerHelper;
 import com.sun.javafx.scene.control.Properties;
 import com.sun.javafx.scene.control.behavior.TextInputControlBehavior;
+import com.sun.javafx.scene.control.skin.Utils;
 import com.sun.javafx.scene.input.ExtendedInputMethodRequests;
 import com.sun.javafx.scene.traversal.Algorithm;
 import com.sun.javafx.scene.traversal.Direction;
@@ -171,18 +172,12 @@ public abstract class ComboBoxPopupControl<T> extends ComboBoxBaseSkin<T> {
                     // Fix for the regression noted in a comment in JDK-8115009.
                     // This forwards the event down into the TextField when
                     // the key event is actually received by the ComboBox.
-                    if (forwardToTextField(ke)) {
+                    if (Utils.dispatchToNode(ke, textField)) {
                         ke.consume();
                     }
                 }
             }
         });
-    }
-
-    // Returns 'true' if the event was consumed.
-    private boolean forwardToTextField(KeyEvent event) {
-        return EventUtil.dispatch(event.copyFor(textField, textField),
-            textField.getEventDispatcher()) == null;
     }
 
     @Override
@@ -614,7 +609,7 @@ public abstract class ComboBoxPopupControl<T> extends ComboBoxBaseSkin<T> {
             if (doConsume && comboBoxBase.getOnAction() != null) {
                 ke.consume();
             } else if (textField != null) {
-                forwardToTextField(ke);
+                Utils.dispatchToNode(ke, textField);
             }
         } else if (ke.getCode() == KeyCode.F10 || ke.getCode() == KeyCode.ESCAPE) {
             // JDK-8115456: The TextField fires F10 and ESCAPE key events
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxPopupControl.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxPopupControl.java
@@ -28,6 +28,7 @@ package javafx.scene.control.skin;
 import javafx.beans.value.ObservableValue;
 import javafx.css.Styleable;
 import javafx.event.EventHandler;
+import javafx.event.EventDispatchChain;
 import javafx.geometry.Bounds;
 import javafx.geometry.HPos;
 import javafx.geometry.Point2D;
@@ -47,6 +48,7 @@ import javafx.scene.input.MouseEvent;
 import javafx.scene.layout.Region;
 import javafx.stage.WindowEvent;
 import javafx.util.StringConverter;
+import com.sun.javafx.event.EventDispatchChainImpl;
 import com.sun.javafx.scene.ParentHelper;
 import com.sun.javafx.scene.control.FakeFocusTextField;
 import com.sun.javafx.scene.control.ListenerHelper;
@@ -170,11 +172,22 @@ public abstract class ComboBoxPopupControl<T> extends ComboBoxBaseSkin<T> {
                     // Fix for the regression noted in a comment in JDK-8115009.
                     // This forwards the event down into the TextField when
                     // the key event is actually received by the ComboBox.
-                    textField.fireEvent(ke.copyFor(textField, textField));
-                    ke.consume();
+                    if (forwardToTextField(ke)) {
+                        ke.consume();
+                    }
                 }
             }
         });
+    }
+
+    // Returns 'true' if the event was consumed.
+    private boolean forwardToTextField(KeyEvent event) {
+        var dispatcher = textField.getEventDispatcher();
+        if (dispatcher == null) return false;
+
+        EventDispatchChain chain = new EventDispatchChainImpl();
+        chain.append(textField.getEventDispatcher());
+        return chain.dispatchEvent(event.copyFor(textField, textField)) == null;
     }
 
     @Override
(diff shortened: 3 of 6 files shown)


modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxPopupControl.java
```diff
@@ -170,13 +172,24 @@ public ComboBoxPopupControl(ComboBoxBase<T> control) {
...
             }
         });
     }
 
+    // Returns 'true' if the event was consumed.
+    private boolean forwardToTextField(KeyEvent event) {
+        var dispatcher = textField.getEventDispatcher();
+        if (dispatcher == null) return false;
+
+        EventDispatchChain chain = new EventDispatchChainImpl();
```
andy-goryachev-oracle: would it make sense to create a public API to dispatch an event specifically to the given `Node`, bypassing the normal path (as in `Event.fireEvent` ?)

something along the lines `EventUtil.fireEvent` ?

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
```diff
@@ -2540,4 +2543,17 @@ public void testSelectionOnItemChangeUsingSetAllOnButtonPress() {
         mouse.fireMousePressAndRelease();
         assertEquals("B", comboBox.getSelectionModel().getSelectedItem());
     }
+
+    // Ensure initial shortcut event is not consumed
+    @Test public void testShortcutNotConsumed() {
```
andy-goryachev-oracle: I'd strongly encourage to place @Test annotation on its own line in all the new code...

andy-goryachev-oracle (commented):
I like this idea very much (will do more testing).

Do you think these tickets are related or similar:

https://bugs.openjdk.org/browse/JDK-8174991
https://bugs.openjdk.org/browse/JDK-8229914
https://bugs.openjdk.org/browse/JDK-8088897
https://bugs.openjdk.org/browse/JDK-8234247
https://bugs.openjdk.org/browse/JDK-8337246
https://bugs.openjdk.org/browse/JDK-8320557

Also, I wonder if other virtualized controls with embedded editors such as `ListView`, `Tree|TableView` might benefit from this fix?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxPopupControl.java
beldenfox: > would it make sense to create a public API to dispatch an event specifically to the given `Node`, bypassing the normal path (as in `Event.fireEvent` ?)

I wouldn't add such a specialized API. It would make more sense to provide an API so clients can easily create event dispatch chains. The EventDispatchChain interface is public but the implementation is not and there's no supported way for clients to create one. With one of those in hand it's easy to pass an event to a single dispatcher or even a set of dispatchers.

beldenfox:
> I like this idea very much (will do more testing).
> 
> Do you think these tickets are related or similar:

I looked into several of these and even prototyped some fixes. The problems with ESCAPE and ENTER being consumed are at a higher level. For example, in some case the InputMaps are auto-consuming ESCAPE when it should be consumed conditionally.

This PR deals with the way ComboBox and Spinner re-fire events. Some of the bug reports mention event re-firing but I think those might be obsolete. There's comments in the code suggesting TextEdits used to re-fire ESCAPE at their parents but I don't think that's happening anymore.

> Also, I wonder if other virtualized controls with embedded editors such as `ListView`, `Tree|TableView` might benefit from this fix?

I'm not familiar with those controls but as far as I know they don't re-fire KeyEvents to drive their editors. I think they set focus directly on their editors when editing in which case this fix isn't relevant. Some quick testing seems to confirm this; they process shortcuts like Cmd+Q just fine.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxPopupControl.java
```diff
@@ -170,13 +172,24 @@ public ComboBoxPopupControl(ComboBoxBase<T> control) {
...
+        chain.append(textField.getEventDispatcher());
```
andy-goryachev-oracle: you already have the dispatcher instance in L185

modules/javafx.controls/src/main/java/javafx/scene/control/skin/SpinnerSkin.java
```diff
@@ -258,6 +246,16 @@ public void executeAccessibleAction(AccessibleAction action, Object... parameter
         });
     }
 
+    // Returns 'true' if the event was consumed.
+    private boolean forwardToTextField(KeyEvent event) {
+        var dispatcher = textField.getEventDispatcher();
+        if (dispatcher == null) return false;
+
+        EventDispatchChain chain = new EventDispatchChainImpl();
+        chain.append(textField.getEventDispatcher());
```
andy-goryachev-oracle: replace with the `dispatcher` pointer?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxPopupControl.java
andy-goryachev-oracle: Maybe I wasn't clear.  You have two methods - this and an identical one in
https://github.com/openjdk/jfx/pull/2166/changes#diff-9117771ea1db4c49f8f0e82634427f7d8205b1f15038c1429959a003a10a5af7R250

that dispatch an event to a component, bypassing the normal chain, correct?

Why not make it a public API?  `Event.dispatchTo(Node)` or something like that?

Or, at a minimum, replace these two with a utility method added to `EventUtil` ?

beldenfox: > Why not make it a public API? `Event.dispatchTo(Node)` or something like that?

Dispatching to a single node is a very specialized case and I don't think it's worth adding a public routine. We would get more mileage out of adding a public API to create an EventDispatchChain. Then clients could write their own version of this short routine or build more complicated chains that dispatch across multiple nodes.

> Or, at a minimum, replace these two with a utility method added to `EventUtil` ?

This routine is only used in two places and I'm hoping it stays that way. The right location for this routine is a common superclass of ComboBox and Spinner (something like ForwardingControl) but that doesn't exist and I don't know how to add it after the fact. As it stands Spinner doesn't handle input method events or requests correctly and to fix that we'll have to copy code verbatim from ComboBox. New controls that need to do this sort of forwarding should start with a common superclass and it should probably pursue a different design.

andy-goryachev-oracle: I suspect it may not be a specialized case (any composite control might want to dispatch an event to its child control for example), and you already have two places with the duplicate code.  

I am ok with this not being a public API, so perhaps it belongs in `EventUtils`.  Certainly not in the ComboBox base class.  I don't quite understand the reason for your objection here.

andy-goryachev-oracle:
You probably want to sync up with the latest master.  This should be a rule for all long-lived PRs (which is pretty much all of them, sorry about that!)

modules/javafx.base/src/main/java/com/sun/javafx/event/EventUtil.java
```diff
@@ -73,4 +74,13 @@ private static Event fireEventImpl(EventDispatchChain eventDispatchChain,
                 eventTarget.buildEventDispatchChain(eventDispatchChain);
         return targetDispatchChain.dispatchEvent(event);
     }
+
+    public static Event dispatch(Event event, EventDispatcher dispatcher) {
```
andy-goryachev-oracle: would it make sense to pass the `Node` instead of the dispatcher, and return a `boolean`?

beldenfox: EventUtil is in the javafx.base module and doesn't have access to Node.

Dispatching an event returns an event (possibly null) and all other routines in EventUtil return that event. It strikes me as artificial to only check whether the event was consumed or not.

andy-goryachev-oracle: This is a utility method to dispatch an `Event` to a `Node`, right?  So returning a boolean which is then passed on makes sense.  The `Node.getEventDispatcher()` is a public API.

Perhaps I don't quite understand your objection?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxPopupControl.java
```diff
@@ -606,7 +614,7 @@ private void handleKeyEvent(KeyEvent ke, boolean doConsume) {
             if (doConsume && comboBoxBase.getOnAction() != null) {
                 ke.consume();
             } else if (textField != null) {
-                textField.fireEvent(ke);
+                forwardToTextField(ke);
```
andy-goryachev-oracle: Should the `ke` event be consumed if its copy was consumed?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxPopupControl.java
```diff
@@ -170,13 +171,20 @@ public ComboBoxPopupControl(ComboBoxBase<T> control) {
...
+                    if (forwardToTextField(ke)) {
+                        ke.consume();
+                    }
                 }
             }
         });
     }
 
+    // Returns 'true' if the event was consumed.
+    private boolean forwardToTextField(KeyEvent event) {
```
andy-goryachev-oracle: this method seems unnecessary, why not fold it into `EventUtil.dispatch()` including copying and and returning a boolean?

and perhaps consuming the original event?

modules/javafx.base/src/main/java/com/sun/javafx/event/EventUtil.java
beldenfox: EventUtil is in the javafx.base module and can't import anything from javafx.graphics, including Node.

I've moved the utility routine to the skin Utils class.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/skin/Utils.java
```diff
@@ -989,4 +992,14 @@ public static String formatHexString(Color c) {
     public static URL getResource(String str) {
         return Utils.class.getResource(str);
     }
+
+    // Dispatches the event to the Node's dispatcher and returns
+    // true if the event was consumed.
+    public static boolean dispatchToNode(Event event, Node node) {
```
andy-goryachev-oracle, line 998: thank you!
last thing (sorry!) - would it make sense to also describe in the comment that this event bypasses the usual chain, or words to this effect?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxPopupControl.java
beldenfox: My policy is to preserve the old logic as much as possible. There's a bunch of tricky code related to handling ENTER and ESCAPE that I don't want to disturb.

In this case the event is KEY_RELEASED for ENTER. I'm pretty sure the TextField will ignore it and it will never be consumed anyway.

andy-goryachev-oracle: So if the application installs a listener on this textField and consumes the event for the purposes of stopping the dispatch, the original event will continue to propagate, unconsumed?

I view this as a bad decision on the part of fx in general, but then again, it's a larger discussion that should happen elsewhere.

In this case, the argument about preserving the existing logic wins.

andy-goryachev-oracle (approved):
Looks good, thank you!

mstr2:
While this PR seems to fix an actual bug, it also runs the risk of preempting the discussion on focus delegation / multi-level focus that we've had several times, but never reached a conclusion. I wouldn't want this proposed solution (directly dispatching the event to its final target) to be the last word on the topic of focus delegation, especially when it comes to a potential API-level solution.

andy-goryachev-oracle:
@mstr2 :  Unless you can bring up a scenario where the proposed solution breaks things, we should probably keep these two things separate.

Once we reach the agreement, it is very likely that changes will be needed everywhere anyway.  I don't want to get stuck in endless discussions that lead nowhere, I would rather see the platform progressing.

Do you have a possible failure mode?

mstr2:
> @mstr2 : Unless you can bring up a scenario where the proposed solution breaks things, we should probably keep these two things separate.
> 
> Once we reach the agreement, it is very likely that changes will be needed everywhere anyway. I don't want to get stuck in endless discussions that lead nowhere, I would rather see the platform progressing.
> 
> Do you have a possible failure mode?

This depends on the definition of "failure mode". With this solution, an event listener that is installed on a node in the skin subgraph that eventually contains the `TextField` will not see the event (because it does not travel through the subgraph, it is dispatched directly to the final target). This is different from how events in the scene graph usually work.

It's a small inconsistency, and it may be completely acceptable. What I'm saying is that I don't want a narrow-scoped solution as the one presented here to sneak into FX and then just by being there become the de-facto solution for focus-delegation-like problems even when we never conclusively decided how we want to solve the bigger picture.

andy-goryachev-oracle:
Right, I completely agree!
I don't think this method imposes any new standard, we are still free to explore all the alternatives.

beldenfox:
> This depends on the definition of "failure mode". With this solution, an event listener that is installed on a node in the skin subgraph that eventually contains the `TextField` will not see the event (because it does not travel through the subgraph, it is dispatched directly to the final target). This is different from how events in the scene graph usually work.

I thought about this but only in the context of the default skin (I doubt we want to encourage clients to attach event listeners to random parts of the skin we create). I forgot that clients can create their own skins and may want to attach filters and handlers. I could alter the code so the dispatch chain covers the entire skin.

Just to be clear I don't like the solution in this PR. I began prototyping some cleaner approaches but was stymied by the control/skin/behavior split which is particularly murky with ComboBox and Spinner. These skins are also playing games with the way they handle ENTER and ESCAPE which would also need to be re-worked.

Here's a question: if you install a KeyEvent filter on a scene and the user is typing in a ComboBox would you prefer the target of the key event to be the ComboBox or the TextField? In the master branch you would see two events, one targeted at the ComboBox and the other at the TextField. If you had to choose just one which would it be?

This is a general question about how key event forwarding should work. If the answer is ComboBox then this crude solution will be enough until something better comes along. If the answer is TextField then we should probably scrap this PR rather than set a bad precedent.

andy-goryachev-oracle:
These are very good questions!

> if you install a KeyEvent filter on a scene and the user is typing in a ComboBox would you prefer the target of the key event to be the ComboBox or the TextField? 

This is why I don't like the `Event.target` at all.  In my opinion (and I understand I am late to the party and the things are unlikely to change) there should only be one event, delivered to its target (which, in the case of a `KeyEvent`, a current focus owner which is the `TextField`).  If someone adds a listener or a filter somewhere in the hierarchy - fine.

There should be no copying of the event objects for the sole purpose of changing its target.  That's how it works in AWT and Swing.  FX chose a different path of complexity, event clones, ignoring the consumed flag, and happy dispatching of already consumed events.  Which means we'll have much fun dealing with the consequences.

mstr2:
> I thought about this but only in the context of the default skin (I doubt we want to encourage clients to attach event listeners to random parts of the skin we create). I forgot that clients can create their own skins and may want to attach filters and handlers. I could alter the code so the dispatch chain covers the entire skin.
> 
> Just to be clear I don't like the solution in this PR. I began prototyping some cleaner approaches but was stymied by the control/skin/behavior split which is particularly murky with ComboBox and Spinner. These skins are also playing games with the way they handle ENTER and ESCAPE which would also need to be re-worked.

For a bugfix, I'm generally okay with any of the possible approaches: 1) do nothing, 2) send the event directly to the target, or 3) dispatch it through the skin subgraph. Since I don't _like_ any of the solutions, I have no clear preference for or against any of them.

Maran23:
> While this PR seems to fix an actual bug, it also runs the risk of preempting the discussion on focus delegation / multi-level focus that we've had several times, but never reached a conclusion. I wouldn't want this proposed solution (directly dispatching the event to its final target) to be the last word on the topic of focus delegation, especially when it comes to a potential API-level solution.

Same for me, +1. But I think we all agree here on this.

> For a bugfix, I'm generally okay with any of the possible approaches: 1) do nothing, 2) send the event directly to the target, or 3) dispatch it through the skin subgraph. Since I don't like any of the solutions, I have no clear preference for or against any of them.

Also +1. This solutions is at least IMO a bit better than the previous one.
Once again, as I don't have a Mac, I might not be able to test this the way I want to - but still watching this discussion and code.

beldenfox:
> For a bugfix, I'm generally okay with any of the possible approaches: 1) do nothing, 2) send the event directly to the target, or 3) dispatch it through the skin subgraph.

The TextField of a ComboBox or Spinner is a direct child so option 2 and 3 are the same. This assumption is baked into the code in a few places including the FakeFocusTextField.

> I wouldn't want this proposed solution (directly dispatching the event to its final target) to be the last word on the topic of focus delegation, especially when it comes to a potential API-level solution.

I've entered draft PR #2181 which sketches out an alternative solution for focus delegation. It shows how a control class can delegate events as a drop-in solution; it doesn't require updating Scene or Node in any way to work.

beldenfox:
Is there anything more I need to do to get this PR approved? Beyond creating a comprehensive focus-delegation framework from scratch?

I'm not in a hurry to get this into JavaFX 27. These controls have all sorts of oddball behavior and bugs and developers have had to resort to some interesting and fragile workarounds. Based on what I'm seeing in the bug database it looks like any time we touch ComboBox or Spinner we break some client's filtering routine and I'd prefer to deal with that at the start of a release cycle rather than the end.

andy-goryachev-oracle:
> Is there anything more I need to do to get this PR approved?

You need two approvals, and possibly some general consensus.  In this case, @mstr2 voiced some concerns re: general purpose focus framework.  Generally, I don't think the desire for better solution should preempt small incremental improvements (since it risks completely stalling the progress).

As @mstr2 correctly pointed out, some of the PRs linger for a very long time, often one approval short of integration.  If everyone would review one PR before submitting their own, things might go differently, but alas.  Any ideas?

mstr2 (approved):
I'll approve with the condition that no one is allowed to bring this up as "it's fixed, why bother" when we'll have our next annual discussion on a general-purpose focus/event delegation mechanism.

kevinrushforth:
@beldenfox This is ready for you to integrate.

PR #2167, opened 2026-05-13
8203870: ProxyBuilder cannot handle Read-Only List Properties properly

stechy1:
ProxyBuilder is used by the FXML loader to instantiate classes whose constructors are annotated with `@NamedArg`. When such a class also exposes a read-only `Map` property (e.g. getProperties() — the same pattern used by `javafx.scene.Node`), setting child elements under that property in FXML caused incorrect behaviour or a runtime error.

**Root cause:** _getReadOnlyProperty()_ always returned an `ArrayListWrapper` regardless of the actual **getter** return type. When the getter returns a `Map`, an `ArrayListWrapper` is the wrong container and the entries are never transferred to the real map on the object.

---------
### Progress

### Issue
 * [JDK-8203870](https://bugs.openjdk.org/browse/JDK-8203870): ProxyBuilder cannot handle Read-Only List Properties properly (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2167/head:pull/2167` \
`$ git checkout pull/2167`

`$ git checkout pull/2167` \
`$ git pull https://git.openjdk.org/jfx.git pull/2167/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2167`

`$ git pr show -t 2167`

</details>

<a href="https://git.openjdk.org/jfx/pull/2167.diff">https://git.openjdk.org/jfx/pull/2167.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2167#issuecomment-4557895977)
</details>


Maran23:
do you have a minimal reproducer? In pure Java, so I can do some tests and if I can reproduce create a ticket?

stechy1:
Hello,

I sent a reproducer to a mailing list, but I can attach it here if it's allowed...

stechy1:
Find a reproducer here: [javafx_bug_report.tar.gz](https://github.com/user-attachments/files/28173230/javafx_bug_report.tar.gz).

Maran23:
@stechy1 I found this issue: https://bugs.openjdk.org/browse/JDK-8203870
It sounds very similar to your issue. Can you confirm it? There are more details in the comments as well.

If it is the same problem, can you also add a test case like in the ticket? And rename the PR title to fit the scheme we use here (see also the contributing guidelines).

stechy1:
Hello @Maran23,

I just checked the ticket issue and my fix does not solve it. I can either take a closer look and include a fix also for that one otherwise new issue will be necessary.

Maran23:
> I just checked the ticket issue and my fix does not solve it. I can either take a closer look and include a fix also for that one otherwise new issue will be necessary.

Ok I see, but I wonder if it is related. It sounds like it is, same pattern with `@NamedArg` and a readonly Collection. So I think a fix should address not just a readonly Map, but also List I think. In the comments, someone mentioned that this was a regression, so it might be worth to check that out as well - and what happens if the code is reverted

stechy1:
> > I just checked the ticket issue and my fix does not solve it. I can either take a closer look and include a fix also for that one otherwise new issue will be necessary.
> 
> Ok I see, but I wonder if it is related. It sounds like it is, same pattern with `@NamedArg` and a readonly Collection. So I think a fix should address not just a readonly Map, but also List I think. In the comments, someone mentioned that this was a regression, so it might be worth to check that out as well - and what happens if the code is reverted

It seems to be related because of the usage of `@NamedArg` annotation, but that's it. Thanks to this annotation the instantiation of properties is handled by "ProxyBuilder" class. But here the relation ends.

I have found a solution for that regression. Do you want me to describe it here with possible solution?

Maran23:
> I have found a solution for that regression. Do you want me to describe it here with possible solution?

Feel free to create a PR for that one as well. That will help a lot to fully understand the full picture around `ProxyBuilder`, not just for me but other reviewers as well.

stechy1:
The fix for the bug you mentioned could be done in the same place I already changed in my first commit. So I decided to include it in this PR as well.

When the constructor is annotated by `@NamedArg`, the `ProxyBuilder` class will handle instantiating. Container support for properties was added but with limitations described in the bug ticket. Every container was treated as [`ArrayListWrapper`](https://github.com/openjdk/jfx/blob/c10ff9f3b67e760d15dbc218cdc132d5d7d63dd2/modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java#L179). This had two consequences: Map properties of the Node were ignored completely and List container was initialized only with a [single value](https://github.com/openjdk/jfx/blob/c10ff9f3b67e760d15dbc218cdc132d5d7d63dd2/modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java#L456).

```java
    private Object getUserValue(String key, Class<?> type) {
        Object val = userValues.get(key);
        if (val == null) {
            return null;
        }

        if (type.isAssignableFrom(val.getClass())) {
            return val;
        }
        ...
```

The type check is crutial for the fix. Up until now, the `type` was always `ArrayListWrapper`, but `val.getClass()` was usually `ObservableList` or `ObservableMap`. Those two types never matched and so few lines later:

```java
        if (ArrayListWrapper.class.equals(val.getClass())) {
            // user given value is an ArrayList but the constructor doesn't
            // accept an ArrayList so the ArrayList comes from
            // the getTemporaryContainer method
            // we take the first argument
            List l = (List) val;
            return l.get(0);
        }
```

By introspecting the needs for return type in the method `getReadOnlyProperty` we can use correct collection so later type will match.

Maran23:
Thanks, will take a closer look soon. Note that your PR body does not exactly match the specification for the AI line. Can you fix that?

Questions:
- Can we write a test for the Readonly List case as well?
- We should probably also check `ObservableSet`, right? So that we handle all collections. If we should do that, can you write a test as well? Would help a lot!

stechy1:
Further I'm going with tests more issues I'm discovering.
So far I fixed creating elements for "standard" and observable collections (`List`, `Set`, `Map`). But I find out that `Arrays` also does not work. It is questionable whether to fix them in this PR as well or leave to another one. It might need bigger changes.

modules/javafx.fxml/src/test/resources/test/javafx/fxml/rt_8203870_a.fxml
```diff
@@ -0,0 +1,55 @@
...
+    <observableMap>
+        <key1>
+            <String fx:value="map_value_1" />
+        </key1>
+        <key2>
+            <String fx:value="mav_value_2" />
+        </key2>
+    </observableMap>
+
+</ClassWithCollections>
```
andy-goryachev-oracle: minor: missing newline (also in ..._b.fxml)

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
```diff
@@ -38,12 +38,17 @@
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
```
andy-goryachev-oracle, line 41: please update the copyright year in all the modified files, as per tradition.

modules/javafx.fxml/src/main/java/javafx/fxml/FXMLLoader.java
```diff
@@ -123,7 +124,7 @@ public boolean isCollection() {
             // Return true if value is a list, or if the value's type defines
             // a default property that is a list
             boolean collection;
-            if (value instanceof List<?>) {
+            if (value instanceof List<?> || value instanceof Set<?>) {
```
andy-goryachev-oracle: shouldn't this be `instanceof Collection` then?
(and I don't think generics are needed for `instanceof` since it's runtime check and type is erased)

modules/javafx.fxml/src/test/java/test/com/sun/javafx/fxml/builder/ClassWithReadOnlySet.java
```diff
@@ -0,0 +1,33 @@
+package test.com.sun.javafx.fxml.builder;
```
andy-goryachev-oracle, line 26: missing copyright header (here and in some other files)

modules/javafx.fxml/src/test/java/test/javafx/fxml/RT_8203870Test.java
```diff
@@ -0,0 +1,47 @@
+package test.javafx.fxml;
+
+import javafx.fxml.FXMLLoader;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class RT_8203870Test {
```
andy-goryachev-oracle: We don't use RT-* prefix anymore; typically, we try to make the names more descriptive, like `TestFXMLoaderReadOnlyProperties` or something like that, your call.  Same goes for the resource files, probably.

modules/javafx.fxml/src/test/java/test/javafx/fxml/ReadOnlyCollectionsCreationTest.java
```diff
@@ -0,0 +1,47 @@
...
+        assertEquals(2, widget.getMap().size());
+
+        assertEquals(3, widget.getObservableList().size());
+        assertEquals(3, widget.getObservableSet().size());
+        assertEquals(2, widget.getObservableMap().size());
+
+//        assertEquals(3, widget.getNames().length);
+//        assertEquals(3, widget.getRatios().length);
+    }
+
```
andy-goryachev-oracle, line 66: minor: extra newlines

modules/javafx.fxml/src/test/java/test/javafx/fxml/ReadOnlyCollectionsCreationTest.java
```diff
@@ -0,0 +1,47 @@
...
+        assertEquals(3, widget.getList().size());
+        assertEquals(3, widget.getSet().size());
+        assertEquals(2, widget.getMap().size());
+
+        assertEquals(3, widget.getObservableList().size());
+        assertEquals(3, widget.getObservableSet().size());
+        assertEquals(2, widget.getObservableMap().size());
+
+//        assertEquals(3, widget.getNames().length);
```
andy-goryachev-oracle: is there a reason to keep commented out lines?

modules/javafx.fxml/src/test/resources/test/javafx/fxml/readonly_collections_creation_a.fxml
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
andy-goryachev-oracle, line 1: needs a copyright header (check other .fxml files for an example)

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
```diff
@@ -180,12 +185,44 @@ private static class ArrayListWrapper<T> extends ArrayList<T> {
...
     }
 
-    // This is used to support read-only collection property.
+    // This is used to support read-only collection/map property.
     private Object getReadOnlyProperty(String propName) {
-        // return ArrayListWrapper now and convert it to proper type later
-        // during the build - once we know which constructor we will use
-        // and what types it accepts
-        return new ArrayListWrapper<>();
+        Method getter = findGetter(propName);
```
andy-goryachev-oracle, line 187: I wonder if this is the right place for this logic.

It looks like the `ArrayListWrapper` is used as a marker container type (the comment seems to be still valid), and the actual logic is handled in `getUserValue()`.  I wonder if even an AtomicReference or a simple holder type would work just as well, with the logic handling different types moved to where it's actually used?

(this private method getReadOnlyProperty() seems completely unnecessary in the first place)

andy-goryachev-oracle (commented):
Some preliminary code review comments, I did not test the solution (yet).

modules/javafx.fxml/src/main/java/javafx/fxml/FXMLLoader.java
Maran23: Without `<?>`, we will usually get a warning. I think its okay to keep. But I'm also wondering if we should check for the generic `Collection` class

modules/javafx.fxml/src/test/java/test/javafx/fxml/RT_8203870Test.java
Maran23: +1, otherwise its hard to guess what happens. If you want you can still write javadoc and reference the JDK ticket

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
Maran23: Mostly wondering the same - why do we have `ArrayListWrapper` but no wrapper classes for the other collections? Maybe we can unify / simplify it more.

I did some tests with `@NamedArg` and it really seems to be buggy with more advanced usecases like this one

modules/javafx.fxml/src/main/java/javafx/fxml/FXMLLoader.java
andy-goryachev-oracle: just curious - what kind of warning?  My Eclipse shows no warnings, but then again, I have some disabled.

Maran23: It shows me this one:

image

andy-goryachev-oracle: ah, right - I have this disabled as it's not really useful.  we've got over 6,000 such warnings in the code base.
thanks!

modules/javafx.fxml/src/test/java/test/javafx/fxml/ReadOnlyCollectionsCreationTest.java
stechy1: Yes, there is a reason. During testing, I find out that array creation in combination of  @NamedArg also does not work. When I was investigating why and trying to find a solution for it, it would require a bit more digging and I'm not sure whether to do it as part of this PR or a new one (I would preffer new one, because I expect bigger changes).

modules/javafx.fxml/src/test/java/test/javafx/fxml/RT_8203870Test.java
stechy1: Understand, I will rename it to proper name.

modules/javafx.fxml/src/main/java/javafx/fxml/FXMLLoader.java
stechy1: I will take a look if `Collection` check would be enough

stechy1: After some investigation I can say there should be no harm to simplify the `instanceof` check to `Collection` in general.

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
Maran23: @stechy1 do you have an idea about this?
Since you are most likely the person that knows the most about this logic. Will help us a lot to review the change

modules/javafx.fxml/src/main/java/javafx/fxml/FXMLLoader.java
```diff
@@ -144,25 +145,25 @@ public void add(Object element) {
...
+            if (value instanceof Collection<?>) {
+                collection = (Collection<Object>) value;
             } else {
                 Class<?> type = value.getClass();
                 DefaultProperty defaultProperty = type.getAnnotation(DefaultProperty.class);
                 String defaultPropertyName = defaultProperty.value();
 
                 // Get the list value
-                list = (List<Object>)getProperties().get(defaultPropertyName);
+                collection = (List<Object>)getProperties().get(defaultPropertyName);
```
andy-goryachev-oracle: should it be `(Collection<Object>)` then?

modules/javafx.fxml/src/main/java/javafx/fxml/FXMLLoader.java
```diff
@@ -144,25 +145,25 @@ public void add(Object element) {
...
 
                 // Coerce the element to the list item type
```
andy-goryachev-oracle, line 159: collection item type?

modules/javafx.fxml/src/main/java/javafx/fxml/FXMLLoader.java
```diff
@@ -144,25 +145,25 @@ public void add(Object element) {
             // If value is a list, add element to it; otherwise, get the value
```
andy-goryachev-oracle: is a collection?

modules/javafx.fxml/src/main/java/javafx/fxml/FXMLLoader.java
```diff
@@ -484,7 +485,7 @@ private void populateListFromString(
             // Split the string and add the values to the list
             List<Object> list = (List<Object>)valueAdapter.get(listPropertyName);
```
andy-goryachev-oracle, line 486: should it be `Collection<Object>` ?

modules/javafx.fxml/src/test/java/test/com/sun/javafx/fxml/builder/ClassWithReadOnlySet.java
andy-goryachev-oracle, line 26: The standard copyright header is still missing.

modules/javafx.fxml/src/test/java/test/javafx/fxml/ClassWithDefaultListProperty.java
```diff
@@ -0,0 +1,17 @@
+package test.javafx.fxml;
```
andy-goryachev-oracle, line 26: missing copyright header - here and in other files

modules/javafx.fxml/src/test/java/test/javafx/fxml/ReadOnlyCollectionsCreationTest.java
andy-goryachev-oracle: please create a follow-up JBS ticket

modules/javafx.fxml/src/test/resources/test/javafx/fxml/class_with_default_list_property.fxml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<?import test.javafx.fxml.ClassWithDefaultListProperty?>
+
+<ClassWithDefaultListProperty>one,two,three</ClassWithDefaultListProperty>
```
andy-goryachev-oracle: minor: missing newline

modules/javafx.fxml/src/test/resources/test/javafx/fxml/class_with_default_set_property.fxml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<?import test.javafx.fxml.ClassWithDefaultSetProperty?>
+
+<ClassWithDefaultSetProperty>one,two,three</ClassWithDefaultSetProperty>
```
andy-goryachev-oracle: minor: missing newline (here and in some other files).  you can see it in the github diff view.

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
```diff
@@ -180,12 +185,44 @@ private static class ArrayListWrapper<T> extends ArrayList<T> {
...
+            }
+
+            if (Map.class.isAssignableFrom(retType)) {
+                return new LinkedHashMap<>();
+            }
+            if (Set.class.isAssignableFrom(retType)) {
+                return new LinkedHashSet<>();
+            }
+            if (Collection.class.isAssignableFrom(retType)) {
+                return new ArrayListWrapper<>();
```
andy-goryachev-oracle: should it be called `CollectionWrapper` ?

or, why have this type at all - the `Set` case returns a `LinkedHashSet` instance, so perhaps this case should return `LinkedList` maybe?

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
```diff
@@ -180,12 +185,44 @@ private static class ArrayListWrapper<T> extends ArrayList<T> {
...
+
+            if (ObservableList.class.isAssignableFrom(retType)) {
+                return FXCollections.observableArrayList();
+            }
+            if (ObservableSet.class.isAssignableFrom(retType)) {
+                return FXCollections.observableSet();
+            }
+            if (ObservableMap.class.isAssignableFrom(retType)) {
+                return FXCollections.observableHashMap();
+            }
```
andy-goryachev-oracle, line 199: Question: the `javafx.collection` package also declares an `ObservableArray` class, should it be handled here?

andy-goryachev-oracle (changes requested):
Thank you for adding tests!

Here some preliminary code review only comments.

modules/javafx.fxml/src/main/java/javafx/fxml/FXMLLoader.java
stechy1: I see this as a one special case. This is designed specifically for handling coma separated values and based on the type where it will be assigned it use `Array` or a `List`.

So in my opinion I would leave this untouched.

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
stechy1: I'm not an expert on this. I just start digging into this because `@NamedArg` was not working as expected. The author of this piece of code must had a reason to write it in this way.

`getReadOnlyProperty(propName)` - inspects the getter's return type and creates a temporary container of the appropriate type (`ObservableList` , `ObservableSet` , `Map` , `Set` , or `ArrayListWrapper` for a plain  `Collection` ). Returns `null`  if the property is not a read-only collection/map.

This method is called here:

```java
    /**
     * This is used to support read-only collection property. This method must
     * return a Collection of the appropriate type if 1. the property is
     * read-only, and 2. the property is a collection. It must return null
     * otherwise.
     *
     */
    private Object getTemporaryContainer(String propName) {
        Object o = containers.get(propName);
        if (o == null) {
            o = getReadOnlyProperty(propName);
            if (o != null) {
                containers.put(propName, o);
            }
        }
        return o;
    }
```

---

Methods `getTemporaryContainer()` and `getUserValue()` are in some kind of relationship. We can look on this as two phases:

# Phase 1 - Parsing / accumulation
 
ProxyBuilder  extends  Map<String, Object> . While the FXML loader is parsing child elements, it calls the standard  Map  methods  containsKey(key)  and  get(key)  on the builder. Both of those override methods delegate directly to  getTemporaryContainer :

```java
@Override
public boolean containsKey(Object key) {
    return (getTemporaryContainer(key.toString()) != null);
}

@Override
public Object get(Object key) {
    return getTemporaryContainer(key.toString());   // ← returns the temp container
}
```

The FXML loader gets back the temporary container (e.g.  `ArrayListWrapper` ,  `ObservableList` …), and populates it with the child elements.

# Phase 2 - Build / resolution

When FXML parsing is done, the loader calls `build()`. At that point:

1. All accumulated _containers_ are flushed into `userValues`.
2. The builder tries to find the best matching constructor / setters.
3. `getUserValue` is called for each parameter/property to retrieve the final, correctly-typed value - including the  `ArrayListWrapper` → scalar unwrapping logic.

modules/javafx.fxml/src/test/resources/test/javafx/fxml/class_with_default_list_property.fxml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
andy-goryachev-oracle, line 1: please make sure *all* new files have the copyright header

andy-goryachev-oracle (changes requested):
missing headers

modules/javafx.fxml/src/test/resources/test/javafx/fxml/class_with_default_list_property.fxml
```diff
@@ -1,4 +1,28 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!--
```
Maran23, line 2: this should be above: `<?xml version="1.0" encoding="UTF-8"?>`

stechy1: Take a look around other XML files. The header is after that line everywhere. I can change it if reálný necessary

andy-goryachev-oracle: I think the first line in xml identifies the format and the encoding, the  header should follow after that.

mstr2: > I think the first line in xml identifies the format and the encoding, the header should follow after that.

This is correct. The most recent [XML specification](https://www.w3.org/TR/xml/) says:
```
document ::= prolog element Misc*
prolog ::= XMLDecl? Misc* (doctypedecl Misc*)?
XMLDecl ::= '<?xml' VersionInfo EncodingDecl? SDDecl? S? '?>'
Misc ::= Comment | PI | S
```

This means that the prolog of an XML document _may_ start with an `<?xml ... ?>` declaration (actually, the spec says it _should_, but it's optional). If it does, it must be the very first thing in the document; comments are only allowed after it.

Maran23: interesting - I just checked and indeed the same pattern is used in the [JDK](https://github.com/openjdk/jdk/blob/master/make/langtools/build.xml). The only exception is the [pom.xml](https://github.com/openjdk/jdk/blob/master/src/utils/LogCompilation/pom.xml) sometimes.

EDIT: Thanks Michael for the link, that explains it.

TIL. So just ignore my comment 😄

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
andy-goryachev-oracle: `ObservableArray` descendants are used in `TriangleMesh`.

stechy1: All right, I will check it and if possible add more tests with `ObservableArray`.

andy-goryachev-oracle (approved):
Looks good, thank you for developing tests!

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
Maran23: Maybe something we could check in a follow up? 
If I understand the author correctly, it is not 100% clear if this is needed or not

modules/javafx.fxml/src/test/java/test/com/sun/javafx/fxml/builder/ProxyBuilderTest.java
```diff
@@ -246,6 +250,204 @@ public void testReadOnlyList() {
...
+    @Test
+    public void testArrayListWrapperUnwrapsFirstElementForScalarArg() {
+        ProxyBuilder pb = new ProxyBuilder(ClassWithPlainCollectionAndScalarArg.class);
+
+        // Simulate what FXMLLoader does: containsKey triggers wrapper creation,
+        // get() returns it, and then the child element is added to the container.
+        assertTrue(pb.containsKey("child"),
+                "containsKey must be true for the plain-List getter");
+
+        @SuppressWarnings("unchecked")
```
Maran23: No need to keep those (in my opinion, those just clutter the code without any real benefit)

Maran23:
I will review and test this later.

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
```diff
@@ -447,8 +490,33 @@ private Object getUserValue(String key, Class<?> type) {
...
 
+        // ObservableIntegerArray needs special treatment
+        if (ObservableIntegerArray.class.isAssignableFrom(type)) {
+            if (val instanceof int[] primitiveInts) {
+                return FXCollections.observableIntegerArray(primitiveInts);
+            }
+            List<?> list = (List<?>) val;
+            int[] ints = new int[list.size()];
+            for (int idx = 0; idx < list.size(); idx++) {
+                ints[idx] = BeanAdapter.coerce(list.get(idx), Integer.class).intValue();
```
Maran23: `.intValue()` is not needed

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
```diff
@@ -447,8 +490,33 @@ private Object getUserValue(String key, Class<?> type) {
...
+
+        // ObservableFloatArray needs special treatment
+        if (ObservableFloatArray.class.isAssignableFrom(type)) {
+            if (val instanceof float[] primitiveFloats) {
+                return FXCollections.observableFloatArray(primitiveFloats);
+            }
+            List<?> list = (List<?>) val;
+            float[] floats = new float[list.size()];
+            for (int idx = 0; idx < list.size(); idx++) {
+                floats[idx] = BeanAdapter.coerce(list.get(idx), Float.class).floatValue();
```
Maran23: `.floatValue()` is not needed

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/BeanAdapter.java
```diff
@@ -652,12 +652,12 @@ public static Type getGenericType(Class<?> sourceType, String key, Class<?> targ
     }
 
     /**
-     * Determines the type of a list item.
+     * Determines the type of a list or set item.
```
Maran23: Determines the type of a `Collection`.

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
```diff
@@ -576,6 +649,39 @@ public void invoke(Object obj, Object argStr) throws Exception {
...
+            super(m, t);
+        }
+
+        @Override
+        public void invoke(Object obj, Object argStr) throws Exception {
+            // we know that this.method returns a Map otherwise it wouldn't be here
+            Map to = (Map) ModuleHelper.invoke(method, obj, new Object[]{});
+            if (argStr instanceof Map) {
+                to.putAll((Map) argStr);
+            }
```
Maran23:
```suggestion
            Map<?, ?> to = (Map<?, ?>) ModuleHelper.invoke(method, obj, new Object[]{});
            if (argStr instanceof Map argMap) {
                to.putAll(argMap);
            }
```

avoid most of the warnings-

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
```diff
@@ -520,6 +588,11 @@ private Map<String, Property> scanForSetters() {
                     Class<?> argType[] = m.getParameterTypes();
                     if (Collection.class.isAssignableFrom(retType) && argType.length == 0) {
                         strsMap.put(propName, new Getter(m, retType));
+                    } else if (Map.class.isAssignableFrom(retType) && argType.length == 0
+                            && !strsMap.containsKey(propName)) {
```
Maran23: Why is this condition needed (only here?)?

modules/javafx.fxml/src/test/java/test/javafx/fxml/ClassWithDefaultPropertyTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.jupiter.api.Test;
+
+public class ClassWithDefaultPropertyTest {
+
+    @Test
+    public void testDefaultListProperty() throws Exception {
+        FXMLLoader fxmlLoader = new FXMLLoader(getClass().getResource("class_with_default_list_property.fxml"));
+        ClassWithDefaultListProperty classWithDefaultProperty = fxmlLoader.load();
+
+        Assertions.assertEquals(1, classWithDefaultProperty.getItems().size());
```
Maran23: I would rather recommend to do `import static org.junit.jupiter.api.Assertions.*;`
and remove `Assertions` from here (also makes this shorter).

modules/javafx.fxml/src/test/java/test/javafx/fxml/ClassWithDefaultPropertyTest.java
```diff
@@ -0,0 +1,54 @@
...
+//        Coma separated values are currently not supported in default list properties
+//        Assertions.assertEquals(3, classWithDefaultProperty.getItems().size());
```
Maran23: maybe remove? I don't think this is something that can be supported without breaking compatibility.

Maran23 (commented):
some comments and suggestions, 1 question, looks good otherwise!

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
Maran23: I replaced `ArrayListWrapper` with a plain `ArrayList` and everything still works and is green.

Maran23: trying out `ClassWithPlainCollectionAndScalarArg` with a `Set` like this:

```
public class ClassWithPlainCollectionAndScalarArg {

    public final String child;
    private final Set<String> childList = new HashSet<>();

    public ClassWithPlainCollectionAndScalarArg(@NamedArg("child") String child) {
        this.child = child;
    }

    public Set<String> getChild() {
        return childList;
    }
}
```

the test fail again because this is not handled by `ArrayListWrapper`.

Which makes we wonder, should we instead do the following in `ProxyBuilder` (L518-524):

```
        if (Collection.class.isAssignableFrom(val.getClass())) { // instead of ArrayListWrapper.class.equals(val.getClass())
            Collection<?> col = (Collection<?>) val;
            return col.iterator().next();
        }
```

and return a normal `ArrayList` instead of the wrapper (L206-218):

```
            if (ObservableArray.class.isAssignableFrom(retType)) {
                return new ArrayList<>(); // instead of ArrayListWrapper
            }

            if (Map.class.isAssignableFrom(retType)) {
                return new LinkedHashMap<>();
            }
            if (Set.class.isAssignableFrom(retType)) {
                return new LinkedHashSet<>();
            }
            if (Collection.class.isAssignableFrom(retType)) {
                return new ArrayList<>(); // instead of ArrayListWrapper
            }
```

all tests are green, no `ArrayListWrapper` needed and the `HashSet` case is also fixed.

Maran23: I created https://github.com/openjdk/jfx/pull/2209 with the changes I described here so you can check the diff.

This is the commit with the changes: https://github.com/openjdk/jfx/pull/2209/changes/c1f95fa29de540e10c143d101ac21e586a008a10

modules/javafx.fxml/src/test/java/test/com/sun/javafx/fxml/builder/ProxyBuilderTest.java
stechy1: You are right.  The `containsKey` method only creates a temporary contaier which is also created by `pb.get` method.

modules/javafx.fxml/src/test/java/test/javafx/fxml/ClassWithDefaultPropertyTest.java
stechy1: Your call. I can remove tests for this case.
I thought it would be great at least to have a test for this case since it was going through the code I was modifying.

Maran23: Not the test, just the commented out `// Assertions.assertEquals(3, classWithDefaultProperty.getItems().size());`

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
stechy1: I like this new approach. We can get rid of the `ArrayListWrapper` and it will still works as expected. Since you have created another PR, what should I do now with mine? I can apply same changes (removing `ArrayListWrapper`) or close this one and keep yours.

Maran23: You can apply or cherrypick my changes / commit if you like.
I just opened my PR to better outline the diff and to show that the all tests are green.

I will close my PR as soon as you applied the changes :)

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
stechy1: Thanks to this condition it is possible to put all values of a `Map` defined inside FXML into a concrete `Map` instance in class. `Map` is not a subtype of a `Collection` and same is for `ObservableArray`.

modules/javafx.fxml/src/test/java/test/com/sun/javafx/fxml/builder/ClassWithPlainListAndScalarArg.java
```diff
@@ -0,0 +1,63 @@
...
+
+package test.com.sun.javafx.fxml.builder;
+
+import javafx.beans.NamedArg;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Fixture class for testing the {@code ArrayListWrapper} unwrapping path in
```
Maran23: One more thing: We should remove the reference of `ArrayListWrapper` in the javadoc here and the other test classes, since we removed it

modules/javafx.fxml/src/test/java/test/com/sun/javafx/fxml/builder/ProxyBuilderTest.java
```diff
@@ -335,11 +335,6 @@ public void testReadOnlyMapWithNamedArg() {
     public void testArrayListWrapperUnwrapsFirstElementForScalarArg() {
         ProxyBuilder pb = new ProxyBuilder(ClassWithPlainCollectionAndScalarArg.class);
 
-        // Simulate what FXMLLoader does: containsKey triggers wrapper creation,
-        // get() returns it, and then the child element is added to the container.
-        assertTrue(pb.containsKey("child"),
```
Maran23: I think I confused you here - sorry. This was fine, you can keep it if you like. What I meant with my comment is that I think we can remove `@SuppressWarnings("unchecked")` below because it does not really help

Maran23:
@stechy1 Also note that the tests do not run in your fork. You need to enable Github Actions.

See here: https://github.com/openjdk/jfx/blob/master/CONTRIBUTING.md#submitting-your-changes-via-a-pull-request

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
Maran23: What I meant is only the condition `!strsMap.containsKey(propName)`, was just wondering because it is only done here and not for the other types

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
stechy1: Changes were overtaken. Thank you for your time and review of the PR.

modules/javafx.fxml/src/main/java/com/sun/javafx/fxml/builder/ProxyBuilder.java
stechy1: By "all other types" you mean `Set` and `List`? They are covered by `Collection` check above.

Maran23: What I mean is, we have the following conditions:
- `Collection.class.isAssignableFrom(retType) && argType.length == 0`
- `Map.class.isAssignableFrom(retType) && argType.length == 0 && !strsMap.containsKey(propName)`
- `ObservableArray.class.isAssignableFrom(retType) && argType.length == 0`

And I was wondering why only for the `Map` if condition we have the additional `!strsMap.containsKey(propName)` check

stechy1: Oh I see it now. It is not necessary. I think it might be just a relic of me testing whether it will perform a bit better with additional check. So no special need for this. Can be removed.

modules/javafx.fxml/src/test/java/test/com/sun/javafx/fxml/builder/ProxyBuilderTest.java
Maran23: must be removed so that the test will run

modules/javafx.fxml/src/test/java/test/javafx/fxml/ReadOnlyCollectionsCreationTest.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Test
+    public void testArrayListUnwrapsFirstElementViaFxml() throws IOException {
+        FXMLLoader fxmlLoader = new FXMLLoader(getClass().getResource("array_list_scalar_arg.fxml"));
+        ClassWithPlainListAndScalarArg result = fxmlLoader.load();
+
+        assertEquals("hello", result.child,
+                "FXMLLoader must unwrap the single ArrayList element and pass it as the scalar @NamedArg constructor argument");
+
```
Maran23: remove the empty newline

modules/javafx.fxml/src/test/resources/test/javafx/fxml/readonly_collections_creation_b.fxml
```diff
@@ -0,0 +1,80 @@
...
+Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+or visit www.oracle.com if you need additional information or have any
+questions.
+-->
+
+<?import java.lang.String?>
+<?import test.javafx.fxml.ClassWithCollections2?>
+<ClassWithCollections2 xmlns:fx="http://javafx.com/fxml"
+                      id="RT_8203870">
+<!--                      names="a, b, c"-->
```
Maran23: I think we should remove the commented out code. I don't think that comma separated values are supported. 
In XML (and FXML), there is no such things as comma separated values equals List/Array elements. 

So I would say lets remove and rather discuss this in the mailing list after this PR is merged.

modules/javafx.fxml/src/test/resources/test/javafx/fxml/readonly_collections_creation_a.fxml
```diff
@@ -0,0 +1,79 @@
...
+
+Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+or visit www.oracle.com if you need additional information or have any
+questions.
+-->
+
+<?import java.lang.String?>
+<?import test.javafx.fxml.ClassWithCollections?>
+<ClassWithCollections xmlns:fx="http://javafx.com/fxml">
+<!--                      names="a, b, c"-->
```
Maran23: same here, remove commented out code

modules/javafx.fxml/src/test/java/test/javafx/fxml/ReadOnlyCollectionsCreationTest.java
Maran23: As I commented in the FXML files already, I would suggest we remove the commented out code because comma separated values were never supported as List / Array. So we should rather discuss this behavior and remove the commented out code for now.

modules/javafx.fxml/src/test/java/test/com/sun/javafx/fxml/builder/ProxyBuilderTest.java
```diff
@@ -108,29 +114,29 @@ public void testImmutableTwoConstructorsWithSameArgNames() {
         pb.put("a", 123);
         pb.put("b", 456);
         ImmutableClass result = (ImmutableClass) pb.build();
-        assertEquals(123, result.a, 1e-10);
-        assertEquals(456, result.b, 1e-10);
+        assertEquals(123, result.a, 0.00001);
```
Maran23: I just saw that below, you were using `1e-5f`, lets use it here and below as well
```suggestion
        assertEquals(123, result.a, 1e-5f);
```

Maran23 (commented):
Looks good. The test is still `@Disabled`, that need to be fixed and I suggest to remove the commented out code, because comma separated values were never supported by FXML so I can't see how this is a bug. More details inline.

modules/javafx.fxml/src/test/java/test/com/sun/javafx/fxml/builder/ProxyBuilderTest.java
stechy1: I have changed this because you asked for change in some other comment. So should I put `1e-5f` there?

Maran23: Yes I said it in another comment. But I reviewed it again and saw that you were using `1e-5f` in the other tests, so I think we should use it here as well to be consistent.

modules/javafx.fxml/src/test/java/test/com/sun/javafx/fxml/builder/ClassWithPlainSetAndScalarArg.java
```diff
@@ -0,0 +1,63 @@
...
+    private final Set<String> childList = new HashSet<>();
+
+    public ClassWithPlainSetAndScalarArg(@NamedArg("child") String child) {
+        this.child = child;
+    }
+
+    public Set<String> getChild() {
+        return childList;
+    }
+}
```
andy-goryachev-oracle: missing newline

modules/javafx.fxml/src/test/java/test/com/sun/javafx/fxml/builder/ProxyBuilderTest.java
```diff
@@ -332,24 +332,37 @@ public void testReadOnlyMapWithNamedArg() {
...
 
-        ClassWithPlainCollectionAndScalarArg result =
-                (ClassWithPlainCollectionAndScalarArg) pb.build();
+        ClassWithPlainSetAndScalarArg result = (ClassWithPlainSetAndScalarArg) pb.build();
 
         assertEquals("expectedValue", result.child,
-                "ProxyBuilder must unwrap the first ArrayListWrapper element "
-                + "and pass it as the scalar @NamedArg constructor argument");
+                "ProxyBuilder must unwrap the first temporary ArrayList element "
+                        + "and pass it as the scalar @NamedArg constructor argument");
```
andy-goryachev-oracle: minor: some funky indentation.

also, we probably should avoid concatenation in tests - these run million times over and over, consuming electricity ;-)  a shorter distinct message might suffice.

modules/javafx.fxml/src/test/resources/test/javafx/fxml/hash_set_scalar_arg.fxml
```diff
@@ -0,0 +1,41 @@
...
+    The @NamedArg constructor parameter "child" has type String (a scalar), so
+    getUserValue must detect the HashSet and extract its first element.
+-->
+<?import java.lang.String?>
+<?import test.com.sun.javafx.fxml.builder.ClassWithPlainSetAndScalarArg?>
+<ClassWithPlainSetAndScalarArg xmlns:fx="http://javafx.com/fxml">
+    <child>
+        <String fx:value="hello"/>
+    </child>
+</ClassWithPlainSetAndScalarArg>
```
andy-goryachev-oracle: missing newline

modules/javafx.fxml/src/test/java/test/com/sun/javafx/fxml/builder/ProxyBuilderTest.java
```diff
@@ -420,7 +419,7 @@ public void testReadOnlyObservableFloatArray() {
     }
 
     @Test
-    public void testReadOnlyObservableIntegerArrayWithNamedArg() {
+    public void testReadOnlyObservableIntegerArrayWxithNamedArg() {
```
andy-goryachev-oracle: oops, `Wxith`

andy-goryachev-oracle (approved):
LGTM.

Maran23 (approved):
Looks good, tested with various applications.

kevinrushforth:
@stechy1 This is ready for you to integrate.

PR #2168, opened 2026-05-15
8379561: Remove UNICODE related code from GLib

sashamatveev:
- Removed/disabled coded which uses `gunichartables.h`. We do not use it and do not need.
- Only dependency on this code is `GVariant` which uses `g_unichar_isprint` to determine if character can be printed or needs to be escaped. This code is not being used currently, but might be in the future. Assert which validates `GVariant` string uses it to log value of `GVariant` in case if assert fails. We might need it someday, so I keep it. Code was updated to use if ascii is printable instead, since we do not use any unicode in our code. In worst case we will escape non-ascii characters in debug log.
- I tested on all platforms with all supported formats.

---------

---------
### Progress

### Issue
 * [JDK-8379561](https://bugs.openjdk.org/browse/JDK-8379561): Remove UNICODE related code from GLib (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2168/head:pull/2168` \
`$ git checkout pull/2168`

`$ git checkout pull/2168` \
`$ git pull https://git.openjdk.org/jfx.git pull/2168/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2168`

`$ git pr show -t 2168`

</details>

<a href="https://git.openjdk.org/jfx/pull/2168.diff">https://git.openjdk.org/jfx/pull/2168.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2168#issuecomment-4462899048)
</details>


lukostyra (commented):
The changes compile correctly and everything seems to work fine. I do have a question though.

I have a general concern relating future gstreamer updates. Inevitably, we'll have to update it on our side anyway, and this seems to add more complexity to the upgrades. Outside of removing code we don't need, what do we gain from doing this?

sashamatveev:
> The changes compile correctly and everything seems to work fine. I do have a question though.
> 
> I have a general concern relating future gstreamer updates. Inevitably, we'll have to update it on our side anyway, and this seems to add more complexity to the upgrades. Outside of removing code we don't need, what do we gain from doing this?

Main benefits:
- Binary size reduction, since UNICODE tables takes space. Example: Windows: glib-lite.dll 674.50 KiB -> 359.50 KiB. MacOS: 1,234,000 bytes -> 1,040,896 bytes.
- GSTREAMER_LITE define does not add complexity too much when merging code with upstream. Removed files will actually reduce effort to merge with upstream, since less code to merge.
- Eventually I would prefer to remove as much unused code as possible. I think we still have plenty of unused code in GStreamer/GLib.

modules/javafx.media/src/main/native/gstreamer/3rd_party/glib/glib/goption.h
```diff
@@ -364,15 +364,20 @@ GLIB_AVAILABLE_IN_ALL
...
                const gchar         *translation_domain);
+#ifndef GSTREAMER_LITE
 GLIB_AVAILABLE_IN_ALL
 gboolean        g_option_context_parse            (GOptionContext      *context,
                gint                *argc,
                gchar             ***argv,
                GError             **error);
+#endif // GSTREAMER_LITE
+
+#ifndef GSTREAMER_LITE
```
kevinrushforth: Minor, you can remove these two lines (no need to close the `ifndef` only to start a new one).

modules/javafx.media/src/main/native/gstreamer/3rd_party/glib/glib/gunicode.h
```diff
@@ -735,13 +736,13 @@ gunichar g_unichar_totitle (gunichar c) G_GNUC_CONST;
    numeric value.  Otherwise return -1.  */
 GLIB_AVAILABLE_IN_ALL
 gint g_unichar_digit_value (gunichar c) G_GNUC_CONST;
-
```
kevinrushforth, line 738: Minor: I recommend reverting this removal of a blank line. Since it is in an `#ifndef GSTREAMER_LITE` its only effect is as a potential future merge conflict.

modules/javafx.media/src/main/native/gstreamer/3rd_party/glib/glib/goption.c
```diff
@@ -663,10 +666,12 @@ context_has_h_entry (GOptionContext *context)
           if (group->entries[i].short_name == 'h')
             return TRUE;
         }
-    }
+  }
```
kevinrushforth: Minor: I recommend reverting this indentation change. Since it is in an `#ifndef GSTREAMER_LITE` its only effect is as a potential future merge conflict.

modules/javafx.media/src/main/native/gstreamer/3rd_party/glib/glib/goption.c
```diff
@@ -663,10 +666,12 @@ context_has_h_entry (GOptionContext *context)
...
   return FALSE;
 }
+#endif // GSTREAMER_LITE
 
+#ifndef GSTREAMER_LITE
```
kevinrushforth: Minor: here is another place where you could remove the `endif` ... `ifndef`

kevinrushforth (commented):
Overall looks good. I left a few minor comments. It would be good to at least revert the unneeded whitespace changes inside `!GSTREAMER_LITE`.

sashamatveev:
8379561: Remove UNICODE related code from GLib [v5]
 - Fixed review comments.

kevinrushforth (approved):

kevinrushforth:
@tiainen Do you want to do a test build? If so, can you be the second reviewer?

lukostyra (approved):
LGTM

PR #2169, opened 2026-05-16
8384806: ComboBox converter does not properly update contained null values

crschnick:
This is a follow-up to the combobox converter PR. It seems like the updateDisplayNode method in the Skin does not properly handle null cases for when null is an item in the combobox. It probably incorrectly assumes that any null values is mapped to null by the stringconverter. However, if you create a custom converter, a null value can have a non-null string value.

---------

---------
### Progress

### Issue
 * [JDK-8384806](https://bugs.openjdk.org/browse/JDK-8384806): ComboBox converter does not properly update contained null values (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2169/head:pull/2169` \
`$ git checkout pull/2169`

`$ git checkout pull/2169` \
`$ git pull https://git.openjdk.org/jfx.git pull/2169/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2169`

`$ git pr show -t 2169`

</details>

<a href="https://git.openjdk.org/jfx/pull/2169.diff">https://git.openjdk.org/jfx/pull/2169.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2169#issuecomment-4466671413)
</details>

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
@@ -359,24 +359,23 @@ public class ComboBoxListViewSkin<T> extends ComboBoxPopupControl<T> {
         } else {
             T value = comboBox.getValue();
             int index = getIndexOfComboBoxValueInItemsList();
+
+            // JDK-8127575 Show the ComboBox value even though it doesn't
+            // exist in the ComboBox items list (part two of fix)
             if (index > -1) {
                 buttonCell.setItem(null);
-                buttonCell.updateIndex(index);
-            } else {
-                // JDK-8127575 Show the ComboBox value even though it doesn't
-                // exist in the ComboBox items list (part two of fix)
-                buttonCell.updateIndex(-1);
-                boolean empty = updateDisplayText(buttonCell, value, false);
-
-                // Note that empty boolean collected above. This is used to resolve
-                // JDK-8124141, where we were getting different styling based on whether
-                // the cell was updated via the updateIndex method above, or just
-                // by directly updating the text. We fake the pseudoclass state
-                // for empty, filled, and selected here.
-                buttonCell.pseudoClassStateChanged(PSEUDO_CLASS_EMPTY,    empty);
-                buttonCell.pseudoClassStateChanged(PSEUDO_CLASS_FILLED,   !empty);
-                buttonCell.pseudoClassStateChanged(PSEUDO_CLASS_SELECTED, true);
             }
+            buttonCell.updateIndex(index);
+            boolean empty = updateDisplayText(buttonCell, value, false);
+
+            // Note that empty boolean collected above. This is used to resolve
+            // JDK-8124141, where we were getting different styling based on whether
+            // the cell was updated via the updateIndex method above, or just
+            // by directly updating the text. We fake the pseudoclass state
+            // for empty, filled, and selected here.
+            buttonCell.pseudoClassStateChanged(PSEUDO_CLASS_EMPTY,    empty);
+            buttonCell.pseudoClassStateChanged(PSEUDO_CLASS_FILLED,   !empty);
+            buttonCell.pseudoClassStateChanged(PSEUDO_CLASS_SELECTED, true);
         }
     }
 
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
@@ -362,9 +362,6 @@ public class ComboBoxListViewSkin<T> extends ComboBoxPopupControl<T> {
 
             // JDK-8127575 Show the ComboBox value even though it doesn't
             // exist in the ComboBox items list (part two of fix)
-            if (index > -1) {
-                buttonCell.setItem(null);
-            }
             buttonCell.updateIndex(index);
             boolean empty = updateDisplayText(buttonCell, value, false);
 
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
@@ -413,6 +413,30 @@ public class ComboBoxTest {
 
         assertEquals("NOT NULL", cell.getText());
     }
+    @Test
+    public void testTextFieldUpdateOnStringConverterChangeWithContainedNullValue() {
+        ObservableList<String> items = FXCollections.observableArrayList(null, "ITEM1", "ITEM2");
+        comboBox.setEditable(true);
+        comboBox.setItems(items);
+        comboBox.setValue(null);
+
+        TextField field = (TextField) getDisplayNode();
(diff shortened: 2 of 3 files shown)


crschnick:
@andy-goryachev-oracle It would nice if the monkey tester combobox page would support adding null values to the items as well to test these cases better

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
```diff
@@ -359,24 +359,23 @@ public final void setHideOnClick(boolean value) {
         } else {
             T value = comboBox.getValue();
             int index = getIndexOfComboBoxValueInItemsList();
+
+            // JDK-8127575 Show the ComboBox value even though it doesn't
+            // exist in the ComboBox items list (part two of fix)
             if (index > -1) {
                 buttonCell.setItem(null);
```
Maran23: Is this even needed? Because when we update the index anyway below, we are usually also setting the item?

crschnick: I am not sure, I just kept the original semantics. Maybe there was a reason why it did it like this

Maran23: I think we should consider removing it. The call to `updateIndex` below will lead to a `updateItem(int oldIndex)` call, which calls `updateItem(T item, boolean empty)`. And this will set the item as expected.

crschnick: Yeah I removed it, the tests still pass

andy-goryachev-oracle:
> It would nice if the monkey tester combobox page would support adding null values to the items as well to test these cases better

Done - grab the latest from 

crschnick:
@andy-goryachev-oracle Thanks, I was able to test the issue with the monkey tester. Without this PR, the button cell does not update when changing the converter while the null value is selected. WIth the PR, it updates as expected

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
```diff
@@ -359,24 +359,20 @@ public final void setHideOnClick(boolean value) {
...
+            boolean empty = updateDisplayText(buttonCell, value, false);
+
+            // Note that empty boolean collected above. This is used to resolve
+            // JDK-8124141, where we were getting different styling based on whether
+            // the cell was updated via the updateIndex method above, or just
+            // by directly updating the text. We fake the pseudoclass state
+            // for empty, filled, and selected here.
+            buttonCell.pseudoClassStateChanged(PSEUDO_CLASS_EMPTY,    empty);
+            buttonCell.pseudoClassStateChanged(PSEUDO_CLASS_FILLED,   !empty);
+            buttonCell.pseudoClassStateChanged(PSEUDO_CLASS_SELECTED, true);
```
andy-goryachev-oracle, line 375: this is not exactly an equivalent change, since we are updating pseudostyles (esp. PSEUDO_CLASS_SELECTED, true).  is it important?  probably not: we'll end up with pseudostyles set on something not connected to the scene graph, probably not a big deal.

andy-goryachev-oracle (commented):
The test passes with the fix and fails without.  I do have a question however:

Since a reproducer is missing from the JBS ticket, what is expected of this change, visually?  To illustrate, first grab the latest monkey tester where I modified the "Quoted" converter to handle null as `◇` symbol.

With a null value selected from the pulldown, I am expecting to see ◇ in the text field, right?  But instead the text field is empty.  Yet when I enter an empty text by clearing the text field and pressing ENTER, I do get `""` as expected.

I don't see any changes in behavior between the master and this PR.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
crschnick: This should correct as that is related to the problem. When you select a value, the empty pseudo style class should always update. It previously did not do that in all cases like when the value was null but the converted string was not null.

crschnick:
@andy-goryachev-oracle It seems like this problem has two different causes depending on whether the combobox is editable or not. I fixed that one as well.

In both cases, the wrong assumption is that a null values is converted to a null string value. Which does not have to be true

andy-goryachev-oracle (approved):
Looks good now, thanks!

Maran23:
I can be the second reviewer, but need some time to do more testing in some projects.

andy-goryachev-oracle:
> I can be the second reviewer

thank you so much!  we've got too many PRs stuck waiting for the second reviewer, or expiring altogether...

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
```diff
@@ -388,6 +388,82 @@ public String fromString(String string) {
...
 
+    @Test
+    public void testButtonCellUpdateOnStringConverterChangeWithContainedNullValue() {
+        ObservableList<String> items = FXCollections.observableArrayList(null, "ITEM1", "ITEM2");
+        comboBox.setEditable(false);
+        comboBox.setItems(items);
+        comboBox.setValue(null);
+
+        ListCell<String> cell = (ListCell<String>) ((ComboBoxListViewSkin<String>) comboBox.getSkin())
+                .getDisplayNode();
```
Maran23: Just call and cast `getDisplayNode();` directly

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
```diff
@@ -388,6 +388,82 @@ public String fromString(String string) {
...
+
+    @Test
+    public void testButtonCellUpdateOnStringConverterChangeWithUncontainedNullValue() {
+        ObservableList<String> items = FXCollections.observableArrayList("ITEM1", "ITEM2");
+        comboBox.setEditable(false);
+        comboBox.setItems(items);
+        comboBox.setValue(null);
+
+        ListCell<String> cell = (ListCell<String>) ((ComboBoxListViewSkin<String>) comboBox.getSkin())
+                .getDisplayNode();
```
Maran23: Just call and cast `getDisplayNode();` directly

Maran23 (commented):
Two minor improvements in the test - otherwise looks good and all my tests were successful.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
crschnick: Fixed

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
crschnick: Fixed

Maran23 (approved):

andy-goryachev-oracle (approved):
looks good

PR #2171, opened 2026-05-18
8384814: [macOS] Trackpad scroll gestures suddenly stop

beldenfox:
On macOS when the user swipes on a trackpad or Magic Mouse JavaFX sees this as a scroll gesture. The scene tries to ensure that all the scroll events generated by the gesture are fired at the same target to ensure that the gesture's scroll events don't get split between two ScrollPanes. For example, while a ScrollPane is being scrolled a child ScrollPane might shift until it's beneath the mouse pointer. Scroll events should not fire at the child but remain with the parent.

During the gesture the scene should target the node that's handling the scrolling (like a ScrollPane) but there's no good way to determine that. Instead it targets a descendant node, whatever was under the mouse pointer when the gesture started. This node is likely to shift during the scroll and if scrolls out of view it might get disconnected from the scene graph. At that point events fired at it will go nowhere and the gesture will abruptly stop.

This PR detects when the gesture target is removed from the scene graph and attempts to find a new target.

This bug does not reproduce on Windows since that platform doesn't generate a SCROLL_STARTED event so the Scene never selects a gesture target. All scroll events are delivered to whatever node is under the mouse pointer which will change as the scroll progresses.

This PR might fix [JDK-8088460](https://bugs.openjdk.org/browse/JDK-8088460) which shows up in Ensemble if anyone wants to test that.

Submitting fix for sanity checking and manual testing. Will investigate creating an automated test for this.

---------

---------
### Progress

### Issue
 * [JDK-8384814](https://bugs.openjdk.org/browse/JDK-8384814): [macOS] Trackpad scroll gestures suddenly stop (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2171/head:pull/2171` \
`$ git checkout pull/2171`

`$ git checkout pull/2171` \
`$ git pull https://git.openjdk.org/jfx.git pull/2171/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2171`

`$ git pr show -t 2171`

</details>

<a href="https://git.openjdk.org/jfx/pull/2171.diff">https://git.openjdk.org/jfx/pull/2171.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2171#issuecomment-4481778650)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
@@ -2011,6 +2011,19 @@ public class Scene implements EventTarget {
 
         if (gesture.target != null && (!gesture.finished || e.isInertia())) {
             pickedTarget = gesture.target.get();
+            if (pickedTarget instanceof Node) {
+                // During a scroll gesture the initial target may have
+                // scrolled out of sight and been removed from the scene
+                // graph.
+                Node pickedNode = (Node) pickedTarget;
+                if (pickedNode.getScene() == null) {
+                    gesture.target = null;
+                    pickedTarget = e.getPickResult().getIntersectedNode();
+                    if (pickedTarget != null) {
+                        gesture.target = new WeakReference<>(pickedTarget);
+                    }
+                }
+            }
         } else {
             pickedTarget = e.getPickResult().getIntersectedNode();
             if (pickedTarget == null) {
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
@@ -2021,6 +2021,8 @@ public class Scene implements EventTarget {
                     pickedTarget = e.getPickResult().getIntersectedNode();
                     if (pickedTarget != null) {
                         gesture.target = new WeakReference<>(pickedTarget);
+                    } else {
+                        pickedTarget = this;
                     }
                 }
             }


Maran23 (commented):
Logic makes sense to me.
Unfortunately, I can not test it, so will not approve. 
Will try to borrow a Mac and test it in some days.

Maran23:
> Will investigate creating an automated test for this

Might be possible by sending manual scroll events, similar of using `SceneHelper.processMouseEvent`.
`SceneTest` contains some tests that uses this mechanism.

youngledo:
While reading the patch, I wondered about one small consistency edge case: in the retargeting branch, if the previous gesture target has been detached and the current pick has no intersected node, `pickedTarget` stays `null`. The existing non-cached path falls back to the `Scene` in that case. Would it be worth keeping the same fallback behavior here?

A related minor thought: would `pickedNode.getScene() != this` be a slightly broader check than `pickedNode.getScene() == null`? It would cover both a detached node and the unlikely case where the old target has been moved to another scene before the rest of the gesture is delivered.

This does not change the main idea of the fix; I am only asking whether this branch should mirror the existing dispatch path more closely.

youngledo:
I tested this PR with the same small ListView reproducer that I used when discussing the issue in #2051.

Environment:
- MacBook Pro with Apple Silicon
- macOS 26.5
- Trackpad scrolling
- PR commit tested: `205de3d292`
- JavaFX runtime shown by the rebuilt local SDK: `27-internal+0-2026-05-28-092752`

How I tested it locally:
1. Built a local SDK from this PR.
2. Ran my small ListView reproducer against that SDK.
3. Reproduced the gesture with continuous slow trackpad scrolling over the ListView.
4. Repeated the same test with both ES2 and Metal pipeline configurations.

For the Metal run, I used:

```bash
-Dprism.order=mtl
-Dprism.verbose=true
```

The verbose output confirmed that the Metal pipeline was selected:

```text
Prism pipeline init order: mtl
Prism pipeline name = com.sun.prism.mtl.MTLPipeline
Initialized prism pipeline: com.sun.prism.mtl.MTLPipeline
```

With the current master / JavaFX 25 behavior, the ListView scroll gesture would stop responding during a continuous slow trackpad scroll. With this PR applied, I can no longer reproduce that problem in the same demo; the ListView continues scrolling normally and feels smooth in my manual test.

I first tested with an ES2-only local SDK using `-Dprism.order=es2` because my Xcode installation was missing `metallib`. After installing the Metal Toolchain, I rebuilt the SDK without that limitation and reran the same demo forcing the Metal pipeline as described above.

This is just a manual reproducer check from an affected user / outside observer, not a full code review, but the original issue I was seeing is no longer reproducible for me with either ES2 or Metal.

youngledo:
Here is a recorded video:
[demo-fx.webm](https://github.com/user-attachments/assets/5c9803cf-fc8f-4707-b889-283a814ff134)

beldenfox:
@youngledo Thanks for testing this. You're right, I should use the same fallback in both cases. I'm away from my development machines right now but will make the fix early next week.

beldenfox:
I've updated the code to be more consistent when the event's picked node is null. It now sends the scroll messages to the scene instead of discarding them. The scene probably doesn't handle them but at least they're not lost entirely. To test this case you need to swipe and then quickly move the cursor so it's no long on the stage. There's a good chance this will lead to inertial scroll events coming in with no picked node.

youngledo:
I also tried the additional case you suggested: start a trackpad scroll on the `ListView`, then quickly move the cursor outside the stage/window while inertial scrolling is still active.

Environment / build:
- Latest PR commit tested: `2e15d433b9`
- JavaFX runtime shown by the local build: `27-internal+0-2026-06-03-022449`
- Metal pipeline enabled (`-Dprism.order=mtl`, confirmed by verbose output showing `com.sun.prism.mtl.MTLPipeline`)

What I observed: after moving the cursor outside the window, the scrollbar stops almost immediately.

This seems consistent with the new fallback sending the event to the `Scene` when the picked node is null, since the `ListView` no longer appears to receive the scroll event in that case. I am not sure whether that behavior is expected, but I wanted to report the result from the suggested test case.

andy-goryachev-oracle (commented):
Does it fix https://bugs.openjdk.org/browse/JDK-8088460 ?
If it does, you probably want to add that issue to this PR.

andy-goryachev-oracle:
I closed the Ensemble ticket as a duplicate.

andy-goryachev-oracle (approved):
looks good on macOS 26.5.1 M1.
no ill effects in the reproducer, monkey tester, and Ensemble.

youngledo:
It seems this PR may address a long-standing trackpad scrolling issue that macOS users have been seeing for many years. 

If the fix is accepted for mainline, is a backport to the JavaFX 25 update line something the project would consider?

kevinrushforth:
> If the fix is accepted for mainline, is a backport to the JavaFX 25 update line something the project would consider?

That would be a question for @johanvos

johanvos:
In general, we try to only include security fixes and critical issues, and not a P4 bug. Backports to LTS releases are resource (time) intensive, and I'd like to focus this time on making 100% sure all security updates are included, and work 100% as intented.
However, this particular fix seems pretty clear and straightforward, and doesn't touch native code, so I think this can be considered.

kevinrushforth:
@arapte can you be the second reviewer?

modules/javafx.graphics/src/main/java/javafx/scene/Scene.java
```diff
@@ -2011,6 +2011,21 @@ private void processGestureEvent(GestureEvent e, TouchGesture gesture) {
 
         if (gesture.target != null && (!gesture.finished || e.isInertia())) {
             pickedTarget = gesture.target.get();
+            if (pickedTarget instanceof Node) {
```
Maran23, line 2014: Minor suggestion

```suggestion
            if (pickedTarget instanceof Node node) {
```

(and remove the cast below)

Maran23 (approved):
Code looks good and reasonable.
Just tested the fix together with a friend on his Mac (v26.5.1). 
Looks good and you can clearly see the broken/laggy scrolling before and the smooth scrolling after.

beldenfox:
Note: I did look into adding system tests. To begin to replicate what's going on here the tests would need to be in Controls and involve a series of ScrollEvents. The events would need to be created with accurate PickResults and I couldn't find existing tests that did this. I'm willing to add tests as a follow-up PR but would appreciate some guidance on how to proceed.

PR #2175, opened 2026-05-26
8385270: Bump JNI_VERSION to 1.8 everywhere

Maran23:
This replaces all `JNI_VERSION` version strings with `JNI_VERSION_1_8`.
All of JavaFX is `JNI_VERSION_1_8` now.

- Tested with JLink+JPackage (with and without JMods)
- Tested with GraalVM (newest version 25)

Sometimes this allowed to replace code that was always built like this:

```
#ifdef JNI_VERSION_1_8
    //min. returned JNI_VERSION required by JDK8 for builtin libraries
    JNIEnv *env;
    if ((*vm)->GetEnv(vm, (void **)&env, JNI_VERSION_1_8) != JNI_OK) {
        return JNI_VERSION_1_4;
    }
   ...
```

with:
```
return JNI_VERSION_1_8;
```

**Question -> Will do as Follow-up** 

~There are a few locations where we have the following pattern;~
```
    if (jvm->GetEnv((void **)&env, JNI_VERSION_1_8)) {
        return JNI_ERR; /* JNI version not supported */
    }
```
~which should not happen anymore, right? 
Most of the code locations do not have this check and just do:~
```
jvm->GetEnv(&env, JNI_VERSION_1_8);
```
~Should I follow that pattern for the few locations left?~

---------

---------
### Progress

### Issue
 * [JDK-8385270](https://bugs.openjdk.org/browse/JDK-8385270): Bump JNI_VERSION to 1.8 everywhere (**Enhancement** - P4)

### Reviewers
 * [Michael Strauß](https://openjdk.org/census#mstrauss) (@mstr2 - **Reviewer**) 🔄 Re-review required (review applies to [6ea68250](https://git.openjdk.org/jfx/pull/2175/files/6ea682505af56bbdac51e294e9bb204284d82dba))

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2175/head:pull/2175` \
`$ git checkout pull/2175`

`$ git checkout pull/2175` \
`$ git pull https://git.openjdk.org/jfx.git pull/2175/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2175`

`$ git pr show -t 2175`

</details>

<a href="https://git.openjdk.org/jfx/pull/2175.diff">https://git.openjdk.org/jfx/pull/2175.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2175#issuecomment-4549332199)
</details>


mstr2:
> There are a few locations where we have the following pattern;
> 
> ```
>     if (jvm->GetEnv((void **)&env, JNI_VERSION_1_8)) {
>         return JNI_ERR; /* JNI version not supported */
>     }
> ```
> 
> which should not happen anymore, right? Most of the code locations do not have this check and just do:
> 
> ```
> jvm->GetEnv(&env, JNI_VERSION_1_8);
> ```
> 
> Should I follow that pattern for the few locations left?

When `GetEnv()` fails, it is documented to return either `JNI_EDETACHED` when called from a non-JVM thread, or `JNI_EVERSION` when the version is not supported. I think we can safely rule out both cases: it's impossible for JavaFX to run on a non-JVM thread, and it is also impossible for JavaFX to run on a JVM < 8.

Maran23:
> When `GetEnv()` fails, it is documented to return either `JNI_EDETACHED` when called from a non-JVM thread, or `JNI_EVERSION` when the version is not supported. I think we can safely rule out both cases: it's impossible for JavaFX to run on a non-JVM thread, and it is also impossible for JavaFX to run on a JVM < 8.

Same thought. Thanks for confirming, will do in the coming days!

Maran23:
What confuses me is that there are checks like `if (envStatus == JNI_EDETACHED)` in e.g. `gtk/GlassApplication.cpp`.
On the other hand returning `JNI_ERR` and the comment `/* JNI version not supported */` while `JNI_EDETACHED` could also be returned is also weird.

mstr2:
> What confuses me is that there are checks like `if (envStatus == JNI_EDETACHED)` in e.g. `gtk/GlassApplication.cpp`. On the other hand returning `JNI_ERR` and the comment `/* JNI version not supported */` while `JNI_EDETACHED` could also be returned is also weird.

`call_runnable()` in `gtk/GlassApplication.cpp` is called from `gtk_main()` on the GTK main-loop thread, which is always a JVM thread (see `GtkApplication.runLoop()`). The code that checks for a detached JVM is pointless, because even if for some reason this function were to be called on a native thread, it would still break JavaFX regardless of the defensive programming because it would call into JavaFX on a thread other than the event thread.

Maran23:
Will create a follow-up task and assign it to me - so we can make this change first, which should be uncontroversial, and then I'll implement the `env` simplification and write in the mailing list again. Especially regarding the Android and Web env changes, I'd at least prefer to make sure this is fine for Gluon as well.

mstr2 (approved):
LGTM

kevinrushforth:
I'll review this right after the fork and after the WebKit 625.1 update is integrated. One thing to note is that we will need to backport the WebKit portion of this fix to earlier release trains to keep the native code in sync across code lines. Since the fixes for each library are independent, this will not be a problem.

modules/javafx.graphics/src/main/native-glass/monocle/EGL.c
```diff
@@ -31,19 +31,9 @@
...
-JNI_OnLoad_glass_monocle(JavaVM *vm, void * reserved) {
-fprintf(stderr, "In JNI_OnLoad_glass)monocle\n");
-#ifdef JNI_VERSION_1_8
-    //min. returned JNI_VERSION required by JDK8 for builtin libraries
-    JNIEnv *env;
-    if ((*vm)->GetEnv(vm, (void **)&env, JNI_VERSION_1_8) != JNI_OK) {
-        return JNI_VERSION_1_4;
-    }
+JNIEXPORT jint JNICALL JNI_OnLoad_glass_monocle(JavaVM *vm, void *reserved) {
+    fprintf(stderr, "In JNI_OnLoad_glass)monocle\n");
```
Maran23, line 35: BTW: Do we want to keep that print here? Seems weird to me but kept it for now

PR #2177, opened 2026-05-28
8385459: Animations should respect reducedMotion preference

mstr2:
Several JavaFX controls use animations to convey state changes, some of which are controllable by developers (for example: `TitledPane.animated` or `Chart.animated`). However, none of those controls respect the `reducedMotion` preference that was introduced with [JDK-8341514](https://bugs.openjdk.org/browse/JDK-8341514).

This enhancement changes the behavior of the following control skins to take the `Scene.Preferences.reducedMotion` preference into account when determining whether to animate a state change:

* `TableRowSkinBase`
* `TitledPaneSkin`
* `TabPaneSkin`
* `PaginationSkin`
* Charts

For controls whose animation can be toggled by developers, it is useful to think of the `animated` property as the application choice, while `reducedMotion` is the user choice.

---------

---------
### Progress

### Issue
 * [JDK-8385459](https://bugs.openjdk.org/browse/JDK-8385459): Animations should respect reducedMotion preference (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2177/head:pull/2177` \
`$ git checkout pull/2177`

`$ git checkout pull/2177` \
`$ git pull https://git.openjdk.org/jfx.git pull/2177/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2177`

`$ git pr show -t 2177`

</details>

<a href="https://git.openjdk.org/jfx/pull/2177.diff">https://git.openjdk.org/jfx/pull/2177.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2177#issuecomment-4559811152)
</details>


modules/javafx.controls/src/main/java/javafx/scene/control/skin/TitledPaneSkin.java
```diff
@@ -290,7 +292,7 @@ private void setExpanded(boolean expanded) {
         }
 
         // we need to perform the transition between expanded / hidden
-        if (Platform.isFxApplicationThread() && getSkinnable().isAnimated()) {
+        if (shouldAnimate()) {
             transitionStartValue = getTransition();
             doAnimationTransition();
         } else {
```
lukostyra, line 300: I'm wondering if we should also check here if the `timeline` is valid and if it is cancel it. Is there a chance `reducedMotion` could be changed to false while we are midway through animating the transition?

mstr2: The question whether `reducedMotion` can change while an animation is in progress is not important: we don't observe the current reduced-motion value, we only sample it once when the animation starts. Allowing an animation to run to completion should always be correct behavior.

However, there is a related problem how we deal with "stale" animations, that is, running animations that are preempted by a new animation. In general, the previous animation should be canceled, and the animated value should either be replaced with the new value, or reconciled with the current value. I've added a bit of code in some places to better clean up old animations.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TitledPaneSkin.java
```diff
@@ -194,6 +196,10 @@ private final DoubleProperty transitionProperty() {
     @Override public void dispose() {
         super.dispose();
 
+        if (timeline != null) {
```
andy-goryachev-oracle: isn't this just `cancelTransitionAnimation()` ?

andy-goryachev-oracle (commented):
It seems to work, the issue that I have with the implementation is a different kind.  Let me explain: I use 'reduce motion' setting to limit the amount of animation in the _OS widgets_.  I was **not** expecting to see the animation disabled when the `animated` flag on the chart is set to `true`.

The original rationale behind disabling animation in charts was to  "save CPU cycles" and it sort of makes sense.  Still questionable (because the earlier discussion about whether 'embedded' in this context should be 'low power' instead).  If the idea is to conserve power, then yes, the animation should be disabled.

On a desktop, however, the new behavior does not seem right.  If the chart is animated via property, the animation must run.

Perhaps what we should do is to control the default value of the animated property(ies) instead.  So if the application does not explicitly set the `animated` property, then it follows the platform, otherwise the application-set value is used.

Furthermore, perhaps there should be another platform preference property like `low power` which **will** disable animations when active.  For example, this property will kick in when the battery drops below 10% or whatever.

What do you think?

mstr2:
The question whether `reducedMotion` should override `animated` or if it should be the other way around is indeed a very good question. I've thought quite a bit about this, and in general I agree with your idea that ideally there would be a "default value" that would depend on platform or scene preferences.

There's a little bit of precedence for this with `Scene.Preferences`, where all properties are "null-coalescing". This means that if set to `null`, the property will instead take on its default value. Since this isn't supported by the Property API, it requires a bit of trickery, where for example `Scene.Preferences.reducedMotion` is defined as:
```java
ObjectProperty<Boolean> reducedMotionProperty();
boolean isReducedMotion();
void setReducedMotion(Boolean);
```

This API allows users to assign `null` to a boolean property, but never see `null` when its value is read. This trick isn't possible with `animated`, which is defined as:
```java
BooleanProperty animatedProperty();
boolean getAnimated();
void setAnimated(boolean); // can't set null
```

We could try to invent another ad-hoc mechanism to support a default value, for example by having an internal flag in the property implementation that tracks whether any code ever attempted to set or bind its value; and if so, treat the property as user-set. However, I don't think that this particular idea is good, because it's too magical and it's irreversible: set the value once, and you can never go back to the default behavior.

I've also entertained the idea of adding another `setAnimated(Boolean)` overload, but that doesn't feel right. Maybe something like `clearAnimated()` could work, which would clear the user-set value and reset the property back to its default value. But that would be yet another ad-hoc invention, and it would also not be reflected in the Property API.

If we can't change the API of `animated`, and if we also don't want to have magical property behavior, there's the obvious question: should `reducedMotion` just override `animated`?

Let's explore both scenarios:
1. `reducedMotion` overrides `animated`:
   In this case, users can change animations application-wide, regardless of what the application author specified for each individual control. If the author _insists_ that users can't disable animations, there's the option of setting `Scene.Preferences.setReducedMotion(false)`, such that it overrides the platform default. I would call this "everyone can have their cake".
2. `animated` overrides `reducedMotion`:
   This is much trickier: application authors can have legitimate reasons to enable animations on some, but not on all controls. But simply by setting the `animated` property, the value of `reducedMotion` is ignored. This can't be easily circumvented, even if the author wants to support a reduced-motion user experience. In this scenario, someone must lose: either the author can't set the `animated` property, or users can't toggle a reduced-motion experience.

> On a desktop, however, the new behavior does not seem right. If the chart is animated via property, the animation must run.

I'd interpret it differently: `Chart.setAnimated(true)` doesn't necessarily mean that the animation must be forced to run, no matter what. For me, another interpretation makes more sense: that the author intended for the animation to run, but not if the user clearly expressed that they don't want animations. Mind you: the author still has full control to ignore the users's wish by just setting `Scene.Preferences.setReducedMotion(false)`.

> Furthermore, perhaps there should be another platform preference property like low power which will disable animations when active. For example, this property will kick in when the battery drops below 10% or whatever.

Even if we had this, I'd still think that `reducedMotion` should affect all animations, if not explicitly disabled by the application author.

andy-goryachev-oracle:
Thank you for a detailed writeup!

I wonder if the issue comes down to whether the chart animation represents "non-essential animation which reduces discomfort for users who experience motion sickness or vertigo" (`Scene.Preferences.reducedMotionProperty()`).

And also whether we want to disable animation to save CPU cycles / battery, the original rationale for the `animated` property.

What do other FX devs/users think?

mstr2:
> I wonder if the issue comes down to whether the chart animation represents "non-essential animation which reduces discomfort for users who experience motion sickness or vertigo" (`Scene.Preferences.reducedMotionProperty()`).

In addition to the language used in the javadoc of the `reducedMotion` property, some users also choose to reduce motion because they perceive it to be [faster](https://www.reddit.com/r/LifeProTips/comments/1expslr/lpt_if_you_want_your_windows_11_pc_to_feel_a) and [snappier](https://news.ycombinator.com/item?id=33222434), so they don't have to wait a few hundred milliseconds for the animation to complete. This is a legitimate use case that we shouldn't dismiss.

In general, I think that chart animations should also qualify as "non-essential". As a matter of fact, when I open up Monkey Tester with charts that have a large number of data points, I think animations add to the confusion instead of helping to resolve it, which can make them not only non-essential, but actively detrimental (and if the number of data points is too large, animations are pointless if it takes a few seconds to render even a single frame).

> And also whether we want to disable animation to save CPU cycles / battery, the original rationale for the `animated` property.

If we add support to query the power-saver mode (most operating systems have it), then we should consider having it affect the `reducedMotion` property directly. The big difference between `reducedMotion` and a hypothetical `lowPower` setting is that the former unambiguously determines the _outcome_, whereas the latter is just a description of some system state, and the outcome depends on what the author thought would be associated with that system state.

andy-goryachev-oracle:
Not sure I agree with lowPower affecting reducedMotion (shouldn't these two be orthogonal?), but I wanted to suggest another possibility: a _static_ property in Chart that controls how the animated handles the platform preferences - whether it overrides it, or the other way around.

What do you think?

Maran23:
> The question whether reducedMotion should override animated or if it should be the other way around is indeed a very good question

I was also wondering how we want to handle or retrofit that. But ...

> I'd interpret it differently: Chart.setAnimated(true) doesn't necessarily mean that the animation must be forced to run, no matter what. For me, another interpretation makes more sense: that the author intended for the animation to run, but not if the user clearly expressed that they don't want animations. Mind you: the author still has full control to ignore the users's wish by just setting Scene.Preferences.setReducedMotion(false).

This convinced me and makes sense.

When I wrote JavaFX applications, I would, for example, enable animations for `TitledPane`s if I thought they looked good, and disable them if I did not think they fit (like in an `Accordion` with many panes - it just took to long to navigate through them). Same for for `Chart`s or `Tab`-closing animations.

I often did things the way I thought was best for usablity and style.
This does not mean every single User may likes that. It is always hard to get that right.

So IMHO, it makes sense that we treat `setAnimated` as the developer choice. And `reducedMotion` as the user choice:
They can decide if they want to see the animations I enabled.

> In addition to the language used in the javadoc of the reducedMotion property, some users also choose to reduce motion because they perceive it to be [faster](https://www.reddit.com/r/LifeProTips/comments/1expslr/lpt_if_you_want_your_windows_11_pc_to_feel_a) and [snappier](https://news.ycombinator.com/item?id=33222434), so they don't have to wait a few hundred milliseconds for the animation to complete. This is a legitimate use case that we shouldn't dismiss.

This is correct and I disabled Animations in Windows 11 for a long time because they took to long - it feels snappier without them.

All that brings me to the same conclusion: We as a developer decide what we want to get animated and what we do not want to. The user can disable that completely with the `reducedMotion` preference and I think we should honor that.

andy-goryachev-oracle:
Thank you all for a thoughtful discussion!

I now agree with the rationale and the behavior, because:

1. animated is the application choice, and reducedMotion is the user choice
2. application may force `Scene.Preferences.setReducedMotion(false)` when needed

(maybe we can mention these two points in the JBS and this PR description, for posterity?)

Also, I wonder if the monkey tester should have a tool to control the platform preferences.   What do you think?

andy-goryachev-oracle (approved):
looks good, with a minor question (code is ok as is through)

andy-goryachev-oracle:
Question: do we want to add a `lowPower` platform preference?  Did anyone encountered a need for it?

I can see at least two possible uses:
- disable animations (in the platform)
- save any outstanding buffers to persistent storage (in the application)

mstr2:
> Not sure I agree with lowPower affecting reducedMotion (shouldn't these two be orthogonal?), but I wanted to suggest another possibility: a _static_ property in Chart that controls how the animated handles the platform preferences - whether it overrides it, or the other way around.
> 
> What do you think?

I'm not sure that a global "animated overrides reducedMotion" toggle would actually be useful in practice: since `animated` is true by default, having it override `reducedMotion` just means that reduced-motion mode is always ignored, and the `animated` property always wins. This ends up being the same behavior as forcing `Scene.Preferences.setReducedMotion(false)`.

While I can at least imagine that a developer might want to have a specific animation to _always_ run, and all other animations to respect reduced-motion mode, a global toggle doesn't give us that capability; it would have to be a per-instance toggle. This is something that we should probably discuss in a broader context if it turns out that there is an actual need to have "overridable properties", I think in the specific case of reduced-motion mode we can first do the easier thing and always respect it, and only if necessary add more fine-grained control later.

mstr2:
> Question: do we want to add a `lowPower` platform preference? Did anyone encountered a need for it?
> 
> I can see at least two possible uses:
> 
> * disable animations (in the platform)
> * save any outstanding buffers to persistent storage (in the application)

I'm not completely opposed to it. Just for reference, there is the [update](https://www.w3.org/TR/mediaqueries-5/#update) media query, which matches `slow` for "severely under-powered devices". There's also a [discussion](https://github.com/w3c/csswg-drafts/issues/7389) about a power-saver query, which went nowhere.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TitledPaneSkin.java
```diff
@@ -194,15 +194,13 @@ private final DoubleProperty transitionProperty() {
...
-        if (timeline != null) {
-            timeline.stop();
-        }
+        cancelTransitionAnimation();
 
         if (behavior != null) {
             behavior.dispose();
         }
+
+        super.dispose();
```
andy-goryachev-oracle, line 203: +1 this is the correct sequence, thank you!

andy-goryachev-oracle (approved):

Maran23:
I will review and test tomorrow. 
`reducedMotion` should be `true` when I disable `Visual effects -> Animation effects` under Windows 11, right?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TitledPaneSkin.java
```diff
@@ -192,11 +194,13 @@ private final DoubleProperty transitionProperty() {
 
     /** {@inheritDoc} */
     @Override public void dispose() {
-        super.dispose();
+        cancelTransitionAnimation();
```
Maran23, line 197: Should we cancel animations in other skins as well? Or is there only a need here?

mstr2: I've added some defensive code in the other controls as well to stop potentially running animations in the `dispose()` method.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TabPaneSkin.java
```diff
@@ -1596,6 +1596,10 @@ private void dispose() {
             listener.dispose();
             setOnContextMenuRequested(null);
             setOnMousePressed(null);
+
+            if (currentAnimation != null) {
```
andy-goryachev-oracle: maybe it should be the first thing called in `dispose()`

modules/javafx.controls/src/main/java/javafx/scene/control/skin/PaginationSkin.java
```diff
@@ -409,6 +418,14 @@ public void dispose() {
         getSkinnable().setClip(null);
         getChildren().removeAll(currentStackPane, nextStackPane, navigation);
 
+        if (deferredStartup != null) {
```
andy-goryachev-oracle, line 418: also here, probably should move it before removing children in L418

andy-goryachev-oracle (approved):

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TitledPaneSkin.java
```diff
@@ -388,6 +394,23 @@ private void doAnimationTransition() {
...
+        if (timeline != null) {
+            timeline.stop();
+            timeline = null;
+        }
+    }
+
+    private boolean shouldAnimate() {
+        TitledPane skinnable = getSkinnable();
+        return skinnable != null
+            && Platform.isFxApplicationThread()
```
Maran23, line 407: Here and in `TabPaneSkin`, we are checking for `Platform.isFxApplicationThread()`. The others do not.
Probably okay, but I thought I would bring it up here.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableRowSkinBase.java
```diff
@@ -142,7 +141,14 @@ public TableRowSkinBase(C control) {
         registerChangeListener(control.indexProperty(), e -> requestCellUpdate());
     }
 
+    @Override
+    public void dispose() {
+        if (currentTransitions != null) {
+            currentTransitions.forEach((_, value) -> value.stop());
```
Maran23, line 147: We should `clear()` the `currentTransitions` here

modules/javafx.controls/src/main/java/javafx/scene/control/skin/PaginationSkin.java
```diff
@@ -406,13 +415,21 @@ public void dispose() {
             return;
         }
 
-        getSkinnable().setClip(null);
-        getChildren().removeAll(currentStackPane, nextStackPane, navigation);
+        if (deferredStartup != null) {
+            deferredStartup.cancel();
```
Maran23, line 419: We should also `null` the `deferredStartup` here

mstr2: This is the same case as in `TabRowSkinBase.dispose()`, it's generally not necessary to null out fields of objects we are disposing.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableRowSkinBase.java
mstr2: This shouldn't be necessary, since the object is being disposed here. After that, it is no longer usable and will be eligible for GC. Nulling out fields really only makes sense when the object is _not_ disposed, when only the interior object should be GC'd, but not the enclosing object.

Maran23: I agree, but my thought behind was not the GC, but rather this:

We can change just the skin at runtime. 
In this case we would still have stale references to the `disclosureNode` here in the `Map`. 
And the `disclosureNode` can still exist after we changed the skin, as the `TableRow` is unchanged (and the `disclosureNode` comes from it). Then this skin might not be GC'ed at all.

Also it is allowed to re-use a skin. So e.g. we could replace the skin of `TableRow` with another one, and then back to the previous one. 
I don't think this will work right now because everything is done in the constructor (and not in `install()`), but that might be possible in the future.

I agree both are rare situation and something you most likely don't do. 
But since it does not hurt to clear/null the fields here, and we are improving the skin in case it is not possible to dispose (GC) it immediately, I think we should rather do it and are safe for whatever a developer might do.

mstr2: > I agree, but my thought behind was not the GC, but rather this:
> 
> We can change just the skin at runtime. In this case we would still have stale references to the `disclosureNode` here in the `Map`. And the `disclosureNode` can still exist after we changed the skin, as the `TableRow` is unchanged (and the `disclosureNode` comes from it). Then this skin might not be GC'ed at all.

The disclosure node must not strongly reference the skin after the skin is removed. If it does, it's a bug that needs to be fixed, because that would mean that a removed skin will stick around and not be collected. I don't think there is a bug here: it's okay if the skin strongly references the disclosure node; it's just the other way that would be a problem.

> Also it is allowed to re-use a skin. So e.g. we could replace the skin of `TableRow` with another one, and then back to the previous one. I don't think this will work right now because everything is done in the constructor (and not in `install()`), but that might be possible in the future.

No, it's not allowed to reuse a skin. In the `SkinBase(C)` constructor, the `control` field is assigned to hold the associated control. This field is nulled in `SkinBase.dispose()`, so installing the same skin again on the control will fail with NPE if the `control` field is dereferenced at any point.

> I agree both are rare situation and something you most likely don't do. But since it does not hurt to clear/null the fields here, and we are improving the skin in case it is not possible to dispose (GC) it immediately, I think we should rather do it and are safe for whatever a developer might do.

It _does_ hurt to null fields of a disposed object, because it suggests that this would do anything meaningful (like fixing a memory leak). But that's not the case: if there was a memory leak (i.e. if the skin wasn't eligible for GC and therefore kept its referenced objects alive), nulling the field would just hide the problem. The solution in that case would be to ensure that a disposed skin is eligible for GC.

Maran23: Just to make it more clear what exactly I was worried about, this situation:

```
treeTableRow.setDisclosureNode(node)
└── TreeTableRowSkin picks it up
    └── stores in Map<Node, FadeTransition>

treeTableRow.setSkin(new TreeTableRowSkin(treeTableRow))
└── Old Skin disposed, new TreeTableRowSkin picks it up
    └── stores in Map<Node, FadeTransition> as well

Old skin has the reference and new one has.
```
I did run some tests and could not find a problem. Still was a bit worried as it was not always clear how to correctly write the skin so it follows the best practices. 
I think in the past the people probably chose to rather clear/null everything to be safe. Or, I saw a comment somewhere IRC that this was done to 'help' the GC. So yeah, a bit out of place and I personally then just followed the `null everything` pattern as I saw it quite often.

Example in the super class of `TableRowSkinBase`:
https://github.com/openjdk/jfx/blob/2abb9a3b0a7e1071bea0b8bf2ca3d83e6fdf25f8/modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java#L246-L252

> No, it's not allowed to reuse a skin

Yes (although this can be achieved when extending from `Skin` and using the `install` method correctly - too much work). Just brought it up so we think about everything. But does not affect us in any way.

Maran23:
> For me, another interpretation makes more sense: that the author intended for the animation to run, but not if the user clearly expressed that they don't want animations. 

I'd just like to add that modern browsers work exactly this way. For example, website developers often set smooth scrolling via CSS. However, your browser will not use/honor it if `reducedMotion` is enabled (automatically). This should leave no doubt that this PR is the right way forward.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableRowSkinBase.java
mstr2: > Just to make it more clear what exactly I was worried about, this situation:
> 
> ```
> treeTableRow.setDisclosureNode(node)
> └── TreeTableRowSkin picks it up
>     └── stores in Map<Node, FadeTransition>
> 
> treeTableRow.setSkin(new TreeTableRowSkin(treeTableRow))
> └── Old Skin disposed, new TreeTableRowSkin picks it up
>     └── stores in Map<Node, FadeTransition> as well
> 
> Old skin has the reference and new one has.
> ```

Okay, but unless I misunderstand something really fundamental here, there shouldn't be a problem if the old skin is not retained. It will simply be collected, along with the `Map` which contains the strong reference. There _would_ be a problem if the old skin were retained, but then again, retaining an object after disposal is not proper use. Disposal must always end the logical lifetime of the object (it's in the name of the operation), and after an object has reached the end of its logical lifetime, actively retaining it in memory is just a leak.

I think you're arguing that it's safer (for some definition of safety) to prevent an accidentally retained object from keeping other objects alive. While I don't think that this is an entirely unreasonable argument, I think it encourages a coding style that borders on wishful thinking; often accompanied by saying "it can't hurt"... sure, but that shouldn't be the criterion on which we judge code. It's a bit like trying to "help" the GC, which almost always is not what we should be doing.

> I did run some tests and could not find a problem. Still was a bit worried as it was not always clear how to correctly write the skin so it follows the best practices. I think in the past the people probably chose to rather clear/null everything to be safe. Or, I saw a comment somewhere IRC that this was done to 'help' the GC. So yeah, a bit out of place and I personally then just followed the `null everything` pattern as I saw it quite often.
> 
> Example in the super class of `TableRowSkinBase`:
> 
> https://github.com/openjdk/jfx/blob/2abb9a3b0a7e1071bea0b8bf2ca3d83e6fdf25f8/modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java#L246-L252
> 
> > No, it's not allowed to reuse a skin
> 
> Yes (although this can be achieved when extending from `Skin` and using the `install` method correctly - too much work). Just brought it up so we think about everything. But does not affect us in any way.

I'm not sure whether we're on the same page here. There's no possibility to reuse a `SkinBase`-based skin other than to implement it incorrectly. The `SkinBase.dispose()` method nulls out the control reference, so reusing a disposed skin will _always_ fail with a NPE, regardless of how the `install()` method was implemented:
```java
var skin = button.getSkin();
button.setSkin(new ButtonSkin(button));
button.setSkin(skin); // <-- NPE: "Skin does not correspond to this Control"
```

The only way to have code like this not instantly fail is by overriding `SkinBase.dispose()` and not calling the base implementation (which would not be a correct implementation of the disposal pattern).

Maran23: > I think you're arguing that it's safer

Yeah, that's what I was thinking. 
I just wanted to make sure we weren't overlooking anything regarding memory leaks (And the fact that setting fields to `null` seems to be a recurring pattern in JavaFX Skins). 
But I checked it using `JMemoryBuddy` and manually, and everything looks fine! 
Sorry for the confusion, and thanks for the explanation.

> I'm not sure whether we're on the same page here. There's no possibility to reuse a SkinBase-based skin other than to implement it incorrectly

Right.
I was referring to skins derived from `Skin` directly, since they could be programmed so that `dispose` removes the inner working and `install` reinstalls them. 
The class that comes closest to this and might work - assuming the logic resides in `install` instead - is JavaFX's `InputFieldSkin`.

But that has nothing to do with this PR. I was just wondering if it might make sense at some point to allow the reuse of all JavaFX skins, since we already have the necessary architecture in place.

Maran23 (approved):
Tested and works fine. 
The behavior matches what modern browsers do.
Skin implementations look fine and as discussed, no issues regarding potential memory leaks (I did some checks with `JMemoryBuddy` as well - all good).

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableRowSkinBase.java
mstr2: > Right. I was referring to skins derived from `Skin` directly, since they could be programmed so that `dispose` removes the inner working and `install` reinstalls them. The class that comes closest to this and might work - assuming the logic resides in `install` instead - is JavaFX's `InputFieldSkin`.
> 
> But that has nothing to do with this PR. I was just wondering if it might make sense at some point to allow the reuse of all JavaFX skins, since we already have the necessary architecture in place.

I will say that it seems a bit weird that _you_ create the skin instance, but the _control_ disposes it. I think it would have been a better design if:
1. The control creates the skin instance on your behalf, and then disposes it if you request another skin, or
2. you create the skin instance, but you are also responsible for disposal.

andy-goryachev-oracle:
> I did some checks with `JMemoryBuddy` as well - all good).

@Maran23 perhaps we should add these checks to the test suite?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableRowSkinBase.java
Maran23: That is a good point.

I've only seen one project that tried to reuse skins (though it didn't work out), so maybe that isn't a practical solution after all (changing the skin if the table is editable or not). 
It just occurred to me, but it has little to do with the PR really and probably deserves its own mailing list post - if it's even necessary, since there are other issues we should tackle first.

Maran23:
> @Maran23 perhaps we should add these checks to the test suite?

Sure, can create a ticket when I have some time. (Also had some ideas regarding `ComboBoxTest` in the PR from Christopher - maybe a ticket to add or improve some tests even?).

PR #2178, opened 2026-05-29
8331464: ChangeListener for TableView.focusWithinProperty( ) is not always called

faneeshh:
The virtual cells (like List Cell and Table Cell) call `setFocused( )` in their `updateFocus( )` method to kind of replicate the control's focus model state. This call goes through the `focused.set( )` override in Node, which adjusts the `focusWithin `reference count on all ancestor nodes.

Since the owning control ( like TableView) is already the Scene focus owner contributing a count of 1, so I believe the cell's `setFocused(true)` increments it to 2. When the Scene later moves focus away, it decrements the count back to 1 (still positive) and so `focusWithinProperty `never fires.

The fix I've used is a focusModelUpdate flag on Node. When `setFocusedViaFocusModel( )` is called, this flag suppresses the `focusWithin ` reference count adjustment inside `focused.set( )` and all normal change notifications and CSS pseudo class updates are firing as usual too.

I've updated the 6 cell types to use `setFocusedViaFocusModel( )` in their `updateFocus( )` methods too.

---------
### Progress

### Error

### Integration blocker
&nbsp;⚠️ Title mismatch between PR and JBS for issue [JDK-8331464](https://bugs.openjdk.org/browse/JDK-8331464)

### Issue
 * [JDK-8331464](https://bugs.openjdk.org/browse/JDK-8331464): ChangeListener for TableView.focusWithinProperty() is not always called (**Bug** - P4) ⚠️ Title mismatch between PR and JBS.

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2178/head:pull/2178` \
`$ git checkout pull/2178`

`$ git checkout pull/2178` \
`$ git pull https://git.openjdk.org/jfx.git pull/2178/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2178`

`$ git pr show -t 2178`

</details>

<a href="https://git.openjdk.org/jfx/pull/2178.diff">https://git.openjdk.org/jfx/pull/2178.diff</a>

</details>

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/ListCell.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/ListCell.java
@@ -575,11 +575,11 @@ public class ListCell<T> extends IndexedCell<T> {
 
         FocusModel<T> fm = listView.getFocusModel();
         if (fm == null) {
-            setFocused(false);
+            setFocusedViaFocusModel(false);
             return;
         }
 
-        setFocused(fm.isFocused(index));
+        setFocusedViaFocusModel(fm.isFocused(index));
     }
 
     private void updateEditing() {
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/TableCell.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/TableCell.java
@@ -549,7 +549,7 @@ public class TableCell<S,T> extends IndexedCell<T> {
         final boolean isFocused = isFocused();
         if (! isInCellSelectionMode()) {
             if (isFocused) {
-                setFocused(false);
+                setFocusedViaFocusModel(false);
             }
             return;
         }
@@ -561,11 +561,11 @@ public class TableCell<S,T> extends IndexedCell<T> {
 
         final TableViewFocusModel<S> fm = tableView.getFocusModel();
         if (fm == null) {
-            setFocused(false);
+            setFocusedViaFocusModel(false);
             return;
         }
 
-        setFocused(fm.isFocused(index, getTableColumn()));
+        setFocusedViaFocusModel(fm.isFocused(index, getTableColumn()));
     }
 
     private void updateEditing() {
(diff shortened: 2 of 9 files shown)


faneeshh:
image
 
That's weird...

mstr2:
You've hit on a sharp edge in the way focus handling is implemented. This is a [known issue](https://bugs.openjdk.org/browse/JDK-8317426), but we can't solve it in the way you propose. You assume that `TableView.focusWithin` should always be `false` when another control outside of the table is focused. This is a sensible assumption, but it would mean incorrectly reporting that no node within a subtree is `focused`, even if it actually is.

The problem is not with `focusWithin`, it's with `setFocused`: cells use the input-focus flag on `Node` to track their internal "focus" state, which is different from the input-focus state that is tracked by `Node`. Superficially, both concepts look similar, but they are very clearly not. The existence of the public `setFocused` method allows subclasses to break the focus model, and cells unfortunately do that. We need to fix that at some point, and it probably means deprecating `setFocused` and making it ineffective.

faneeshh:
Okay thank u for the feedback! @Maran23 told me the same but I just needed second opinions on my approach but yeah I agree with you both. You can close the PR then if u need to.

faneeshh:
I was also thinking if it be acceptable to address this at the Cell level instead? The idea would be to override `setFocused( ) `in Cell to be deprecated and introduce a package private `setFocusedByFocusModel(boolean)` on Cell that updates the :focused pseudo-class directlyy without going through Node's focus machinery or affecting focusWithin

This would also keep the existing CSS working and fix the focusWithin bug and scope all changes to controls.

Maran23:
As mentioned in the ticket, this one is tricky indeed.

We know that:
- We should deprecate `setFocused` - it is a defect, working around `requestFocus` and will not play nice with `focusWithin`
- We somehow need a way to kind of (fake) focus a node. I think in `TableCell` we could get around faking it completely as it really is there for CSS - the commit on focus loss can also work without that
- `FakeFocusTextField` is more ticky as all the `focusListener` registered in the skin need to run still while we do not steal the focus with `requestFocus`

PR #2179, opened 2026-06-01
8385666: Fix for JDK-8384806 broke combo box cell graphics

crschnick:
This is a much simpler fix for JDK-8384806 which does not have any side effects.

This restores the old code with only a one line change instead. For reference, see https://github.com/openjdk/jfx/commit/8d917ae738120e12ac12cd0957879b7c00e59b03. We now fix the issue by clearing the cell with `buttonCell.updateIndex(-1);` as using `buttonCell.setItem(null);` was causing the original issue when the item was already null.

---------

---------
### Progress

### Issue
 * [JDK-8385666](https://bugs.openjdk.org/browse/JDK-8385666): Fix for JDK-8384806 broke combo box cell graphics (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2179/head:pull/2179` \
`$ git checkout pull/2179`

`$ git checkout pull/2179` \
`$ git pull https://git.openjdk.org/jfx.git pull/2179/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2179`

`$ git pr show -t 2179`

</details>

<a href="https://git.openjdk.org/jfx/pull/2179.diff">https://git.openjdk.org/jfx/pull/2179.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2179#issuecomment-4588976445)
</details>

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
@@ -359,15 +359,16 @@ public class ComboBoxListViewSkin<T> extends ComboBoxPopupControl<T> {
         } else {
             T value = comboBox.getValue();
             int index = getIndexOfComboBoxValueInItemsList();
+
+            // This guarantees that the cell is cleared even when the
+            // value is null and contained in the items (which is allowed)
+            buttonCell.updateIndex(-1);
+
             if (index > -1) {
-                // This guarantees that the cell is cleared even when the
-                // value is null and contained in the items (which is allowed)
-                buttonCell.updateIndex(-1);
                 buttonCell.updateIndex(index);
             } else {
                 // JDK-8127575 Show the ComboBox value even though it doesn't
                 // exist in the ComboBox items list (part two of fix)
-                buttonCell.updateIndex(-1);
                 boolean empty = updateDisplayText(buttonCell, value, false);
 
                 // Note that empty boolean collected above. This is used to resolve
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
@@ -359,20 +359,24 @@ public class ComboBoxListViewSkin<T> extends ComboBoxPopupControl<T> {
         } else {
             T value = comboBox.getValue();
             int index = getIndexOfComboBoxValueInItemsList();
+            if (index > -1) {
+                buttonCell.updateIndex(-1);
+                buttonCell.updateIndex(index);
+            } else {
+                // JDK-8127575 Show the ComboBox value even though it doesn't
+                // exist in the ComboBox items list (part two of fix)
+                buttonCell.updateIndex(-1);
+                boolean empty = updateDisplayText(buttonCell, value, false);
 
-            // JDK-8127575 Show the ComboBox value even though it doesn't
-            // exist in the ComboBox items list (part two of fix)
-            buttonCell.updateIndex(index);
-            boolean empty = updateDisplayText(buttonCell, value, false);
-
-            // Note that empty boolean collected above. This is used to resolve
-            // JDK-8124141, where we were getting different styling based on whether
-            // the cell was updated via the updateIndex method above, or just
-            // by directly updating the text. We fake the pseudoclass state
-            // for empty, filled, and selected here.
-            buttonCell.pseudoClassStateChanged(PSEUDO_CLASS_EMPTY,    empty);
-            buttonCell.pseudoClassStateChanged(PSEUDO_CLASS_FILLED,   !empty);
-            buttonCell.pseudoClassStateChanged(PSEUDO_CLASS_SELECTED, true);
+                // Note that empty boolean collected above. This is used to resolve
+                // JDK-8124141, where we were getting different styling based on whether
+                // the cell was updated via the updateIndex method above, or just
+                // by directly updating the text. We fake the pseudoclass state
+                // for empty, filled, and selected here.
+                buttonCell.pseudoClassStateChanged(PSEUDO_CLASS_EMPTY,    empty);
+                buttonCell.pseudoClassStateChanged(PSEUDO_CLASS_FILLED,   !empty);
+                buttonCell.pseudoClassStateChanged(PSEUDO_CLASS_SELECTED, true);
+            }
         }
     }
 
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
@@ -360,6 +360,8 @@ public class ComboBoxListViewSkin<T> extends ComboBoxPopupControl<T> {
             T value = comboBox.getValue();
             int index = getIndexOfComboBoxValueInItemsList();
             if (index > -1) {
+                // This guarantees that the cell is cleared even when the
+                // value is null and contained in the items (which is allowed)
                 buttonCell.updateIndex(-1);
                 buttonCell.updateIndex(index);
             } else {
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
@@ -599,7 +599,8 @@ public class ComboBoxTest {
         assertNotNull(comboBox.getButtonCell().getGraphic());
         assertEquals("null-text", ((Label) comboBox.getButtonCell().getGraphic()).getText());
 
-        // Calls updateDisplayNode()
+        // A call to getDisplayNode() will call updateDisplayNode() in its implementation
+        // as that is how the skin does it
         getDisplayNode();
         assertNotNull(comboBox.getButtonCell().getGraphic());
         assertEquals("null-text", ((Label) comboBox.getButtonCell().getGraphic()).getText());
--- a/modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
+++ b/modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
@@ -411,6 +411,7 @@ public class ComboBoxTest {
 
         assertEquals("NOT NULL", cell.getText());
     }
+
     @Test
     public void testTextFieldUpdateOnStringConverterChangeWithContainedNullValue() {
         ObservableList<String> items = FXCollections.observableArrayList(null, "ITEM1", "ITEM2");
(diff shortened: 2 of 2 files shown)


crschnick:
@andy-goryachev-oracle Would it be possible to have some basic graphics setting for the combobox in the monkey tester? That would make testing easier

andy-goryachev-oracle (commented):
I've added the button cell and the cell factory options to the monkey tester

I do see some irregularities in the master, would it be possible to demonstrate the issue using the monkey tester or a standalone reproducer, for the sake of reviewers?

edit: "irregularities" were probably because I was editing the code while the MT was running, it usually works in eclipse unless lambdas or listeners are involved, and they definitely were.

andy-goryachev-oracle:
I don't see any difference between this PR and the master.

Also, I don't see any differences when changing the button cell in the MT, perhaps I did something wrong?

Maran23:
I'm also wondering in what usecase this is broken.

Background: Setting the index to a `Cell`, even if it is the very same index will trigger the `indexChanged` code. In there, we will check if the item is the same as before and not call `updateItem` then.

Resetting the index to `-1` and then back to the original index is therefore only needed when something changed that the `Cell` can not observe and we really want kind of forcefully call `updateItem`.

This is done in `VirtualFlow` (reset index, set again), because a row might not change but a cell in the row could.
Example: We change the name of a `Person`. The `Person` object is the same, but the name is not (new `String`). But in `ListCell`, things are easier, so I just wonder if there is a case we really need this (then this PR is the correct approach) or something else is off.

crschnick:
About reprodution: Set the button cell to "With Icon" and select any value from the list. Then, change the converter. This should remove the graphic. There are probably more niche cases where the graphics gets removed on master.

@Maran23 I looked through the code and it seems like the ComboBox realizes updates in the button cell by setting the value to null and then reselecting the value to force a call to updateItem(). This is correct as something happened that we can't observe, i.e. the converter changing. Touching anything on that side leads to unintended side effects like this one, so I chose to just keep it. This PR is only an effective one-line change now from the original code, much better than before.

The bug with a selected contained null value not updating is because calling setItem(null) does not reset the cell as intended if null is already the selected value. That is why it uses setIndex() now

andy-goryachev-oracle:
@crschnick could you provide step-by-step instructions how to reproduce please?  I still don't see the issue.

crschnick:
@andy-goryachev-oracle Set it to this: 
image

Then, set the converter to anything else

andy-goryachev-oracle:
thanks!  so the missing step is to have a non-editable combo box, then change the converter.

I've updated the JBS description to list the steps:

To reproduce using the monkey tester https://github.com/andy-goryachev-oracle/MonkeyTest :
- navigate to ComboBox page
- make sure the combo box editable is off
- set button cell to "With Icon", cell factory to null, converter to <default>
- select a value, observe the icon on the button
- change converter
expected: the icon remains
observed: the icon disappears

is this right?

crschnick:
Oh I forgot about specifying that one, it is just the default setting for me. MonkeyTester is saving all the settings in the background.

So yes, this is right.

andy-goryachev-oracle (approved):
The fix looks good, now that we can tell everyone what the problem was ;-)

(You may want to update the description of this PR with the steps to reproduce).

@Maran23 would you like to be the second reviewer?

Maran23:
Yes, will have a deeper look soon.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
```diff
@@ -359,20 +359,26 @@ public final void setHideOnClick(boolean value) {
...
-            buttonCell.pseudoClassStateChanged(PSEUDO_CLASS_SELECTED, true);
+            if (index > -1) {
+                // This guarantees that the cell is cleared even when the
+                // value is null and contained in the items (which is allowed)
+                buttonCell.updateIndex(-1);
+                buttonCell.updateIndex(index);
+            } else {
+                // JDK-8127575 Show the ComboBox value even though it doesn't
+                // exist in the ComboBox items list (part two of fix)
+                buttonCell.updateIndex(-1);
```
arapte: a very quick comment,  `buttonCell.updateIndex(-1); ` is common in both if and else block, may be move it before the if-else block, and use appropriate comments to explain the behavior ?

Maran23:
Can we instead do something like this in `ComboBoxListViewSkin`:

```
...
        lh.addChangeListener(control.converterProperty(), e -> {
            listView.refresh();
            buttonCell.updateIndex(-1); // <--- NEW
            updateDisplayNode();
        });
...
```

This makes more sense to me. I understand the problem and your reasoning. 
If the converter changes, we want to refresh the `ListView`. But we should also force the `buttonCell` to refresh, hence we should reset to -1. 

Then you can revert `updateDisplayNode` to not call `buttonCell.updateIndex(-1);`. All tests still pass for me and the fix makes more sense - in this case the `buttonCell` is 100% stale and we need to consider that.

hjohn:
The code that sets the list index to -1 is already a hack, and only really works for cells that haven't been modified (especially the updateDisplayNode part). This part just hard assumes the cell is a plain text label:

                // JDK-8127575 Show the ComboBox value even though it doesn't
                // exist in the ComboBox items list (part two of fix)
                buttonCell.updateIndex(-1);
                boolean empty = updateDisplayText(buttonCell, value, false);
                
Some ComboBoxes do customize their cell display (adding additional things like tags/descriptions or a small image) and updating just the text (and not all the other stuff it may have) therefore is going to give you a cell that doesn't behave like your customized cell at all.

This issue details the problem: https://bugs.openjdk.org/browse/JDK-8221722

crschnick:
@Maran23 For reviewing, this PR is essential a revert of https://github.com/openjdk/jfx/commit/8d917ae738120e12ac12cd0957879b7c00e59b03 and a redo of the fix but simpler. I want to keep it minimal now as I now know how sensitive the combo box is to any changes, especially for custom button cells. I think the same bug will occur when you change the prompt text (https://github.com/openjdk/jfx/blob/7378d4079910653763626ce6e707b00adbec1459/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java#L175), so just adding the updateIndex() call in one listener is only a partial fix.

@hjohn I agree, the whole skin is a hack. I got burned by changing too much, that is why this issue exists in the first place. It would be easier to just create a `ComboBoxListViewSkinV2` for all the larger issues.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
```diff
@@ -492,6 +493,54 @@ public String fromString(String string) {
...
+
+            @Override
+            public String fromString(String string) {
+                return "?";
+            }
+        });
+        assertEquals("null-text", ((Label) comboBox.getButtonCell().getGraphic()).getText());
+
+        comboBox.setValue("ITEM1");
+        assertEquals("ITEM1", ((Label) comboBox.getButtonCell().getGraphic()).getText());
```
Maran23, line 572: `getDisplayNode()`

Maran23 (commented):
Reverting is fine and I agree updating the index is better then manually setting the item. @arapte suggested an improvement that I also would like to see.

I still think the `buttonCell` should be updated to -1. I don't think this affects the prompt text - in any case, we could agree to do that later as a follow up. We have many more tests in this area, and the code is less hacky, so overall a win.

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
crschnick: Actually, the call to getDisplayNode() calls updateDisplayNode() in it, so to not skew the tests, the current version is better imo

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
crschnick: Done

crschnick:
@Maran23 I updated the test case to show that setting the prompt text is also affected. It now runs multiple methods that all call updateDisplayNode() in some way, which breaks in the same way

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
```diff
@@ -492,6 +493,61 @@ public String fromString(String string) {
...
+        // if the updateDisplayNode() method is misbehaving like for JDK-8384806.
+        // Force update the cell again to have a graphic
+        // so that we can test the actual methods below
+        comboBox.getButtonCell().updateIndex(-1);
+        comboBox.getButtonCell().updateIndex(0);
+
+        // All of those invocations call updateDisplayNode()
+        // They should not change the button cell graphic to null
+        List<Runnable> methods = List.of(
+                () -> comboBox.setPromptText("abc"),
```
Maran23: I think this is probably better as a new separate test. Now we are testing two different operations in one test. Even we end up copy pasting much of this test, I would like a second test only for the `setPromptText` much more

crschnick: Alright I split it into multiple similar test methods

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
```diff
@@ -492,6 +493,124 @@ public String fromString(String string) {
...
+        // so that we can test the actual methods below
+        comboBox.getButtonCell().updateIndex(-1);
+        comboBox.getButtonCell().updateIndex(0);
+
+        comboBox.setValue(null);
+        assertNotNull(comboBox.getButtonCell().getGraphic());
+        assertEquals("null-text", ((Label) comboBox.getButtonCell().getGraphic()).getText());
+
+        // Calls updateDisplayNode()
+        getDisplayNode();
```
Maran23, line 604: This looks a bit weird - but should be the `buttonCell`, right?

So you may can just do: `ListCell<?> buttonCell = getDisplayNode()`
and then use it below for the assertions
```
        assertNotNull(buttonCell.getGraphic());
        assertEquals("null-text", ((Label) buttonCell.getGraphic()).getText());

        comboBox.setValue("ITEM1");
        assertEquals("ITEM1", ((Label) buttonCell.getGraphic()).getText());

        comboBox.setValue("ITEM2");
        assertEquals("ITEM2", ((Label) buttonCell.getGraphic()).getText());
```

crschnick: I think I mentioned that before, this is done to prevent calls to updateDisplayNode(). E.g. the method getDisplayNode() calls updateDisplayNode(), which would defeat the point of the test

Maran23: Yes, thats why my idea is to assign it once here and just reuse it below

crschnick: I mean at this point we have 3 almost identical duplicated tests, I don't think that would make a big difference when it comes to code quality. Especially considering that one still needs to be aware of the side effect of calling that method. If I don't call it, I don't need to think about it

Maran23: Okay, I wonder if we should either replace or rename the method in the test then. If it would just be called updateXXX and returns a display node, that will look better and is more honest about the side effect (because the PR before I was not aware of the side effect - so it already tricked me when reading the test)

I think having 3 similar tests is not a problem, we have that a lot for tables as well. There is just no better way, many tests often require a similar setup. The advantage is that if some fail, you know exactly what is broken (unless they are called jdkXYZ or rtXYZ).

crschnick: I updated the comment, maybe that is better

Maran23 (approved):

andy-goryachev-oracle (approved):

modules/javafx.controls/src/test/java/test/javafx/scene/control/ComboBoxTest.java
Maran23: For me it's fine, but it probably makes sense to refine the tests in a follow up. 
The most simple thing is to rename (only in the test class) `getDisplayNode()` to `updateDisplayNode()`. And maybe we find a solution to make the tests smaller, if the initial setup is similar, we could extract the `ComboBox` setup.

Example what I did in the past here:

https://github.com/openjdk/jfx/blob/b1eccb6ed0d984ca3e854b6f46163d56bd8f84d1/modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/TableRowSkinTest.java#L298-L317

PR #2182, opened 2026-06-04
JDK 8210037 : ComboBoxes in GridPane resize when popup opens

faneeshh:
So the issue was that when two or more ComboBox controls with `Priority.ALWAYS` HGrow are placed in a GridPane, opening and closing one ComboBox's popup caused the column widths to become unequal.

The root cause I believe is in r`econfigurePopup( )` in `ComboBoxPopupControl. ` While the popup is showing, it calls `setPrefSize( )` on the popup content (the ListView) to stabilize its displayed size. This pins the ListView's preferred width to the current popup dimensions. When the popup closes and the GridPane performs its next layout pass, `computePrefWidth `in `ComboBoxListViewSkin `reads `listView.prefWidth( )` which for the ComboBox whose popup was just open returns the stale pinned value, while the other ComboBox computes dynamically. This asymmetry causes GridPane to allocate unequal column widths.

I first tried resetting setPrefSize to `USE_COMPUTED_SIZE` in the `WINDOW_HIDDEN` handler. This approach does look clean I guess but doesn't work cause the GridPane layout pass that reads computePrefWidth happens while the popup is still showing, so the reset fires too late and the asymmetric value is already committed.

The fix I've used :
Guard computePrefWidth in `ComboBoxListViewSkin`: when the popup is not showing, we compute the ListView width dynamically via a new private method `computeListViewPrefWidth( )` instead of reading the potentially stale `listView.prefWidth( )`.

**When the popup is showing, the pinned value is still used directly and this preserves the existing popup stability behavior from JDK-8116801 and JDK-8123876.**

The measurement logic in `computeListViewPrefWidth` was already present in the anonymous ListView subclass's computePrefWidth override. All I did was extract it into a named method to eliminate the duplication.

---------

---------
### Progress
- \[ \] Commit message must refer to an issue

### Error

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2182/head:pull/2182` \
`$ git checkout pull/2182`

`$ git checkout pull/2182` \
`$ git pull https://git.openjdk.org/jfx.git pull/2182/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2182`

`$ git pr show -t 2182`

</details>

<a href="https://git.openjdk.org/jfx/pull/2182.diff">https://git.openjdk.org/jfx/pull/2182.diff</a>

</details>

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
@@ -289,13 +289,26 @@ public class ComboBoxListViewSkin<T> extends ComboBoxPopupControl<T> {
 
     /** {@inheritDoc} */
     @Override protected double computePrefWidth(double height, double topInset, double rightInset, double bottomInset, double leftInset) {
-        double superPrefWidth = super.computePrefWidth(height, topInset, rightInset, bottomInset, leftInset);
-        double listViewWidth = listView.prefWidth(height);
-        double pw = Math.max(superPrefWidth, listViewWidth);
-
         reconfigurePopup();
-
-        return pw;
+        double superPrefWidth = super.computePrefWidth(height, topInset, rightInset, bottomInset, leftInset);
+        // reconfigurePopup() calls setPrefSize(newWidth, newHeight) on the listView to
+        // stabilise the popup's displayed size while it is open (relates to JDK-8116801, JDK-8123876).
+        // That call pins listView.prefWidth() to the popup's current pixel width. Once the
+        // popup closes, the pinned value is stale: a parent layout container (like a GridPane)
+        // that re measures this ComboBox will read the inflated popup width instead of the
+        // natural control width, producing asymmetric column widths ( related to JDK-8210037).
+        // To avoid this, bypass the potentially pinned prefWidth when the popup is not
+        // showing and compute the natural list view width directly instead.
+        double listViewWidth;
+        if (comboBox.isShowing()) {
+            listViewWidth = listView.prefWidth(height);
+        } else {
+            listViewWidth = computeListViewPrefWidth(height);
+            if (listView.getItems().isEmpty() && listView.getPlaceholder() != null) {
+                listViewWidth = Math.max(listView.getPlaceholder().prefWidth(height), listViewWidth);
+            }
+        }
+        return Math.max(superPrefWidth, listViewWidth);
     }
 
     /** {@inheritDoc} */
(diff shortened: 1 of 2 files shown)


modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
```diff
@@ -538,31 +590,7 @@ private ListView<T> createListView() {
...
-
-                    pw = Math.max(comboBox.getWidth(), skin.getMaxCellWidth(rowsToMeasure) + 30);
-                } else {
-                    pw = Math.max(100, comboBox.getWidth());
-                }
-
-                // need to check the ListView pref height in the case that the
-                // placeholder node is showing
-                if (getItems().isEmpty() && getPlaceholder() != null) {
-                    pw = Math.max(super.computePrefWidth(height), pw);
```
Maran23, line 569: The logic is a little bit different as the inner `ListView#computePrefWidth` did a super call, which you replace with a `skin.computePrefWidth` call.

I rather wonder if we should keep that logic there.

Something like:

```
var width = computeListViewPrefWidth(height);
if (getItems().isEmpty() && getPlaceholder() != null) {
    width = Math.max(super.computePrefWidth(height), width);
}
```

and of course also adjust the other caller of `computeListViewPrefWidth` as well.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
```diff
@@ -381,6 +388,51 @@ public final void setHideOnClick(boolean value) {
         return behavior;
     }
 
+    /**
+     * Computes the preferred width of the popup {@link ListView} by measuring
+     * actual cell content, without relying on any pref size value that may have
+     * been pinned by a prior {@link #reconfigurePopup()} call.
+     *
+     * <p>This computation is shared by two callers:
```
Maran23: It does not really matter from where this is called - this is usually not interesting and therefore need to be in the javadoc.

Rather explain what the method does and why.
This code somewhat reminds me of the measurement done in tables:

https://github.com/openjdk/jfx/blob/8020a8595426c475ee29a1d09f0abdde6a58e1cf/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableColumnHeader.java#L617-L630

So maybe have a look there as well.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
```diff
@@ -289,13 +289,20 @@ public final void setHideOnClick(boolean value) {
 
     /** {@inheritDoc} */
     @Override protected double computePrefWidth(double height, double topInset, double rightInset, double bottomInset, double leftInset) {
-        double superPrefWidth = super.computePrefWidth(height, topInset, rightInset, bottomInset, leftInset);
-        double listViewWidth = listView.prefWidth(height);
```
Maran23, line 293: Isn't this calling `ListView#computePrefWidth` which will call the measurement logic (now `computeListViewPrefWidth`)? 
Normally this is how it works in JavaFX, so this makes me wonder why we even need to call `computeListViewPrefWidth` again?

Maran23 (changes requested):
some initial comments and thoughts to better understand the problem and fix.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
faneeshh: `Region.prefWidth( ) `is final and when `reconfigurePopup( )` calls `listView.setPrefSize(newWidth, newHeight)`, that stores `newWidth `in the pref width property. Every subsequent `listView.prefWidth(h)` call returns that stored value directly nd `computePrefWidth( )` isn't called then ever. The override in the anonymous `ListView` subclass is irrelevant once the pref has been pinned. That is exactly why calling `computeListViewPrefWidth( )` directly is necessary.

In short : it bypasses the short circuit in `Region.prefWidth( )` and checks the measurement logic unconditionally

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
faneeshh: Yeah I overlooked that. Good catch

modules/javafx.controls/src/main/java/javafx/scene/control/skin/ComboBoxListViewSkin.java
faneeshh: Okay will do!

PR #2184, opened 2026-06-08
8385959: Prompt text should be visible in empty focused TextInputControls

Ziad-Mid:
This fixes a `TextInputControl` styling issue where prompt text is hidden when an empty text input control receives focus.
The fix removes the focused-state override `-fx-prompt-text-fill: transparent;` so focused empty text inputs continue to use the normal Modena prompt text fill.
The prompt text now stay visible when a text input control is focused.

---------

---------
### Progress

### Issue
 * [JDK-8385959](https://bugs.openjdk.org/browse/JDK-8385959): Prompt text should be visible in empty focused TextInputControls (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2184/head:pull/2184` \
`$ git checkout pull/2184`

`$ git checkout pull/2184` \
`$ git pull https://git.openjdk.org/jfx.git pull/2184/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2184`

`$ git pr show -t 2184`

</details>

<a href="https://git.openjdk.org/jfx/pull/2184.diff">https://git.openjdk.org/jfx/pull/2184.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2184#issuecomment-4652358155)
</details>


modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/skin/modena/TextInputControlModenaTest.java
```diff
@@ -94,6 +99,14 @@ private void removeUncaughtExceptionHandler() {
         Thread.currentThread().setUncaughtExceptionHandler(null);
     }
 
+    private static Collection<Class<? extends TextInputControl>> promptTextControls() {
```
andy-goryachev-oracle: there is already `parameters()` with these values in L62

Ziad-Mid: Oh my bad I missed it and created a new one, I will remove the one I added.

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/skin/modena/TextInputControlModenaTest.java
```diff
@@ -136,4 +142,46 @@ public void testHighlightTextInput(String accentColor, String expectedTextColor)
...
+    }
+
+    private static Text getPromptNode(TextInputControl control) {
+        if (control instanceof TextArea) {
+            return TextInputSkinShim.getPromptNode((TextArea) control);
+        }
+
+        return TextInputSkinShim.getPromptNode((TextField) control);
+    }
+
```
andy-goryachev-oracle: minor: extra newline

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/skin/modena/TextInputControlModenaTest.java
```diff
@@ -136,4 +142,46 @@ public void testHighlightTextInput(String accentColor, String expectedTextColor)
...
+            control.setPromptText("Prompt text");
+
+            StackPane root = new StackPane(control);
+            Scene scene = new Scene(root, 400, 200);
+            stage.setScene(scene);
+            stage.show();
+
+            control.requestFocus();
+            Toolkit.getToolkit().firePulse();
+            control.applyCss();
```
andy-goryachev-oracle: is this necessary?  it should work without `applyCss()`, I think.

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/skin/modena/TextInputControlModenaTest.java
```diff
@@ -136,4 +142,46 @@ public void testHighlightTextInput(String accentColor, String expectedTextColor)
...
+            assertTrue(promptNode.isVisible(), type.getSimpleName() + " prompt text should be visible");
+            assertNotEquals(Color.TRANSPARENT, promptNode.getFill(),
+                    type.getSimpleName() + " prompt text fill should not be transparent");
+        } finally {
+            stage.hide();
+            Application.setUserAgentStylesheet(userAgentStylesheet);
+        }
+    }
+
+    private static Text getPromptNode(TextInputControl control) {
```
andy-goryachev-oracle: I wonder if we should move it to the `TextInputSkinShim` instead, something like this:
```
    public static Text getPromptNode(TextInputControl c) {
var skin = c.getSkin();
if(skin instanceof TextFieldSkin skin) {
        return skin.getPromptNode();
    } else if(skin instanceof TextAreaSkin skin) {
        return skin.getPromptNode();
    }
...
```

it will impact other tests, but probably cleaner long term.

andy-goryachev-oracle (commented):
the fix looks good!
some minor comments related to the test.

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/skin/modena/TextInputControlModenaTest.java
Ziad-Mid: Yes not really necessary it just forces CSS to be applied, it works fine without it too.

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/skin/modena/TextInputControlModenaTest.java
Ziad-Mid: Yes it is cleaner, currently there is a `getPromptNode()` method for `TextField` and `TextArea` only in `TextInputSkinShim`, but adding this new method would require us to make a change to many tests, maybe we can do this in another ticket and refactor it? what do you think ?

andy-goryachev-oracle: No, it will not touch the other tests, only their execution.
Basically, instead of two methods `getPromptNode(TextField)` and `getPromptNode(TextArea)` in the shim, you'll have one - `getPromptNode(TextInputControl)`.  No other changes are needed.

Ziad-Mid: Yes I see , I can implement it right here then

Ziad-Mid:
I have added method `getPromptNode(TextInputControl control) ` to `TextInputSkinShim` and removed the old once and fixed minor issues suggested.

andy-goryachev-oracle (approved):
perfect, thanks!

Ziad-Mid:
@arapte or @kevinrushforth could one of you please be second reviewer ?

arapte (approved):

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/skin/modena/TextInputControlModenaTest.java
```diff
@@ -136,4 +141,36 @@ public void testHighlightTextInput(String accentColor, String expectedTextColor)
...
 
+    @ParameterizedTest
+    @MethodSource("parameters")
+    public void promptTextIsVisibleWhenEmptyFocusedTextInput(Class<? extends TextInputControl> type)
+            throws Exception {
+        String userAgentStylesheet = Application.getUserAgentStylesheet();
+        Stage stage = new Stage();
+
+        try {
+            Application.setUserAgentStylesheet(Application.STYLESHEET_MODENA);
```
Maran23: Why is that done here? Never saw it before so I just removed it locally and the test still pass and fails without the fix.

Maran23 (changes requested):
Copyright header was not adjusted but since @arapte runs a script at one point -> probably optional.

Fix is simple, tests looks good but I wonder about the user agent stylesheet.

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/skin/modena/TextInputControlModenaTest.java
Maran23: Ok I think I know what may was going on. 

There is another test that will not correctly reset the application user agent stylesheet.
I created https://bugs.openjdk.org/browse/JDK-8387020 with more details and will create a PR ~soon~. done.

In my opinion, we should remove `Application.setUserAgentStylesheet(Application.STYLESHEET_MODENA);` here, because for one, it's the default anyway, and if another test incorrectly sets the user-agent stylesheet, this test shouldn't hide that - it should fail.

andy-goryachev-oracle: I think this is good as is - the bug calls for Modena.css specifically, so the test ensures the right stylesheet.
And thanks for filing JDK-8387020 - it is indeed a weird choice of data url.

Maran23: Okay but then we may should do that in the `setup` and reset in `cleanup` perhaps? Because the other test `testHighlightTextInput` failed because of JDK-8387020, so it is also affected when modena is not set.

andy-goryachev-oracle: Right, but the test failures are a separate problem that you've addressed in JDK-8387020.  This test is correct in its setting the specific stylesheet, while the rest are probably ok because they are supposed to work with the default stylesheet (it happens to be modena.css today, but who knows what happens tomorrow).
Right?

Maran23: Yes correct.

But what I mean is that the preexisting test `TextInputControlModenaTest#testHighlightTextInput` failed as well without JDK-8387020.
And this test does not set the modena stylesheet with `Application.setUserAgentStylesheet(Application.STYLESHEET_MODENA);` but should do as well because it will fail if another stylesheet is used tomorrow.

Thats why I propose to either remove `Application.setUserAgentStylesheet(Application.STYLESHEET_MODENA);` or set it in an `@BeforeEach` method - here in `TextInputControlModenaTest` and reset in an `@AfterEach`, because not just this new test depends on modena, but the existing one does as well.

Ziad-Mid: But a false cleanup from another test (in this case the test mentioned in JDK-8387020) that changes the Application user agent stylesheet value doesn't really mean that our test should fail no ? as it is a wrong implementation i guess in the that test. I tried it too and yes `TextInputControlModenaTest#testHighlightTextInput` fails  without JDK-8387020 as it uses the previous value. Maybe with that new fix it is fine to remove `Application.setUserAgentStylesheet(Application.STYLESHEET_MODENA);`

Maran23: Right, either 
- remove: `Application.setUserAgentStylesheet(Application.STYLESHEET_MODENA);` from here
- or add it to the `TextInputControlModenaTest#testHighlightTextInput` test (or better in `@BeforeEach` with a reset in `@AfterEach`, so it will run for all tests in `TextInputControlModenaTest` now and in the future)

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/skin/modena/TextInputControlModenaTest.java
```diff
@@ -136,4 +141,34 @@ public void testHighlightTextInput(String accentColor, String expectedTextColor)
         }
     }
 
+    @ParameterizedTest
+    @MethodSource("parameters")
+    public void promptTextIsVisibleWhenEmptyFocusedTextInput(Class<? extends TextInputControl> type)
+            throws Exception {
+        String userAgentStylesheet = Application.getUserAgentStylesheet();
```
andy-goryachev-oracle: If we are not setting the stylesheet, the lines L148 and L171 are not necessary and should be removed, right?

Stepping back, I think there are a couple of pre-existing issues with this test in general, something that we might want to address in a follow-up:

1. the test name contains Modena yet it does not set `modena.css` explicitly.  Once/if the default stylesheet changes, we'll have to rename the test, right?
2. we may run the newly added test with `caspian.css` as well, just to make sure. Strictly speaking, the bug says we don't have this problem with caspian, but there is a value in having this test as a regression test.  So we may want to consider extracting this test into a separate class, one that iterates over supported stylesheets.

Maran23: Yes, both are unused now.

I wonder how many tests will may fail if the application Stylesheet changed. Probably not that much, but I think it is worth exploring.

mstr2: I really don't like this ad-hoc way of testing some random aspect of a style theme. Maybe we should have a way to test themes more systematically, for example by comparing a scene graph with a style theme against a known reference image. We could probably use the headless platform to render a scene graph in a platform-independent way. Of course, this would also require us to have some tooling to produce the reference images.

Ziad-Mid: The test fails when setting `Application.setUserAgentStylesheet(Application.STYLESHEET_CASPIAN);` to use caspian stylesheet as in `caspian.css` we have same issue as modena it sets the value of `-fx-prompt-text-fill` to transparent

```
.text-field:focused {
...
    -fx-prompt-text-fill: transparent;
}
.text-area:focused {
...
    -fx-prompt-text-fill: transparent;
}
```

Ziad-Mid: the test succeeds when removing those lines as we did in modena

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/skin/modena/TextInputControlModenaTest.java
```diff
@@ -67,36 +71,28 @@ private static Collection<Class> parameters() {
...
-        setUncaughtExceptionHandler();
-        try {
-            textInput = (TextInputControl)type.getDeclaredConstructor().newInstance();
-        } catch (Exception e) {
-            fail(e);
-        }
+    @BeforeEach
+    public void setup() {
+        userAgentStylesheet = Application.getUserAgentStylesheet();
+        Application.setUserAgentStylesheet(Application.STYLESHEET_MODENA);
```
andy-goryachev-oracle, line 77: we could probably do this in `@BeforeClass`, but it does not matter (much), only makes the test run slightly longer.

andy-goryachev-oracle (approved):
Left a minor suggestion, but this lgtm as is.

modules/javafx.controls/src/main/resources/com/sun/javafx/scene/control/skin/modena/modena.css
```diff
@@ -1300,7 +1300,6 @@ is being used to size a border should also be in pixels.
         linear-gradient(from 0px 0px to 0px 5px, derive(-fx-control-inner-background, -9%), -fx-control-inner-background);
     -fx-background-insets: -0.2, 1, -1.4, 3;
     -fx-background-radius: 3, 2, 4, 0;
-    -fx-prompt-text-fill: transparent;
```
Maran23, line 1303: One note here, maybe @andy-goryachev-oracle can tell:

This is still done for the RTA:

```
.rich-text-area:focused {
    -fx-prompt-text-fill: transparent;
}
```

Is this something we should solve as well?

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/skin/modena/TextInputControlModenaTest.java
arapte: something like this would be good to do
```
    @BeforeAll
    public static void setup() {
        userAgentStylesheet = Application.getUserAgentStylesheet();
        Application.setUserAgentStylesheet(Application.STYLESHEET_MODENA);
    }

    @AfterAll
    public static void cleanup() {
        Application.setUserAgentStylesheet(userAgentStylesheet);
    }

    @AfterEach
    public void cleanupTest() {
        if (stage != null) {
            stage.hide();
            stage = null;
        }
    }
```

modules/javafx.controls/src/main/resources/com/sun/javafx/scene/control/skin/modena/modena.css
andy-goryachev-oracle: good catch, but there is no prompt text in the `RichTextArea` yet.
I can deal with it in https://bugs.openjdk.org/browse/JDK-8359862

Maran23: Sounds good!

arapte (approved):

andy-goryachev-oracle (approved):
lgtm

@Maran23 are you ok with the latest changes?

Maran23 (approved):
Looks good.

Ziad-Mid:
Thank you for your reviews !

PR #2186, opened 2026-06-10
8340564: [macOS] Intermittent build failure in native font with parallel compilation

kevinrushforth:
This PR fixes a long-standing intermittent build failure on macOS (although it could have hit any platform, we never saw it on Windows or Linux). The problem was a race condition in the custom Groovy CCTask used by gradle to compile native C / C++ code. A set of C/C++ command line options is passed in to the custom CCTask and in the case of C++ files, mutated (to remove an unwanted option). The bug is that the original list, which is passed in to each invocation of CCTask is mutated rather than making a copy and mutating that. When compilation tasks are run in parallel (which they are by default), this can lead to the exception if one thread mutates the list while another thread reads the list.

Even without the concurrency error, mutating the input list of options directly is wrong, and should never have been done. The fix is to make a copy of the input args and operate on the copy.

I was able to provoke this bug on my local system by creating a few extra dummy files in the native-font directory and then running `gradle cleanNativeFont; gradle :graphics:ccMacFont` in a loop. Without the fix it fails for me after about 10-30 iterations. With the fix, I've run > 100 and still running.

---------

---------
### Progress

### Issue
 * [JDK-8340564](https://bugs.openjdk.org/browse/JDK-8340564): [macOS] Intermittent build failure in native font with parallel compilation (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2186/head:pull/2186` \
`$ git checkout pull/2186`

`$ git checkout pull/2186` \
`$ git pull https://git.openjdk.org/jfx.git pull/2186/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2186`

`$ git pr show -t 2186`

</details>

<a href="https://git.openjdk.org/jfx/pull/2186.diff">https://git.openjdk.org/jfx/pull/2186.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2186#issuecomment-4670393917)
</details>

### DIFF
--- a/buildSrc/src/main/groovy/com/sun/javafx/gradle/CCTask.groovy
+++ b/buildSrc/src/main/groovy/com/sun/javafx/gradle/CCTask.groovy
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013, 2025, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2013, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -65,11 +65,11 @@ class CCTask extends NativeCompileTask {
             if (params != null) {
                 // A little hack. Only use the -std=c99 flag if compiling .c or .m
                 if (sourceFile.name.endsWith(".cpp") || sourceFile.name.endsWith(".cc") || sourceFile.name.endsWith(".mm")) {
-                    def stripped = params;
+                    def stripped = new ArrayList<String>(params);
                     stripped.remove("-std=c99");
                     spec.args(stripped)
                 } else {
-                    spec.args(params)
+                    spec.args(new ArrayList<String>(params));
                 }
             }
 


buildSrc/src/main/groovy/com/sun/javafx/gradle/CCTask.groovy
```diff
@@ -65,11 +65,11 @@ class CCTask extends NativeCompileTask {
             if (params != null) {
                 // A little hack. Only use the -std=c99 flag if compiling .c or .m
                 if (sourceFile.name.endsWith(".cpp") || sourceFile.name.endsWith(".cc") || sourceFile.name.endsWith(".mm")) {
-                    def stripped = params;
+                    def stripped = new ArrayList<String>(params);
```
andy-goryachev-oracle, line 68: pre-existing: the comment in L66 does not exactly correspond to code...

andy-goryachev-oracle (approved):

buildSrc/src/main/groovy/com/sun/javafx/gradle/CCTask.groovy
kevinrushforth: Right. I decided to leave it, since it's conceptually accurate, if imprecise.

kevinrushforth:

buildSrc/src/main/groovy/com/sun/javafx/gradle/CCTask.groovy
andy-goryachev-oracle: it will break the moment someone adds a .c++ file, but right now we are ok.

kevinrushforth: I suspect other things could break if we try to use a weird extension like `c++`. However, this is fragile, so point taken.

FWIW, this is a total hack anyway. Any such filtering should be done in the gradle files, not here. A follow-up might be in order, although I doubt it would be high priority.

arapte (approved):
lgtm+

PR #2187, opened 2026-06-12
8386590: GridPane - content-biased child spanning multiple columns/rows is sized ignoring hgap/vgap

FlorianKirmaier:
This PR fixes the computeSize methods not considering the gap in some cases.

It also fixes a missed spot, in the previous PR for JDK-8092379.

Each fix is covered by simple tests which fail before and work after the change.

---------

---------
### Progress

### Issue
 * [JDK-8386590](https://bugs.openjdk.org/browse/JDK-8386590): GridPane - content-biased child spanning multiple columns/rows is sized ignoring hgap/vgap (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2187/head:pull/2187` \
`$ git checkout pull/2187`

`$ git checkout pull/2187` \
`$ git pull https://git.openjdk.org/jfx.git pull/2187/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2187`

`$ git pr show -t 2187`

</details>

<a href="https://git.openjdk.org/jfx/pull/2187.diff">https://git.openjdk.org/jfx/pull/2187.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2187#issuecomment-4689803859)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/layout/GridPane.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/layout/GridPane.java
@@ -1230,7 +1230,7 @@ public class GridPane extends Pane {
         computeGridMetrics();
         performingLayout = true;
         try {
-            final double[] heights = height == -1 ? null : computeHeightsToFit(height).asArray();
+            final CompositeSize heights = height == -1 ? null : computeHeightsToFit(height);
 
             return snapSpaceX(getInsets().getLeft()) +
                     computeMinWidths(heights).computeTotalWithMultiSize() +
@@ -1245,7 +1245,7 @@ public class GridPane extends Pane {
         computeGridMetrics();
         performingLayout = true;
         try {
-            final double[] widths = width == -1 ? null : computeWidthsToFit(width).asArray();
+            final CompositeSize widths = width == -1 ? null : computeWidthsToFit(width);
 
             return snapSpaceY(getInsets().getTop()) +
                     computeMinHeights(widths).computeTotalWithMultiSize() +
@@ -1259,7 +1259,7 @@ public class GridPane extends Pane {
         computeGridMetrics();
         performingLayout = true;
         try {
-            final double[] heights = height == -1 ? null : computeHeightsToFit(height).asArray();
+            final CompositeSize heights = height == -1 ? null : computeHeightsToFit(height);
 
             return snapSpaceX(getInsets().getLeft()) +
                     computePrefWidths(heights).computeTotalWithMultiSize() +
@@ -1273,7 +1273,7 @@ public class GridPane extends Pane {
         computeGridMetrics();
         performingLayout = true;
         try {
-            final double[] widths = width == -1 ? null : computeWidthsToFit(width).asArray();
+            final CompositeSize widths = width == -1 ? null : computeWidthsToFit(width);
 
             return snapSpaceY(getInsets().getTop()) +
                     computePrefHeights(widths).computeTotalWithMultiSize() +
@@ -1369,15 +1369,11 @@ public class GridPane extends Pane {
         return true;
     }
 
-    private double getTotalWidthOfNodeColumns(Node child, double[] widths) {
+    private double getTotalWidthOfNodeColumns(Node child, CompositeSize widths) {
         if (getNodeColumnSpan(child) == 1) {
-            return widths[getNodeColumnIndex(child)];
+            return widths.getSize(getNodeColumnIndex(child));
         } else {
-            double total = 0;
-            for (int i = getNodeColumnIndex(child), last = getNodeColumnEndConvertRemaining(child); i <= last; ++i) {
-                total += widths[i];
-            }
-            return total;
+            return widths.computeTotal(getNodeColumnIndex(child), getNodeColumnEndConvertRemaining(child) + 1);
         }
     }
 
@@ -1411,7 +1407,7 @@ public class GridPane extends Pane {
         return rowMaxHeight;
     }
 
-    private CompositeSize computePrefHeights(double[] widths) {
+    private CompositeSize computePrefHeights(CompositeSize widths) {
         CompositeSize result;
         if (widths == null) {
             if (rowPrefHeight != null) {
@@ -1465,7 +1461,7 @@ public class GridPane extends Pane {
         return result;
     }
 
-    private CompositeSize computeMinHeights(double[] widths) {
+    private CompositeSize computeMinHeights(CompositeSize widths) {
         CompositeSize result;
         if (widths == null) {
             if (rowMinHeight != null) {
@@ -1511,15 +1507,11 @@ public class GridPane extends Pane {
         return result;
     }
 
-    private double getTotalHeightOfNodeRows(Node child, double[] heights) {
+    private double getTotalHeightOfNodeRows(Node child, CompositeSize heights) {
         if (getNodeRowSpan(child) == 1) {
-            return heights[getNodeRowIndex(child)];
+            return heights.getSize(getNodeRowIndex(child));
         } else {
-            double total = 0;
-            for (int i = getNodeRowIndex(child), last = getNodeRowEndConvertRemaining(child); i <= last; ++i) {
-                total += heights[i];
-            }
-            return total;
+            return heights.computeTotal(getNodeRowIndex(child), getNodeRowEndConvertRemaining(child) + 1);
         }
     }
 
@@ -1553,7 +1545,7 @@ public class GridPane extends Pane {
         return columnMaxWidth;
     }
 
-    private CompositeSize computePrefWidths(double[] heights) {
+    private CompositeSize computePrefWidths(CompositeSize heights) {
         CompositeSize result;
         if (heights == null) {
             if (columnPrefWidth != null) {
@@ -1610,7 +1602,7 @@ public class GridPane extends Pane {
         return result;
     }
 
-    private CompositeSize computeMinWidths(double[] heights) {
+    private CompositeSize computeMinWidths(CompositeSize heights) {
         CompositeSize result;
         if (heights == null) {
             if (columnMinWidth != null) {
@@ -1741,12 +1733,12 @@ public class GridPane extends Pane {
             } else if (contentBias == Orientation.HORIZONTAL) {
                 widths = (CompositeSize) computePrefWidths(null).clone();
                 columnTotal = adjustColumnWidths(widths, width);
-                heights = computePrefHeights(widths.asArray());
+                heights = computePrefHeights(widths);
                 rowTotal = adjustRowHeights(heights, height);
             } else {
                 heights = (CompositeSize) computePrefHeights(null).clone();
                 rowTotal = adjustRowHeights(heights, height);
-                widths = computePrefWidths(heights.asArray());
+                widths = computePrefWidths(heights);
                 columnTotal = adjustColumnWidths(widths, width);
             }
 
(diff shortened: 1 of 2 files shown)


modules/javafx.graphics/src/main/java/javafx/scene/layout/GridPane.java
```diff
@@ -1820,8 +1812,9 @@ private CompositeSize computeWidthsToFit(double width) {
                                         cs = widths.getLength() - c;
                                     }
                                     double w = widths.getSize(c);
-                                    for (int j = 2; j <= cs; j++) {
-                                        w += widths.getSize(c + j - 1) + snaphgap;
+                                    for (int j = c + 1; j < cs + c; j++) {
+                                        w += widths.hasGapBefore(j) ? snaphgap : 0;
+                                        w += widths.getSize(j);
```
hjohn, line 1817: This was the spot I missed in JDK-8092379 I guess? :)

FlorianKirmaier: Yes.

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/GridPaneTest.java
```diff
@@ -3169,4 +3169,82 @@ public void testGridPaneHgapVgapConstructor() {
...
+
+        for (int i = 0; i < 5; i++) {
+            gridpane.add(new MockResizable(10,10,10,10,10,10), i, 1);
+        }
+
+        gridpane.resize(70, 500);
+        gridpane.layout();
+
+        assertEquals(70, fixedArea.getWidth());
+        fixedArea.checkSize();
```
Maran23: I would rather prefer an accurate assertions here and below. This also makes it more understandable what the test does IMO.

The you can also remove the `checkSize` method.

```suggestion
        assertEquals(143, fixedArea.getHeight());
```

or

```suggestion
        assertEquals(Math.round(FixedAreaRegion.AREA / 70), fixedArea.getHeight());
```

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/GridPaneTest.java
```diff
@@ -3169,4 +3169,82 @@ public void testGridPaneHgapVgapConstructor() {
...
+
+        for (int i = 0; i < 5; i++) {
+            gridpane.add(new MockResizable(10,10,10,10,10,10), 1, i);
+        }
+
+        gridpane.resize(500, 70);
+        gridpane.layout();
+
+        assertEquals(70, fixedArea.getHeight());
+        fixedArea.checkSize();
```
Maran23:
```suggestion
        assertEquals(143, fixedArea.getWidth());
```

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/GridPaneTest.java
```diff
@@ -3169,4 +3169,82 @@ public void testGridPaneHgapVgapConstructor() {
...
+        gridpane.add(fixedArea, 0, 0, 5, 1);
+
+        GridPane.setValignment(fixedArea, VPos.BASELINE);
+        GridPane.setFillHeight(fixedArea, false);
+
+        gridpane.resize(100, 500);
+        gridpane.layout();
+
+        assertEquals(100, fixedArea.getWidth());
+        fixedArea.checkSize();
```
Maran23:
```suggestion
        assertEquals(100, fixedArea.getHeight());
```

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/GridPaneTest.java
```diff
@@ -3169,4 +3169,82 @@ public void testGridPaneHgapVgapConstructor() {
         assertEquals(3, gridpane.getHgap(), 0);
         assertEquals(8, gridpane.getVgap(), 0);
     }
+
+    static class FixedAreaRegion extends Region {
+        final double AREA = 100*100;
```
Maran23:
```suggestion
        final double AREA = 10000;
```

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/GridPaneTest.java
andy-goryachev-oracle: floating point operations might be inexact, so a small tolerance is required in `assertEquals` in general case (and should be anticipated in comparisons).  

In this test, there is no chance to see fractional scale or non-integer sizes, so we are ok with exact `assertEquals`.

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/GridPaneTest.java
andy-goryachev-oracle: or perhaps declare `HEIGHT`, `WIDTH`, and `AREA = HEIGHT * WIDTH`

(but 100*100 is more meaningful in the context)

andy-goryachev-oracle (commented):
There is no reproducer, so only the code review for now.

FlorianKirmaier:
Honestly, I prefer the version where the Node itself declares the invariant.

Let me explain why I think this is a more scalable testing method.

I could just check the width and height against the values I already know.
But by checking an invariant in the Node, I could in theory create a random GridPane
configuration and verify for every Node whether its invariant is fulfilled.

This way I could basically create randomized test cases - a bit like the
Haskell QuickCheck library. For that, it's important that the check is defined in
only one place, and as part of the Node.

If we get more test cases for FixedAreaRegion, this would also make it easier to
ensure all test cases are correct.

I also have more test cases in the pipeline that would reuse the invariant from
FixedAreaRegion.

If I can't convince you, I'll change it to the suggested version.

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/GridPaneTest.java
FlorianKirmaier: I added my reasoning for using the checkSize method instead of having the asserts inside the tests.
If its not convincing, I'll change them to the suggested versions.

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/GridPaneTest.java
```diff
@@ -3169,4 +3169,83 @@ public void testGridPaneHgapVgapConstructor() {
         assertEquals(3, gridpane.getHgap(), 0);
         assertEquals(8, gridpane.getVgap(), 0);
     }
+
+    static class FixedAreaRegion extends Region {
+        final double SIDE = 100;
+        final double AREA = SIDE*SIDE;
```
Maran23:
```suggestion
        final double AREA = SIDE * SIDE;
```

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/GridPaneTest.java
```diff
@@ -3169,4 +3169,83 @@ public void testGridPaneHgapVgapConstructor() {
...
+        Orientation bias;
```
Maran23: Should also be `final`

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/GridPaneTest.java
```diff
@@ -3169,4 +3169,83 @@ public void testGridPaneHgapVgapConstructor() {
...
+        protected double computePrefWidth(double height) {
+            return height <= 0 ? SIDE : AREA / height;
+        }
+
+        @Override
+        protected double computePrefHeight(double width) {
+            return width <= 0 ? SIDE : AREA / width;
+        }
+
+        void checkSize() {
```
Maran23: If we want to keep that, I would suggest that we rename it something like `assertSize` to make clear that this is an assertion

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/GridPaneTest.java
```diff
@@ -3169,4 +3169,83 @@ public void testGridPaneHgapVgapConstructor() {
...
+            // This should work when the Node is laid out properly.
+            assertEquals(AREA, getWidth() * getHeight(), 20);
```
Maran23, line 3198: From my testig, a `delta` of 10 was enough. Thats why I was a bit confused and also suggested to use precise assertions instead. 

I understand you want to keep that for future tests, which is fine for me. 
But I'm a bit afraid that a high delta may masks some bugs where the test should fail but will pass because it fits into the delta of 20. Especially since v/hgap is usually set to 4 or 5, and we only have 2 cols/rows, a delta of 20 can be enough to e.g. hide the fact that a gap was not calculated in correctly or similar situations.

Does that makes sense?

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/GridPaneTest.java
FlorianKirmaier: done

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/GridPaneTest.java
FlorianKirmaier: done

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/GridPaneTest.java
FlorianKirmaier: Thank you for the quick response.

It only looks high because it compares the area - this basically
represents 0.2 pixels. I'm not sure whether the snap-to-pixel logic
will require a higher value in some other cases. A value up to
1.0 pixels (100) would be in a range that makes sense to me.

I could do something like the following, if that's considered more readable:
```
AverageSide = Math.sqrt(getWidth() * getHeight());
assertEquals(100, AverageSide, 0.2);
```

andy-goryachev-oracle (changes requested):
Please provide steps to reproduce in the JBS - either an application or maybe how to reproduce it in the monkey tester's GridPane.

FlorianKirmaier:
I've added a simple test-app to the Ticket - including a screenshot showing the wrong test-wrap.

andy-goryachev-oracle:
Thank you for adding a reproducer!

Unfortunately, I still not sure what I am looking at.  Could you tell what is expected vs. what is observed?

Also, the reproducer seems to contain questionable code:
there are multiple overlapping labels

Screenshot 2026-06-25 at 08 17 32


later constraints override the earlier ones:

```
result.add(innerContent, 2, 1); // col=2, row=1
... then
GridPane.setConstraints(innerContent, 0, 0, 2, 1); // col=0, row=0
```

FlorianKirmaier:
I've slightly adapted the Test-Application.

The "Overlapping elements" are required to reproduce the bug.
With 25 or before JDK-8092379 they are not required - but afterwards they are needed to reproduce the bug.

The overlapping-element no longer has text but a colored background instead. This makes the app look less buggy.
Removed the overwritten position for the main-element - which didn't have much effect but was a codesmell.

But still basically the same app.

When the app is resized so the TextFlow can render with 1 line without much space left,
then the rendering bug is visible.


With JFX26:
Screenshot 2026-06-30 at 12 04 29

With build from PR:
Screenshot 2026-06-30 at 12 05 12

andy-goryachev-oracle (approved):
Thank you for explaining what to look for - the exact use case wasn't immediately apparent.  Now I see the problem.

lgtm

arapte (approved):
lgtm+, both change and sanity testing looks good.

PR #2188, opened 2026-06-13
8386617: JavaFX controls in the title bar

mstr2:
JavaFX controls in the title bar were previewed with [JDK-8313424](https://bugs.openjdk.org/browse/JDK-8313424) (delivered in JFX 25) and the API was subsequently refined with [JDK-8369836](https://bugs.openjdk.org/browse/JDK-8369836) and [JDK-8370446](https://bugs.openjdk.org/browse/JDK-8370446) (delivered in JFX 26). The feedback collected in two preview rounds suggests that the API is fit for purpose, and no fundamental changes are required. This means that the feature is ready to become final and permanent in JFX 27.

I propose a few small changes to smooth out remaining edges:
1. **Rename `HeaderBar.prefButtonHeight` to `HeaderBar.systemButtonHeight`**
   This more clearly expresses that it is the height of the _system-provided_ buttons.
2. **Rename `HeaderBar.minSystemHeight` to `HeaderBar.systemMinHeight`**
   What is meant is not a _minimum system height_ (which makes no sense), but a _system-provided minimum height_.
3. **Add `HeaderBar.systemColorScheme`**
   In the previous iteration, the color scheme of the header buttons on Windows and Linux was determined by the brightness of the `Scene.fill` background. On macOS, which uses header buttons drawn by the operating system, the color scheme of the header buttons was instead determined by the color scheme of the window itself.

   The newly added `HeaderBar.systemColorScheme` attached property allows applications to specify the color scheme for system-provided header buttons directly, without falling back to either the scene background or the window color scheme. The default value of the property is `null`, which means "no preference"; in this case, the color scheme for header buttons will match the color scheme for the window on all platforms. Setting this property to a non-null value is only necessary in the rare case when an application needs different color schemes for header buttons and the window (i.e. when it has a bright title bar in dark mode, or a dark title bar in light mode).

Since these changes are only cosmetic (in the case of renamed properties) and don't fundamentally change the API of the enhancement (in the case of `HeaderBar.systemColorScheme`), they shouldn't need to be previewed for another round.

The rest of the significant changes in this PR are a clean-up pass over the documentation, and having the macOS header button implementation support the new `HeaderBar.systemColorScheme` property.

## Note on the use of attached properties
This enhancement uses attached properties, which JavaFX has had since its inception, in a new way. Instead of only offering a simple getter and setter, it also adds a property getter:
```java
public class HeaderBar {
    public static void setDragType(Node child, HeaderDragType value);
    public static HeaderDragType getDragType(Node child);
    public static ObjectProperty<HeaderDragType> dragTypeProperty(Node child); // <-- this is new
}
```

This allows developers to use the Property API to interact with attached properties, and in the future may allow attached properties to become CSS-styleable (#1714). This really is a missing piece that we should address, as it makes life [much easier](https://github.com/openjdk/jfx/pull/2015#issuecomment-3863351123) in several scenarios. There is also an enhancement underway to formalize attached properties in the Property API (#2015).

If we end up deciding that the Property API should _not_ support attached properties (which would be a weird decision considering that it is a natural extension with many downstream benefits), then the attached property getters in the `HeaderBar` class will look a bit out of place.

/reviewers 2
/csr

---------

---------
### Progress
- \[x\] Change requires CSR request [JDK-8387000](https://bugs.openjdk.org/browse/JDK-8387000) to be approved

### Issues
 * [JDK-8386617](https://bugs.openjdk.org/browse/JDK-8386617): JavaFX controls in the title bar (**Enhancement** - P4)
 * [JDK-8358823](https://bugs.openjdk.org/browse/JDK-8358823): Improve documentation of custom header bars (**Enhancement** - P4)
 * [JDK-8387000](https://bugs.openjdk.org/browse/JDK-8387000): JavaFX controls in the title bar (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2188/head:pull/2188` \
`$ git checkout pull/2188`

`$ git checkout pull/2188` \
`$ git pull https://git.openjdk.org/jfx.git pull/2188/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2188`

`$ git pr show -t 2188`

</details>

<a href="https://git.openjdk.org/jfx/pull/2188.diff">https://git.openjdk.org/jfx/pull/2188.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2188#issuecomment-4697139907)
</details>


palexdev:
Sorry for repeating myself but since we are talking about:
> The feedback collected in two preview rounds suggests that the API is fit for purpose

The last thing missing to make it truly production ready is to have round corners on Linux systems too. Because as it is now, apps on those systems using the `EXTENDED` style are just unreleasable!

mstr2:
> Sorry for repeating myself but since we are talking about:
> 
> > The feedback collected in two preview rounds suggests that the API is fit for purpose
> 
> The last thing missing to make it truly production ready is to have round corners on Linux systems too. Because as it is now, apps on those systems using the `EXTENDED` style are just unreleasable!

I agree, but the purpose of this PR is to finalize the API, which wouldn't change when we polish the native Linux implementation to support rounded corners.

kevinrushforth:
We don't have much time in 27 to finalize it, but let's see where we get. One thing to note: If we do finalize this in 27 (or, in 28 if it misses 27), all API elements need to change to `@since 27`. When a preview feature is finalized, the entire API is "since" the release that it went final, even if there were no changes in this round. For new classes (e.g., HeaderBar), only the class needs to retain the `@since`.

mstr2:
I've changed all API elements to `@since 27`.

andy-goryachev-oracle (commented):
I'll take a look.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/Window.java
```diff
@@ -309,11 +309,11 @@ protected Window(Window owner, Screen screen, int styleMask) {
...
      */
-    private final DoubleProperty prefHeaderButtonHeight =
-        new SimpleDoubleProperty(this, "prefHeaderButtonHeight", HeaderBar.USE_DEFAULT_SIZE);
+    private final DoubleProperty headerButtonHeight =
+        new SimpleDoubleProperty(this, "headerButtonHeight", HeaderBar.USE_DEFAULT_SIZE);
 
-    public final ReadOnlyDoubleProperty prefHeaderButtonHeightProperty() {
-        return prefHeaderButtonHeight;
+    public final ReadOnlyDoubleProperty headerButtonHeightProperty() {
+        return headerButtonHeight;
```
andy-goryachev-oracle: returns a `SimpleDoubleProperty` not `ReadOnlyDoubleProperty`

modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
andy-goryachev-oracle: rm Preview Feature

modules/javafx.graphics/src/main/java/com/sun/glass/ui/Window.java
```diff
@@ -322,8 +322,28 @@ public final ReadOnlyDoubleProperty prefHeaderButtonHeightProperty() {
...
+    }
+
+    /**
+     * Specifies whether header buttons are displayed in a style appropriate for dark mode.
+     * Sub-classes can use this value in their header button visualization, but they are not required to do so.
+     */
+    private final BooleanProperty headerButtonDarkStyle = new SimpleBooleanProperty(this, "headerButtonDarkStyle");
+
+    public final ReadOnlyBooleanProperty headerButtonDarkStyleProperty() {
+        return headerButtonDarkStyle;
```
andy-goryachev-oracle: returns not a `ReadOnlyBooleanProperty`

modules/javafx.graphics/src/main/java/javafx/scene/layout/HeaderBar.java
```diff
@@ -267,14 +269,13 @@ public static HeaderButtonType getButtonType(Node child) {
     /**
      * Specifies the {@code HeaderButtonType} of the child, indicating its semantic use in the header bar.
      * <p>
-     * This property can be set on any {@link Node}. Specifying a header button type also provides the behavior
-     * associated with the button type. If the default behavior is not desired, applications can register an
-     * event filter on the child node that consumes the {@link MouseEvent#MOUSE_RELEASED} event.
+     * This property can be set on any {@link Node}. Specifying a {@code HeaderButtonType} also provides the
```
andy-goryachev-oracle, line 264: provide a `{@link` instead of `{@code` perhaps?

modules/javafx.graphics/src/main/java/javafx/scene/layout/HeaderBar.java
```diff
@@ -304,35 +305,73 @@ protected void invalidated() {
...
+     * Gets the value of the {@link #systemButtonHeightProperty(Stage) systemButtonHeight}
+     * attached property for the specified {@code Stage}.
      *
      * @param stage the {@code Stage}
      * @return the preferred height of the system-provided header buttons
      */
-    public static double getPrefButtonHeight(Stage stage) {
-        return AttachedProperties.of(stage).prefButtonHeight.get();
+    public static double getSystemButtonHeight(Stage stage) {
+        return AttachedProperties.of(stage).systemButtonHeight.get();
```
andy-goryachev-oracle: side node: if we decide to put rarely used properties in a hashtable (see https://github.com/andy-goryachev-oracle/Test/blob/main/doc/Experiments/NodeProperties.md ) we could get rid of the `AttachedProperties` machinery (or roll it into the `NodeProperties`).

andy-goryachev-oracle (commented):
preliminary code review, more tomorrow.

modules/javafx.graphics/src/main/java/javafx/scene/layout/HeaderBar.java
mstr2: Yes, attached properties shouldn't be stored in `Node.getProperties()` at all, where they are publicly accessible and can also be changed or removed by completely unrelated code.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/SceneContext.java
```diff
@@ -123,7 +124,16 @@ public boolean isFullScreen() {
     }
 
     private final MediaProperty<ColorScheme> colorScheme = new MediaProperty<>(
-            "colorScheme", PlatformImpl.getPlatformPreferences().colorSchemeProperty());
+            "colorScheme", PlatformImpl.getPlatformPreferences().colorSchemeProperty()) {
+                @Override
+                protected void onInvalidated() {
+                    if (scene.getWindow() instanceof Stage stage) {
```
andy-goryachev-oracle: question: if the Scene had the colorScheme property set before it's added to a Stage, will the stage color scheme be updated correctly?

modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacWindow.java
```diff
@@ -49,7 +49,8 @@ protected MacWindow(Window owner, Screen screen, int styleMask) {
         super(owner, screen, styleMask);
 
         if (isExtendedWindow()) {
-            prefHeaderButtonHeightProperty().subscribe(this::onPrefHeaderButtonHeightChanged);
+            headerButtonHeightProperty().subscribe(this::onHeaderButtonHeightChanged);
+            headerButtonDarkStyleProperty().subscribe(this::onHeaderButtonDarkStyleChanged);
```
andy-goryachev-oracle, line 53: question: here we bind to the `headerButtonDarkStyleProperty`, but in `GtkWindow` and `WinWindow` we do not.  Is this right?

andy-goryachev-oracle (commented):
testing looks good on macOS, I did not test on Windows/Linux.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/SceneContext.java
mstr2: Yes, the actual scene color scheme is queried unconditionally just before the stage is shown in `Stage.isHeaderButtonDarkStyle()`. The `StageHelper.notifyColorSchemeChanged()` path is only relevant if the scene color scheme is changed while the stage is showing.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacWindow.java
mstr2: It's bound in `WinWindow.createHeaderButtonOverlay()`, but not in `GtkWindow.createHeaderButtonOverlay()`. Good catch!

andy-goryachev-oracle: what about `WinWindow` ?

mstr2: It's already used in `WinWindow`.

andy-goryachev-oracle (approved):
lgtm, needs to be re-tested on Window and Linuxes.

kevinrushforth (commented):
I reviewed the public API and all looks good.

I recommend that you update the CSR and move it to Proposed as soon as you can. The CSR should describe the differences between the existing preview API and the proposed final API. Attach the diffs of the public API. Additionally, attach the final "new in 27" API -- effectively a list of the entire final API surface.

In parallel with the CSR review, I'll review and test the implementation changes (which are fairly small and targeted).

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/SceneContext.java
kevinrushforth: ~~While its true that the mechanism will work when initially setting a scene on a stage and then showing that stage, it will fail when switching to a new scene on a stage that is already showing a different scene.~~

UPDATE: between the time I wrote this (then pending) comment and now, I see that you pushed a new update that changes the way this is done. I can confirm that this is now fixed.

mstr2:
I've decided to move `HeaderBar.AttachedProperties` to `Stage.ExtendedPropertiesImpl`. The move itself is quite straightforward and the class implementation doesn't change much; what changes is a significant removal of inter-class communication. 

Previously, some values were stored as fields directly on `Stage`, while others were stored on `HeaderBar.AttachedProperties`. Both needed to be carefully kept in sync. For example, when a property like `HeaderBar.systemButtonHeight(Stage)` was retrieved, it was initialized with the current value stored on `Stage`. Then, if the attached property value was changed by the user, its new value was pushed back to `Stage`. Moving all of the implementation to `Stage` means that most of the inter-class communciation is not needed.

A second and final change is the specification of `HeaderBar.systemColorScheme(Stage)`: previously, this was specified so that `null` means "use the scene-provided color scheme". While that works in principle, it doesn't allow users of `HeaderBar` to easily observe the scene-provided effective color scheme of the header buttons. I've changed this property to match the specification of `Scene.Preferences.colorScheme` as a _null-coalescing_ property: if set to `null`, its value resolves to the value provided by the scene. If set to a non-null value, it overrides the scene value.

modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
```diff
@@ -1303,52 +1292,154 @@ public final ObjectProperty<String> fullScreenExitHintProperty() {
...
+                        peer.setHeaderButtonHeight(get());
+                    }
+                }
+            };
+
+            this.systemColorScheme = new NullCoalescingPropertyBase<>(
+                    stage.sceneProperty()
+                        .map(Scene::getPreferences)
+                        .flatMap(Scene.Preferences::colorSchemeProperty)
+                        .orElse(ColorScheme.LIGHT)) {
```
kevinrushforth: Would it be better to map this to `PlatformImpl.getPlatformPreferences().colorSchemeProperty()`, which is the default when there is a scene, but color scheme hasn't been set?

kevinrushforth (commented):
The API change to make `HeaderBar::systemColorScheme` a null-coalescing property is good (and fixes a bug), as is the change to move / consolidate the implementation of the attached properties.

I left one question, but I'll leave it up to you as to whether / how to address it. I'll do final testing next week.

modules/javafx.graphics/src/main/java/javafx/stage/Stage.java
mstr2: It makes sense to track the platform color scheme when we have no scene. Since this is not possible with the fluent API (there's no `orElse(ObservableValue)`), I've added a custom binding.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/Window.java
```diff
@@ -322,8 +324,29 @@ public final ReadOnlyDoubleProperty prefHeaderButtonHeightProperty() {
...
-        prefHeaderButtonHeight.set(height);
+    public final void setHeaderButtonHeight(double height) {
+        headerButtonHeight.set(height);
+    }
+
+    /**
+     * Specifies whether header buttons are displayed in a style appropriate for dark mode.
+     * Sub-classes can use this value in their header button visualization, but they are not required to do so.
+     */
+    private final ReadOnlyBooleanWrapper headerButtonDarkStyle =
```
Maran23: Can this be a lazy property? From what I saw when reviewing locally, this is only called when we are an extended window, otherwise it is not. If this correct, then I think it makes sense to save a bit of memory (especially since this class is also used for plain `Popup`s).

modules/javafx.graphics/src/main/java/com/sun/glass/ui/Window.java
```diff
@@ -309,11 +311,11 @@ protected Window(Window owner, Screen screen, int styleMask) {
      * Implementations should choose a sensible default height for their header button visualization if
      * {@link HeaderBar#USE_DEFAULT_SIZE} is specified.
      */
-    private final DoubleProperty prefHeaderButtonHeight =
-        new SimpleDoubleProperty(this, "prefHeaderButtonHeight", HeaderBar.USE_DEFAULT_SIZE);
+    private final ReadOnlyDoubleWrapper headerButtonHeight =
```
Maran23: same question as for `headerButtonDarkStyle`

modules/javafx.graphics/src/main/java/com/sun/glass/ui/HeaderButtonOverlay.java
```diff
@@ -318,31 +330,24 @@ public HeaderButtonOverlay(ObservableValue<String> stylesheet, boolean modalOrOw
...
-            .subscribe(_ -> updateStyleClass()); // use a value subscriber, not an invalidation subscriber
-
         subscriptions = Subscription.combine(
             focusedSubscription,
             resizableSubscription,
             maximizedSubscription,
-            updateStylesheetSubscription,
             stylesheet.subscribe(this::updateStylesheet));
 
         getStyleClass().setAll("-FX-INTERNAL-header-button-container");
```
Maran23, line 353: Should this be `-FX-INTERNAL` still? I thought this is named like that simply because it's a preview feature?

Maran23 (commented):
some preliminary review comments, will do more testing and reviewing later this day

Maran23:
Some things I have found which might be bugs. Some could very well be pre-existing bugs (unrelated to this feature).

1. The top border is cut by 1px. Maybe intended? But looks weird.
image

2. This is very weird because the issue can not be reproduced as soon as I record it. But I can make a screenshot.
I hover over the settings button and then right next to the minimize button and the hover state is not cleared from the settings button. 
image

3. The minimize button looks blurry
image

Information

- Windows 11
- Display scale 100%
- I just created a normal `Button`, called `headerBar.setRight(settingsBtn);` and after the `Stage` is shown I call `HeaderBar.setSystemButtonHeight(primaryStage, headerBar.prefHeight(Region.USE_PREF_SIZE));`.
- CSS for the border: `-fx-border-width: 2px;`

modules/javafx.graphics/src/main/java/com/sun/glass/ui/HeaderButtonOverlay.java
mstr2: No, it's named this way because it is an implementation detail and not API. Applications must not assume that the header buttons are styleable in any particular way, or at all.

mstr2:
> Some things I have found which might be bugs. Some could very well be pre-existing bugs (unrelated to this feature).
> 
> 1. The top border is cut by 1px. Maybe intended? But looks weird.

Doesn't look intentional to me. I've noticed issues like these with several layout containers, and I guess it's a snapping issue.

> 2. This is very weird because the issue can not be reproduced as soon as I record it. But I can make a screenshot.
>    I hover over the settings button and then right next to the minimize button and the hover state is not cleared from the settings button.

This looks like a bug. Can you create a JBS ticket to track it?

> 3. The minimize button looks blurry

Yes, there's a noticeable difference between the way JavaFX and Windows rasterize shapes, especially on low-DPI screens. I've tried to tweak shape sizes a little bit on Windows to make them look more similar, but there will always be a difference. One solution could be to use images instead of SVGs. This might be a future improvement.

modules/javafx.graphics/src/main/java/com/sun/glass/ui/HeaderButtonOverlay.java
Maran23: Got you, makes sense!

modules/javafx.graphics/src/main/java/javafx/scene/layout/HeaderBar.java
```diff
@@ -512,7 +537,6 @@ public HeaderBar(Node left, Node center, Node right) {
      * The left area of the {@code HeaderBar}.
      *
      * @defaultValue {@code null}
-     * @since 26
      */
     private final ObjectProperty<Node> left = new NodeProperty("left");
```
Maran23: I wonder if this properties (including `leftSystemPadding` and `rightSystemPadding` can be made lazy as well. The `xxxProperty` method are unused by default, so unless developers bind it or use the setter, the properties will not be initialized.

modules/javafx.graphics/src/main/java/javafx/scene/layout/HeaderDragType.java
Maran23: `TRANSPARENT` was the the hardest for me to understand. I think you did a good job already explaining it - But to fully understand, one more sentence what it means that the hit testing continues would help a lot to clarify IMO.

modules/javafx.controls/src/main/java/javafx/scene/control/DialogPane.java
```diff
@@ -435,23 +434,19 @@ public CssMetaData<DialogPane,String> getCssMetaData() {
...
+     * @since 27
      */
-    @Deprecated(since = "26")
+    private ObjectProperty<HeaderBar> headerBar;
+
     public final ObjectProperty<HeaderBar> headerBarProperty() {
         if (headerBar == null) {
-            PreviewFeature.HEADER_BAR.checkEnabled();
             headerBar = new SimpleObjectProperty<>(this, "headerBar") {
                 WeakReference<HeaderBar> wref = new WeakReference<>(null);
```
Maran23: `oldHeaderBar` might be a bit better and follows the other properties in other controls

modules/javafx.graphics/src/main/java/javafx/stage/StageStyle.java
```diff
@@ -114,27 +112,13 @@ public enum StageStyle {
...
-     * interface. Applications should set the scene fill to a color that matches the user interface of the header
-     * bar area, even if the scene fill is not visible because it is obscured by other controls.
-     *
-     * <h4>Custom header buttons</h4>
-     * If more control over the header buttons is desired, applications can opt out of the default header buttons
-     * by setting {@link HeaderBar#setPrefButtonHeight(Stage, double)} to zero and providing custom header buttons
-     * instead. Any JavaFX control can be used as a custom header button by setting its semantic type with the
-     * {@link HeaderBar#setButtonType(Node, HeaderButtonType)} method.
-     *
      * <h4>Title text</h4>
```
Maran23: Not just the title text but also the application icon in case of Windows and probably most of Linux. Probably good to mention as well? Since for both, clicking on the icon gives a `ContextMenu` which is also lost - so good to know for the developer.

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/HeaderBarTest.java
```diff
@@ -125,13 +151,128 @@ void buttonType_attachedProperty() {
...
         assertEquals(123, minHeight[0]);
-        assertEquals(123, HeaderBar.getPrefButtonHeight(stage));
+        assertEquals(123, HeaderBar.getSystemButtonHeight(stage));
+    }
+
+    @Nested
+    class SystemColorSchemeTest {
+        @Test
+        void attachedPropertyPublishesExplicitValue() {
+            var colorScheme = new ColorScheme[1];
```
Maran23, line 167: Minor: `AtomicReference` could be used. I personally like it a bit more

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/HeaderBarTest.java
```diff
@@ -71,14 +81,30 @@ void setup() {
...
+        setPlatformColorScheme(colorScheme);
+    }
+
+    void setPlatformColorScheme(ColorScheme colorScheme) {
+        PlatformImpl.getPlatformPreferences().update(Map.of(
+            TEST_FOREGROUND_COLOR, colorScheme == ColorScheme.DARK ? Color.WHITE : Color.BLACK,
+            TEST_BACKGROUND_COLOR, colorScheme == ColorScheme.DARK ? Color.BLACK : Color.WHITE));
     }
 
     <T> T getAttachedProperty(String name) {
```
Maran23, line 103: Should this be called `getExtendedProperties` as well?

Maran23 (commented):
Some more findings. My testing looks good. This was on my list for a long time but did not used it until now.

Javadoc is very well written. I used it mostly to figure out the API and it guided me very well.

Maran23:
> > 1. The top border is cut by 1px. Maybe intended? But looks weird.
> 
> Doesn't look intentional to me. I've noticed issues like these with several layout containers, and I guess it's a snapping issue.

Could be, but I directly set the `Button` as right `HeaderBar` node. So there is no container inbetween. Could still be a snapping problem in `Button` or so, but it could also be something else. Shall I create a ticket?

> > 2. I hover over the settings button and then right next to the minimize button and the hover state is not cleared from the settings button.
>
> This looks like a bug. Can you create a JBS ticket to track it?


Created: https://bugs.openjdk.org/browse/JDK-8387727

> > 3. The minimize button looks blurry
> 
> Yes, there's a noticeable difference between the way JavaFX and Windows rasterize shapes, especially on low-DPI screens. I've tried to tweak shape sizes a little bit on Windows to make them look more similar, but there will always be a difference. One solution could be to use images instead of SVGs. This might be a future improvement.

Or maybe we can improve the SVG rasterization. For rendering SVG paths, it seems to make a difference how the nodes are built and if the SVG is used as `Shape`. I just used the `minimize` SVG path in [SvgNode](https://github.com/Maran23/svgnode) and it looks good:
image
When I wrote it, I found out that the node structure and using a `SVGPath` as shapes influences how the SVG looks (e.g. if it looks blurry). So maybe we can improve that in the future!

modules/javafx.controls/src/main/java/javafx/scene/control/DialogPane.java
mstr2: I'd say it's not the old value, but the current value of the property. Changed it to `currentValue`.

modules/javafx.graphics/src/main/java/javafx/scene/layout/HeaderDragType.java
mstr2: I've added a sentence that hit-testing continues with the next eligible node under the transparent node.

modules/javafx.graphics/src/main/java/javafx/stage/StageStyle.java
mstr2: I've added language that an application icon is not provided for free. However, I didn't include any mention of a context menu: this is subsumed by "Applications [...] need to provide their own implementation", as this specification is not the right place to enumerate platform-specific behaviors.

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/HeaderBarTest.java
mstr2: I don't like that one: `AtomicReference` is an atomic reference, not a general-purpose mutable reference wrapper.

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/HeaderBarTest.java
mstr2: I'd say no, because this method specifically retrieves a single attached property from the extended stage.

mstr2:
> Could be, but I directly set the `Button` as right `HeaderBar` node. So there is no container inbetween. Could still be a snapping problem in `Button` or so, but it could also be something else. Shall I create a ticket?

Yeah, I think we should at least investigate and find out why that happens.

> Created: https://bugs.openjdk.org/browse/JDK-8387727

👍 

> Or maybe we can improve the SVG rasterization. For rendering SVG paths, it seems to make a difference how the nodes are built and if the SVG is used as `Shape`. I just used the `minimize` SVG path in [SvgNode](https://github.com/Maran23/svgnode) and it looks good: image When I wrote it, I found out that the node structure and using a `SVGPath` as shapes influences how the SVG looks (e.g. if it looks blurry). So maybe we can improve that in the future!

Good idea, I think that's also something we should investigate.

modules/javafx.graphics/src/main/java/javafx/stage/StageStyle.java
Maran23: >  I didn't include any mention of a context menu

Yes, I wouldn't add it either, since this is very platform-dependent (like many other things).

I'm happy with the icon addition.

Maran23 (approved):
I'm happy with the latest changes.

Seems like your GHA catched a timing where eclipse was down.

drmarmac:
I also did some testing of this PR, on Windows 11 and KDE Linux, with no obvious regressions on either platform. KDE Linux also shows the mentioned pre-existing issue with missing rounded corners, which should be fixable later without API changes. 

The height property renames are an improvement. The new `HeaderBar.systemColorScheme` toggles the system button brightness independently of everything else, which is what I expected from the javadocs.
I found the description given in this PR, "Setting this property to a non-null value is only necessary in the rare case when an application needs different color schemes for header buttons and the window (i.e. when it has a bright title bar in dark mode, or a dark title bar in light mode)", quite useful for understanding when to use the property. Maybe it's worth adding that to the javadocs as well, unless it's too verbose?

+1 for making attached properties first-class. I don't see any reason not to do PR #2015, which fills a longstanding gap in jfx APIs.

Maran23:
> > Could be, but I directly set the `Button` as right `HeaderBar` node. So there is no container inbetween. Could still be a snapping problem in `Button` or so, but it could also be something else. Shall I create a ticket?
> 
> Yeah, I think we should at least investigate and find out why that happens.

Created https://bugs.openjdk.org/browse/JDK-8387728

Regarding the SVG rendering, I will create one the coming days, currently experimenting / checking what the JavaFX Controls do + `modena.css` and collecting all the different approaches.

For example, this is how the `TitledPane` is configured. Looks a bit hacky (magic padding numbers that need to be configured exactly for the specific SVG shape). 

https://github.com/openjdk/jfx/blob/ca5d26623ce68d555d9a699c90b6dbd3e389f83e/modules/javafx.controls/src/main/resources/com/sun/javafx/scene/control/skin/modena/modena.css#L1863-L1874

modules/javafx.graphics/src/test/java/test/javafx/scene/layout/HeaderBarTest.java
andy-goryachev-oracle: I also think AtomicReference will be a better choice.  An array is probably ok here because it's a single-threaded test.

modules/javafx.graphics/src/main/java/javafx/scene/layout/HeaderBar.java
```diff
@@ -711,14 +738,14 @@ protected double computeMinWidth(double height) {
             : null;
 
         if (stage != null) {
-            var attachedProperties = AttachedProperties.of(stage);
+            var properties = StageHelper.getExtendedProperties(stage);
 
             if (scene.getEffectiveNodeOrientation() != getEffectiveNodeOrientation()) {
-                leftSystemPaddingWidth = isLeftSystemPadding() ? attachedProperties.rightSystemInset.get().getWidth() : 0;
-                rightSystemPaddingWidth = isRightSystemPadding() ? attachedProperties.leftSystemInset.get().getWidth() : 0;
+                leftSystemPaddingWidth = isLeftSystemPadding() ? properties.rightSystemInsetProperty().get().getWidth() : 0;
```
andy-goryachev-oracle, line 744: still think it was a mistake with left/right instead of leading/trailing.

andy-goryachev-oracle (approved):

modules/javafx.graphics/src/main/java/javafx/scene/layout/HeaderBar.java
mstr2: I agree, but this mistake was made over a decade ago. At this point, introducing inconsistency doesn't help anyone...

kevinrushforth: Yes, that ship has sailed.

andy-goryachev-oracle: Titanic!

kevinrushforth (approved):
Looks good to me, including the latest changes to the docs.

Maran23 (approved):
Changes look good.

PR #2191, opened 2026-06-17
8386663: Stylesheet/StyleClass list should be lazily initialized

Maran23:
This PR is an optimization for `Node.styleClass` and `Parent.stylesheets`.
Instead of always initializing both properties with an empty list, we are creating (therefore allocating) the list on the first access instead.
Similar to many other lazy properties.

Why?
- `Parent.getStylesheets()` is very rarely used by developers and JavaFX code. So this list is very often completely unused and empty. Most developers usually only add stylesheets at the `Scene`
- `Node.getStyleClass()` is usually not used (empty) for layout containers such as `Pane` or `Group` or shapes. A JavaFX App usually consists of a good amount of such containers (or shapes)

So that our CSS code is not initializing both lists on access, I added related `NodeHelper` and `ParentHelper` methods to return `null` when both lists were not initialized and therefore used. Otherwise we return the list as before.
This will save us some memory and allocation, which is both good for the memory consumption but also for `Node` / `Parent` creation (time).

Added documentation and tests. Will do some measurements with some apps very soon and attach it here.

---

# Benchmarks

I wrote a very small scene graph analyzer snippet to measure the memory gain.
Feel free to test this on your own apps!
1. Get the `SceneGraphAnalyzer` here: [SceneGraphAnalyzer](https://gist.github.com/Maran23/38beca5b043e547e1a84749e3162c0b2)
2. Add this code to your `Scene` and press the shortcut `F12` when all of the UI is loaded:
```
        scene.setOnKeyPressed(event -> {
            if (event.getCode() == KeyCode.F12) {
                var analyzer = new SceneGraphAnalyzer();
                var res = analyzer.analyze(scene);
                res.print();
            }
        });
```
3. If you have a modular app, add the following VM argument: `--add-opens javafx.graphics/javafx.scene=yourapp`

## Projects

1. Tested with [JFXCentral](https://github.com/dlsc-software-consulting-gmbh/jfxcentral2)

```
  +---------------------------------------------------------------+
  | Scene Graph Analysis                                          |
  +---------------------------------------------------------------+
  |  Nodes: 654 total, 388 parents, 266 leaves                    |
  +---------------------------------------------------------------+
  |   Field         Null   Total       %    N/A    Avg      Saved |
  |   ----------------------------------------------------------- |
  |   stylesheets    386 /   388   99.5%      -      -   39.20 KB |
  |   styleClass      19 /   654    2.9%      -      -    1.93 KB |
  +---------------------------------------------------------------+
  |  Total saved: 41.13 KB                                        |
  |  (at 104 bytes / empty list)                                  |
  +---------------------------------------------------------------+
```

1.1. Test with an Icon Library at JFXCentral

```
  +---------------------------------------------------------------+
  | Scene Graph Analysis                                          |
  +---------------------------------------------------------------+
  |  Nodes: 1,032 total, 640 parents, 392 leaves                  |
  +---------------------------------------------------------------+
  |   Field         Null   Total       %    N/A    Avg      Saved |
  |   ----------------------------------------------------------- |
  |   stylesheets    639 /   640   99.8%      -      -   64.90 KB |
  |   styleClass      17 / 1,032    1.6%      -      -    1.73 KB |
  +---------------------------------------------------------------+
  |  Total saved: 66.63 KB                                        |
  |  (at 104 bytes / empty list)                                  |
  +---------------------------------------------------------------+
```

2. Tested with [XPipe](https://github.com/xpipe-io/xpipe)

```
  +---------------------------------------------------------------+
  | Scene Graph Analysis                                          |
  +---------------------------------------------------------------+
  |  Nodes: 3,535 total, 2,733 parents, 802 leaves                |
  +---------------------------------------------------------------+
  |   Field         Null   Total       %    N/A    Avg      Saved |
  |   ----------------------------------------------------------- |
  |   stylesheets  2,732 / 2,733  100.0%      -      -  277.47 KB |
  |   styleClass     656 / 3,535   18.6%      -      -   66.63 KB |
  +---------------------------------------------------------------+
  |  Total saved: 344.09 KB                                       |
  |  (at 104 bytes / empty list)                                  |
  +---------------------------------------------------------------+
```

3. Tested with [JabRef](https://github.com/JabRef/jabref)

```
  +---------------------------------------------------------------+
  | Scene Graph Analysis                                          |
  +---------------------------------------------------------------+
  |  Nodes: 3,353 total, 2,379 parents, 974 leaves                |
  +---------------------------------------------------------------+
  |   Field         Null   Total       %    N/A    Avg      Saved |
  |   ----------------------------------------------------------- |
  |   stylesheets  2,372 / 2,379   99.7%      -      -  240.91 KB |
  |   styleClass     444 / 3,353   13.2%      -      -   45.09 KB |
  +---------------------------------------------------------------+
  |  Total saved: 286.00 KB                                       |
  |  (at 104 bytes / empty list)                                  |
  +---------------------------------------------------------------+
```

4. Table heavy app

```
  +---------------------------------------------------------------+
  | Scene Graph Analysis                                          |
  +---------------------------------------------------------------+
  |  Nodes: 747 total, 431 parents, 316 leaves                    |
  +---------------------------------------------------------------+
  |   Field         Null   Total       %    N/A    Avg      Saved |
  |   ----------------------------------------------------------- |
  |   stylesheets    431 /   431  100.0%      -      -   43.77 KB |
  |   styleClass      35 /   747    4.7%      -      -    3.55 KB |
  +---------------------------------------------------------------+
  |  Total saved: 47.33 KB                                        |
  |  (at 104 bytes / empty list)                                  |
  +---------------------------------------------------------------+
```

5. Chart heavy app

```
  +---------------------------------------------------------------+
  | Scene Graph Analysis                                          |
  +---------------------------------------------------------------+
  |  Nodes: 566 total, 345 parents, 221 leaves                    |
  +---------------------------------------------------------------+
  |   Field         Null   Total       %    N/A    Avg      Saved |
  |   ----------------------------------------------------------- |
  |   stylesheets    341 /   345   98.8%      -      -   34.63 KB |
  |   styleClass     146 /   566   25.8%      -      -   14.83 KB |
  +---------------------------------------------------------------+
  |  Total saved: 49.46 KB                                        |
  |  (at 104 bytes / empty list)                                  |
  +---------------------------------------------------------------+
```

---------

---------
### Progress

### Issue
 * [JDK-8386663](https://bugs.openjdk.org/browse/JDK-8386663): Stylesheet/StyleClass list should be lazily initialized (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2191/head:pull/2191` \
`$ git checkout pull/2191`

`$ git checkout pull/2191` \
`$ git pull https://git.openjdk.org/jfx.git pull/2191/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2191`

`$ git pr show -t 2191`

</details>

<a href="https://git.openjdk.org/jfx/pull/2191.diff">https://git.openjdk.org/jfx/pull/2191.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2191#issuecomment-4727932558)
</details>


Maran23:
Today I learned about https://github.com/openjdk/jol to measure the exact memory footprint of objects and used those insights to measure the memory gain of some apps - results in the PR description.

Memory stats:

```
com.sun.javafx.collections.ObservableListWrapper@76bbca23d footprint:
     COUNT       AVG       SUM   DESCRIPTION
         1        16        16   [Ljava.lang.Object;
         1        40        40   com.sun.javafx.collections.ObservableListWrapper
         1        24        24   java.util.ArrayList
         3                  80   (total)
```
With `-XX:+UseCompactObjectHeaders` (always enabled in JDK-27, so I used this value as baseline):
```
com.sun.javafx.collections.ObservableListWrapper@7ffb7d94d footprint:
     COUNT       AVG       SUM   DESCRIPTION
         1        16        16   [Ljava.lang.Object;
         1        32        32   com.sun.javafx.collections.ObservableListWrapper
         1        24        24   java.util.ArrayList
         3                  72   (total)
```

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -1326,42 +1331,48 @@ public String getName() {
...
+                        return get(0);
+                    } else {
+                        StringBuilder buf = new StringBuilder();
+                        for (int i = 0; i < size(); i++) {
+                            buf.append(get(i));
+                            if (i + 1 < size()) {
+                                buf.append(' ');
+                            }
+                        }
+                        return buf.toString();
```
nlisker: If you prefer:
```suggestion
			return switch (size()) {
				case 0 -> "";
				case 1 -> get(0);
				default -> {
					   StringBuilder buf = new StringBuilder();
		                for (int i = 0; i < size(); i++) {
		                    buf.append(get(i));
		                    if (i + 1 < size()) {
		                        buf.append(' ');
		                    }
		                }
		                yield buf.toString();
				}
			};
```

hjohn:
> * `Node.getStyleClass()` is usually not used (empty) for layout containers such as `Pane` or `Group` or shapes. A JavaFX App usually consists of a good amount of such containers (or shapes)

The stats you showed lead me to a different conclusion: style class is usually used, and most things do have styles associated with them, including containers; most containers I create definitely have a style, they just don't have one by default.

I fully agree on the StyleSheets property.

Maran23:
> The stats you showed lead me to a different conclusion: style class is usually used, and most things do have styles associated with them, including containers; most containers I create definitely have a style, they just don't have one by default.

Yes. It might still be a good idea though, but I agree that it seems to have the biggest advantage for Chart/`Shape` heavy apps from what I measured. 

It would be interesting to also measure an app that uses many `Shape3D` instances. As those have no `styleClass` as well by default. So I think those applications should also benefit from this. But I unfortunately don't have any myself, maybe I will check the `Jfx3dViewerApp`.

I still plan to test with some more application, but currently not at home so need to setup some of the projects on my laptop first.

nlisker:
> It would be interesting to also measure an app that uses many `Shape3D` instances. As those have no `styleClass` as well by default. So I think those applications should also benefit from this. But I unfortunately don't have any myself, maybe I will check the `Jfx3dViewerApp`.

There's https://github.com/openjdk/jfx/tree/master/tests/performance/3DLighting.

drmarmac:
I tested this in an app with lots of 2D shapes:

```
╔══════════════════════════════════════════════════╗
║ Scene Graph Analysis                             ║
╠══════════════════════════════════════════════════╣
║  Total nodes:               6,836                ║
║    ├─ Parent nodes:         4,850                ║
║    └─ Leaf nodes:           1,986                ║
╠══════════════════════════════════════════════════╣
║  Null styleClass:      2,048 / 6,836  ( 30.0%)   ║
║  Null stylesheets:     4,845 / 4,850  ( 99.9%)   ║
╠══════════════════════════════════════════════════╣
║  Saved (styleClass):      144.00 KB              ║
║  Saved (stylesheets):     340.66 KB              ║
║  Saved (total):           484.66 KB              ║
║                   (at 72 bytes / empty list)     ║
╚══════════════════════════════════════════════════╝
```

`Node`s with null styleClass are indeed mainly containers and `Shape`s.
I'd say it's worth doing both optimizations.

andy-goryachev-oracle:
I think this effort is similar to the idea of stashing rarely used attributes into a hidden hashmap.

@Maran23 would you like to cooperate (review, etc.) with me on this?  Once we get through with the rare properties, `Node`, `Shape`s, and possibly other classes, we might see substantial improvement in RAM consumption.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/SimpleSelector.java
```diff
@@ -340,7 +347,7 @@ private boolean matchesStyleClasses(List<String> styleClassNames) {
          * of this selector:
          */
 
-        if (styleClassNames.size() < selectorStyleClassNames.size()) {
+        if (styleClassNames == null || styleClassNames.size() < selectorStyleClassNames.size()) {
```
andy-goryachev-oracle, line 350: would it make more sense, instead of adding the null check here, move it up to after L278, to avoid unnecessary check?

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
andy-goryachev-oracle: or a regular switch(size()) to minimize the diff

andy-goryachev-oracle (approved):
I like it.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
Maran23: Good idea, although I wonder if we should just straight up remove the `toString()` from here and just put it into the `Node.toString()` method, directly into the `StringBuilder` that is created there anyway.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/SimpleSelector.java
Maran23: I think this will not really matter. 
Just made the change locally to check what looks/feels better and personally it felt better in `matchesStyleClasses`.

Because it bascially reads like "well, a null `styleClass` will not match anything, so just return". But just my personally feeling here really.

Maran23:
> I think this effort is similar to the idea of stashing rarely used attributes into a hidden hashmap.
> 
> @Maran23 would you like to cooperate (review, etc.) with me on this? Once we get through with the rare properties, `Node`, `Shape`s, and possibly other classes, we might see substantial improvement in RAM consumption.

Yes. I think we could probably reuse and adjust the `SceneGraphAnalyzer` approach for some of the measurements we should do in the future for this one.

Maran23:
Added two more Open Source projects to the benchmark section.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
Maran23: Moved the logic directly into `Node.toString()` because IMO it makes more sense and improves the readability as well. 
Because there is no `toString` method to visually parse anymore in `getStyleClass()`.
And in `Node.toString()`, we already have a `StringBuilder` so it got shorter and more readable as well (compared to if-else or switch-case).

And wrote tests for the default `Node.toString()` method as well to make sure nothing breaks.

Maran23:
Seems like the latest GHA catched a race condition:

```
TaskEventTest > onCancelledCalledWhenSetViaProperty() FAILED
        at app//test.javafx.concurrent.TaskEventTest.onCancelledCalledWhenSetViaProperty(TaskEventTest.java:375)
```

andy-goryachev-oracle:
> Seems like the latest GHA catched a race condition:

It's a known issue: https://bugs.openjdk.org/browse/JDK-8357459

modules/javafx.controls/src/test/java/test/javafx/scene/control/PopupControlTest.java
```diff
@@ -108,18 +108,18 @@ public void setup() {
...
         popup.getStyleClass().add("Hello");
-        assertEquals("Hello", popup.getStyleClass().toString());
+        assertEquals("[Hello]", popup.getStyleClass().toString());
     }
 
     @Test public void getStyleClassStringable3() {
         popup.getStyleClass().add("Hello");
         popup.getStyleClass().add("Goodbye");
-        assertEquals("Hello Goodbye", popup.getStyleClass().toString());
+        assertEquals("[Hello, Goodbye]", popup.getStyleClass().toString());
```
andy-goryachev-oracle, line 122: wouldn't this be a breaking change?
if someone is relying on style classes being separated with a space instead of json?

hjohn: It's a bit odd it was overridden in the first place to do something different from what most others `List`s do (although it does look a bit more like what you see in CSS that way, but still without the leading dots).

The format is not documented (you'd have to document it on `ObservableList` or on the `getStyleClass` method) so we would be allowed to make this change. Nobody should be relying on `toString` output, of essentially an `ObservableList`, being in a specific format.

andy-goryachev-oracle: so there is a potential for breaking a) `CssParser` and b) some weird application that relies on the existing behavior for some reason.

two questions then:
1. what is the rationale behind this change?
2. could we demonstrate that at least `CssParser` logic is not broken?

Maran23: I agree with @hjohn here hence the changes I made.
There is no change for the `CssParser`.
If applications rely on the unspecified and not documented `toString`, then yes, they may break. But that would be bad decision from the start - why would you do that instead of calling `contains`?

andy-goryachev-oracle: Well, I basically agree with you both, but I still want to get an answer

1. what is the rationale for this change?  was there something that the old format couldn't represent correctly?
2. I did not ask whether the `CssParser` was changed, but whether it relied in any way on the old format

andy-goryachev-oracle: one more thing: we should probably mention this change in the description and possibly in the JBS ticket as well.

Maran23: Ah maybe I misunderstood you, regarding your questions:
1. No, it is just a cleanup which originated from the comments in this PR
2. Ahh no, it does not

modules/javafx.graphics/src/test/java/test/javafx/scene/NodeTest.java
```diff
@@ -171,6 +189,73 @@ public class NodeTest {
...
+        assertEquals(1, styleClass.size());
+    }
+
+    /**
+     * StyleClass list is lazily initialized.
+     * CSS processing should not initialize it when null.
+     */
+    @Test
+    public void testCssProcessingDoesNotInitializeStyleClass() {
+        Rectangle node = new Rectangle();
```
andy-goryachev-oracle, line 220: the `node` is never attached to the scene graph, so the pulses do nothing, and also css processing is not happening, right?

modules/javafx.graphics/src/test/java/test/javafx/scene/ParentTest.java
```diff
@@ -75,6 +77,53 @@ public void tearDown() {
...
+     * CSS processing should not initialize it when null.
+     */
+    @Test
+    public void testCssProcessingDoesNotInitializeStylesheets() {
+        MockParent node = new MockParent();
+
+        List<String> stylesheets = ParentHelper.getStylesheetsOrNull(node);
+        assertNull(stylesheets);
+
+        Group g = new Group();
```
andy-goryachev-oracle: The `node` is never attached to the `Group` or the scene graph, same issue.

andy-goryachev-oracle (commented):
Please document the toString() format change in the JBS ticket and this PR description.

I don't know whether a CSR is needed, since the actual format has not been documented, but the fact that multiple tests fail with this change indicate that the applications that rely on the exact string (and not on `.contains()`) will break.

modules/javafx.graphics/src/test/java/test/javafx/scene/ParentTest.java
Maran23: Good catch, I missed that somehow. Should be inside the `Group`. The `stageLoader` can not be used here unfortunately so I somehow missed that when writing the stage setup manually.

Maran23:
> I don't know whether a CSR is needed, since the actual format has not been documented, but the fact that multiple tests fail with this change indicate that the applications that rely on the exact string (and not on `.contains()`) will break.

But note that the tests did check for whole `String`. If we would change e.g. `Node.toString()`, then tests will also fail. If we would change a `toString()` method from any other list (e.g. `getChildren`), then there might be a test that checked the excact pattern - but not because it is needed but because we have tests to even check the whole `String`.

A normal `contains` on the list will work as before, even a `contains` on the `String` will work as before. Just that the whole sequence looks like one would expect from a list. I can't see a reason why application would use the value from `toString` other than debugging

andy-goryachev-oracle (approved):
Mentioning this change in the `toString()` behavior in the JBS ticket should be sufficient, I think.

kevinrushforth:
> Mentioning this change in the toString() behavior in the JBS ticket should be sufficient, I think.

I agree. No need for a CSR (nor a release note).

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -713,6 +713,11 @@ public TransitionTimer findTransitionTimer(Node node, String propertyName) {
             public MediaQueryContext getMediaQueryContext(Node node) {
                 return node.getMediaQueryContext();
             }
+
+            @Override
+            public List<String> getStyleClassOrNull(Node node) {
+                return node.styleClass != null ? node.getStyleClass() : null;
```
mstr2:
```suggestion
                return node.styleClass;
```

modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
```diff
@@ -151,6 +151,11 @@ public ParentTraversalEngine getTraversalEngine(Parent parent) {
             public List<String> doGetAllParentStylesheets(Parent parent) {
                 return parent.doGetAllParentStylesheets();
             }
+
+            @Override
+            public List<String> getStylesheetsOrNull(Parent parent) {
+                return parent.stylesheets != null ? parent.getStylesheets() : null;
```
mstr2:
```suggestion
                return parent.stylesheets;
```

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
Maran23: looks a bit weird at first (since normally always a getter is used for lazy properties) but actually a good idea!

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/NodeHelper.java
```diff
@@ -134,6 +134,23 @@ public static void processCSS(Node node) {
...
+     * it is null as it will be lazily initialized only when accessed
+     * for the first time by the developer or JavaFX code.
+     * <p>
+     * Layout containers (such as Pane) usually never have any styleClass set (unless set by a developer),
+     * so we save an allocation by returning null when the styleClass is not initialized.
+     *
+     * @param node the node
+     * @return the styleClass or null
+     */
+    public static List<String> getStyleClassOrNull(Node node) {
```
mstr2, line 150: Most (all?) helper classes that implement the private accessor pattern don't have documentation, because it is implied that it's a simple forwarding method.

kevinrushforth:
It doesn't look like this will be ready before the jfx27 RDP1 fork, but in the unlikely event that it is, wait until after the fork. Any CSS lazy eval fix has a relatively high risk of regression, so this is the sort of change that is better earlier in the release cycle.

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/NodeHelper.java
Maran23: My thought was to explain why this method exists. Do you have a suggestion, or should we just remove it entirely?

Maran23:
> It doesn't look like this will be ready before the jfx27 RDP1 fork, but in the unlikely event that it is, wait until after the fork. Any CSS lazy eval fix has a relatively high risk of regression, so this is the sort of change that is better earlier in the release cycle.

That would be my suggestion, too. 
I also think that's the advantage of the release train model. Either it fits in on time, or it doesn't (which is not a bad thing). 
And usually there is no rush.

(On that, I wouldn't want to merge https://github.com/openjdk/jfx/pull/1970, https://github.com/openjdk/jfx/pull/2175 or https://github.com/openjdk/jfx/pull/2167 right before RDP1 either)

andy-goryachev-oracle (approved):

modules/javafx.graphics/src/main/java/com/sun/javafx/scene/NodeHelper.java
mstr2: I don't think that accessor methods should contain specification (that should probably live on the actual implementation), but you decide if you want to keep it.

Maran23: will remove.

Maran23:
I updated the `SceneGraphAnalyzer` a bit, also because I use it to measure the amount of CSS styleHelper for my CSS PR (and some upcoming changes).

Found out that I measured the footprint wrong, we actually save more bytes than I originally thought.
The way I know measure it now is by using: `GraphLayout.parseInstance(...).toFootprint();` and put in an instance of `TrackableObservableList`.

As it registers a listener inside the constructor (which will initialize the `listenerHelper`), the footprint is a bit higher.

modules/javafx.controls/src/main/java/javafx/scene/control/MenuItem.java
```diff
@@ -638,7 +638,12 @@ public final ObservableSet<PseudoClass> getPseudoClassStates() {
             if (!hasId) sbuf.append('[');
             else sbuf.append(", ");
             sbuf.append("styleClass=");
-            sbuf.append(getStyleClass());
+            for (int i = 0; i < getStyleClass().size(); i++) {
```
mstr2: You could replace all of that with `String.join(" ", getStyleClass())`.

modules/javafx.graphics/src/main/java/javafx/scene/Node.java
```diff
@@ -8676,7 +8669,12 @@ public String toString() {
             if (!hasId) sbuf.append('[');
             else sbuf.append(", ");
             sbuf.append("styleClass=");
-            sbuf.append(getStyleClass());
+            for (int i = 0; i < getStyleClass().size(); i++) {
```
mstr2: Same suggestion to use `String.join`.

modules/javafx.controls/src/main/java/javafx/scene/control/MenuItem.java
Maran23: good idea, done.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/StyleManager.java
```diff
@@ -1621,21 +1621,26 @@ public StyleMap findMatchingStyles(Node node, SubScene subScene, Set<PseudoClass
...
 
-                final String styleClass = styleClasses.get(n);
-                if (styleClass == null || styleClass.isEmpty()) continue;
+            if (styleClasses != null) {
+                key.styleClasses = FixedCapacitySet.of(styleClasses.size());
+                for (int n=0, nMax=styleClasses.size(); n<nMax; n++) {
 
-                key.styleClasses.add(styleClass);
+                    final String styleClass = styleClasses.get(n);
+                    if (styleClass == null || styleClass.isEmpty()) continue;
```
andy-goryachev-oracle: please use curly braces and place continue on its own line

modules/javafx.graphics/src/test/java/test/javafx/scene/NodeTest.java
```diff
@@ -99,6 +103,20 @@
  */
 public class NodeTest {
 
+    private StubToolkit toolkit;
+    private Stage stage;
+
+    @BeforeEach
+    public void setUp() {
+        toolkit = (StubToolkit) Toolkit.getToolkit();
+        stage = new Stage();
```
andy-goryachev-oracle: 1. many other tests do `((StubToolkit) Toolkit.getToolkit())` so it probably makes sense either do the same, or fix the other test to use this reference
2. this Stage is being created for each test, but only used it in one, is this right?

what do you think?

Maran23: changed. Regarding `StubToolkit`, normally the cast is not needed. But `StubToolkit` has some methods that might be needed (rarely) for tests.

Change the `Stage` logic. Note that the `StageLoader` would be perfect here, but only accessible in `javafx.controls`. I would really like to move it in a testbug PR, but the diff will probably be huge. What do you think? If this is okay, I will create a ticket.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/StyleManager.java
Maran23: note that this is the same as before. But changed as the diff is showing it anyway.

modules/javafx.graphics/src/test/java/test/javafx/scene/NodeTest.java
andy-goryachev-oracle: I think it's ok for now, only one test needs a `Stage`.

andy-goryachev-oracle (approved):
lgtm, thanks!

Maran23:
@mstr2 Do you want to re-review as well?

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/behavior/TwoLevelFocusListBehavior.java
```diff
@@ -192,7 +192,7 @@ private Event postDispatchTidyup(Event event) {
                 */
                 Parent p = tlNode.getParent();
                 if (p != null) {
-                    if (Properties.COMBO_BOX_STYLE_CLASS.equals(p.getStyleClass().toString())) {
+                    if (p.getStyleClass().size() == 1 && p.getStyleClass().contains(Properties.COMBO_BOX_STYLE_CLASS)) {
```
mstr2: This seems very strange. We detect a combo box here, but not if it has another (possibly unrelated) style class.

Maran23: Strange code indeed, but preexisting. But I just realized, we could use `NodeHelper` here as well.

The old code was even more weird by checking the `toString()` result, so it is a bit better now.

PR #2161, opened 2026-05-03
8374804: ConditionalFeature media queries

mstr2:
This PR adds the `-fx-supports-conditional-feature` media query, which allows applications to adapt their stylesheets to different platforms with varying conditional feature support. Currently, the built-in themes use hard-coded logic to include or exclude conditional-feature stylesheets. The new media query allows us to potentially remove the hard-coded logic in the future, and gives third-party themes an API to query conditional feature support.

---------
- [x] I confirm that I make this contribution in accordance with the [OpenJDK Interim AI Policy](https://openjdk.org/legal/ai).

---------
### Progress
- \[x\] Change must not contain extraneous whitespace
- \[x\] Change requires CSR request [JDK-8384139](https://bugs.openjdk.org/browse/JDK-8384139) to be approved
- \[x\] Commit message must refer to an issue
- \[x\] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issues
 * [JDK-8374804](https://bugs.openjdk.org/browse/JDK-8374804): ConditionalFeature media queries (**Enhancement** - P4)
 * [JDK-8384139](https://bugs.openjdk.org/browse/JDK-8384139): ConditionalFeature media queries (**CSR**)

### Reviewers
 * [Andy Goryachev](https://openjdk.org/census#angorya) (@andy-goryachev-oracle - **Reviewer**)
 * [Marius Hanl](https://openjdk.org/census#mhanl) (@Maran23 - Committer)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2161/head:pull/2161` \
`$ git checkout pull/2161`

Update a local copy of the PR: \
`$ git checkout pull/2161` \
`$ git pull https://git.openjdk.org/jfx.git pull/2161/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 2161`

View PR using the GUI difftool: \
`$ git pr show -t 2161`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jfx/pull/2161.diff">https://git.openjdk.org/jfx/pull/2161.diff</a>

</details>
<details><summary>Using Webrev</summary>

[...]

PR #2193, opened 2026-06-19
8374822: Platform media query

mstr2:
Add the `-fx-platform` media query (similar to `-moz-platform` in Firefox) that allows applications to switch styles based on the operating system on which the application is running.

/reviewers 2
/csr

---------

---------
### Progress
- \[x\] Change requires CSR request [JDK-8386998](https://bugs.openjdk.org/browse/JDK-8386998) to be approved

### Issues
 * [JDK-8374822](https://bugs.openjdk.org/browse/JDK-8374822): Platform media query (**Enhancement** - P4)
 * [JDK-8386998](https://bugs.openjdk.org/browse/JDK-8386998): Platform media query (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2193/head:pull/2193` \
`$ git checkout pull/2193`

`$ git checkout pull/2193` \
`$ git pull https://git.openjdk.org/jfx.git pull/2193/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2193`

`$ git pr show -t 2193`

</details>

<a href="https://git.openjdk.org/jfx/pull/2193.diff">https://git.openjdk.org/jfx/pull/2193.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2193#issuecomment-4754682199)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
+++ b/modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
@@ -1049,7 +1049,7 @@
             </tr>
             <tr>
                 <td class="value nowrap">-fx-platform</td>
-                <td class="value">android | embedded | headless | ios | linux | macos | unix | windows</td>
+                <td class="value">android | embedded | headless | ios | linux | macos | windows</td>
                 <td>discrete</td>
                 <td>This media feature cannot be evaluated in a <a href="#mf-boolean-context">boolean context</a>.</td>
             </tr>
--- a/modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
+++ b/modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
@@ -1047,6 +1047,12 @@
                     conditional feature.<br>
                     This media feature cannot be evaluated in a <a href="#mf-boolean-context">boolean context</a>.</td>
             </tr>
+            <tr>
+                <td class="value nowrap">-fx-platform</td>
+                <td class="value">android | embedded | headless | ios | linux | macos | unix | windows</td>
+                <td>discrete</td>
+                <td>This media feature cannot be evaluated in a <a href="#mf-boolean-context">boolean context</a>.</td>
+            </tr>
         </tbody>
     </table>
     <h3><a id="introexamples">Examples</a></h3>
--- a/modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
+++ b/modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
@@ -1049,7 +1049,7 @@
             </tr>
             <tr>
                 <td class="value nowrap">-fx-platform</td>
-                <td class="value">android | embedded | headless | ios | linux | macos | windows</td>
+                <td class="value">android | ios | linux | macos | windows</td>
                 <td>discrete</td>
                 <td>This media feature cannot be evaluated in a <a href="#mf-boolean-context">boolean context</a>.</td>
             </tr>
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaFeatures.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaFeatures.java
@@ -54,8 +54,6 @@ public final class MediaFeatures {
     // Supported values for the -fx-platform media query.
     static Map<String, BooleanSupplier> PLATFORMS = Map.of(
         "android", PlatformUtil::isAndroid,
-        "embedded", PlatformUtil::isEmbedded,
-        "headless", PlatformUtil::isHeadless,
         "ios", PlatformUtil::isIOS,
         "linux", PlatformUtil::isLinux,
         "macos", PlatformUtil::isMac,
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaFeatures.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaFeatures.java
@@ -59,7 +59,6 @@ public final class MediaFeatures {
         "ios", PlatformUtil::isIOS,
         "linux", PlatformUtil::isLinux,
         "macos", PlatformUtil::isMac,
-        "unix", PlatformUtil::isUnix,
         "windows", PlatformUtil::isWindows
     );
 
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaFeatures.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaFeatures.java
@@ -25,6 +25,7 @@
 
 package com.sun.javafx.css.media;
 
+import com.sun.javafx.PlatformUtil;
 import com.sun.javafx.application.PlatformImpl;
 import com.sun.javafx.css.media.expression.ConjunctionExpression;
 import com.sun.javafx.css.media.expression.FunctionExpression;
@@ -33,7 +34,9 @@ import com.sun.javafx.css.parser.Token;
 import javafx.application.ColorScheme;
 import javafx.application.ConditionalFeature;
 import java.util.Locale;
+import java.util.Map;
 import java.util.function.BiFunction;
+import java.util.function.BooleanSupplier;
 import java.util.function.Function;
 
 /**
@@ -44,9 +47,21 @@ public final class MediaFeatures {
     private MediaFeatures() {}
 
     // Extension point for unit tests to add media features
-    public static BiFunction<String, Token, MediaQuery> DEFAULT = (featureName, _) -> {
+    static BiFunction<String, Token, MediaQuery> DEFAULT = (featureName, _) -> {
         throw new IllegalArgumentException(String.format("Unknown media feature <%s>", featureName));
     };
+
+    // Supported values for the -fx-platform media query.
+    static Map<String, BooleanSupplier> PLATFORMS = Map.of(
+        "android", PlatformUtil::isAndroid,
+        "embedded", PlatformUtil::isEmbedded,
+        "headless", PlatformUtil::isHeadless,
+        "ios", PlatformUtil::isIOS,
+        "linux", PlatformUtil::isLinux,
+        "macos", PlatformUtil::isMac,
+        "unix", PlatformUtil::isUnix,
+        "windows", PlatformUtil::isWindows
+    );
 
     /**
      * Returns a {@code MediaQuery} that evaluates the specified feature in a discrete context.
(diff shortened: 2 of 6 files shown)


Maran23 (approved):
Looks good and the logic is very similar to the previous `ConditionalFeature` PR: https://github.com/openjdk/jfx/pull/2161

I also like the change that the `DEFAULT` variable is not public anymore (I commented on that back in the `Conditional stylesheet imports` PR: https://github.com/openjdk/jfx/pull/2031#discussion_r2937363946).

Conceptually, it is a bit strange to have `unix` as a standalone platform. It seems to include the `Linux` and `Solaris` platform when looking into `PlatformUtil`. But not `Mac`, which is also a Unix platform and even Unix certified AFAIK.

So that might be something we should have a look at before baking it into the new `-fx-platform` API. If `Solaris` is not important anymore, we could probably replace `isUnix` with `isLinux` (only used 2 times in prod code).

mstr2:
> Conceptually, it is a bit strange to have `unix` as a standalone platform. It seems to include the `Linux` and `Solaris` platform when looking into `PlatformUtil`. But not `Mac`, which is also a Unix platform and even Unix certified AFAIK.
> 
> So that might be something we should have a look at before baking it into the new `-fx-platform` API. If `Solaris` is not important anymore, we could probably replace `isUnix` with `isLinux` (only used 2 times in prod code).

You're right, let's remove "unix" for now because its semantics are unclear. It's not even used in CSS, and we can always add it later if we see the need to do so.

andy-goryachev-oracle:
I am not using these features, preferring to build stylesheets programmatically, so please forgive my silly question:

Should the platform media query be tied to a particular look and feel instead of the hardware platform?  While it makes sense in many areas to have things like `PlatformUtil.isMac()` as a general guidance (like popup trigger on the mouse press instead of mouse release, or the use of `command` key vs `ctrl` key), we might have more than one look and feel per platform.  Boomer example: Windows 98 vs Windows Vista vs Windows 10 vs Windows 11 might need somewhat different stylesheets, or various desktop environments on Linux.

mstr2:
> I am not using these features, preferring to build stylesheets programmatically, so please forgive my silly question:
> 
> Should the platform media query be tied to a particular look and feel instead of the hardware platform? While it makes sense in many areas to have things like `PlatformUtil.isMac()` as a general guidance (like popup trigger on the mouse press instead of mouse release, or the use of `command` key vs `ctrl` key), we might have more than one look and feel per platform. Boomer example: Windows 98 vs Windows Vista vs Windows 10 vs Windows 11 might need somewhat different stylesheets, or various desktop environments on Linux.

Good question. I think that themes shouldn't excessively change styles based on the platform, because that's a very broad category. On the web, you'll see a trend of moving to finer-grained capabilities (for example, `prefers-color-scheme` or `pointer: fine`) instead of using the platform as a decision proxy. The reason that I'm proposing to add `-fx-platform` is not so much that I think themes should be doing _more_ of these tests, especially not new themes. This media query is a relatively crude helper that allows us to define the old Caspian and Modena themes in CSS without hard-coded optional stylesheets.

In an earlier revision of this media query, I've had sub-categories like `windows-11`, `linux-gnome`, and so on. If we find out in the future that there's an absolutely compelling need to have these sub-categories, we can add them. For the moment, I think that just supporting the minimum feature set that Caspian and Modena actually use is the safer choice.

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
```diff
@@ -1047,6 +1047,12 @@ <h5>Combining Media Features</h5>
                     conditional feature.<br>
                     This media feature cannot be evaluated in a <a href="#mf-boolean-context">boolean context</a>.</td>
             </tr>
+            <tr>
+                <td class="value nowrap">-fx-platform</td>
+                <td class="value">android | embedded | headless | ios | linux | macos | windows</td>
```
andy-goryachev-oracle: what does `embedded` mean in this context?
is an embedded linux a `linux` or `embedded`?

maybe just remove it, like `unix`?

Maran23:
> Good question. I think that themes shouldn't excessively change styles based on the platform, because that's a very broad category. On the web, you'll see a trend of moving to finer-grained capabilities (for example, `prefers-color-scheme` or `pointer: fine`) instead of using the platform as a decision proxy. The reason that I'm proposing to add `-fx-platform` is not so much that I think themes should be doing _more_ of these tests, especially not new themes. This media query is a relatively crude helper that allows us to define the old Caspian and Modena themes in CSS without hard-coded optional stylesheets.

I agree with your point. However, I think it's worth noting that on the desktop side, there are legitimate use cases beyond refining Modena and Caspian for the platform media query.

The distinction here is that browsers already provide the native look and feel (header bar, window controls), and the website lives inside the browser. I'm not aware of a common use case where websites aim to look different on each OS (And unstyled HTML also usually does not match the native look and feel).
Desktop applications, on the other hand, are the top-level window - so matching the native look and feel is a much more common and valid goal.

And there are already community projects with such themes, one example is [JMetro](https://www.pixelduke.com/java-javafx-theme-jmetro/).

So I think the use case for platform queries is valid and will make it easy to change some CSS variables or things like the `HeaderBar` (+icons) rather than requiring programmatic stylesheet switching or other workarounds.

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
Maran23: The problem is that a stylesheet is loaded when `embedded` is `true`.

https://github.com/openjdk/jfx/blob/edf20b532fa9eeff14c06f9e041c93babf82bbe3/modules/javafx.graphics/src/main/java/com/sun/javafx/application/PlatformImpl.java#L789-L792

So it is needed to replace that. As far as I understand, `embedded` is like `headless`, a platform that can be used together with e.g. `linux`. You could just set the system property (like we do for our headless tests) and it will match the platform.

mstr2:
I've decided to remove both `embedded` and `headless`, so the platform query now only detects the actual OS on which JavaFX is running. I think this is simpler than overloading `-fx-platform` with JavaFX concepts like what counts as an embedded platform.

Of course, as @Maran23 pointed out, the current themes _do_ switch stylesheets based on whatever "embedded" means. However, the "embedded" stylesheet does nothing more than remove `-fx-effect` from all kinds of controls. This is obviously very crude, because there are certainly embedded devices that are not too slow to render effects.

I think that we should instead add the [update](https://www.w3.org/TR/mediaqueries-5/#update) media query for this purpose, which matches `slow` for severely under-powered devices. That would be better than using `-fx-platform: embedded` to mean "remove effects".

andy-goryachev-oracle (commented):
Should there be another, catch-all case where the platform does not fit the linux/macos/windows triad?  Solaris, for example?

Also, the CSR will need to be updated, and perhaps it should contain some explanation of the case when the platform does not map to the three defined choices.

Maran23 (approved):
Looks good.
CSR need to be updated.

Maran23:
> Should there be another, catch-all case where the platform does not fit the linux/macos/windows triad? Solaris, for example?

I think this should be requirement-driven. If there is a need for another platform like Solaris, someone will likely file an issue or post in the mailing list at one point.

andy-goryachev-oracle:
I agree with the sentiment, my question was "what happens".  It will not evaluate the `@media` block with any of the declared values, correct?  Should it be mentioned in the `cssref.html`, or is it self-evident?

Maran23:
> I agree with the sentiment, my question was "what happens". It will not evaluate the `@media` block with any of the declared values, correct? Should it be mentioned in the `cssref.html`, or is it self-evident?

Ahh got you, yes, it will match none in this case. Similar as for `-fx-supports-conditional-feature`, where they can be media queries that will not match (like `web`, if it is not on the classpath).

I think thats why Michael said it should not be excessively used. 
If no platform match, good default values should be used.

Btw., adding to what I wrote above regarding theming the application based on the OS and this comment, there is actually one usecase in JavaFX where things are different depending of the OS:

https://github.com/openjdk/jfx/blob/edf20b532fa9eeff14c06f9e041c93babf82bbe3/modules/javafx.controls/src/main/java/javafx/scene/control/ButtonBar.java#L475-L487

Which shows how I think `-fx-platform` should be used as well: Have a good default and may change the theming inside specialized platform media queries when needed.

@mstr2 unrelated to this PR, do you think the code in `ButtonBar` is a candidate that we should rather solve with the platform media queries instead?

andy-goryachev-oracle (approved):
All the explanations make sense, thanks!

mstr2:
I think we can define `linux` as "Unix, but not macOS and not Android" so that it includes systems like Solaris. From a UI perspective, there's no need to differentiate between different flavors of Unix, so long as all of those flavors use our GTK pipeline.

Calling the feature value `unix` would be confusing, because it doesn't match for unix-like systems where our rendering platform is not GTK. And from a developer perspective, I think "linux" is generally understood to also include systems with different branding.

The developers of [Firefox](https://github.com/mozilla-firefox/firefox/blob/c686530f70e9d4b8b4285d931a5630504939fca8/servo/components/style/gecko/media_features.rs#L562) also chose those semantics with the same rationale:

```rust
pub enum Platform {
    /// Matches any Android version.
    Android,
    /// For our purposes here, "linux" is just "gtk" (so unix-but-not-mac).
    /// There's no need for our front-end code to differentiate between those
    /// platforms and they already use the "linux" string elsewhere (e.g.,
    /// toolkit/themes/linux).
    Linux,
    /// Matches any iOS version.
    Ios,
    /// Matches any macOS version.
    Macos,
    /// Matches any Windows version.
    Windows,
}
```

mstr2:
> @mstr2 unrelated to this PR, do you think the code in `ButtonBar` is a candidate that we should rather solve with the platform media queries instead?

Yes, this looks like a good candidate for the new media query.

modules/javafx.graphics/src/main/java/com/sun/javafx/css/media/MediaFeatures.java
```diff
@@ -44,10 +47,19 @@ public final class MediaFeatures {
...
-    public static BiFunction<String, Token, MediaQuery> DEFAULT = (featureName, _) -> {
+    static BiFunction<String, Token, MediaQuery> DEFAULT = (featureName, _) -> {
         throw new IllegalArgumentException(String.format("Unknown media feature <%s>", featureName));
     };
 
+    // Supported values for the -fx-platform media query.
+    static Map<String, BooleanSupplier> PLATFORMS = Map.of(
+        "android", PlatformUtil::isAndroid,
+        "ios", PlatformUtil::isIOS,
+        "linux", PlatformUtil::isUnix, // for our purposes, "linux" means "unix but not macOS"
```
andy-goryachev-oracle, line 58: I wonder if the GTK aspect is more important here, and also in the `cssref.html`?
It's not about file path separators or some such, it's about GUI toolkit, isn't it?

mstr2: Yes, you're right that the purpose of this media query is not so much technicalities like file separators. I'd say that it is more about the "look and feel" of the OS, and how style themes can adapt to it. In this sense, it also isn't about whether JavaFX uses GTK to show the window, or whether it uses some other way to achieve the same thing.

So it's not about GTK, but then again "linux" is also not a precisely-defined look-and-feel. However, if you conducted a survey, I think you'd find that in actual usage, "linux" _does_ imply a certain look-and-feel, at least to an extent (because most people don't use exotic distros). And that's probably the only utility you'll get out of this particular media query.

andy-goryachev-oracle (approved):
Thank you for clarifications!  I think is should be clear enough for everyone.

Maran23 (approved):
Makes sense. Maybe we should rename `isUnix` to `isLinuxOrSolaris` or `isUnixButNotMac`.

Mostly for better readablity in the other code locations, but thats for another PR.

PR #2196, opened 2026-06-23
8366198: RichTextArea: embedded image, text background, wavy underline attributes

andy-goryachev-oracle:
# Support for Embedded Images, Text Highlights, Wavy Underlines

## Summary

This PR adds new attributes to support embedded images, text highlights, and wavy underlines.

The `RichEditorDemoApp` gains new context menus and enables drag and drop to showcase the new functionality.

Screenshot 2026-06-18 at 12 40 28

## Problem

Among the feedback received after publishing the RichTextArea incubating module, a number of users complained about the need to support embedded images.  Other users complained about difficulty to get the text highlighted or wavy underlined, and having these decorations controlled via CSS.

## Solution

Support for images was intentionally missing in the initial release; this PR adds the feature.

Adding highlights and underlines is a bit more complicated, for several reasons.  One reason is that the existing highlighting mechanism supports mixing of arbitrary number of translucent colors - something that is pretty much impossible to do with CSS alone.

A useful compromise is to offer a limited number of standard attributes that can be controlled via CSS: 5 highlights and 3 wavy underlines.

### StyleAttributeMap Attributes

New attributes:
- EMBEDDED_IMAGE
- TEXT_HIGHLIGHT_1 ... TEXT_HIGHLIGHT_5
- UNDERLINE_WAVY_1 ... UNDERLINE_WAVY_3

### CSS Styling

The styling of highlights and wavy underlines is determined by the stylesheet and can be changed by the user.

The style selectors are:

`.rich-text-area .text-highlight-1` ... `.rich-text-area .text-highlight-5`
`.rich-text-area .underline-wavy-1` ... `.rich-text-area .underline-wavy-3`

### New Public APIs

`EmbeddedImage`
`FileListFormatHandler`
`StyleAttribute.inlineNode()`
`StyleAttribute.isInlineNode()`
`StyledSegment.ofInlineNode()`
`CellContext.RunDecor`
`CellContext.decorateRun()`
`RichTextArea.dropTargetProperty`
`RichTextArea.clearDropTarget()`
`RichTextArea.getStyleAttributeMap()`

### File Format Changes

The data transfer format (also used by the demo app to save the .rich files) version string is changed, making the old files unreadable (no backward compatibility will be provided during incubation period).

### Breaking API Changes

`DataFormatHandler.createStyledInput()`
`StyledTextModel.getStyleAttributeMap()`
`StyledSegment.ofInlineNode()`

## Testing

In addition to the `RichEditorDemoApp`, the standalone Monkey Tester [1] provides additional options via context menu in the `RichTextArea` page.

## Questions to the Reviewers

1. is the number of text highlight (5) and wavy underline (3) attributes sufficient?
2. can the default colors, selected to be similar to colors used by charts, be improved?

## References

[0] [RichTextArea DataFormat (Incubator)](https://github.com/andy-goryachev-oracle/jfx/blob/8366198.new.attributes/doc-files/controls/RichTextArea/RichTextArea_DataFormat_Incubator.md)

[1] [Monkey Tester with added support for the new attributes](https://github.com/andy-goryachev-oracle/MonkeyTest/tree/8366198.attributes)

[2] [JDK-8386850 [StyledTextModel] rethink marshalling mechanism](https://bugs.openjdk.org/browse/JDK-8386850)

---------

---------
### Progress
- \[x\] Change requires CSR request [JDK-8388059](https://bugs.openjdk.org/browse/JDK-8388059) to be approved

### Issues
 * [JDK-8366198](https://bugs.openjdk.org/browse/JDK-8366198): RichTextArea: embedded image, text background, wavy underline attributes (**Enhancement** - P3)
 * [JDK-8388059](https://bugs.openjdk.org/browse/JDK-8388059): RichTextArea: embedded image, text background, wavy underline attributes (**CSR**)

### Reviewers
 * [Ajit Ghaisas](https://openjdk.org/census#aghaisas) (@aghaisas - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2196/head:pull/2196` \
`$ git checkout pull/2196`

`$ git checkout pull/2196` \
`$ git pull https://git.openjdk.org/jfx.git pull/2196/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2196`

`$ git pr show -t 2196`

</details>

<a href="https://git.openjdk.org/jfx/pull/2196.diff">https://git.openjdk.org/jfx/pull/2196.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2196#issuecomment-4782855001)
</details>


modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledTextModel.java
```diff
@@ -772,12 +791,13 @@ private final TextPos replace(StyleResolver resolver, TextPos start, TextPos end
                 insertParagraph(index, gen);
                 break;
             case TEXT:
+            case INLINE_NODE:
```
Jugen, line 792: I realize that this has only been included in order to handle embedded images, and that maybe you're going to  make it more specific in another PR - if so then that's great. Otherwise, I suggest adding the missing protected method `insertNodeSegment` and invoking it here for INLINE_NODE. The _StyledTextModel_ implementation of the method can just forward it to `insertTextSegment` as being done here. These changes will then also allow custom models to override `insertNodeSegment` and make copy paste as well as import operations possible with inline nodes.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttributeMap.java
```diff
@@ -94,9 +100,57 @@ public final class StyleAttributeMap {
...
+     * Text highlight character attribute for background color 4.
+     * @since 27
+     */
+    public static final StyleAttribute<Boolean> TEXT_HIGHLIGHT_4 = StyleAttribute.character("TEXT_HIGHLIGHT_4", Boolean.class);
+
+    /**
+     * Text highlight character attribute for background color 5.
+     * @since 27
+     */
+    public static final StyleAttribute<Boolean> TEXT_HIGHLIGHT_5 = StyleAttribute.character("TEXT_HIGHLIGHT_5", Boolean.class);
```
Jugen: The JavaDoc comments here should mention that the color is specified via CSS and what the default color is.
Something like: "Text highlight attribute. The color is specified via CSS text-highlight-1 which has a default color of xxxx."

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttributeMap.java
```diff
@@ -94,9 +100,57 @@ public final class StyleAttributeMap {
...
+     * Wavy underline character attribute for color 2.
+     * @since 27
+     */
+    public static final StyleAttribute<Boolean> UNDERLINE_WAVY_2 = StyleAttribute.character("UNDERLINE_WAVY_2", Boolean.class);
+
+    /**
+     * Wavy underline character attribute for color 3.
+     * @since 27
+     */
+    public static final StyleAttribute<Boolean> UNDERLINE_WAVY_3 = StyleAttribute.character("UNDERLINE_WAVY_3", Boolean.class);
```
Jugen: The JavaDoc comments here should mention that the color is specified via CSS and what the default color is.
Something like: "Wavy underline attribute. The color is specified via CSS underline-wavy-1 which has a default color of red."

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/RichTextArea.java
```diff
@@ -2450,12 +2523,60 @@ private static StyleHandlerRegistry initStyleHandlerRegistry() {
             cx.addStyle("-fx-fill:" + color + ";");
         });
 
+        b.setSegHandler(StyleAttributeMap.TEXT_HIGHLIGHT_1, (cc, cx, v) -> {
+            if (v) {
+                cx.decorateRun(StyleAttributeMap.TEXT_HIGHLIGHT_1, CellContext.RunDecor.HIGHLIGHT, Params.TEXT_HIGHLIGHT_1);
+            }
```
Jugen, line 2529: Since a developer could define their own highlight or wavy underline attribute like:
> INFO_HIGHLIGHT = new StyleAttribute<>( "INFO_HIGHLIGHT", Color.class, false );

and want to add it to the _StyleHandlerRegistry_, it would be helpful if `decorateRun` was overloaded with a Color parameter:
> cx.decorateRun( INFO_HIGHLIGHT, CellContext.RunDecor.HIGHLIGHT, v );

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttributeMap.java
andy-goryachev-oracle: this is a good idea, thanks!

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/RichTextArea.java
andy-goryachev-oracle: +1

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttributeMap.java
andy-goryachev-oracle: a quick note: the actual color is not a part of API, but the style name definitely is.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledTextModel.java
andy-goryachev-oracle: That's right, we have a separate ticket
https://bugs.openjdk.org/browse/JDK-8365948

andy-goryachev-oracle:
One question I wanted to ask @Jugen and the others is whether the number of text highlight (5) and wavy underline (3) attributes is sufficient.  To a lesser extent, whether the default colors are ok or can be improved.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/RichTextArea.java
andy-goryachev-oracle: On second thought, this will needlessly complicate things.

The only reason I had `RichParagraph.Builder.addHighlight(..., Color)` is to support a varying number of ad-hoc highlights generated programmatically.  To give you an example, a log viewer with 10-20 different tokens highlighted in different colors.

In this PR, we are adding a limited set of attributes to address 95% of most common use cases.

Since you simply want to use what looks like a single `INFO_HIGHLIGHT` attribute, you could either use the style name and your own stylesheet, or repurpose one of the standard `TEXT_HIGHLIGHT_*` with a custom stylesheet.

Does it make sense?

koppor:
Does this PR also support to change the mouse cursor on links? Currently, the mouse cursor gets the text selection cursor when hovering on links. (Refs https://github.com/JabRef/html-to-node/issues/5)

(Moreover, Claude found something more: https://github.com/JabRef/html-to-node/issues/3 - I did not "unzip" these items yet).

andy-goryachev-oracle:
@koppor : 
the ability to set the cursor is already there: you can declare your own character `StyleAttribute`, and then use `CellContext.getNode()` to get the pointer to the `Text` node, as seen here:

https://github.com/andy-goryachev-oracle/jfx/blob/0142cc6f1077cca9ad51e285cabe57f253a2b394/apps/samples/RichTextAreaDemo/src/com/oracle/demo/richtext/rta/RichTextAreaDemoPane.java#L117

(edit: link)

doc-files/controls/RichTextArea/RichTextArea_DataFormat_Incubator.md
```diff
@@ -38,7 +38,7 @@ As an example, the following rich text
 is represented by the following file:
 
 ```
-{@RichText-v2-incubator}{#tabs|99.5}{ff|System}{fs|12.0}tabs:	1	2	3{!tabs|149.0,190.0,229.0}
+{@RichText-v4-incubator}{#tabs|99.5}{ff|System}{fs|12.0}tabs:	1	2	3{!tabs|149.0,190.0,229.0}
```
kevinrushforth: Shouldn't that be "v3" -- not sure why it needs to jump from v2 to v4...

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/RichTextArea.java
Jugen: Yes for a single attribute, but if there are multiple attributes or you don't want to use a stylesheet then it's needful because otherwise the developer has to override `prepareParagraph` and do the overlapping mechanics themselves. This PR with the overloaded `decorateRun` I suggested would make it possible to rely on the built-in implementation instead, which would be very helpful.

andy-goryachev-oracle: Ok, I see your point.  

My initial thinking was that the highlights require additional support from the model, which means a custom model with its own `prepareParagraph()`.

I've filed a separate ticket https://bugs.openjdk.org/browse/JDK-8387983 .

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/DataFormatHandler.java
```diff
@@ -38,17 +38,17 @@
  */
 public abstract class DataFormatHandler {
     /**
-     * Creates a StyledInput for the given input string.  When pasting, the caller may pass
+     * Creates a StyledInput for the given input data.  When pasting, the caller may pass
      * the style attributes {@code attr} at the insertion point.  This argument may be used by
      * the implementation if the format contains no styles on its own (for example, in the plain text format case).
      *
-     * @param input the input string
+     * @param input the input data, never null
```
kevinrushforth, line 45: Now that you take an Object instead of a String, you need to define what subtypes are expected and what happens if an unexpected type is passed in. Presumably "IllegalArgumentException" or "ClassCastException", but maybe it is a no-op?

Some documentation should be on this abstract method. If the answer depends on the subclass, then document this with an `@implSpec` section.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/DataFormatHandler.java
```diff
@@ -38,17 +38,17 @@
...
      * @param attr the style attributes (can be null)
      * @return the StyledInput
      * @throws IOException when operation is not supported or an I/O error occurs
      * @throws UnsupportedOperationException if the copy operation is not supported
      */
-    public abstract StyledInput createStyledInput(String input, StyleAttributeMap attr) throws IOException;
+    public abstract StyledInput createStyledInput(Object input, StyleAttributeMap attr) throws IOException;
```
kevinrushforth, line 53: This needs `@since 27`, since it is a binary incompatible change that is effectively removing the old and adding a new method.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledSegment.java
```diff
@@ -288,9 +288,10 @@ public String toString() {
     /**
      * Creates a StyledSegment which consists of a single inline Node.
      * @param generator the code to create a Node instance
+     * @param a the segment styles, can be null
      * @return the StyledSegment instance
      */
-    public static StyledSegment ofInlineNode(Supplier<Node> generator) {
+    public static StyledSegment ofInlineNode(Supplier<Node> generator, StyleAttributeMap a) {
```
kevinrushforth: Needs `@since 27`

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledSegment.java
```diff
@@ -288,9 +288,10 @@ public String toString() {
     /**
      * Creates a StyledSegment which consists of a single inline Node.
      * @param generator the code to create a Node instance
+     * @param a the segment styles, can be null
```
kevinrushforth: What does "null" do? Is it different from an empty style attribute map?

doc-files/controls/RichTextArea/RichTextArea_DataFormat_Incubator.md
```diff
@@ -107,19 +110,57 @@ Example:
...
+|hi3     |TEXT_HIGHLIGHT_3      |boolean     |Text highlight color 3
+|hi4     |TEXT_HIGHLIGHT_4      |boolean     |Text highlight color 4
+|hi5     |TEXT_HIGHLIGHT_5      |boolean     |Text highlight color 5
+|i       |ITALIC                |boolean     |Italic typeface
+|ss      |STRIKE_THROUGH        |boolean     |Strike through
+|tc      |TEXT_COLOR            |Color       |Text color (Note 3)
+|u       |UNDERLINE             |boolean     |Underline
+|uw1     |UNDERLINE_WAVY_1      |boolean     |Wavy underline color 1
+|uw2     |UNDERLINE_WAVY_2      |boolean     |Wavy underline color 2
+|uw3     |UNDERLINE_WAVY_3      |boolean     |Wavy underline color 3
```
kevinrushforth: I think `WAVY_UNDERLINE_N` (here and in other places) would be a more consistent naming.

doc-files/controls/RichTextArea/RichTextArea_DataFormat_Incubator.md
```diff
@@ -107,19 +110,57 @@ Example:
...
+
+1. Value of `0.0` indicates that the rendered image height or width should be computed according to the image
+   intrinsic aspect ratio.
+2. Value of `-1.0` indicates that the rendered image width should not exceed the view's wrapped text width.
+   Value of `-2.0` indicates that the rendered image width should always fit the view's wrapped text width.
+
+
+Example:
+
+`{img|w,138.0,h,102.0,tw,-1.0,th,-1.0,a,true,b,iVBORw0KGgoAAA...}`
```
kevinrushforth: The example has `th,-1.0` but the above note says that -1 is only for the width.

modules/javafx.graphics/src/main/docs/javafx/scene/doc-files/cssref.html
```diff
@@ -6623,6 +6623,39 @@ <h3 class="package">jfx.incubator.scene.control.richtext</h3>
...
+            text-highlight-4<br>
+            text-highlight-5
+          </th>
+          <td>The text highlight styles.</td>
+        </tr>
+        <tr>
+          <th scope="row" class="propertyname">
+            underline-wavy-1<br>
+            underline-wavy-1<br>
+            underline-wavy-3
```
kevinrushforth: Copy / paste error: the second `-1` items should be `-2`. And here is another place where `wavy-underline-N` would be a more consistent name.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/FileListFormatHandler.java
```diff
@@ -0,0 +1,142 @@
...
+     * Returns the singleton instance of {@code FileListFormatHandler}.
+     * @return the singleton instance of {@code FileListFormatHandler}
+     */
+    public static final FileListFormatHandler getInstance() {
+        return instance;
+    }
+
+    @Override
+    public StyledInput createStyledInput(Object input, StyleAttributeMap attr) throws IOException {
+        List<File> files = (List<File>)input;
```
kevinrushforth, line 72: This is an example of what I mentioned in the superclass method. This class should provide docs and at least add an `@throws ClassCastException` (if that's what we decide is the cleanest behavior, else document the behavior you want and change the impl to check for instanceof and implement that behavior).

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/FileListFormatHandler.java
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     * Inserts the dropped files as inline images into the {@link RichTextArea}.
+     * If a file cannot be loaded as an image, the file name is inserted instead.
+     * This method clears the existing selection.
+     *
+     * @param t the target control
+     * @param p the text position
+     * @param files the list of files to be inserted
+     */
+    public static void handleDrop(RichTextArea t, TextPos p, List<File> files) {
```
kevinrushforth: Who calls this static method?

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/EmbeddedImage.java
```diff
@@ -0,0 +1,311 @@
...
+
+    private final byte[] bytes;
+    private final double width;
+    private final double height;
+    private final double targetWidth;
+    private final double targetHeight;
+    private final boolean keepAspectRatio;
+
+    /// Private constructor that DOES NOT make a defensive copy of the bytes.
+    private EmbeddedImage(
```
kevinrushforth, line 92: Minor: We don't usually put javadoc comments on private methods of public classes. Suggestion: consider changing this to an ordinary inline comment (although I see that this pattern is used elsewhere, so this could be deferred).

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/EmbeddedImage.java
```diff
@@ -0,0 +1,311 @@
...
+
+    /**
+     * Creates a copy of this {@code EmbeddedImage} with the specified target width, height, and aspect ratio.
+     * @param targetWidth the new target width
+     * @param targetHeight the new target height
+     * @param keepAspectRatio whether to keep the aspect ratio
+     * @return the new instance
+     */
+    public EmbeddedImage copy(double targetWidth, double targetHeight, boolean keepAspectRatio) {
+        return new EmbeddedImage(bytes, width, height, targetWidth, targetHeight, keepAspectRatio);
```
kevinrushforth, line 225: Not an API issue, but this doesn't make a defensive copy. Is the `bytes` array ever modified? Or is it effectively deeply immutable?

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/RichTextArea.java
```diff
@@ -520,6 +522,54 @@ public final boolean isDisplayCaret() {
         return displayCaret == null ? Params.DEFAULT_DISPLAY_CARET : displayCaret.get();
     }
 
+    /**
+     * This property indicates the target text position during the drag-and-drop operations.
```
kevinrushforth: s/the drag-and-drop/drag-and-drop/

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/RichTextArea.java
```diff
@@ -520,6 +522,54 @@ public final boolean isDisplayCaret() {
...
+     * Attempts to set the drop target position at the specified screen coordinates.
+     * If the drop target is visible, the value of the
+     * {@link #dropTargetProperty() dropTarget} property will be updated to reflect the insertion point in the model,
+     * otherwise it will be set to {@code null}.
+     *
+     * @param screenX the screen x coordinate
+     * @param screenY the screen y coordinate
+     * @since 27
+     */
+    public final void setDropTarget(double screenX, double screenY) {
```
kevinrushforth, line 550: Should there be a variant of this method that takes a `TextPos`?

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/RichTextArea.java
```diff
@@ -1417,8 +1467,35 @@ public final void extendSelection(TextPos pos) {
...
+     * segment's attributes will be returned so inserted/typed text has the same style as
+     * preceding text.  When {@code forInsert} is {@code false}, the returned attributes correspond exactly
+     * to the symbol indicated by the specified position.
+     * <p>
+     * This method returns the attributes of the last character at the end of the document.
+     * An empty {@code StyleAttributeMap} is returned for a {@code null} model.
+     *
+     * @param pos the text position
+     * @param forInsert whether to pick preceding style at the segment boundary
+     * @return the style attributes, non-null
```
kevinrushforth: Minor: we usually say "never null"

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledTextModel.java
```diff
@@ -244,13 +244,22 @@ public interface Listener {
...
+     * This method returns the attributes of the last character at the end of the document.
      *
      * @param resolver the style resolver
      * @param pos the text position
+     * @param forInsert whether to pick preceding style at the segment boundary
      * @return the style attributes, non-null
+     * @since 27
      */
-    public abstract StyleAttributeMap getStyleAttributeMap(StyleResolver resolver, TextPos pos);
+    public abstract StyleAttributeMap getStyleAttributeMap(StyleResolver resolver, TextPos pos, boolean forInsert);
```
kevinrushforth, line 260: Rather than removing the existing method, would it be better to leave it and have it call the new method with a default value of forInsert (probably false, unless you think true is more common)?

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledTextModel.java
```diff
@@ -244,13 +244,22 @@ public interface Listener {
...
      * Returns the {@link StyleAttributeMap} of the character at the specified position's {@code charIndex}.
-     * When at the end of the document, returns the attributes of the last character.
+     * <p>
+     * When the position points to the boundary between two text segments with different attributes,
+     * the value of {@code forInsert} determines which segment will be used: when {@code true}, the preceding
+     * segment's attributes will be returned so inserted/typed text has the same style as
+     * preceding text.  When {@code forInsert} is {@code false}, the returned attributes correspond exactly
+     * to the symbol indicated by the specified position.
+     * <p>
+     * This method returns the attributes of the last character at the end of the document.
```
kevinrushforth: This is a little confusing. My first reading of this sentence was that it always returned the attributes of the last character (that is, the character at the end of the document), which I'm sure isn't what you are trying to say.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledTextModel.java
```diff
@@ -684,7 +693,7 @@ public final TextPos getEndOfParagraphTextPos(int index) {
      * <p>
      * This is a convenience method which eventually calls
      * {@link #replace(StyleResolver, TextPos, TextPos, StyledInput)}
-     * with the attributes provided by {@link #getStyleAttributeMap(StyleResolver, TextPos)} at the
+     * with the attributes provided by {@link #getStyleAttributeMap(StyleResolver, TextPos, boolean)} at the
```
kevinrushforth: The doc should explicitly say that it calls `getStyleAttributeMap` with `true` for the last param.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttribute.java
```diff
@@ -75,6 +75,18 @@ public static <P> StyleAttribute<P> document(String name, Class<P> type) {
...
+    /**
+     * Creates an inline node character attribute.
+     * @param <P> the attribute value type
+     * @param name the attribute name
+     * @param type the attribute value type
+     * @return the new attribute instance
+     * @since 27
+     */
+    public static <P> StyleAttribute<P> inlineNode(String name, Class<P> type) {
+        return new InlineNodeStyleAttribute<P>(name, type);
```
kevinrushforth, line 87: Rather than a separate `inlineNode` factory method, could this instead be done with a check on the class type in `character()` factory method? I could ask the same of the other factory methods, so maybe there is some value in doing this way?

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextFormatHandler.java
```diff
@@ -185,8 +195,9 @@ public StyledOutput createStyledOutput(RichTextFormatHandler h, StyleResolver r,
     }
 
     @Override
-    public StyledInput createStyledInput(String input, StyleAttributeMap attr) {
-        return new RichStyledInput(input);
+    public StyledInput createStyledInput(Object input, StyleAttributeMap attr) {
```
kevinrushforth, line 198: Check to see whether this inherits the `@since 27`

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextModel.java
```diff
@@ -233,13 +235,14 @@ protected void applyStyle(int index, int start, int end, StyleAttributeMap attrs
     }
 
     @Override
-    public StyleAttributeMap getStyleAttributeMap(StyleResolver resolver, TextPos pos) {
+    public StyleAttributeMap getStyleAttributeMap(StyleResolver resolver, TextPos pos, boolean forInsert) {
```
kevinrushforth, line 238: Check whether `@since 27` inherits

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextModel.java
```diff
@@ -414,19 +417,33 @@ public int getTextLength() {
...
-         * @return the style info
+         * Retrieves the style attributes at the specified position.
+         * @param charIndex the character index
+         * @param leading the leading/trailing bias
+         * @param forInsert whether to pick preceding style at the segment boundary
+         * @return the style attributes, non-null
+         * @since 27
          */
-        public StyleAttributeMap getStyleAttributeMap(int offset) {
+        public StyleAttributeMap getStyleAttributeMap(int charIndex, boolean leading, boolean forInsert) {
```
kevinrushforth, line 426: This is in a private class. It probably shouldn't have any javadoc, much less an `@since`

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/SimpleViewOnlyStyledModel.java
```diff
@@ -327,13 +327,14 @@ public SimpleViewOnlyStyledModel nl(int count) {
     }
 
     @Override
-    public StyleAttributeMap getStyleAttributeMap(StyleResolver r, TextPos pos) {
+    public StyleAttributeMap getStyleAttributeMap(StyleResolver r, TextPos pos, boolean forInsert) {
```
kevinrushforth, line 330: Check `@since 27`

kevinrushforth (commented):
I've reviewed the public API and it generally looks good. I left mostly javadoc comments and questions with a couple API questions.

I haven't reviewed cssref yet, but that's next.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/DataFormatHandler.java
andy-goryachev-oracle: documented `ClassCastException`

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/EmbeddedImage.java
andy-goryachev-oracle: Eclipse shows javadoc when I click on a name, so it is useful.
Javadoc tool ignores these comments.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/EmbeddedImage.java
andy-goryachev-oracle: The only public way of creating an `EmbeddedImage` instance is `EmbeddedImage.of()` which does make a defensive copy.

Also, the byte array is never exposed to the public.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/FileListFormatHandler.java
andy-goryachev-oracle: RichEditorDemoWindow:118

this is a convenience method for DnD.  I think this class is the right place to expose this functionality, but I am open to suggestions.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextFormatHandler.java
andy-goryachev-oracle: does not:

Image

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttribute.java
andy-goryachev-oracle: not sure I understand the problem here.  it's a way to create an inline node attribute.  I don't want to expose a single-purpose constructor or various types, thus the factory method.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledSegment.java
andy-goryachev-oracle: yes: `getStyleAttributeMap()` is allowed to return `null`.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledTextModel.java
andy-goryachev-oracle: no, I'd rather change the method signature during incubating period.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/RichTextArea.java
andy-goryachev-oracle: Not really, the caller of this method is a mouse event handler, and it's better to use screen coordinates, similar to `getTextPosition()`

andy-goryachev-oracle:

@aghaisas would you mind lending a hand in reviewing this PR, if you have some time?

andy-goryachev-oracle:
@kevinrushforth thank you for reviewing!
The CSR is ready.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/FileListFormatHandler.java
kevinrushforth: The superclass now says that a `ClassCastException` is thrown if the the input data type is not supported. Do you think it is worth adding a sentence to this method saying what type is supported? Maybe something like this?

```suggestion
    /**
     * {@inheritDoc}
     *
     * <p>The type of {@code input} must be {@code List<File>}.
     */
    @Override
    public StyledInput createStyledInput(Object input, StyleAttributeMap attr) throws IOException {
        List<File> files = (List<File>)input;
```

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/FileListFormatHandler.java
kevinrushforth: I see what this is doing now. This is not the right place to put this convenience method. It is in a class that otherwise isn't accessed by applications. Further, it is a static method that does not operate on a DataFormatHandler at all; it operates on a RichTextArea not on the model, so I don't see why it would even be anywhere in the model package.

The method also is not named well. The fact that an application can use it to handle a drop operation doesn't describe what it does.

How about moving this static helper method to be an instance method in RichTextArea and renaming it something like `insertImages` or `insertImageFiles` or ... ?

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextFormatHandler.java
kevinrushforth: OK. Looking closer this is good, since the overridden method points to the parent method, and it is that parent method that was added in 27. No changes needed.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextModel.java
kevinrushforth: Not an issue.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextModel.java
kevinrushforth: OK to leave the javadoc. Thanks for getting rid of the `@since`

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/SimpleViewOnlyStyledModel.java
kevinrushforth: No changes needed.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledSegment.java
kevinrushforth: OK, but that still doesn't tell me what null does and how it is different from an empty list. That's really a larger issue though, so OK to defer.

kevinrushforth (commented):
The changes all look good. There is one issue related to the out-of-place convenience method. I looked more closely and I think this has to be changed (moved and renamed).

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/FileListFormatHandler.java
kevinrushforth: I see that the implementation of this method constructs an instance of the private `FileListStyledInput` class, so if the helper method moves out of this class, it would need to change slightly, calling `FileListFormatHandler.getInstance().createStyledInput(files, null)`.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/FileListFormatHandler.java
andy-goryachev-oracle: I think the base class description should be good enough - it says 

> @throws ClassCastException if the input data type is not supported by the **handler**

or do we need a separate and duplicate javadoc?  or, is there a way to append to the base class javadoc?

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/FileListFormatHandler.java
andy-goryachev-oracle: removed the method.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/FileListFormatHandler.java
kevinrushforth: > is there a way to append to the base class javadoc?

Yes. Look at my suggestion right before your comment.

apps/samples/RichTextAreaDemo/src/com/oracle/demo/richtext/editor/RichEditorDemoWindow.java
```diff
@@ -115,7 +116,13 @@ public RichEditorDemoWindow() {
                 } else {
                     TextPos p = editor.getDropTarget();
                     if (p != null) {
-                        FileListFormatHandler.handleDrop(editor, p, files);
+                        try {
+                            StyledInput in = FileListFormatHandler.getInstance().createStyledInput(files, null);
+                            editor.clearSelection();
+                            editor.replaceText(p, p, in);
```
kevinrushforth: Even better. Thanks.

kevinrushforth (commented):
I answered your question about appending to the base class docs: yes, it's possible.

Btw, it isn't an essential change, but a "nice to have".

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyledSegment.java
```diff
@@ -286,12 +286,16 @@ public String toString() {
     }
 
     /**
-     * Creates a StyledSegment which consists of a single inline Node.
+     * Creates a StyledSegment which consists of a single inline Node
+     * with the specified styles (the styles can be {@code null} which is equivalent
+     * to {@link StyleAttributeMap#EMPTY}.
```
kevinrushforth: Missing closing paren before the final `.` of the paragraph.

Minor: maybe add a `,` before `which`?

kevinrushforth (commented):
API looks good, with one missing `)` in the docs.

After you correct it, update the CSR and then I'll review.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/StyleAttribute.java
```diff
@@ -183,6 +204,23 @@ public boolean isDocumentAttribute() {
         }
     }
 
+    /// inline node style attribute
+    private static final class InlineNodeStyleAttribute<X> extends StyleAttribute<X> {
```
kevinrushforth, line 212: This addition has exposed a pair of related problems with the implementation of `equals` vs `hashCode` in `StyleAttribute`. The specific bug is that `equals` does not include a check on the new `isInlineNode()` attribute, whereas `hashCode`, by virtue of the fact that it includes the hash of the concrete subclass in its computation, effectively does. So at a minimum, you need to add a test for `isInlineImage()` in equals.

However, it also raises a more general problem: you should not include `getClass()` in the hash code unless it is also explicitly included in the check for `equals` or you risk breaking the equals / hashCode contract (as happened here). For many cases, the concrete type of the class should not matter when computing the hash, but if it does, then `equals` must check it. Having `equals` check the type one way (using the isXXXX() methods) and `hashCode` check it another (using the concrete class type) is fragile.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextModel.java
```diff
@@ -898,9 +918,14 @@ private static int whichCase(int off, int max, int start, int end) {
...
             for (RSegment seg : this) {
-                String text = seg.text();
                 StyleAttributeMap a = seg.attrs();
-                b.addSegment(text, a);
+                EmbeddedImage im = a.get(StyleAttributeMap.EMBEDDED_IMAGE);
+                if (im == null) {
+                    String text = seg.text();
+                    b.addSegment(text, a);
+                } else {
+                    b.addInlineNode(im::createNode, a);
```
kevinrushforth, line 927: What if an inline node also has a non-empty text string? It looks like this will silently drop all of the text. Will this cause problems with API that uses TextPos? Would it be better to split the segment or else throw an error in this case?

kevinrushforth (commented):
I left a couple inline comments / questions.

modules/jfx.incubator.richtext/src/main/java/com/sun/jfx/incubator/scene/control/richtext/TextCell.java
```diff
@@ -478,8 +499,113 @@ public void requestLayout() {
...
+        }
+
+        public void addRun(int length, StyleAttribute<?> a, CellContext.RunDecor type, String styleName) {
+            // compute offset
+            int count = flow.getChildren().size();
+            for (int i = lastCount; i < count; i++) {
+                Node n = flow.getChildren().get(i);
+                if (n instanceof Text t) {
+                    offset += t.getText().length();
+                } else if(n.isManaged()) { // ignoring highlights added in VFlow:824
```
kevinrushforth: Minor: missing space after `if`

apps/samples/RichTextAreaDemo/src/com/oracle/demo/richtext/rta/DemoModel.java
```diff
@@ -43,13 +43,28 @@
...
 /**
  * RichTextArea demo model.
  *
  * @author Andy Goryachev
  */
 public class DemoModel extends SimpleViewOnlyStyledModel {
+    private static final StyleAttributeMap OVERLAP1 = StyleAttributeMap.builder().
+        set(StyleAttributeMap.TEXT_HIGHLIGHT_1, Boolean.TRUE).
+        setFontSize(16).
+        build();
```
kevinrushforth: Minor suggestion: the usual pattern for multi-line fluent calls (e.g., builders, streams) is to put the `.` on the same line as the method.


```suggestion
    private static final StyleAttributeMap OVERLAP1 = StyleAttributeMap.builder()
        .set(StyleAttributeMap.TEXT_HIGHLIGHT_1, Boolean.TRUE)
        .setFontSize(16)
        .build();
```

modules/jfx.incubator.richtext/src/main/java/com/sun/jfx/incubator/scene/control/richtext/util/RichUtils.java
```diff
@@ -173,13 +175,16 @@ public static boolean isTouchSupported() {
...
+    /// ignoring unmanagement Nodes (highlights) and counting non-Text nodes
+    /// as having length=1.
     public static int getTextLength(TextFlow f) {
         int len = 0;
         for (Node n : f.getChildrenUnmodifiable()) {
             if (n instanceof Text t) {
                 len += t.getText().length();
-            } else {
-                // treat non-Text nodes as having 1 character
+            } else if(n.isManaged()) {
```
kevinrushforth: Minor: missing space after `if`.

modules/jfx.incubator.richtext/src/main/java/com/sun/jfx/incubator/scene/control/richtext/TextCell.java
```diff
@@ -478,8 +499,113 @@ public void requestLayout() {
...
+        }
+    }
+
+    /// keeps track of coalesced decorated runs
+    private static class Decorator {
+        private final TextFlow flow;
+        private int offset;
+        private int lastCount;
+        private final ArrayList<DecorationRun> runs = new ArrayList<>(4);
+        private HashMap<StyleAttribute<?>,DecorationRun> byType = new HashMap<>();
```
kevinrushforth: Minor: missing space after `,`

kevinrushforth (commented):
I reviewed most of the the code changes and didn't spot anything else. I did leave a couple minor formatting issues inline.

I ran the RTA editor demo, and tried both a copy/paste and a drag/drop of an image, but nothing happened. What am I doing wrong?

Other things I noticed:

If I highlight text and type in the middle of it, it doesn't apply the highlight to the inserted text. Character styles work as expected. Is this intentional?

A few follow-up Enhancements that seem worth considering:

1. Add another EmbeddedImage factory method that takes a URL

2. Store / read the mime type as part of the native RTA output format

3. Consider another EmbeddedImage factory method that takes an already-constructed Image. This might be a bit tricky when it comes to writing it: you would either need to encode it, e.g., to a lossless PNG format, or come up with a "raw" format that stores the raw image data and enough metadata to allow it to be decoded. This may or may not be worth it.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextModel.java
andy-goryachev-oracle: hm, not sure I understand: `addInlineNode()` does not care about the node properties; the model treats it as having a single character (' ', but perhaps it should be '\ufffd').  StyledSegment:307

kevinrushforth: What happens if you call `applyStyle` to a range of text using a `StyleAttributeMap` that contains `StyleAttribute.EMBEDDED_IMAGE`?

andy-goryachev-oracle: good point!  should be impossible.

kevinrushforth: Should be impossible? Or should be _prevented_? There is nothing I can see in the current API that would stop an application from doing that. Perhaps it should be explicitly disallowed? That could be handled with a follow-up bug, since it seems like a corner case that a well-behaved application wouldn't hit.

andy-goryachev-oracle: yes, should be impossible from the application view point, should be prevented/handled by the model.  will address this in a follow-up.

andy-goryachev-oracle:
> If I highlight text and type in the middle of it, it doesn't apply the highlight to the inserted text. Character styles work as expected. Is this intentional?

it's a bug in the rich editor demo, will addressed in a follow-up.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/FileListFormatHandler.java
```diff
@@ -0,0 +1,131 @@
...
+ * @since 27
+ */
+public class FileListFormatHandler extends DataFormatHandler {
+
+    private static final FileListFormatHandler instance = new FileListFormatHandler();
+
+    /**
+     * Constructor.
+     */
+    public FileListFormatHandler() {
```
kevinrushforth: This class is intended to be a singleton. Why does it have a public constructor?

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/FileListFormatHandler.java
```diff
@@ -0,0 +1,131 @@
...
+                        byte[] b = Files.readAllBytes(f.toPath());
+                        Image im = new Image(new ByteArrayInputStream(b), false);
+                        if (!im.isError()) {
+                            double w = im.getWidth();
+                            double h = im.getHeight();
+                            EmbeddedImage em = EmbeddedImageHelper.create(b, w, h, EmbeddedImage.FIT_WIDTH, EmbeddedImage.AUTO, true);
+                            StyleAttributeMap a = StyleAttributeMap.of(StyleAttributeMap.EMBEDDED_IMAGE, em);
+                            return StyledSegment.of(" ", a);
+                        }
+                    } catch (Throwable e) {
```
kevinrushforth, line 121: This should probably be be `catch (Exception e)` instead (we generally don't catch `Error` in normal operation)

kevinrushforth (commented):
I left a couple more inline comments. One more thing I spotted:

`RtfStyledOutput` doesn't handle embedded images. File a follow-up for this? I see a placeholder for PARAGRAPH attributes, would it be helpful to add a similar placeholder for inline images?

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/FileListFormatHandler.java
andy-goryachev-oracle: singleton to allow reuse of the default implementation, public constructor to allow subclassing for custom models.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/FileListFormatHandler.java
andy-goryachev-oracle: The intent was to catch _anything_ - there is little value in bubbling anything up at this point.

andy-goryachev-oracle:
> `RtfStyledOutput` doesn't handle embedded images. File a follow-up for this?

This is on the list - you can see an embedded image in RTF files on windows, but on mac you don't.  On mac, an attempt to save a file with an embedded image results in .rtfd file which is something else.

There is no RTF import either.

andy-goryachev-oracle:
Fixed an issue with pasting an image and typing next to an embedded image.  If you still see the problem, please let me know the exact steps to reproduce.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/FileListFormatHandler.java
kevinrushforth: None of the other `DataFormatHandler` subclasses have a public constructor. What makes this one different? How might a custom model override this?

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/FileListFormatHandler.java
kevinrushforth: _Anything_ usually doesn't include Error unless in a "run loop" (e.g., on the renderer thread) or when calling a handler on the JavaFX application thread, or similar, but we are not consistent on that at all in the JavaFX runtime, so I won't object.

kevinrushforth:
> Fixed an issue with pasting an image and typing next to an embedded image. If you still see the problem, please let me know the exact steps to reproduce.

Steps:

1. Run RichEditorDemo
2. Select any image from another app
3. Copy it to the clipboard
4. Paste it into the RichEditorDemo window

BUG: no image is pasted

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/FileListFormatHandler.java
andy-goryachev-oracle: made private

andy-goryachev-oracle:
`RichEditorDemo` does not import images (yet), only drag-and-drop.

The main issue is the process of converting `javafx.scene.image.Image` to a byte array for storage.  An application might decide to store all the images as PNG, or try to identify the best format between JPG/PNG, or do something completely different.

edit: filed https://bugs.openjdk.org/browse/JDK-8388276

kevinrushforth:
> RichEditorDemo does not import images (yet), only drag-and-drop.

I see. And more particularly, only drag and drop of local files it seems (I had tried doing a drag / drop of an https image from a browser and that didn't work).

I guess this gets back to RFE 3 that I suggested above about image support.

OK, I think all my points have been addressed. I'll do some final testing before approving.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/FileListFormatHandler.java
kevinrushforth: Thanks. Worth noting is that changing the constructor from public to private is a change in the API surface, so this will require amending the CSR.

andy-goryachev-oracle: updated the CSR.

modules/jfx.incubator.richtext/src/main/java/jfx/incubator/scene/control/richtext/model/RichTextModel.java
```diff
@@ -479,7 +479,7 @@ public void insertText(int offset, String text, StyleAttributeMap attrs) {
                         }
                         if (ix < toSplit.length()) {
                             String s2 = toSplit.substring(ix);
-                            insertSegment2(i, s2, a);
+                            insertSegment2(i, s2, RichUtils.filterOutNodeAttributes(a));
```
kevinrushforth, line 482: This (along with the later fix in this file) looks like it will handle the important case of text insertion not getting mistakenly styled with the EMBEDDED_IMAGE style. I don't think it will handle the case of calling `insert` or `applyStyle` programmatically, though. Since that would be an application error, it seems fine to do it as a follow-up bug.

kevinrushforth (approved):
LGTM. I left one comment recommending a follow-up bug.

aghaisas (approved):
Sanity checked the code changes. They look fine.

I tested on macOS.
I saw an issue in the demo (apart from the issues deemed to be fixed as a follow-up)
**Issue :**  I could not find a way to remove highlight or wavy underline once set.
Undo option removes only the last highlight/wavy underline. There should be an option to remove any selected highlight or wavy underline.

Ziad-Mid (approved):
LGTM, new attributes tested and works fine with latest changes.

andy-goryachev-oracle:
Thank you all for the review!

/integrate

andy-goryachev-oracle:
> **Issue :** I could not find a way to remove highlight or wavy underline once set.

It's a RichEditorDemo issue: the menu should have a check mark.  Select the text and use the same "Highlight -> Color *" menu to remove highlights.

PR #2197, opened 2026-06-25
8373452: DataFormat threading and API issues

andy-goryachev-oracle:
## Summary

This PR changes the behavior of `DataFormat` by allowing multiple instances that contain the same set of mime types (ids).

## Problem

There seems to be several issues with DataFormat API and implementation discovered during review of the `Clipboard`-related code:

1. `static DataFormat::lookupMimeType(String)` is not thread safe: while iterating over previously registered entries in the `DATA_FORMAT_LIST` another thread might create a new instance (DataFormat L227)

2. `public DataFormat(String...)` constructor might throw an `IllegalArgumentException` if one of the given mime types is already assigned to another `DataFormat`. The origin of this requirement is unclear, but one possible issue I can see is if the application has two libraries that both attempt to create a `DataFormat` for let's say `"text/css"`. Then, depending on the timing or the exact code path, an exception will be thrown for which the library(-ies) might not be prepared. The constructor is also not thread safe.

3. To avoid a situation mentioned in bullet 2, a developer would typically call `lookupMimeType()` to obtain the already registered instance, followed by a constructor call if such an instance has not been found. An example of such code can be seen in webkit/UIClientImpl:299 - but even then, despite that two-step process being synchronized, the code might still fail if *some other* library or the application attempts to create a new instance of DataFormat, since the constructor itself is not synchronized.

4. `DataFormat(new String[] { null })` is allowed but makes no sense!

5. The current implementation uses the `WeakReferenceQueue` which theoretically might, under certain conditions, allow the application to create mismatched `DataFormat`s.

Why do we need to have the registry of previously created instances? Unclear. My theory is that the DataFormat allows to have multiple mime-types (ids) - example being `DataFormat.FILES = new DataFormat("application/x-java-file-list", "java.file-list");` - and the registry was added to prevent creation of a `DataFormat` with just one id for some reason.

Also, I could not find the origin of the multi-id requirement, or the origin of the `java.file-list` id itself (it is not a valid mime type).

## Solution

The proposed solution is to relax the constraint on the constructor to allow multiple instances with the same set of mime types (ids).  This might require changing the application code from identity compare (`==`) to `.equals()`.

The change also adds synchronization on the registry inside the constructor and `DataFormat::lookupMimeType(String)`.

Additionally, the constructor checks for `null` ids, throwing an `IllegalArgumentException`.  (Quite possibly, we might want to disallow `new DataFormat((String[])null)` which is currently allowed.

## Compatibility Impact

The applications that relied on identity check ( `if(dataFormat == DataFormat.FILES)` ) must be changed to use the `equals()` method.  This might be a rare problem because most of the time the applications declare custom format as a static constant and use it to access the clipboard / dragboard data.

The constructor throws a `NullPointerException` if any of the ids is null.

## Alternatives

One alternative is to deprecate the multi-id constructor and only allow one mime type per `DataFormat`, also removing the `java.file-list` id.  We could then still keep `Set<String> DataFormat.getIdentifiers()` which will always return a set with one element.

## Notes

1. This PR presents an alternative solution to https://github.com/openjdk/jfx/pull/2006 (please refer to some very good comments there made by @mstr2 and @nlisker )
2. I think there is enough behavioral differences that a CSR is warranted.

---------

---------
### Progress

### Issue
 * [JDK-8373452](https://bugs.openjdk.org/browse/JDK-8373452): DataFormat threading and API issues (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2197/head:pull/2197` \
`$ git checkout pull/2197`

`$ git checkout pull/2197` \
`$ git pull https://git.openjdk.org/jfx.git pull/2197/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2197`

`$ git pr show -t 2197`

</details>

<a href="https://git.openjdk.org/jfx/pull/2197.diff">https://git.openjdk.org/jfx/pull/2197.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2197#issuecomment-4802664276)
</details>


modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
```diff
@@ -126,25 +120,39 @@ public class DataFormat {
...
      */
     public DataFormat(@NamedArg("ids") String... ids) {
-        DATA_FORMAT_LIST.cleanup();
-        if (ids != null) {
-            for (String id : ids) {
-                if (lookupMimeType(id) != null) {
-                    throw new IllegalArgumentException("DataFormat '" + id +
-                            "' already exists.");
+        if (ids == null) {
+            this.identifier = Collections.<String>emptySet();
```
nlisker:
```suggestion
            this.identifier = Set.of()
```

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
```diff
@@ -126,25 +120,39 @@ public class DataFormat {
...
+            // don't care about registry
+        } else {
+            this.identifier = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(ids)));
```
nlisker:
```suggestion
            this.identifier = Set.of(ids);
```
This also throws on `null`s (but NPE, not IAE; you can rethrow). It also throws on duplicates which we need to do anyway.

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
```diff
@@ -126,25 +120,39 @@ public class DataFormat {
...
+            // check for mismatched formats
+            boolean isNew = true;
+            synchronized (registry) {
+                for (String id : ids) {
+                    if (id == null) {
+                        throw new IllegalArgumentException("DataFormat id must not be null.");
+                    }
```
nlisker:
```suggestion
```

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
```diff
@@ -199,17 +207,13 @@ public final Set<String> getIdentifiers() {
...
-        if (identifier.equals(otherDataFormat.identifier)) {
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == this) {
             return true;
+        } else if (obj instanceof DataFormat f) {
+            return identifier.equals(f.identifier);
         }
-
         return false;
```
nlisker:
```suggestion
        }
        return (obj instanceof DataFormat df && identifier.equals(df.identifier);
```

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
```diff
@@ -126,25 +120,39 @@ public class DataFormat {
...
+                    if (id == null) {
+                        throw new IllegalArgumentException("DataFormat id must not be null.");
+                    }
+                    DataFormat f = registry.get(id);
+                    if (f != null) {
+                        if (!this.identifier.equals(f.identifier)) {
+                            throw new IllegalArgumentException("DataFormat '" + id + "' already exists.");
+                        }
+                        isNew = false;
+                    }
```
nlisker:
```suggestion
                    boolean noneExist = Collections.disjoint(this.identifie, registry.keySet());
```

nlisker (commented):
Added some suggestions, but haven't done a serious review.

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
andy-goryachev-oracle: the proposed code is no different?  in any case, throwing an NPE might be better in this case.

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
andy-goryachev-oracle: my version is better as it avoids scanning twice

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
nlisker: I'm proposing to remove this since the `null` check in `ids` was done with `Set.of(ids)`. It's unreachable code (unless `ids` is changed in a background thread maybe, but I don't think this is a real concern, and you can work with `identifier` instead anyway).

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
```diff
@@ -126,25 +117,40 @@ public class DataFormat {
...
     public DataFormat(@NamedArg("ids") String... ids) {
-        DATA_FORMAT_LIST.cleanup();
-        if (ids != null) {
-            for (String id : ids) {
-                if (lookupMimeType(id) != null) {
-                    throw new IllegalArgumentException("DataFormat '" + id +
-                            "' already exists.");
+        if (ids == null) {
+            this.identifier = Set.of();
+            // don't care about registry
```
nlisker: You can `return` here and avoid the big `else` branch.

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
nlisker: This claim might be too strong. From a performance perspective, the sizes of the collections are very small (a `DataFormat` will realistically have at most 3 ids and the total number of ids/mime-types that exist in the wild is probably in the dozens only), so performance is not really a consideration. However, if we do look at it, your implementation iterates over the array, which is O(n), with a hashed `get` which is O(1). `Collections.disjoint` is optimized to the given collections - it iterates over the smaller collection, which is O(n), and checks contains which is O(1). So I don't think your code gives better performance.

From a readability perspective, what your code says is "for each id, find a data format that uses it" (and if the `DataFormat` shares 2 or more ids, you will find and compare the same `DataFormat` twice or more). `disjoint` is clearer: "does any of the new ids exist already?"

In fact, I'm not sure why you're throwing only when not all the identifiers are the same, that is, why you specified "if one of the given ids is already assigned to another DataFormat **with a different set of ids**". If some are the same it's an exception, but if all of them are the same then it's not? Maybe I'm missing something.

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
andy-goryachev-oracle: I agree (I think I got confused by the github diff in the conversation).

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
andy-goryachev-oracle: my version is faster, but the difference is so minute it's irrelevant.  I can use `disjoint()` here.

> If some are the same it's an exception, but if all of them are the same then it's not?

This is the essence of the change.  We are to allow multiple DataFormat instances if they are equal, so as to avoid the unnecessary exceptions.

The reason we can't allow disjoint formats is because the `lookupMimeType()` logic will break, and that would be a more drastic breaking change.

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
andy-goryachev-oracle: I like big `else` branches :-)

andy-goryachev-oracle:
> If some are the same it's an exception, but if all of them are the same then it's not? Maybe I'm missing something.

We've got this `lookupMimeType()` which is supposed to return the right multi-id data format, and it will break if one registers mismatched formats:

```
DataFormat f1 = new DataFormat("a", "b");
DataFormat f2 = new DataFormat("b", "c");

// say whaat?
var what = DataFormat.lookupMimeType("b");
```

A better solution would be to disallow multi-id formats and require the single format id to be a valid mime type, but that would be a more drastic change.

andy-goryachev-oracle:
re: disjoint(): sorry, I don't want to deal with peculiarities of it (empty set etc.).

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
```diff
@@ -126,25 +117,37 @@ public class DataFormat {
...
+     * @throws NullPointerException if any of the ids is null
      */
     public DataFormat(@NamedArg("ids") String... ids) {
-        DATA_FORMAT_LIST.cleanup();
-        if (ids != null) {
-            for (String id : ids) {
-                if (lookupMimeType(id) != null) {
-                    throw new IllegalArgumentException("DataFormat '" + id +
-                            "' already exists.");
+        if (ids == null) {
```
nlisker:
```suggestion
        if (ids == null || ids.length == 0) {
```
I think we can shortcut an empty id list.

nlisker:
> re: disjoint(): sorry, I don't want to deal with peculiarities of it (empty set etc.).

What peculiarities? We are talking about sets of `String`s. How are empty sets a problem? If a set has no elements then it is disjoint; the docs are clear "Returns `true` if the two specified collections have no elements in common.", so any empty set returns `true`.

It's a suggestions to make the code clearer, ignore it if you want. As it is, it takes quite some effort to understand what the constructor does and it's mostly complexity for irrelevant optimizations (the registry will not be larger than a few dozen entries, and this is a stretch probably). I'd value readability here over saving a microsecond and write:
```java
	public DataFormat(String... ids) {
		if (ids == null || ids.length == 0) {
			identifier = Set.of();
			return;
		}

		identifier = Set.of(ids);

		if (registry.containsValue(this)) {
			// all ids exist in an existing data format - no need to change registry
			return;
		}
		if (!Collections.disjoint(identifier, registry.keySet())) {
			// existing ids can't be overridden - throw
			throw new IllegalArgumentException("An id already exists");
		}
		// no existing ids - add
		identifier.forEach(id -> registry.put(id, this));
	}
```

I also find the name `identifier` confusing, I would use `identifiers` since it's a collection.

modules/javafx.graphics/src/test/java/test/javafx/scene/input/DataFormatTest.java
```diff
@@ -101,4 +116,18 @@ public void testEqualsAndHashCode(DataFormat format, String mime1, String mime2)
...
+        assertThrows(NullPointerException.class, () -> {
+            String mime = null;
+            new DataFormat(mime);
+        });
+    }
+
+    @Test
+    public void nullArrayIsAllowedForCompatibilityReasons() {
+        String[] mimes = null;
+        new DataFormat(mimes);
```
Maran23:
```suggestion
        assertDoesNotThrow(() -> new DataFormat(mimes));
```

modules/javafx.graphics/src/test/java/test/javafx/scene/input/DataFormatTest.java
```diff
@@ -101,4 +116,18 @@ public void testEqualsAndHashCode(DataFormat format, String mime1, String mime2)
         assertFalse(uniqueFormat.equals(format));
         assertFalse(uniqueFormat.hashCode() == format.hashCode());
     }
+
+    @Test
+    public void noMoreNullMimeTypes() {
+        assertThrows(NullPointerException.class, () -> {
+            String mime = null;
+            new DataFormat(mime);
+        });
```
Maran23, line 131:
```suggestion
        String mime = null;
        assertThrows(NullPointerException.class, () -> {
            new DataFormat(mime);
        });
```

Recommendation is to always have one statement in the `assertThrows` / `assertDoesNotThrows`

andy-goryachev-oracle:
re: disjoint

`Collections.disjoint()` returns true with two empty sets.

And you are right, it's rather unlikely for any DataFormat (except FILES) to contain more than one element, so we are basically bikeshedding.

The main question is whether this PR solves the problem sufficiently, or should we ban multi-id data formats altogether?  I could not find an origin or any use of `java.file-list` id, it's not even a valid mime type.

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
nlisker, line 91: `DRAG_IMAGE` and `DRAG_IMAGE_OFFSET` are unused (and were unused before this PR). I don't know where they were supposed to be used, maybe there's some missing wiring somewhere. They could possibly be removed.

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
```diff
@@ -41,14 +37,9 @@
...
-     * A static cache of all DataFormats created and currently in use. This is needed
-     * by the underlying implementation, such that, given a mime type, we can determine
-     * the associated DataFormat. The OS level is going to supply us with a mime type
-     * (or other string based key), and we need to be able to map this back to the FX
-     * DataFormat.
-     */
-    private static final WeakReferenceQueue<DataFormat> DATA_FORMAT_LIST = new WeakReferenceQueue<>();
+    // A static registry of DataFormats for the purposes of checking against constructing DataFormats
+    // that contain mismatched mime types.
+    private static final HashMap<String,DataFormat> registry = new HashMap<>();
```
nlisker:
```suggestion
    private static final Map<String, DataFormat> registry = new HashMap<>();
```

We usually code to an interface, not an implementation.

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
```diff
@@ -187,7 +190,7 @@ public final Set<String> getIdentifiers() {
     @Override public int hashCode() {
         int hash = 7;
 
-        for (String id : identifier) {
+        for (String id : identifiers) {
             hash = 31 * hash + id.hashCode();
```
nlisker: The hash should be `identifiers.hashCode()` to match the `equals` check.

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
nlisker, line 217:
```suggestion
        if (mimeType == null || mimeType.isEmpty()) {
```

Minor and preexisting, so up to you.

nlisker (changes requested):
> The main question is whether this PR solves the problem sufficiently, or should we ban multi-id data formats altogether? I could not find an origin or any use of `java.file-list` id, it's not even a valid mime type.

In my opinion, banning multi-ids makes the model more reasonable. You can register all the IDs you want 1-by-1 (unless there are conflicts). I would even say that removing this class makes the model simpler because identifiers can be registered directly as strings, but I have a feeling that this will break a bit of code...
Whatever we can get away with is a plus as I see it.

I had an AI go over the fix because concurrency is tricky. Its findings, in short (these things are very verbose):
* It agrees that the concurrent access issue is resolved.
* It also came up with the scenario of having duplicate identifiers: `new DataFormat("a", "a")`, but they are handled by the contract of `Set.of`. I'm not sure what we want to do with such an oddity, but it might be worth a test case.
* It found the equals-inconsistent hash function that I commented about.
* It reminded that now the registered IDs will be held forever because we're not using a weak reference anymore. I don't think that's a problem, although what we have is first-come-first-serve, which is another argument for exposing the `Map` methods directly where you can use `putIfAbsent` or override with `put`.

Regarding the tests, if we're fixing a concurrency issue, there should be a concurrent construction test.

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
andy-goryachev-oracle: these are being used in `c.s.glass.ui.Clipboard` and `glass_dnd.cpp`

I think we should add a comment (it won't be visible in the API spec)

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
andy-goryachev-oracle: I prefer to know what I am dealing with.  The rule you mentioned is applicable to methods, or when the field can be instantiated with different types.  does not apply in this case.

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
andy-goryachev-oracle: please elaborate - the current implementation complies with the hashCode/equals contract (and is effectively unchanged btw).

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
nlisker: Definitely needs a comment.

modules/javafx.graphics/src/main/java/javafx/scene/input/DataFormat.java
nlisker: Two `equals` objects must have the same hashcode. This is not the case here: `DataFormat#equals` calls `Set#equals`, whose contract says that the elements need to be equal (order-independent). The current hash function iterates the elements in an unspecified order, producing order-dependent hashes. `Set.of("a", "b")` and `Set.of("b", "a")` are `equals` but have different hashes. By delegating to `Set#hashCode`, we align it with `Set#equals` and have `Set` worry about matching its contract.

nlisker: Regarding the difference from the current `master` code, creating a `DataFormat` with identical IDs in a different order to an existing one used to throw, but now it's ignored, so you get your new instance. If you now try to `Map#get` with it, you will get `null` because the first-come-first-serve `DataFormat` has a different hashcode than yours has. Imagine the surprise.

andy-goryachev-oracle: you are right, this needs to be fixed.  thanks!

nlisker: The AI found this, I'll give it positive reinforcement.

andy-goryachev-oracle: (though I think it will not be ignored because it the ids will clash)

andy-goryachev-oracle: > The AI found this,

we all have our little secrets, the important thing is that no text or code was copied during the process per the [OpenJDK Interim Policy on Generative AI](https://openjdk.org/legal/ai)

nlisker: It's a review, so we're in the clear as per points 4 and 5 :)

andy-goryachev-oracle: yep.

modules/javafx.graphics/src/test/java/test/javafx/scene/input/DataFormatTest.java
```diff
@@ -101,4 +122,83 @@ public void testEqualsAndHashCode(DataFormat format, String mime1, String mime2)
...
+                public void run() {
+                    int num = 0;
+                    while (run.get()) {
+                        String id = "test-data-format-" + num;
+                        ++num;
+                        if (num > 10) {
+                            num = 0;
+                        }
+                        try {
+                            DataFormat f = new DataFormat(id);
```
nlisker, line 164: `f` is unused, so no need for the LHS.

modules/javafx.graphics/src/test/java/test/javafx/scene/input/DataFormatTest.java
```diff
@@ -101,4 +122,83 @@ public void testEqualsAndHashCode(DataFormat format, String mime1, String mime2)
...
+        String[] mimes = null;
+        assertDoesNotThrow(() -> {
+            new DataFormat(mimes);
+        });
+    }
+
+    @Timeout(10)
+    @Test
+    public void concurrencyWithSingleID() throws Exception {
+        long duration = 5_000;
```
nlisker: 5 seconds for a single test is probably too long. Running all the tests in the controls module takes about 2 minutes. I tested with 0.5 seconds and there's no issue.

The way I do these race tests is with a `@RepeatedTest(M)` instead of the sleep. Each tests runs N concurrent threads (N being "a lot") and the time it takes them to finish is hardware-based, but running them M times deterministically gives the non-flakiness guarantee. This way you don't need to set a hardware-independent duration and the tests completes as fast as possible. It's not like faster hardware requires more runs.

modules/javafx.graphics/src/test/java/test/javafx/scene/input/DataFormatTest.java
andy-goryachev-oracle: I was going to add another check and forgot.

modules/javafx.graphics/src/test/java/test/javafx/scene/input/DataFormatTest.java
andy-goryachev-oracle: switched to `@RepeatedTest and 100k iterations per test (~0.2 sec per test), total test is about 2.3 seconds on mac M1

nlisker: The number of iterations is way more than is realistically needed for a test. Takes 6 seconds for me with a `threadCount` of 49.
In any case, does the test fail for you without the added synchronization?

andy-goryachev-oracle: what do you mean "without the added synchronization"?

what would you suggest for the parameter values?

andy-goryachev-oracle: oh, you are asking whether it fails without synchronization - no it does not, the starts should align just right for it to fail, really.

I don't think we should test the concurrency contract of the `HashMap` though.   Do you have a specific suggestion?

nlisker: I mean the `synchronized` blocks. You're testing concurrent access. If I remove those blocks, I expect the test to fail.

The parameter values I use depend on what produces a failure reliably.

nlisker: Well, what does this test test then? Your points 1 and 5 are about lack of thread safety. A test that they are resolved should fail before the fix and not fail after, unless you think that it's trivial, and then there's no need for a test.

andy-goryachev-oracle: well, it did fail twice in a row with threadCount = 49, but would pass afterwards.

we **know** the `HashMap` will fail if it's not synchronized, what is the point of testing it?

andy-goryachev-oracle: > unless you think that it's trivial

that's why I did not include the test initially.  I can either keep the test (esp. since it did fail for me twice w/o syn'd keyword), or remove it.

nlisker: I'll leave it to another reviewer since I can't test the correctness here.

nlisker (approved):
Looks good, except for the concurrency test that I can't reproduce.

PR #454, opened 2021-04-03
8264770: BidirectionalBinding should use InvalidationListener to prevent boxing

mstr2:
The internal BidirectionalBinding class implements bidirectional bindings for JavaFX properties. The design intent of this class is to provide specializations for primitive value types to prevent boxing conversions (cf. specializations of the Property class with a similar design intent).

However, the primitive BidirectionalBinding implementations do not meet the design goal of preventing boxing conversions, because they implement ChangeListener.

ChangeListener is a generic SAM interface, which makes it impossibe to invoke an implementation of ChangeListener::changed with a primitive value (i.e. any primitive value will be auto-boxed).

The boxing conversion happens, as with all ChangeListeners, at the invocation site (for example, in ExpressionHelper). Since the boxing conversion has already happened by the time any of the BidirectionalBinding implementations is invoked, there's no point in using primitive specializations of BidirectionalBinding after the fact.

This issue can be solved by having BidirectionalBinding implement InvalidationListener instead, which by itself does not incur a boxing conversion. Because bidirectional bindings are eagerly evaluated, the observable behavior remains the same.

I've filed a bug report with the same title.

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [x] Change must be properly reviewed

### Issue
 * [JDK-8264770](https://bugs.openjdk.java.net/browse/JDK-8264770): BidirectionalBinding should use InvalidationListener to prevent boxing

### Reviewers
 * [Ambarish Rapte](https://openjdk.java.net/census#arapte) (@arapte - **Reviewer**)
 * [Kevin Rushforth](https://openjdk.java.net/census#kcr) (@kevinrushforth - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jfx pull/454/head:pull/454` \
`$ git checkout pull/454`

Update a local copy of the PR: \
`$ git checkout pull/454` \
`$ git pull https://git.openjdk.java.net/jfx pull/454/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 454`

View PR using the GUI difftool: \
`$ git pr show -t 454`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.java.net/jfx/pull/454.diff">https://git.openjdk.java.net/jfx/pull/454.diff</a>

</details>

PR #2200, opened 2026-07-03
8387655: java.lang.NullPointerException: Cannot invoke "java.lang.Boolean.booleanValue()" because "newValue" is null

jperedadnr:
This PR captures the `currentValue` of the fired value change event in `ExpressionHelper$Generic::fireValueChangedEvent`,  before entering the notification loop, in order to pass it down to all the change listeners. 

This prevents a potential issue that could happen if all the change listeners are removed before the notification loop ends: the `currentValue` is nullified, and the pending listeners would unexpectedly receive null, instead of the value that was correctly sent to the previous listeners.

Some unit tests have been included, all of them but one pass before and after the proposed fix, and the last one fails before the fix and passes after it. 

The scenario that describes this failing test is: 
- A property has attached a few InvalidationListeners and a few ChangeListeners (so the Generic path is followed, and not the SingleChange). 
- At some point the property changes to a non-null `newValue`, and all the listeners get notified, one by one, in the same order as they were registered. 
- If, for any reason, and _while_ the different change listeners are being notified, all the change listeners are removed,  the pending listeners will, unexpectedly,  receive `null`, though the listeners that were notified before that happens received the expected `newValue`.

For simplicity, the test simulates that exact moment by removing all the change listeners when `listenerA` receives `newValue` from the `changed` notification. Then the `currentValue` gets nullified, and then the pending listeners in the notification list `curChangeList` (`listenerB` and `listenerC`, as they were registered after `listenerA`) get `null`. There could be other triggers, as long as they happen on the same thread, as part of the same call stack that started `fireValueChangedEvent`.

With the fix, by capturing the `currentValue` into `newValue` before entering the loop, then `B` and `C` will still receive `newValue`, regardless any possible change done to `currentValue` in the process.

---------

---------
### Progress

### Issue
 * [JDK-8387655](https://bugs.openjdk.org/browse/JDK-8387655): java.lang.NullPointerException: Cannot invoke "java.lang.Boolean.booleanValue()" because "newValue" is null (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2200/head:pull/2200` \
`$ git checkout pull/2200`

`$ git checkout pull/2200` \
`$ git pull https://git.openjdk.org/jfx.git pull/2200/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2200`

`$ git pr show -t 2200`

</details>

<a href="https://git.openjdk.org/jfx/pull/2200.diff">https://git.openjdk.org/jfx/pull/2200.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2200#issuecomment-4876428595)
</details>

### DIFF
--- a/modules/javafx.base/src/main/java/com/sun/javafx/binding/ExpressionHelper.java
+++ b/modules/javafx.base/src/main/java/com/sun/javafx/binding/ExpressionHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -381,9 +381,13 @@ public abstract class ExpressionHelper<T> extends ExpressionHelperBase {
                 if (curChangeSize > 0) {
                     final boolean changed = (currentValue == null)? (oldValue != null) : !currentValue.equals(oldValue);
                     if (changed) {
+                        // Capture currentValue, to pass down the same value to all listeners, given that
+                        // currentValue could be nullified if all the change listeners are removed. If that happens
+                        // during the notification loop, the pending listeners should still receive the correct value.
+                        final T newValue = currentValue;
                         for (int i = 0; i < curChangeSize; i++) {
                             try {
-                                curChangeList[i].changed(observable, oldValue, currentValue);
+                                curChangeList[i].changed(observable, oldValue, newValue);
                             } catch (Exception e) {
                                 Thread.currentThread().getUncaughtExceptionHandler().uncaughtException(Thread.currentThread(), e);
                             }
--- a/modules/javafx.base/src/test/java/test/com/sun/javafx/binding/ExpressionHelperTest.java
+++ b/modules/javafx.base/src/test/java/test/com/sun/javafx/binding/ExpressionHelperTest.java
@@ -812,7 +812,7 @@ public class ExpressionHelperTest {
             newValueRef.set(newValue);
         };
         final ChangeListener<Object> listenerA = (observable, oldValue, newValue) ->
-                ExpressionHelper.removeListener(helper, listenerB);
+                helper = ExpressionHelper.removeListener(helper, listenerB);
 
         helper = ExpressionHelper.addListener(helper, observable, listenerA);
         helper = ExpressionHelper.addListener(helper, observable, listenerB);
@@ -838,7 +838,7 @@ public class ExpressionHelperTest {
                 list.add(newValue);
 
         final ChangeListener<Object> listenerA = (observable, oldValue, newValue) ->
-                ExpressionHelper.removeListener(helper, listenerB);
+                helper = ExpressionHelper.removeListener(helper, listenerB);
 
         helper = ExpressionHelper.addListener(helper, observable, listenerA);
         helper = ExpressionHelper.addListener(helper, observable, listenerB);
--- a/modules/javafx.base/src/test/java/test/com/sun/javafx/binding/ExpressionHelperTest.java
+++ b/modules/javafx.base/src/test/java/test/com/sun/javafx/binding/ExpressionHelperTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2025, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -29,6 +29,8 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.ArrayList;
 import java.util.BitSet;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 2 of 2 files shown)


kevinrushforth:
As a reminder, per the [OpenJDK Interim AI Policy](https://openjdk.org/legal/ai), please do not enable Copilot to post review comments to a PR in any repo the openjdk GitHub organization. You are welcome to use Copilot or other AI tools offline to review pull requests as long as you don't contribute code or comments generated by those tools.

tsayao:
I has do be disabled, comes enabled by default.

kevinrushforth:
> I has do be disabled, comes enabled by default.

Yes, but unless it has changed recently, you need to explicitly request a review from Copilot by adding it as a Reviewer of a PR. That's the step we are asking developers not to take.

jperedadnr:
Sorry, my bad. I had "Automatic Copilot code review" as Thiago describes, so I didn't request a review from Copilot. 
I've disable it now, so should I close this PR and open a new one?

kevinrushforth:
> Sorry, my bad. I had "Automatic Copilot code review" as Thiago describes, so I didn't request a review from Copilot.

Thanks, that's good to know. And no worries.

> I've disable it now, so should I close this PR and open a new one?

No need. Let's proceed with this.

mstr2:
While I appreciate the effort, I think that a better use of the time spent fixing this edge case in `ExpressionHelper` would have been to review #1081. This narrow patch ensures that listeners will not see a null `newValue` in a very specific case, but they can still see plainly wrong new values in many other cases.

`ExpressionHelper` is not salvageable.

jperedadnr:
Hmm, one other thing that went wrong with this PR: I took the wrong JBS issue...

It should be: https://bugs.openjdk.org/browse/JDK-8387655 instead of https://bugs.openjdk.org/browse/JDK-8387626, for which I created #2202.

I've edited the title and description now. 

@kevinrushforth Should I close this and create a new one instead?

@mstr2 Thanks for your review. Given that the JBS info was not related to this PR, could you please have a look again? Also, check this [comment](https://bugs.openjdk.org/browse/JDK-8387655?focusedId=14890782&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14890782).

kevinrushforth:
> Hmm, one other thing that went wrong with this PR: I took the wrong JBS issue...
> 
> It should be: https://bugs.openjdk.org/browse/JDK-8387655 instead of https://bugs.openjdk.org/browse/JDK-8387626, for which I created #2202.
> 
> I've edited the title and description now.
> 
> @kevinrushforth Should I close this and create a new one instead?

No need. Skara will track the changed title and apply the PR to the right bug now that you fixed it.

andy-goryachev-oracle:
> While I appreciate the effort, I think that a better use of the time spent fixing this edge case in ExpressionHelper would have been to review https://github.com/openjdk/jfx/pull/1081.

#1081 has a merge conflict and no movement.  I'd rather see the platform moving forward than wait for years for some grand improvement that never materializes.  Having said that, does #1081 fix this particular issue (no way to check quickly, it seems to have a merge conflict).

mstr2:
> > While I appreciate the effort, I think that a better use of the time spent fixing this edge case in ExpressionHelper would have been to review #1081.
> 
> #1081 has a merge conflict and no movement. I'd rather see the platform moving forward than wait for years for some grand improvement that never materializes. Having said that, does #1081 fix this particular issue (no way to check quickly, it seems to have a merge conflict).

It has already materialized, the code exists. Framing John’s work as “no movement” is a bit rich when you (and others) have been stalling this PR for three years through non-participation. What a way to treat a community of volunteers... 😞

Maran23:
> Having said that, does #1081 fix this particular issue (no way to check quickly, it seems to have a merge conflict).

I just checked that, the PR fixes this issue. 
cc @hjohn, as you wrote in https://github.com/openjdk/jfx/pull/1081#issuecomment-4509121523 - Here is a real bug that would have already been solved by your changes.
Additionally, the `listenerB` does not run as it is removed - as expected.

I understand that the PR is bigger than this localized fix. On the other hand, the PR has been around for three years now, and John is very knowledgeable about properties and bindings. And I tested with multiple big applications back then and found no regressions.

And on the other hand, most developers don't use bindings that 'complex' or learned to avoid them (like me), because removing a listener in another is not that common and can create problems very fast, as we can see in the reproducer.

andy-goryachev-oracle:
What should we do then?  Should we focus on #1081, or attempt this localized fix?

kevinrushforth:
> What should we do then? Should we focus on #1081, or attempt this localized fix?

Ultimately, we should spend the time on #1081 but it won't make jfx27. It's a large fix with a behavioral change that will need a lot of testing and should go in earlier in the release cycle of a release, and not a week before rampdown.

If this is a good and safe fix, and it can be reviewed in time to make jfx27, then this localized fix might make sense (which is what I mentioned in the JBS comment). So the question is: Is it?

mstr2:
> What should we do then? Should we focus on #1081, or attempt this localized fix?

At surface level, this PR is a very small change: it captures the current value before invoking listeners, it's a one-line change. So it might seem like a pretty safe and easy fix, an easy win.

However, we already know for a fact that capturing the current value doesn't ensure that the reported values are correct. It's simply not possible. This PR proposes to change listener notifications to make it work in one edge case, but it won't solve the underlying problem and fail to work in many other edge cases. How do we know that this is actually a good trade-off, given that we're dealing with core infrastructure? How do we ensure that we don't introduce regressions in cases that have been working so far, possibly only accidentally? This solution is not _correct_, at best it's _more correct_ than what we have, but still defective.

I'm not against this PR in principle, and I can be convinced that _more correct_ is better than _less correct_, but it's still a defective solution compared to #1081, and one that needs quite a bit of testing. I've been bitten by something like this before: With #454, I proposed to have bidirectional bindings use invalidation listeners instead of change listeners. It seemed like a easy improvement, went through code review without problems, and just like that, we introduced a serious regression that was only discovered after release.

hjohn:
I've resolved the conflict in #1081 just in case.

kevinrushforth:
We have two choices for this PR:

1. Proceed with this PR for jfx27. Even though it is at best an incomplete fix for the problem, it does avoid an NPE in a specific use case
2. Abandon this PR and wait until jfx28 when the underlying problem will be fixed by #1081

Presuming this PR has no ill effects, that is, there are no cases where it misbehaves or has worse behavior than today, then 1 seems a reasonable option. Paraphrasing @mstr2, If we think this is "_more correct_" than what we have today, then we can proceed on the principle "that _more correct_ is better than _less correct_".

Either way, we should continue to push on #1081 and get it in early in jfx28 -- some time this month would be great. I note that if we go with option 1, we would likely want to revert it as part of #1081, since it would no longer be necessary.

@mstr2 @hjohn What is your recommendation?

nlisker:
> 2\. Abandon this PR and wait until jfx28 when the underlying problem will be fixed by [8290310: ChangeListener events are incorrect or misleading when a nested change occurs #1081](https://github.com/openjdk/jfx/pull/1081)

If #1081 is really going to be implemented in 28 then the potential damages are limited to 27 only: if this PR is abandoned, we live with this issue for 1 (more) version only; if it's integrated, we will live with whatever regression it causes (if it causes) also for 1 version only. I would ask @jperedadnr how significant are the issues caused by this and can they be worked around in the meantime?

hjohn:
> We have two choices for this PR:
> 
> 1. Proceed with this PR for jfx27. Even though it is at best an incomplete fix for the problem, it does avoid an NPE in a specific use case
> 2. Abandon this PR and wait until jfx28 when the underlying problem will be fixed by [8290310: ChangeListener events are incorrect or misleading when a nested change occurs #1081](https://github.com/openjdk/jfx/pull/1081)
> 
> Presuming this PR has no ill effects, that is, there are no cases where it misbehaves or has worse behavior than today, then 1 seems a reasonable option. Paraphrasing @mstr2, If we think this is "_more correct_" than what we have today, then we can proceed on the principle "that _more correct_ is better than _less correct_".
> 
> Either way, we should continue to push on #1081 and get it in early in jfx28 -- some time this month would be great. I note that if we go with option 1, we would likely want to revert it as part of #1081, since it would no longer be necessary.
> 
> @mstr2 @hjohn What is your recommendation?

`ExpressionHelper` won't disappear entirely yet with #1081 -- it is still used by the Bean adapters and (surprisingly) in `TextInputControl`.

I could include a full removal in #1081 -- but as that PR is already tough to review, I think that's best done as a follow up. So, with that in mind, I think we can include this "more correct" solution, and it will disappear automatically with the complete removal of `ExpressionHelper`.

If #1081 is indeed in early in JFX 28, then I think it is likely `ExpressionHelper` can be fully removed before 28 is released.

hjohn (approved):
I took a closer look at the change, and I think it is fairly harmless (famous last words :) ).

The `currentValue` is a field that is sometimes updated when the listener lists changes. Specifically, it is `null`ed (so it isn't referring to something that could be GC'd) when all change listeners are removed.

If that happens while a notification is in progress, then instead of the clearly intended value, `null` is sent because every listener callback uses the field's value directly. Capturing the value at the start of the loop solves this.

andy-goryachev-oracle (approved):
LGTM.  I agree with @hjohn - this does look harmless, hopefully https://github.com/openjdk/jfx/pull/1081 will get in early in jfx28.

My only comment is whether tests in general should use public APIs instead of internal classes like `ExpressionHelper`.  Once #1081 is in these tests will likely break, right?

jperedadnr:
> My only comment is whether tests in general should use public APIs instead of internal classes like ExpressionHelper. Once https://github.com/openjdk/jfx/pull/1081 is in these tests will likely break, right?

But that will happen with the already existing `ExpressionHelperTest` file (more than 30 tests), and many other tests that are using internal classes in test.com.sun.javafx... It is not always possible to just use public API/shims to get to test internal classes, is it?

andy-goryachev-oracle:
Right, there is a value in testing some aspects of the implementation.  I am just poining out the obvious - the tests that rely on public spec survive refactoring/redesign.

jperedadnr:
> I would ask @jperedadnr how significant are the issues caused by this and can they be worked around in the meantime?

@nlisker as I described in the [JBS ticket](https://bugs.openjdk.org/browse/JDK-8387655?focusedId=14890583&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14890583), the issue happened with an internal implementation in JavaFX classes (PopupWindows and focusedProperties), but I could finally isolate to the much more simple BooleanPropertyTest test. 

There could probably be some workarounds, given that the root cause of the problem is now well known, like adding a dummy change listener just to not get _all_ the listeners removed in the real scenario, but of course this shouldn't be needed once this small fix is in (and hopefully #1081).

PR #2201, opened 2026-07-03
8268657: Looked-up color fails for -fx-background-color in JavaFX CSS file

Maran23:
This one took me several days to fix. Below I will try to explain every information I found out while debugging and fixing this.

Originally I found this weird CSS issue already years ago, hence my comment in the issue.
It happened because `ControlsFX` secretly changes your `Scene` root to its `DecorationPane` (and attaches your old root under it). This seems to trigger CSS errors sometimes that do not make sense and the UI seems to be fine as well.

I tried to write several tests and scenarios that should work or not work. So some tests succeed without the fix and also after.
The fix is only in `CssStyleHelper` and consists of two changes. Both for the `firstStyleableAncestor`, which has some issues.

The `firstStyleableAncestor` was introcuded in https://bugs.openjdk.org/browse/JDK-8090462, so this is technically somewhat a regression from there.

### The fix

A child can change the scene structure (e.g. change the root) as a result from processing its CSS. I wrote multiple tests to show that. Because of this scenario, we need to change how the `firstStyleableAncestor` is found.

1.  The `firstStyleableAncestor` is now found by eagerly creating its `styleHelper` (before, we just checked if it has one). If this worked (-> that node is styleable), we have our ancestor.
  - 1.1 In `createStyleHelper`, we already traverse the parents to get the `depth`, so we can optimize our `firstStyleableAncestor` search by hanging into the loop and check if this is our styleable ancestor

2. In very rare cases `firstStyleableAncestor` can be null for a non-root `Node`. This, again can happen if something changes while we are processing the CSS. Normally CSS processing is top-down. But when a child changes a parent (or root), the structure changes and in this case we could no longer have a `firstStyleableAncestor`. So both CSS processing methods `resolveRef` and `getInheritedStyle` will search for a styleable ancestor if there is none set.
  - 2.1. This usually never happens and for `Scene` roots this will be a noop.

#### Performance

Performance wise, I could not see any problem or regression. We still cache our `firstStyleableAncestor` (and even reuse a preexisting loop). 
We might create a parent `styleHelper` more early in rare scenarios, but those will be created anyway the next pulse and usually are reused then.

I also avoid to create any null `WeakReference` objects, as `resolveRef` could be hot path if the user hovers e.g. over table entries. So no objects are created in this case.

There is https://bugs.openjdk.org/browse/JDK-8187955 which has an idea how to get rid of a recursion which I think is a good step if we want to optimize this further. But the ticket has not much information, so it is hard to completely get what the author meant.

##### What I tried before

At first, I thought this problem is only related to the `Scene` root node. So I first implemented a fix on the `rootProperty`, always prcoessing css for the new root (immediately, so we create a new `styleHelper` now and not the next pulse) or nulling the `styleHelper` on the old root.

But then I checked if this issue can be reproduced with non-root `Node`s as well, so I wrote tests for it that proves that this is a generic problem not related to the root `Node` (it just so happens that `modena.css` and usually all other stylesheets usually set the colors on `.root { ... }` and not on a child node).

With that information, the problem must be inside `CssStyleHelper` and not in `Scene#rootProperty` or `Node#setScenes`.

---------

---------
### Progress

### Issue
 * [JDK-8268657](https://bugs.openjdk.org/browse/JDK-8268657): Looked-up color fails for -fx-background-color in JavaFX CSS file (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2201/head:pull/2201` \
`$ git checkout pull/2201`

`$ git checkout pull/2201` \
`$ git pull https://git.openjdk.org/jfx.git pull/2201/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2201`

`$ git pr show -t 2201`

</details>

<a href="https://git.openjdk.org/jfx/pull/2201.diff">https://git.openjdk.org/jfx/pull/2201.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2201#issuecomment-4878140002)
</details>

### DIFF
--- a/modules/javafx.graphics/src/test/java/test/javafx/scene/CssStyleHelperTest.java
+++ b/modules/javafx.graphics/src/test/java/test/javafx/scene/CssStyleHelperTest.java
@@ -972,7 +972,7 @@ public class CssStyleHelperTest {
      */
     @Test
     void testLookupResolvesWithPseudoClassAndIntermediatePane() {
-        var errors = StyleManager.errorsProperty();
+        var errors = CssParser.errorsProperty();
         errors.clear();
 
         String theme = toDataURL("""
@@ -1006,7 +1006,7 @@ public class CssStyleHelperTest {
      */
     @Test
     void testColorLookupClassCastExceptionWhenStylesheetNotOnRoot() {
-        var errors = StyleManager.errorsProperty();
+        var errors = CssParser.errorsProperty();
         errors.clear();
 
         String themeA = toDataURL("""
@@ -1038,7 +1038,7 @@ public class CssStyleHelperTest {
      */
     @Test
     void testColorLookupClassCastExceptionAfterNewRootSwap() {
-        var errors = StyleManager.errorsProperty();
+        var errors = CssParser.errorsProperty();
         errors.clear();
 
         String themeA = toDataURL("""
@@ -1072,7 +1072,7 @@ public class CssStyleHelperTest {
      */
     @Test
     void testLookupResolvesAfterRootTransitionToStateRootSwap() {
-        var errors = StyleManager.errorsProperty();
+        var errors = CssParser.errorsProperty();
         errors.clear();
 
         String theme = toDataURL("""
@@ -1115,7 +1115,7 @@ public class CssStyleHelperTest {
      */
     @Test
     void testLookupResolvesAfterChildTransitionToStateRootSwap() {
-        var errors = StyleManager.errorsProperty();
+        var errors = CssParser.errorsProperty();
         errors.clear();
 
         String theme = toDataURL("""
--- a/modules/javafx.graphics/src/test/java/test/javafx/scene/CssStyleHelperTest.java
+++ b/modules/javafx.graphics/src/test/java/test/javafx/scene/CssStyleHelperTest.java
@@ -29,6 +29,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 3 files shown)


modules/javafx.graphics/src/test/java/test/javafx/scene/CssStyleHelperTest.java
```diff
@@ -965,6 +967,196 @@ public void mediaQueryRemovalShouldNotInterruptTransitionsDuringReset() {
         assertEquals(List.of(1.0, 2.0, 1.5, 1.0), trace);
     }
 
+    /**
+     * PseudoClass should correctly pass down even if there are intermediate unstyled panes inbetween.
+     */
+    @Test
+    void testLookupResolvesWithPseudoClassAndIntermediatePane() {
+        var errors = StyleManager.errorsProperty();
```
andy-goryachev-oracle: should we use `CssParser.errorsProperty()` instead here?

(they are the same, but the `CssParser` is a public API while this one is not)

andy-goryachev-oracle (approved):
This looks **very** good, thanks for the set of tests (which fail as expected in master).

modules/javafx.graphics/src/test/java/test/javafx/scene/CssStyleHelperTest.java
Maran23: I don't see a reason against using `CssParser`, so sure, changed just now

andy-goryachev-oracle (approved):

hjohn (approved):
Nice fix!

PR #2202, opened 2026-07-06
8387626: java.lang.NullPointerException: Cannot invoke "com.sun.glass.ui.Window.isMinimized()" because "this.platformWindow" is null

jperedadnr:
This PR removes the stage from the active windows list after a destroy notification, preventing a possible NPE in `WindowStage::setIconified`. 

A system test that fails before the proposed fix, and passes after it, has been included.

---------

---------
### Progress

### Issue
 * [JDK-8387626](https://bugs.openjdk.org/browse/JDK-8387626): java.lang.NullPointerException: Cannot invoke "com.sun.glass.ui.Window.isMinimized()" because "this.platformWindow" is null (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2202/head:pull/2202` \
`$ git checkout pull/2202`

`$ git checkout pull/2202` \
`$ git pull https://git.openjdk.org/jfx.git pull/2202/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2202`

`$ git pr show -t 2202`

</details>

<a href="https://git.openjdk.org/jfx/pull/2202.diff">https://git.openjdk.org/jfx/pull/2202.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2202#issuecomment-4893182044)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassWindowEventHandler.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassWindowEventHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008, 2025, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -133,6 +133,7 @@ class GlassWindowEventHandler extends Window.EventHandler implements Supplier<Vo
                 break;
             case WindowEvent.DESTROY:
                 stage.setPlatformWindowClosed();
+                WindowStage.removeActiveWindow(stage);
                 stage.stageListener.closed();
                 break;
             case WindowEvent.CLOSE:
--- a/tests/system/src/test/java/test/javafx/stage/DestroyWindowTest.java
+++ b/tests/system/src/test/java/test/javafx/stage/DestroyWindowTest.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright (c) 2026, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+package test.javafx.stage;
+
+import javafx.application.Application;
+import javafx.application.Platform;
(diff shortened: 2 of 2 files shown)


modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassWindowEventHandler.java
```diff
@@ -133,6 +133,7 @@ public Void get() {
                 break;
             case WindowEvent.DESTROY:
                 stage.setPlatformWindowClosed();
+                WindowStage.removeActiveWindow(stage);
```
andy-goryachev-oracle, line 136: The fix seems to be working (tested on macOS and windows).

One peculiar thing I noticed is that if I set a breakpoint in WindowStage:582 in the the master branch and run the test hitting the breakpoint, it won't fail in macOS (but fails in windows).

The main question is why `removeActiveWindow()` is called before the `stage.stageListener.closed()`, somehow it does not seem right.

I wonder if the order should have been 
```
            case WindowEvent.DESTROY:
                stage.stageListener.closed();
                stage.setPlatformWindowClosed();
```

In fact, the test passes in this case without needing `removeActiveWindow()`, as it's being called inside `stage.stageListener.closed();`

What do you think about that?

jperedadnr: As it is before this PR:

When `stage.setPlatformWindowClosed()` is called, it clears `WindowStage::platformWindow`,
`stage.stageListener.closed()` ends up calling `WindowStage::handleFocusDisabled`, which takes the last windowStage in the `activeWindows` list, which is not null, but its platformWindow is, causing the NPE.

With my proposed fix:
`stage.setPlatformWindowClosed()` -> clears `WindowStage::platformWindow`,
`WindowStage.removeActiveWindow(stage)` -> removes windowStage from the `activeWindows` list,
`stage.stageListener.closed()` -> `WindowStage::handleFocusDisabled` bails out, no active windows, so no NPE.

With your suggestion:
`stage.stageListener.closed()` -> `WindowStage::handleFocusDisabled` which takes the last windowStage, which still has a valid `platformWindow`, so no NPE, and then `setVisible(false)` removes windowStage from the `activeWindows` list,
`stage.setPlatformWindowClosed()` -> does nothing, the platformWindow was already cleared, so this call seems redundant.

In both cases the NPE is fixed, but it seems that this will do it with less calls: 
```
case WindowEvent.DESTROY:
                stage.stageListener.closed();
```

Edit: However, having a closer look, it seems that `setPlatformWindowClosed()` is needed in the first place after all: `Window.notifyDestroy()` emits `WindowEvent.DESTROY` before it sets `ptr = 0`, so
if `stage.stageListener.closed()` runs first, it can call `window.hide()` -> `peer.close()` -> `WindowStage.close()` -> `platformWindow.close()` while the Glass window is still not destroyed, possibly causing a second native close attempt. Therefore the comment:

> // ... This state is necessary to prevent the platform
> // window from being closed more than once.
 
And it is also needed when `stage.stageListener.closed()` bails out if `!window.isShowing()`: without setPlatformWindowClosed(), `platformWindow` won't be cleared (possible memory leak).

So I'd say we should keep the order of calls in the current proposal.

andy-goryachev-oracle (approved):
lgtm, thanks!

modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassWindowEventHandler.java
johanvos: > // ... This state is necessary to prevent the platform
> // window from being closed more than once.

What flow is expected when the window if being closed more than once? I assume it is always initiated from the native window system? In that case, it will happen on the FX Thread, and it will not be started until the previous notification listener code has finished.

kevinrushforth: Everything happens on the FX thread, so the concern isn't about multi-threading. Rather it is that we can get multiple close events, I think there are also multiple things that can trigger it (e.g., when an owner window is closed, all the child windows get closed).

In any case, Jose's fix doesn't change anything related to this. We already have defensive calls in place to avoid problems if close is called more than once.

The current fix is a good fix, regardless, given what `removeActiveWindow()` does. Also, I think it is the least intrusive fix for the problem. It would have been _way_ more risky to swap the order of the two existing calls (and, as Jose mentioned, it doesn't work to just eliminate one of them).

kevinrushforth (commented):
LGTM. The logic is sound and my local testing looks good.

I'll run a CI headful test build to be sure and approve if there are no problems.

kevinrushforth (approved):
All good. Testing is green on all platforms.

PR #2203, opened 2026-07-06
8388141: JavaFX.setOpenURIHandler() only receives event if application is already launched

pabulaner:
# Problem description

When You create a JavaFX native application with *jpackage* for macOS You will run into an issue when trying to handle URIs. When Your application isn’t running and You have it registered for handling certain types of URIs it won’t receive the first URI. 

## Here is a simple example: 

If You have a JavaFX application that handles URIs for example of the form *openurifx://your-content* and You have it installed with *jpackage* for macOS and now You try to handle an URI by running the following command: 

``` open openurifx://hello-world ``` 

The application will open, but it won’t receive the above URI. If You call the same command after the application is already running, everything will work fine and the application will receive the URI.

# Problem cause

The cause of the problem is that when You build a JavaFX application and want to handle URIs You will have to combine JavaFX and AWT as AWT provides the URI handling capabilities. So the startup process is as follows:

1. JavaFX is initialized
2. JavaFX forwards the URI event to AWT
3. AWT is initialized 
4. AWT is ready to handle URI events from JavaFX

The issue is that the URI event is delivered before AWT can handle it and therefore is simple dropped. That’s why later URIs will be received and are handled properly. 

# Fix

The fix is that JavaFX will wait until it receives a ready event from AWT so JavaFX knows that AWT is able to handle events now. After that it will forward all buffered URIs and therefore everything will be handled correctly. This means that the fix involves both the JFX and JDK repositories and needs to be integrated in both. 

**Note**: In the case that a user has only a fix for the JFX part the URIs won’t be delivered at all, but there won’t be any error. This means that JavaFX applications won’t be able to handle any URIs so it might be an idea to add an environment variable to tell JavaFX to not wait for an AWT ready event. Then the first URI event would be lost, but at least all the following events would be handled properly.

# Tests

To test the error and the fix You can clone this repository: https://github.com/pabulaner/openurifx and checkout the branch *first-url*. There You will find a README.md with further instructions.

PR inside the JDK: https://github.com/openjdk/jdk/pull/31791

---------
### Progress
- \[ \] Change requires CSR request [JDK-8391252](https://bugs.openjdk.org/browse/JDK-8391252) to be approved

### Issues
 * [JDK-8388141](https://bugs.openjdk.org/browse/JDK-8388141): JavaFX.setOpenURIHandler() only receives event if application is already launched (**Bug** - P4)
 * [JDK-8391252](https://bugs.openjdk.org/browse/JDK-8391252): JavaFX.setOpenURIHandler() only receives event if application is already launched (**CSR**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2203/head:pull/2203` \
`$ git checkout pull/2203`

`$ git checkout pull/2203` \
`$ git pull https://git.openjdk.org/jfx.git pull/2203/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2203`

`$ git pr show -t 2203`

</details>

<a href="https://git.openjdk.org/jfx/pull/2203.diff">https://git.openjdk.org/jfx/pull/2203.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2203#issuecomment-4894602155)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.h
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.h
@@ -53,6 +53,8 @@
     jboolean        jTaskBarApp;
     jlong           jshareContextPtr;
 
+    NSMutableArray<NSURL *> *bufferedURLs;
+
     // local and intra-app event monitoring
     //
     // id              localMonitor;
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.h
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.h
@@ -53,8 +53,6 @@
     jboolean        jTaskBarApp;
     jlong           jshareContextPtr;
 
-    NSMutableArray<NSURL *> *bufferedURLs;
-
     // local and intra-app event monitoring
     //
     // id              localMonitor;
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassApplication.m
@@ -77,18 +77,13 @@ static NSString* JavaRunLoopMode = @"AWTRunLoopMode";
 // don't deadlock.
 static NSArray<NSString*> *runLoopModes = nil;
 
-// Custom event that is emitted by AWT to let libraries like JavaFX
-// know that it is ready to receive embedded events. Until this event is
-// emitted libraries should buffer the already received events.
-static NSString* awtEmbeddedEventReady = @"AWTEmbeddedEventReady";
+// Specifies whether awt is initialized to receive embedded events.
+static bool awtInitialized = false;
 
 // Custom event that is provided by AWT to allow libraries like
 // JavaFX to forward native events to AWT even if AWT runs in
 // embedded mode.
 static NSString* awtEmbeddedEvent = @"AWTEmbeddedEvent";
-
-// Set to true once AWT has emitted the embedded ready event.
-static BOOL awtEmbeddedEventReadyReceived = false;
 
 #ifdef STATIC_BUILD
 jint JNICALL JNI_OnLoad_glass(JavaVM *vm, void *reserved)
@@ -197,13 +192,6 @@ static NSMutableArray<GlassRunnable*> *deferredRunnables = nil;
         {
             [GlassHelper SetGlassClassLoader:classLoader withEnv:env];
         }
-
-        self->bufferedURLs = [NSMutableArray array];
-
-        // register AWT embedded event ready listener
-        NSNotificationCenter *ctr = [NSNotificationCenter defaultCenter];
-        Class clz = [GlassApplication class];
-        [ctr addObserver:clz selector:@selector(_embeddedEventReady:) name:awtEmbeddedEventReady object:nil];
     }
     return self;
 }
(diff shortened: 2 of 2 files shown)


kevinrushforth:
@pabulaner You _must not_ use the same JBS bug ID for a fix that spans JavaFX and JDK. You will need to file a new JBS issue for JavaFX and change the title of this jfx PR to use that new bug ID.

Also, this will need careful coordination and some discussion before it proceeds. I won't even have time to look at it before the jfx27 fork (and since JDK has already forked for JDK 28, this isn't a candidate for jfx27 anyway).

kevinrushforth:
> @pabulaner You _must not_ use the same JBS bug ID for a fix that spans JavaFX and JDK. You will need to file a new JBS issue for JavaFX and change the title of this jfx PR to use that new bug ID.

I see that this is an old bug and was originally filed against the old javapackager (although is still relevant to running JavaFX with jpackage). I wonder if it would be better to use that original bug for the JavaFX half and file a new Bug or Enhancement request for AWT? Or maybe file a pair of new JBS issues -- one for JavaFX and one for AWT -- to better describe the problem.

Either way, you will need two bug ID: one for JavaFX and one for AWT.

crschnick:
Whenever I used both AWT and JavaFX together, I force AWT to initialize before JavaFX as several other things can break as well when doing it the other way around. Is initializing JavaFX before AWT even officially supported?

pabulaner:
I filed a new bug report with the help of @FlorianKirmaier and now the JDK and JFX PR have different JBS issues. 
In this case the JFX issue just references the already present JDK issue as this seemed to be the most simple solution.

prrace:
I don't see any harm in AWT sending such a message, although perhaps a CSR is needed.
The impact seems to be on the FX side, I don't know that I like that you essentially force the JDK-baseline for FX to be one that has this fix. Perhaps some other way to determine that AWT is loaded can be used. Or there should be a "cap" on how long it waits. I posted this comment in the JDK PR too.

pabulaner:
I have thought about this issue too. Maybe it would be a solution to add an environment variable to toggle this behavior on and off on the JavaFX side. Something like "DISABLE_AWT_EMBEDDED_READY_EVENT".

Or maybe You have some idea on how to achieve the same effect with a different implementation.

kevinrushforth:
> I don't see any harm in AWT sending such a message, although perhaps a CSR is needed.

Quite possibly this would need a CSR depending on how it is done.

> The impact seems to be on the FX side, I don't know that I like that you essentially force the JDK-baseline for FX to be one that has this fix.

Yeah, I don't like this aspect of the fix at all. As it stands, it will queue up these messages indefinitely if you run an FX that has this fix with JDK 27.

> Perhaps some other way to determine that AWT is loaded can be used. Or there should be a "cap" on how long it waits.

Or both. However it is implemented, I think we need to do the following:

1. A timeout and/or a max queue length so that it doesn't queue up an effectively unlimited number of these messages
2. Either implement a detection mechanism that doesn't require an AWT event (not sure whether it is feasible), or add a version check to disable this feature if the JDK version is < 28. This would be in addition to a system property or env var to disable it.

Thinking out loud... another alternative would be for JavaFX to initialize AWT via a call to `Toolkit.getToolkit()` when we receive this event. That would mean the fix could be done entirely in JavaFX and you might not need to worry about queuing the messages, presuming the call could be done synchronously. This would need discussion and a CSR.

pabulaner:
So I actually implemented Your feedback and the new implementation just requires a change on the JFX side. It uses a method invocation in the native part of the code to just call Toolkit.getDefaultToolkit() and therefore initializes AWT when receiving url events.

pabulaner:
If this approach is considered better than the old one, I will close the pull request on the JDK side.

kevinrushforth:
>> Thinking out loud... another alternative would be for JavaFX to initialize AWT via a call to `Toolkit.getToolkit()` when we receive this event. That would mean the fix could be done entirely in JavaFX and you might not need to worry about queuing the messages, presuming the call could be done synchronously. This would need discussion and a CSR.

...

> So I actually implemented Your feedback and the new implementation just requires a change on the JFX side. It uses a method invocation in the native part of the code to just call Toolkit.getDefaultToolkit() and therefore initializes AWT when receiving url events.

I see that you did a proof of concept of my alternative thought. Presuming that receiving the `NSApplicationDelegate application:openURLs:` event means that it is reasonable for the AWT toolkit to be initialized, something like this might work. I do like this better than the alternative of needing coordinated changes on the AWT side. Have you tested it in various scenarios to ensure that there are no problems initializing the toolkit from the AppKit thread? You will also need to test what happens when the AWT Toolkit is already initialized (although I can't think of any potential problems).

As I mentioned above, this will need some discussion on the mailing list and it will need a CSR, since it is a behavioral change.

/csr needed

FlorianKirmaier:
@kevinrushforth 
I've created a CSR based on that PR together with @pabulaner 
Here is the link: https://bugs.openjdk.org/browse/JDK-8391252
Any feedback on that - is there something we should change?

PR #2205, opened 2026-07-07
8387684: [PrismFontFile] Possibly unsafe double check locking

andy-goryachev-oracle:
Fixes double check locking by adding the `volatile` keyword.

During normal operation, the access seems to happen from two threads:
```
Thread[#29,QuantumRenderer-0,5,main]
Thread[#34,JavaFX Application Thread,5,main]
```

On one hand, this access is likely to be mutually exclusive, on the other hand, javafx does allow the nodes to be created in a background thread, and it might be possible to invoke, for example, `GlyphLayout.breakRuns()` from a background thread.

---------

---------
### Progress

### Issue
 * [JDK-8387684](https://bugs.openjdk.org/browse/JDK-8387684): [PrismFontFile] Possibly unsafe double check locking (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2205/head:pull/2205` \
`$ git checkout pull/2205`

`$ git checkout pull/2205` \
`$ git pull https://git.openjdk.org/jfx.git pull/2205/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2205`

`$ git pr show -t 2205`

</details>

<a href="https://git.openjdk.org/jfx/pull/2205.diff">https://git.openjdk.org/jfx/pull/2205.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2205#issuecomment-4908322107)
</details>


modules/javafx.graphics/src/main/java/com/sun/javafx/font/PrismFontFile.java
```diff
@@ -1436,7 +1436,7 @@ boolean hasGlyph(int gid) {
        }
    }
 
-   ColorGlyphStrike[] sbixStrikes = null;
+   private volatile ColorGlyphStrike[] sbixStrikes;
```
kevinrushforth, line 1439: I don't have time to look at this for jfx27, so I recommend leaving it for after the fork, but I see at one other related threading issue. Both should be fixed at the same time (else there is little point in changing anything). The `buildSbixStrikeTables()` could be called twice, since it isn't inside the inner `if (sbixStrikes == null)`. It won't be called concurrently, but will still do the initialization more than once which is, at best, unnecessary.

Minor: I prefer to leave the `= null` that you removed, since code is checking / relying on the initial value (yes I know it doesn't actually matter, but I find it more self-documenting with the explicit initialization).

andy-goryachev-oracle: Good point, the whole thing is funky.  I've restructured the code to make sure it is a proper double check locking and the `buildSbixStrikeTables()` is called only once.

For the null field, it's not necessary per JLS 4.12.5
https://docs.oracle.com/javase/specs/jls/se26/html/jls-4.html#jls-4.12.5 :

> For all reference types ([§4.3](https://docs.oracle.com/javase/specs/jls/se26/html/jls-4.html#jls-4.3)), the default value is null.

Besides, it annoys me when the debugger steps at each pointless assignments each time.

kevinrushforth (approved):
LGTM

andy-goryachev-oracle:
@azuev-java could you be the second reviewer?

arapte (approved):
This change improves correctness and does not alter behavior. +1

andy-goryachev-oracle:
thank you all for the review!

PR #2206, opened 2026-07-08
8331458: Replace qtdemux with MP4 demux based on Media Foundation

sashamatveev:
* Replaced `qtdemux` with a Media Foundation-based MP4 demuxer on Windows.
* The Media Foundation-based MP4 demuxer supports **pull mode only** due to the design of the Media Foundation APIs. For fragmented MP4 (fMP4) segments, each segment is demuxed independently because of API limitations. After processing each segment, the demuxer internally creates a new Source Reader.
* **HLSConnectionHolder:** Headers are now sent with every segment because the MP4 demuxer reloads the Source Reader for each segment. Headers are intentionally not cached to avoid unnecessary connections. Additionally, HLS streams with separate audio renditions now seek to the last segment when seeking to the end of the stream, matching the behavior of standard HLS streams.
* **JavaSource:** Removed the unused `update` flag and moved caps setting into a dedicated switch case.
* Added trace support for Media Foundation-based plugins (both the demuxer and decoder). The tracing infrastructure is still a work in progress and will be expanded as needed for debugging. It is disabled by default and can only be enabled at compile time.
* H.264 video from MP4 and fMP4 (HLS) is now decoded using Media Foundation instead of DirectShow. H.264 video in TS-based HLS streams continues to use DirectShow. This change avoids several DirectShow issues encountered with the new MP4 demuxer and reduces implementation complexity.
* Updated `MFWrapper` to properly support H.264 decoding. Unlike H.265, H.264 requires using the display aperture for the display resolution, does not signal frame availability, requires handling format changes, and must fully drain output before accepting new input samples.
* Modified the cache implementation to support pull mode for the HLS progress buffer.
* Added pull mode support to the HLS progress buffer.
* Changed `progressbuffer` to use a static source pad instead of a dynamic one, simplifying pipeline construction. The need for a dynamic source pad is unclear.
* Performed minor cleanup by removing dead and obsolete code.
* The progress buffer is now forced when `JavaSource` does not support pull mode (for example, with `jar` and `jrt` protocols). This provides a simple workaround until native pull mode support is implemented for these protocols. In particular, `jar` does not provide true random access at the Java layer anyway.

---------

---------
### Progress

### Issues
 * [JDK-8331458](https://bugs.openjdk.org/browse/JDK-8331458): Replace qtdemux with MP4 demux based on Media Foundation (**Enhancement** - P3)
 * [JDK-8305842](https://bugs.openjdk.org/browse/JDK-8305842): Video sometimes does not start when reinitializing in Windows 11 (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2206/head:pull/2206` \
`$ git checkout pull/2206`

`$ git checkout pull/2206` \
`$ git pull https://git.openjdk.org/jfx.git pull/2206/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2206`

`$ git pr show -t 2206`

</details>

<a href="https://git.openjdk.org/jfx/pull/2206.diff">https://git.openjdk.org/jfx/pull/2206.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2206#issuecomment-4910430110)
</details>


sashamatveev:
8305842: Replace qtdemux with MP4 demux based on Media Foundation [v4]
- Fixed mimetype issue on Linux. qtdemux only accepts `video/quicktime`, so we need to map `video/mp4` and `video/x-m4v` to `video/quicktime`.

sashamatveev:
8305842: Replace qtdemux with MP4 demux based on Media Foundation [v5]
- Changed on how HLS select media file index when variant playlist is switched. It was based on file name, but it is not reliable, since file names are not guarantee to match. Instead index is adjusted based on sequence number which should not drift more then +/- 1 when switch is done. Sequence number is like index of first segment file in live streams.
- Fixed issue when variant playlist was not detected as live even it was live. Playlist with variant playlists was detected correctly as live.
- Fixed reload time. Thread sleep is expecting milliseconds and we provided (seconds / 2) as milliseconds without conversion. As are result we where requesting playlist way too often.
- Redone mechanism on how we waiting for live segments to be added. Old approach had a bug with semaphore and if we added two segments in one run it will cause code to assume that we have segments when we did not. As a result EOS was signaled for live streams. This issue does not reproduce often, but might happen.
- Old playlist data will be cleared when we switching bitrate, otherwise we will attempt to play old segments which are gone long time.
- Fixed removing old segments not to remove init header for fMP4.
- Added additional tracing to mfdemux and mfwrapper.
- mfdemux will not push caps event if caps did not change. When event is pushed decoder will flush and cause video glitch.
- mfdemux will adjust segment event to first PTS. Needed for live streams, since their PTS does not start with 0.
- Increase queue buffers for HLS Live on Windows with mfdemux since we reload source reader for each segment to eliminate stream break ups.

sashamatveev:
8305842: Replace qtdemux with MP4 demux based on Media Foundation [v6]
- Fixed merge issues.

sashamatveev:
8305842: Replace qtdemux with MP4 demux based on Media Foundation [v7]
- Undo unrelated HLS fixes and moved them under https://github.com/openjdk/jfx/pull/2227

sashamatveev:
8305842: Replace qtdemux with MP4 demux based on Media Foundation [v8]
- mfdemux will convert `BYTE` segments to `TIME` to handle HLS Live change introduced in https://github.com/openjdk/jfx/pull/2227. With https://github.com/openjdk/jfx/pull/2227 `hlsprogressbuffer `will send `BYTE` segment, so qtdemux or any another demux will handle PTS at demux level.

modules/javafx.media/src/main/native/gstreamer/gstreamer-lite/projects/plugins/gstplugins-lite.c
```diff
@@ -41,8 +41,7 @@ gboolean lite_plugins_init (GstPlugin * plugin)
       !plugin_init_wavparse(plugin) ||
       !plugin_init_aiff(plugin) ||
       !plugin_init_app(plugin) ||
-      !plugin_init_audioparsers(plugin) ||
-      !plugin_init_qtdemux(plugin))
```
kevinrushforth, line 45: This used to be called on all platforms. I see you moved this initialization from a platform-independent call to a Linux-only call. That means it won't be called for macOS. Since you also removed qtdemux from the mac Makefile, I presume that this is intentional. Was qtdemux ever used on macOS prior to this?

modules/javafx.media/src/main/native/gstreamer/plugins/dshowwrapper/dshowwrapper.cpp
```diff
@@ -835,7 +835,6 @@ int dshowwrapper_deliver(GstBuffer *pBuffer, sUserData *pUserData)
                 g_print("JFXMEDIA OUTPUT AAC  -1\n");
         }
 #endif
-
```
kevinrushforth, line 838: This file has two whitespace-only changes (removal of two blank lines). I recommend reverting it.

modules/javafx.media/src/main/native/gstreamer/plugins/mfwrapper/mfwrapper.h
```diff
@@ -73,16 +74,21 @@ struct _GstMFWrapper
     gboolean is_eos_received;
     gboolean is_eos;
     gboolean is_decoder_initialized;
+    gboolean force_discontinuity;
```
kevinrushforth: This field is used, but never set. Is there a reason to add what looks like a duplicate of the existing `is_force_discontinuity` field?

modules/javafx.media/src/main/native/gstreamer/plugins/mfwrapper/mfwrapper.cpp
```diff
@@ -489,146 +577,36 @@ static void mfwrapper_print_output_media_formats(IMFTransform *pMFTrasnform, con
...
-    if (SUCCEEDED(hr))
-        hr = MFCreateMemoryBuffer(dwBufferSize, &pBuffer);
-
-    if (SUCCEEDED(hr))
-        hr = pBuffer->SetCurrentLength(dwBufferSize);
-
-    if (SUCCEEDED(hr))
-        hr = pBuffer->Lock(&pbBuffer, NULL, NULL);
+    HRESULT hr = mfwrapper_create_sample_from_gst_buffer(&pSample, buf,
+            decoder->force_discontinuity);
```
kevinrushforth: This field is never set.

modules/javafx.media/src/main/native/gstreamer/plugins/mfwrapper/mfdemux.cpp
```diff
@@ -0,0 +1,1728 @@
...
+        demux->pGSTMFByteStream = new (nothrow) CMFGSTByteStream((QWORD)data_length, demux->sink_pad, demux->is_hls);
+        if (demux->pGSTMFByteStream == NULL)
+            return FALSE;
+
+        hr = demux->pGSTMFByteStream->QueryInterface(IID_IMFByteStream, (void**)&demux->pIMFByteStream);
+        if (FAILED(hr) || demux->pIMFByteStream == NULL)
+            return FALSE;
+    }
+
+    hr = MFCreateSourceReaderFromByteStream(demux->pIMFByteStream, NULL, &demux->pSourceReader);
```
kevinrushforth, line 763: This is called by `mfdemux_loop` on a task thread. I think you need to initialize COM on this thread (with `CoInitializeEx`).

modules/javafx.media/src/main/native/gstreamer/plugins/mfwrapper/mfgstbytestream.cpp
```diff
@@ -0,0 +1,498 @@
...
+        return E_POINTER;
+
+    if (pResult != NULL)
+        pResult->SetStatus(m_readResult);
+
+    *pcbRead = m_cbBytesRead;
+
+    SafeRelease(&m_pAsyncResult);
+
+    return S_OK;
```
kevinrushforth: Shouldn't this return `m_readResult`?

modules/javafx.media/src/main/native/gstreamer/plugins/progressbuffer/posix/filecache.c
```diff
@@ -132,35 +132,53 @@ gint64 cache_read_buffer(Cache* cache, GstBuffer** buffer)
...
+
+    if (cache == NULL)
+        return GST_FLOW_ERROR;
+
+    guint64 bytes_available = cache_bytes_available(cache, start_position);
+    if (bytes_available == 0)
+        return GST_FLOW_FLUSHING;
+
+    if ((guint64)size < bytes_available)
+        size = bytes_available;
```
kevinrushforth: This test looks backwards (compare with `win32/filecache.c` which looks correct)

modules/javafx.media/src/main/native/gstreamer/plugins/mfwrapper/mfgstbytestream.cpp
```diff
@@ -0,0 +1,498 @@
...
+}
+
+HRESULT CMFGSTByteStream::BeginWrite(const BYTE *pb, ULONG cb, IMFAsyncCallback *pCallback, IUnknown *punkState)
+{
+    return E_NOTIMPL;
+}
+
+HRESULT CMFGSTByteStream::Close()
+{
+    // Nothing to close
```
kevinrushforth: Could there be a problem with a pending async read? If so, do you need to cancel it here?

modules/javafx.media/src/main/native/gstreamer/plugins/mfwrapper/mfdemux.cpp
```diff
@@ -0,0 +1,1728 @@
...
+            output_segment.position = GST_BUFFER_TIMESTAMP(pBuffer);
+        }
+
+        GstEvent *event = gst_event_new_segment(&output_segment);
+        gst_event_unref(demux->cached_segment_event);
+        demux->cached_segment_event = NULL;
+
+#if TRACE_ENABLE
+        TRACE(DEMUX_SRC_EVENTS, "GST_EVENT_SEGMENT start=%lld time=%lld position=%lld\n",
+                segment.start, segment.time, segment.position);
```
kevinrushforth: `segment` is undefined here.

kevinrushforth (commented):
I haven't done much testing, but left a few inline comments / questions.

modules/javafx.media/src/main/native/gstreamer/gstreamer-lite/projects/plugins/gstplugins-lite.c
sashamatveev: Yes, we used qtdemux upto jfx20 and usage was removed since JDK-8287822. For some reason I did not cleanup makefiles and plugin registration. On macOS since JDK-8287822 (jfx20) only aiff files are played via GStreamer and rest if via AVFoundation.

modules/javafx.media/src/main/native/gstreamer/plugins/mfwrapper/mfwrapper.h
sashamatveev: Merge conflict. We should use `is_force_discontinuity`. I will fix it by removing `force_discontinuity`.

modules/javafx.media/src/main/native/gstreamer/plugins/mfwrapper/mfwrapper.cpp
sashamatveev: Changed to `is_force_discontinuity`.

modules/javafx.media/src/main/native/gstreamer/plugins/mfwrapper/mfgstbytestream.cpp
sashamatveev: Yes, we need to cancel pending async read on Close. I will fix it.

modules/javafx.media/src/main/native/gstreamer/plugins/mfwrapper/mfgstbytestream.cpp
sashamatveev: Yes. I will fix it.

sashamatveev:
Replace qtdemux with MP4 demux based on Media Foundation [v9]
- Fixed latest review comments.

kevinrushforth (approved):
LGTM. The changes look good. Testing is green.

arapte (approved):

PR #2210, opened 2026-07-13
8388140: Significant CSS performance regression after JDK-8268657

Maran23:
Since the last [PR](https://github.com/openjdk/jfx/pull/2201) made the parent `styleHelper` creation more eager, creating and adding many nodes will be slower (initially).

The fix is actually something I already had in mind when I did the previous PR: Find the ancestor lazily.
Now, when we create a `CssStyleHelper` we will not yet set the `firstStyleableAncestor`. 
This will improve the performance to the level before while all tests still pass.

This makes sense because `CssStyleHelper` is created (top-down) but not immediately used. It can also be (re)created when a `Node` changes its `Parent` (or deferred from there).

Later, `transitionToState` is called (top-down) which will then need the `firstStyleableAncestor` (and this is also where we will compute it now). Since in the meantime the scene structure could change, it makes sense to do that as late as possible.

The days before I also investigated [JDK-8187955](https://bugs.openjdk.org/browse/JDK-8187955) and found a good improvement as well (no recursion needed, we can again use the preexisting loop). Since we now create the `firstStyleableAncestor` lazy, it makes sense to combine it, as otherwise the `isUserSetFont` need to be changed which I think is more risky.

Also did some digging about `isUserSetFont`, and was added/changed in this tickets:
- `isUserSetFont` code added: https://bugs.openjdk.org/browse/JDK-8123731
- `isUserSetFont` code changed: https://bugs.openjdk.org/browse/JDK-8093577

/issue add JDK-8187955

---------

---------
### Progress

### Integration blocker
&nbsp;⚠️ Title mismatch between PR and JBS for issue [JDK-8388140](https://bugs.openjdk.org/browse/JDK-8388140)

### Issues
 * [JDK-8388140](https://bugs.openjdk.org/browse/JDK-8388140): [BACKOUT] Looked-up color fails for -fx-background-color in JavaFX CSS file (**Bug** - P2) ⚠️ Title mismatch between PR and JBS.
 * [JDK-8187955](https://bugs.openjdk.org/browse/JDK-8187955): CssStyleHelper : recursion in isUserSetFont() can be avoided (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2210/head:pull/2210` \
`$ git checkout pull/2210`

`$ git checkout pull/2210` \
`$ git pull https://git.openjdk.org/jfx.git pull/2210/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2210`

`$ git pr show -t 2210`

</details>

<a href="https://git.openjdk.org/jfx/pull/2210.diff">https://git.openjdk.org/jfx/pull/2210.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2210#issuecomment-4962442978)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
@@ -87,9 +87,15 @@ final class CssStyleHelper {
      * Creates a new StyleHelper.
      */
     static CssStyleHelper createStyleHelper(final Node node) {
-        // need to know how far we are to root in order to init arrays.
-        // TODO: should we hang onto depth to avoid this nonsense later?
-        // TODO: is there some other way of knowing how far from the root a node is?
+        boolean userSetFont;
+        if (node.styleHelper == null) {
+            // Node styleHelper can not be reused later, because it does not exist.
+            // We can therefore safely set this true and ignore the property for the rest of this method.
+            userSetFont = true;
+        } else {
+            userSetFont = isUserSetFont(node);
+        }
+
         Node styleableAncestor = null;
         Styleable parent = node;
         int depth = 0;
@@ -97,13 +103,14 @@ final class CssStyleHelper {
             depth++;
             parent = parent.getStyleableParent();
 
-            if (styleableAncestor == null && parent instanceof Node parentNode && isStyleableAncestor(parentNode)) {
-                styleableAncestor = parentNode;
+            if (parent instanceof Node parentNode) {
+                if (styleableAncestor == null && parentNode.styleHelper != null) {
+                    styleableAncestor = parentNode;
+                }
+                if (!userSetFont) {
+                    userSetFont = isUserSetFont(parentNode);
+                }
             }
-        }
-
-        if (node.styleHelper != null) {
-            setFirstStyleableAncestor(node.styleHelper, styleableAncestor);
         }
 
         // The List<CacheEntry> should only contain entries for those
@@ -123,7 +130,7 @@ final class CssStyleHelper {
         //
         // reuse the existing styleHelper if possible.
         //
-        if ( canReuseStyleHelper(node, styleMap) ) {
+        if ( canReuseStyleHelper(node, styleMap, styleableAncestor) ) {
             //
             // JDK-8123731
             //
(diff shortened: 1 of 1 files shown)


andy-goryachev-oracle (commented):
I want to explore a couple of scenarios - this might take some time.

Maran23:
> I want to explore a couple of scenarios - this might take some time.

Thats fine! I also found one weird usecase I want to further investigate as well. I really hate the `CssStyleHelper`. No tests, no documentation (Just for fun I removed `isUserSetFont` and everything is still green). It is hard to fix something without breaking performance for another usecase it seems.

kevinrushforth:
I think this is a good direction, but I worry about the risk of functional regression. CSS is, as you know, notoriously prone to regressions. It's difficult to test and there are a lot of paths to consider. And any time we do lazy evaluation, it needs to be tested very carefully.

One concrete concern: The lazy eval approach will change when `CssStyleHelper parentHelper` is created. That could affect the validity of the cache key.

Given that we are T-47 hours to the JavaFX 27 RDP1 fork, I think we are better off backing out JDK-8268657 for 27, and then redoing it, along with _this_ fix in JavaFX 28 when we we will have the time to do more evaluation and testing to catch regressions before they happen, and more time to deal with any regressions that might be missed.

What do other reviewers think?

andy-goryachev-oracle:
I am planning to do more experiments today, but I had the same initial thought: it might be better to back out JDK-8268657 given the proximity to RDP1 and the risk of regression.

The bigger question is how can we ensure the CSS works as advertised?  Perhaps we ought to invest more time into writing a comprehensive test suite against the spec, which will be a monumental task.

Maran23:
> Given that we are T-47 hours to the JavaFX 27 RDP1 fork, I think we are better off backing out JDK-8268657 for 27, and then redoing it, along with _this_ fix in JavaFX 28 when we we will have the time to do more evaluation and testing to catch regressions before they happen, and more time to deal with any regressions that might be missed.
> 
> What do other reviewers think?

I thought about this today as well and also came to the same conclusion. So lets revert it.
In my profiling, this PR caused an increase in calls to `getCachedFont`, which I don't like but I haven't yet found a better way to deal with this without giving up the cached styleable ancestor.

kevinrushforth:
OK, I'll help with the bookkeeping:

As noted in the [JDK Developers Guide](https://openjdk.org/guide/#backing-out-a-change), I will mark the original issue as "Fix failed" and change [JDK-8388140](https://bugs.openjdk.org/browse/JDK-8388140) to be the backout issue.

You can close this PR and create a new "backout" pr with the changeset reverted. I'll also file a "REDO" bug for the original bug.

kevinrushforth:
I've made the changes in JBS as described above. Please close this PR. If you don't have time to create the backout PR today, I can do it.

Maran23:
Closing in favor of https://github.com/openjdk/jfx/pull/2212

PR #2213, opened 2026-07-15
8388304: Scene.setNodeOrientation triggers instant CSS but also doesn't update properly

FlorianKirmaier:
### Fix
The PR replaces applyCss() with reapplyCSS in the nodeOrientation code of Scene.
applyCss() reapplied styles eagerly - and didn't rebuild the "style maps" resulting in wrong css.
reapplyCSS() rematches correctly - and also defers it to the next pulse.

I've added a unit test to: Node_effectiveOrientation_Css_Test.

### Test improvements
Because the whole test class was disabled, I've also investigated which tests are working - and reenabled the working tests.
This PR also fixes 2 of the previously failing tests in Node_effectiveOrientation_Css_Test - which are now enabled.

Which are the following tests:
```
Node_effectiveOrientation_Css_Test.test_dir_pseudoClass_functions_on_scene_effective_orientation_not_node
Node_effectiveOrientation_Css_Test.test_SimpleSelector_dir_pseudoClass_with_scene_effective_orientation_rtl
```

---------

---------
### Progress

### Issue
 * [JDK-8388304](https://bugs.openjdk.org/browse/JDK-8388304): Scene.setNodeOrientation triggers instant CSS but also doesn't update properly (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2213/head:pull/2213` \
`$ git checkout pull/2213`

`$ git checkout pull/2213` \
`$ git pull https://git.openjdk.org/jfx.git pull/2213/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2213`

`$ git pr show -t 2213`

</details>

<a href="https://git.openjdk.org/jfx/pull/2213.diff">https://git.openjdk.org/jfx/pull/2213.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2213#issuecomment-5682938440)
</details>

### DIFF
--- a/modules/javafx.graphics/src/test/java/test/javafx/scene/Node_effectiveOrientation_Css_Test.java
+++ b/modules/javafx.graphics/src/test/java/test/javafx/scene/Node_effectiveOrientation_Css_Test.java
@@ -44,7 +44,6 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 /**
  * Test :dir functional pseudo-class
  */
-@Disabled("JDK-8234152")
 public class Node_effectiveOrientation_Css_Test {
 
     private Group root;
@@ -119,6 +118,7 @@ public class Node_effectiveOrientation_Css_Test {
         assertEquals(Color.web("#ff0000"), rect.getFill());
     }
 
+    @Disabled("JDK-8234152")
     @Test
     public void test_CompounSelector_dir_pseudoClass_on_parent_with_scene_effective_orientation_ltr() {
         Stylesheet stylesheet = new CssParser().parse(
@@ -139,6 +139,8 @@ public class Node_effectiveOrientation_Css_Test {
         assertEquals(Color.web("#00ff00"), rect.getFill());
     }
 
+
+    @Disabled("JDK-8234152")
     @Test
     public void test_CompoundSelector_dir_pseudoClass_on_parent_with_scene_effective_orientation_rtl() {
         Stylesheet stylesheet = new CssParser().parse(
@@ -160,6 +162,7 @@ public class Node_effectiveOrientation_Css_Test {
         assertEquals(Color.web("#ff0000"), rect.getFill());
     }
 
+    @Disabled("JDK-8234152")
     @Test
     public void test_CompounSelector_dir_pseudoClass_on_child_with_scene_effective_orientation_ltr() {
         Stylesheet stylesheet = new CssParser().parse(
@@ -180,6 +183,7 @@ public class Node_effectiveOrientation_Css_Test {
         assertEquals(Color.web("#00ff00"), rect.getFill());
     }
 
+    @Disabled("JDK-8234152")
     @Test
     public void test_CompoundSelector_dir_pseudoClass_on_child_with_scene_effective_orientation_rtl() {
         Stylesheet stylesheet = new CssParser().parse(
@@ -234,6 +238,37 @@ public class Node_effectiveOrientation_Css_Test {
         // :dir() pseudo-class functions on scene effective orientation, not node
         assertEquals(Color.web("#ff0000"), rect.getFill());
 
+    }
+
+    @Test
+    public void testCssUpdates() {
+        Group root = new Group();
+        Scene scene = new Scene(root);
+        Stage stage = new Stage();
+        stage.setScene(scene);
+        stage.show();
+
+        Stylesheet stylesheet = new CssParser().parse(
+                ".rect:dir(rtl) { -fx-fill: #ff0000; }" +
+                ".rect:dir(ltr) { -fx-fill: #00ff00; }" +
+                ".rect { -fx-fill: #0000ff; }"
+        );
+        StyleManager.getInstance().setDefaultUserAgentStylesheet(stylesheet);
+
+        Rectangle rect = new Rectangle();
+        rect.getStyleClass().add("rect");
+        root.getChildren().add(rect);
+
+        root.applyCss();
+        assertEquals(Color.web("#00ff00"), rect.getFill());
+
+        scene.setNodeOrientation(RIGHT_TO_LEFT);
+        root.applyCss();
+        assertEquals(Color.web("#ff0000"), rect.getFill());
+
+        scene.setNodeOrientation(LEFT_TO_RIGHT);
+        root.applyCss();
+        assertEquals(Color.web("#00ff00"), rect.getFill());
     }
 
 }
--- a/modules/javafx.graphics/src/test/java/test/javafx/scene/Node_effectiveOrientation_Css_Test.java
+++ b/modules/javafx.graphics/src/test/java/test/javafx/scene/Node_effectiveOrientation_Css_Test.java
@@ -120,7 +120,7 @@ public class Node_effectiveOrientation_Css_Test {
 
     @Disabled("JDK-8234152")
     @Test
-    public void test_CompounSelector_dir_pseudoClass_on_parent_with_scene_effective_orientation_ltr() {
+    public void test_CompoundSelector_dir_pseudoClass_on_parent_with_scene_effective_orientation_ltr() {
         Stylesheet stylesheet = new CssParser().parse(
                 ".root:dir(rtl) .rect { -fx-fill: #ff0000; }" +
                 ".root:dir(ltr) .rect { -fx-fill: #00ff00; }" +
@@ -164,7 +164,7 @@ public class Node_effectiveOrientation_Css_Test {
 
     @Disabled("JDK-8234152")
     @Test
-    public void test_CompounSelector_dir_pseudoClass_on_child_with_scene_effective_orientation_ltr() {
+    public void test_CompoundSelector_dir_pseudoClass_on_child_with_scene_effective_orientation_ltr() {
         Stylesheet stylesheet = new CssParser().parse(
                 ".root .rect:dir(rtl) { -fx-fill: #ff0000; }" +
                 ".root .rect:dir(ltr) { -fx-fill: #00ff00; }" +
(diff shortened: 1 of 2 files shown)


kevinrushforth:
This seems like a relatively safe fix on the surface, but we will not have time to adequately review it before the `jfx27` RDP1 fork. Since it touches CSS and since you have a somewhat related PR #2215 also out for review, it seems better to target this for `jfx28` anyway.

FlorianKirmaier:
I could imagine, that some parts of the code-base rely on the forced CSS computation.
But otherwise i have a good feeling that its correct.

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_effectiveOrientation_Css_Test.java
```diff
@@ -160,6 +162,7 @@ public void test_CompoundSelector_dir_pseudoClass_on_parent_with_scene_effective
         assertEquals(Color.web("#ff0000"), rect.getFill());
     }
 
+    @Disabled("JDK-8234152")
     @Test
     public void test_CompounSelector_dir_pseudoClass_on_child_with_scene_effective_orientation_ltr() {
```
Maran23: While here, can we fix the typo in `test_Compoun` -> `test_Compound` here and some other tests aswell

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_effectiveOrientation_Css_Test.java
```diff
@@ -236,4 +240,35 @@ public void test_dir_pseudoClass_functions_on_scene_effective_orientation_not_no
 
     }
 
+    @Test
+    public void testCssUpdates() {
```
Maran23: Maybe we should name it similar as what you wrote in your description - maybe `testChangeNodeOrientationWillReapplyCss` or `testChangeNodeOrientationWillBatchReapplyCss`

Maran23 (commented):
Fix looks good to me and makes sense. This is in-line with other operations. 

`applyCss` is usually only ever called for newly added children so that they are computed correctly immediately.
Although this is very fragile, as there is also:

https://github.com/openjdk/jfx/blob/17189589cb82a2cec6397e1e14024593f5ce30c3/modules/javafx.graphics/src/main/java/javafx/scene/Node.java#L9961-L9974

Which sometimes seems to not work correctly, probably because of nested parent.
Me and John were wondered about this at one point as well. But that is another story altogether.

So the `nodeOrientation` really is the odd one here. I don't see a reason we should call `applyCss` instead of the `reapplyCSS`.

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_effectiveOrientation_Css_Test.java
FlorianKirmaier: Fixed the 2 Compun - couldn't find another spelling error.

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_effectiveOrientation_Css_Test.java
FlorianKirmaier: It's now named testChangeNodeOrientationWillReapplyCss.

Maran23:
Looks good, tested on several applications and all good. Note that you need to fix the warnings / errors that are mentioned in the PR body.

This is unrelated to this PR (so should not block it):

I just had a look why the other tests fail. It is always the same reason which I think might not even a bug.

When I change:

```java
       Stylesheet stylesheet = new CssParser().parse(
                ".root:dir(rtl) .rect { -fx-fill: #ff0000; }" +
                ".root:dir(ltr) .rect { -fx-fill: #00ff00; }" +
                ".root .rect { -fx-fill: #0000ff; }"
        );
```

to:

```java
        Stylesheet stylesheet = new CssParser().parse(
                ".root .rect { -fx-fill: #0000ff; }" +
                ".root .rect:dir(rtl) { -fx-fill: #ff0000; }" +
                ".root .rect:dir(ltr) { -fx-fill: #00ff00; }"
        );
```

all tests are green. 

In CSS, the order matters. Although the CSS specification documents that the [specifity](https://www.w3.org/TR/selectors-4/#specificity-rules) should be higher with a pseudoclass (so in this case, the order should NOT matter). So JavaFX is not following the CSS specification here, but might be intended (although weird - I would say this is a bug).

cc @mstr2, @hjohn Do you have some thoughts on this?

hjohn:
> When I change:
> 
> ```java
>        Stylesheet stylesheet = new CssParser().parse(
>                 ".root:dir(rtl) .rect { -fx-fill: #ff0000; }" +
>                 ".root:dir(ltr) .rect { -fx-fill: #00ff00; }" +
>                 ".root .rect { -fx-fill: #0000ff; }"
>         );
> ```
> 
> to:
> 
> ```java
>         Stylesheet stylesheet = new CssParser().parse(
>                 ".root .rect { -fx-fill: #0000ff; }" +
>                 ".root .rect:dir(rtl) { -fx-fill: #ff0000; }" +
>                 ".root .rect:dir(ltr) { -fx-fill: #00ff00; }"
>         );
> ```
> 
> all tests are green.
> 
> In CSS, the order matters. Although the CSS specification documents that the [specifity](https://www.w3.org/TR/selectors-4/#specificity-rules) should be higher with a pseudoclass (so in this case, the order should NOT matter). So JavaFX is not following the CSS specification here, but might be intended (although weird - I would say this is a bug).

I don't know, specificity is encoded a bit weird (only 4 bits, and doesn't guard overflows) in `Match`, and is lacking specificity for types (but perhaps FX handles this elsewhere).  However, specifically for node orientation, there is some exception being made, so perhaps that has something to do with what you're seeing?

```
    Match(final Selector selector, Set<PseudoClass> pseudoClasses, int idCount, int styleClassCount) {
        Objects.requireNonNull(selector);
        Objects.requireNonNull(pseudoClasses);

        this.selector = selector;
        this.idCount = idCount;
        this.styleClassCount = styleClassCount;
        this.pseudoClasses = ImmutablePseudoClassSetsCache.of(pseudoClasses);
        int nPseudoClasses = pseudoClasses.size();
        if (selector instanceof SimpleSelector simple) {
            if (simple.getNodeOrientation() != INHERIT) {
                nPseudoClasses += 1;
            }
        }
        specificity = (idCount << 8) | (styleClassCount << 4) | nPseudoClasses;
    }
```

Maran23:
>         if (selector instanceof SimpleSelector simple) {
>             if (simple.getNodeOrientation() != INHERIT) {
>                 nPseudoClasses += 1;
>             }
>         }
>         specificity = (idCount << 8) | (styleClassCount << 4) | nPseudoClasses;

Good catch! Thats an odd check. Together with:

https://github.com/openjdk/jfx/blob/1e7e7d03e85f9aa3bca97f4e6ec64deaf9c0bdf6/modules/javafx.graphics/src/main/java/com/sun/javafx/css/SimpleSelector.java#L243-L246

this gives us the full picture. Unrelated to this PR, but very weird.

Maran23:
FYI @FlorianKirmaier this looks good to me. But you have still outstanding errors/warnings, see your PR body.
You probably should also merge in master again, while here.

FlorianKirmaier:
Merged with main,
removed trailing dot in title.
So everything should be ready.

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_effectiveOrientation_Css_Test.java
```diff
@@ -236,4 +240,35 @@ public void test_dir_pseudoClass_functions_on_scene_effective_orientation_not_no
 
     }
 
+    @Test
+    public void testChangeNodeOrientationWillReapplyCss() {
+        Group root = new Group();
+        Scene scene = new Scene(root);
+        Stage stage = new Stage();
```
andy-goryachev-oracle: this stage is leaking (never hidden etc.)
should probably use the class field instead `Node_effectiveOrientation_Css_Test.stage` ?

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_effectiveOrientation_Css_Test.java
```diff
@@ -236,4 +240,35 @@ public void test_dir_pseudoClass_functions_on_scene_effective_orientation_not_no
...
+
+        scene.setNodeOrientation(RIGHT_TO_LEFT);
+        root.applyCss();
+        assertEquals(Color.web("#ff0000"), rect.getFill());
+
+        scene.setNodeOrientation(LEFT_TO_RIGHT);
+        root.applyCss();
+        assertEquals(Color.web("#00ff00"), rect.getFill());
+    }
+
```
andy-goryachev-oracle: extra newline

andy-goryachev-oracle (commented):
Nice, localized fix.

Minor comment in one of the tests.  The test itself fails in master and runs with the fix on macOS 26.6.2.

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_effectiveOrientation_Css_Test.java
FlorianKirmaier: You are right. It nor properly reuses the part from @BeforeAll.

modules/javafx.graphics/src/test/java/test/javafx/scene/Node_effectiveOrientation_Css_Test.java
FlorianKirmaier: newline removed.

FlorianKirmaier:
Great - I've added the 2 changes.
@BeforeAll is now properly used in the test. And the newline is removed.

andy-goryachev-oracle (approved):
lgtm, thanks!

Maran23 (approved):
LGTM.
I linked the findings in regards to the disabled tests in [JDK-8234152](https://bugs.openjdk.org/browse/JDK-8234152).

PR #2217, opened 2026-07-16
8384483: Create implementation of NSAccessibilityTable protocol

azuev-java:
- To preserve functionality the separate native peers for tables and lists are created;
- The list peer for now reports its role as a table for backward compatibility, that can be changed later;
- Fixed the long existed bug where table reported incorrectly currently selected rows;
- Added accessibilityFocusedUIElement to AccessibleBase;
- Added functionality for array attribute count and array attribute values retrieval to AccessibleBase;

---------

---------
### Progress
- \[x\] Change requires CSR request [JDK-8390448](https://bugs.openjdk.org/browse/JDK-8390448) to be approved

### Issues
 * [JDK-8384483](https://bugs.openjdk.org/browse/JDK-8384483): Create implementation of NSAccessibilityTable protocol (**Enhancement** - P3)
 * [JDK-8390448](https://bugs.openjdk.org/browse/JDK-8390448): Create implementation of NSAccessibilityTable protocol (**CSR**)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2217/head:pull/2217` \
`$ git checkout pull/2217`

`$ git checkout pull/2217` \
`$ git pull https://git.openjdk.org/jfx.git pull/2217/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2217`

`$ git pr show -t 2217`

</details>

<a href="https://git.openjdk.org/jfx/pull/2217.diff">https://git.openjdk.org/jfx/pull/2217.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2217#issuecomment-4995226950)
</details>

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
@@ -1066,6 +1066,15 @@ public abstract class TableViewSkinBase<M, S, C extends Control, I extends Index
                  * look for column headers */
                 return getTableHeaderRow();
             }
+            case VISIBLE_ITEM_RANGE: {
+                I firstVisibleCell = flow.getFirstVisibleCellWithinViewport();
+                I lastVisibleCell = flow.getLastVisibleCellWithinViewport();
+                if (firstVisibleCell == null || lastVisibleCell == null) {
+                    return new int[] { 0, 0 };
+                }
+                int firstIndex = firstVisibleCell.getIndex();
+                return new int[] { firstIndex, lastVisibleCell.getIndex() - firstIndex + 1 };
+            }
             case VERTICAL_SCROLLBAR: return flow.getVbar();
             case HORIZONTAL_SCROLLBAR: return flow.getHbar();
             default: return super.queryAccessibleAttribute(attribute, parameters);
--- a/modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
@@ -810,6 +810,8 @@ public enum AccessibleAttribute {
      *     number of visible items </li>
      * <li>Parameters: </li>
      * </ul>
+     *
+     * @since 28
      */
     VISIBLE_ITEM_RANGE(int[].class),
 
--- a/modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
@@ -802,6 +802,18 @@ public enum AccessibleAttribute {
     VERTICAL_SCROLLBAR(Node.class),
 
     /**
+     * Returns the range of visible items in an indexed control.
+     * <ul>
+     * <li>Used by: ListView and TableView </li>
+     * <li>Needs notify: no </li>
+     * <li>Return Type: {@code int[]} containing the first item index and the
+     *     number of visible items </li>
+     * <li>Parameters: </li>
+     * </ul>
+     */
+    VISIBLE_ITEM_RANGE(int[].class),
+
+    /**
      * Returns true if node is visible, otherwise false.
      * <ul>
      * <li>Used by: Node and ContextMenu </li>
--- a/modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
@@ -807,7 +807,8 @@ public enum AccessibleAttribute {
      * <li>Used by: ListView and TableView </li>
      * <li>Needs notify: no </li>
      * <li>Return Type: {@code int[]} containing the first item index and the
-     *     number of visible items </li>
+     *     number of visible items. If no items are fully visible the return
+           value should be {0, 0} </li>
      * <li>Parameters: </li>
      * </ul>
      *
--- a/modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
@@ -808,7 +808,7 @@ public enum AccessibleAttribute {
      * <li>Needs notify: no </li>
      * <li>Return Type: {@code int[]} containing the first item index and the
      *     number of visible items. If no items are fully visible the return
-           value should be {0, 0} </li>
+     *     value should be {0, 0} </li>
      * <li>Parameters: </li>
      * </ul>
      *
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ListViewSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/ListViewSkin.java
@@ -376,6 +376,15 @@ public class ListViewSkin<T> extends VirtualContainerBase<ListView<T>, ListCell<
                 }
                 return null;
             }
+            case VISIBLE_ITEM_RANGE: {
+                ListCell<T> firstVisibleCell = flow.getFirstVisibleCellWithinViewport();
+                ListCell<T> lastVisibleCell = flow.getLastVisibleCellWithinViewport();
+                if (firstVisibleCell == null || lastVisibleCell == null) {
+                    return new int[] { 0, 0 };
+                }
+                int firstIndex = firstVisibleCell.getIndex();
+                return new int[] { firstIndex, lastVisibleCell.getIndex() - firstIndex + 1 };
+            }
             case SELECTED_ITEMS: {
                 MultipleSelectionModel<T> sm = getSkinnable().getSelectionModel();
                 if (sm == null) {
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkin.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkin.java
@@ -147,6 +147,10 @@ public class TableViewSkin<T> extends TableViewSkinBase<T, T, TableView<T>, Tabl
     /** {@inheritDoc} */
     @Override public Object queryAccessibleAttribute(AccessibleAttribute attribute, Object... parameters) {
         switch (attribute) {
+            case ROW_AT_INDEX: {
+                final int rowIndex = (Integer)parameters[0];
+                return rowIndex < 0 ? null : flow.getPrivateCell(rowIndex);
+            }
             case SELECTED_ITEMS: {
                 List<Node> selection = new ArrayList<>();
                 TableViewSelectionModel<T> sm = getSkinnable().getSelectionModel();
--- a/modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacAccessible.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacAccessible.java
@@ -121,6 +121,7 @@ final class MacAccessible extends Accessible {
         AXMenuItemCmdGlyph(ACCELERATOR, MacVariant::createNSNumberForInt),
         AXMenuItemCmdModifiers(ACCELERATOR, MacVariant::createNSNumberForInt),
         AXMenuItemMarkChar(SELECTED, MacVariant::createNSString),
+        AXVisibleItemRange(VISIBLE_ITEM_RANGE, MacVariant::createNSValueForRange),
         AXDateTimeComponents(null, MacVariant::createNSNumberForInt),
 
         // NSAccessibilityMenuRole
--- a/modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacAccessible.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacAccessible.java
@@ -1723,12 +1723,15 @@ final class MacAccessible extends Accessible {
                     if (variant != null && variant.longArray != null && variant.longArray.length > 0) {
                         long[] ids = variant.longArray;
                         ObservableList<Node> items = FXCollections.observableArrayList();
+                        AccessibleRole controlRole = (AccessibleRole)getAttribute(ROLE);
                         for (long id : ids) {
                             MacAccessible acc = GlassAccessibleToMacAccessible(id);
                             if (acc != null) {
                                 Integer index = (Integer)acc.getAttribute(INDEX);
                                 if (index != null) {
-                                    Node cell = (Node)getAttribute(ROW_AT_INDEX, index);
+                                    AccessibleAttribute itemAttribute = controlRole == AccessibleRole.LIST_VIEW
+                                            ? ITEM_AT_INDEX : ROW_AT_INDEX;
+                                    Node cell = (Node)getAttribute(itemAttribute, index);
                                     if (cell != null) {
                                         items.add(cell);
                                     }
--- a/modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacAccessible.java
+++ b/modules/javafx.graphics/src/main/java/com/sun/glass/ui/mac/MacAccessible.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2013, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -27,7 +27,9 @@ package com.sun.glass.ui.mac;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.function.Function;
 import javafx.collections.FXCollections;
 import javafx.collections.ObservableList;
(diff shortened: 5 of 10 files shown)


kevinrushforth:

andy-goryachev-oracle (changes requested):
- opened monkey tester
- selected table view page
- chose Items: 500,000 rows
- enabled voice over

and it locked up with the beach ball.  are you rendering all the rows somehow?  I see it iterating in `TableViewSkin:152` with rowIndex 143390 which isn't right.

MacAccessible:1152 iterates over the whole table (maxCount=500000).

edit: this looks like something this PR introduced, as it's not reproducible in master.

andy-goryachev-oracle:
Just FYI, there seems to be multiple accessibility issues in TableView (present in master, so they are separate, pre-existing issues):

1. command-A to select all, the accessibility rectangle goes outside of the visible area (i.e. when the table width exceeds the scroll pane view port).  tried to create a screenshot, but mac does not screenshot the accessibility decorations.

2. select a table cell, then click on the scroll bar with the mouse to show next/previous page.  the accessibility focus rectangle remains in the same place, now selecting some other cell, while the logical selection remains unchanged it seems.  using control-option-arrow keys, the viewport jumps to the previously selected cell.  also, the accessibility focus rectangle might go outside of the view port area if the table width is larger than the scroll pane viewport.

3. unexpectedly, using accessibility navigation control-option-arrow keys, the menu is selected on all key released (because of the option key release).

azuev-java:
> * opened monkey tester
> * selected table view page
> * chose Items: 500,000 rows
> * enabled voice over
> 
> and it locked up with the beach ball. are you rendering all the rows somehow? I see it iterating in `TableViewSkin:152` with rowIndex 143390 which isn't right.
> 
> MacAccessible:1152 iterates over the whole table (maxCount=500000).
> 
> edit: this looks like something this PR introduced, as it's not reproducible in master.

I have fixed it in the latest commit, should be working now.

azuev-java:
> Just FYI, there seems to be multiple accessibility issues in TableView (present in master, so they are separate, pre-existing issues):
> 
> 1. command-A to select all, the accessibility rectangle goes outside of the visible area (i.e. when the table width exceeds the scroll pane view port).  tried to create a screenshot, but mac does not screenshot the accessibility decorations.
> 2. select a table cell, then click on the scroll bar with the mouse to show next/previous page.  the accessibility focus rectangle remains in the same place, now selecting some other cell, while the logical selection remains unchanged it seems.  using control-option-arrow keys, the viewport jumps to the previously selected cell.  also, the accessibility focus rectangle might go outside of the view port area if the table width is larger than the scroll pane viewport.
> 3. unexpectedly, using accessibility navigation control-option-arrow keys, the menu is selected on all key released (because of the option key release).

I will try to address these issues with the implementation for the table rows, if i will not be able to - i will create a separate bug to tackle it later.

andy-goryachev-oracle:
How do I trigger VISIBLE_ITEM_RANGE code path?

andy-goryachev-oracle:
Noted a weird thing (though pre-existing in master): after a cell in the `ListView` is edited, it cannot receive accessibility focus rectangle and the content of it does not announce:

Screenshot 2026-08-11 at 10 13 16

also, the tooltip shows `edit to 'update' to commit the change` which is confusing and makes little sense.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
```diff
@@ -1066,6 +1066,15 @@ private boolean isCellFocused(int row) {
...
                 return getTableHeaderRow();
             }
+            case VISIBLE_ITEM_RANGE: {
+                I firstVisibleCell = flow.getFirstVisibleCellWithinViewport();
+                I lastVisibleCell = flow.getLastVisibleCellWithinViewport();
+                if (firstVisibleCell == null || lastVisibleCell == null) {
+                    return new int[] { 0, 0 };
+                }
+                int firstIndex = firstVisibleCell.getIndex();
+                return new int[] { firstIndex, lastVisibleCell.getIndex() - firstIndex + 1 };
```
andy-goryachev-oracle, line 1076: this code is not being hit when scrolling with the mouse.

should the change in the viewport origin result in VISIBLE_ITEM_RANGE being updated?

modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
```diff
@@ -801,6 +801,18 @@ public enum AccessibleAttribute {
...
+     * Returns the range of visible items in an indexed control.
+     * <ul>
+     * <li>Used by: ListView and TableView </li>
+     * <li>Needs notify: no </li>
+     * <li>Return Type: {@code int[]} containing the first item index and the
+     *     number of visible items </li>
+     * <li>Parameters: </li>
+     * </ul>
+     */
+    VISIBLE_ITEM_RANGE(int[].class),
```
andy-goryachev-oracle, line 817: 1. this is a new attribute, so this PR needs as CSR, correct?
2. would it make more sense to use a more descriptive record instead of int[]?  

minor: the values in this enum are not sorted alphabetically - should they?  do we care?

andy-goryachev-oracle (commented):
basically works identically to the master branch.

two questions:
1. accessibility cursor sort of breaks after editing a cell in both ListView and TableView
2. question about updating visible range in case of mouse scrolling

azuev-java:
> How do I trigger VISIBLE_ITEM_RANGE code path?

It is being called from the a11y code to query the currently visible rows range when VO requests all rows. So just create a large list or table (like 100000 rows), put focus on one of the first rows and start VoiceOver. Then using mouse scroll somewhere far away from the currently focussed row and click on any element within the row. The VO should re-request the list of the rows and it should correctly narrate the newly selected element. If this query fails the VO will narrate "Empty row" instead of the actual content of the selected row or cell.

modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
azuev-java: > 1. this is a new attribute, so this PR needs as CSR, correct?

I have no idea. If that would be a change in the public API i would definitely say yes, for the attribute - since we add it and not changing the meaning of the existing attribute - may be? Gray area for me honestly. @kevinrushforth Any comments?

> 2. would it make more sense to use a more descriptive record instead of int[]?

That would require more processing on the native side so i would just leave it as is. It is a technical attribute that is required to pass information from the skin level to the native a11y helper - it is not supposed to be available for any other purposes.

> minor: the values in this enum are not sorted alphabetically - should they? do we care?

Not really. Changing the existing code just to sort the enum - i would hate to do so, i prefer to keep the history cleaner, it's an old code, technical binding of the parameters to the native bound variables, nobody really looks at it except the compiler.

modules/javafx.controls/src/main/java/javafx/scene/control/skin/TableViewSkinBase.java
azuev-java: > this code is not being hit when scrolling with the mouse.
> 
> should the change in the viewport origin result in VISIBLE_ITEM_RANGE being updated?

Not until the VO requests the data. Which happens when it detects change in the selected row or when we enter/exit the table or list with VO selection shortcuts.

azuev-java:
> Noted a weird thing (though pre-existing in master): after a cell in the `ListView` is edited, it cannot receive accessibility focus rectangle and the content of it does not announce:
> 
> Screenshot 2026-08-11 at 10 13 16
> also, the tooltip shows `edit to 'update' to commit the change` which is confusing and makes little sense.

Thats... interesting. Seems like another thing i might need to address while working on the row a11y peer. Totally unexpected. Where on earth did it get data for that that tooltip?

modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
andy-goryachev-oracle: AccessibleAttribute is public API.  The new attribute also needs `@since 28`.
CSR is needed.

(it's fine to keep it an `int[]`)

andy-goryachev-oracle:
> `edit to 'update' to commit the change`

oh sorry, this tooltip comes from the monkey tester...

modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
azuev-java: > AccessibleAttribute is public API. The new attribute also needs `@since 28`. CSR is needed.

Oh well. Time to learn how to create CSR for JavaFX. Yaaaaay.... [/sarcasm off]

andy-goryachev-oracle:
Re: scrolling.

In theory, there is `Node.notifyAccessibleAttributeChanged()`.
It's typically called in an invalidated() method when a11y-related properties change.

andy-goryachev-oracle:
Another issue:

1. choose a large model (500,000 rows)
2. enable VoiceOver
3. select all (command-A)

locks up both `ListView` and `TableView`.  This is pre-existing, as it also locks up in master.

This is not a big issue for `ListView` since it is not intended for large models, but it may be a problem for the `TableView`.

modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
kevinrushforth: > > 1. this is a new attribute, so this PR needs as CSR, correct?
> 
> I have no idea. If that would be a change in the public API i would definitely say yes, for the attribute - since we add it and not changing the meaning of the existing attribute - may be? Gray area for me honestly. @kevinrushforth Any comments?

It's new API, so yes, a CSR is in order.

> > minor: the values in this enum are not sorted alphabetically - should they? do we care?
> 
> Not really. Changing the existing code just to sort the enum - i would hate to do so, i prefer to keep the history cleaner, it's an old code, technical binding of the parameters to the native bound variables, nobody really looks at it except the compiler.

Yeah, let's not do this. In addition to creating diff churn for no good reason, this would techincally be an incompatible change, since it would change the result of the "ordinal()" method. In practice I can't imagine that wouldn't matter, but ...

> > AccessibleAttribute is public API. The new attribute also needs `@since 28`. CSR is needed.
> 
> Oh well. Time to learn how to create CSR for JavaFX. Yaaaaay.... [/sarcasm off]

Same way as you would for the JDK. :)

kevinrushforth: > Yeah, let's not do this. In addition to creating diff churn for no good reason, this would techincally be an incompatible change, since it would change the result of the "ordinal()" method. In practice I can't imagine that wouldn't matter, but ...

Although, I see that we've inserted enums in the middle before, and this does too, so nm about the compatibility point.

Still, let's not have the churn.

modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
```diff
@@ -801,6 +801,20 @@ public enum AccessibleAttribute {
...
     VERTICAL_SCROLLBAR(Node.class),
 
+    /**
+     * Returns the range of visible items in an indexed control.
+     * <ul>
+     * <li>Used by: ListView and TableView </li>
+     * <li>Needs notify: no </li>
+     * <li>Return Type: {@code int[]} containing the first item index and the
+     *     number of visible items </li>
+     * <li>Parameters: </li>
```
andy-goryachev-oracle, line 812: Parameters: should describe the structure (see `BOUNDS_FOR_RANGE` for example)

modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
```diff
@@ -801,6 +801,20 @@ public enum AccessibleAttribute {
      */
     VERTICAL_SCROLLBAR(Node.class),
 
+    /**
+     * Returns the range of visible items in an indexed control.
+     * <ul>
+     * <li>Used by: ListView and TableView </li>
+     * <li>Needs notify: no </li>
+     * <li>Return Type: {@code int[]} containing the first item index and the
```
andy-goryachev-oracle, line 809: please specify how it encodes a no-range case.

it looks like the code returns [0,0], but is this right?
should it return an empty array or `null` instead?

andy-goryachev-oracle:
Created https://bugs.openjdk.org/browse/JDK-8390342 for VoiceOver lockup after select-all with large models.

modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
azuev-java: The {0, 0} is the valid range, we should not return empty array or null in any case.

modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
azuev-java: We do not pass any parameters to the VISIBLE_ITEM_RANGE property so the parameters list should be empty.

andy-goryachev-oracle: what do you mean?

https://github.com/openjdk/jfx/pull/2217/changes#diff-bdee521c329d386e86fab0d6870c31750481dddb7cb5ca6aa65854948ca649f8R379

we must document the parameters (and exceptional cases like no visible range).

modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
andy-goryachev-oracle: this exceptional case needs to be documented.

modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
kevinrushforth: The parameters here are _input parameters_. As Alex says, there aren't any (unlike the `BOUNDS_FOR_RANGE` attribute where there are).

modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
kevinrushforth: A brief note saying that `[0,0]` is returned in the case of null or empty might be helpful.

modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
andy-goryachev-oracle: I don't understand - it does not matter what kind of parameters they are, they need to be documented.  How am I supposed to provide VISIBLE_ITEM_RANGE in RichTextArea?

`ListViewSkin` L386 returns the `int[]` with `{ firstIndex, count }` which needs to be documented.  

Specifically,
- `int[] { firstIndex, count }` must be documented
- `int[] { 0, 0 }` case must be documented
- the fact that it is a non-null int[2] should probably be mentioned

kevinrushforth: I was referring to the first inline comment that you made above on this line, which is `* <li>Parameters: </li>`:

> Parameters: should describe the structure (see BOUNDS_FOR_RANGE for example)

There are no parameters.

> ListViewSkin L386 returns the int[] with { firstIndex, count } which needs to be documented.

So you are saying that the _return value_ (not parameters) needs to better documented. Quite possibly.

andy-goryachev-oracle: I am using terminology used in the `AccessibleAttribute` class.  For example, L91 specifies the "Parameters:", whatever that is.  Perhaps it should be renamed "Structure:" or something like that.

edit: Also, we should probably drop the "Parameters:" `<li>` when it's empty.

kevinrushforth: Parameters is consistently used in this class to mean _input_ parameters. The example from L91 (`BOUNDS_FOR_RANGE `) are input parameters. For this new attribute, as with many (most) others, there are none.

> Perhaps it should be renamed "Structure:" or something like that.

Why? Parameters is used elsewhere (e.g., for method parameters) to mean effectively the same thing. No need to invent a new term.

> Also, we should probably drop the "Parameters:" `<li>` when it's empty.

Maybe, but that's tangential to this PR.

andy-goryachev-oracle: I see - perhaps they should be called "Input Parameters:" then?

And in case of the `VISIBLE_ITEM_RANGE` we need to document the return value structure, because `Return Type: int[]` is insufficient.

kevinrushforth: > I see - perhaps they should be called "Input Parameters:" then?

I think it's fine as is, but file a follow-up issue if you want to clarify it a bit (maybe in the class header would be better).

> And in case of the VISIBLE_ITEM_RANGE we need to document the return value structure, because Return Type: int[] is insufficient.

Yes, that makes sense. Since `int[]` isn't a class with named parameters, it would be a good idea to describe it.

azuev-java: I've added some clarification about thye return value. If it is Ok i will amend the CSR and it will be ready for review.

andy-goryachev-oracle (approved):
thanks for making the changes!  lgtm

modules/javafx.graphics/src/main/java/javafx/scene/AccessibleAttribute.java
```diff
@@ -801,6 +801,21 @@ public enum AccessibleAttribute {
...
+     *     number of visible items. If no items are fully visible the return
+           value should be {0, 0} </li>
```
kevinrushforth: You are missing the `*` comment char.

azuev-java: Thanks, fixed.

andy-goryachev-oracle (approved):

arapte (approved):
Verified the behavior before and after, lgtm

PR #2219, opened 2026-07-21
8388508: [macOS] MenuItem graphics in the system menu bar is corrupted or disappears, or JVM crashes in CoreGraphics, under memory pressure

jperedadnr:
This PR creates a native owned copy of the pixel buffer passed to GlassPixels::getImage, to let AppKit start, at any later time, the rasterization of the images for the system menu bars with a native object that outlives the Java buffer. Therefore, if under certain conditions of high memory pressure, the latter is reclaimed and collected, the former is still available.

A manual test has been included. 

Since this is a memory pressure/timing issue, it is not deterministic, and depends on several conditions. Therefore, it needs to be run several times to spot the issue (before the proposed fix). After the fix, the issue should never happen again.

When running it from `tests/manual/systemmenu`, with:
```
$JAVA_HOME/bin/java @../../../build/run.args MacOSSystemMenuImageGraphicTest.java
```
it might not trigger the crash, but still the icons might be missing or garbled most of the runs.

However, with:

```
$JAVA_HOME/bin/javac @../../../build/run.args MacOSSystemMenuImageGraphicTest.java // remove --enable-native-access from run.args
$JAVA_HOME/bin/java @../../../build/run.args MacOSSystemMenuImageGraphicTest
```
the crash is more likely to happen in different runs.

Other options to increase the likelihood of failure is to use a low Xmx, or to increase the `COUNTER` value.

---------

---------
### Progress

### Issue
 * [JDK-8388508](https://bugs.openjdk.org/browse/JDK-8388508): [macOS] MenuItem graphics in the system menu bar is corrupted or disappears, or JVM crashes in CoreGraphics, under memory pressure (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2219/head:pull/2219` \
`$ git checkout pull/2219`

`$ git checkout pull/2219` \
`$ git pull https://git.openjdk.org/jfx.git pull/2219/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2219`

`$ git pr show -t 2219`

</details>

<a href="https://git.openjdk.org/jfx/pull/2219.diff">https://git.openjdk.org/jfx/pull/2219.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2219#issuecomment-5032946105)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/native-glass/mac/GlassPixels.m
+++ b/modules/javafx.graphics/src/main/native-glass/mac/GlassPixels.m
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2011, 2024, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2011, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -65,13 +65,16 @@ NSImage* getImage(u_int8_t* data, int jWidth, int jHeight, int jOffset) {
             size_t width = (size_t) jWidth;
             size_t height = (size_t) jHeight;
 
-            CGDataProviderRef provider = CGDataProviderCreateWithData(NULL,
-                                            data + jOffset, width * height * 4, NULL);
-            if (provider != NULL) {
-                cgImage = CGImageCreate(width, height, 8, 32, 4 * width, colorSpace,
-                        kCGImageAlphaPremultipliedFirst | kCGBitmapByteOrder32Little,
-                        provider, NULL, 1, kCGRenderingIntentDefault);
-                CGDataProviderRelease(provider);
+            CFDataRef cfData = CFDataCreate(NULL, data + jOffset, width * height * 4);
+            if (cfData != NULL) {
+                CGDataProviderRef provider = CGDataProviderCreateWithCFData(cfData);
+                if (provider != NULL) {
+                    cgImage = CGImageCreate(width, height, 8, 32, 4 * width, colorSpace,
+                            kCGImageAlphaPremultipliedFirst | kCGBitmapByteOrder32Little,
+                            provider, NULL, 1, kCGRenderingIntentDefault);
+                    CGDataProviderRelease(provider);
+                }
+                CFRelease(cfData);
             }
         }
         CGColorSpaceRelease(colorSpace);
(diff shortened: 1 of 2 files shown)


beldenfox:
I could easily reproduce the bug in the master branch and this PR fixes it. Linux and Windows also copy the data (it's deferred on Windows but it does get copied). The reference counting in this PR looks correct and overall the code looks good.

I think there's an existing bug which leaks one of these NSImages. In GlassPasteboard.m look for jPixelsAttachData (currently on line 270). The NSImage created there is not released and since it's created using `alloc` it won't be autoreleased either. The two other calls to jPixelsAttachData are releasing their NSImages.

jperedadnr:
Thanks @beldenfox. 

As you say, GlassPasteboard creates a `NSImage` that is allocated in `GlassPixels::getImage` (which precisely gets modified in this PR), but never released (while GlassCursor and GlassMenu do release it properly). While that is a clear leak in GlassPasteboard, I fail to see this as related to the same bug/JBS ticket.

While the fix for this leak would be really small (just adding `[image release];` to GlassPasteboard), I'm not sure it should be part of this PR? Shouldn't we file a separated JBS ticket for it instead?

beldenfox:
> While the fix for this leak would be really small (just adding `[image release];` to GlassPasteboard), I'm not sure it should be part of this PR? Shouldn't we file a separated JBS ticket for it instead?

Yeah, we should probably treat this as a separate bug even though the fix is so small.

kevinrushforth:
> > While the fix for this leak would be really small (just adding `[image release];` to GlassPasteboard), I'm not sure it should be part of this PR? Shouldn't we file a separated JBS ticket for it instead?
> 
> Yeah, we should probably treat this as a separate bug even though the fix is so small.

Yes, please. Can one of you file a new bug for this?

jperedadnr:
I can do it

jperedadnr:
I've filed https://bugs.openjdk.org/browse/JDK-8388786 for the `NSImage` leak.

beldenfox (approved):
Looks good to me.

kevinrushforth (approved):
Looks good. The provided test program does not render the menus correctly and sometime crashes. All good with the fix.

kevinrushforth:
@jperedadnr Sorry for the delay. This fixes a serious bug, and it would be suitable to consider for jfx27 during RPD2 if you want to make a fix request.

jperedadnr:
Thanks, @kevinrushforth . 
I've added the `jfx27-fix-request` label.

kevinrushforth:
> I've added the `jfx27-fix-request` label.

Approved.

PR #2225, opened 2026-07-27
8388277: [REDO] Looked-up color fails for -fx-background-color in JavaFX CSS file

Maran23:
Another much better try to fix the issue.
I recommend to read: https://github.com/openjdk/jfx/pull/2201 first. All tests from there are included. 
I added some new ones that succeed before and after, a first step for more CSS tests as discussed in: https://github.com/openjdk/jfx/pull/2218#issuecomment-5094495306

My new idea is now the following constraint, which I think is also a much better approach:
- A `CssStyleHelper` always has a correct `firstStyleableAncestor`. We can at any time trust and rely on it.
- We will also reuse the existing loop for the `isUserSetFont` check to improve the performance a bit

Implementation:
- We now save a flag to exactly know in which CSS state the `Node` is.
- We will collect all `Node`s in the scene tree once and then reuse the list when we need to process a stale `styleHelper` from a parent.
  - This will make sure we do no process the parent again and again when we have some stale `styleHelper`s in the chain 
- Not every `Node` has a `styleHelper` - it is only created when needed, so we can not attach the flag in there.

This fixes the issue while a deep (optionally unstyled) scene graph has no performance penality.
The approach is similar than my previous PR, but more smart. And with the set constraint mentioned above.

---

I do think we can improve the `CssStyleHelper` more. But for another day. 
Maybe at one point, with more tests and when all requirements are clear, we can find a way without `CssHelperState` and without creating an empty `CssStyleHelper` just to hold trigger states (because of that, we need to check `styleHelper.cacheContainer != null` a lot of times).

/issue add JDK-8187955

---------

---------
### Progress

### Issues
 * [JDK-8388277](https://bugs.openjdk.org/browse/JDK-8388277): [REDO] Looked-up color fails for -fx-background-color in JavaFX CSS file (**Bug** - P4)
 * [JDK-8187955](https://bugs.openjdk.org/browse/JDK-8187955): CssStyleHelper : recursion in isUserSetFont() can be avoided (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2225/head:pull/2225` \
`$ git checkout pull/2225`

`$ git checkout pull/2225` \
`$ git pull https://git.openjdk.org/jfx.git pull/2225/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2225`

`$ git pr show -t 2225`

</details>

<a href="https://git.openjdk.org/jfx/pull/2225.diff">https://git.openjdk.org/jfx/pull/2225.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2225#issuecomment-5096796561)
</details>


andy-goryachev-oracle (commented):
please resolve the merge conflict.

subhramit:
Our project [JabRef](https://github.com/JabRef/jabref) is currently dependent on this fix for a [critical performance bug report](https://github.com/JabRef/jabref/issues/16343).
Would really appreciate if it could be expedited.

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
```diff
@@ -204,93 +222,77 @@ private static void updateParentTriggerStates(Styleable styleable, int depth, Ps
...
-    //
-    private static void setTrue(WritableValue<Boolean> booleanProperty) {
-        if (booleanProperty != null) booleanProperty.setValue(true);
+    /**
+     * Whether {@code parentNode} can act as a styleable ancestor, i.e. whether it has a helper that
+     * can actually contribute styles.
+     */
+    private static boolean isStyleableAncestor(Node parentNode) {
+        if (parentNode.cssHelperState == Node.CssHelperState.STALE) {
+            parentNode.cssHelperState = Node.CssHelperState.RESOLVED_EARLY;
```
andy-goryachev-oracle, line 285: there might be a problem here: this code sets `RESOLVED_EARLY` before creating the `styleHelper`.  createStyleHelper() can invoke application listeners that might, for example, add new nodes and force `applyCss()` immediately.  The re-entrance descendants see `RESOLVED_EARLY` and trust the old, stale, `styleHelper`.

would it make sense to add another state, or perhaps move the state into `styleHelper` ?

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
```diff
@@ -87,15 +89,34 @@ private CssStyleHelper() {
...
+        }
 
-        // need to know how far we are to root in order to init arrays.
-        // TODO: should we hang onto depth to avoid this nonsense later?
-        // TODO: is there some other way of knowing how far from the root a node is?
+        Node styleableAncestor = null;
         Styleable parent = node;
         int depth = 0;
-        while(parent != null) {
+        while (parent != null) {
```
andy-goryachev-oracle, line 104: there might be potential quadratic execution time here.

imagine a long chain of nodes
```
root
  + node1
      + node2
         ...
            + nodeN
```

- start by adding a child to nodeN, this makes the entire chain dirty.
- change style class in all the nodes in the chain
- add a second child to nodeN

adding the second child recursively rebuilds every stale ancestor, walking all the way to the root.

kevinrushforth:
> Our project [JabRef](https://github.com/JabRef/jabref) is currently dependent on this fix for a [critical performance bug report](https://github.com/JabRef/jabref/issues/16343). Would really appreciate if it could be expedited.

GIven that the earlier version caused a regression, and that this touches an area of the code that is particularly tricky, it will need sufficient time to review and test it. It will then go into jfx28 and is unlikely to be backported, given the risk, at least not right away.

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
```diff
@@ -87,15 +89,34 @@ private CssStyleHelper() {
...
+                    styleableAncestor = parentNode;
+                }
+                if (!userSetFont) {
+                    userSetFont = isUserSetFont(parentNode);
+                }
+            }
+        }
+
+        if (node.styleHelper != null) {
+            setFirstStyleableAncestor(node.styleHelper, styleableAncestor);
```
andy-goryachev-oracle, line 119: this changes the existing helper's ancestor unconditionally.  the helper might contain stale styles (especially during style or hierarchy updates from within the listeners)

Maran23: not sure I follow: This just sets the ancestor, the same way we did before

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
Maran23: this will only rebuild the style helper that needs it. And only the first ancestor. I can't see how this could be a problem - do you have a unit test in mind? 
I tested several scenarios and could not spot any problem. Note that this is a very rare case that usually only happens for the scenarios I implemented as tests

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
Maran23: Moving to `styleHelper` was my very first idea but many nodes will have not have one (mostly layout containers). So it is not possible to do that. I can change the order of the statements though.

A layout container may have no style helper but because a parent style changed and now affects it, it needs one (and we need to set the state accordingly).

In the future, I would really like to revisit this. I hope this can be simplified, but there are more changes needed. Especially the creation of an empty `CssStyleHelper` just to hold parent states is weird. Maybe, on the way we will find an even better way, when other things are refactored/improved first.

modules/javafx.graphics/src/test/java/test/javafx/scene/CssStyleHelperTest.java
```diff
@@ -1248,6 +1252,371 @@ public void mediaQueryRemovalShouldNotInterruptTransitionsDuringReset() {
...
+        for (int i = 0; i < depth; i++) {
+            StackPane pane = new StackPane();
+            current.getChildren().add(pane);
+            chain.add(pane);
+
+            current = pane;
+        }
+        return chain;
+    }
+
```
andy-goryachev-oracle, line 1619: toDataUrl: similar to https://github.com/openjdk/jfx/pull/2236#discussion_r3751856321

perhaps we could create a Utils class with this method.  no need to replicate utility code all over the place.

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
andy-goryachev-oracle: ok, so here is the test that passes in master and fails spectacularly with this PR:

```
    @Test
    void checkQuadraticPerformace() {
        scene.getStylesheets().add(toDataURL(
            """
            .old {
               -fx-padding: 1.0;
            }
            .new {
               -fx-padding: 99.0;
            }
            """));

        AtomicInteger counter = new AtomicInteger();

        class TPane extends Pane {
            public TPane(String style) {
                getStyleClass().add("style");
            }

            @Override
            public Styleable getStyleableParent() {
                counter.incrementAndGet();
                return super.getStyleableParent();
            }
        }

        int number = 16;
        ArrayList<TPane> chain = new ArrayList<>();

        TPane top = new TPane("old");
        chain.add(top);

        TPane p = top;
        for (int i = 1; i < number; i++) {
            TPane ch = new TPane("old");
            p.getChildren().add(ch);
            chain.add(ch);
            p = ch;
        }

        scene.setRoot(top);
        top.applyCss();

        counter.set(0);
        // mark the chain dirty
        p.getChildren().add(new Pane());
        int baseline = counter.get();

        for (TPane pane : chain) {
            pane.getStyleClass().setAll("new");
        }

        counter.set(0);
        // should not result in quadratic performance
        p.getChildren().add(new Pane());
        int newCount = counter.get();

        assertTrue(newCount <= baseline * 4, () -> {
            return MessageFormat.format("Baseline={0}, observed={1}", baseline, newCount);
        });
    }
```

modules/javafx.graphics/src/test/java/test/javafx/scene/CssStyleHelperTest.java
Maran23: yes, will do after JUnit 6!

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
Maran23: Thanks, very helpful! I thought about it for a while, and did come up with another implementation that I think is a bit more simple and the other hand more optimized for that scenario.

Now I will collect all nodes just once and reuse that list to initialize all stale `styleHelper`. That should be much more optimized and I actually think it improved the code as well a little bit.
I first thought this may get much more complex, but it isn't IMO.

Maran23: (Also added your test as regression test!)

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
```diff
@@ -173,124 +203,111 @@ static CssStyleHelper createStyleHelper(final Node node) {
...
 
             helper.cacheContainer.cssSetProperties.putAll(remainingProperties);
         }
 
         return helper;
     }
 
-    private static void updateParentTriggerStates(Styleable styleable, int depth, PseudoClassState[] triggerStates) {
+    private static Styleable[] styleablePath(Styleable styleable) {
+        Styleable[] path = new Styleable[8];
```
Maran23: I wonder if a `List` makes more sense. So open for opinions, I used an array as we also do that with e.g. `triggerStates` and to squeeze out a bit of performance. But not sure if it is worth it.

Maran23:
Okay after testing more, I found one more case where I got a CSS warning.
Added a test and fix. It is a very similar test where the hierarchy is just a bit different to trigger the problem.

So thanks @andy-goryachev-oracle for you test case, that really helped for this particular usecase.

I'm quite happy with the solution and also the performance.

Very interesting: I just retested if [JDK-8291853](https://bugs.openjdk.org/browse/JDK-8291853) is now fixed as well (so with the PR reverted), and indeed, it is fixed now. 
Back then I already had the suspicion that is has something to do with the CSS processing. Which makes me even more confident that I hopefully found the correct solution for all problems and all performance cases.

So we could also revert the fix as a follow-up, as now we found the underlying probem / root cause.
The PR as reference: https://git.openjdk.org/jfx/pull/947.

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
andy-goryachev-oracle: A `List` might be a better option as it avoids growing the array manually, enables streams

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
```diff
@@ -84,18 +87,47 @@ private CssStyleHelper() {
...
     static CssStyleHelper createStyleHelper(final Node node) {
+        Styleable[] path = styleablePath(node);
+
+        // We first recreate the style helper that are stale.
+        // This usually only happens when a child changes the scene tree while its CSS is processed.
+        for (int index = path.length - 1; index > 0; index--) {
+            if (path[index] instanceof Node ancestor && ancestor.cssHelperStale) {
+                ancestor.cssHelperStale = false;
+                ancestor.cssHelperResolvedEarly = true;
+                ancestor.styleHelper = createStyleHelper(ancestor, path, index);
```
andy-goryachev-oracle: `cssHelperStale` is reset before the the new helper is created, which will cause problems if the new helper resets css properties, invoking listeners.  if such a listener styles another descendant, or adds another descendant, the CSS code will see `cssHelperStale=false` but observe the old helper instance.

we might want to consider either a "resolving" state or detect this condition and retry.

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
```diff
@@ -84,18 +87,47 @@ private CssStyleHelper() {
...
 
     /**
-     * Creates a new StyleHelper.
+     * Creates a new {@link CssStyleHelper} for the {@link Node},
+     * or null if it does not need one due to no style matching it.
+     *
+     * @return the {@link CssStyleHelper} or null
      */
     static CssStyleHelper createStyleHelper(final Node node) {
+        Styleable[] path = styleablePath(node);
```
andy-goryachev-oracle: similar problem with a stale `path`: a listener might make this path invalid.
while `findMatchingStyles()` uses the new hierarchy, while CacheContainer, `updateParentTriggerStates()`, `firstStyleableAncestor` still use old path.

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
```diff
@@ -84,18 +87,47 @@ private CssStyleHelper() {
...
+
+        // We first recreate the style helper that are stale.
+        // This usually only happens when a child changes the scene tree while its CSS is processed.
+        for (int index = path.length - 1; index > 0; index--) {
```
andy-goryachev-oracle: so my ai buddy suggests the process remains quadratic:
while `getStyleableParent()` is used only once, its overloads are called for each stale ancestor.  each scans the path to find a user-set font, following to the root, and a new `CacheContainer` scans again.

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
Maran23: Not sure I understand. Why would it do that? We just create it and move on. No other method is called.

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
Maran23: I fail to understand that. All methods in the `createStyleHelper` will use the parent chain we created once. How could we still end up being quadratic? We will go through the aprent chain, yes, but we did that before as well

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
Maran23: Will have a look tomorrow about this, especially regarding the cssHelperStale / cssProperties path

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
andy-goryachev-oracle: the use case is this:
- layout() is called at the top of the hierarchy
- a child is added somewhere in layoutChildren(), which invokes reapplyCSS(), which calls createStyleHelper()
- createStyleHelper() triggers another property change, and let's say that property has a listener that adds another child
- since isPerformingLayout() is true, the child's css is processed immediately

the issue is that while `ancestor.cssHelperStale = false`, the `ancestor.styleHelper` is still old, causing wrong css to be applied (which represents a regression)

Maran23: Okay then I probably understood you right, that's exactly what I've been checking all day (inspired by your other comment). Wrote some failing tests already but need to do more testing.

I might make a PR to simplify the `triggerStates` before, since that currently makes the logic worse to understand (and may enable some optimizations as well).

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
andy-goryachev-oracle: ok, this might be a contrived scenario, but it does illustrate the problem.  the test passes in master and fails with this PR:

```java
    @Test
    void testQuadraticPerformace() {
        scene.getStylesheets().add(toDataURL(
            """
            .old { -fx-padding: 11; }
            .new { -fx-padding: 22; }
            """));

        int depth = 50;
        AtomicInteger count = new AtomicInteger();

        class TPane extends Pane {
            private final SimpleStyleableObjectProperty<Font> font;
            private final List<CssMetaData<? extends Styleable, ?>> metadata;

            public TPane(CssMetaData<TPane, Font> fontMetadata) {
                font = new SimpleStyleableObjectProperty<>(fontMetadata, this, "font", Font.getDefault());
                // we should have a utility for this
                ArrayList<CssMetaData<? extends Styleable, ?>> md = new ArrayList<>(Pane.getClassCssMetaData());
                md.add(fontMetadata);
                this.metadata = List.copyOf(md);
                getStyleClass().add("old");
            }

            @Override
            public List<CssMetaData<? extends Styleable, ?>> getCssMetaData() {
                return metadata;
            }
        }

        CssMetaData<TPane, Font> fontMD = new FontCssMetaData<TPane>("-fx-font", Font.getDefault()) {
            @Override
            public StyleableProperty<Font> getStyleableProperty(TPane p) {
                count.incrementAndGet();
                return p.font;
            }
            
            @Override
            public boolean isSettable(TPane p) {
                return !p.font.isBound();
            }
        };

        ArrayList<TPane> panes = new ArrayList<>(depth);
        TPane top = new TPane(fontMD);
        panes.add(top);

        TPane p = top;
        for (int i = 1; i < depth; i++) {
            TPane ch = new TPane(fontMD);
            p.getChildren().add(ch);
            panes.add(ch);
            p = ch;
        }

        scene.setRoot(top);
        top.applyCss();

        p.getChildren().add(new Pane());

        for (TPane pane : panes) {
            pane.getStyleClass().setAll("new");
        }

        count.set(0);

        p.getChildren().add(new Pane());

        int ct = count.get();
        assertTrue(ct <= depth * 4, "Expected<=" + (depth*4) + " observed=" + ct + " for depth=" + depth);
    }
```

andy-goryachev-oracle:
One of the emerging theme here is the use of multiple data points where atomicity is required (the helper and its state).  I wonder if we could avoid that by let's say merging the state into the helper somehow?

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
Maran23: Just verified, works with my changes.

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
Maran23: pushed

Maran23:
> One of the emerging theme here is the use of multiple data points where atomicity is required (the helper and its state). I wonder if we could avoid that by let's say merging the state into the helper somehow?

Yes that is what I need to come up with. I moved `cssResetInProgress` out to `Node`.

And I will have a look to move `triggerStates` to `Node` as well, as this will enable us to not create unneeded nearly empty `CssStyleHelper` and will improve the code quite a bit, as currently we need to check if a `styleHelper` has a null `cacheContainer` so that we know it is an empty `CssStyleHelper` "shell" just to hold the `triggerStates`. This will probably make the code here just a tiny bit easier as well.

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
```diff
@@ -84,20 +86,61 @@ private CssStyleHelper() {
...
+    private static CssStyleHelper updateStyleHelper(Node node, List<Styleable> path, int index,
+                                                    Node styleableAncestor, boolean ancestorUserSetFont) {
+        final CssStyleHelper currentHelper = node.styleHelper;
+        final boolean userSetFont = currentHelper == null || ancestorUserSetFont || isUserSetFont(node);
+
+        if (currentHelper != null) {
+            setFirstStyleableAncestor(currentHelper, styleableAncestor);
+        }
+
+        node.cssHelperStale = false;
```
andy-goryachev-oracle: I think we are hitting the same problem as before: the `cssHelperStale` flag is cleared while the old helper is still there.

This is a problem because code that runs after that might execute application logic that re-enters the CSS processing and sees the stale helper.

I don't know what the solution might be - either a separate `RESOLVING` state, or moving this flag to the helper itself, or discarding the helper, or what?

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
```diff
@@ -84,20 +86,61 @@ private CssStyleHelper() {
...
+                recreatedAncestor = true;
+            }
+
+            userSetFont = userSetFont || isUserSetFont(ancestor);
+            if (isStyleableAncestor(ancestor)) {
+                styleableAncestor = ancestor;
+            }
         }
 
+        if (recreatedAncestor && !isPathValid(path)) {
```
andy-goryachev-oracle: another scenario: reparenting an ancestor under a different parent.  in this case, the remaining loop rebuilds the ancestor using old path and clears the `cssHelperStale` flag in L142.  the recursive call follows the new path, but because it sees the state no longer stale it skips the ancestor.

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
```diff
@@ -84,20 +86,61 @@ private CssStyleHelper() {
...
+        boolean recreatedAncestor = false;
+
+        for (int index = path.size() - 1; index > 0; index--) {
+            if (!(path.get(index) instanceof Node ancestor)) {
+                continue;
+            }
+
+            if (ancestor.cssHelperStale) {
+                ancestor.cssHelperResolvedEarly = true;
+                updateStyleHelper(ancestor, path, index, styleableAncestor, userSetFont);
```
andy-goryachev-oracle: it looks like we still have quadratic work here: for each stale ancestor, we allocate `triggerStates` array at L153 and when the style map changed, CacheContainer scans another ancestor at L419

CssStyleHelperTest counts only `getStyleableParent()` but track neither these scans nor allocations.

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
```diff
@@ -107,16 +151,16 @@ static CssStyleHelper createStyleHelper(final Node node) {
         // are gotten. By comparing the actual pseudo-class state to the
         // pseudo-class states that apply, a CacheEntry can be created or
         // fetched using only those pseudoclasses that matter.
-        final PseudoClassState[] triggerStates = new PseudoClassState[depth];
+        final PseudoClassState[] triggerStates = new PseudoClassState[path.size() - index];
 
         final StyleMap styleMap =
                 StyleManager.getInstance().findMatchingStyles(node, node.getSubScene(), triggerStates);
```
andy-goryachev-oracle, line 157: I think we still have a problem: `findMatchingStyles` can re-enter CSS and see the ancestors as stale, so it will try to re-build them, possibly in a loop.

a solution might be to introduce `RESOLVING` state which prevents that (similar to `Parent.performingLayout` flag)

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
```diff
@@ -84,18 +86,60 @@ private CssStyleHelper() {
...
+                continue;
+            }
+
+            if (ancestor.cssHelperStale) {
+                ancestor.cssHelperResolvedEarly = true;
+                boolean propertiesReset = updateStyleHelper(ancestor, path, index, styleableAncestor, userSetFont);
+
+                // Listeners running while the old helper reset properties may have changed the hierarchy.
+                // The remaining stale ancestors must not be rebuilt for an outdated path, so start over.
+                if (propertiesReset && !isPathValid(path)) {
```
andy-goryachev-oracle, line 111: a listener can make the newly installed helper stale again (for example, by changing the ancestor style).  we probably need to restart when the updated node becomes stale again (in addition to checking the path).

modules/javafx.graphics/src/main/java/javafx/scene/CssStyleHelper.java
```diff
@@ -374,10 +411,10 @@ private final static class CacheContainer {
         private CacheContainer(
                 Node node,
                 final StyleMap styleMap,
-                int depth) {
-
+                List<Styleable> path,
+                int startIndex) {
             int ctr = 0;
-            int[] smapIds = new int[depth];
+            int[] smapIds = new int[path.size() - startIndex];
```
andy-goryachev-oracle, line 417: we still seem to have quadratic allocation: each stale ancestor allocates a `triggerStates` array in L154, and each changed helper allocates and scans another suffix in its cache key.

the existing test tracks neither allocations nor scans.

PR #2227, opened 2026-07-29
8389211: Fix HLS Live issues found during testing of JDK-8331458 and various live streams

sashamatveev:
- Changed on how HLS select media file index when variant playlist is switched. It was based on file name, but it is not reliable, since file names are not guaranteed to match. Instead, index is adjusted based on sequence number which should not drift more than +/- 1 when switch is done. Sequence number is like index of first segment file in live streams.
Fixed issue when variant playlist was not detected as live even it was live. Playlist with variant playlists was detected correctly as live.
- Fixed reload time. Thread sleep is expecting milliseconds, and we provided (seconds / 2) as milliseconds without conversion. As have resulted we were requesting playlist way too often.
- Redone mechanism on how we are waiting for live segments to be added. Old approach had a bug with semaphore and if we added two segments in one run it will cause code to assume that we have segments when we did not. As a result, EOS was signaled for live streams. This issue does not reproduce often but might happen.
- Old playlist data will be cleared when we are switching bitrate, otherwise we will attempt to play old segments which are gone long time.
- Fixed start time for live streams by sending `BYTE` segment vs `TIME`, so `qtdemux` will handle initial start time. Live streams will not start with 0 PTS as on demand.
- `javasource` will report `GST_CLOCK_TIME_NONE` for unknown duration instead of large negative number.
- `#EXT-X-MEDIA` streams will be seek to last segment similar to streams without `#EXT-X-MEDIA` when seek to the end.

---------

---------
### Progress

### Issue
 * [JDK-8389211](https://bugs.openjdk.org/browse/JDK-8389211): Fix HLS Live issues found during testing of JDK-8331458 and various live streams (**Bug** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2227/head:pull/2227` \
`$ git checkout pull/2227`

`$ git checkout pull/2227` \
`$ git pull https://git.openjdk.org/jfx.git pull/2227/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2227`

`$ git pr show -t 2227`

</details>

<a href="https://git.openjdk.org/jfx/pull/2227.diff">https://git.openjdk.org/jfx/pull/2227.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2227#issuecomment-5111596784)
</details>

### DIFF
--- a/modules/javafx.media/src/main/java/com/sun/media/jfxmedia/locator/HLSConnectionHolder.java
+++ b/modules/javafx.media/src/main/java/com/sun/media/jfxmedia/locator/HLSConnectionHolder.java
@@ -28,7 +28,9 @@ import com.sun.media.jfxmedia.MediaError;
 import com.sun.media.jfxmedia.MediaException;
 import com.sun.media.jfxmediaimpl.MediaUtils;
 import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
@@ -45,7 +47,7 @@ import java.util.List;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.Semaphore;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Stream;
 
 final class HLSConnectionHolder extends ConnectionHolder {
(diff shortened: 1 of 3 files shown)


sashamatveev:
8389211: Fix HLS Live issues found during testing of JDK-8331458 and various live streams [v2]
- Playlist switch will done using index and not a sequence number. Using sequence number is not correct.
- Adjusted reload time. Reload time will be target duration after playlist was updated or target duration / 2 if playlist did not change.
- If live changed to VOD playlist we will stop reloading.

PR #2234, opened 2026-07-31
8389515: Memory leak in dshowwrapper - decoded buffers delivered by CSink are never unmapped

wkia:
This is the fix for "JDK-8389515: Memory leak in dshowwrapper - decoded buffers delivered by CSink are never unmapped"

To prevent leaks, the `CSample` class was refactored with new `SetGstBuffer` and `TakeGstBuffer` methods that automatically track and manage the mapping state of the underlying `GstBuffer`. The `Allocator` and `Sink` are updated to use these new methods, ensuring buffers are always properly unmapped and released when their ownership is transferred.

The reproducer (`NativeMemoryLeakReproducer.java`) log with the fix applied:
```
+ d:/work/jdk/jdk25/bin/java.exe @jfx/build/run.args NativeMemoryLeakReproducer.java demo.mp4
Java:   25 (OpenJDK 64-Bit Server VM)
JavaFX: 28-internal (runtime 28-internal+0-2026-08-04-080015)
451.6 MiB
459.3 MiB
452.4 MiB
463.1 MiB
465.6 MiB
469.9 MiB
474.9 MiB
494.9 MiB
498.5 MiB
496.0 MiB
509.4 MiB
```

---------

---------
### Progress

### Issue
 * [JDK-8389515](https://bugs.openjdk.org/browse/JDK-8389515): Memory leak in dshowwrapper - decoded buffers delivered by CSink are never unmapped (**Bug** - P3)

### Reviewers
 * [Alexander Matveev](https://openjdk.org/census#almatvee) (@sashamatveev - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2234/head:pull/2234` \
`$ git checkout pull/2234`

`$ git checkout pull/2234` \
`$ git pull https://git.openjdk.org/jfx.git pull/2234/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2234`

`$ git pr show -t 2234`

</details>

<a href="https://git.openjdk.org/jfx/pull/2234.diff">https://git.openjdk.org/jfx/pull/2234.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2234#issuecomment-5146394276)
</details>

### DIFF
--- a/modules/javafx.media/src/main/native/gstreamer/plugins/dshowwrapper/Allocator.cpp
+++ b/modules/javafx.media/src/main/native/gstreamer/plugins/dshowwrapper/Allocator.cpp
@@ -55,20 +55,29 @@ HRESULT CAllocator::GetBuffer(IMediaSample **ppBuffer, REFERENCE_TIME *pStartTim
     if (FAILED(hr))
         return hr;
 
+    pSample = (CSample*)*ppBuffer;
+
     if (m_pBuffer == NULL)
     {
         if (GetGstBuffer != NULL)
             GetGstBuffer(&m_pBuffer, m_lSize, &m_UserData);
 
         if (m_pBuffer == NULL)
+        {
+            pSample->Release();
+            *ppBuffer = NULL;
             return E_FAIL;
+        }
     }
 
-    pSample = (CSample*)*ppBuffer;
     hr = pSample->SetGstBuffer(m_pBuffer);
     m_pBuffer = NULL;
     if (FAILED(hr))
+    {
+        pSample->Release();
+        *ppBuffer = NULL;
         return hr;
+    }
 
     return S_OK;
 }
--- a/modules/javafx.media/src/main/native/gstreamer/plugins/dshowwrapper/Allocator.cpp
+++ b/modules/javafx.media/src/main/native/gstreamer/plugins/dshowwrapper/Allocator.cpp
@@ -69,6 +69,7 @@ HRESULT CAllocator::GetBuffer(IMediaSample **ppBuffer, REFERENCE_TIME *pStartTim
         return hr;
 
     pSample->m_pGstBuffer = m_pBuffer;
+    pSample->m_pMappedGstBuffer = m_pBuffer;
     pSample->m_GstMapInfo = m_MapInfo;
 
     hr = pSample->SetPointer(m_MapInfo.data, m_MapInfo.size);
@@ -86,7 +87,11 @@ HRESULT CAllocator::ReleaseBuffer(IMediaSample *pBuffer)
     CSample *pSample = (CSample*)pBuffer;
     if (ReleaseSample != NULL)
     {
-        gst_buffer_unmap(pSample->m_pGstBuffer, &pSample->m_GstMapInfo);
+        if (pSample->m_pMappedGstBuffer != NULL)
+        {
+            gst_buffer_unmap(pSample->m_pMappedGstBuffer, &pSample->m_GstMapInfo);
+            pSample->m_pMappedGstBuffer = NULL;
+        }
         ReleaseSample(pSample->m_pGstBuffer, &m_UserData);
         pSample->m_pGstBuffer = NULL;
     }
--- a/modules/javafx.media/src/main/native/gstreamer/plugins/dshowwrapper/Allocator.cpp
+++ b/modules/javafx.media/src/main/native/gstreamer/plugins/dshowwrapper/Allocator.cpp
@@ -65,14 +65,7 @@ HRESULT CAllocator::GetBuffer(IMediaSample **ppBuffer, REFERENCE_TIME *pStartTim
     }
 
     pSample = (CSample*)*ppBuffer;
-    if (!gst_buffer_map(m_pBuffer, &m_MapInfo, GST_MAP_WRITE))
-        return hr;
-
-    pSample->m_pGstBuffer = m_pBuffer;
-    pSample->m_pMappedGstBuffer = m_pBuffer;
-    pSample->m_GstMapInfo = m_MapInfo;
-
-    hr = pSample->SetPointer(m_MapInfo.data, m_MapInfo.size);
+    hr = pSample->SetGstBuffer(m_pBuffer);
     m_pBuffer = NULL;
     if (FAILED(hr))
         return hr;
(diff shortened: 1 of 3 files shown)


kevinrushforth:
Reviewers: @sashamatveev @kevinrushforth

kevinrushforth:
The Windows GHA failure is an unrelated intermittent failure, so can be ignored (we don't compile native media code anyway).

modules/javafx.media/src/main/native/gstreamer/plugins/dshowwrapper/Allocator.cpp
```diff
@@ -86,7 +87,11 @@ HRESULT CAllocator::ReleaseBuffer(IMediaSample *pBuffer)
     CSample *pSample = (CSample*)pBuffer;
     if (ReleaseSample != NULL)
     {
-        gst_buffer_unmap(pSample->m_pGstBuffer, &pSample->m_GstMapInfo);
+        if (pSample->m_pMappedGstBuffer != NULL)
+        {
+            gst_buffer_unmap(pSample->m_pMappedGstBuffer, &pSample->m_GstMapInfo);
```
kevinrushforth: `m_pMappedGstBuffer` is a copy of the pointer that was used to initialize `m_pGstBuffer`, but there is nothing that will ensure ownership of the buffer. If there is any code path where the buffer is freed prior to `CAllocator::ReleaseBuffer` being called -- and it looks like there might be -- this could lead to a use after free.

I'd like @sashamatveev to take a closer look.

kevinrushforth (commented):
It will fix the leak, but I'm not sure it is safe from a lifecycle point of view.

modules/javafx.media/src/main/native/gstreamer/plugins/dshowwrapper/Allocator.cpp
wkia: @kevinrushforth 
> If there is any code path where the buffer is freed prior to CAllocator::ReleaseBuffer ... this could lead to a use after free.

I would only like to point out that the proposed change is intended to _upmap_ the buffer properly, and repeats the original behavior that has been disrupted after disabling the macro. 

OTOH we could move unmapping closer to ownership change in `CSink::DoRenderSampleInternal()`.

sashamatveev: I think we should fix it correctly. Original implementation was not right to unmap memory on GstBuffer which might be released already.

My suggestion is to add functions below in CSample:
`HRESULT SetGstBuffer(GstBuffer *pGstBuffer);`
`GstBuffer *TakeGstBuffer();`

Add `m_bGstBufferMapped` flag as well to CSample.

Allocator will call `SetGstBuffer` and `SetGstBuffer` will map it. `CSink::DoRenderSampleInternal()` will call `TakeGstBuffer` and this function will unmap it and set `m_pGstBuffer` to NULL. Call `TakeGstBuffer` in `CAllocator::ReleaseBuffer` to release it in case if it was not delivered. `m_pGstBuffer` and `m_GstMapInfo` should be private.

wkia: I've made the requested changes, and updated the test results.

modules/javafx.media/src/main/native/gstreamer/plugins/dshowwrapper/Allocator.cpp
```diff
@@ -65,13 +65,7 @@ HRESULT CAllocator::GetBuffer(IMediaSample **ppBuffer, REFERENCE_TIME *pStartTim
...
-        return hr;
-
-    pSample->m_pGstBuffer = m_pBuffer;
-    pSample->m_GstMapInfo = m_MapInfo;
-
-    hr = pSample->SetPointer(m_MapInfo.data, m_MapInfo.size);
+    hr = pSample->SetGstBuffer(m_pBuffer);
     m_pBuffer = NULL;
     if (FAILED(hr))
         return hr;
```
kevinrushforth, line 79: If `SetGstBuffer` fails, I think you need to release the sample and clear `*ppBuffer` to avoid leaking sample buffers, since the caller doesn't in the case of a reported failure. Maybe something like this?

```
    if (FAILED(hr)) {
        pSample->Release();
        *ppBuffer = null;
        return hr;
    }
```

kevinrushforth (commented):
Yes, this is better. I left one inline comment about another cleanup case that I think is needed. @sashamatveev can confirm.

modules/javafx.media/src/main/native/gstreamer/plugins/dshowwrapper/Allocator.cpp
wkia: @kevinrushforth 
If so, then I think we need to release the sample after a failed `GetGstBuffer` as well, am I right?
Like
```
        if (GetGstBuffer != NULL)
            GetGstBuffer(&m_pBuffer, m_lSize, &m_UserData);

        if (m_pBuffer == NULL)
        {
            pSample->Release();
            *ppBuffer = NULL;
            return E_FAIL;
        }
```

kevinrushforth: Yes, I think you are right, except that `pSample` hasn't yet been assigned (so the assignment of `pSample` would need to be moved up).

wkia: Pushed the changes

kevinrushforth (commented):
Updated fix looks good. I'll run a few tests then approve. One suggested bit of cleanup:. `m_MapInfo` is now unused and can be removed.

wkia:
@kevinrushforth 
> Updated fix looks good. I'll run a few tests then approve. One suggested bit of cleanup:. `m_MapInfo` is now unused and can be removed.

I forgot about that, sorry. Done

kevinrushforth (approved):
LGTM

sashamatveev:
Looks good. Please update copyright year to 2026 for modified files.

wkia:
@sashamatveev 
> update copyright year to 2026

Done

kevinrushforth (approved):

sashamatveev (approved):
Looks good.

kevinrushforth:
Thanks for fixing this. This would be a good candidate to backport to `jfx27`.

/sponsor

kevinrushforth:
> This would be a good candidate to backport to `jfx27`.

I should add "...`jfx27` if the backport PR can be done today". JavaFX 27 RDP2 starts tomorrow, after which, this can go into jfx27u (for 27.0.1) instead.

PR #525, opened 2021-05-31
8263095: Provide a way for a custom control to indicate that its userAgentStyleSheet has changed

palexdev:
This change allows custom control to change their style dynamically
When the user-agent stylesheet changes the property automatically calls `NodeHelper.reapplyCSS(Region.this);` to recompute the CSS for the node and its children.
To make this work the StyleManager class must be fixed too.
The line `regionUserAgentStylesheet = weakRegionUserAgentStylesheetMap.computeIfAbsent((Region)region, Region::getUserAgentStylesheet);
` is problematic because if the region is already present in the map but the user-agent stylesheet changed, the value of `regionUserAgentStylesheet` is not updated and this leads to controls having the wrong style.
To fix this issue I added this check:
```
if (((Region) region).getUserAgentStylesheet() != null && !((Region) region).getUserAgentStylesheet().equals(regionUserAgentStylesheet)) {
    weakRegionUserAgentStylesheetMap.put((Region) region, ((Region) region).getUserAgentStylesheet());
    regionUserAgentStylesheet = ((Region) region).getUserAgentStylesheet();
}

```

---------
### Progress
- [x] Change must not contain extraneous whitespace
- [x] Commit message must refer to an issue
- [ ] Change requires a CSR request matching fixVersion (No fixVersion in .jcheck/conf) to be approved (needs to be created)
- [ ] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Issue
 * [JDK-8263095](https://bugs.openjdk.org/browse/JDK-8263095): Provide a way for a custom control to indicate that its userAgentStyleSheet has changed

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx pull/525/head:pull/525` \
`$ git checkout pull/525`

Update a local copy of the PR: \
`$ git checkout pull/525` \
`$ git pull https://git.openjdk.org/jfx pull/525/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 525`

View PR using the GUI difftool: \
`$ git pr show -t 525`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
<a href="https://git.openjdk.org/jfx/pull/525.diff">https://git.openjdk.org/jfx/pull/525.diff</a>

</details>

PR #2236, opened 2026-07-31
8389092: PopupWindow.show() unconditionally overwrites user-set scene stylesheets with the owner's stylesheets

Maran23:
`PopupWindow` will always overwrite its `Scene` (user agent) stylesheets when an owner was set and `show` is called.
Code like this:

```java
        final Popup popup = new Popup();
        popup.getScene().setUserAgentStylesheet(popupUserAgentStylesheet);
        popup.getScene().getStylesheets().add(popupStylesheet);
```

will do nothing, because your added stylesheets will be later overwritten when `show` is called. 
Andy and I were already wondering about this behavior two years ago: https://github.com/openjdk/jfx/pull/1394#discussion_r1617620949

---

I can't see any reason why we should do that. Instead, this PR will only add the stylesheets of the owner if they do not exist already. 
Additionally, we will not overwrite the user agent stylesheet if it was already set.

Added tests for all combinations I can think of. This PR also fixes NPEs that can happen when the owner window has no 'root window'.

There is already a test for the `Cursor` behavior, but there was none that verifies that the `Cursor` is not overwritten, so added one as well.

https://github.com/openjdk/jfx/blob/05a7b6d0db5e799395da27ab43a93d4337001e11/modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java#L479-L481

Now, we will never overwrite anything that the developer set (before `show`ing). And as we can see above, this was already done this way with the `Cursor`.

---------

---------
### Progress

### Issue
 * [JDK-8389092](https://bugs.openjdk.org/browse/JDK-8389092): PopupWindow.show() unconditionally overwrites user-set scene stylesheets with the owner's stylesheets (**Bug** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2236/head:pull/2236` \
`$ git checkout pull/2236`

`$ git checkout pull/2236` \
`$ git pull https://git.openjdk.org/jfx.git pull/2236/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2236`

`$ git pr show -t 2236`

</details>

<a href="https://git.openjdk.org/jfx/pull/2236.diff">https://git.openjdk.org/jfx/pull/2236.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2236#issuecomment-5147015356)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
+++ b/modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2010, 2025, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2010, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -458,31 +458,34 @@ public abstract class PopupWindow extends Window {
     }
 
     private void showImpl(final Window owner) {
+        Window rootWindow = getRootWindow(owner);
+        if (rootWindow == null) {
+            return;
+        }
+
         // Update the owner field
         this.ownerWindow.set(owner);
         if (owner instanceof PopupWindow) {
             ((PopupWindow)owner).children.add(this);
         }
+
         // PopupWindow should disappear when owner node is not visible
-        if (owner != null) {
-            owner.showingProperty().addListener(weakOwnerNodeListener);
-        }
+        owner.showingProperty().addListener(weakOwnerNodeListener);
 
         final Scene sceneValue = getScene();
         SceneHelper.parentEffectiveOrientationInvalidated(sceneValue);
 
-        // JDK-8116444
-        applyStylesheetFromOwner(owner);
-
-        final Scene ownerScene = getRootWindow(owner).getScene();
+        final Scene ownerScene = rootWindow.getScene();
         if (ownerScene != null) {
+            copyStylesheetFromOwnerScene(ownerScene);
+
             if (sceneValue.getCursor() == null) {
                 sceneValue.setCursor(ownerScene.getCursor());
             }
         }
 
         // It is required that the root window exist and be visible to show the popup.
-        if (getRootWindow(owner).isShowing()) {
+        if (rootWindow.isShowing()) {
             // We do show() first so that the width and height of the
             // popup window are initialized. This way the x,y location of the
             // popup calculated below uses the right width and height values for
@@ -498,14 +501,34 @@ public abstract class PopupWindow extends Window {
      * @param owner the owner {@link Window}
      */
     void applyStylesheetFromOwner(Window owner) {
-        Scene scene = getScene();
-        final Scene ownerScene = getRootWindow(owner).getScene();
-        if (ownerScene != null) {
-            if (ownerScene.getUserAgentStylesheet() != null) {
-                scene.setUserAgentStylesheet(ownerScene.getUserAgentStylesheet());
-            }
-            scene.getStylesheets().setAll(ownerScene.getStylesheets());
+        // JDK-8116444
+        Window rootWindow = getRootWindow(owner);
+        if (rootWindow == null) {
+            return;
         }
+
+        final Scene ownerScene = rootWindow.getScene();
+        if (ownerScene == null) {
+            return;
+        }
+
+        copyStylesheetFromOwnerScene(ownerScene);
+    }
+
+    private void copyStylesheetFromOwnerScene(Scene ownerScene) {
+        Scene scene = getScene();
+        if (scene.getUserAgentStylesheet() == null && ownerScene.getUserAgentStylesheet() != null) {
+            scene.setUserAgentStylesheet(ownerScene.getUserAgentStylesheet());
+        }
+
+        List<String> newStylesheets = new ArrayList<>();
+        for (String stylesheet : ownerScene.getStylesheets()) {
+            if (!scene.getStylesheets().contains(stylesheet)) {
+                newStylesheets.add(stylesheet);
+            }
+        }
+
+        scene.getStylesheets().addAll(newStylesheets);
     }
 
     /**
--- a/modules/javafx.graphics/src/test/java/test/javafx/stage/PopupTest.java
+++ b/modules/javafx.graphics/src/test/java/test/javafx/stage/PopupTest.java
@@ -870,6 +870,34 @@ public class PopupTest {
         assertEquals(List.of(popupStylesheet, ownerStylesheet), popup.getScene().getStylesheets());
     }
 
+    @Test
+    public void testShowTwiceAddsStylesheetOfOwnerOnlyOnce() {
+        final String ownerStylesheet = toBase64(".owner { -fx-fill: green; }");
+        scene.getStylesheets().add(ownerStylesheet);
+
+        final Popup popup = new Popup();
+
+        popup.show(stage);
+        popup.hide();
+        popup.show(stage);
+
+        assertEquals(List.of(ownerStylesheet), popup.getScene().getStylesheets());
+    }
+
+    @Test
+    public void testShowDoesNotAddStylesheetTwiceWhenPopupAndOwnerShareIt() {
+        final String sharedStylesheet = toBase64(".owner { -fx-fill: green; }");
+
+        scene.getStylesheets().addAll(sharedStylesheet);
+
+        final Popup popup = new Popup();
+        popup.getScene().getStylesheets().add(sharedStylesheet);
+
+        popup.show(stage);
+
+        assertEquals(List.of(sharedStylesheet), popup.getScene().getStylesheets());
+    }
+
     private String toBase64(String stylesheet) {
         return "data:base64," + Base64.getEncoder().encodeToString(stylesheet.getBytes(StandardCharsets.UTF_8));
     }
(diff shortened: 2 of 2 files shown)


modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
```diff
@@ -458,31 +458,34 @@ public void show(Window ownerWindow, double anchorX, double anchorY) {
     }
 
     private void showImpl(final Window owner) {
+        Window rootWindow = getRootWindow(owner);
+        if (rootWindow == null) {
```
andy-goryachev-oracle, line 462: not equivalent change: the old code set this.ownerWindow (to null in this case, see L467)

modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
```diff
@@ -458,31 +458,34 @@ public void show(Window ownerWindow, double anchorX, double anchorY) {
...
+            return;
```
andy-goryachev-oracle, line 463: I am just a bit confused here.  Referring to old version' line numbers, it looks like we should not have a null value returned from `getRootWindow(owner)`: it's being dereferenced in L477 and also L485.

modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
```diff
@@ -498,14 +501,34 @@ private void showImpl(final Window owner) {
      * @param owner the owner {@link Window}
      */
     void applyStylesheetFromOwner(Window owner) {
+        // JDK-8116444
+        Window rootWindow = getRootWindow(owner);
+        if (rootWindow == null) {
```
andy-goryachev-oracle, line 506: same thing - is it even possible for the `rootWindow` to be `null`?

Maran23: Yes, as you can see in the tests I wrote.

modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
```diff
@@ -498,14 +501,34 @@ private void showImpl(final Window owner) {
...
-            if (ownerScene.getUserAgentStylesheet() != null) {
-                scene.setUserAgentStylesheet(ownerScene.getUserAgentStylesheet());
+        if (scene.getUserAgentStylesheet() == null && ownerScene.getUserAgentStylesheet() != null) {
+            scene.setUserAgentStylesheet(ownerScene.getUserAgentStylesheet());
+        }
+
+        List<String> newStylesheets = new ArrayList<>();
+        for (String stylesheet : ownerScene.getStylesheets()) {
+            if (!scene.getStylesheets().contains(stylesheet)) {
+                newStylesheets.add(stylesheet);
```
andy-goryachev-oracle, line 527: can a PopupWindow be reused?  if so, the stylesheets might accumulate from the earlier cycles.  is this ok?  should we keep track of the stylesheets added here and remove them on `hide()`?

modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
```diff
@@ -498,14 +501,34 @@ private void showImpl(final Window owner) {
...
+    }
+
+    private void copyStylesheetFromOwnerScene(Scene ownerScene) {
         Scene scene = getScene();
-        final Scene ownerScene = getRootWindow(owner).getScene();
-        if (ownerScene != null) {
-            if (ownerScene.getUserAgentStylesheet() != null) {
-                scene.setUserAgentStylesheet(ownerScene.getUserAgentStylesheet());
+        if (scene.getUserAgentStylesheet() == null && ownerScene.getUserAgentStylesheet() != null) {
+            scene.setUserAgentStylesheet(ownerScene.getUserAgentStylesheet());
```
andy-goryachev-oracle, line 521: there might be another issue:
- owner A has the user agent stylesheet A.css
- a popup with no stylesheet is shown with the owner A.  A.css is shown
- the popup gets hidden
- either owner A changes the stylesheet to B.css, or the popup is reusing with a different owner
- the popup gets shown

since `scene.getUserAgentStylesheet()` is already A.css, this code does not set the new stylesheet, leaving the popup with a wrong style.

I know reusing the popup (context menu, etc.) is a bad idea, but surprisingly, I saw this happen many times in the past.

modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
Maran23: Does not happen as we will check with `contains`. There is also a test for this scenario.

modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
Maran23: AFAIK, there is no way that you can actually change the user agent stylesheet. It is read once and then cached.
See e.g. https://github.com/openjdk/jfx/pull/525, where a user proposed to change that. 

So I think we are actually good here

modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
Maran23: You have a point but I don't see how this could be a problem. I can't imagine a usecase where you want to call `show` just to reset the owner

andy-goryachev-oracle: I was pointing out that the change is not equivalent - in the old code, a `null` owner causes `this.ownerWindow` be set to null, in the new code it does not.

modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
andy-goryachev-oracle: This is not a question about the tests you wrote.
Sorry, I need to spell my question out.  Looking at (old) L502:

`final Scene ownerScene = getRootWindow(owner).getScene();`

you see that `getScene()` dereferences the root window, so it can't be `null` there, I guess, otherwise we had seen an NPE.

modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
andy-goryachev-oracle: Here is the case for the `PopupTest` that works in master and fails with this fix:

```
    @Test
    public void updateStylesheetFromOwnerStage() {
        String cssA = toBase64(".root { -fx-fill: green; }");
        String cssB = toBase64(".root { -fx-fill: red; }");
        scene.setUserAgentStylesheet(cssA);

        Popup p = new Popup();
        p.show(stage);
        assertEquals(cssA, p.getScene().getUserAgentStylesheet());

        p.hide();
        scene.setUserAgentStylesheet(cssB);
        p.show(stage);
        assertEquals(cssB, p.getScene().getUserAgentStylesheet());
    }
```

modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
andy-goryachev-oracle: I suppose I need to spell this out.  L527 adds new stylesheets, but there is not removing operation.  So if you keep adding stylesheets and not removing the old stylesheets, they just get accumulated.

Think of a scenario when a stylesheet is not sourced from a file/resource, but generated programmatically via a data url.

modules/javafx.graphics/src/test/java/test/javafx/stage/PopupTest.java
```diff
@@ -802,6 +804,104 @@ public void testCursorInheritance() {
...
+
+        final Popup popup = new Popup();
+        popup.getScene().getStylesheets().add(sharedStylesheet);
+
+        popup.show(stage);
+
+        assertEquals(List.of(sharedStylesheet), popup.getScene().getStylesheets());
+    }
+
+    private String toBase64(String stylesheet) {
```
andy-goryachev-oracle, line 901: should probably be static...
and since we are using it more than once, maybe we can move it to a new graphics-specific `Utils` class?

Maran23: yes, this is already on my list. Waiting for JUnit 6 to be merged first.

modules/javafx.graphics/src/main/java/javafx/stage/PopupWindow.java
Maran23: I guess this question is also about reusing `PopupWindow`?

The problem is that the stylesheets and the user agent stylesheet are hard to get right. 
We don't want to overwrite any (user) stylesheets but still add the owner ones.

Even if we try to track that, we may still accidently remove a stylesheet set from the developer. Or that was intended to stay. 
Think about we add the owner stylesheets, the developer checks the stylesheets with `contains` later and does not add a stylesheet because it is already there - fine. We show it again, and e.g. the owner changed, the stylesheets is removed but the developer really wanted that one stylesheet to stay.

I think this can't be solved perfectly, because the current magic logic is already bad. We could only try to solve this on a best-effort basis.

andy-goryachev-oracle: I buy this argument.  Reuse is bad.

Maran23: I will move this to a draft, need to think about it more. My current idea is what you also suggested, keeping track of the owner stylesheets we added. I don't like it, but that is probably the most best-effort basis fix I can think of.

I had hoped this is easier to solve, but it isn't (too many possible usecases as you also mentioned).

PR #2240, opened 2026-08-03
8389584: Allow a shared Gradle properties file

nlisker:
Allows using a Gradle properties file at the project level by requiring the user-specific one to be present in the user's Gradle dir. A Gradle properties file is necessary for a proper Gradle build system's configuration. The user file takes precedence over the project one.

---------

---------
### Progress

### Issue
 * [JDK-8389584](https://bugs.openjdk.org/browse/JDK-8389584): Allow a shared Gradle properties file (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2240/head:pull/2240` \
`$ git checkout pull/2240`

`$ git checkout pull/2240` \
`$ git pull https://git.openjdk.org/jfx.git pull/2240/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2240`

`$ git pr show -t 2240`

</details>

<a href="https://git.openjdk.org/jfx/pull/2240.diff">https://git.openjdk.org/jfx/pull/2240.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2240#issuecomment-5164911183)
</details>


nlisker:
@kevinrushforth Please review. Someone from Gluon might also want to take a look.

A notification on the mailing list prior to integration could help give a heads-up with regards to the git conflict that may arise.

AnirvanSarkar:
I have a concern about requiring OpenJFX specific configuration in the user-level `~/.gradle/gradle.properties`.

This makes OpenJFX build properties visible to every Gradle build on the machine. 
In particular, generically named project defined properties such as `JDK_HOME`, `CONF`, `INCREMENTAL`, and `LINT` could conflict with unrelated projects or plugins.

It also prevents developers from keeping OpenJFX only Gradle settings scoped to this repository, for example:

```properties
org.gradle.java.home=C:/JDK/jdk-25.0.1
org.gradle.dependency.verification=lenient
org.gradle.daemon=false
```

Is there a reason we cannot continue to support a project local, gitignored `gradle.properties` instead? 
This would still allow each developer to configure their own environment while keeping the configuration isolated to the OpenJFX build.

andy-goryachev-oracle:
I have the same concern as @AnirvanSarkar

nlisker:
> This makes OpenJFX build properties visible to every Gradle build on the machine.
> In particular, generically named project defined properties such as JDK_HOME, CONF, INCREMENTAL, and LINT could conflict with unrelated projects or plugins.
> 
> It also prevents developers from keeping OpenJFX only Gradle settings scoped to this repository, for example:

These are valid concerns.

> Is there a reason we cannot continue to support a project local, gitignored `gradle.properties` instead?

As explained above, it is required for project configuration such as the toolchain and build properties, e.g.,:
```properties
org.gradle.caching=true
org.gradle.parallel=true
org.gradle.configuration-cache=true
```
Having this file gitignored means we can't improve the build (we're stuck with Gradle 2 or 3). It is a committed file in every project I've seen (I can make a list if it helps). Figuratively, the current setup "hijacks" the file.
What we can do is add an uncommitted `local.properties` instead with user properties. It'll be read during configuration and replace the current user-specified `gradle.properties` (`JDK_HOME`, `CONF`, `INCREMENTAL`). This is also what's being done in some repos.
The `org.gradle*` properties that are required before configuration time (like the deamon switch) can be passed as command flags. For example, the daemon is very useful when developing, but can be disabled for CI/CD, so this is a good option to have on be default and turned off in the CI/CD script (note that OpenJFX [builds without config flags](https://github.com/openjdk/jfx/blob/f8e39e78b742534bb37a097d5c7e34b687e62c24/.github/workflows/submit.yml#L139)). A `GRADLE_OPTS` env var can aggregate them if it helps.
Also, git allows locally untracked files if the user really doesn't want the out-of-the-box one, but I don't think it's a good option.

It boils down to what the purpose of `gradle.properties` here is. If it's part of the build, like the `build.gradle` and `settings.gradle` files are (and it usually is), then it needs to contain the build configuration that matches the capabilities of these files. If it's a user-configuration, then the user needs to know what can be turned on. For example, making the build file support parallel execution will do nothing until the user "figures out" they need to create a `gradle.properties` locally that enables it. This is a rather terrible user experience, especially for features that have been mainstay for many years now.

I'd like to know which gradle settings users want to change and why.

AnirvanSarkar:
Thanks for the clarification.

Using a gitignored `gradle-local.properties` for `CONF` and other OpenJFX specific settings works for me.

I suggest `gradle-local.properties` rather than `local.properties`, since the latter is conventionally associated with Android builds [1].

The existing `gradle.properties.template` could therefore be renamed to `gradle-local.properties.template`.

The settings I plan to override locally are:

```properties
org.gradle.dependency.verification=lenient
org.gradle.daemon=false
```

Other developers may also want to override settings such as:

```properties
org.gradle.jvmargs=-Xmx...
```

The dependency verification override is needed in my workflow because I usually do not compile the WebKit and Media sources. Instead, I set the `STUB_RUNTIME_OPENJFX` Gradle property so that the WebKit and Media libraries are downloaded automatically from Maven Central [2].

Because the SHA-256 hashes of these stub runtimes are not currently included in `gradle/verification-metadata.xml`, I need to set: `org.gradle.dependency.verification=lenient`.

If SNAPSHOT builds of OpenJFX were available, this override would not be necessary [3].

Gradle daemon is disabled due to known issues on Windows [4][5].

*Building OpenJFX* wiki page [6] would also require an update if this pull request is merged.

---------------------------------

[1] - https://developer.android.com/build#properties-files
[2] - https://github.com/openjdk/jfx/blob/28%2B2/WEBKIT-MEDIA-STUBS.md#officially-released-libraries
[3] - https://mail.openjdk.org/archives/list/openjfx-dev@openjdk.org/thread/OPTPHQLEGL7DB3MGKWHM2L4DHTB7JBHZ/
[4] - https://bugs.openjdk.org/browse/JDK-8290976?focusedId=14512413&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14512413
[5] - https://bugs.openjdk.org/browse/JDK-8359899
[6] - https://wiki.openjdk.org/display/OpenJFX/Building+OpenJFX

nlisker:
> The settings I plan to override locally are:
> 
> ```ini
> org.gradle.dependency.verification=lenient
> org.gradle.daemon=false
> ```

To make sure we're on the same page, as explained above, these options can't be overridden in a local properties file that is read during the build configuration phase because they are used before the configuration phase. Those would have to be passed in as command flags (however they can be aggregated into one flag).

AnirvanSarkar:
Yes, that's my understanding as well.

nlisker:
@AnirvanSarkar Try this branch if you're able to and see if it answers your needs. Thanks.

AnirvanSarkar:
@nlisker  

Thanks for the changes! 

I noticed one behavior change around overriding OpenJFX project properties from the command line using `-P`.

For example, with the following in `gradle-local.properties`:

```properties
CONF = Release
```

**Running:**

```bash
./gradlew -PCONF=DebugNative properties | grep CONF
```

**Expected:**
```text
CONF: DebugNative
CONF: DebugNative
```

**Current output:**
```text
CONF: Release
CONF: Release
```

I believe it would be preferable for the command-line value (`-PCONF=DebugNative`) to take precedence over the value specified in `gradle-local.properties`, to be consistent with the existing behavior.

One possible way to preserve this behavior could be to make `localProps` global and have `defineProperty(String name, String defaultValue)` delegate to `defineProperty(String name, Properties props, String defaultValue)`, passing `localProps`.

nlisker:
> I believe it would be preferable for the command-line value (`-PCONF=DebugNative`) to take precedence over the value specified in `gradle-local.properties`, to be consistent with the existing behavior.

True. It should work now.

kevinrushforth:
I'll want to do both a few local tests and a CI build to ensure that it doesn't cause problems.

@arapte Can you also review?

build.gradle
```diff
@@ -165,6 +165,29 @@ void loadProperties(String sourceFileName) {
...
+    def propFile = new File(sourceFileName)
+    if (!propFile.canRead()) {
+        return
+    }
+
+    def localProps = new Properties()
+    propFile.withInputStream { localProps.load(it) }
+
+    def cliProps = gradle.startParameter.projectProperties
+    localProps.forEach((name, value) -> {
```
Maran23, line 183: I wonder if we could do that in the `settings.gradle` instead?
To be honest I don't know that the best practise is (or if there even is 'best way'), just something I was wondering

nlisker: > I wonder if we could do that in the settings.gradle instead?

Probably, but it has Oracle's `buildClosed` configuration there that I didn't want to conflict with. I try to go for minimal friction for the time being. This step is crucial for freeing the *de-facto* properties file for its intended use, which will allow to use the proper Java toolchain and enable parallel builds and caches. When things are wired up correctly and the entanglement reduces, we can have a look at this again.
The solution there is probably for Oracle to unify whatever it is they are doing in there with the public hooks that are provided, like this one.

> To be honest I don't know that the best practise is (or if there even is 'best way'), just something I was wondering

The best practice for what? Reading a user's properties file? Not that I know of. The Gradle team admits it's a gap, they're just not sure what to do about it. See https://github.com/gradle/gradle/issues/12283.

nlisker:
I fixed the merge conflict from #1970. It's ready for review again.

@kevinrushforth / @arapte, check that the closed build isn't being interfered with. If it's using the properties file, they'd need to be moved to the new local one.

build.gradle
Maran23: Makes sense. I will review later as well.

> This step is crucial for freeing the de-facto properties file for its intended use, which will allow to use the proper Java toolchain and enable parallel builds and caches

I totally agree on this. And this will in the end benefit all of us. And it is similar to other projects, which usually have some kind of git-ignored gradle property file for local customization (Usually the `gradle.properties`, but I also saw projects with a similar name like the one in this PR, and as you pointed out there is no standard name yet. And renaming a file later on is easy, e.g. when there finally is a convention).

.gitignore
```diff
@@ -6,13 +6,15 @@ Release/
...
 build/
 dist/
-gradle.properties
 testbin/
 **/Ensemble8/lib/
 **/Ensemble8/src/generated/resources/ensemble/search/index/write.lock
 **/nbproject/private/
 hs_err_pid*
 
+# JavaFX user-local properties for the build
```
kevinrushforth: This line has trailing whitespace. It's not an error, but would be nice to fix.

Maran23:
Tested on Arch Linux: The file is correcty loaded as expected. But it does not seem to actually use the settings.
In my case, I tried the following settings (they should fail the Gradle build):
```properties
org.gradle.configuration-cache=true
org.gradle.configuration-cache.parallel=true
org.gradle.caching=true
org.gradle.parallel=true
```

When I put them into the `gradle.properties`, build will fail. In the `gradle-local.properties`, they are read but not processed and build will succeed. Or may I misunderstand something?

nlisker:
> Tested on Arch Linux: The file is correcty loaded as expected. But it does not seem to actually use the settings. In my case, I tried the following settings (they should fail the Gradle build):
> 
> ```ini
> org.gradle.configuration-cache=true
> org.gradle.configuration-cache.parallel=true
> org.gradle.caching=true
> org.gradle.parallel=true
> ```
> 
> When I put them into the `gradle.properties`, build will fail. In the `gradle-local.properties`, they are read but not processed and build will succeed. Or may I misunderstand something?

I'm confused by the phrasing (which of them is "The file"?), but the rules are as follows: `gradle-local.properties` is for JavaFX's properties that are read during configuration time, as is stated in the [template file](https://github.com/openjdk/jfx/pull/2240/changes#diff-fe53c14ef2920a0b77cd43ba28c73d2cd13bf3677519b0cf9e6785da2a139bf2). `gradle.properties` is the only project-local file that can declare the `org.gradle` properties before configuration (special treatment). If you want to override them, you can use the command line or the user-local file (which will affect all gradle builds).

So, I would expect the build to fail currently with those properties on `gradle.properties` because they are read in time and the build file is not ready for them. It will be if I have my way with it. :)
I would expect the build to not do anything differently if they are in `gradle-local.properties` because they are read after they are needed, so they do absolutely nothing. You can keep them there for decorations if you are so inclined. :)

Maran23:
> I'm confused by the phrasing (which of them is "The file"?), but the rules are as follows: `gradle-local.properties` is for JavaFX's properties

I see, that was my mistake. I thought I can also set Gradle properties. 
I probably thought so because it is named `gradle-local.properties`, so it somewhat sounds like: `gradle.properties`, but for you locally (and git-ignored). While it is rather a 'Configure the JavaFX build'.

> I would expect the build to not do anything differently if they are in `gradle-local.properties`

So yes, can confirm that. :smile:

nlisker:
> I probably thought so because it is named `gradle-local.properties`

It's an arbitrary name chosen by convention, it can be named anything. It's [loaded manually by name](https://github.com/openjdk/jfx/pull/2240/changes#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7R326), same with our existing `build.properties`. The `gradle.properties` one is [special](https://docs.gradle.org/current/userguide/build_environment.html#sec:the_gradle_properties_file).

nlisker:
> `gradle.properties`, but for you locally (and git-ignored)

That's the gap that https://github.com/gradle/gradle/issues/12283 was asked to fix.

gradle-local.properties.template
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2013, 2026, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2013, 2023, Oracle and/or its affiliates. All rights reserved.
```
arapte: minor: revert to 2026

gradle-local.properties.template
```diff
@@ -26,8 +26,11 @@
 #####################################################################################
 #                                                                                   #
 # This properties file allows for a developer to override various properties used   #
-# when building JavaFX. Any settings you configure in the gradle.properties file    #
-# will have precedence over any values specified in the build script itself.        #
+# when building JavaFX. Any settings you configure in the gradle-local.properties   #
+# file will have precedence over any values specified in the build script itself.   #
+# Don't place Gradle properties (org.gradle.*) or system properties (systemProp.*)  #
```
arapte, line 31: Gradle properties (org.gradle.*) or system properties (systemProp.*): Developers can mistakenly add such properties to this file and fail to read this note, and the struggle to root cause it.

Would it be convenient to add a warning or fail the build in `loadLocalProperties` function, if such properties are added ?

arapte (changes requested):
Change looks good, local testing is fine. 
This change would not alter regular builds unless one creates gradle-local.properties.

There is a reference to `gradle.properties` in file `modules/javafx.media/src/main/native/xcode_project/java_home.sh` 
Could you please check if it requires any attention ?

Suggested two changes.

gradle-local.properties.template
nlisker: Good idea. I added a build failure to be on the safe side. Can reduce to logging a warning if others think otherwise.

build.gradle
```diff
@@ -300,6 +327,7 @@ void setFileTimestamps(Provider<RegularFile> archive, Instant instant) {
  *****************************************************************************/
 
 loadProperties("$projectDir/build.properties")
+loadLocalProperties("$projectDir/gradle-local.properties")
```
kevinrushforth, line 330: Question: loadProperties and loadLocalProperties are semantically different. The former is used for loading `build.properties` and turns dot-separated names into camel-case names, so `jfx.build.jdk.version` becomes `jfxBuildJdkVersion`. The latter leaves the names unchanged. Do you think it is worth adding a comment to that effect to the new method (and. maybe the existing method as well)?

gradle-local.properties.template
kevinrushforth: I think an error is fine.

kevinrushforth:
> I fixed the merge conflict from #1970. It's ready for review again.
> 
> @kevinrushforth / @arapte, check that the closed build isn't being interfered with. If it's using the properties file, they'd need to be moved to the new local one.

I checked and there are no problems. I also did a local build and a Jenkins build. I'll finish my review shortly.

nlisker:
> There is a reference to `gradle.properties` in file `modules/javafx.media/src/main/native/xcode_project/java_home.sh`
> Could you please check if it requires any attention ?

I have a couple of issues here. First, this is a macOS workaround and I can't test it. Secondly, I don't know what `JFX_MEDIA_DIR` is. It's set in the cryptic `/media/src/main/native/xcode_project/JFXMedia.xcodeproj/project.pbxproj` file *after* it's read:
```ini
JFX_MEDIA_BUILD_DIR = "$(JFX_MEDIA_DIR)/build";
JFX_MEDIA_DIR = ../../../..;
```
I don't know when this code is run. If it's during the build (doesn't seem like it), then a `gradle.properties` file is already irrelevant. If it's a setup for the build, and it generates a `gradle.properties` in a subproject, it will supposedly override the parent one. There's a discussion on it in https://github.com/gradle/gradle/issues/35029.

Does this file still get executed or is it a relic? It also has a fallback to 1.8, do we still support building on that version?

kevinrushforth (commented):
Looks good with a couple minor points.

Regarding the reported reference to `gradle.properties`, it is unused by the build -- Xcode is only possibly used by a developer when debugging on macOS.

I suspect this script is unused -- and in any case the JDK 8 fallback would completely break.

@sashamatveev can say whether he uses this script in media in connection with Xcode.

build.gradle
nlisker: I think it's more of an implementation detail and I'd say it's a discrepancy in the properties files rather than in the implementation. Why does `build.properties` use lowercase dotted names while the local properties one uses uppercase underscore names?

andy-goryachev-oracle:
the title is weird - "a shared properties"?

build.gradle
kevinrushforth: The main reason is historical -- our initial build system was ant-based. The `build.properties` file, along with similar closed property files are parsed by more than just gradle (e.g., closed shell scripts). We could consider an RFE to change it, but it would require a coordinated open/closed change along with making sure we don't have any other place that uses it.

Anyway, I think it's OK to leave it undocumented.

kevinrushforth (approved):

arapte (approved):

PR #2245, opened 2026-08-05
8386859: WebView crashes when calling document.startViewTransition()

jaybhaskar:
Issue: document.startViewTransition triggers texture update which is trying to access GL based data which is not supported in JavaFX Webkit. 

Solution: After 623.1 upgrade, BitmapTexture and BitmapTextureJava are unrelated class, since BitmapTextureJava::updateContent was no op code in 622.1 upgrade, whose drawing method was commented out. JavaFX Webkit does not have the OpenGL texture-upload and surface state hence making BitmapTexture:updateContent method no op code for JavaFX Webkit platform which restore 622.1 behavior. 

The TextureMapperTest.java test program covers JavaFX TextureMapper behavior, the test pass with fix and failed without fix.

---------

---------
### Progress

### Issue
 * [JDK-8386859](https://bugs.openjdk.org/browse/JDK-8386859): WebView crashes when calling document.startViewTransition() (**Bug** - P2)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2245/head:pull/2245` \
`$ git checkout pull/2245`

`$ git checkout pull/2245` \
`$ git pull https://git.openjdk.org/jfx.git pull/2245/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2245`

`$ git pr show -t 2245`

</details>

<a href="https://git.openjdk.org/jfx/pull/2245.diff">https://git.openjdk.org/jfx/pull/2245.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2245#issuecomment-5194065059)
</details>

### DIFF
--- a/modules/javafx.web/src/main/native/Source/WebCore/platform/graphics/texmap/TextureMapper.cpp
+++ b/modules/javafx.web/src/main/native/Source/WebCore/platform/graphics/texmap/TextureMapper.cpp
@@ -244,7 +244,11 @@ void TextureMapper::releaseUnusedTexturesNow()
 
 ClipStack& TextureMapper::clipStack()
 {
+#if PLATFORM(JAVA)
+    return m_clipStack;
+#else
     return data().currentSurface ? data().currentSurface->clipStack() : m_clipStack;
+#endif
 }
 
 void TextureMapper::beginPainting(FlipY flipY, BitmapTexture* surface)
@@ -1400,7 +1404,11 @@ void TextureMapper::bindSurface(BitmapTexture *surface)
 
 BitmapTexture* TextureMapper::currentSurface()
 {
+#if PLATFORM(JAVA)
+    return nullptr;
+#else
     return data().currentSurface.get();
+#endif
 }
 
 bool TextureMapper::beginScissorClip(const TransformationMatrix& modelViewMatrix, const FloatRect& targetRect)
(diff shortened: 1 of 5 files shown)


andy-goryachev-oracle (commented):
the reproducer works with the fix on macbook pro M1 26.5.2

modules/javafx.web/src/main/native/Source/WebCore/platform/graphics/texmap/TextureMapper.cpp
```diff
@@ -1506,6 +1514,9 @@ void TextureMapper::beginClip(const TransformationMatrix& modelViewMatrix, const
     // Increase stencilIndex and apply stencil testing.
     clipStack().setStencilIndex(stencilIndex * 2);
     clipStack().applyIfNeeded();
+#else
+    clipStack().push();
+    clipStack().intersect(enclosingIntRect(modelViewMatrix.mapRect(targetRect.rect())));
```
kevinrushforth, line 1519: @arapte Can you verify that this is equivalent? Even if it isn't, we might decide to fix it in a follow-up, given that it is provably better to do this than crash with the above code.

kevinrushforth (approved):
I confirm that this fixes the problem. The included test crashes without the fix and passes with the fix. Even if the clipping code isn't 100% equivalent, it's still better to take this fix than to knowingly call code that will crash.

modules/javafx.web/src/main/native/Source/WebCore/platform/graphics/texmap/TextureMapper.cpp
arapte: These seem different to me. The GL path can handle rounded corner rectangle clip / non-rectangular clip. I could not simulate the difference yet.

arapte (approved):

PR #2247, opened 2026-08-05
8389847: Gradle project import fails dependency verification

mstr2:
Loading the project in IntelliJ fails during Gradle synchronization with dependency verification errors for Groovy 4.0.28 artifacts. A normal command-line task such as `gradlew help` does not reproduce the problem because it does not request source artifacts for Gradle’s bundled Groovy libraries. The underlying cause is that the Gradle 9.2 upgrade introduced Groovy 4.0.28, but the committed verification metadata was never expanded for the metadata files used specifically during IDE source resolution.

When IntelliJ imports the project with source downloading enabled, Gradle attempts to attach sources for its local Groovy dependencies. Gradle performs this resolution through its internal repository named Gradle Libs. The project enables verification of dependency metadata:
```xml
<verify-metadata>true</verify-metadata>
```

Source JARs are already trusted by the existing `*-sources.jar` rule. Resolving those JARs, however, also requires Gradle module metadata and the Groovy BOM metadata. These .module and .pom files are not present in verification-metadata.xml, so Gradle rejects them.

This problem is fixed by adding checksums for the module metadata of every Groovy 4.0.28 library bundled with Gradle 9.2.

---------

---------
### Progress

### Issue
 * [JDK-8389847](https://bugs.openjdk.org/browse/JDK-8389847): Gradle project import fails dependency verification (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2247/head:pull/2247` \
`$ git checkout pull/2247`

`$ git checkout pull/2247` \
`$ git pull https://git.openjdk.org/jfx.git pull/2247/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2247`

`$ git pr show -t 2247`

</details>

<a href="https://git.openjdk.org/jfx/pull/2247.diff">https://git.openjdk.org/jfx/pull/2247.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2247#issuecomment-5198030089)
</details>


Maran23 (approved):
I also have this issue and solved it by adding:
`<trust file=".*groovy.*" regex="true"/>`, 
which really is not a good solution but worked for me. So thanks for solving this!

I can confirm this fixes the issue and is the correct way.

One question: How can we generate those entries in the future, e.g. for new Gradle versions?
When I run the command in the `README.txt` next to `verification-metadata.xml`, those entries are not added for me. Maybe we should update / add the steps into this file for the future?

mstr2:
> One question: How can we generate those entries in the future, e.g. for new Gradle versions? When I run the command in the `README.txt` next to `verification-metadata.xml`, those entries are not added for me. Maybe we should update / add the steps into this file for the future?

I've added a section to gradle/README.txt

Maran23 (approved):

nlisker:
When I generate the verification metadata, I get entries that are different than yours: [verification-metadata.xml](https://github.com/user-attachments/files/30904771/verification-metadata.xml). Do we add the entries that are required by all IDEs?

mstr2:
> When I generate the verification metadata, I get entries that are different than yours: [verification-metadata.xml](https://github.com/user-attachments/files/30904771/verification-metadata.xml). Do we add the entries that are required by all IDEs?

Well, you do get the same entries for Groovy, which is good. I see no reason not to add entries that are required by other IDEs, as otherwise working with JavaFX is just gets a little harder. The alternative to adding individual entries is a blanket `<trust/>` entry (see Marius' comment).

nlisker:
In this PR do you only want to add the Groovy dependencies without the IDE ones?

mstr2:
> In this PR do you only want to add the Groovy dependencies without the IDE ones?

For this PR, I'm only adding the dependencies that are required to make JavaFX work with IntelliJ. I'm not familiar with other IDEs and how people use them with JavaFX, so that's something I'd leave for others to figure out.

kevinrushforth:
@arapte I believe you use IntelliJ. Can you review this?

kevinrushforth:
FWIW, this seems fine to me.

gradle/README.txt
```diff
@@ -41,7 +41,27 @@ Recreate the dependency verification file as follows:
...
+   add the following setting to gradle.properties:
+
+   org.gradle.dependency.verification=lenient
+
+   Reload the Gradle project in IntelliJ with source downloading enabled.
+   Inspect the dependency verification report under:
+
+   build/reports/dependency-verification/
+
+   After adding the artifact entries, remove the temporary lenient setting
```
arapte: It requires to manually update the gradle/verification-metadata.xml after this step?
If Yes, may be mention it explicitly.
And, Is there a way to avoid this manual step ? Any option that updates the gradle/verification-metadata.xml file directly ?

arapte (changes requested):
> This problem is fixed by adding checksums for the module metadata of every Groovy 4.0.28 library bundled with Gradle 9.2.

These entries are for Gradle 9.2.0.
The gradle version is now updated to 9.7.0 : So, do the entries in `gradle/verification-metadata.xml` needs to be updated ?

mstr2:
I now think that Marius' solution is better. This is an IDE-only dependency, it's not required to build JavaFX. I have added a narrowly scoped `<trust/>` element, which means that we don't need to keep track of which concrete version of Groovy is used by Gradle.

Maran23 (approved):
Good idea and better than my regex.

I did a global search on GitHub and found that people always use a variation of this solution: https://github.com/search?q=%3Ctrust+file%3D%22groovy&type=code

So looks good to me. Tested on Windows 11, for some reason I don't have this problem on Linux.

gradle/verification-metadata.xml
```diff
@@ -5,6 +5,7 @@
       <verify-signatures>false</verify-signatures>
       <trusted-artifacts>
          <trust file=".*-sources[.]jar" regex="true"/>
+         <trust group="org[.]apache[.]groovy" file="groovy-.*[.](?:pom|module)" regex="true"/>
```
arapte, line 11: Minor: please add a short comment before this line.
Suggestion: Required for importing the JavaFX Gradle project in IntelliJ IDEA

arapte (changes requested):
Looks good, Seems better to avoid checksum churn with every Gradle update.
Minor change requested.

Maran23:
Got this issue under Linux today as well and can also confirm that it fixes the issue there as well, as expected.

gradle/verification-metadata.xml
mstr2: I added a comment, and also updated README.txt

Maran23 (approved):

arapte (approved):

PR #2250, opened 2026-08-06
8387301: ListView, ComboBox, TableView, TreeTableView fail when item type is a value class

kevinrushforth:
This PR creates a WeakReferenceWrapper object to replace direct uses of WeakReference in controls where the referent is a user-supplied object of an unknown type.

As noted in JEP 401, which is now integrated into JDK 28, "The garbage collection APIs in java.lang.ref ... do not allow developers to manually manage value objects in the heap. Attempts to create Reference objects for value objects throw IdentityException at run time."

Several core JDK classes such as all of the primitive wrappers (e.g., `Integer`, `Character`), `Optional`, `LocalDateTime`, and a few others are now value types if JDK 28 is run with the `--enable-preview` option.

The `ListView`, `ComboBox`, `TableView`, and `TreeTableView` controls take a parameterized item type and hold items of that type. The following places in the implementation create weak references to an item. If that item type is a value class -- meaning that it does not have identity -- creating the `WeakReference` fails.

As noted in the JBS issue, there are 3 cases to consider.

1. `SelectedItemsReadOnlyObservableList<E>` -- `E` is the item type (created by `MultipleSelectionModelBase<T>`) : `ListView`, `TableView`, `ComboBox` (due to its skin creating a `ListView<T>`) -- replace with `WeakReferenceWrapper`

2. `TablePosition<S,T>` -- `S` is the item type : `TableView` -- the reference is unused, so I removed it

3. `TableCell<S,T>` and `TreeTableCell<S,T>` -- `S` is the item type : `TableView`, `TreeTableView` -- replace with `WeakReferenceWrapper`

The new `WeakReferenceWrapper` class takes a referent of any type and either creates a WeakReference (if it has identity) or directly stores the reference (if it is null or does not have identity). I added a test for the wrapper.

All of the controls tests pass with this fix. I did three test runs as follows:

1. JDK 25
2. JDK 28 without `--enable-preview`
3. JDK 28 with `--enable-preview`

Without the fix, 31 controls tests fails on the 3rd run.

---------

---------
### Progress

### Issue
 * [JDK-8387301](https://bugs.openjdk.org/browse/JDK-8387301): ListView, ComboBox, TableView, TreeTableView fail when item type is a value class (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2250/head:pull/2250` \
`$ git checkout pull/2250`

`$ git checkout pull/2250` \
`$ git pull https://git.openjdk.org/jfx.git pull/2250/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2250`

`$ git pr show -t 2250`

</details>

<a href="https://git.openjdk.org/jfx/pull/2250.diff">https://git.openjdk.org/jfx/pull/2250.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2250#issuecomment-5297444667)
</details>

### DIFF
--- a/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/SelectedItemsReadOnlyObservableList.java
+++ b/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/SelectedItemsReadOnlyObservableList.java
@@ -28,7 +28,6 @@ import javafx.collections.ListChangeListener;
 import javafx.collections.ObservableList;
 import javafx.collections.ObservableListBase;
 
-import java.lang.ref.WeakReference;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.function.Supplier;
@@ -38,7 +37,7 @@ public abstract class SelectedItemsReadOnlyObservableList<E> extends ObservableL
     // This is the actual observable list of selected indices used in the selection model
     private final ObservableList<Integer> selectedIndices;
     private final Supplier<Integer> modelSizeSupplier;
-    private final List<WeakReference<E>> itemsRefList;
+    private final List<WeakReferenceWrapper<E>> itemsRefList;
 
     public SelectedItemsReadOnlyObservableList(ObservableList<Integer> selectedIndices, Supplier<Integer> modelSizeSupplier) {
         this.modelSizeSupplier = modelSizeSupplier;
@@ -85,7 +84,7 @@ public abstract class SelectedItemsReadOnlyObservableList<E> extends ObservableL
             // FIXME we could make this more efficient by only making the reported changes to the list
             itemsRefList.clear();
             for (int selectedIndex : selectedIndices) {
-                itemsRefList.add(new WeakReference<>(getModelItem(selectedIndex)));
+                itemsRefList.add(new WeakReferenceWrapper<>(getModelItem(selectedIndex)));
             }
 
             endChange();
--- a/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/SelectedItemsReadOnlyObservableList.java
+++ b/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/SelectedItemsReadOnlyObservableList.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, 2021, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2016, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
--- a/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/WeakReferenceWrapper.java
+++ b/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/WeakReferenceWrapper.java
@@ -38,6 +38,22 @@ import java.util.Objects;
  * In the case of a value object, the referent is never collected, so it is only
  * suitable for uses that do not rely on the object being placed onto a reference
  * queue.
+ * This class might not be suitable for large trees of value objects or for
+ * value objects that wrap an identity object (for example, {@code Optional<Object>}.
+ * Using it for such cases can lead to memory being held longer than expected
+ * (as long as the {@code WeakReferenceWrapper} itself is held).
+ * <p>
+ * NOTE: we might be able to modify or eliminate this class in the future based
+ * on one the following:
+ *
+ * <ol>
+ * <li>A future version of the value objects preview (a follow-on to JEP 401) might
+ * relax the restriction on weak references of value objects. Depending on how
+ * this is done, we might change this wrapper to take advantage of that.</li>
+ * <li>A future improvement to list-based UI controls might eliminate the need
+ * for some or all of the weak references to application-proved items. If all are
+ * eliminated, then we can eliminate this wrapper class.</li>
+ * </ol>
  *
  * @param <T> the type of the referent
  */
--- a/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/WeakReferenceWrapper.java
+++ b/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/WeakReferenceWrapper.java
@@ -33,7 +33,7 @@ import java.util.Objects;
 /**
  * Wrapper class that holds a reference to an object with or without identity.
  * If the object is non-null and has identity, a weak reference is created and
- * stored; otherwise it holds a reference to the object itself,
+ * stored; otherwise it holds a reference to the object itself.
  * <p>
  * In the case of a value object, the referent is never collected, so it is only
  * suitable for uses that do not rely on the object being placed onto a reference
@@ -50,7 +50,7 @@ public class WeakReferenceWrapper<T> {
     static {
         Method meth;
         try {
-            meth = Objects.class.getDeclaredMethod("hasIdentity", Object.class);
+            meth = Objects.class.getMethod("hasIdentity", Object.class);
         } catch (NoSuchMethodException ex) {
             meth = null;
         }
@@ -59,12 +59,13 @@ public class WeakReferenceWrapper<T> {
     }
 
     /**
-     * Helper function that calls Object.hasIdentity via reflection.
-     * If `obj` is null, treat is as having no identity (matching JDK 28's
-     * behavior with or without --enable-preview) and return false.
-     * If `obj` is not null, check whether Objects.hasIdentity exists: if the
-     * method doesn't exist or cannot be invoked, return true; otherwise, call
-     * Objects.hasIdentity(obj) reflectively and return its value.
+     * Helper method that reflectively calls Objects.hasIdentity to determine
+     * whether we should create and hold a weak reference to the given object.
+     * If {@code obj} is null, treat it as having no identity (matching JDK 28's
+     * behavior with or without {@code --enable-preview}) and return false.
+     * If {@code obj} is not null, check whether {@code Objects.hasIdentity} exists:
+     * if the method doesn't exist or cannot be invoked, return true; otherwise,
+     * call {@code Objects.hasIdentity(obj)} reflectively and return its value.
      */
     private static boolean useWeakRef(Object obj) {
         if (obj == null) {
@@ -86,7 +87,9 @@ public class WeakReferenceWrapper<T> {
     /**
      * Creates a new reference that refers to the given object.
      * If the object is non-null and has identity, a weak reference is created and
-     * stored; otherwise it holds a reference to the object itself,
+     * stored; otherwise it holds a reference to the object itself.
+     *
+     * @param obj the object this reference will refer to
      */
     public WeakReferenceWrapper(T obj) {
         if (useWeakRef(obj)) {
--- a/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/WeakReferenceWrapper.java
+++ b/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/WeakReferenceWrapper.java
@@ -70,7 +70,7 @@ public class WeakReferenceWrapper<T> {
             return true;
         } else {
             try {
-                System.err.println("[3] Calling Ojbects.hasIdentityMethod(obj)");
+                System.err.println("[3] Calling Objects.hasIdentityMethod(obj)");
                 return (Boolean)hasIdentityMethod.invoke(null, obj);
             } catch (Exception ex) {
                 return true;
--- a/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/WeakReferenceWrapper.java
+++ b/modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/WeakReferenceWrapper.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) 2026, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.  Oracle designates this
+ * particular file as subject to the "Classpath" exception as provided
+ * by Oracle in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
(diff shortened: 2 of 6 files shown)


modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/WeakReferenceWrapper.java
```diff
@@ -0,0 +1,98 @@
...
+    private static boolean hasIdentity(Object obj) {
+        if (obj == null) {
+            System.err.println("[1] null");
+            return false;
+        } else if (hasIdentityMethod == null) {
+            System.err.println("[2] hasIdentityMethod == null");
+            return true;
+        } else {
+            try {
+                System.err.println("[3] Calling Ojbects.hasIdentityMethod(obj)");
```
andy-goryachev-oracle: Ojbects

kevinrushforth: Fixed. My fingers do that all the time!

kevinrushforth:

@dansmithcode Your comments would be welcome if you have time to look at this.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/WeakReferenceWrapper.java
```diff
@@ -0,0 +1,110 @@
...
+ * If the object is non-null and has identity, a weak reference is created and
+ * stored; otherwise it holds a reference to the object itself.
+ * <p>
+ * In the case of a value object, the referent is never collected, so it is only
+ * suitable for uses that do not rely on the object being placed onto a reference
+ * queue.
+ *
+ * @param <T> the type of the referent
+ */
+public class WeakReferenceWrapper<T> {
```
andy-goryachev-oracle, line 60: An application or a library should not be forced to do this, in my opinion.

WeakReference(T) should work as before - in case of a value object it should hold the reference indefinitely because it's the same behavior as before.  It may work differently with the WeakReference(T,ReferenceQueue) constructor which is ok because it would affect a much smaller space of use cases.

It is probably ok to do it right now, to avoid javafx breaking with the value objects preview enabled.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/WeakReferenceWrapper.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.Objects;
+
+/**
+ * Wrapper class that holds a reference to an object with or without identity.
+ * If the object is non-null and has identity, a weak reference is created and
+ * stored; otherwise it holds a reference to the object itself.
+ * <p>
+ * In the case of a value object, the referent is never collected, so it is only
+ * suitable for uses that do not rely on the object being placed onto a reference
+ * queue.
```
andy-goryachev-oracle, line 40: I might suggest to include the JBS number for when we need to undo / redo things due to inevitable change in the value objects JEP (and possibly a link to the JEP itself).

Perhaps also say a couple of words about the fact that this class should not exist had they decided to make the WeakReference implementation handle this case transparently.

modules/javafx.controls/src/main/java/javafx/scene/control/TableCell.java
```diff
@@ -631,7 +632,7 @@ private boolean isInCellSelectionMode() {
 
     private boolean isFirstRun = true;
 
-    private WeakReference<S> oldRowItemRef;
+    private WeakReferenceWrapper<S> oldRowItemRef;
```
andy-goryachev-oracle, line 635: While it is ok to apply this workaround (WeakReferenceWrapper) here, this kind of change in the WeakReference behavior is just awful: we should never force the application developers (r a third party library developers) to make a change like this.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/WeakReferenceWrapper.java
kevinrushforth: > WeakReference(T) should work as before

But that isn't the case.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/WeakReferenceWrapper.java
kevinrushforth: There is no JBS issue or JEP that will eliminate the need for this. The restriction is intentional with no current plan to change it.

I could add a comment that this class _might_ become unnecessary in the future, if a there is a change in the way `WeakReference` deals with value objects, but it is uncertain if or when that might be.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/SelectedItemsReadOnlyObservableList.java
```diff
@@ -38,7 +37,7 @@ public abstract class SelectedItemsReadOnlyObservableList<E> extends ObservableL
     // This is the actual observable list of selected indices used in the selection model
     private final ObservableList<Integer> selectedIndices;
     private final Supplier<Integer> modelSizeSupplier;
-    private final List<WeakReference<E>> itemsRefList;
+    private final List<WeakReferenceWrapper<E>> itemsRefList;
```
Maran23, line 40: This is still confusing me as also mentioned in the ticket. For example, the `SelectionModel` has a `selectedItem` property, where we save the item `<T>`. So the item `T` is saved, without any `WeakReference` handling.

And this is used in e.g. `ComboBox`, which does not seems to leak memory. So I wonder why we use a `WeakReference` here.

The `WeakReference` usage in the `TableCell` is also questionable.
Maybe worth to research in a follow-up?

kevinrushforth: Yes, this is definitely worth a follow-up.

This is a very odd use of weak references. It's a short-lived list of previously removed items used in a subsequent change event.

Dan Smith @dlsmith-dev suggested something similar in this thread:

https://mail.openjdk.org/archives/list/valhalla-dev@openjdk.org/thread/TD7AQH4SNRS4QAQAXTISKIERNK6LINYB/

The use of weak references suggests that when this was implemented, they were worried about a possible leak, so we would need to show that it can't leak in practice.

If we can prove using strong references won't lead to a leak, that would be the best solution anyway.

kevinrushforth:
I updated the docs for the `WeakReferenceWrapper` class to indicate that we might modify or eliminate it if value objects change to allow use by weak references or if we can eliminate the need to use weak references to application-provided items in list-based controls.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/WeakReferenceWrapper.java
```diff
@@ -0,0 +1,126 @@
...
+ * eliminated, then we can eliminate this wrapper class.</li>
+ * </ol>
+ *
+ * @param <T> the type of the referent
+ */
+public class WeakReferenceWrapper<T> {
+
+    private static final Method hasIdentityMethod;
+    private final T obj;
+    private final WeakReference<T> ref;
```
andy-goryachev-oracle, line 64: We could, in theory, save one pointer and just use 

`private Object ref;`

which is either a `WeakReference` or the object itself, and use `(instanceof WeakReference r)` in `get()`.

This code is ok since we won't have many instances of this wrapper anyway.

andy-goryachev-oracle (approved):

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/SelectedItemsReadOnlyObservableList.java
kevinrushforth: I filed [JDK-8390875](https://bugs.openjdk.org/browse/JDK-8390875) and [JDK-8390878](https://bugs.openjdk.org/browse/JDK-8390878) to track the follow-up issues.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/WeakReferenceWrapper.java
kevinrushforth: I had initially considered that approach, but ended up going with two fields. Thanks.

kevinrushforth:
@Maran23 @arapte Can one of you be the second reviewer?

@dlsmith-dev do you have any comments?

Maran23:
Yes, will review soon.

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/WeakReferenceWrapper.java
```diff
@@ -0,0 +1,126 @@
...
+ * relax the restriction on weak references of value objects. Depending on how
+ * this is done, we might change this wrapper to take advantage of that.</li>
+ * <li>A future improvement to list-based UI controls might eliminate the need
+ * for some or all of the weak references to application-proved items. If all are
+ * eliminated, then we can eliminate this wrapper class.</li>
+ * </ol>
+ *
+ * @param <T> the type of the referent
+ */
+public class WeakReferenceWrapper<T> {
```
ExE-Boss, line 60: Personally, I’d use a sealed interface with a strong and weak implementation instead, something like:
```java
import java.lang.invoke.MethodHandle;
import java.util.Objects;
import static java.lang.invoke.MethodHandles.publicLookup;
import static java.lang.invoke.MethodType.methodType;

import org.jspecify.annotations.*;

@NullMarked
sealed interface WeakRefWrapper<T> {
	private static boolean hasIdentity(final @Nullable Object obj) {
		if (obj == null) {
			return false;
		}

		final @Namespace class Holder {
			static final @Nullable MethodHandle OBJECTS_HAS_IDENTITY;
			static {
				MethodHandle hasIdentity = null;
				try {
					hasIdentity = publicLookup().findStatic(
						Objects.class,
						"hasIdentity",
						methodType(boolean.class, Object.class)
					);
				} catch (final IllegalAccessException ex) {
					// `Exceptions::fromROE` (fromReflectiveOperationException) behaves mostly
					// like `java.lang.invoke.MethodHandleNatives::mapLookupExceptionToError`
					throw Exceptions.fromROE(ex);
				} catch (NoSuchMethodException _) {
					// ok
				}
				OBJECTS_HAS_IDENTITY = hasIdentity;
			}
		}

		final var hasIdentity = Holder.OBJECTS_HAS_IDENTITY;
		if (hasIdentity == null) {
			return true;
		}

		try {
			return (boolean) hasIdentity.invokeExact(obj);
		} catch (final Throwable t) {
			// `Exceptions::asUnchecked` rethrows the `Throwable`
			// as if it were an unchecked exception
			throw Exceptions.asUnchecked(t);
		}
	}

	public static <T extends @Nullable Object> WeakRefWrapper<T> of(final T value) {
		if (hasIdentity(value)) {
			return new WeakImpl<>(value);
		} else {
			return new StrongImpl<>(value);
		}
	}

	public @Nullable T get();

	private static final class WeakImpl<T> extends WeakReference<T>
		implements WeakRefWrapper<T> {

		WeakImpl(final T referent) {
			super(Objects.requireNonNull(referent));
		}
	}

	private static final class StrongImpl<T extends @Nullable Object>
		implements WeakRefWrapper<T> {

		private final T value;
		StrongImpl(final T value) {
			this.value = value;
		}

		@Override
		public T get() {
			return this.value;
		}
	}
}
```

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/WeakReferenceWrapperTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/*
+ * Test WeakReferenceWrapper utility.
+ */
+public class WeakReferenceWrapperTest {
+
+    // POJO with identity
+    static class POJO {
```
Maran23, line 40: Minor: Could be a `record`

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/WeakReferenceWrapperTest.java
```diff
@@ -0,0 +1,98 @@
...
+        POJO(int i) {
+            this.i = i;
+        }
+    }
+
+    // Test WeakReferenceWrapper with null, Integer (could be a value object),
+    // String (always an identity object), and a POJO (identity object).
+    @Test
+    public void testWeakReferenceWrapper() {
+        var nullRef = new WeakReferenceWrapper<Object>(null);
```
Maran23, line 66: Minor: Here and below, the type argument `<>` can also be removed.
Just checked, the left `var` side is always able to infer the correct type.

Maran23 (approved):
Looks good to me. Left some minor comments. 

I can confirm that the other `WeakReference` usages in `TableCell`, `TreeTableCell` and in both `TablePosition` classes are used for the `TableView` or `TreeItem`.

So everything that references an item is correctly replaced by the `WeakReferenceWrapper`. 

Thanks for creating the follow-up investigation tickets, when I have more time I can have a look into the details of those!

modules/javafx.controls/src/main/java/com/sun/javafx/scene/control/WeakReferenceWrapper.java
kevinrushforth: > Personally, I’d use a sealed interface with a strong and weak implementation instead

That seems cleaner than the option I had considered (and Andy also mentioned) to eliminate the extra field. I prototyped it using an abstract class, which I like better than an interface for this case, so I might go with it.

I opted to keep it simple to minimize the changes (for example, there is no need for the Holder class and I dispensed with the annotations).

ExE-Boss: > I prototyped it using an abstract class, which I like better than an interface for this case, so I might go with it.

That can’t use `extends WeakReference`

kevinrushforth: Ah, good point. And having `WeakImpl` extend `WeakReference` would eliminate a second object for each reference wrapper (and a level of indirection on the get call).

kevinrushforth: Given that we already have follow-up issues to consider eliminating this WeakReferenceWrapper, I've decided to leave this wrapper as it currently stands: a single concrete helper class with two fields. I'll file a follow-up issue to improve it if it ends up sticking around.

It's a cleaner implementation and better encapsulation using a class (which matters because the wrapper class needs to be public since it is used from more than one package), even though there is an extra indirection when calling `wr.get()`. Reworking it to eliminate the extra indirection, while keeping the existing encapsulation, would be possible with a little work, but it isn't worth it at this time.

We don't expect a huge number of these and they aren't in the hot path (they are short-lived items created and used only when items are removed).

I did notice a couple typos in the docs, which I'll fix, but I'll defer any other improvements until we know whether it is worth the effort.

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/WeakReferenceWrapperTest.java
kevinrushforth: Yes, it could have been since records are identity classes by default. I think I'll leave it as is though.

modules/javafx.controls/src/test/java/test/com/sun/javafx/scene/control/WeakReferenceWrapperTest.java
kevinrushforth: I wasn't sure that would work for null, but I prefer to leave it as is to be more self-documenting (although it really wouldn't matter hear, that's one of the drawbacks of `var` some times).

andy-goryachev-oracle (approved):

Maran23 (approved):
doc changes look good.

PR #2253, opened 2026-08-07
8389970: StringIndexOutOfBoundsException when a text containing a mnemonic is removed

eduardsdv:
This PR fixes the StringIndexOutOfBoundsException, that occurs when an empty text is set to a Labeled, that previously contained a mnemonic.

The reason for the error is that the ``updateDisplayedText(double, double)``, which also updates ``containsMnemonic`` flag, was not invoked if the text was empty. The value of this flag was still ``true`` but the index of the mnemonic character inside ``MnemonicInfo`` had already been updated to -1. This led to the StringIndexOutOfBoundsException in the line 611.

I fixed it by moving the call to ``updateDisplayedText(double, double)`` outside the if-clause, so that it is always called when the text is being laid out. This should not affect performance because the method already checks whether recalculation is required. If not, it exits quickly.

---------

---------
### Progress

### Issue
 * [JDK-8389970](https://bugs.openjdk.org/browse/JDK-8389970): StringIndexOutOfBoundsException when a text containing a mnemonic is removed (**Bug** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2253/head:pull/2253` \
`$ git checkout pull/2253`

`$ git checkout pull/2253` \
`$ git pull https://git.openjdk.org/jfx.git pull/2253/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2253`

`$ git pr show -t 2253`

</details>

<a href="https://git.openjdk.org/jfx/pull/2253.diff">https://git.openjdk.org/jfx/pull/2253.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2253#issuecomment-5219876511)
</details>

### DIFF
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java
@@ -548,11 +548,12 @@ public abstract class LabeledSkinBase<C extends Labeled> extends SkinBase<C> {
             graphicHeight = graphic.getLayoutBounds().getHeight();
         }
 
+        updateDisplayedText(w, h);
+
         if (ignoreText) {
             textWidth  = textHeight = 0;
             text.setText("");
         } else {
-            updateDisplayedText(w, h); // Have to do this just in case it needs to be recomputed
             textWidth  = snapSizeX(Math.min(text.getLayoutBounds().getWidth(),  wrapWidth));
             textHeight = snapSizeY(Math.min(text.getLayoutBounds().getHeight(), wrapHeight));
         }
@@ -599,14 +600,19 @@ public abstract class LabeledSkinBase<C extends Labeled> extends SkinBase<C> {
         Point2D mnemonicPos = null;
         double mnemonicWidth = 0.0;
         double mnemonicHeight = 0.0;
-        int mnemonicIndex = mnemonicInfo != null ? mnemonicInfo.getMnemonicIndex() : -1;
-        if (containsMnemonic && mnemonicIndex >= 0) {
-            final Font font = text.getFont();
-            String preSt = mnemonicInfo.getText();
-            boolean isRTL = (labeledNode.getEffectiveNodeOrientation() == NodeOrientation.RIGHT_TO_LEFT);
-            mnemonicPos = Utils.computeMnemonicPosition(font, preSt, mnemonicIndex, this.wrapWidth, labeled.getLineSpacing(), isRTL);
-            mnemonicWidth = Utils.computeTextWidth(font, preSt.substring(mnemonicIndex, mnemonicIndex + 1), 0);
-            mnemonicHeight = Utils.computeTextHeight(font, "_", 0, text.getBoundsType());
+        if (containsMnemonic && mnemonicInfo != null) {
+            int mnemonicIndex = mnemonicInfo.getMnemonicIndex();
+            String cleanText = mnemonicInfo.getText();
+            // The mnemonic KeyCode can also be defined in the following form: Exit_(q)
+            // In this case, no mnemonic character is available and the mnemonicIndex value is equal to the text.length().
+            // This means we cannot substring the character from the text.
+            if (mnemonicIndex >= 0 && mnemonicInfo.getExtendedMnemonicText() == null) {
+                final Font font = text.getFont();
+                boolean isRTL = (labeledNode.getEffectiveNodeOrientation() == NodeOrientation.RIGHT_TO_LEFT);
+                mnemonicPos = Utils.computeMnemonicPosition(font, cleanText, mnemonicIndex, this.wrapWidth, labeled.getLineSpacing(), isRTL);
+                mnemonicWidth = Utils.computeTextWidth(font, cleanText.substring(mnemonicIndex, mnemonicIndex + 1), 0);
+                mnemonicHeight = Utils.computeTextHeight(font, "_", 0, text.getBoundsType());
+            }
         }
 
 
@@ -969,7 +975,9 @@ public abstract class LabeledSkinBase<C extends Labeled> extends SkinBase<C> {
             if (cleanText != null && cleanText.length() > 0
                     && mnemonicInfo != null
                     && !com.sun.javafx.PlatformUtil.isMac()
-                    && getSkinnable().isMnemonicParsing()) {
+                    && getSkinnable().isMnemonicParsing()
+                    // Consider the mnemonic as not present if the text is not visible at all
+                    && labeled.getContentDisplay() != ContentDisplay.GRAPHIC_ONLY) {
                 /*
                 ** the Labeled has a MnemonicParsing property,
                 ** if set true, then auto-parsing will check for
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java
@@ -548,11 +548,15 @@ public abstract class LabeledSkinBase<C extends Labeled> extends SkinBase<C> {
             graphicHeight = graphic.getLayoutBounds().getHeight();
         }
 
+        // Have to do this just in case it needs to be recomputed
+        // This is especially important because it also recomputes the 'containsMnemonic' value,
+        // which is used by this method later on.
+        updateDisplayedText(w, h);
+
         if (ignoreText) {
             textWidth  = textHeight = 0;
             text.setText("");
         } else {
-            updateDisplayedText(w, h); // Have to do this just in case it needs to be recomputed
             textWidth  = snapSizeX(Math.min(text.getLayoutBounds().getWidth(),  wrapWidth));
             textHeight = snapSizeY(Math.min(text.getLayoutBounds().getHeight(), wrapHeight));
         }
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java
@@ -1104,7 +1104,7 @@ public abstract class LabeledSkinBase<C extends Labeled> extends SkinBase<C> {
         Labeled labeled = getSkinnable();
         String sourceText = labeled.getText();
 
-        if (labeled.isMnemonicParsing()) {
+        if (labeled.isMnemonicParsing() && !com.sun.javafx.PlatformUtil.isMac()) {
             if (mnemonicInfo == null) {
                 mnemonicInfo = new MnemonicInfo(sourceText);
             } else {
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java
@@ -551,12 +551,12 @@ public abstract class LabeledSkinBase<C extends Labeled> extends SkinBase<C> {
         // Have to do this just in case it needs to be recomputed
         // This is especially important because it also recomputes the 'containsMnemonic' value,
         // which is used by this method later on.
-        updateDisplayedText(w, h);
 
         if (ignoreText) {
             textWidth  = textHeight = 0;
             text.setText("");
         } else {
+            updateDisplayedText(w, h);
             textWidth  = snapSizeX(Math.min(text.getLayoutBounds().getWidth(),  wrapWidth));
             textHeight = snapSizeY(Math.min(text.getLayoutBounds().getHeight(), wrapHeight));
         }
@@ -603,12 +603,13 @@ public abstract class LabeledSkinBase<C extends Labeled> extends SkinBase<C> {
         Point2D mnemonicPos = null;
         double mnemonicWidth = 0.0;
         double mnemonicHeight = 0.0;
-        if (containsMnemonic) {
+        int mnemonicIndex = mnemonicInfo != null ? mnemonicInfo.getMnemonicIndex() : -1;
+        if (containsMnemonic && mnemonicIndex >= 0) {
             final Font font = text.getFont();
             String preSt = mnemonicInfo.getText();
             boolean isRTL = (labeledNode.getEffectiveNodeOrientation() == NodeOrientation.RIGHT_TO_LEFT);
-            mnemonicPos = Utils.computeMnemonicPosition(font, preSt, mnemonicInfo.getMnemonicIndex(), this.wrapWidth, labeled.getLineSpacing(), isRTL);
-            mnemonicWidth = Utils.computeTextWidth(font, preSt.substring(mnemonicInfo.getMnemonicIndex(), mnemonicInfo.getMnemonicIndex() + 1), 0);
+            mnemonicPos = Utils.computeMnemonicPosition(font, preSt, mnemonicIndex, this.wrapWidth, labeled.getLineSpacing(), isRTL);
+            mnemonicWidth = Utils.computeTextWidth(font, preSt.substring(mnemonicIndex, mnemonicIndex + 1), 0);
             mnemonicHeight = Utils.computeTextHeight(font, "_", 0, text.getBoundsType());
         }
 
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java
@@ -975,9 +975,7 @@ public abstract class LabeledSkinBase<C extends Labeled> extends SkinBase<C> {
             if (cleanText != null && cleanText.length() > 0
                     && mnemonicInfo != null
                     && !com.sun.javafx.PlatformUtil.isMac()
-                    && getSkinnable().isMnemonicParsing()
-                    // Consider the mnemonic as not present if the text is not visible at all
-                    && labeled.getContentDisplay() != ContentDisplay.GRAPHIC_ONLY) {
+                    && getSkinnable().isMnemonicParsing()) {
                 /*
                 ** the Labeled has a MnemonicParsing property,
                 ** if set true, then auto-parsing will check for
@@ -1030,16 +1028,19 @@ public abstract class LabeledSkinBase<C extends Labeled> extends SkinBase<C> {
             }
 
             if (containsMnemonic) {
-                if (mnemonic_underscore == null) {
-                    mnemonic_underscore = new Line();
-                    mnemonic_underscore.setStartX(0.0f);
-                    mnemonic_underscore.setStartY(0.0f);
-                    mnemonic_underscore.setEndY(0.0f);
-                    mnemonic_underscore.getStyleClass().clear();
-                    mnemonic_underscore.getStyleClass().setAll("mnemonic-underline");
-                }
-                if (!getChildren().contains(mnemonic_underscore)) {
-                    getChildren().add(mnemonic_underscore);
+                // If only the graphic is visible, we do not need mnemonic_underscore node
+                if (labeled.getContentDisplay() != ContentDisplay.GRAPHIC_ONLY) {
+                    if (mnemonic_underscore == null) {
+                        mnemonic_underscore = new Line();
+                        mnemonic_underscore.setStartX(0.0f);
+                        mnemonic_underscore.setStartY(0.0f);
+                        mnemonic_underscore.setEndY(0.0f);
+                        mnemonic_underscore.getStyleClass().clear();
+                        mnemonic_underscore.getStyleClass().setAll("mnemonic-underline");
+                    }
+                    if (!getChildren().contains(mnemonic_underscore)) {
+                        getChildren().add(mnemonic_underscore);
+                    }
                 }
             } else if (mnemonic_underscore != null && getChildren().contains(mnemonic_underscore)) {
                 Platform.runLater(() -> {
--- a/modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java
+++ b/modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java
@@ -946,24 +946,26 @@ public abstract class LabeledSkinBase<C extends Labeled> extends SkinBase<C> {
             final Labeled labeled = getSkinnable();
             String cleanText = getCleanText();
 
-            if (mnemonicInfo != null && mnemonicInfo.getMnemonicIndex() >= 0) {
-                // If only the graphic is visible, we do not need mnemonic_underscore node
-                if (labeled.getContentDisplay() != ContentDisplay.GRAPHIC_ONLY) {
-                    if (mnemonic_underscore == null) {
-                        mnemonic_underscore = new Line();
-                        mnemonic_underscore.setStartX(0.0f);
-                        mnemonic_underscore.setStartY(0.0f);
-                        mnemonic_underscore.setEndY(0.0f);
-                        mnemonic_underscore.getStyleClass().clear();
-                        mnemonic_underscore.getStyleClass().setAll("mnemonic-underline");
+            if (!com.sun.javafx.PlatformUtil.isMac()) {
+                if (mnemonicInfo != null && mnemonicInfo.getMnemonicIndex() >= 0) {
+                    // If only the graphic is visible, we do not need mnemonic_underscore node
+                    if (labeled.getContentDisplay() != ContentDisplay.GRAPHIC_ONLY) {
+                        if (mnemonic_underscore == null) {
+                            mnemonic_underscore = new Line();
+                            mnemonic_underscore.setStartX(0.0f);
+                            mnemonic_underscore.setStartY(0.0f);
+                            mnemonic_underscore.setEndY(0.0f);
+                            mnemonic_underscore.getStyleClass().clear();
+                            mnemonic_underscore.getStyleClass().setAll("mnemonic-underline");
+                        }
+                        if (!getChildren().contains(mnemonic_underscore)) {
+                            getChildren().add(mnemonic_underscore);
+                        }
                     }
-                    if (!getChildren().contains(mnemonic_underscore)) {
-                        getChildren().add(mnemonic_underscore);
-                    }
+                } else if (mnemonic_underscore != null && getChildren().contains(mnemonic_underscore)) {
+                    getChildren().remove(mnemonic_underscore);
+                    mnemonic_underscore = null;
                 }
-            } else if (mnemonic_underscore != null && getChildren().contains(mnemonic_underscore)) {
-                getChildren().remove(mnemonic_underscore);
-                mnemonic_underscore = null;
             }
 
             int len = cleanText != null ? cleanText.length() : 0;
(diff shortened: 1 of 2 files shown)


andy-goryachev-oracle:
Thank you for logging the bug and offering to fix it!
The reproducer works fine with macOS; I'll test it on windows soon.

andy-goryachev-oracle (approved):
Looks like a good solution.

andy-goryachev-oracle:
as I commented in the ticket, the mnemonic decorations are not visible on macOS so this issue does not occur there.  It still needs to be tested on linux.

@arapte could you be the second reviewer please?

eduardsdv:
It would be nice to have this reviewed. Could you be the second reviewer, @arapte?

modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java
```diff
@@ -548,11 +548,15 @@ protected void layoutLabelInArea(double x, double y, double w, double h, Pos ali
...
+        // This is especially important because it also recomputes the 'containsMnemonic' value,
+        // which is used by this method later on.
+        updateDisplayedText(w, h);
+
         if (ignoreText) {
             textWidth  = textHeight = 0;
             text.setText("");
         } else {
-            updateDisplayedText(w, h); // Have to do this just in case it needs to be recomputed
             textWidth  = snapSizeX(Math.min(text.getLayoutBounds().getWidth(),  wrapWidth));
```
arapte: `ignoreText` would be true in two cases, either 
- when the text is empty / null or 
- `ContentDisplay.GRAPHIC_ONLY` is true. refer [ LabeledSkinBase.isIgnoreText()](https://github.com/openjdk/jfx/blob/3b54913c53f5ac079278a934081d87ce7d55173f/modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java#L1241)

With this change, `updateDisplayedText()` gets invoked unconditionally even for the scenario when the text is non-empty, but needs to be ignored because `ContentDisplay.GRAPHIC_ONLY` is true.

Please check if this can cause any issue.

arapte (changes requested):

modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java
eduardsdv: You are right. 
This is my approach to fix the issue by calling the ``updateDisplayedText()`` regardless of the value of ``ignoreText``. Otherwise, the ``containsMnemonic`` value may be stale.

I added additional tests to check combinations ``ContentDisplay.GRAPHIC_ONLY`` and ``ContentDisplay.TEXT_ONLY`` with empty and null text.

arapte: When `ContentDisplay.GRAPHIC_ONLY` is true, and text is **non-empty**: Calling `updateDisplayedText()` may result in invoking `addMnemonic();` and `getChildren().add(mnemonic_underscore);`. [ refer the method `updateDisplayedText()` , unable to share links due to Github outage ]
This can result in addition of an un-required node to scenegraph. 

The proposed change fixes the exception, but it seems to be introducing a regression.

kevinrushforth: Perhaps another option would be to reset the `containsMnemonic` flag in the `if (ignoreText)` block rather than calling `updateDisplayedText`?

eduardsdv: I also tested a new condition (``!ignoreText``) in the ``layoutLabelInArea()`` method in the following if-clause.
The test are running green and also no additional children are added in case of ``ContentDisplay.GRAPHIC_ONLY``.

Would this be a preferable solution?

````java
Point2D mnemonicPos = null;
double mnemonicWidth = 0.0;
double mnemonicHeight = 0.0;
if (!ignoreText && containsMnemonic) {
    final Font font = text.getFont();
    String preSt = mnemonicInfo.getText();
    boolean isRTL = (labeledNode.getEffectiveNodeOrientation() == NodeOrientation.RIGHT_TO_LEFT);
    mnemonicPos = Utils.computeMnemonicPosition(font, preSt, mnemonicInfo.getMnemonicIndex(), this.wrapWidth, labeled.getLineSpacing(), isRTL);
    mnemonicWidth = Utils.computeTextWidth(font, preSt.substring(mnemonicInfo.getMnemonicIndex(), mnemonicInfo.getMnemonicIndex() + 1), 0);
    mnemonicHeight = Utils.computeTextHeight(font, "_", 0, text.getBoundsType());
}
````

Comment (kevinrushforth) [reply modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java]: That might be OK, but I'll defer to @arapte and @andy-goryachev-oracle

Comment (andy-goryachev-oracle) [reply modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java]: alternatively, to keep the changes to a minimum, we only need to clear `containsMnemonic` in `updateDisplayedText()` in case of `ContentDisplay.GRAPHIC_ONLY` ?

Comment (eduardsdv) [reply modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java]: >  to keep the changes to a minimum

This is the key point.

The StringIndexOutOfBoundsException occurs because the ``mnemonicIndex`` is -1.

Since the ``mnemonicIndex`` is checked for being >= 0 in other places in ``javafx.scene.control.skin.LabeledSkinBase`` and ``com.sun.javafx.scene.control.skin.Util`` classes, I decided to move the ``updateDisplayedText()`` back behind the ``ignoreText`` condition and to add a new ``mnemonicIndex >=0`` condition to the if-clause instead.

This avoids the error and minimizes the risk of regressions.

````java
Point2D mnemonicPos = null;
double mnemonicWidth = 0.0;
double mnemonicHeight = 0.0;
int mnemonicIndex = mnemonicInfo != null ? mnemonicInfo.getMnemonicIndex() : -1;
if (containsMnemonic && mnemonicIndex >= 0) {
    final Font font = text.getFont();
    String preSt = mnemonicInfo.getText();
    boolean isRTL = (labeledNode.getEffectiveNodeOrientation() == NodeOrientation.RIGHT_TO_LEFT);
    mnemonicPos = Utils.computeMnemonicPosition(font, preSt, mnemonicIndex, this.wrapWidth, labeled.getLineSpacing(), isRTL);
    mnemonicWidth = Utils.computeTextWidth(font, preSt.substring(mnemonicIndex, mnemonicIndex + 1), 0);
    mnemonicHeight = Utils.computeTextHeight(font, "_", 0, text.getBoundsType());
}
````

File: modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java [hunk 2]
```diff
@@ -2092,6 +2095,97 @@ public void testEscapedMnemonicSymbolIsNotProcessedWhenParsingIsDisabled() {
...
+        Supplier<Boolean> containsMnemonicNode = () -> label.getChildrenUnmodifiable().stream().anyMatch(p -> p instanceof Line);
+
+        label.setText("foo_bar");
+        tk.firePulse();
+        // Mac does not implement or add mnemonics, skip this test on a Mac
+        if (!com.sun.javafx.PlatformUtil.isMac()) {
+            assertTrue(containsMnemonicNode.get());
+        }
+
+        label.setContentDisplay(ContentDisplay.GRAPHIC_ONLY);
```
andy-goryachev-oracle, line 2142: here we tested mnemonic behavior when content display was set to `GRAPHIC_ONLY`.

would it make sense to set the content display to `TEXT_ONLY` and re-check as a part of this test?

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
```diff
@@ -2092,6 +2095,97 @@ public void testEscapedMnemonicSymbolIsNotProcessedWhenParsingIsDisabled() {
...
+
+        label.setText("xxx");
+        tk.firePulse();
+        assertFalse(containsMnemonicNode.get());
+
+        label.setText("foo_bar");
+        tk.firePulse();
+        assertFalse(containsMnemonicNode.get());
+
+        sl.dispose();
```
andy-goryachev-oracle: suggestion: to make sure the test cleans the environment even when it fails, you could create a `private StageLoader stageLoader` field instead of a local variable, and dispose of it if it's not `null` in `@AfterEach`.

alternatively, you'll need to dispose of it in a `finally` block.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
```diff
@@ -54,6 +56,7 @@
 import com.sun.javafx.tk.Toolkit;
 import test.com.sun.javafx.scene.control.infrastructure.StageLoader;
 
+
```
andy-goryachev-oracle: unnecessary blank line, please remove.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
```diff
@@ -2092,6 +2095,97 @@ public void testEscapedMnemonicSymbolIsNotProcessedWhenParsingIsDisabled() {
...
+        label.setText("xxx");
+        tk.firePulse();
+
+        // text -> mnemonic
+        label.setText("foo_bar");
+        tk.firePulse();
+
+        // mnemonic -> empty
+        label.setText("");
+        tk.firePulse();
```
andy-goryachev-oracle, line 2313: could we also check the mnemonic registration?
the registration should exist before L2171 and be removed after L2172

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
```diff
@@ -2092,6 +2095,97 @@ public void testEscapedMnemonicSymbolIsNotProcessedWhenParsingIsDisabled() {
...
+
+        checkMnemonics();
+    }
+
+    private void checkMnemonics() {
+        Toolkit tk = Toolkit.getToolkit();
+        StageLoader sl = new StageLoader(label);
+        tk.firePulse();
+
+        label.setText("foo_bar");
```
andy-goryachev-oracle, line 2294: I am getting SOOB exception if I do 

`label.setText("test_(t)");`

here

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
eduardsdv: Sure, I've added testing for content display TEXT_ONLY.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
eduardsdv: Done

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
eduardsdv: Done.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
eduardsdv: Good point. I've added the check for the mnemonic registration to the scene.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
eduardsdv: Yes. I got the same error.
I've adjusted the code and the tests to catch this error in the future.

eduardsdv:
The tests now also check that no additional nodes are added to the scene graph when the mnemonics are unavailable. This is the case when:

- Platform is Mac
- MnemonicParsing is false
- ContentDisplay is ``ContentDisplay.GRAPHIC_ONLY``
- text is null, empty or does not contain mnemonic

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
```diff
@@ -80,6 +91,13 @@ public void setup() {
         text = (Text) SkinBaseShim.getChildren(skin).get(0);
     }
 
+    @AfterEach
+    public void teardown() {
+        if (stageLoader != null) {
+            stageLoader.dispose();
```
andy-goryachev-oracle, line 104: `stageLoader = null;`

to avoid using stale stageLoader left over from an earlier run?

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
```diff
@@ -2092,6 +2113,116 @@ public void testEscapedMnemonicSymbolIsNotProcessedWhenParsingIsDisabled() {
...
+        Supplier<Boolean> mnemonicNodePresenceChecker = () -> label.getChildrenUnmodifiable().stream().anyMatch(p -> p instanceof Line);
+
+        Function<Boolean, String> mnemonicErrorMessageSupplier = expected ->
+                (expected
+                        ? "Mnemonic registration expected but not found: "
+                        : "No mnemonic expected but was: ")
+                        + mnemonics;
+
+        // Mac does not support mnemonics
+        boolean expectedMnemonics = PlatformUtil.isMac() ? false : label.isMnemonicParsing();
```
andy-goryachev-oracle, line 2282: could we also check whether mnemonic is actually registered with the scene, using  `label.getScene().getMnemonic()`?

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
```diff
@@ -2126,6 +2257,13 @@ public void testBaselineAlignmentWhenTextIsSetAndGraphicIsSet() {
         sl.dispose();
     }
 
+    private StageLoader createStageLoader(Node node) {
+        if (stageLoader == null) {
```
andy-goryachev-oracle: this works only if `stageLoader` is null'ed in `@AfterEach`

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
```diff
@@ -2092,6 +2113,116 @@ public void testEscapedMnemonicSymbolIsNotProcessedWhenParsingIsDisabled() {
...
+        label.setMnemonicParsing(true);
+        checkMnemonics();
+
+        label.setMnemonicParsing(false);
+        checkMnemonics();
+
+        label.setMnemonicParsing(true);
+        checkMnemonics();
+    }
+
```
andy-goryachev-oracle, line 2134: I was trying to check whether mnemonic is registered with the scene, but this test fails with NPE before that:

```
    @Test
    public void withGraphic() {
        assumeFalse(PlatformUtil.isMac());
        label.setText("_test");
        label.setMnemonicParsing(true);
        label.setGraphic(new Label());
        label.setContentDisplay(ContentDisplay.GRAPHIC_ONLY);
        stageLoader = new StageLoader(label);
        Toolkit.getToolkit().firePulse(); // NPE
```

andy-goryachev-oracle (changes requested):
What looked like a simple change proved to be exceedingly tricky...  I wonder if we are hitting pre-existing design issues.

For example, what I might have done is to eagerly discard mnemonicInfo when any properties that affect it change, and re-create it each time it's needed (making sure not to lose sync with the scene mnemonics).  The fact that we have SOOBEs and NPEs indicates that we are using stale values, so perhaps we need to make sure we re-compute them correctly.

What do you think?

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
eduardsdv: As far as I know, it is not necessary, because the Junit creates a new instance of the test class for each test method call.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
eduardsdv: Done.
The scene is now always accessed via the ``label``.
See the ``label.getScene().getMnemonics()`` in ``checkMnemonics()`` method.

````java
Function<String, Boolean> mnemonicRegistrationChecker = character ->
        label.getScene().getMnemonics()
....
````

File: modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java (hunk 9 above)
Comment (eduardsdv) [reply modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java]: I've removed this method.

File: modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java (hunk 10 above)
Comment (eduardsdv) [reply modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java:2134]: Interesting. I've added extra tests that permutate properties involved into mnemonic processing and I got the error too.

I've reworked some methods and removed obsolete flags.

Some cases were not handled properly:
- the ``getCleanText()`` method did not update the ``MnemonicInfo`` if the text was null
- presence of two methods to add/remove mnemonic registration. These methods were called from different places, but not in pairs. This led to multiple mnemonic registrations. I replaced them by ``updateMnemonicRegistration()``. It concentrates all the knowledge, whether the mnemonic is already added, needs to be replaced or removed.
- multiple flags made it difficult to oversee the state and the correct value combinations. The 'MnemonicInfo' instance is now the only source of truth.

Now all tests running green again.

File: modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java [hunk 11]
```diff
@@ -1186,33 +1106,55 @@ private String getCleanText() {
...
     }
 
-    private void addMnemonic() {
-        if (labeledNode != null) {
-            mnemonicScene = labeledNode.getScene();
-            if (mnemonicScene != null) {
-                mnemonicScene.addMnemonic(new Mnemonic(labeledNode, mnemonicCode));
-            }
+    private void updateMnemonicRegistration() {
+        if (com.sun.javafx.PlatformUtil.isMac()) {
```
andy-goryachev-oracle, line 1124: please use the import statement instead of the FQCN

modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java
```diff
@@ -1186,33 +1106,55 @@ private String getCleanText() {
...
+            return;
+        }
+        Control skinnable = getSkinnable();
+        Scene skinnableScene = skinnable.getScene();
+        Node mnemonicNode = skinnable;
+        if (skinnable instanceof Label l && l.getLabelFor() != null) {
+            mnemonicNode =  l.getLabelFor();
```
andy-goryachev-oracle, line 1131: extra space

modules/javafx.controls/src/main/java/javafx/scene/control/skin/LabeledSkinBase.java
```diff
@@ -1186,33 +1106,55 @@ private String getCleanText() {
...
-                mnemonicScene.addMnemonic(new Mnemonic(labeledNode, mnemonicCode));
-            }
+    private void updateMnemonicRegistration() {
+        if (com.sun.javafx.PlatformUtil.isMac()) {
+            return;
+        }
+        Control skinnable = getSkinnable();
+        Scene skinnableScene = skinnable.getScene();
+        Node mnemonicNode = skinnable;
+        if (skinnable instanceof Label l && l.getLabelFor() != null) {
```
andy-goryachev-oracle, line 1130: minor: might want to use a name other than l as it is similar to 1 or |

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
```diff
@@ -2142,25 +2149,134 @@ public void testMnemonics_setContentDisplay() {
...
+    @Test
+    void testMnemonics_with_same_Labeled_and_Scene() {
+        StackPane root = new StackPane();
+        root.getChildren().add(label);
+        stageLoader = new StageLoader(root);
+
+        StringBuilder sb = new StringBuilder();
+
+        provide_testMnemonics_parameters().forEach(a -> {
```
andy-goryachev-oracle, line 2189: could this be converted to a simple parameterized test?
or the sequence is important?

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
```diff
@@ -2142,25 +2149,134 @@ public void testMnemonics_setContentDisplay() {
...
+                        }
+                    }
+                }
+            }
+        }
+
+        return arguments.stream();
+    }
+
+    private static <T> List<T> listOf(T... a) {
```
andy-goryachev-oracle, line 2252: isn't this a standard `List.of()`?

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
andy-goryachev-oracle: This may or may not be true, as it depends on junit5 configuration
```
junit.jupiter.testinstance.lifecycle.default = per_class
```
there is no harm in adhering to good coding practices

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
```diff
@@ -2092,6 +2120,225 @@ public void testEscapedMnemonicSymbolIsNotProcessedWhenParsingIsDisabled() {
...
+        label.setText("foo_bar");
+        tk.firePulse();
+        assertEquals(expectedMnemonics, mnemonicRegistrationChecker.apply("b"), mnemonicErrorMessageSupplier.apply(expectedMnemonics));
+        assertEquals(expectedMnemonicsNode, mnemonicNodePresenceChecker.get());
+
+        // extended mnemonic
+        label.setText("test_(t)");
+        tk.firePulse();
+        assertEquals(expectedMnemonics, mnemonicRegistrationChecker.apply("t"), mnemonicErrorMessageSupplier.apply(expectedMnemonics));
+        assertEquals(expectedMnemonicsNode, mnemonicNodePresenceChecker.get());
```
andy-goryachev-oracle, line 2339: I think we still might have a problem.  "Extended mnemonic" which is not publicly specified anywhere (which is another issue which I just filed https://bugs.openjdk.org/browse/JDK-8390898 ) means that the underline should be removed since the corresponding symbol may not be present.

We'll have failures if we add 
```
assertFalse(mnemonicNodePresenceChecker.get());
```
everywhere the extended mnemonic is used.

andy-goryachev-oracle (commented):
I think you are getting close!  This is a complicated subject.

I've filed https://bugs.openjdk.org/browse/JDK-8390898 to make sure the extended mnemonic behavior is properly specified.

There is also an existing ticket

[JDK-8336590](https://bugs.openjdk.org/browse/JDK-8336590)
Mnemonic Rendering incorrect (different) if Button has a graphic

can you check if your fix solves that issue please?

andy-goryachev-oracle:
Also, probably in a follow-up, we might want to check whether the `MenuItem` class needs similar fixes applied (especially related to mnemonic registration).

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
Maran23: I think we usually do not `null` the `stageLoader`.
In any case I have some test improvements planned for this after JUnit 6 is merged

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
Maran23: Yes, should be used instead

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
```diff
@@ -2092,6 +2120,225 @@ public void testEscapedMnemonicSymbolIsNotProcessedWhenParsingIsDisabled() {
...
+
+    private static <T> List<T> listOf(T... a) {
+        List<T> result = new ArrayList<>();
+        for (T o : a) {
+            result.add(o);
+        }
+        return result;
+    }
+
+    private void checkMnemonics() {
```
Maran23, line 2260: Minor, how about we name this `verifyMnemonics` to make clear we do JUnit verifications here.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
```diff
@@ -2092,6 +2120,225 @@ public void testEscapedMnemonicSymbolIsNotProcessedWhenParsingIsDisabled() {
...
+        Toolkit tk = Toolkit.getToolkit();
+        if (stageLoader == null) {
+            stageLoader = new StageLoader(label);
+        }
```
Maran23, line 2264: I would recommend to do that once in the test as a first step and remove it from here. Will make the separation of concerns much easier.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
```diff
@@ -2092,6 +2120,225 @@ public void testEscapedMnemonicSymbolIsNotProcessedWhenParsingIsDisabled() {
...
+            label.setLabelFor(labelFor);
+
+            sb.append(mnemonicParsing + ", " + contentDisplay + ", " + text + ", " + graphic + ", " + labelFor).append("\n");
+            try {
+                checkMnemonics();
+            } catch (Throwable e) {
+                // This makes it easier to identify the failed step
+                System.out.println(sb.toString());
+                throw e;
+            }
```
Maran23, line 2217: Usually a `catch` is not needed, JUnit will show the error and the stacktrace. This is usually rather an anti-pattern.

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
Maran23: Same question, should be a `@MethodSource("provide_testMnemonics_parameters")` as well

modules/javafx.controls/src/test/java/test/javafx/scene/control/skin/LabelSkinTest.java
```diff
@@ -2092,6 +2120,225 @@ public void testEscapedMnemonicSymbolIsNotProcessedWhenParsingIsDisabled() {
...
+
+    @Test
+    public void withGraphic() {
+        Assumptions.assumeFalse(PlatformUtil.isMac());
+        label.setText("_test");
+        label.setMnemonicParsing(true);
+        label.setGraphic(new Label());
+        label.setContentDisplay(ContentDisplay.GRAPHIC_ONLY);
+        stageLoader = new StageLoader(label);
+        Toolkit.getToolkit().firePulse(); // No NPE
```
Maran23, line 2160:
```suggestion
        assertDoesNotThrow(() -> Toolkit.getToolkit().firePulse());
```

PR #2224, opened 2026-07-27
8388276: RichTextArea: image import

andy-goryachev-oracle:
Enables importing of images from the clipboard into the RichTextModel by adding a data format handler for the DataFormat.IMAGE mime type.

This change includes new utility added by [JDK-8390345](https://bugs.openjdk.org/browse/JDK-8390345).

Note: the banana slug image is a public domain added with author's permission [0].

## References

[0] https://en.wikipedia.org/wiki/Banana_slug

/csr

---------
- [x] I confirm that I make this contribution in accordance with the [OpenJDK Interim AI Policy](https://openjdk.org/legal/ai).

---------
### Progress
- \[x\] Change must not contain extraneous whitespace
- \[x\] Commit message must refer to an issue
- \[ \] Change requires CSR request [JDK-8389549](https://bugs.openjdk.org/browse/JDK-8389549) to be approved
- \[ \] Change must be properly reviewed (2 reviews required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer), 1 [Author](https://openjdk.org/bylaws#author))

### Warnings
&nbsp;⚠️ Patch contains a binary file (tests/system/src/test/resources/test/com/sun/javafx/images/banana-slug.jpg)
&nbsp;⚠️ Patch contains a binary file (tests/system/src/test/resources/test/com/sun/javafx/images/underlines.png)

### Issues
 * [JDK-8388276](https://bugs.openjdk.org/browse/JDK-8388276): RichTextArea: image import (**Enhancement** - P4)
 * [JDK-8389549](https://bugs.openjdk.org/browse/JDK-8389549): RichTextArea: image import (**CSR**)

### Reviewers
 * [Ajit Ghaisas](https://openjdk.org/census#aghaisas) (@aghaisas - **Reviewer**)

### Reviewing
<details><summary>Using <code>git</code></summary>

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2224/head:pull/2224` \
`$ git checkout pull/2224`

Update a local copy of the PR: \
`$ git checkout pull/2224` \
`$ git pull https://git.openjdk.org/jfx.git pull/2224/head`

</details>
<details><summary>Using Skara CLI tools</summary>

Checkout this PR locally: \
`$ git pr checkout 2224`

View PR using the GUI difftool: \
`$ git pr show -t 2224`

</details>
<details><summary>Using diff file</summary>

Download this PR as a diff file: \
[...]

PR #2254, opened 2026-08-10
8388450: ImageIO.write(SwingFXUtils.fromFXImage()) creates 0 length JPEG for some images

prsadhuk:
If an image has RGBA encoding and all pixels are opaque, then `SwingFXUtils.fromFXImage(image, null)` selects TYPE_INT_ARGB_PRE pixel format but the JPEG ImageIO writer cannot encode this alpha-bearing image pixel format since it has no support for it and returns false; so an empty byte array is created.

Issue is when `bimg` (used to store the returned pixel data from `fromFXImage`) is null`,  fromFXImage` uses the JavaFX PixelReader’s storage format, not the alpha values of individual pixels. 
For an RGBA PNG, JavaFX commonly decodes the pixels into premultiplied ARGB/BGRA, so `fxFormat.getType()` is one of INT_ARGB_PRE/BYTE_BGRA_PRE so `getBestBufferedImageType` returns `BufferedImage.TYPE_INT_ARGB_PRE` because the pixel format has an alpha component. It does not inspect whether every alpha value happens to be 255.

The existing `fromFXImage` code only calls `checkFXImageOpaque()` when the caller supplies a non-null `bImg` BufferedImage. 
If `bImg `is null, all-opaque RGBA PNG produces an alpha-capable BufferedImage i.e., BufferedImage.TYPE_INT_ARGB_PRE
```
if (bimg == null) {
    bimg = new BufferedImage(iw, ih, prefBimgType);
}
```
and since JPEG has no standard alpha channel so it cannot store transparency, so when `ImageIO.write(image, "jpg", out)` runs, ImageIO looks for a registered JPEG writer which can encode that pre-multiplied-alpha image type,
 but the JPEG writer rejects an alpha-bearing BufferedImage, so `ImageIO.write` finds no suitable writer and returns false 
so OutputStream is not written into and have 0 bytes 
[Basically the JPEG writer does not check whether the alpha values are all 255; it only sees that the input image has an alpha channel and declines to write it]

The proposed JavaFX change avoids the rejection for an RGBA-formatted but fully opaque image by returning TYPE_INT_RGB, which JPEG can encode. 
ie., for a JavaFX image with an alpha-capable format but only opaque pixels, `fromFXImage(image, null)` is made to choose RGB format rather than ARGB.
Additionally, checkFXImageOpaque is improved to scan one row at a time instead of costly full-image scan so that unnecessary Color object for each pixels is not created just to inspect alpha.

A regression subtest is added to existing testcase

---------

---------
### Progress

### Issue
 * [JDK-8388450](https://bugs.openjdk.org/browse/JDK-8388450): ImageIO.write(SwingFXUtils.fromFXImage()) creates 0 length JPEG for some images (**Bug** - P3)

### Reviewers
 * [Alexander Zvegintsev](https://openjdk.org/census#azvegint) (@azvegint - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2254/head:pull/2254` \
`$ git checkout pull/2254`

`$ git checkout pull/2254` \
`$ git pull https://git.openjdk.org/jfx.git pull/2254/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2254`

`$ git pr show -t 2254`

</details>

<a href="https://git.openjdk.org/jfx/pull/2254.diff">https://git.openjdk.org/jfx/pull/2254.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2254#issuecomment-5238075648)
</details>

### DIFF
--- a/modules/javafx.swing/src/main/java/javafx/embed/swing/SwingFXUtils.java
+++ b/modules/javafx.swing/src/main/java/javafx/embed/swing/SwingFXUtils.java
@@ -205,8 +205,7 @@ public class SwingFXUtils {
 
     private static boolean checkFXImageOpaque(PixelReader pr, int iw, int ih) {
         int[] pixels = new int[iw];
-        WritablePixelFormat<IntBuffer> format =
-                PixelFormat.getIntArgbPreInstance();
+        WritablePixelFormat<IntBuffer> format = PixelFormat.getIntArgbPreInstance();
         for (int y = 0; y < ih; y++) {
             pr.getPixels(0, y, iw, 1, format, pixels, 0, iw);
             for (int pixel : pixels) {
@@ -256,14 +255,15 @@ public class SwingFXUtils {
         int ih = (int) img.getHeight();
         PixelFormat<?> fxFormat = pr.getPixelFormat();
         boolean srcPixelsAreOpaque = false;
-        boolean opacityMatters = bimg == null ||
-                bimg.getType() == BufferedImage.TYPE_INT_BGR ||
-                bimg.getType() == BufferedImage.TYPE_INT_RGB;
         switch (fxFormat.getType()) {
             case INT_ARGB_PRE:
             case INT_ARGB:
             case BYTE_BGRA_PRE:
+            case BYTE_BGRA:
             case BYTE_INDEXED:
+                boolean opacityMatters = bimg == null ||
+                        bimg.getType() == BufferedImage.TYPE_INT_BGR ||
+                        bimg.getType() == BufferedImage.TYPE_INT_RGB;
                 if (opacityMatters) {
                     srcPixelsAreOpaque = checkFXImageOpaque(pr, iw, ih);
                 }
--- a/modules/javafx.swing/src/main/java/javafx/embed/swing/SwingFXUtils.java
+++ b/modules/javafx.swing/src/main/java/javafx/embed/swing/SwingFXUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2012, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2012, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -156,6 +156,9 @@ public class SwingFXUtils {
                 return bimgType;
             }
         }
+        if (isOpaque) {
+            return BufferedImage.TYPE_INT_RGB;
+        }
         switch (fxFormat.getType()) {
             default:
             case BYTE_BGRA_PRE:
@@ -201,10 +204,13 @@ public class SwingFXUtils {
     }
 
     private static boolean checkFXImageOpaque(PixelReader pr, int iw, int ih) {
-        for (int x = 0; x < iw; x++) {
-            for (int y = 0; y < ih; y++) {
-                Color color = pr.getColor(x,y);
-                if (color.getOpacity() != 1.0) {
+        int[] pixels = new int[iw];
+        WritablePixelFormat<IntBuffer> format =
+                PixelFormat.getIntArgbPreInstance();
+        for (int y = 0; y < ih; y++) {
+            pr.getPixels(0, y, iw, 1, format, pixels, 0, iw);
+            for (int pixel : pixels) {
+                if ((pixel >>> 24) != 0xff) {
                     return false;
                 }
             }
@@ -250,16 +256,15 @@ public class SwingFXUtils {
         int ih = (int) img.getHeight();
         PixelFormat<?> fxFormat = pr.getPixelFormat();
         boolean srcPixelsAreOpaque = false;
+        boolean opacityMatters = bimg == null ||
+                bimg.getType() == BufferedImage.TYPE_INT_BGR ||
+                bimg.getType() == BufferedImage.TYPE_INT_RGB;
         switch (fxFormat.getType()) {
             case INT_ARGB_PRE:
             case INT_ARGB:
             case BYTE_BGRA_PRE:
-            case BYTE_BGRA:
-                // Check fx image opacity only if
-                // supplied BufferedImage is without alpha channel
-                if (bimg != null &&
-                        (bimg.getType() == BufferedImage.TYPE_INT_BGR ||
-                         bimg.getType() == BufferedImage.TYPE_INT_RGB)) {
+            case BYTE_INDEXED:
+                if (opacityMatters) {
                     srcPixelsAreOpaque = checkFXImageOpaque(pr, iw, ih);
                 }
                 break;
--- a/tests/system/src/test/java/test/javafx/embed/swing/SwingFXUtilsTest.java
+++ b/tests/system/src/test/java/test/javafx/embed/swing/SwingFXUtilsTest.java
@@ -28,7 +28,9 @@ package test.javafx.embed.swing;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import java.io.ByteArrayOutputStream;
 import java.awt.image.BufferedImage;
+import javax.imageio.ImageIO;
 import java.util.concurrent.CountDownLatch;
 import javafx.application.Application;
 import javafx.application.Platform;
@@ -36,6 +38,7 @@ import javafx.embed.swing.SwingFXUtils;
 import javafx.scene.image.Image;
 import javafx.scene.image.PixelFormat;
 import javafx.scene.image.PixelReader;
+import javafx.scene.image.WritableImage;
 import javafx.stage.Stage;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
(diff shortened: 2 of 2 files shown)


modules/javafx.swing/src/main/java/javafx/embed/swing/SwingFXUtils.java
```diff
@@ -201,10 +204,13 @@ public static WritableImage toFXImage(BufferedImage bimg, WritableImage wimg) {
...
-            for (int y = 0; y < ih; y++) {
-                Color color = pr.getColor(x,y);
-                if (color.getOpacity() != 1.0) {
+        int[] pixels = new int[iw];
+        WritablePixelFormat<IntBuffer> format =
+                PixelFormat.getIntArgbPreInstance();
+        for (int y = 0; y < ih; y++) {
+            pr.getPixels(0, y, iw, 1, format, pixels, 0, iw);
+            for (int pixel : pixels) {
+                if ((pixel >>> 24) != 0xff) {
```
andy-goryachev-oracle, line 212: I would have done
`((pixel & 0xff000000) != 0xff000000)`

but I think there is no difference in performance whatsoever

modules/javafx.swing/src/main/java/javafx/embed/swing/SwingFXUtils.java
```diff
@@ -201,10 +204,13 @@ public static WritableImage toFXImage(BufferedImage bimg, WritableImage wimg) {
     }
 
     private static boolean checkFXImageOpaque(PixelReader pr, int iw, int ih) {
-        for (int x = 0; x < iw; x++) {
-            for (int y = 0; y < ih; y++) {
-                Color color = pr.getColor(x,y);
-                if (color.getOpacity() != 1.0) {
+        int[] pixels = new int[iw];
+        WritablePixelFormat<IntBuffer> format =
+                PixelFormat.getIntArgbPreInstance();
```
andy-goryachev-oracle: just curious: why is this line broken?  it fits in 120 columns just fine.  time to update the formatting rules?

modules/javafx.swing/src/main/java/javafx/embed/swing/SwingFXUtils.java
```diff
@@ -250,16 +256,15 @@ public static BufferedImage fromFXImage(Image img, BufferedImage bimg) {
         int ih = (int) img.getHeight();
         PixelFormat<?> fxFormat = pr.getPixelFormat();
         boolean srcPixelsAreOpaque = false;
+        boolean opacityMatters = bimg == null ||
```
andy-goryachev-oracle: minor: this is calculated even when it's not needed.  could it be moved to L267?

modules/javafx.swing/src/main/java/javafx/embed/swing/SwingFXUtils.java
```diff
@@ -250,16 +256,15 @@ public static BufferedImage fromFXImage(Image img, BufferedImage bimg) {
...
             case INT_ARGB_PRE:
             case INT_ARGB:
             case BYTE_BGRA_PRE:
-            case BYTE_BGRA:
-                // Check fx image opacity only if
-                // supplied BufferedImage is without alpha channel
-                if (bimg != null &&
-                        (bimg.getType() == BufferedImage.TYPE_INT_BGR ||
-                         bimg.getType() == BufferedImage.TYPE_INT_RGB)) {
+            case BYTE_INDEXED:
```
andy-goryachev-oracle, line 263: question: this switch statement is missing `BYTE_BGRA`.  is this a problem?

tests/system/src/test/java/test/javafx/embed/swing/SwingFXUtilsTest.java
andy-goryachev-oracle, line 89: would it make sense to iterate over every `PixelFormat.Type` using `WritableImage(PixelBuffer)` constructor  to make sure we are getting a meaningful result in each case?

modules/javafx.swing/src/main/java/javafx/embed/swing/SwingFXUtils.java
prsadhuk: then let it remain same :-)

modules/javafx.swing/src/main/java/javafx/embed/swing/SwingFXUtils.java
prsadhuk: ok..modified

modules/javafx.swing/src/main/java/javafx/embed/swing/SwingFXUtils.java
prsadhuk: yes, missed...added

tests/system/src/test/java/test/javafx/embed/swing/SwingFXUtilsTest.java
prsadhuk: I guess PixelBuffer supports only INT_ARGB_PRE and BYTE_BGRA_PRE

> Pixel data should be stored either in an [IntBuffer](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/IntBuffer.html) using a [PixelFormat](https://openjfx.io/javadoc/21/javafx.graphics/javafx/scene/image/PixelFormat.html) of type INT_ARGB_PRE or in a [ByteBuffer](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/ByteBuffer.html) using a [PixelFormat](https://openjfx.io/javadoc/21/javafx.graphics/javafx/scene/image/PixelFormat.html) of type BYTE_BGRA_PRE.

INT_ARGB_PRE is already being tested..there's not much test coverage to iterate so I guess let it  
stay small and target the reported behavior

modules/javafx.swing/src/main/java/javafx/embed/swing/SwingFXUtils.java
prsadhuk: updated..it was just to keep line consistent with the javadoc beneath

andy-goryachev-oracle:
Something is wrong: the reproducer fails to render the PNG image attached to the ticket,
`ImageIO.getWriter()` L1558 returns an empty iterator, so `ImageIO.write()` fails

```
    private static byte[] writeImage(Image im, String format) throws IOException {
        ByteArrayOutputStream out = new ByteArrayOutputStream(65536);
        try {
            // using disk cache slows things down
            boolean old = ImageIO.getUseCache();
            ImageIO.setUseCache(false);
            try {
                var bi = ImgUtil.fromFXImage(im, null);
                ImageIO.write(bi, format, out);
            } finally {
                ImageIO.setUseCache(old);
            }
        } finally {
            out.close();
        }
        return out.toByteArray();
    }
```

prsadhuk:
> Something is wrong: the reproducer fails to render the PNG image attached to the ticket, `ImageIO.getWriter()` L1558 returns an empty iterator, so `ImageIO.write()` fails
> 
> ```
>     private static byte[] writeImage(Image im, String format) throws IOException {
>         ByteArrayOutputStream out = new ByteArrayOutputStream(65536);
>         try {
>             // using disk cache slows things down
>             boolean old = ImageIO.getUseCache();
>             ImageIO.setUseCache(false);
>             try {
>                 var bi = ImgUtil.fromFXImage(im, null);
>                 ImageIO.write(bi, format, out);
>             } finally {
>                 ImageIO.setUseCache(old);
>             }
>         } finally {
>             out.close();
>         }
>         return out.toByteArray();
>     }
> ```

I guess it will fail as ImgUtil class which is copied implementation of SwingFXUtils is not having this PR fix..
If I change ImageWriteTest_8388450.java to call SwingFXUtils.fromFXImage it passes

./jdk/bin/java @C:/Users/Prasantas/dev/javafx/jfx/rt/build/run.args ImageWriteTest_8388450.java
 OK eclipse-key-mappings.png test

 ```
private static byte[] writeImage(Image im, String format) throws IOException {
        ByteArrayOutputStream out = new ByteArrayOutputStream(65536);
        try {
            // using disk cache slows things down
            boolean old = ImageIO.getUseCache();
            ImageIO.setUseCache(false);
            try {
                var bi = SwingFXUtils.fromFXImage(im, null);
                ImageIO.write(bi, format, out);
            } finally {
                ImageIO.setUseCache(old);
            }
        } finally {
            out.close();
        }
        return out.toByteArray();
    }
```

andy-goryachev-oracle:
> ImgUtil class which is copied implementation of SwingFXUtils is not having this PR fix.

you are right, I am sorry.

this also means the `RichTextArea` incubator module should use `SwingFXUtils` directly, adding dependency on `javafx.swing`.

andy-goryachev-oracle (approved):

kevinrushforth:
> > ImgUtil class which is copied implementation of SwingFXUtils is not having this PR fix.
> 
> you are right, I am sorry.
> 
> this also means the `RichTextArea` incubator module should use `SwingFXUtils` directly, adding dependency on `javafx.swing`.

Hmm. Maybe as a stop-gap, but it would not be acceptable for any `javafx.*` modules to depend on `javafx.swing`, so this would have to be replaced before we could ever finalize RichTextArea.

andy-goryachev-oracle:
> it would not be acceptable for any javafx.* modules to depend on javafx.swing

What can we do?

We could provide a utility in `javafx.graphics` (?) to read/write a subset of formats natively (PNG/JPG should be sufficient) that works directly with `Image` instead of `BufferedImage` and does not involve `ImageIO`.

Or we can keep duplicating the code and potentially risk tripping over the same situation in the future.

andy-goryachev-oracle:
or we could port ImageIO parts to work with JPG/PNG in javafx.

andy-goryachev-oracle:
Applied these changes to `ImgUtil` in #2224.

andy-goryachev-oracle:

andy-goryachev-oracle:
I've filed https://bugs.openjdk.org/browse/JDK-8390345 to avoid duplicating image i/o code.

azvegint (approved):

PR #2260, opened 2026-08-15
8390433: Add pixel-snapping documentation

mstr2:
Pixel snapping is really hard to get right (in fact, it's so hard that even JavaFX itself gets it wrong in so many places).
I've compiled a list of things that I've learned, because there isn't really any good documentation as of yet.

---------

---------
### Progress

### Issue
 * [JDK-8390433](https://bugs.openjdk.org/browse/JDK-8390433): Add pixel-snapping documentation (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jfx.git pull/2260/head:pull/2260` \
`$ git checkout pull/2260`

`$ git checkout pull/2260` \
`$ git pull https://git.openjdk.org/jfx.git pull/2260/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 2260`

`$ git pr show -t 2260`

</details>

<a href="https://git.openjdk.org/jfx/pull/2260.diff">https://git.openjdk.org/jfx/pull/2260.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jfx/pull/2260#issuecomment-5303675933)
</details>

### DIFF
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
@@ -1146,7 +1146,6 @@ public abstract non-sealed class Parent extends Node {
         }
     }
 
-    // PENDING_DOC_REVIEW
     /**
      * Computes the preferred width for the specified height.
      * <p>
@@ -1176,7 +1175,6 @@ public abstract non-sealed class Parent extends Node {
         return maxX - minX;
     }
 
-    // PENDING_DOC_REVIEW
     /**
      * Computes the preferred height for the specified width.
      * <p>
@@ -1223,7 +1221,6 @@ public abstract non-sealed class Parent extends Node {
         return prefWidth(height);
     }
 
-    // PENDING_DOC_REVIEW
     /**
      * Computes the minimum height for the specified width.
      * <p>
--- a/modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
@@ -1148,14 +1148,19 @@ public abstract non-sealed class Parent extends Node {
 
     // PENDING_DOC_REVIEW
     /**
-     * Calculates the preferred width of this {@code Parent}. The default
-     * implementation calculates this width as the width of the area occupied
-     * by its managed children when they are positioned at their
-     * current positions at their preferred widths.
+     * Computes the preferred width for the specified height.
+     * <p>
+     * An overriding implementation should be consistent with the layout policy implemented by
+     * {@link #layoutChildren()}. If pixel snapping is used, corresponding measurement and layout
+     * calculations must use the same snapped values.
      *
-     * @param height the height that should be used if preferred width depends
-     *      on it
-     * @return the calculated preferred width
+     * @implNote The default implementation computes the horizontal span containing the parent's origin
+     *           and all managed children. It uses each child's current layout position and preferred width,
+     *           constrained by the child's minimum and maximum widths.
+     *           The {@code height} parameter is ignored.
+     * @param height the height on which to base the preferred width, or {@code -1} if no height is specified
+     * @return the computed preferred width
+     * @see <a href="layout/package-summary.html#pixel-snapping">Pixel Snapping</a>
      */
     protected double computePrefWidth(double height) {
         double minX = 0;
@@ -1173,14 +1178,19 @@ public abstract non-sealed class Parent extends Node {
 
     // PENDING_DOC_REVIEW
     /**
-     * Calculates the preferred height of this {@code Parent}. The default
-     * implementation calculates this height as the height of the area occupied
-     * by its managed children when they are positioned at their current
-     * positions at their preferred heights.
+     * Computes the preferred height for the specified width.
+     * <p>
+     * An overriding implementation should be consistent with the layout policy implemented by
+     * {@link #layoutChildren()}. If pixel snapping is used, corresponding measurement and layout
+     * calculations must use the same snapped values.
      *
-     * @param width the width that should be used if preferred height depends
-     *      on it
-     * @return the calculated preferred height
+     * @implNote The default implementation computes the vertical span containing the parent's origin
+     *           and all managed children. It uses each child's current layout position and preferred height,
+     *           constrained by the child's minimum and maximum heights.
+     *           The {@code width} parameter is ignored.
+     * @param width the width on which to base the preferred height, or {@code -1} if no width is specified
+     * @return the computed preferred height
+     * @see <a href="layout/package-summary.html#pixel-snapping">Pixel Snapping</a>
      */
     protected double computePrefHeight(double width) {
         double minY = 0;
@@ -1197,12 +1207,16 @@ public abstract non-sealed class Parent extends Node {
     }
 
     /**
-     * Calculates the minimum width of this {@code Parent}. The default
-     * implementation simply returns the pref width.
+     * Computes the minimum width for the specified height.
+     * <p>
+     * An overriding implementation should be consistent with the layout policy implemented by
+     * {@link #layoutChildren()}. If pixel snapping is used, corresponding measurement and layout
+     * calculations must use the same snapped values.
      *
-     * @param height the height that should be used if min width depends
-     *      on it
-     * @return the calculated min width
+     * @implNote The default implementation returns {@link #prefWidth(double) prefWidth(height)}.
+     * @param height the height on which to base the minimum width, or {@code -1} if no height is specified
+     * @return the computed minimum width
+     * @see <a href="layout/package-summary.html#pixel-snapping">Pixel Snapping</a>
      * @since JavaFX 2.1
      */
     protected double computeMinWidth(double height) {
@@ -1211,12 +1225,16 @@ public abstract non-sealed class Parent extends Node {
 
     // PENDING_DOC_REVIEW
     /**
-     * Calculates the min height of this {@code Parent}. The default
-     * implementation simply returns the pref height;
+     * Computes the minimum height for the specified width.
+     * <p>
+     * An overriding implementation should be consistent with the layout policy implemented by
+     * {@link #layoutChildren()}. If pixel snapping is used, corresponding measurement and layout
+     * calculations must use the same snapped values.
      *
-     * @param width the width that should be used if min height depends
-     *      on it
-     * @return the calculated min height
+     * @implNote The default implementation returns {@link #prefHeight(double) prefHeight(width)}.
+     * @param width the width on which to base the minimum height, or {@code -1} if no width is specified
+     * @return the computed minimum height
+     * @see <a href="layout/package-summary.html#pixel-snapping">Pixel Snapping</a>
      * @since JavaFX 2.1
      */
     protected double computeMinHeight(double width) {
@@ -1299,12 +1317,16 @@ public abstract non-sealed class Parent extends Node {
     }
 
     /**
-     * Invoked during the layout pass to layout the children in this
-     * {@code Parent}. By default it will only set the size of managed,
-     * resizable content to their preferred sizes and does not do any node
-     * positioning.
+     * Invoked by {@link #layout()} during a layout pass to position and resize the managed children.
      * <p>
-     * Subclasses should override this function to layout content as needed.
+     * Subclasses should override this method to implement their layout policy. When pixel snapping is used,
+     * this parent owns the snapping policy for the positions and sizes it allocates to its children.
+     * The snapped allocations made here must be consistent with the corresponding minimum-size and
+     * preferred-size calculations.
+     *
+     * @implSpec The default implementation invokes {@link Node#autosize()} on each managed,
+     *           resizable child and leaves all child positions unchanged.
+     * @see <a href="layout/package-summary.html#pixel-snapping">Pixel Snapping</a>
      */
     protected void layoutChildren() {
         for (int i=0, max=children.size(); i<max; i++) {
--- a/modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
@@ -1591,41 +1591,61 @@ public class Region extends Parent {
     }
 
     /**
-     * Computes the minimum width of this region.
-     * Returns the sum of the left and right insets by default.
-     * region subclasses should override this method to return an appropriate
-     * value based on their content and layout strategy.  If the subclass
-     * doesn't have a VERTICAL content bias, then the height parameter can be
-     * ignored.
+     * Computes the minimum width of this region for the specified height.
+     * <p>
+     * This method supplies the value used by {@link #minWidth(double)} when the {@link #minWidth}
+     * property is set to {@link #USE_COMPUTED_SIZE}.
+     * <p>
+     * Subclasses should override this method when their content or layout policy requires a different
+     * minimum width. If {@link #getContentBias()} is {@link Orientation#VERTICAL}, the computation should
+     * use {@code height}; otherwise {@code height} can be ignored. An overriding implementation should be
+     * consistent with its {@link #layoutChildren()} implementation, including its pixel-snapping decisions.
      *
-     * @return the computed minimum width of this region
+     * @implNote The default implementation returns the sum of the left and right {@linkplain #getInsets() insets}.
+     * @param height the height on which to base the minimum width, or {@code -1} if no height is specified
+     * @return the computed minimum width
+     * @see <a href="package-summary.html#pixel-snapping">Pixel Snapping</a>
      */
     @Override protected double computeMinWidth(double height) {
         return getInsets().getLeft() + getInsets().getRight();
     }
 
     /**
-     * Computes the minimum height of this region.
-     * Returns the sum of the top and bottom insets by default.
-     * Region subclasses should override this method to return an appropriate
-     * value based on their content and layout strategy.  If the subclass
-     * doesn't have a HORIZONTAL content bias, then the width parameter can be
-     * ignored.
+     * Computes the minimum height of this region for the specified width.
+     * <p>
+     * This method supplies the value used by {@link #minHeight(double)} when the {@link #minHeight}
+     * property is set to {@link #USE_COMPUTED_SIZE}.
+     * <p>
+     * Subclasses should override this method when their content or layout policy requires a different
+     * minimum height. If {@link #getContentBias()} is {@link Orientation#HORIZONTAL}, the computation should
+     * use {@code width}; otherwise {@code width} can be ignored. An overriding implementation should be
+     * consistent with its {@link #layoutChildren()} implementation, including its pixel-snapping decisions.
      *
-     * @return the computed minimum height for this region
+     * @implNote The default implementation returns the sum of the top and bottom {@linkplain #getInsets() insets}.
+     * @param width the width on which to base the minimum height, or {@code -1} if no width is specified
+     * @return the computed minimum height
+     * @see <a href="package-summary.html#pixel-snapping">Pixel Snapping</a>
      */
     @Override protected double computeMinHeight(double width) {
         return getInsets().getTop() + getInsets().getBottom();
     }
 
     /**
-     * Computes the preferred width of this region for the given height.
-     * Region subclasses should override this method to return an appropriate
-     * value based on their content and layout strategy.  If the subclass
-     * doesn't have a VERTICAL content bias, then the height parameter can be
-     * ignored.
+     * Computes the preferred width of this region for the specified height.
+     * <p>
+     * This method supplies the value used by {@link #prefWidth(double)} when the {@link #prefWidth}
+     * property is set to {@link #USE_COMPUTED_SIZE}.
+     * <p>
+     * Subclasses that implement a custom layout policy should override this method to compute the width
+     * needed by that policy. If {@link #getContentBias()} is {@link Orientation#VERTICAL}, the computation
+     * should use {@code height}; otherwise {@code height} can be ignored. An overriding implementation should
+     * be consistent with its {@link #layoutChildren()} implementation, including its pixel-snapping decisions.
      *
-     * @return the computed preferred width for this region
+     * @implNote The default implementation adds the left and right {@linkplain #getInsets() insets} to the preferred
+     *           width computed by the {@link Parent#computePrefWidth(double) superclass implementation}.
+     * @param height the height on which to base the preferred width, or {@code -1} if no height is specified
+     * @return the computed preferred width
+     * @see <a href="package-summary.html#pixel-snapping">Pixel Snapping</a>
      */
     @Override protected double computePrefWidth(double height) {
         final double w = super.computePrefWidth(height);
@@ -1633,13 +1653,21 @@ public class Region extends Parent {
     }
 
     /**
-     * Computes the preferred height of this region for the given width;
-     * Region subclasses should override this method to return an appropriate
-     * value based on their content and layout strategy.  If the subclass
-     * doesn't have a HORIZONTAL content bias, then the width parameter can be
-     * ignored.
+     * Computes the preferred height of this region for the specified width.
+     * <p>
+     * This method supplies the value used by {@link #prefHeight(double)} when the {@link #prefHeight}
+     * property is set to {@link #USE_COMPUTED_SIZE}.
+     * <p>
+     * Subclasses that implement a custom layout policy should override this method to compute the height
+     * needed by that policy. If {@link #getContentBias()} is {@link Orientation#HORIZONTAL}, the computation
+     * should use {@code width}; otherwise {@code width} can be ignored. An overriding implementation should
+     * be consistent with its {@link #layoutChildren()} implementation, including its pixel-snapping decisions.
      *
-     * @return the computed preferred height for this region
+     * @implNote The default implementation adds the top and bottom {@linkplain #getInsets() insets} to the preferred
+     *           height computed by the {@link Parent#computePrefHeight(double) superclass implementation}.
+     * @param width the width on which to base the preferred height, or {@code -1} if no width is specified
+     * @return the computed preferred height
+     * @see <a href="package-summary.html#pixel-snapping">Pixel Snapping</a>
      */
     @Override protected double computePrefHeight(double width) {
         final double h = super.computePrefHeight(width);
@@ -1647,32 +1675,42 @@ public class Region extends Parent {
     }
 
     /**
-     * Computes the maximum width for this region.
-     * Returns Double.MAX_VALUE by default.
-     * Region subclasses may override this method to return an different
-     * value based on their content and layout strategy.  If the subclass
-     * doesn't have a VERTICAL content bias, then the height parameter can be
-     * ignored.
+     * Computes the maximum width of this region for the specified height.
+     * <p>
+     * This method supplies the value used by {@link #maxWidth(double)} when the {@link #maxWidth}
+     * property is set to {@link #USE_COMPUTED_SIZE}.
+     * <p>
+     * Subclasses may override this method to impose an upper bound based on their content or layout policy.
+     * If {@link #getContentBias()} is {@link Orientation#VERTICAL}, the computation should use {@code height};
+     * otherwise {@code height} can be ignored. A finite maximum derived from layout geometry should be consistent
+     * with the calculations and pixel-snapping decisions used by {@link #layoutChildren()}.
      *
-     * @param height The height of the Region, in case this value might dictate
-     * the maximum width
-     * @return the computed maximum width for this region
+     * @implNote The default implementation returns {@link Double#MAX_VALUE}, indicating that
+     *           the region has no finite maximum width.
+     * @param height the height on which to base the maximum width, or {@code -1} if no height is specified
+     * @return the computed maximum width; {@link Double#MAX_VALUE} indicates no finite maximum
+     * @see <a href="package-summary.html#pixel-snapping">Pixel Snapping</a>
      */
     protected double computeMaxWidth(double height) {
         return Double.MAX_VALUE;
     }
 
     /**
-     * Computes the maximum height of this region.
-     * Returns Double.MAX_VALUE by default.
-     * Region subclasses may override this method to return a different
-     * value based on their content and layout strategy.  If the subclass
-     * doesn't have a HORIZONTAL content bias, then the width parameter can be
-     * ignored.
+     * Computes the maximum height of this region for the specified width.
+     * <p>
+     * This method supplies the value used by {@link #maxHeight(double)} when the {@link #maxHeight}
+     * property is set to {@link #USE_COMPUTED_SIZE}.
+     * <p>
+     * Subclasses may override this method to impose an upper bound based on their content or layout policy.
+     * If {@link #getContentBias()} is {@link Orientation#HORIZONTAL}, the computation should use {@code width};
+     * otherwise {@code width} can be ignored. A finite maximum derived from layout geometry should be consistent
+     * with the calculations and pixel-snapping decisions used by {@link #layoutChildren()}.
      *
-     * @param width The width of the Region, in case this value might dictate
-     * the maximum height
-     * @return the computed maximum height for this region
+     * @implNote The default implementation returns {@link Double#MAX_VALUE}, indicating that
+     *           the region has no finite maximum height.
+     * @param width the width on which to base the maximum height, or {@code -1} if no width is specified
+     * @return the computed maximum height; {@link Double#MAX_VALUE} indicates no finite maximum
+     * @see <a href="package-summary.html#pixel-snapping">Pixel Snapping</a>
      */
     protected double computeMaxHeight(double width) {
         return Double.MAX_VALUE;
--- a/modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
+++ b/modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
@@ -303,7 +303,6 @@
  *     <li>A position is rounded to the nearest pixel so that an edge stays close to its requested coordinate.
  *     <li>Sizes use ceiling so that snapping itself does not reduce the measured content allocation.
  *     <li>Empty space can usually become slightly smaller without losing content.
- *     <li>{@code snapSpaceX/Y} can also be used to remove floating-point drift from the result of a computation.
  * </ul>
  *
  * The {@linkplain javafx.scene.layout.Region#snappedTopInset() snapped inset} methods return the combined border
@@ -344,8 +343,8 @@
  * combined sum of such elements, each semantic role can require a different decision:
  * <ul>
  *     <li>Snapping too early can lose precision, while snapping too late can introduce errors.
- *     <li>Snapping several children independently can avoid clipping, but at the same time risk exceeding
- *         the content size defined by the region.
+ *     <li>Snapping several children independently can avoid clipping, but the region's measurement and layout
+ *         methods must use consistent calculations so that the computed content size includes those allocations.
  * </ul>
  *
  * Additionally, adding or subtracting already-snapped values can introduce floating-point drift that may cause a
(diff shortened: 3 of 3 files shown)


modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -217,5 +217,303 @@
...
+ *
+ * <h3>Why snapping is difficult</h3>
+ *
+ * There is no <em>single</em> correct direction in which to round a value, and no <em>single</em> place in a layout
+ * algorithm where snapping must occur. Since a value can represent a position, content size, empty space, or a
+ * combined sum of such elements, each semantic role can require a different decision:
+ * <ul>
+ *     <li>Snapping too early can lose precision, while snapping too late can introduce errors.
+ *     <li>Snapping several children independently can avoid clipping, but at the same time risk exceeding
+ *         the content size defined by the region.
```
hjohn: There should be no risk, as the compute results should take snapping into account; this can basically only occur if the calculation used by compute methods is different from the one used during layoutChildren.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -217,5 +217,303 @@
...
+ *         In the correct example, {@code snapSizeX} is used to create each child's size allocation independently.
+ *         Then, both size allocations are added and {@code snapSpaceX} removes floating-point drift.
+ *         The same rule applies to distinct margins and gaps: snap each one with {@code snapSpaceX/Y}, then re-snap
+ *         the final sum with {@code snapSpaceX/Y} again. For more information, refer to <em>Re-snap after
+ *         calculations, using the meaning of the result</em>.
+ *         <p>
+ *         Note that this only applies to <em>independent</em> allocations. If several children must fit within a
+ *         fixed allocated space, the algorithm must consider all children together and coordinate rounding children
+ *         up or down so that their sum does not exceed the allocated space.
+ *     <li><b>Do not repeatedly ceil the same semantic size.</b><br>
```
hjohn: I think this applies to all of them (you specifically mentioned `ceil`), but it is especially dangerous with the ones that use `ceil` (although I can think we can mitigate this with a small adjustment to the `ceil` code used if we're willing to be only accurate to say 1 millionth of a pixel).

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -217,5 +217,303 @@
...
+ *         double width = snapSizeX(measuredWidth + measurementAdjustment);
+ *
+ *         // Incorrect: snapping early throws away information and the second snapSizeX can add another pixel
+ *         double width = snapSizeX(snapSizeX(measuredWidth) + measurementAdjustment);
+ *         }
+ *         At scale {@code 1.0}, if both values are {@code 0.2}, the first form returns {@code 1.0}, while the second
+ *         (incorrect) form returns {@code 2.0}. This rule does not conflict with the preceding rule: two independent
+ *         pieces of content are two allocations, while two intermediate terms describing one piece of content are
+ *         one allocation.
+ *     <li><b>Re-snap after calculations, using the meaning of the result.</b><br>
```
hjohn: I think we may need to check if this actually helps (it sometimes definitely will), even though I've saying this as well there is still a problem:

The `snapSize` operation first applies its own floating point calculation (multiplying by render scale which can already introduce a tiny error) before calling `Math.ceil`.  So even a correctly snapped end result (say 0.66...667) multiplied by the renderscale 1.5 can become 1.00...001 which is then ceil'd to 2.

This happens because the result of a child's `compute`, even if snapped, gets used by the parent that treats it as content (using `snapSizeX/Y` ceiling).

This is why I'm now of the opinion that we should subtract a constant value before ceiling, so the operation becomes:

    ceil((v * renderScale) - epsilon) / renderScale

Where the epsilon is set to 1 millionth of a logical pixel (1e-6).

The reasoning to use 1 millionth is:
- too large a value may become noticable (ie. 1/10th of a pixel may introduce slight blurriness)
- too small a value may not absorb floating point errors that have been multiplied (a spacing * number of children), are using values of fairly high magnitudes (a 100000 pixel screen or group of screens) or are using a fairly high renderscale
- 1 millionth of a pixel is still unobservable and it is fair to say that a value within 1 millionth of a pixel can be considered to be *that* pixel

Why not `ulp`?
- It doesn't help correct the result when the calculations have accumulated more than 1 floating point error
- It works poorly when the value to snap value is `Double.MAX_VALUE` (or something equally large) which is used through-out FX (an `ulp` at that magnitude is like 1e292 pixels) -- `ceiling` a `Double.MAX_VALUE` should yield `Double.MAX_VALUE` not `Double.MAX_VALUE - 1e292`).

So resnapping may only be needed after significant number of calculations have been done with values that were snapped originally; tiny floating point errors should be absorbed by the parent's snapping (after `ceil` has been fixed) or by the rendering hardware (usually only accurate up to `float` precision).

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -217,5 +217,303 @@
...
+ *         policy of its parent), it must not reposition or resize itself. If the allocated width or height is not
+ *         aligned, the region cannot both preserve the exact allocation and make its complete bounds pixel-aligned.
+ *         Apply the snapping policy of the region deliberately:
+ *         {@snippet :
+ *         double rawAvailableWidth = getWidth();
+ *         double rawAvailableHeight = getHeight();
+ *
+ *         // Depending on this region's fitting and overflow policy, snap this region's
+ *         // raw position and raw allocated size to lay out its children.
+ *         }
```
hjohn: This is a bit unclear to me, the snippet does not include any snapping?

hjohn (commented):
Thanks a lot for documenting this!

I think you highlighted many good points, and I agree with all of them.

I've added a small discussion around the re-snapping (which I think doesn't solve all issue without an adjustment to our `snapSize` logic) but is about as accurate as it can be with the current state.

That in itself shouldn't stop us from adding these docs, so will approve once the other small points have been discussed/resolved.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -217,5 +217,303 @@
...
+ *         double total = snapSpaceX(snapSizeX(firstWidth) + snapSizeX(secondWidth)); // 2.0
+ *
+ *         // Incorrect: 0.4 + 0.4 is treated as one content allocation
+ *         double total = snapSizeX(firstWidth + secondWidth); // 1.0
+ *         }
+ *         In the correct example, {@code snapSizeX} is used to create each child's size allocation independently.
+ *         Then, both size allocations are added and {@code snapSpaceX} removes floating-point drift.
+ *         The same rule applies to distinct margins and gaps: snap each one with {@code snapSpaceX/Y}, then re-snap
+ *         the final sum with {@code snapSpaceX/Y} again. For more information, refer to <em>Re-snap after
+ *         calculations, using the meaning of the result</em>.
```
nlisker: This referred-to section hasn't appeared yet.
```suggestion
 *         the final sum with {@code snapSpaceX/Y} again. For more information, refer to <em>Re-snap after
 *         calculations, using the meaning of the result</em> below.
```

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -217,5 +217,303 @@
...
+ *         Positions and spaces use nearest-pixel rounding; content sizes use ceiling.
+ *         This distinction prevents gaps from growing unnecessarily and content from being clipped:
+ *         {@snippet :
+ *         double x = snapPositionX(rawX);       // coordinate
+ *         double gap = snapSpaceX(rawGap);      // empty space
+ *         double width = snapSizeX(rawWidth);   // content size
+ *
+ *         // Incorrect: a 0.1-unit gap becomes a full pixel
+ *         double gap = snapSizeX(0.1);
+ *         }
```
nlisker: Just below there's an example,
```
// Correct: each content item receives an independent allocation
double total = snapSpaceX(snapSizeX(firstWidth) + snapSizeX(secondWidth)); // 2.0
 ```
that snaps a size (sum of snapped widths) with `snapSpace`. That's confusing. It looks like `snapSpace` is for more than just empty spaces.

File: modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java [hunk 7]
```diff
@@ -217,5 +217,303 @@
...
+ *         double firstWidth = snapSizeX(firstChildWidth);
+ *         double gapWidth = snapSpaceX(getGap());
+ *         double secondWidth = snapSizeX(secondChildWidth);
+ *         double allocatedWidth = snapSpaceX(firstWidth + gapWidth + secondWidth);
+ *
+ *         // Incorrect: arithmetic can add floating-point noise to snapped values
+ *         double allocatedWidth = firstWidth + gapWidth + secondWidth;
+ *
+ *         // Incorrect: snapSizeX can turn floating-point noise into an extra pixel
+ *         double allocatedWidth = snapSizeX(firstWidth + gapWidth + secondWidth);
```
nlisker: The correct example uses `snapSpaceX` (after snapping individually), but this incorrect example uses `snapSizeX` without snapping individually. If the idea is to demonstrate re-snapping, then the same method should be used, otherwise it introduces another incorrectness.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
nlisker: Additionally, I think this "for more information" part appears too late. I'd think it belongs in the section above about classification since it looks like a direct contradiction to what it wrote (sizes are snapped with `space` suddenly).

The rules about snapping a sum of snapped numbers it split between sections. For sizes, it is explained both here and in *Re-snapping* (use `snapSpace`), gaps/spaces are explained here (use `snapSpace`), and coordinates are explained in *Re-snapping* (use `snapPosition`):
```java
 * The final {@code snapSpaceX} does not mean that the allocated width is empty space, it merely uses
 * the snapping method to remove a tiny amount of floating-point drift.
 * <p>
 * The same principle applies to coordinates: after calculating a final coordinate from snapped values, use
 * {@code snapPositionX/Y} to remove potential floating-point drift.
```
I'd put all 3 of them in one place before they are used and avoid the repetition.

I'll note that it was mentioned previously in *Choosing the snapping operation* with
```
<li>{@code snapSpaceX/Y} can also be used to remove floating-point drift from the result of a computation.
```
so the point about removing floating-point drift is now in 3 places.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -217,5 +217,303 @@
...
+ *         the addition of these terms produces the final allocated width, but it can introduce floating-point error.
+ *         The final {@code snapSpaceX} does not mean that the allocated width is empty space, it merely uses
+ *         the snapping method to remove a tiny amount of floating-point drift.
+ *         <p>
+ *         The same principle applies to coordinates: after calculating a final coordinate from snapped values, use
+ *         {@code snapPositionX/Y} to remove potential floating-point drift.
+ *     <li><b>Deliberately apply the snapping policy to values determined by the parent.</b><br>
+ *         Since a region's position and allocated size are determined by its parent (and the {@code isSnapToPixel}
+ *         policy of its parent), it must not reposition or resize itself. If the allocated width or height is not
+ *         aligned, the region cannot both preserve the exact allocation and make its complete bounds pixel-aligned.
```
nlisker: "...is not aligned" with what? pixel-aligned?

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -217,5 +217,303 @@
...
+ *             child.maxHeight(snappedChildWidth));
+ *
+ *         // Snap the height the child will receive
+ *         double snappedChildHeight = snapSizeY(rawChildHeight);
+ *
+ *         // Incorrect: height is measured for rawChildWidth
+ *         double snappedChildHeight = snapSizeY(boundedSize(
+ *             child.prefHeight(rawChildWidth),
+ *             child.minHeight(rawChildWidth),
+ *             child.maxHeight(rawChildWidth)));
```
nlisker: The incorrect example combines the 2 calculations above it, but the mistake is only in the first one. Combining them hides the difference. I would write for the incorrect example:
```java
double rawChildHeight = boundedSize(
    child.prefHeight(rawChildWidth),
    child.minHeight(rawChildWidth),
    child.maxHeight(rawChildWidth)));
```
for a direct comparison.

nlisker (commented):
A very nice guide. I've left some comments on readability.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
mstr2: I've expanded the sample to allocate a single, resizable child.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
mstr2: I've reordered the bullet points, so that arithmetic drift is explained first. I've also reworded the other bullet points a bit.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
mstr2: I'm not sure what you mean here. The individual components are always snapped in this example, the mistake is using `snapSizeX` instead of `snapSpaceX`.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
mstr2: I've reworded this a bit.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -217,5 +217,310 @@
  * mode, a mirroring transform is automatically applied to nodes, flipping the visual flow in the horizontal
  * direction. If this behavior is not desired, nodes can override {@link javafx.scene.Node#usesMirroring()}
  * and return {@code false}.
+ *
+ * <h2>Pixel Snapping</h2>
+ *
+ * In JavaFX, layout coordinates use {@code double} values, which means that a position or size can be fractional.
```
andy-goryachev-oracle, line 223: I think it's the other way around: JavaFX supports fractional scale (example: 125%) which means the logical coordinates cannot be integers.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -217,5 +217,310 @@
...
+ * Layout is expressed in logical coordinates, and the {@linkplain javafx.stage.Window#getRenderScaleX() horizontal}
+ * and {@linkplain javafx.stage.Window#getRenderScaleY() vertical} render scales of a window convert logical units
+ * to pixels in the window's rendering buffer:
+ *
+ * <pre>{@code
+ *     physical pixels = logical units * render scale
+ * }</pre>
+ *
+ * At a render scale of {@code 1.0}, one logical unit is one pixel. At a render scale of {@code 1.5}, one pixel is
+ * {@code 1 / 1.5} ≈ {@code 0.6667}, logical units. A correctly snapped value is therefore not necessarily an integer,
```
andy-goryachev-oracle: 0.6666... mathematically, or 0.6666666666666666 in IEEE 754, or ~0.6667

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -217,5 +217,310 @@
...
+ * In this case, even if the child snaps its descendants in its local coordinates, the entire child subtree remains
+ * shifted by half a pixel. Conversely, a parent can allocate a child pixel-aligned outer bounds even when that child
+ * has disabled snapping for its own layout.
+ * <p>
+ * Transforms are applied after the layout has been computed. For example, at render scale {@code 1.0},
+ * the following translation undoes the alignment established during layout:
+ * {@snippet :
+ * child.relocate(snapPositionX(10), snapPositionY(20));
+ * child.setTranslateX(0.5); // The rendered X coordinate is now between pixels
+ * }
```
nlisker, line 338: Is there some conclusion or advice to be drawn from here? I feel it leaves the reader hanging a bit.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
nlisker: I misplaced the comment. It was supposed to be at the example in *Snap independent allocations independently*,
```
 *         // Correct: each content item receives an independent allocation
 *         double total = snapSpaceX(snapSizeX(firstWidth) + snapSizeX(secondWidth)); // 2.0
 *
 *         // Incorrect: 0.4 + 0.4 is treated as one content allocation
 *         double total = snapSizeX(firstWidth + secondWidth); // 1.0
```
which showed both a different snapping method and a difference in re-snapping. You've changed the incorrect example to use `snapSpaceX`, so now the incorrect part is clear.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
mstr2: I've reframed the section around "why snapped content can still look blurry", and the conclusion is that if that happens, check the ancestors and transforms of the snapped content.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
mstr2: You probably mean that if the layout falls on physical pixel boundaries, then the logical coordinates might not be integers. That's right, but even more fundamentally, just by having coordinates be `double`, I can assign any non-integer value.

andy-goryachev-oracle: I mean cause-and-effect, especially at the start of the section.

The reason we have double is because of fractional scale, the rest is derived from that fact.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -217,5 +217,314 @@
...
+ *     <li>Empty space can usually become slightly smaller without losing content.
+ * </ul>
+ *
+ * The {@linkplain javafx.scene.layout.Region#snappedTopInset() snapped inset} methods return the combined border
+ * inset and padding; custom controls should normally use these methods instead of snapping
+ * {@linkplain javafx.scene.layout.Region#getInsets() raw insets} manually.
+ *
+ * <h3>Why snapped content can still look blurry</h3>
+ *
+ * Enabling pixel snapping on a region does not guarantee that the region or its descendants will be rendered on
```
andy-goryachev-oracle, line 314: The way I see it (I could be wrong), ideally all layout operations should use snapped coordinates, that is, be performed using display pixels.  Assuming, of course, the entire parent chain is snapped.

The only downside of this method is more calls to snap* methods, but it seems to be the only method to guarantee correct sizing.

andy-goryachev-oracle (commented):
I like the explanations very much (modulo minor comments).

One question - should we add backreferences to this new section from relevant methods in `Region` and `SkinBase`?

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
mstr2: I think that's too strong a claim. Even if we didn't have different render scales, there would still be a need to have sub-pixel positioning (for example, for smooth animation).

andy-goryachev-oracle: hmm, good point.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
mstr2: I've thought about this quite a bit, and here's how I understand it.

1. When we re-snap after calculations, we only use `snapPosition`/`snapSpace`, never `snapSize`.

   The proposed text already says that. The reason for this is that the terms of the calculation were already snapped to begin with, so the _idealized_ result is guaranteed to be pixel-aligned. This is a critical piece of information that we, as the authors of the calculation, know: that any deviation from the pixel grid is not genuine, it's merely an artifact of floating-point arithmetic. We know for a fact that we can safely discard it, and `snapSpace` is guaranteed to give us the correctly snapped result as long as float drift does not exceed 0.5 pixels (which is a safe assumption).

2. You point out that the parent can use this snapped size, and then use it in its own calculations by `snapSize`-ing it again.

   However, isn't idempotency in this situation what the current `ScaledMath` implementation already guarantees? Granted, it only explicitly guarantees (for pixel-aligned values less than 10^15):
   ```
   ceil(ceil(value, scale), scale) == ceil(value, scale)
   ```
   It doesn't explicitly guarantee cross-operation idempotency like in our scenario. But since the values returned from `ScaledMath` functions are all pixel-aligned, I think the guarantee can be extended across operations.

   If this is true, then your example (0.666... with scale 1.5) won't ever ceil to 2 for any realistic render scale.

3. Fundamentally, when we apply such a correction in `ScaledMath`, we've lost important information: whether the tiny excess is genuine, or whether it is a result of arithmetic drift. Only the caller of the scaled rounding methods knows this. The difference between the two is not meaningless, in the end it is the difference between a pixel added or removed. And that is, in turn, the difference between a layout staying as it is, or reflowing to the next line.

It is true that your idea improves perturbation stability of sloppily-snapped calculations, but I think that it doesn't improve the result if the snapping was done correctly to begin with (that is, discard any arithmetic excess if you know that your calculation was done on pixel-aligned values).

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -217,5 +217,314 @@
...
+ * Applications should use the snapping methods on {@code Region} instead of implementing the formula above or
+ * caching its result. The built-in methods use the appropriate current render scale and account for floating-point
+ * error near pixel boundaries.
+ *
+ * <h3>Choosing the snapping operation</h3>
+ *
+ * {@code Region} provides three pairs of axis-specific snapping methods. Their names describe the semantic role of
+ * the value being snapped, which influences the snapping direction:
+ *
+ * <table border="1">
```
kevinrushforth: KCR TODO: check a11y for table.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
andy-goryachev-oracle: > the terms of the calculation were already snapped to begin with, so the idealized result is guaranteed to be pixel-aligned

Can we emphasize this please?

mstr2: > Can we emphasize this please?

I've changed this paragraph to emphasize it.

mstr2:
> One question - should we add backreferences to this new section from relevant methods in `Region` and `SkinBase`?

I've updated the documentation of the layout methods in `Parent` and `Region` and included a `@see` reference to the pixel snapping documentation.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
hjohn: There's two things here:

1. Using `ScaledMath.ceil` where it subtracts the `ulp` works incorrectly when inputs are so large that they don't have a significant fractional part anymore (all digits are allocated to the integer part) -- this causes `Double.MAX_VALUE` to get mangled which shows up in unit tests that can't just use `== Double.MAX_VALUE` -- the rounding functions have no such problems
2. The current `ceil` function forces absolute correct use of snapping, so that even with pre-snapped values `left + size + right` must be snapped again. With a more lenient `ceil` this final snapping becomes unnecessary as the total error introduced with pre-snapped values will never become significant enough that we can't `ceil` to the **intended pixel** if we were using a constant instead of the `ulp`.

## Example for situation 1:
### Using ULP
Initial value | ULP | ceil(initial − ULP) | Difference
---|---|---|---
1.797693134862**3157**E308 | 1.9958403095347198E292 | 1.797693134862**3155**E308 | 2E292
2.0E20 | 3.2768E4 | 1.9999999999999997E20 | 30000
2.0E10 | 3.814697265625E-6 | 2.0E10 | 0
2.0E5 | 2.9103830456733704E-11 | 2.0E5 | 0
2.0E0 | 4.440892098500626E-16 | 2.0E0 | 0

### Using tiny fractional constant
Initial value | Constant | ceil(initial − Constant) | Difference
---|---|---|---
1.7976931348623157E308 | 1.0E-6 | 1.7976931348623157E308 | 0
2.0E20 | 1.0E-6 | 2.0E20 | 0
2.0E10 | 1.0E-6 | 2.0E10 | 0
2.0E5 | 1.0E-6 | 2.0E5 | 0
2.0E0 | 1.0E-6 | 2.0E0 | 0

mstr2: We've still lost the information to decide whether 10.0000001 is a tiny, but intended offset that should round to 11, or whether it is noise and the result should round to 10.

But let's assume we make the ceil function a bit fuzzy, so that it tolerates arithmetic on snapped values where the final result is not snapped again. That doesn't make the snapping API a good idea, or even just remotely useful for custom controls, because even with the snap-the-result rule gone, there are still a lot of other ways in which developers will definitely get snapping wrong.

I think the only way how third-party developers can create custom layouts that actually work is to offer an API to compose them from well-tested primitives, like your virtual layout container proposal.

hjohn: > It is true that your idea improves perturbation stability of sloppily-snapped calculations, but I think that it doesn't improve the result if the snapping was done correctly to begin with (that is, discard any arithmetic excess if you know that your calculation was done on pixel-aligned values)

Yeah, but that does set a very high bar. Even a simple addition like `return left + size + right` is wrong even when all values were snapped here. What I'm suggesting is that with a slightly less "edgy" ceil function, you only need one rule: 

- snap all input values before use

If we didn't have `ceil` anywhere, but all functions used rounding, then that would have been sufficient. And that seems to have been the thought process behind the implementation of almost all layout containers (they snap their inputs, then calculate without further snapping).

However `snapSize`, which intent is to **snap a content-size to a size so all of it is visible**, was implemented in a strict mathematical sense that is sensitive to tiny floating point errors; in reality it was intended to ensure all content pixels were visible; do we really need to ensure that an overflow of 0.00000000000002 is "visible" ? That extra pixel allocated there will be blank (even with anti-aliasing).

So I agree that `ceil` as implemented now is correct in the mathematical sense, but that's not what we're trying to achieve. We're trying to achieve correct snapping for content sizes, not ceiling.

hjohn: It looks like my 2nd reply (sorry) cross with this one.

> We've still lost the information to decide whether 10.0000001 is a tiny, but intended offset that should round to 11, or whether it is noise and the result should round to 10.

The point I tried to make in the reply above was what exactly does a content size of `10.0000001` mean?  Is that 0.0000001 extra pixel going to be visible and relevant that it must be rendered?  

> But let's assume we make the ceil function a bit fuzzy, so that it tolerates arithmetic on snapped values where the final result is not snapped again. That doesn't make the snapping API a good idea, or even just remotely useful for custom controls, because even with the snap-the-result rule gone, there are still a lot of other ways in which developers will definitely get snapping wrong.

I lost you at the point where you went to "That doesn't make the snapping API a good idea..." -- what API?

I'm not criticizing this documentation update; I think it's great. I am merely asking ourselves the question: are we going to let floating point artifacts and the current implementation of a mathematically accurate `ceil` function dictate how a large part of snapping must work?  As we're now about to formalize this in API docs.

The original intent of `snapSize` is clear: to snap widths/heights of content areas that users interact with, in such a way that no information is lost visually.  Now I contend that it was never the intention of this function to ensure a potential 0.0000001 pixel overflow (or since we're using doubles, currently even much worse, a 1e-14 overflow) **must** be displayed as otherwise important information is lost.

Rounding 1 millionth of a pixel doesn't lose information.

I'm not saying we change `ScaledMath` here (as that defines a `ceil` function which is correct as is).  I am saying `snapSize` should not be using `ScaledMath.ceil` but something more relevant to the task at hand, where fractional pixels, too tiny to be relevant for the goal of displaying "content", have no bearing on the outcome.

I mean if my control says its width must preferably be 10.0000001 (which it is only allowed to return if unsnapped), even if intended, it would still be invisible if the parent choses to render this as 10 pixels.

Now to potentially poke a hole in my own argument: this does mean that if `computeMinWidth` returns `10.000001`, I'm allowed to interpret that as 10; just like we now allow a value that strictly exceeds `10` by 1 ulp (even if `10` can be exactly represented) to be interpreted as `10`.  I'm merely suggesting to extend this value (and `ulp` is still incorrect for other reasons, see `Double.MAX_VALUE`) by something that is more in tune with what we're dealing with here, finite device pixels.

> I think the only way how third-party developers can create custom layouts that actually work is to offer an API to compose them from well-tested primitives, like your virtual layout container proposal.

Any control author still must implement the `compute` functions; and most of the well tested primitives that they may want to use are package private on `Region` (and to be honest, they better remain that way as their API is too poor to expose).

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -217,5 +217,320 @@
...
+ * Layout is expressed in logical coordinates, and the {@linkplain javafx.stage.Window#getRenderScaleX() horizontal}
+ * and {@linkplain javafx.stage.Window#getRenderScaleY() vertical} render scales of a window convert logical units
+ * to pixels in the window's rendering buffer:
+ *
+ * <pre>{@code
+ *     physical pixels = logical units * render scale
+ * }</pre>
+ *
+ * At a render scale of {@code 1.0}, one logical unit is one pixel. At a render scale of {@code 1.5}, one pixel is
+ * {@code 1 / 1.5} ≈ {@code 0.6667}, logical units. A correctly snapped value is therefore not necessarily an integer,
```
Maran23: Is it just me or does the comma feels off here: `one pixel is ... 0.6667, logical units`

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -217,5 +217,320 @@
...
+ * Therefore, if content appears blurry despite pixel snapping being enabled, check both how its ancestors position
+ * and size it and whether transforms affect its final rendered position.
+ *
+ * <h3>Why snapping is difficult</h3>
+ *
+ * There is no <em>single</em> correct direction in which to round a value, and no <em>single</em> place in a layout
+ * algorithm where snapping must occur. Since a value can represent a position, content size, empty space, or a
+ * combined sum of such elements, each semantic role can require a different decision:
+ * <ul>
+ *     <li>Snapping too early can lose precision, while snapping too late can introduce errors.
```
Maran23, line 349: There is no closing `</li>` tag here and below. 
I'm not sure if this is really needed. But probably best practise to do so.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -217,5 +217,320 @@
...
+ * <ol>
+ *     <li><b>Use the axis-specific snapping method.</b><br>
+ *         A horizontal value uses the X scale and a vertical value uses the Y scale.
+ *         This matters when those scales differ:
+ *         {@snippet :
+ *         double left = snapSpaceX(margin.getLeft()); // Correct
+ *         double top = snapSpaceY(margin.getTop());   // Correct
+ *
+ *         double top = snapSpaceX(margin.getTop());   // Incorrect axis
+ *         double gap = snapSpace(rawGap);             // Deprecated and ambiguous
```
Maran23, line 370: I wonder if we should remove those deprecated methods at one point. They are deprecated since JavaFX 9, so years ago.

We should probably start by marking them with `forRemoval = true`. (Follow-up)

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -217,5 +217,320 @@
...
+ *         // Correct: coordinate calculated from snapped values is re-snapped as a position
+ *         double childX = snapPositionX(snappedLeftInset() + allocatedWidth);
+ *
+ *         // Incorrect: snapSizeX can turn floating-point noise into an extra pixel
+ *         double allocatedWidth = snapSizeX(firstWidth + gapWidth + secondWidth);
+ *
+ *         // Incorrect: the sum of snapped values can add floating-point drift
+ *         double allocatedWidth = firstWidth + gapWidth + secondWidth;
+ *         }
+ *     <li><b>Snap independent allocations independently.</b><br>
```
Maran23, line 427: This is very helpful and a nice example!

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
mstr2: > I lost you at the point where you went to "That doesn't make the snapping API a good idea..." -- what API?

What I meant was that the current API (overriding measurement any layout methods, combined with manual snapping) is not very useful to create custom layouts. Most developers won't have an intuition of the intricate rules that is strong enough to pull it off without making mistakes (as can be clearly seen in JavaFX itself), and the failure modes of defective code are often not easily visible.

Aside from that, I think that you've convinced me that "fuzzy snapping" is probably a good idea. However, not as something we should do in JavaFX itself, and neither as something we should recommend. The problem that I still have with it is that it allows off-grid values to be exposed in the scene graph; the correction is only applied when those values are consumed by the layout system itself. It does not prevent non-layout-related code from seeing the off-grid values. I can't imagine that this will be a problem for any reasonable application, but it still makes it feel a bit "ugly".

I'd like to see built-in layout containers continue to expose canonical grid-aligned values. This will account for the majority of layout activity in any given JavaFX application. Fuzzy snapping would then be more like a fallback system to have custom third-party layouts (which are probably going to have defects) behave more gracefully.

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
```diff
@@ -1591,88 +1591,126 @@ public final BooleanProperty cacheShapeProperty() {
...
+     * consistent with its {@link #layoutChildren()} implementation, including its pixel-snapping decisions.
      *
-     * @return the computed minimum width of this region
+     * @implNote The default implementation returns the sum of the left and right {@linkplain #getInsets() insets}.
+     * @param height the height on which to base the minimum width, or {@code -1} if no height is specified
+     * @return the computed minimum width
+     * @see <a href="package-summary.html#pixel-snapping">Pixel Snapping</a>
      */
     @Override protected double computeMinWidth(double height) {
         return getInsets().getLeft() + getInsets().getRight();
```
Maran23, line 1610: not relevant for this PR, but should this be snapped in the future? Or do you think this is better as default?

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
andy-goryachev-oracle: yep, comma is not needed here.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -391,29 +394,58 @@
...
+ *         double gapWidth = snapSpaceX(getGap());
+ *         double secondWidth = snapSizeX(secondChildWidth);
+ *
+ *         // Correct: re-snap the final allocated span with snapSpaceX after arithmetic
+ *         double allocatedWidth = snapSpaceX(firstWidth + gapWidth + secondWidth);
+ *
+ *         // Correct: coordinate calculated from snapped values is re-snapped as a position
+ *         double childX = snapPositionX(snappedLeftInset() + allocatedWidth);
+ *
+ *         // Incorrect: snapSizeX can turn floating-point noise into an extra pixel
```
andy-goryachev-oracle, line 421: I am wondering if we should remove incorrect examples (there are many more ways to produce an incorrect result).

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
```diff
@@ -217,5 +217,320 @@
...
+ *
+ * <ol>
+ *     <li><b>Use the axis-specific snapping method.</b><br>
+ *         A horizontal value uses the X scale and a vertical value uses the Y scale.
+ *         This matters when those scales differ:
+ *         {@snippet :
+ *         double left = snapSpaceX(margin.getLeft()); // Correct
+ *         double top = snapSpaceY(margin.getTop());   // Correct
+ *
+ *         double top = snapSpaceX(margin.getTop());   // Incorrect axis
```
andy-goryachev-oracle, line 369: wonder if we should remove incorrect examples here as well.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
Maran23: I found them really helpful. This is one of that things where negative examples really help to understand the full picture IMO.

modules/javafx.graphics/src/main/java/javafx/scene/layout/package-info.java
mstr2: It's not needed, and we almost always omit it in javadocs. At least that's my impression for the past few years.

modules/javafx.graphics/src/main/java/javafx/scene/layout/Region.java
mstr2: Yes, it should be snapped.

Maran23:
One last question from my side: There is also this weird method which is only used once: `snapPortionX/Y` (only used in `HBox` / `VBox`).
Is this something we need, or will this be eventually cleaned up?

mstr2:
> One last question from my side: There is also this weird method which is only used once: `snapPortionX/Y` (only used in `HBox` / `VBox`). Is this something we need, or will this be eventually cleaned up?

This one rounds the value towards zero: 2.6 rounds to 2, and -2.6 rounds to -2. It is marginally useful for HBox/VBox for coordinated growing or shrinking. Whether it should live on Region... doubtful.

andy-goryachev-oracle (approved):
excellent work!

Maran23 (approved):
Nice documentation and guide. All my questions/points are resolved, so approving.

modules/javafx.graphics/src/main/java/javafx/scene/Parent.java
```diff
@@ -1148,14 +1148,19 @@ void clearSizeCache() {
 
     // PENDING_DOC_REVIEW
```
kevinrushforth: Minor: I think this can be removed now.

kevinrushforth (approved):
This looks great. Very clear and understandable. I left one minor suggestion. Will reapprove if you make the change.

Maran23 (approved):

hjohn (approved):
Looks good

kevinrushforth (approved):

andy-goryachev-oracle (approved):

